PR #3247, opened 2019-07-25
Avoid crash on Android 4 by getting a vector icon from the local context instead of the application context.

### LINKED ISSUE #3246: App crashes when Geo question tries to open Mapbox on Android 4.x
## Software and hardware versions 
Collect v1.23 - master
Android v4.x

#### Problem description
App crashes when the user opens some Goe question when Mapbox is set

1. The user goes to `General Settings` > `Maps`
2. User sets Mapbox
3. User opens GeoPoint / GeoTrace or GeoShape question
4. User clicks on the button to open map
5. App crashes 

Screenshot from 2019-07-25 12-53-22

#### Expected behavior
Collect does not crash

#### Other information 
Probably missed during testing #3164

zestyping:
Issues: Closes #3246.
Scope: MapboxMapFragment
Requested reviewers: @lognaturel 

#### User-visible changes

A crash when opening a Mapbox map on Android 4.x is avoided.

#### Internal changes

I was able to replicate and confirm this crash in an Android 4.x emulator.  I was puzzled by this, and don't understand why it's happening.

First, I tried switching to `VectorDrawableCompat.create()` as recommended by [this Stack Overflow post](https://stackoverflow.com/questions/48285547/resourcesnotfoundexception-for-xml-vector-drawable-on-api-19).  This did avoid the crash.

But then I went looking and found that there are many places where we call `ContextCompat.getDrawable` to get a vector icon (e.g. `FileArrayAdapter`, `FormHierarchyActivity`), and when I exercised those code paths in my Android 4.4 emulator, they did not crash.  There is even a similar call in `OsmDroidMapFragment` and it does not crash.

Upon closer inspection I found that these other calls did not use `getApplicationContext()`, but instead got a context from a local Activity or View.  So I tried switching `MapboxMapFragment` to also use the fragment's context instead of the application context, and that seems to work.

I don't understand why it works, though.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
@@ -45,7 +45,6 @@ import com.mapbox.mapboxsdk.utils.ColorUtils;
 
 import org.odk.collect.android.BuildConfig;
 import org.odk.collect.android.R;
-import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.geo.MbtilesFile.LayerType;
 import org.odk.collect.android.geo.MbtilesFile.MbtilesException;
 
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel:
🤦🏻‍♀️Looks like some kind of Android bug to me. Thanks for diving in.

mmarciniak90:
**Tested with success**

Verified on Android: 4.4, 4.2 - crash does not occur
Verified also on Android 6.0 and 7.0

Issue #3253, opened 2019-07-26
User-Agent string lists system before product, contrary to RFC 7231 recommendation


lognaturel:
As I looked at addressing #3252, I noticed that the User-Agent string currently used in requests does not follow the order recommended by [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.5.3):

> By convention, the product identifiers are listed in decreasing order of their significance for identifying the user agent software.

As described by #800 and shown in tests, Collect's User-Agent string looks like:
`Dalvik/2.1.0 (Linux; U; Android 5.1.1; hi6210sft Build/LMY47X) org.odk.collect.android/v1.5.1-10-ge20fa334-dirty`

The RFC only makes a recommendation so this is not a big deal but I wanted to at least make the decision explicit.

I'm guessing that it was done this way because the application ID looks odd at the head. 

~Using the application ID is also unorthodox, I think. It has the nice property of ensuring that forks by default provide an identifying User-Agent string. But we could also easily put a string in a config file that forks don't get and have forks be identified as something like "UnidentifiedODKCollectFork".~ @zestyping discovered in https://github.com/opendatakit/collect/issues/3253#issuecomment-515638028 that Mapbox uses the application ID as well. Thinking about it more, I think it's the right way to go.

Is it worth changing either the order ~or the product name or both~? I think this is unlikely to be disruptive so I would tend to change it but I may not be thinking of ways this could break downstream tools. Naturally, it would reset quotas based on User-Agent strings.

CC @zestyping @yanokwa

zestyping:
Putting it first (before `Dalvik`) seems reasonable.  I don't think we need to replace the app ID; I agree it's unusual but it also works well.  Just my two cents. :smile:

zestyping:
Side note: I confirmed that the tile requests from OSMDroid do not use the same HTTP infrastructure.  Those requests look like this, and would need to have their headers set separately:

```
GET /2/3/2.png HTTP/1.1
User-Agent: osmdroid
Host: 192.168.1.67:9999
Connection: Keep-Alive
Accept-Encoding: gzip
```

lognaturel:
@zestyping I'm hoping you'll enjoy https://github.com/opendatakit/collect/pull/3254 (writing the description up right now).

zestyping:
I just tried the same experiment with Mapbox, and discovered that it makes its own User-Agent string as well:

```
GET /17/21017/50624.png HTTP/1.1
[...]

PR #3254, opened 2019-07-27
Set User-Agent string for osmdroid

### LINKED ISSUE #3248: OpenStreetMap does not display on Androids 5 and higher
Collect v1.23.x, Android v5.1, 6.0, 7.0, and 8.1, device used...

#### Problem description
OpenStreetMap does not display when using Geo widgets.
Screenshot_2019-07-26-11-31-46

1. Choose in General Settings Maps/OpenStreetMap
2. Open Geopoint/GeoTrace/GeoShape widget
3. Click on Start button

#### Expected behavior
OpenStreetMap should load.

#### Other information 
Other types of map display correctly on Android 5.1 and higher
Interesting is that it displays correctly on Android 4.4 and lower.

### LINKED ISSUE #3252: Configure osmdroid to provide Collect User-Agent string to tile servers
We are supposed to identify the Collect app to tile servers: https://github.com/osmdroid/osmdroid/wiki/Important-notes-on-using-osmdroid-in-your-app#set-the-http-user-agent-variable

lognaturel:
Closes #3252
Closes #3248

I have communicated with OSM and they have advised us to change the User-Agent string.

Verified that we follow [the osmdroid instructions](https://github.com/osmdroid/osmdroid/wiki/Important-notes-on-using-osmdroid-in-your-app#set-the-http-user-agent-variable).

Used the [network profiler](https://developer.android.com/studio/profile/network-profiler) to look at the request details while loading an osmdroid-rendered map and verified that the User-Agent header is as expected (Dalvik/2.1.0 (Linux; U; Android 5.1.1; SM-T285 Build/LMY47V) org.odk.collect.android/v1.23.0-beta.4-9-g4cb530c69).

I moved the `getUserAgentString` method to the application level so all components that send user agent strings can use a consistent one.

I considered also addressing https://github.com/opendatakit/collect/issues/3253 but since I'm not totally sure it's without risk and it's not really a fix, I figured we can do this for the next release.

It should not affect users.

It does affect tile servers that we pull from, though, because now they'll be seeing a new User-Agent string. In particular, this will bring back access to OpenStreetMap tiles which violates their terms of service. We should wait until we learn more about #3248 before merging.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -29,6 +29,10 @@ import android.os.SystemClock;
 import android.preference.PreferenceManager;
 import android.util.Log;
 
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatDelegate;
+import androidx.multidex.MultiDex;
+
 import com.crashlytics.android.Crashlytics;
 import com.evernote.android.job.JobManager;
 import com.evernote.android.job.JobManagerCreateException;
@@ -68,9 +72,6 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.util.Locale;
 
-import androidx.annotation.Nullable;
-import androidx.appcompat.app.AppCompatDelegate;
-import androidx.multidex.MultiDex;
 import timber.log.Timber;
 
 import static org.odk.collect.android.logic.PropertyManager.PROPMGR_USERNAME;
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -208,6 +209,19 @@ public String getVersionedAppName() {
         return getString(R.string.app_name) + versionName;
     }
 
+    /**
+     * Get a User-Agent string that provides the platform details followed by the application ID
+     * and application version name: {@code Dalvik/<version> (platform info) org.odk.collect.android/v<version>}.
+     *
+     * This deviates from the recommended format as described in https://github.com/opendatakit/collect/issues/3253.
```
zestyping, line 216: Not a big deal, but I like putting the app ID first, at least in the long term.  Did you keep `Dalvik` first because you'd prefer to keep it first, or just to minimize changes for now?

zestyping (approved):

zestyping:
This looks great!

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
lognaturel: Yes, I do as well. I thought it was nice to have a single place where we generate a User-Agent string. And since changing it for all HTTP requests is more of an enhancement than a fix, I'd rather give a chance for #3253 to get comments and do it after v1.23 is out so it can spend some time in beta. I don't think changing it for osmdroid then will have any repercussions.

mmarciniak90:
@lognaturel this pr does not have assigned version. Do we want to have it in v1.23?

lognaturel:
Sorry about that. Yes, please. I initially wasn't sure it would be a fix for #3248 but it turns out to be.

mmarciniak90:
**Tested with success**

Verified on Android: 4.2, 4.4, 6.0, 7.0, 8.1

I confirm cases:
- OpenStreetMap displays correctly
- Offline layers are able to be applied on OpenStreetMap
- Verified GeoPoint, GeoShape, GeoTrace

PR #3255, opened 2019-07-27
Show attribution text for OSM, USGS, Stamen, and CARTO maps.

### LINKED ISSUE #3251: Provide proper attribution for map tiles
Collect makes available beautiful basemaps from a number of free web services. I just did some licensing due diligence and realized that we violate several:

- Carto: the [terms of service](https://drive.google.com/file/d/0B3OBExqwT6KJNHp3U3VUamx6U1U/view) state: "When using the Services, you must provide attribution to both OpenStreetMap and CARTO, as described at https://carto.com/attributions." It looks like it should be "© OpenStreetMap contributors, © CARTO"
- Stamen: from [the maps service landing page](http://maps.stamen.com/#toner), the attribution should be "Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>." 
- USGS: the [terms of service](https://www.usgs.gov/faqs/what-are-terms-uselicensing-map-services-and-data-national-map?qt-news_science_products=0#qt-news_science_products) state: "request that the following acknowledgment statement of the originating agency be included in products and data derived from our map services when citing, copying, or reprinting: "Map services and data available from U.S. Geological Survey, National Geospatial Program.""

zestyping:
Issues: Closes #3251.
Scope: OsmDroidMapFragment
Requested reviewers: @lognaturel 

#### User-visible changes

Attribution text now appears in the bottom-right corner for the OSM, USGS, Stamen, and CARTO maps.  The text is in the primary text colour for the selected theme.

The text is not always readable, depending on the theme and the lightness/darkness of the map tiles.  The text is also temporarily obscured when the zoom buttons appear (eventually the zoom buttons fade away).

The text follows the recommendations in #3251, though it does not provide clickable hyperlinks.

#### Screenshots

attr-stamen
attr-usgs-unreadable
attr-usgs-dark
attr-carto-positron

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -639,7 +639,7 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
         public static final int FONT_SIZE = 12;
         public static final int MARGIN_DP = 10;
 
-        private Paint paint;
+        private final Paint paint;
 
         public AttributionOverlay(Context context) {
             super();
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -20,6 +20,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.graphics.Bitmap;
+import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Paint;
 import android.location.Location;
@@ -47,9 +48,9 @@ import org.osmdroid.tileprovider.IRegisterReceiver;
 import org.osmdroid.util.BoundingBox;
 import org.osmdroid.util.GeoPoint;
 import org.osmdroid.views.MapView;
-import org.osmdroid.views.overlay.CopyrightOverlay;
 import org.osmdroid.views.overlay.MapEventsOverlay;
 import org.osmdroid.views.overlay.Marker;
+import org.osmdroid.views.overlay.Overlay;
 import org.osmdroid.views.overlay.Polyline;
 import org.osmdroid.views.overlay.TilesOverlay;
 import org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay;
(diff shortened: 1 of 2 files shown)


zestyping:
Hmm.  Unfortunately, the built-in `CopyrightOverlay` does not wrap the text, so long attributions can run off the left side on small screens.

zestyping:
Well... here's what it looks like if I render multi-line text.  It's not the prettiest, but I'm trying to keep the implementation as minimal as I can.  What do you think, @lognaturel ?

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -629,4 +633,41 @@ public void removeLastPoint() {
             }
         }
     }
+
+    /** An overlay that draws an attribution message in the lower-right corner. */
+    private static class AttributionOverlay extends Overlay {
+        public static final int FONT_SIZE = 12;
+        public static final int MARGIN_DP = 10;
+
+        private Paint paint;
```
lognaturel: final, per pmd

lognaturel:
TL;DR: as soon as the build is green this is good to go!

For truly tiny screens, it will still go off on the left and possibly interfere with the buttons on the right:
device-2019-07-26-202256

We're going to need to redo this when we switch to Mapbox for now so I think what's important is reacting quickly now that we know the problem exists. I think it's a good faith effort and not so terrible even in the tiny screen case.

zestyping:
Alas.

Oh, also, I've hardcoded these strings and not made them translatable.  Is that the right choice here?

lognaturel (approved):
I agree with the choice to make these always displayed in English. I think that is typical with this kind of attribution. The only one that won't make much sense to someone who doesn't speak English is the USGS one and it only applies to US maps.

Thanks!

mmarciniak90:
@lognaturel this pr does not have assigned version. Do we want to have it in v1.23?

lognaturel:
Yes, please. Sigh, I put the issue in the milestone but not the PR. I think that once we notice we are doing something incorrectly we should react as quickly as possible.

mmarciniak90:
@zestyping I noticed the problem on 4 maps on **Geopoint** widget:
1. Attribution text disappears when map zooms to the current location
2. Attribution text disappears when device orientation is changed



---
**Verified cases:**
* Geopoint, Geotrace, GeoShape
* reopened map with added location points
* offline layers
* rotation

I agree with @zestyping
> The text is not always readable, depending on the theme and the lightness/darkness of the map tiles.
> The text is also temporarily obscured when the zoom buttons appear (eventually the zoom buttons fade away).

Text can be also obscured by buttons visible on the right side.
Screenshot_20190730-135215

mmarciniak90:

zestyping:
@lognaturel Please let me know if you think this needs attention—I don't have a plan to work on this at the moment, but can if you'd like.

lognaturel:
Thanks as always for the thorough verification, @mmarciniak90. I have merged and filed https://github.com/opendatakit/collect/issues/3263. The important thing is that the attribution text is available at some point so this is pretty low priority. I'm guessing there are some deeper lifecycle issues that may take some work.

mmarciniak90:

PR #3256, opened 2019-07-27
Removed the mappings present in the CollectContentTypeMapping enum wh…

### LINKED ISSUE #3173: Remove custom handling for file types
From https://github.com/opendatakit/collect/pull/3163

"For a rainy day / good first issue, some of the types that Collect matches against are handled by the Android type mapper. It would be great for someone to go through and remove custom handling for those, I think."

arun-y99:
…ich are already handled by the MimeTypeMap mapper.

Closes #3173 

Changes:
All the mappings that are covered in the MimeTypeMap.java file have been removed in the custom map.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
@@ -32,21 +32,15 @@ public class CollectThenSystemContentTypeMapper implements OpenRosaHttpInterface
     }
 
     private enum CollectContentTypeMappings {
-        XML("xml",  "text/xml"),
-        _3GPP("3gpp", "audio/3gpp"),
-        _3GP("3gp",  "video/3gpp"),
         AVI("avi",  "video/avi"),
         AMR("amr",  "audio/amr"),
         CSV("csv",  "text/csv"),
-        JPG("jpg",  "image/jpeg"),
         MP3("mp3",  "audio/mp3"),
-        MP4("mp4",  "video/mp4"),
         OGA("oga",  "audio/ogg"),
         OGG("ogg",  "audio/ogg"),
         OGV("ogv",  "video/ogg"),
         WAV("wav",  "audio/wav"),
-        WEBM("webm", "video/webm"),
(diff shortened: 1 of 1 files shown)


lognaturel:
Hello and welcome, @arun-y99! It looks like one of the tests is failing. Could you please take a look at that?

lognaturel:
Hi @arun-y99! If you have any questions about tests, consider hopping into #collect-code [in Slack](https://slack.opendatakit.org/).

arun-y99:
Hi @lognaturel , I am just trying to figure out why the MimeTypeMap is not handling the mappings removed by me in the custom mapping.
For e.g. the "text/xml","xml" mapping is handled by MimeTypeMap, but the value returned is "application-octet-stream", which is the default string returned if no mapping is found.

lognaturel:
The failure I'm seeing is `expected:<a[udio/3gpp]> but was:<a[pplication/octet-stream]>`. This makes sense because 3gpp was defined in the custom map but not in Android. You removed it from the custom map so there's no support anymore. 

How did you pick which mappings to remove? The ones that are in Android should be removed. The ones that are not in Android should be maintained.

seadowg (changes requested):
Hey @arun-y99 👋 

Android may well handle the mappings you've removed but our tests use a mock `MimeTypeMap` and basically don't make any assumptions around what mappings it may handle. It'd be great as part of this PR if you'd be able to link some documentation about what mappings Android handles and then remove those mappings from both the object and the test. If we know Android handles a file extension then we don't need a test to check our code handles it - the test `whenExtensionIsNotRecognized_returnsTypeFromAndroid` covers us falling back to Android. If you have questions [our Slack](https://slack.opendatakit.org/) (as @lognaturel linked to) is good place to reach out!

It'd also be helpful if you could make sure to include the answers to the questions and the checkboxes that our [PULL_REQUEST_TEMPLATE](https://github.com/opendatakit/collect/blob/master/.github/PULL_REQUEST_TEMPLATE.md) should have prompted you with. This helps us understand the changes and lets us know that the PR is ready to review.

lognaturel:
Closing due to inactivity. Please reopen if you'd like to address the comments above, @arun-y99!

PR #3259, opened 2019-07-29
Make raster tiles 100% opaque on Mapbox maps.

zestyping:
As discussed, this makes raster offline tile layers appear 100% opaque (instead of 50% transparent) on Mapbox maps, for consistency with the rendering of offline tiles on Google and OSMDroid maps.

- Built and ran this APK on both Android 6.0.1 and Android 9.0
- Selected the "satellite.mbtiles" raster layer (attached)
- Opened GeoShape with a Google map, a Mapbox map, and the USGS topographic map
    - In each case, I zoomed out until the satellite tiles appeared, and confirmed they were drawn fully opaque.
    - In each case, I used the Layers button to turn the layer on and off and confirmed it still appears properly.

The main reason we decided to do this is that raster layers are drawn fully opaque on Google maps and OSMDroid maps, and we wanted to make this behaviour consistent on Mapbox maps.

The intentional change is that the raster layer is 100% opaque.

However, there is now a different inconsistency.  MBTiles files contain, in their metadata, a minimum and maximum zoom level declaration.  For the low-res satellite.mbtiles file, this range is 0 to 5.

On Google and OSMDroid maps, the raster layer is only drawn at the zoom levels for which it is declared.  On Mapbox maps, the raster layer is always drawn.  So when the map zooms to your current location, on Google and OSMDroid, you will see the basemap, but on Mapbox, the entire screen will be a solid colour (dark green on land, typically).

#### Open questions

Should we implement our own check to make the raster layer disappear when the zoom level is outside of the declared range?
  - Pro: The behaviour will exactly match that in Google and OSMDroid.
  - Con: The sudden appearance/disappearance of the raster layer in Google and OSMDroid is jarring and has confused me in the past.  At the default zoom level after a GPS fix (level 16), the raster layer typically doesn't appear, so I think it's not working, and there's no indication in the UI what the zoom level is or what zoom levels it should be expected to appear at.  It's actually kind of nice to be able to see that the raster layer is there.

In theory, this is a reference layer, so it isn't really supposed to be used for images that are opaque everywhere. But of course this is the only option we are providing right now; we don't offer a way to specify an offline "basemap" in the preferences.

One could imagine a nice compromise behaviour where the opacity is 100% within the declared range (in this case, 0 to 5), and then smoothly falls off to something like 30–40% outside of that range (maybe at zoom level 6 it's 70%, at zoom level 7 it's 40%, that sort of thing).  I'm pretty sure Mapbox can do this nicely if we want it to, though this would be inconsistent with Google and OSMDroid.

The regular GPS widgets in the "All widgets" form are sufficient for testing.

Documentation issue: https://github.com/opendatakit/docs/issues/1073

- [x] run `./gradlew checkCode` and confirmed all checks pass
- [ ] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally (this started 20 minutes ago and is still running so I'll post this PR for now and update this when it finishes)
- [ ] verified that any code or assets from external sources **(no new code or assets)**
- [ ] verified that any new UI elements use theme colors **(no new UI elements)**

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
@@ -8,6 +8,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.view.Gravity;
 
+import com.google.common.collect.ImmutableSet;
 import com.mapbox.android.core.location.LocationEngine;
 import com.mapbox.android.core.location.LocationEngineCallback;
 import com.mapbox.android.core.location.LocationEngineProvider;
@@ -53,6 +54,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -592,8 +594,35 @@ public class MapboxMapFragment extends org.odk.collect.android.geo.mapboxsdk.Map
     }
 
     private void addOverlayLayer(Layer layer) {
-        map.getStyle().addLayer(layer);
         overlayLayers.add(layer);
+
+        // If there is a LocationComponent, it will have added some layers to the
+        // style.  The SymbolManager and LineManager also add their own layers
+        // where they place their symbols and lines.  We need to insert the new
+        // overlay just under all these upper layers to keep it from covering up
+        // the crosshairs, the point markers, and the traced lines.
+        Set<String> upperLayerIds = ImmutableSet.of(
+            SymbolManager.ID_GEOJSON_LAYER,
+            LineManager.ID_GEOJSON_LAYER,
+
+            // These are exactly the layer IDs defined in LocationComponentConstants,
+            // but unfortunately we can't refer to them because it's package-private.
+            "mapbox-location-shadow-layer",
+            "mapbox-location-foreground-layer",
+            "mapbox-location-background-layer",
+            "mapbox-location-accuracy-layer",
+            "mapbox-location-bearing-layer"
+        );
+
+        for (Layer l : map.getStyle().getLayers()) {
+            if (upperLayerIds.contains(l.getId())) {
+                // We've found the first (lowest) upper layer; insert just below it.
+                map.getStyle().addLayerBelow(layer, l.getId());
+                return;
+            }
+        }
+        // No upper layers were found, so let's put the overlay on top.
+        map.getStyle().addLayer(layer);
     }
 
     private void addOverlaySource(Source source) {
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
@@ -65,7 +65,6 @@ import timber.log.Timber;
 import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineColor;
 import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineOpacity;
 import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineWidth;
-import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.rasterOpacity;
 
 public class MapboxMapFragment extends org.odk.collect.android.geo.mapboxsdk.MapFragment
     implements MapFragment, OnMapReadyCallback,
@@ -515,6 +514,9 @@ public class MapboxMapFragment extends org.odk.collect.android.geo.mapboxsdk.Map
             List<MbtilesFile.VectorLayer> layers = mbtiles.getVectorLayers();
             for (MbtilesFile.VectorLayer layer : layers) {
                 // Pick a colour that's a function of the filename and layer name.
+                // The colour will appear essentially random; the only purpose here
+                // is to try to assign different colours to different layers, such
+                // that each individual layer appears in its own consistent colour.
                 int hue = (((id + "." + layer.name).hashCode()) & 0x7fffffff) % 360;
                 addOverlayLayer(new LineLayer(id + "/" + layer.name, id).withProperties(
                     lineColor(Color.HSVToColor(new float[] {hue, 0.7f, 1})),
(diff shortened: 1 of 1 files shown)


zestyping:
Okay, `checkAll` finally finished.  It took about 20 minutes to run.  Three tests failed:

```
org.odk.collect.android.http.OkHttpConnectionGetRequestTest > withCredentials_whenHttps_retriesWithCredentials FAILED
    java.net.ConnectException
        Caused by: java.net.ConnectException

org.odk.collect.android.http.OkHttpConnectionHeadRequestTest > withCredentials_whenHttps_retriesWithCredentials FAILED
    java.net.ConnectException
        Caused by: java.net.ConnectException

org.odk.collect.android.http.OkHttpConnectionPostRequest > withCredentials_whenHttps_retriesWithCredentials FAILED
    java.net.ConnectException
        Caused by: java.net.ConnectException
```

The rest of the tests passed.  I assume this is due to a specific HTTP server not being available; is this typical?

lognaturel (approved):
Thanks for laying out the behavior options. I think sticking with this default behavior of showing only the raster fill at high levels of zoom where tiles are unavailable is ok. I agree that the osmdroid/Google behavior is confusing -- how about filing an issue to make that one consistent with this?

And I like that augmented comment, too. 👍

lognaturel:
> It took about 20 minutes to run

We do have a problem with instrumented tests taking too long and also with them failing on some devices (e.g. on Firebase Lab). Maybe what we should do is look into whether we can run them remotely on all branches.

I have seen those test failures as well. They don't fail on CI but we should definitely get them fixed.

mmarciniak90:
@zestyping I noticed one problem:
1. User sets Mapbox and selects satellite.mbtiles
2. User opens some Geo widgets, e.g. GeoTrace
3. User sees pointer
4. User changes Layer to `None`
5. User sees pointer
6. User changes Layer to satellite.mbtiles
7. A pointer is not visible. It appears if device orientation is changed.



---
I have also doubt about
>The main reason we decided to do this is that raster layers are drawn fully opaque on Google maps and OSMDroid maps, and we wanted to make this behaviour consistent on Mapbox maps.


|Google|OSM|Mapbox without fix|Mapbox with fix|
|---|---|---|---|
|Screenshot_20190730-160717|Screenshot_20190730-160752|Screenshot_20190730-151433|Screenshot_20190730-151313|
|Screenshot_20190730-160816|Screenshot_20190730-160711|Screenshot_20190730-151538|Screenshot_20190730-151406

mmarciniak90:

lognaturel:
> I have also doubt about

For the zoomed out case, I think the confusion comes from the fact that Google and OSM are currently inconsistent when it comes to labels, right? This is because of https://github.com/opendatakit/collect/issues/3004 -- labels are rendered on top of the reference layer in Google Maps. In the Mapbox case, the labels show on master because the reference layer is partly transparent. The layer itself could have transparency so it makes more sense to draw it as it is defined, without modifying its transparency. 

For the zoomed in case, there's not a great option when high-res tiles are not available as part of the layer. In the Google and OSM case, the layer just disappears which is pretty confusing if you think of that layer as a basemap. In the new Mapbox case, it just blows up the most zoomed in layer available which doesn't have much detail so looks like a solid color. Not great either but at least it's more consistent with the idea that an opaque reference layer is meant to be treated like a basemap.

I agree we should revisit all of this but in the short term, rendering the raster layer with the transparency of the tiles seems like an improvement to me.

> A pointer is not visible. It appears if device orientation is changed.

Sounds like the pointer is rendered under the reference layer and it was only visible before because the layer always had some transparency. That would be good to address. @zestyping, would you have time to look at it? If not, I can take it on during my Wednesday.

zestyping:
@mmarciniak90 Good catch! @lognaturel I'm looking at it.

zestyping:
An update on this as I'm investigating.  Here's what I've tried and learned so far:

  - Disabling and re-enabling the `LocationComponent` doesn't make it visible.
  - Stopping and re-starting the `LocationComponent` doesn't make it visible.
  - The `LocationComponent` adds 5 layers to the map in its `addLayers` method, and there is no opportunity (visible in its source code) to remove and re-add those layers.  A flag ensures that initialization of those layers can happen only once, so the only way to remove them is to destroy and re-create the entire `LocationComponent`.
  - Visibility of the location crosshairs is not the whole problem; the red lines drawn by the `LineManager` and red dot markers drawn by the `SymbolManager` are also covered up by reference layer overlays.
  - The `LocationComponent`, `LineManager`, and `SymbolManager` don't provide access to the layers they create.  The only way I've found so far to identify these layers is by the spelling of the layer IDs (e.g. "mapbox-location-shadow-layer", "mapbox-android-symbol-layer").  It's fragile, but I haven't found an alternative.


So, my current tack is to try to guess which layers belong to the `LocationComponent`, `LineManager`, and `SymbolManager` by looking at their IDs, and insert the reference layer just under them.

zestyping:
Okay, I found constants with these layer IDs, and have pushed a change that looks for these layers and inserts the overlay under them.  Unfortunately, some of the constants are not accessible to our code, so I had to retype the strings.  But it does appear to work.

As a side discovery, I found that when there wasn't a valid Mapbox access token (which causes the map to fall back to OSM tiles), the red circles were not appearing.  I fixed this in https://github.com/opendatakit/collect/pull/3259/commits/75d002706ea5281cda3746f8996d83a58f0d02b1.

Expected behaviour:
1. On a Mapbox map, adding and removing reference layers (using the button with the layers icon) should not ever cover up the location crosshairs, the accuracy indicator circle, the collected points (red circles), or the collected traces (red lines).
2. When the Mapbox map first loads, the location layers (crosshairs and accuracy circle) are on top of the annotation layers (the collected points and traces).  
3. Adding a reference layer is known to swap the order of the location layers and the annotation layers, which makes it possible to cover up parts of the crosshairs with red circles and lines.
4. If you build an APK without a Mapbox access token (an empty `collect_app/secrets.properties` file), collected points should appear as the usual small red circles.  Prior to this fix, these circles were not drawn when there was no access token.

#2, I think, is as it should be.  #3 is unfortunate but quite difficult to fix, as it is caused by the order in which event listeners are invoked to set up the location layers and annotation layers.  We have control of the setup order when first creating the map so we can cause the location layers to be added last; but we don't have control over the sequence in which the event listeners are invoked.  Further, there doesn't appear to be an exposed method that reorders existing layers.  There are only methods for inserting new layers or removing existing layers, and removing a layer destroys it so it cannot then be inserted elsewhere in the layer list.

zestyping:
Filed https://github.com/mapbox/mapbox-gl-native/issues/15262.

kkrawczyk123:
I have re-tested pointer issue reported by @mmarciniak90 and I was not able to reproduce it.
Verified on Androids 4.2, 4.4, 5.1, 6.0 and 8.1. 

PR #3260, opened 2019-07-29
Use high accuracy priority to get points in absence of wifi or cell

### LINKED ISSUE #3244: Setgeopoint action doesn't get frequent enough updates in the absence of wifi or cell access
Collect v1.23.beta-0+

#### Problem description
The setgeopoint action does not receive location updates frequently enough to be useful in the absence of wifi or cell.

Use a form like the setgeopoint test form on http://opendatakit.appspot.com/ that triggers the action multiple times. Walk around an area without cell or wifi, triggering the action. Notice that the location access icon is rarely shown in the Android status bar. Look at the submission and see that very few locations were captured.

#### Expected behavior
Locations are generally captured, even in the absence of wifi or cell.

lognaturel:
Closes #3244

Went to a large park with clear view of the sky and no wifi. Used a device with no SIM to fill out the `Collect v1.23 Beta: Background Geopoint Test` form from the default server. Did it first from master and then from this branch.

I expected that [PRIORITY_BALANCED_POWER_ACCURACY](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest.html#PRIORITY_BALANCED_POWER_ACCURACY) would still access GPS when no location was available from other providers but that does not appear to be the case or at least it doesn't attempt it frequently. After re-reading [the description of the accuracy levels](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest), I realized that we really do want the most accurate location available for the background geopoint feature rather than "block level" accuracy. If a form designer is explicitly asking for location in the form, it's not acceptable that it only work reliably in an environment with wifi and cell, especially given our users. The battery-saving measure we have in place is that we only request location for 20 seconds and give up if none is available by then (this might need to be revisited as well or perhaps become configurable).

The GPS will be accessed more frequently to capture location in the background. When that occurs, the user will see the location access icon in the device status bar. This should go away after at most 20 seconds for each background location request (in the case of the beta form, each multiple selection question triggers a location request). This may result in more battery usage if wifi or cell localization is unavailable.

`Collect v1.23 Beta: Background Geopoint Test` form from the default server. I'm not entirely sure how to verify this in an urban environment. I don't believe all devices make it possible to entirely turn off cell and wifi localization (that is, even if you're not connected to a wifi network, I believe local networks can be used). I had to get out of any network range to really see the bad performance of master.

Since this now matches the accuracy setting of the audit log and mapping features, perhaps it doesn't need QA?

Once the pyxform implementation is in, the feature will get full documentation.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/location/client/MaxAccuracyWithinTimeoutLocationClient.java
+++ b/collect_app/src/main/java/org/odk/collect/android/location/client/MaxAccuracyWithinTimeoutLocationClient.java
@@ -27,7 +27,7 @@ public class MaxAccuracyWithinTimeoutLocationClient implements LocationClient.Lo
     private Location highestAccuracyReading;
     private final Handler timerHandler;
 
-    private static final LocationClient.Priority DEFAULT_PRIORITY = LocationClient.Priority.PRIORITY_BALANCED_POWER_ACCURACY;
+    private static final LocationClient.Priority DEFAULT_PRIORITY = LocationClient.Priority.PRIORITY_HIGH_ACCURACY;
 
     public MaxAccuracyWithinTimeoutLocationClient(LocationClient locationClient, LocationListener listener) {
         this.locationClient = locationClient;


grzesiek2010 (approved):

grzesiek2010:
> perhaps it doesn't need QA?

I agree.

PR #3261, opened 2019-07-30
Fixed navigationButtons_ShouldBeVisibleWhenAreSetInTheMiddleOfForm test

grzesiek2010:
One espresso test is failing https://console.firebase.google.com/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/6777832822726188216

I was able to reproduce the issue and confirm that my changes fix it.

`openContextualActionModeOverflowMenu()` method which I used is just a more reliable way to click on `three-dots`.

It doesn't require testing

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/FormEntry.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/FormEntry.java
@@ -1,5 +1,6 @@
 package org.odk.collect.android.espressoutils;
 
+import androidx.test.espresso.Espresso;
 import androidx.test.espresso.matcher.PreferenceMatchers;
 import androidx.test.rule.ActivityTestRule;
 import org.odk.collect.android.R;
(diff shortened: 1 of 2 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/FormEntry.java
```diff
@@ -123,7 +123,7 @@ public static void clickOk() {
     }
 
     public static void clickOptionsIcon() {
-        onView(withContentDescription("More options")).perform(click());
+        Espresso.openContextualActionModeOverflowMenu();
```
seadowg: Should this be `openActionBarOverflowOrOptionsMenu` instead? Looking at the [docs](https://developer.android.com/reference/androidx/test/espresso/Espresso.html#openContextualActionModeOverflowMenu()) I'm not 100% sure as is it's a little confusing. It seems `openContextualActionModeOverflowMenu` would also work on "floating ActionModes" (I'm assuming like the context menus on long presses on text).

grzesiek2010: Yeah, both work well and I'm also not 100% sure but openActionBarOverflowOrOptionsMenu seems like a better method here so fixed.

seadowg (approved):

PR #3265, opened 2019-07-31
Fail encrypted form finalization if form does not specify instanceID

### LINKED ISSUE #3264: Don't finalize encrypted form if definition has no instanceID and show explanation to user
Collect v1.23.0 and prior

#### Problem description
When a form definition does not include an instanceID, it can't be encrypted because an instanceID is necessary to identify the record in the submission manifest. Currently, encryption just silently fails and the submission is still saved as finalized. A user may not realize that their data is not encrypted and could reveal sensitive information.

This was discovered when implementing encryption in Central.

1. Load a form with an encryption key but no instanceID such as [Birds-encrypted.xml.txt](https://github.com/opendatakit/collect/files/3449217/Birds-encrypted.xml.txt)
1. Fill out the form
1. Save as finalized
1. Edit saved form -- if it were encrypted, it should not be possible to edit.

#### Expected behavior
When the user attempts to save as finalized, a toast is displayed saying "This form does not specify an instanceID. You must specify one to enable encryption." The form is not saved as finalized.

lognaturel:
Closes #3264

Note: this is an unlikely case because all known form builders add an `instanceID`. That's why the error message is not translated. I did move it to the top, though, since it feels like an invariant.

I changed logging from `e` to `d` where the error is caused by the form or device configuration since there's nothing that ODK developers can do about those -- they're not software defects.

Filled form from issue, verified that finalization is disabled and that a toast is shown.

This uses the existing infrastructure to stop finalization and present an error to the user. I considered keeping the Timber log but it's not a software error so there's no point in logging it to Crashlytics.

The user will be unable to finalize forms with encryption keys and no instanceIDs. No regression risk that I can think of.

[Birds-encrypted.xml.txt](https://github.com/opendatakit/collect/files/3449331/Birds-encrypted.xml.txt)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
@@ -17,8 +17,9 @@ package org.odk.collect.android.utilities;
 import android.content.ContentResolver;
 import android.database.Cursor;
 import android.net.Uri;
-import androidx.annotation.Nullable;
 import android.util.Base64;
+
+import androidx.annotation.Nullable;
 
 import org.apache.commons.io.IOUtils;
 import org.kxml2.io.KXmlSerializer;
@@ -249,9 +250,14 @@ public class EncryptionUtils {
      * Retrieve the encryption information for this uri.
      *
      * @param uri either an instance URI (if previously saved) or a form URI
+     * @throws EncryptionException if the record can't be encrypted
      */
     public static EncryptedFormInformation getEncryptedFormInformation(Uri uri,
             InstanceMetadata instanceMetadata) throws EncryptionException {
+        // submission must have an OpenRosa metadata block with a non-null instanceID
+        if (instanceMetadata.instanceId == null) {
+            throw new EncryptionException("This form does not specify an instanceID. You must specify one to enable encryption.", null);
+        }
 
         ContentResolver cr = Collect.getInstance().getContentResolver();
 
@@ -298,7 +304,7 @@ public class EncryptionUtils {
 
                 if (formCursor.getCount() != 1) {
                     String msg = Collect.getInstance().getString(R.string.not_exactly_one_blank_form_for_this_form_id);
-                    Timber.e(msg);
+                    Timber.d(msg);
                     throw new EncryptionException(msg, null);
                 }
                 formCursor.moveToFirst();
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
```diff
@@ -355,8 +356,7 @@ public static EncryptedFormInformation getEncryptedFormInformation(Uri uri,
         // submission must have an OpenRosa metadata block with a non-null
         // instanceID value.
         if (instanceMetadata.instanceId == null) {
-            Timber.e("No OpenRosa metadata block or no instanceId defined in that block");
-            return null;
+            throw new EncryptionException("This form does not specify an instanceID. You must specify one to enable encryption.", null);
```
seadowg: I think we should probably have that string in `string.xml`

seadowg (changes requested):
I think it'd be good to have test of at least `getEncryptedFormInformation` to drive out this case. Eventually it would be nice to have some coverage of the infrastructure that results in the `Toast` being shown but given we've chatted about moving the Task tests I don't think right now is a good time for that particular backfill.

collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
lognaturel: Putting strings in `string.xml` is not "free" in the sense that it adds more work for translators. I think that for edge cases like this one, it's preferable to avoid having it translated. This is extremely unlikely since pyxform and Build add instanceID. It's really only likely when someone comes across a really old form definition (generally used for testing).

I've added 557a826 to explain the exception and hopefully you find it reasonable.

lognaturel:
@seadowg I hope that you will find my test sufficient. It's not ambitious but this is mostly for developers who have old test forms lying around so I don't want to spend too much time here.

seadowg (approved):
I like it! Simple.

bognasoldev89:
*Tested with success*

Verified on Android: 
4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 9.0

Verified cases:
- Filling, saving and finalizing attached form

On all of the Androids user is not able to finalize encrypted form if definitions has no instanceID. Form is saved and visible in Edit Saved Form. Toast is displayed after marking form as finalized, saying “This form does not specify an instanceID. You must specify one to enable encryption. Form has not been saved as finalized.”

Screenshot_2019-09-13-12-35-29

PR #3269, opened 2019-07-31
Fixed header dates

### LINKED ISSUE #3266: Unable to open Get Blank Form list with Aggregate as Server and date with unicode characters
Collect v1.23.x, Android 4.4, 5.1, 6.0, 7.0. 

#### Problem description
I have encountered a weird bug when performing regression testing. I had chosen the Spanish language in devices settings and tried to open Get Blank Form. I had default Aggregate in Server settings I saw an error message: 
Screenshot (Jul 31, 2019 1_11_33 PM)

I started to investigate and these are the results that I have obtained:
- The error message was NOT visible when I changed the language to English in Collect or Devices Settings
- The error message was NOT visible when I changed the Date to 30th of July and to 1st August this year in devices Settings
- The similar error message was visible when I chose Date in Devices Settings to 29th February of 2020, I can't find any reason why 31th of July is any kind of suspicious date
- I tried with different languages, I was able to reproduce this error with Spanish, Polish, Arabic, Czech and Russian. Probably it occurs also for some other languages which name I don't know.
- I was NOT able to reproduce it on Collect 1.22.4 and 1.22.0
- This is the logcat:
`2019-07-31 12:25:33.831 4830-6253/org.odk.collect.android E/CollectServerClient: Parsing failed with Unexpected char 0xe9 at 2 in Date value: mié., 31 jul. 2019 10:25:33 GMT+00:00 while accessing https://opendatakit.appspot.com/formList
2019-07-31 12:25:34.029 4830-4830/org.odk.collect.android E/ViewRootImpl: sendUserActionEvent() mView == null`

1. Try today or change date in devices settings to 31th of July 2019
2. Set language in Collect or Device Settings Spanish/Polish/Arabic/Czech/Russian.
3. Try to open Get Blank Form

#### Expected behavior
I can't see any reason why the user should not be able to open Get Blank Form on 31th of July.

grzesiek2010:
Closes #3266 

Nice catch @kkrawczyk123 
Explanation:
The problem is that we attach date headers to request:
[here](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java#L308), [here](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java#L318) and [here](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java#L328)

those dates are generated [here](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java#L333) and they are in the same language that the app is using so if you use Spanish it would be `mié., 31 jul. 2019 08:27:51 GMT` what is not acceptable (the `é` char) and OkHttp throws an exception if we try to add a header with such a string.

I tested the mentioned scenario, reproduced the issue and confirmed that the fix works well.

I decided to use dates in English no matter what language is set in the app. It fixes the problem and I don't think that it's needed to have translatable dates in request headers.

It doesn't change anything apart from the language used to create dates we add to request headers, so I don't think it requires much testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
@@ -1,7 +1,5 @@
 package org.odk.collect.android.http;
 
-import android.text.format.DateFormat;
-
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
@@ -24,8 +22,8 @@ import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -331,9 +329,9 @@ public class OkHttpConnection implements OpenRosaHttpInterface {
     }
 
     private String getHeaderDate() {
-        GregorianCalendar g = new GregorianCalendar(TimeZone.getTimeZone("GMT"));
-        g.setTime(new Date());
-        return DateFormat.format("E, dd MMM yyyy hh:mm:ss zz", g).toString();
+        SimpleDateFormat dateFormatGmt = new SimpleDateFormat("E, dd MMM yyyy hh:mm:ss zz", Locale.US);
+        dateFormatGmt.setTimeZone(TimeZone.getTimeZone("GMT"));
+        return dateFormatGmt.format(new Date());
     }
 
     /**


grzesiek2010:
@lognaturel 
I think it should be added to v1.23

yanokwa:
I got a message on facebook for an issue that might be related.

> If I'm using Collect with "device language" (Hungarian), I can not download empty form, because in hungarian we use " á, í, é, ó, ú, ő, ű" e.g.: July = Július or in short júl., and it throw an error message. "Form listing failed. Oarsing failed with Unexpected char 0xfa at 8 in Date value: Cs, 25 júl. 2019 07:26:52 GMT+00:00 while accessing...." Simple workaround if I change the Collector language to english.

grzesiek2010:
> I got a message on facebook for an issue that might be related.

yeah, it's the same issue.

lognaturel:
Agreed with using English dates everywhere. It's really pretty nutty to send localized dates. ~@issa-tseng is going to have a laugh/cry about the fact that Aggregate didn't even support localized dates properly, I'm sure.~ (it's the change in HTTP lib that causes the exception, as @grzesiek2010 noted) 

We can take the opportunity to patch the spec to make it clear it should be an [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.1.2) date **in English**.

lognaturel (approved):

kkrawczyk123:
**Tested with success!!**
Verified on Androids 4.2, 4.4, 5.1, 6.0 and 8.1

PR #3272, opened 2019-08-01
Bypass caching of internal secondary instances for performance

### LINKED ISSUE #3271: Take advantage of form definition parsing performance enhancement in JavaRosa 2.15.0
https://github.com/opendatakit/javarosa/pull/445 introduced a significant performance improvement to forms with internal secondary instances. Taking advantage of it requires using the new `XFormUtils.getFormFromFormXml` instead of `XFormUtils.getFormFromInputStream`.

Once that change is made, the time taken to load a form with a large internal secondary instance such as [nigeria_wards_internal.xml.txt](https://github.com/opendatakit/collect/files/3454544/nigeria_wards_internal.xml.txt) should decrease significantly.

lognaturel:
Closes #3271

1. Loaded [nigeria_wards_internal.xml.txt](https://github.com/opendatakit/collect/files/3454550/nigeria_wards_internal.xml.txt) in v1.22.0, both right after pushing it to the device (no caching) and after that first load (from cache). Confirmed that both took a long time.
1. **Cleared the form load cache** from Admin Settings > Reset Application... > Form load cache
1. Installed from this branch
1. Loaded [nigeria_wards_internal.xml.txt](https://github.com/opendatakit/collect/files/3454550/nigeria_wards_internal.xml.txt) twice and confirmed that it was much faster both times (both from XML and from cache).

It's the narrowest change to use the new method.

The only intended change is that forms with internal secondary instances (basically, selects with choice filters) load much faster.

What's happening behind the scenes is that the caching of those internal secondary instances is skipped. There's a risk of bad behavior when upgrading or downgrading. There's a risk to any form with a select and a choice filter. 

The Google Sheets implementation also uses this code to load a form definition on submission (to figure out the column headers). There's a risk of regression there as well.

Since this is a change to caching, issues would show up on second load or later.

[nigeria_wards_internal.xml.txt](https://github.com/opendatakit/collect/files/3454550/nigeria_wards_internal.xml.txt)

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
@@ -18,7 +18,6 @@ import android.content.Intent;
 import android.database.Cursor;
 import android.os.AsyncTask;
 
-import org.apache.commons.io.IOUtils;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.FormIndex;
 import org.javarosa.core.model.instance.InstanceInitializationFactory;
@@ -51,7 +50,6 @@ import org.odk.collect.android.utilities.ZipUtils;
 
 import java.io.File;
 import java.io.FileFilter;
-import java.io.FileInputStream;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


lognaturel:
I was certain that @JohnTheBeloved has put in a Collect PR for this but it looks like we didn't coordinate on that. I'm disappointed because we announced the improvement in the beta but it's too late to get merged for v1.23. Let's try to get it merged as quickly as possible for v1.24 so those new code paths can get exercised and users can get the benefits. @yanokwa, please edit the release announcements for v1.23 accordingly.

seadowg (approved):
As far as I can see this is mainly just switching the method over so looks good to me!

mmarciniak90:
**Tested with success**

Verified cases:
* main case - **loading form takes less time on all devices**
* upgrade/downgrade : v.1.22 -> version with fix (started `Nigeria Wards` form) -> v1.22 (opened form from cache)
* upgrade/downgrade : v.1.22.x (started `Nigeria Wards` form) -> version with fix (opened form from cache)
* adding answers and reload it from a cache
* worked with logcat to find new errors
* bunch of other forms to check for possible regressions

PR #3273, opened 2019-08-01
Fixed crash in FormDownloadList

### LINKED ISSUE #3267: App crashes when Storage permission is unselected and user tries to return to `Get Blank Form` view
Collect v1.22.4 (older versions were not verified)

#### Problem description
App crashes when Storage permission is unselected and user tries to return to `Get Blank Form` view

1. User uses Collect. **Aggregate** is set as a server.
2. User opens Get Blank Form
3. Storage permission is unselected
4. User returns to Collect
5. App crashes

Screenshot from 2019-07-31 14-12-56

#### Expected behavior
App does not crash
User is able to continue work

#### Other information 
The crash is not visible when Google Drive is set as a server

grzesiek2010:
Closes #3267 

Reproduced the described crash and confirmed it fixes the problem.

I just added a check to avoid calling `updateAdapter()` if storage permissions are not granted. It's the only solution here.

It's a safe fix. It won't change anything else apart from fixing the issue.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadList.java
@@ -339,7 +339,6 @@ public class FormDownloadList extends FormListActivity implements FormListDownlo
     @Override
     protected void onRestoreInstanceState(Bundle state) {
         super.onRestoreInstanceState(state);
-        toggleButtonLabel(toggleButton, listView);
         updateAdapter();
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadList.java
@@ -339,7 +339,9 @@ public class FormDownloadList extends FormListActivity implements FormListDownlo
     @Override
     protected void onRestoreInstanceState(Bundle state) {
         super.onRestoreInstanceState(state);
-        updateAdapter();
+        if (PermissionUtils.areStoragePermissionsGranted(this)) {
+            updateAdapter();
+        }
     }
 
     @Override


collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadList.java
```diff
@@ -339,8 +339,9 @@ private void downloadFormList() {
     @Override
     protected void onRestoreInstanceState(Bundle state) {
         super.onRestoreInstanceState(state);
-        toggleButtonLabel(toggleButton, listView);
-        updateAdapter();
+        if (PermissionUtils.areStoragePermissionsGranted(this)) {
+            updateAdapter();
```
grzesiek2010, line 343: I can't recall why it's needed here, maybe it's not at all but I guess there is a tricky case that needs this fix so I decided to keep it and add if statement instead of removing what also would fix the crash but could bring some strange problems.

lognaturel (approved):

lognaturel:
Thanks for separating out that first commit -- I had a moment where I wondered where that call went, was about to look into it, but the answer was right there! ✨

grzesiek2010:
No testing? The solution seems obvious but we both have seen even more obvious that we merged without testing and there was something wrong.

lognaturel:
Sigh, I suppose you're right. I think we're ok to release it but @grzesiek2010 is right that it'd be good to get a check @kkrawczyk123, @mmarciniak90.

mmarciniak90:
**Tested with success**

Verified on Android: 9.0, 8.1, 7.0, 6.0
Verified on v1.23 downloaded from `Play Store`

Verified cases:
* Storage permission denied on `Get Blank Form` when **Aggregate** was set as a server
* Storage permission denied on `Get Blank Form` when **Google Drive** was set as a server

Issue #3205, opened 2019-07-11
Espresso regression tests failing sporadically


seadowg:
The `DrawWidgetTest` and `SignatureWidgetTest` appear to be failing on most runs on Test Lab but not in a predictable manner. The actual failure appears to be:

```
androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id: org.odk.collect.android:id/jumpEndButton 
```

grzesiek2010:
I'm pretty sure the problem is related to orientation changes...
in those tests, we use DrawActivity which works in landscape mode. When a user returns from that activity to DrawWidget/SignatureWidget a rotation will take place landscape->portrait. It might take some time and probably it's the cause.

What do you think about calling [waitForIdleSync](https://developer.android.com/reference/android/app/Instrumentation.html#waitForIdleSync()) every time we get back from that activity to wait for the rotation end?

https://stackoverflow.com/a/38860384

seadowg:
@grzesiek2010 my impression is that Espresso should be doing that for us but it's definitely worth a try. I hadn't thought about the screen rotations and it wouldn't surprise me if that messes with Espresso...

grzesiek2010:
@seadowg so maybe let's try it https://github.com/opendatakit/collect/pull/3277

seadowg:
@grzesiek2010 do you think it's worth merging in #3276 first as a temporary measure or are you pretty close with #3277?

seadowg:
Sorry I should have closed this with #3276.

PR #3276, opened 2019-08-02
Ignore flakey tests

seadowg:
This is a temporary measure until we have time to have a look at stabilizing them. This means losing the coverage these tests bring but gives us confidence in our test suite again.

This is temporary relief for #3205.

Double checked the tests are actually ignored by running the suite.

We could of course delete the tests but I think it's better to ignore right now so that we can have a look at these tests when someone has time. We have an issue open so I'm confident we can get to it.


grzesiek2010 (approved):
I need more time to investigate and fix those failing tests so we can ignore them temporarily.

PR #3277, opened 2019-08-02
Fix tests that contain screen rotation

### LINKED ISSUE #3205: Espresso regression tests failing sporadically

grzesiek2010:
Closes #3205 

We can't reproduce the issue so nothing.

I just guess that it might fix failing tests as described in #3205 

No risk.

No.

No.


seadowg:
@grzesiek2010 did this fix it on Test Lab?

grzesiek2010:
Ok, I've just realized I can do that without merging let me play with it a bit :)

grzesiek2010:
@seadowg I tested `waitForIdleSync()` but it doesn't fix the problem. The only way that seems pretty reliable is using `Thread.sleep()` what I did. I know we don't like it but 5 test passed in a row on Test Lab so looks as if it works.

seadowg:
Ach that's annoying. One thing I've done when I've run into a scenario like this is use a "looping wait" rather than a solo `sleep`. In a test that's waiting for some text to appear that says "Hello" this would look like this:

```java
waitFor(() -> {
    onView(withText("Hello")).check(matches(isDisplayed()));
});
```

The actual implementation of the method could be:

```java
public static void waitFor(Runnable runnable) throws TimeoutException {
    long startTime = System.currentTimeMillis();

    do {
        try {
            runnable.run();
            return;
        catch (Exception e) {
            Thread.sleep(500);
        }
    } while ((System.currentTimeMillis() - startTime) < TIMEOUT);

    throw new TimeoutException();
}
```

This is a bit nicer than a solo `sleep` as we check several times and then get out as early as we can. This means on a fast device we probably don't wait as long as we would have set the sleep and a slow device will pass without needing to bump the sleep time up.

For these tests we could use a `waitFor` with an assertion on what we expect the screen to look like after the rotation.

What do you think?

seadowg (changes requested):
As earlier comments - just making sure this is in the right state!

grzesiek2010:
I tried improving the solution but to no avail: https://github.com/opendatakit/collect/commit/ca06a2ce2973806266957b90f292f3195dcc3331
looks as if ids are available before and the method returns too quickly.
I talked to @seadowg and we decided to leave it as-is for now and maybe in the future we will be able to do it better.

@seadowg I resolved conflict so you can merge.

seadowg:
Looks good! Gave this a run on test lab as well.

seadowg (approved):

PR #3282, opened 2019-08-02
Upgrade OKHttp3 to allow unicode in headers

### LINKED ISSUE #3281: OkHttp failure when submission filename contains unicode
Collect v1.23.0

#### Problem description
Submission filenames are based on the form name which can contain unicode. Trying to send a submission with such a form name causes a crash.

See reproduction steps and full error message at https://forum.opendatakit.org/t/unable-to-upload-odk-record-to-server/21230

lognaturel:
Closes #3281 

Successfully submitted an instance from the form definition in the issue.

I was going to add a test for this but since it turns out to be an OKHttp3 bug (https://github.com/square/okhttp/issues/4564) and [a test was introduced for the fix](https://github.com/square/okhttp/pull/4618/files#diff-c0fbf53145399b63adc71321a0d0b864R275), I don't think we need to add one at this level.

Since the OKHttp fix was backported to OKHttp3 (which we need to use because we support API levels lower than 21), this is the simplest possible fix.

The goal of this PR is to allow users to send submissions with filenames that contain unicode. Since the fix involves an update to the HTTP library, there could be side effects in anything that involves data transfers. I've reviewed [the changelog](https://github.com/square/okhttp/compare/okhttp_3.12.x) and everything looks safe, well-tested, and like it will provide user benefit.

See issue.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,5 +51,5 @@ ext {
     daggerVersion = '2.16'
     rxLifecycleVersion = '2.2.1'
     workManagerVersion = "1.0.0-alpha10"
-    okhttp3Version = '3.12.0'
+    okhttp3Version = '3.12.3'
 }


lognaturel:
I don't think there's much meaningful review or QA to do here since it's a dependency bug. I propose we release this as soon as possible since users are likely to run into it and then QA can evaluate and decide whether it makes sense to do a quick regression check.

yanokwa:
Ordinarily, I'd want to hold a hot fix a few days after a release, but this affects submissions and there are a likely a number of submissions with Unicode.

yanokwa (approved):

seadowg:
For the moment I agree it's an OkHttp responsibility but it might still be nice long term to add something to the`OpenRosaXRequestTest` classes as they are technically independent of the underlying HTTP client. This would give us protection if we move to another HTTP client in the (hopefully far future). What do you think? Happy to take that on or write an issue up (could be a "good first issue").

lognaturel:
@seadowg I'm not convinced that we need to have an HTTP-specific test but sleeping on this has led me to think we should have a form with a unicode name involved in some end-to-end regression test (which would include submission). I think it's the kind of thing that is likely to bite us in some other unexpected way. I had actually completely forgotten that the form name as opposed to form_id was used to create the submission filename. That means we have paths that have unicode in them and I can imagine all kinds of odd things happening if we, for example, make changes to the storage layer. Will be interested to hear QA's thoughts on where that best fits in.

mmarciniak90:
**Tested with success on v1.23.1 from Play Store**

Verified form: [ReunionRecuperacion.xml.txt](https://github.com/opendatakit/collect/files/3466337/ReunionRecuperacion.xml.txt)

Verified on Android: 9.0, 8.1, 7.0, 6.0, 5.1, 4.4, 4.2

Verified cases:
- send with success the form to Aggregate
- send with success the form to Google Sheet

seadowg:
@lognaturel yeah that makes sense. There are a lot of places for something like that to go wrong.

PR #3286, opened 2019-08-05
Don't attempt to write user agent in Robolectric tests

seadowg:
This was causing a (non fatal) error which made test output confusing.

Ran `testDebug` locally to make sure no tests relied on OSMDroid's user agent being configured.

We could pull the OSMDroid stuff and inject a fake/mock in test but more the moment using an override with Robolectric's test application object seems the simplest if we can get away with just not calling the offending method.


lognaturel (approved):

lognaturel:
Whoops, thank you! I had not looked at the output.

PR #3163, opened 2019-06-17
Switch to OKHttp

seadowg:
Closes #2869 

#### What has been done to verify that this works as intended?

Tests run locally and a quick play with the app on the emulator. Also ran the build through test lab as a precaution.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

This should not effect users at all but as we're changing the backbone of the communication with Open Rosa servers it would be good to check Collect thoroughly against Aggregate/Central (and any other common Open Rosa servers).

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #3289, opened 2019-08-06
Ensure OpenRosa requests use correct auth schemes

### LINKED ISSUE #3287: Introduction of OkHttp changed authorization type over https
Hi,

The authorization type changed from **Digest** to **Authorization-Basic** over HTTPS when OKHttp was introduced.

@lognaturel / @yanokwa  do you post anywhere potential breaks for 3rd party software that uses ODK Collect (Like Ona, kobotoolbox, etc) when introducing changes?

seadowg:
Closes #3287.

Added tests based on the docs [here](https://docs.opendatakit.org/openrosa-authentication/) around proactively sending credentials. At first it seemed like nothing was wrong but then I discovered that we ran into a problem only when using HTTPS (because `Basic` auth isn't enabled over HTTP).

It seems the key to problem here is that `BasicAuthenticator` was being added before `DigestAuthenticator` to our list of `CachingAuthenticator`s. This meant that when we made a request our setup would try and proactively send credentials and use the first `Authenticator` that was able to build an authenticated request. This would result in it always sending a `Basic` attempt as `BasicAuthenticator` can always build a header - it just needs the credentials. It's not a problem the other way round as `DigestAuthenticator` is unable to build a header without having previously authenticated a request - it needs realm, nonce etc. 🙄

Aggregate didn't seem to have a problem with `Basic` being sent before `Digest` as it would just return a `Digest` challenge. I don't think the behavior was actually "incorrect" but it would create a bunch more network chatter and went against what you'd expect from the docs which had led to problems for a 3rd party server.

The key was writing and green-ing `withCredentials_onceDigestChallenged_whenHttps_proactivelySendsCredentials` as this was the only failing test added and also matched the behaviour seen in #3287. I was also able to reproduce and verify using the Android Studio Network profiler.

For now I think this is good but I'd like to follow up with work to extract some of the authentication logic (at least the tests) so the code around that is easier to read and debug. It'd also be nice to not have to duplicate authentication tests between all the different HTTP verbs (right now only GET checks this case but we know the authentication is the same for all verbs).

This should only have a positive effect of reducing the number of network requests made to carry out unnecessary authentication but it would be good to test this against known Open Rosa servers with Digest and Basic auth.

For testing it would be good to do the same pass as we carried out on #3163.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
@@ -277,13 +277,13 @@ public class OkHttpConnection implements OpenRosaHttpInterface {
         }
 
         Credentials cred = new Credentials(credentials.getUsername(), credentials.getPassword());
-        DispatchingAuthenticator.Builder daBuilder = new DispatchingAuthenticator.Builder();
 
+        DispatchingAuthenticator.Builder daBuilder = new DispatchingAuthenticator.Builder();
+        daBuilder.with("digest", new DigestAuthenticator(cred));
         if (scheme.equalsIgnoreCase("https")) {
             daBuilder.with("basic", new BasicAuthenticator(cred));
         }
 
-        daBuilder.with("digest", new DigestAuthenticator(cred));
         DispatchingAuthenticator authenticator = daBuilder.build();
 
         initializeHttpClient(); // Need to initalise the http client again to get rid of the cached credentials
--- a/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/OkHttpConnection.java
@@ -276,9 +276,7 @@ public class OkHttpConnection implements OpenRosaHttpInterface {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


grzesiek2010 (approved):

mmarciniak90:
Verified on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1

List of verified cases:

* Aggregate

   * Upgrade when Aggregate 2.0.3 is configured
   * Aggregate 2.0.3 download and send
   * 1000 submissions send to Aggregate
   * Automatically download the upgraded version of the form
   * Autosend Aggregate 2.0.3
   * Aggregate 1.7 download, send and autosend
   * Autosend specified in form
   * Autosend submission with media

* Google Drive

   * Upgrade when Google Drive is configured
   * 1000 submissions send to Google sheet (crash reported in #3296)
   * Download form from Google Drive
   * Send to Google Sheet
   * Autosend to Google Sheet
   * Send the form with specified submission URL

* Central

   * Central download and send
   * 1000 submissions send to Central 
   * Autosend Central

* Other
   * Ona download and send
   * Ona autosend
   * Kobo download and send
   * Kobo autosend

PR #3302, opened 2019-08-08
Extract authentication and common headers from OkHttpConnection

seadowg:
This is a follow up to #3289 and is based on the [post mortem of the issue](https://gist.github.com/seadowg/925999bd28504fc97efa918b0a5d85d9).

The two key changes here are:

1. Remove static state from `OkHttpConnection` and have it be an `@Singleton` instead. This keeps the state that needs to persist between instances of `CollectServerClient` but stops tests being affected by the static variables and also should make the class slightly easier to reason about.

2. Extract the authentication and common header logic out of `OkHttpConnection` and into `OkHttpOpenRosaServerClientFactory`. This also means our tests for that behaviour move to that object (which don't need Robolectric). I've kept these tests in a contact style so we don't lose the ability to switch HTTP clients in the future.

Ran tests locally and played around with the app. Also had a check with the Network Profiler to make sure we weren't generating any extra/unexpected traffic.

We're still not in the best possible world as it's still unclear where the boundaries sit between `CollectServerClient` and `OpenRosaHttpInterface` but I think this is a step in the right direction as we know have a class that is responsible for communicating with an Open Rosa server that sits outside of the Android/Collect world.

There shouldn't be any change in behaviour. Is still think it would be good to run the QA checks we did for #3289 as we don't yet have E2E tests that actually hit a server.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
@@ -18,6 +18,7 @@ import org.odk.collect.android.utilities.Clock;
 
 import java.io.IOException;
 import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
@@ -90,14 +91,14 @@ public class OkHttpOpenRosaServerClientFactory implements OpenRosaServerClientFa
             return client.newCall(request.newBuilder()
                     .addHeader(USER_AGENT_HEADER, userAgent)
                     .addHeader(OPEN_ROSA_VERSION_HEADER, OPEN_ROSA_VERSION)
-                    .addHeader(DATE_HEADER, getHeaderDate())
+                    .addHeader(DATE_HEADER, getHeaderDate(clock.getCurrentTime()))
                     .build()).execute();
         }
 
-        private String getHeaderDate() {
+        private static String getHeaderDate(Date currentTime) {
             SimpleDateFormat dateFormatGmt = new SimpleDateFormat("E, dd MMM yyyy hh:mm:ss zz", Locale.US);
             dateFormatGmt.setTimeZone(TimeZone.getTimeZone("GMT"));
-            return dateFormatGmt.format(clock.getCurrentTime());
+            return dateFormatGmt.format(currentTime);
         }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
@@ -87,15 +87,11 @@ public class OkHttpOpenRosaServerClientFactory implements OpenRosaServerClientFa
 
         @Override
         public Response makeRequest(Request request) throws IOException {
-            return client.newCall(addHeaders(request)).execute();
-        }
-
-        private Request addHeaders(Request request) {
-            return request.newBuilder()
+            return client.newCall(request.newBuilder()
                     .addHeader(USER_AGENT_HEADER, userAgent)
                     .addHeader(OPEN_ROSA_VERSION_HEADER, OPEN_ROSA_VERSION)
                     .addHeader(DATE_HEADER, getHeaderDate())
-                    .build();
+                    .build()).execute();
         }
 
         private String getHeaderDate() {
--- a/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
@@ -14,10 +14,10 @@ import com.burgstaller.okhttp.digest.DigestAuthenticator;
 import org.odk.collect.android.http.HttpCredentialsInterface;
 import org.odk.collect.android.http.openrosa.OpenRosaServerClient;
 import org.odk.collect.android.http.openrosa.OpenRosaServerClientFactory;
+import org.odk.collect.android.utilities.Clock;
 
 import java.io.IOException;
 import java.text.SimpleDateFormat;
-import java.util.Date;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
@@ -39,9 +39,11 @@ public class OkHttpOpenRosaServerClientFactory implements OpenRosaServerClientFa
     private static final String DATE_HEADER = "Date";
 
     private final OkHttpClient.Builder baseClient;
+    private final Clock clock;
 
-    public OkHttpOpenRosaServerClientFactory(@NonNull OkHttpClient.Builder baseClient) {
+    public OkHttpOpenRosaServerClientFactory(@NonNull OkHttpClient.Builder baseClient, Clock clock) {
         this.baseClient = baseClient;
+        this.clock = clock;
     }
 
     @Override
@@ -68,17 +70,19 @@ public class OkHttpOpenRosaServerClientFactory implements OpenRosaServerClientFa
                     .addInterceptor(new AuthenticationCacheInterceptor(authCache)).build();
         }
 
-        return new OkHttpOpenRosaServerClient(builder.build(), userAgent);
+        return new OkHttpOpenRosaServerClient(builder.build(), userAgent, clock);
     }
 
     private static class OkHttpOpenRosaServerClient implements OpenRosaServerClient {
 
         private final OkHttpClient client;
         private final String userAgent;
+        private final Clock clock;
 
-        OkHttpOpenRosaServerClient(OkHttpClient client, String userAgent) {
+        OkHttpOpenRosaServerClient(OkHttpClient client, String userAgent, Clock clock) {
             this.client = client;
             this.userAgent = userAgent;
+            this.clock = clock;
         }
 
         @Override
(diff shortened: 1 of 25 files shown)


collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientFactoryTest.java
```diff
@@ -55,6 +66,23 @@ public void sendsOpenRosaHeaders() throws Exception {
         assertThat(request.getHeader("X-OpenRosa-Version"), equalTo("1.0"));
     }
 
+    @Test
+    public void sendsDateHeader() throws Exception {
```
lognaturel, line 70: This test verifies three things -- that a date header is added, that the date header's format is correct and that the `getCurrentTime` method is called. I strongly agree with verifying the first one. I think the second is not a bad idea as documentation of intended behavior (especially that the date is always in English since that has caused problems). I'm not sure the third is all that meaningful. It's the third that requires introducing the `Clock` interface and that feels like dependency inversion gone a bit too far to me. Naturally, I can be convinced otherwise. 

Would you consider, instead of verifying that the date/time in the header is an exact match to a date/time produced by a mock, just checking the format of the header date/time? I think you could do that by parsing the date string with the desired formatter, formatting the resulting Date and ensuring it matches the original string.

collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+            this.userAgent = userAgent;
+            this.clock = clock;
+        }
+
+        @Override
+        public Response makeRequest(Request request) throws IOException {
+            return client.newCall(addHeaders(request)).execute();
+        }
+
+        private Request addHeaders(Request request) {
```
lognaturel: How about making this static to make it really clear that it doesn't use anything from the instance? Maybe even calling it `withHeaders`? Or leaving it inline in `makeRequest`? For some reason I got tripped up for a second reading it.

collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+            return request.newBuilder()
+                    .addHeader(USER_AGENT_HEADER, userAgent)
+                    .addHeader(OPEN_ROSA_VERSION_HEADER, OPEN_ROSA_VERSION)
+                    .addHeader(DATE_HEADER, getHeaderDate())
+                    .build();
+        }
+
+        private String getHeaderDate() {
```
lognaturel: This would need to be static too which feels right as well.

lognaturel (commented):
Agreed this is a good change. One tiny thing that tripped me up (maybe not worth doing anything about) and one more philosophical question.

collect_app/src/main/java/org/odk/collect/android/http/okhttp/OkHttpOpenRosaServerClientFactory.java
seadowg: Right yeah I can totally see that. I think originally it was being called a few times but maybe now inlining is best. I'll give it a go.

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientFactoryTest.java
seadowg: I can see your argument about it feeling like the test is checking a call to `getCurrentTime`. I didn't really intended it that way but more as a verification we are using the right time. Mockito maybe muddies that though. I'm not sure I like the idea of just checking the format. Looking at the [specs for a request](https://docs.opendatakit.org/openrosa-http/#http-requests) as far as I can see we need the `Date` header to be the "The date on the device" and also in that format. Given that it does feel like the things we need to drive out with a test are:

1. There is a date header
2. The date is in the right format
3. The date represents the correct point in time - the device's "now"

I can imagine an argument that we don't need the time to be dead on and that we could use fuzzy "time is within/near" matchers but I'm never quite comfortable introducing that non determinism when we can just treat time like the variant that it is and have it as a dependency on our object/function. 

I can totally see that the `Clock` seems like overkill for this one object but I'd be surprised if it wouldn't be useful in other objects and their tests (any of the time based Widgets for example). Do you think maybe using a "fake" would be nicer? This is a pattern I've actually used more than Mockito for something like this:

```java
class SettableClock implements Clock {

    private Date time;

    @Override
    public Date getTime() {
        if (time != null) {
            return time;
        } else {
            return new Date();
        }
    }

    public void setTime(Date time) {
        this.time = time;
    }
}

fakeClock.setTime(new Date());
subject = new OkHttpOpenRosaServerClientFactory(baseClient, clock);
```

seadowg:
Rebasing this to double check it fixes unit test problems we're seeing on master

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientFactoryTest.java
lognaturel: I guess there are two separate things that I don't love: that the mock doesn't make it feel like the test really checks for current time and that the `Clock` dependency comes in the `OkHttpOpenRosaServerClientFactory` constructor.

For the first point, it's too bad we can't use `java.time`'s `Clock` (because of API level). It's of course very similar to what you have but the fact that the interface comes from the same package and that everything is designed to work together feels better. The fake vs. mock approach doesn't make a big difference to me on that dimension. Maybe it's the best that can be done.

For the second, I was hoping that perhaps the dependency could come in more locally but I guess that with `OkHttpOpenRosaServerClient` being private static, there's not really much that can be done there.

Ok, I give up, this is probably as good as can be.

lognaturel:
@opendatakit/testers This would be a great one to prioritize since it fixes the build! 🙏 Up to you to decide whether you think it's better to do all the #3289 regression checks now or whether to do a few high-level checks, merge, and then know that there will be another pass during regression testing. I think the risk here is relatively low. I did try to follow everything that was moved around but there is a risk that somehow something accidentally got removed that isn't covered by a test.

lognaturel (approved):

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientFactoryTest.java
seadowg: Yeah I kind of get you here. Having this "system level" abstraction here maybe feels a bit much. I think I'm maybe jumping ahead because I've used a `Clock` abstraction before but usually at more of a `Fragment`/`Activity` level. 

Maybe the way to do it at this level is just have `currentTime` come in as an argument to `makeRequest`. Seeing as we have this approved now I'll hold off but I'm happy to follow up with an attempt at something we can be happier with 😁

lognaturel: > Having this "system level" abstraction here maybe feels a bit much.

Precisely.

Agree that an argument to `makeRequest` would be great. But you'd have to restructure things since you can't test `makeRequest` directly, right? Sounds good to keep thinking about these kinds of things for future improvements. It's good to have had the conversation so we get synced up on each other's taste.

I think this kind of passing around of low-level dependencies through a bunch of layers is how people end up feeling like they need a DI framework so I want to be mindful of it.

seadowg: It's pretty doable as `makeRequest` is called from the tests as it's the method on the `OpenRosaServerClient`! I'll play around with it when we get this in 👍

mmarciniak90:
@seadowg I noticed that auto-send fails with this message. 
I checked it on 
* Aggregate v1.7 
* Aggregate v2.0
* Central
* ONA

seadowg:
@mmarciniak90 hmm interesting. Thanks for that I'll check it out!

mmarciniak90:
@seadowg yes, it is! because this problem is not visible when I send it manually or when auto-sent is to Google Sheet

seadowg:
@mmarciniak90 could you give me the steps you go through to get to this? I'm not as familiar with the autosend functionality.

mmarciniak90:
Repro steps
1. Go to `General settings` -> `Form management`
2. In `Auto send` option select eg. `Wifi or cellular`
3. Fill some form and wait 
4. A notification with auto-send status appears

seadowg:
@mmarciniak90 yeah seeing exactly the same thing. It looks like the autosend functionality has been broken a small mistake in the code. I'm double checking our Espresso tests as I'm surprised nothing caught this. Will fix the problem and then ping you!

mmarciniak90:
Take a look also on case where auto-send is set directly in the form.

Repro steps:

1. Add form [Auto_Submit_True_Auto_Delete_True.xml.txt](https://github.com/opendatakit/collect/files/3496410/Auto_Submit_True_Auto_Delete_True.xml.txt) to Aggregate and download it in Collect
2. Go to General settings -> Form management
3. In Auto send option select OFF
4. Fill this form and wait
5. A notification with auto-send status appears

seadowg:
@mmarciniak90 ok that should be fixed now!

mmarciniak90:
@seadowg yes, I confirm that this change fixed described problem. I'm continuing the testing.

mmarciniak90:
**Tested with success**

**Verified on Android**: 4.2, 4.4, 6.0, 7.0, 8.1
I slightly reduced the number of cases compared with #3289 

**List of verified cases:**

* Aggregate

   * Upgrade when Aggregate 2.0.3 is configured
   * Aggregate 2.0.3 download and send
   * 1000 submissions send to Aggregate (verified on Android 7.0, 5.1, 4.2)
   * Automatically download the upgraded version of the form
   * Autosend Aggregate 2.0.3
   * Aggregate 1.7 download, send and autosend
   * Autosend specified in form
   * Autosend submission with media

* Google Drive

   * 1000 submissions send to Google sheet (verified on Android 8.1, 6.0, 4.4)
      * noticed error : Too many requests per 100 seconds
   * Download form from Google Drive
   * Send to Google Sheet
   * Autosend to Google Sheet
   * Send the form with specified submission URL

* Central 

   * Central download and send
   * 1000 submissions send to Central (verified on Android 7.0, 4.2)
   * Autosend Central

* Other
   * Ona download and send
   * Ona autosend
   * Kobo download and send
   * Kobo autosend

PR #3311, opened 2019-08-19
Fix basemap attribution text disappearing

### LINKED ISSUE #3263: Basemap attribution text disappears on rotation and zoom for geopoint with maps and placement-map appearances
Collect v1.23.0

#### Problem description
#3255 added attribution text for basemap options from sources other than Mapbox and Google. These basemap options are displayed by the osmdroid mapping engine. Rotating the screen or zooming in to the current location makes these disappear.

1. Select a basemap source other than Mapbox or Google in General Settings > Maps > Basemap source.
1. Open a form such as All Widgets on the default server with geopoint questions with appearance maps and placement-map
1. Tap on Start GeoPoint on one of those questions
1. Notice that the attribution text is in the lower right corner
1. Rotate the screen
1. Notice that the attribution text is gone
1. Tap the back button
1. Tap on Start GeoPoint
1. Notice that the attribution text is in the lower right corner
1. Zoom to current position

#### Expected behavior
Attribution text should always be visible as it is with the geotrace and geoshape question types.

#### Other information 
This looks like a lifecycle management issue that may go deeper than just this.

nshestiuk:
Closes #3263

I reproduced basemap attribution text disappearing on Android 7.1 (text disappears on rotation, zooming and also when the pointer is placed) and Android 9 (text disappears only when the pointer is placed) and confirmed that this PR fixes the issue on both.

**Android 9** (text disappears only when the pointer is placed)
with_issue
**Android 9** (fixed)
fixed

I've added missed code. That is obvious and simple solution.

I do not see any risk of regression.

no
no

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -155,8 +155,7 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
         map.getController().setCenter(toGeoPoint(INITIAL_CENTER));
         map.getController().setZoom((int) INITIAL_ZOOM);
         map.setTilesScaledToDpi(true);
-        map.getOverlays().add(new AttributionOverlay(getContext()));
-        map.getOverlays().add(new MapEventsOverlay(this));
+        addAttributionAndMapEventsOverlays();
         loadReferenceOverlay();
         addMapLayoutChangeListener(map);
         myLocationOverlay = new MyLocationNewOverlay(map);
@@ -302,7 +301,7 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
 
     @Override public void clearFeatures() {
         map.getOverlays().clear();
-        map.getOverlays().add(new MapEventsOverlay(this));
+        addAttributionAndMapEventsOverlays();
         map.getOverlays().add(myLocationOverlay);
         map.invalidate();
         features.clear();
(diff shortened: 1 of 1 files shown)


collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -302,6 +302,7 @@
 
     @Override public void clearFeatures() {
         map.getOverlays().clear();
+        map.getOverlays().add(new AttributionOverlay(getContext()));
```
seadowg: It looks like we have duplicated work we need to do to set up the view in both `clearFeatures` and `onCreateView` (this line and the line below for instance). I think it'd be good to extract this to a shared method so we're less likely to run into inconsistencies in the future. Does that make sense?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
nshestiuk: Done. 
If you have better idea for the method name, I whoud be happy to update PR

seadowg: 👍 Seems good for now!

mmarciniak90:
**Tested with success**

Verified cases:
* OSM, USGS, Stamen, and CARTO maps
* zoom
* rotation
* GeoPoint widget
* GeoShape and GeoTrace - no regression noticed

seadowg (approved):

lognaturel:
Nice to see you again, @nshestiuk, and thanks for taking this on!

PR #3312, opened 2019-08-19
Automated tests and fixes for existing ones

kkrawczyk123:
Adding new regression automated tests and fixing existing ones.


seadowg:
@kkrawczyk123 Oh wow this is an interesting one. I pulled this down and saw that the PMD error is `Avoid calling finalize() explicitly`. It looks like line 47 in `FormManagementTest` is incorrectly triggering this check as the string resource used is `R.string.constraint_behavior_on_finalize`. 

I'm guessing the simplest thing to do is change the string name but I'm surprised this broke!

grzesiek2010:
Yes, it's a pmd bug I even tried to update pmd but the problem is not solved so renaming if you can come up with a good name is fine or suppressing otherwise.

mmarciniak90 (approved):

grzesiek2010 (approved):

seadowg:
@grzesiek2010 sorry I'd talked to @kkrawczyk123  but not posted here about it looking like we had Test Lab failures with these changes. Should we revert this until we know the suite is green? It looks like the fix we needed for `spinnerList_ShouldDisplay` is something we could bring in seperately.

grzesiek2010:
Let's keep it, for now, I'll look into it.

PR #3314, opened 2019-08-20
Hierarchy tests

grzesiek2010:
I just confirmed that tests work as expected.

Form hierarchy is a tricky part of the app. We have had a lot of issues in that area like:
https://github.com/opendatakit/collect/issues/2871
https://github.com/opendatakit/collect/issues/2944
https://github.com/opendatakit/collect/issues/2936
https://github.com/opendatakit/collect/issues/2942

since the functionality is not completed and we will work on the code I added some tests to avoid breaking something that we have already fixed.

It doesn't require testing.

No.

No.


collect_app/src/androidTest/java/org/odk/collect/android/formentry/FormHierarchyTest.java
```diff
@@ -0,0 +1,143 @@
...
+        FormEntry.clickGoToIconInForm();
+        onView(withId(R.id.list)).check(matches(RecyclerViewMatcher.withListSize(R.id.list, 3)));
+        FormEntry.clickOnText("Group 1");
+        onView(withId(R.id.list)).check(matches(RecyclerViewMatcher.withListSize(R.id.list, 3)));
+        FormEntry.checkIfTextDoesNotExist("Repeat Group 1");
+    }
+
+    @Test
+    //https://github.com/opendatakit/collect/issues/2942
+    public void test4() {
```
seadowg: I think it would be nice for these tests to be named around what feature they are driving out. For instance could this one be `canDeleteLastGroup_whenInNestedRepeat`? It's definitely nice to have the github links but just thinking through if one of these tests broke it would be nicer (for me personally) to be able to get a sense of what this test was doing quickly.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/RecyclerViewMatcher.java
```diff
@@ -69,4 +65,17 @@ public boolean matchesSafely(View view) {
             }
         };
     }
+
+    public static Matcher<View> withListSize(int listId, final int size) {
```
seadowg: This is nice. 

One thing I'm unclear on: Are you intending this to match the total number of items "on screen" or in the `RecyclerView`'s adapter though? If it's the former maybe we should rename to like "withVisibleItems" or something? Otherwise, I believe it needs to call `recyclerView.getAdapter().getItemCount()` instead.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/RecyclerViewMatcher.java
```diff
@@ -69,4 +65,17 @@ public boolean matchesSafely(View view) {
...
+        return new TypeSafeMatcher<View>() {
+            @Override public boolean matchesSafely(final View view) {
+                RecyclerView recyclerView = view.getRootView().findViewById(listId);
```
seadowg: Will the `View` pass here not be your `RecyclerView` (assuming you give `onView` the right matchers)?

seadowg (changes requested):
Just a couple of wee changes. Also ran this through Test Lab and it looked good 👍

collect_app/src/androidTest/java/org/odk/collect/android/formentry/FormHierarchyTest.java
grzesiek2010: Improved.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/RecyclerViewMatcher.java
grzesiek2010: Right, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/RecyclerViewMatcher.java
grzesiek2010: You are right I need all elements, fixed.

grzesiek2010:
>Just a couple of wee changes. Also ran this through Test Lab and it looked good

Thanks, I added changes.

seadowg (approved):

PR #3317, opened 2019-08-26
Pass activity to general settings Espresso helper

seadowg:
Fixes broken tests in `FillBlankFormTest` that were failing due to the test loading strings in a different language from the app.

This change lets us use the `Activity` under test as the source of resources so we don't run into problems with locale changes.

Run tests locally and on Test Lab.

I didn't feel comfortable changing the way the app switches `Locale` config as the feature does seem to be working. I think after we have this merged it would be good to look at having our Espresso helpers use a page object style pattern so we can pass the `Activity` in at a constructor but that would be quite a large diff so thought it would be best to avoid as a quick fix.

Only changes tests so shouldn't be any regression risks.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
@@ -158,7 +158,7 @@ public class FillBlankFormTest extends BaseRegressionTest {
 
         //TestCase37
         MainMenu.clickOnMenu();
-        MainMenu.clickGeneralSettings();
+        MainMenu.clickGeneralSettings(main.getActivity());
         Settings.clickOnUserInterface();
         Settings.clickOnLanguage();
         Settings.clickOnSelectedLanguage("Deutsch");
@@ -196,7 +196,7 @@ public class FillBlankFormTest extends BaseRegressionTest {
         pressBack();
         pressBack();
         MainMenu.clickOnMenu();
-        MainMenu.clickGeneralSettings();
+        MainMenu.clickGeneralSettings(main.getActivity());
         Settings.clickOnUserInterface();
         Settings.clickOnLanguage();
         Settings.clickOnSelectedLanguage("English");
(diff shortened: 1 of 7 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
```diff
@@ -158,7 +158,7 @@ public void sortByDialog_ShouldBeTranslatedAndDisplayProperIcons() {
 
         //TestCase37
         MainMenu.clickOnMenu();
-        MainMenu.clickGeneralSettings();
+        MainMenu.clickGeneralSettings(main.getActivity());
```
grzesiek2010, line 161: ~~Isn't it a context of MainMenuActivity? Probably it would work as well in this case but seems strange. What about using [ActivityHelpers.getActivity()](https://github.com/opendatakit/collect/blob/master/collect_app/src/androidTest/java/org/odk/collect/android/support/ActivityHelpers.java#L22) which returns the current activity?~~

ohh my mistake I thought it's used in FormEntry but it's in MainMenu so that's fine.

grzesiek2010:
You solved just the issue with clickGeneralSettings() method but theoretically, it's a broader problem which can take place in each method we use strings so maybe we should devise a broader solution instead of fixing such bugs one by one?

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
seadowg: Oh we could use that in future though. I knew there must be a way of getting the "current" Activity rather than just the one on the test rule. That'll be useful thanks!

seadowg:
@grzesiek2010 completely agree this needs a broader solution 👍. My thinking with this PR was just the absolute simplest change to get the build fixed. I'd then started working on a follow up that moves our helpers to more of a [page object](https://www.martinfowler.com/bliki/PageObject.html) setup where the pages extend an abstract class with the basic helpers like a `getString()` (that is safe for translations). I've pushed a draft of it (#3320) so you can see it!

I'd still vote for merging this as a specific fix so we're working in the green again and then we can get the general solution in.

grzesiek2010 (approved):

PR #3320, opened 2019-08-27
Introduce Page Objects to Espresso tests

seadowg:
Follow up to #3317.

This introduces helpes to Espresso in the from of [Page Objects](https://www.martinfowler.com/bliki/PageObject.html) as a replacement of the static helpers we currently use. I've only converted one test for the moment (`FillBlankFormTest`).

We'd talked before about doing this eventually but the changes in #3317 made it feel like a good time as we can hold references in page objects to things we might need (rather than pass them to the helpers everytime) - in that case it was the `ActivityTestRule` for fetching the `Activity` and then translated strings. There are a few other advantages to them (especially using a fluent API style) in my mind:

1. Having common helpers (`pressBack` for example) exposed on a super type makes them more discoverable than a static helper in my opinion.
2. We get documentation about what "actions" can be performed on a page which can make it easier to write (we get automcomplete while chaining toghether) and understand tests.
3. We also get documentation about what operations transition from one page to another (these operations return a different page type) - this is really helpful when reading/debugging a test and trying to understand where the test **thinks** it is at any one point.
4. Use the `assertOnPage` method on actions that do transition between pages can give us more robust tests - we should probably always make an assertion that we're on the page we think we are but let's face it, it's easy to forget.
4. It can make it easier to make navigational changes in the app. For instance, changing the return type of a method from one page type to another would cause compilation errors in the tests now.

Ran tests locally.

I did consider doing a full sweep but wanted to prioritize discussing this (and hopefully get it in). Have the static helpers clearly deprecated should mean we can move to the new style gradually.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
@@ -17,6 +17,9 @@ import static org.hamcrest.CoreMatchers.not;
 /**
  * Base class for Page Objects (https://www.martinfowler.com/bliki/PageObject.html)
  * used in Espresso tests. Provides shared helpers/setup.
+ *
+ * This class's generic type is designed to it's extending class so that shared methods
+ * can return the extending class.
  */
 
 abstract class Page<T extends Page<T>> {
@@ -27,6 +30,10 @@ abstract class Page<T extends Page<T>> {
         this.rule = rule;
     }
 
+    /**
+     * Presses back and then returns the Page object defined by
+     * `destination`.
+     */
     public <D extends Page> D pressBack(Class<D> destination) {
         Espresso.pressBack();
 
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
@@ -15,11 +15,20 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
 
 /**
- * Base class for Page Objects (https://www.martinfowler.com/bliki/PageObject.html)
- * used in Espresso tests. Provides shared helpers/setup.
+ * Base class for Page Objects used in Espresso tests. Provides shared helpers/setup.
+ * <p>
+ * Sub classes of {@code Page} should represent a page or part of a "page" that the user would
+ * interact with. Operations on these objects should return {@code this} (unless
+ * transitioning to a new page) so that they can be used in a fluent style.
+ * <p>
+ * The generic typing is a little strange here but enables shared methods such as
+ * {@link Page#closeSoftKeyboard()} to return the sub class type rather than {@code Page} so
+ * operations can be chained without casting. For example {@code FooPage} would extend
+ * {@code Page<FooPage>} and calling {@code fooPage.closeSoftKeyboard()} would return
+ * a {@code FooPage}.
  *
- * This class's generic type is designed to it's extending class so that shared methods
- * can return the extending class.
+ * @see <a href="https://www.martinfowler.com/bliki/PageObject.html">Page Objects</a>
+ * @see <a href="https://en.wikipedia.org/wiki/Fluent_interface">Fluent Interfaces</a>
  */
 
 abstract class Page<T extends Page<T>> {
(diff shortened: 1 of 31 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
```diff
@@ -67,108 +68,114 @@
...
-        MainMenu.clickMenuFilter();
-        MainMenu.searchInBar("Aaa");
-        pressBack();
-        pressBack();
-
+        new MainMenuPage(main)
+                .clickFillBlankForm()
+                .clickMenuFilter()
+                .searchInBar("Aaa")
+                .pressBack(new BlankFormSearchPage(main))
```
seadowg, line 99: We could also have each Page implement its own `pressBack` (similar to `assertOnaPage`) but in my experience where "back" is ends up being pretty dynamic and I also like how this ends up giving you extra documentation in the test around where you are.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
```diff
@@ -340,79 +353,79 @@ public void values_ShouldBeRandom() {
...
+        new MainMenuPage(main)
+                .startBlankFormWithError("g6Error")
+                .clickOK(new FormEntryPage("g6Error", main))
+                .swipeToNextQuestion()
+                .clickSaveAndExit()
+                .checkIsToastWithMessageDisplayed("Form successfully saved!");
+
+        new MainMenuPage(main).startBlankForm("g6Error2")
+                .putText("bla")
+                .swipeToNextQuestionWithError()
```
seadowg, line 365: I've adopted this pattern of giving actions a `WithError` suffix and returning an `ErrorDialog` when we expect there to be an error. A lot of other ways to handle this but let's see how this evolves.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
```diff
@@ -17,6 +17,9 @@
 /**
  * Base class for Page Objects (https://www.martinfowler.com/bliki/PageObject.html)
  * used in Espresso tests. Provides shared helpers/setup.
+ *
+ * This class's generic type is designed to it's extending class so that shared methods
```
lognaturel: it's -> its. But I think there may also be missing words? Maybe "designed to be its"?

How about being a little more explicit and detailed about the pattern? Some ideas on how to frame it:

"When defining a new subclass, you must specify that new subclass name as the type parameter. For example, `FooPage` must extend `Page<FooPage>`. This is so that the methods shared by `Page` subclasses can all return a `Page` of their type, enabling a fluent API. For example, if `FooPage` introduces a `foo()` method, it is possible to write `fooPage.closeSoftKeyboard().foo()` because `closeSoftKeyboard` returns a `FooPage` object."

Seems it would be great to have a `This` keyword or something that makes it possible to refer to the actual type of the object. Maybe there's an obvious reason that wouldn't be a good idea but I can't think of it.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java
lognaturel: I think this exact explanation would be helpful to have on the `pressBack` method.

lognaturel (changes requested):
I've made a couple commenting suggestions but other than that, I like the API a lot. It'd be good to get impressions from @opendatakit/testers and @grzesiek2010 since they'll be using this the most.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
seadowg: > it's -> its

Noooooooo. I feel like I spend half my life thinking about this. Will fix the typos.

> How about being a little more explicit and detailed about the pattern

Yeah I agree. Will improve the docs. 

> Seems it would be great to have a This keyword

Haha yeah the fluent pattern really pushes Java's generics here - defining methods on a super class that return an inheriting type. As a fun aside, Kotlin introduced `inline` and `reified` to give you the ability to turn generic types into "real" types that make all this a little easier: https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters

lognaturel (approved):
Looks great to me! Just need one of @opendatakit/testers and @grzesiek2010 to take a look and make sure they can work with the pattern.

mmarciniak90 (approved):

lognaturel:
@seadowg could you please file an issue with what remains to be done so we don't lose track? Perhaps with a checklist? Thanks!

seadowg:
Added at #3338

PR #3323, opened 2019-08-28
Improved managing databases + automated tests

### LINKED ISSUE #3299: Revisit forms database migrations and add testing
https://github.com/opendatakit/collect/pull/3250 added tests for the instances database migration and fixed a number of migration bugs. Do the same for the forms database as appropriate.

grzesiek2010:
Closes #3299 

Since it's a long-standing issue which has bothered me (and not only me) and it's pretty complicated, it requires a good explanation:

ODK Collect might be installed on the same device along with its forks. If it takes place all apps share the same directory (`sdcard/odk/`). That means the same database file (`forms.db` or `instances.db`) 
might be used by two or more apps. Forks might be out of date compared to ODK Collect and use older database versions. In such a case ODK Collect will upgrade its databases every time it's needed but a fork will downgrade them in similar cases. Theoretically, everything should work fine and `onUpgrade()`/`onDowngrade()` methods should be called properly but there is one problem that breaks it.
The problem is that in [FormsProvider](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java) and [InstanceProvider](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java) classes [dbHelper](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java#L56) objects are singletons (in order to avoid leaks https://github.com/opendatakit/collect/commit/d4201b0721061661c7dbba32dada7adf00459130)
so a case like the one described below is possible:
1. We have ODK Collect and its fork (KoboCollect) installed on the same device.
2. ODK Collect uses forms.db v7 but KoboCollect is out of date and uses forms.db v5.
3. A user starts ODK Collect and performs an operation that uses forms.db eg. clicks on `Fill Blank Form` in the main menu (that creates forms.db v7)
4. A user starts KoboCollect and also performs an operation that uses forms.db like in the previous step (forms.db is downgraded to v5).
5. A user starts ODK Collect and performs an operation that uses forms.db again and it causes a crash.

To understand what causes the crash we need to understand what happens under the hood:

1. In step 3 we created [Collect.FormsProvider.dbHelper](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java#L67) object v7 
2. In step 4 the forms.db file was changed (downgraded) via KoboCollect
3. In step 5 a user tries to use forms.db via Collect again, [Collect.FormsProvider.dbHelper](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java#L66) still exists (the app hasn't been killed so singleton object exists) 
but Collect doesn't know about the fact that KoboCollect downgraded forms.db file, so `onUpgrade()` is not called becuase Collect.FormsProvider.dbHelper still represents db v7 which is expected.

That cause conflicts, there is a crash, the app is killed (Collect.FormsProvider.dbHelper as well of course), so now if a user opens the app again `onUpgrade()` will be called properly because the
obsolete Collect.FormsProvider.dbHelper which claimed that we already have a database v7 no longer exist ([a new one is created](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java#L67)).

I fixed that problem in two first commits:
https://github.com/opendatakit/collect/commit/c4356623ccec5273bba7e3ba79816f3d04519ec6
https://github.com/opendatakit/collect/commit/9aade598ba2076c408c691972b2f36467fcdfdc7
the solution is improved in the next commits but that is the general idea.

I think it wasn't the only problem that could cause crashes. Another one is upgrading forms or instances databases from lower versions but with the same columns, we are trying to add. It might seem crazy a bit but we can't be sure what our forks do with our databases, maybe the downgrade them but don't remove newly added columns etc.
I fixed that problem in this commit https://github.com/opendatakit/collect/commit/197bcaf5c73970a15c91314a15046890cb369fb1

I wasn't able to find anything more but it doesn't mean I'm sure it fixes every issue related to our databases. Time will tell, we will test and merge it and we can wait for reports from firebase.

I performed manual tests and added automated tests.

A perfect (another) solution (in terms of being crash-free) would be to stop sharing the same directory with our forks but I know it would be difficult to achieve and probably some users need that so the would complain. Taking into account that it's not possible, this is the only solution.

There won't be any visible change, everything should work like before but databases should be managed in a better way to avoid crashes.
Forms and instances, databases should be tested so saving forms, upgrading/downgrading, everything related to them.

No.

No.


collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java
```diff
@@ -385,6 +385,12 @@ public int update(@NonNull Uri uri, ContentValues values, String where, String[]
         return count;
     }
 
+    private boolean isDatabaseHelperOutOfDate() {
```
lognaturel: How about either making this static or moving it to `InstanceDatabaseHelper` and calling it something like `databaseNeedsUpgrade`? I have a slight preference for moving it to the helper.

I've suggested making instance methods static a few times recently so I think it's worth taking a moment to chat about whether this is a good idea or not. I personally find it really helpful when all methods that don't relate to object state are static because:
- it's 100% sure there are no unintended side effects
- it's clear that the method could be moved somewhere else if other methods that group with it are discovered
- it may suggest a better place for the method to go (in this case, `InstanceDatabaseHelper`)

I know there's a general dislike for static methods in Java but to me, it's more about methods that aren't object oriented (don't access state) rather than about the `static` keyword. There are more arguments on both sides at https://softwareengineering.stackexchange.com/questions/215826/make-methods-that-do-not-depend-on-instance-fields-static

Also interested in @seadowg's take so we can agree on a general principle and not have to think about it again in the future.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseContext.java
```diff
@@ -35,7 +35,7 @@ public DatabaseContext(String path) {
 
     @Override
     public File getDatabasePath(String name) {
```
lognaturel: This isn't right given the method signature of the base method -- `name` should be used in some way. If you want to pass in the full path to the db file, then I think `DatabaseContext` needs to be redesigned.

collect_app/src/main/java/org/odk/collect/android/utilities/SQLiteUtils.java
```diff
@@ -0,0 +1,18 @@
+package org.odk.collect.android.utilities;
+
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+
+public class SQLiteUtils {
```
lognaturel, line 10: There are now lots of different classes that provide convenience wrappers over SQL queries. Any chance this could fit in an existing one? Maybe `CustomSQLiteQueryBuilder`? The DAOs?

collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
```diff
@@ -92,6 +94,7 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         }
 
         Timber.i("Upgrading database from version %d to %d completed with success.", oldVersion, newVersion);
+        isDatabaseBeingMigrated = false;
```
lognaturel: If there's a failure and rollback, the ongoing migration flag will be true until a relaunch, right? Does that mean that if the failure was caused by something that can be fixed, the user will need to also kill the app and relaunch for the migration to run? If so, that doesn't feel right to me.

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static junit.framework.TestCase.assertTrue;
+import static org.odk.collect.android.database.helpers.FormsDatabaseHelper.FORMS_TABLE_NAME;
+import static org.odk.collect.android.database.helpers.InstancesDatabaseHelper.INSTANCES_TABLE_NAME;
+import static org.odk.collect.android.test.FileUtils.copyFileFromAssets;
+
+public class SQLiteUtilsTest {
+
+    @Test
+    public void doesColumnExistTest() throws IOException {
+        copyFileFromAssets("database" + File.separator + "formsV7.db", FormsDatabaseHelper.DATABASE_PATH);
```
lognaturel: How did you generate these db files? That is, were they copied straight from Collect or did you manipulate them in some way? I think it'd be helpful to use some kind of convention to indicate that. I used a `_real` prefix to indicate it was straight from Collect. In this case, if these two DB files were taken straight from Collect, I'd call them `forms_v7_real.db` and `instances_v5_real.db`.  I don't mind changing the convention as long as things are consistent.

Edit: it looks like 9ac52a5ee6d4339a81e1da841c2c68d8007f3cfd changes the convention to just use `tableName_vN.db` when the database comes straight from Collect. Is that accurate (that is, can you confirm all the ones you added are straight from Collect)? If so, it sounds good to me.

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static junit.framework.TestCase.assertFalse;
+import static junit.framework.TestCase.assertTrue;
+import static org.odk.collect.android.database.helpers.FormsDatabaseHelper.FORMS_TABLE_NAME;
+import static org.odk.collect.android.database.helpers.InstancesDatabaseHelper.INSTANCES_TABLE_NAME;
+import static org.odk.collect.android.test.FileUtils.copyFileFromAssets;
+
+public class SQLiteUtilsTest {
+
+    @Test
+    public void doesColumnExistTest() throws IOException {
```
lognaturel: I often struggle with reviewing tests and making sure they're meaningful and helpfully-named so I'm interested in bringing @seadowg in here as well.

I got confused initially because I thought this was checking something about the migrations whereas the intent is to verify the `doesColumnExist` method. Since this method doesn't have anything to do with the specific database tables in Collect or their versions, I would suggest doing this as a Robolectric test. I would further claim that all you need is to verify one column name that exists and one that doesn't since there's no logic in the method.

collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
```diff
@@ -344,4 +351,14 @@ public static void databaseMigrationStarted() {
     public static boolean isDatabaseBeingMigrated() {
         return isDatabaseBeingMigrated;
     }
+
+    static List<String> getFormsColumnNames(SQLiteDatabase db) {
```
lognaturel: Maybe it's time to pull this out of here and the instance helper?

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/FormsDatabaseHelperTest.java
```diff
@@ -61,5 +64,17 @@ public void testMigration() throws IOException {
         List<String> newColumnNames = FormsDatabaseHelper.getFormsColumnNames(db);
 
         assertThat(newColumnNames, contains(FormsDatabaseHelper.CURRENT_VERSION_COLUMN_NAMES));
+
+        if (description.startsWith("Upgrading")) {
```
lognaturel: This is a good check to add. How about adding a boolean parameter instead? Checking the description seems error-prone.

collect_app/src/androidTest/java/org/odk/collect/android/test/FileUtils.java
```diff
@@ -0,0 +1,28 @@
...
+public class FileUtils {
+
+    private FileUtils() {
+    }
+
+    public static void copyFileFromAssets(String fileSourcePath, String fileDestPath) throws IOException {
+        AssetManager assetManager = InstrumentationRegistry.getInstrumentation().getContext().getAssets();
+        try (InputStream input = assetManager.open(fileSourcePath);
+             OutputStream output = new FileOutputStream(fileDestPath)) {
+            byte[] buffer = new byte[1024];
```
lognaturel: Maybe it'd be better to use `IOUtils.copy` here as in https://github.com/opendatakit/collect/pull/3323/files#diff-95bba6de31f4296ffadad5868939acb3L116? I forgot about it when I introduced this method.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -523,6 +527,18 @@ public int update(Uri uri, ContentValues values, String where,
         return newWhereArgs;
     }
 
+    private boolean isDatabaseHelperOutOfDate() {
```
lognaturel: Same comment as with `InstanceProvider`: move to the db helper or make static.

lognaturel (changes requested):
Thanks for the really good explanation of what's going on and how you've addressed it.

Two things that I think really need to be addressed -- `DatabaseContext.getDatabasePath` ignoring its parameter and the `isDatabaseBeingMigrated` flag staying true after a failure and rollback. The rest are questions or suggestions.

collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java
seadowg: I'm generally fine with static helpers inside a class - especially for the reasoning that it makes them easier to move. I've even found it very useful in heavier objects to even use static methods that just take in instance fields as parameters as sometimes this can end up more readable. They can be problematic when you realize that a static method is a dependency from some object you'd like to test but usually it's pretty easy to then create an interface where that method should live - this was done recently with the `PermissionUtils`. The fact that we can't have methods/functions outside an actual class makes this more controversial in Java that other languages in my opinion.

In this case it makes more sense to me to move the helper to `InstanceDatabaseHelper` as a public method. For me then you're just asking the object "are you out of date?" and then the `InstancesProvider` can decide what to do with it. What would be really nice would be to follow the ["tell don't ask"](https://www.martinfowler.com/bliki/TellDontAsk.html) principle here and call something on `InstancesDatabaseHelper` that makes it upgrade itself if needed. I don't think that's possible though... 🤷‍♂️

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
seadowg: Yeah Robolectric runs against real SQLite so there should be no harm in moving that there. I like the idea of the test creating it's own database (using [`SQLiteDatabase`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase)) with a couple of columns and then checking `doesColumnExist` for an existing and a non existing column. That gives us a test that checks that `doesColumnExist` does its job without having to worry about how databases look in Collect.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/test/FileUtils.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseContext.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
grzesiek2010: Yes, it might end up in such a state. Fixed.

collect_app/src/main/java/org/odk/collect/android/utilities/SQLiteUtils.java
grzesiek2010: I was thinking about this but CustomSQLiteQueryBuilder doesn't seem like a good place either DAOs so I decided to leave it as-is.

collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
grzesiek2010: All of them are from Collect but some of them I needed to edit adding a column, removing etc.
If it's not edited but just copied from collect the name is like 
`forms_v4.db`
`instances_v5.db`
but if it's edited the name contains more info:
`instances_v7000_removed_jrVersion.db`
and has a version that indicates it's not an original file.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/FormsDatabaseHelperTest.java
grzesiek2010: How do you imagine such a check? I would need to open the database like I do here for example: https://github.com/opendatakit/collect/pull/3323/files#diff-0c930d81466089bfec86df65dbef3a4eR365 and compare versions but is it worth it?

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/FormsDatabaseHelperTest.java
lognaturel: I just meant that in addition to the `description` and `dbFilename` parameters, it could have an `upgrading` parameter that gets set when the test is created. That way it's more self-documenting than the fact that the description has to have a particular name.

collect_app/src/test/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+                + FORM_FILE_PATH + " text not null, "
+                + LANGUAGE + " text, "
+                + SUBMISSION_URI + " text, "
+                + BASE64_RSA_PUBLIC_KEY + " text, "
+                + JRCACHE_FILE_PATH + " text not null, "
+                + AUTO_SEND + " text, "
+                + AUTO_DELETE + " text, "
+                + LAST_DETECTED_FORM_VERSION_HASH + " text);");
+
+        assertTrue(SQLiteUtils.doesColumnExist(db, FORMS_TABLE_NAME, _ID));
```
lognaturel: I really think you only need one of these assertions and that the database doesn't need to be related to Collect at all! But this is already much easier to understand and much faster to run so I'm fine with leaving it and simplifying in the future if needed.

collect_app/src/androidTest/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
lognaturel: Works for me.

collect_app/src/test/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
seadowg: Yeah I think to make it really clear that the behaviour of `doesColumnExist` is independent from the structure of Collect's database it could just create a database with `COLUMN_A` and `COLUMN_B`, test they exist and then test that `COLUMN_C` doesn't. This means you don't have to import the contants from `FormsProviderAPI`. It would also cut down on the size of the test potentially making it easier for someone new to the codebase to understand quickly.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/FormsDatabaseHelperTest.java
grzesiek2010: ok done.

collect_app/src/test/java/org/odk/collect/android/utilities/SQLiteUtilsTest.java
grzesiek2010: ok done.

seadowg: Love it!

seadowg (approved):

lognaturel (approved):
Love the new test!

grzesiek2010:
@mmarciniak90 I did a rebase so testing this pr you can also test #3329 at the same time.

mmarciniak90:
**Tested with success**

**Verified on Android**: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 9.0
Collect does not crash. The crash was only visible for KoBoCollect app.

Verified cases:
- used Kobo v1.14
- used Kobo v1.4.3
- opening apps from the main menu
- switching between both apps

PR #3229, opened 2019-07-23
Audio playback refactor

seadowg:
Closes #2521

There's already been some good work and discussion around how to fix some of the bugs and quirks around audio playback in the app (#2522, #2863 and #2817 for example). For people not familiar with the bugs: there are multiple problems with play/stop buttons interacting with one another. After having a look at the problem, it made sense to give it another stab now that we have some new tooling maturing in the Android Jetpack world.

After digging into the current architecture it made sense to me that a good way forward would be to separate out audio playback functionality entirely from the question world so that it could be tested and implemented in isolation. To help myself out I ended up drawing a picture of this in [Miro](https://miro.com):

Screenshot 2019-07-23 at 10 20 28

This draft starts with an initial pass at that that uses an `AudioPlayerViewModel` to coordinate the `MediaPlayer` and expose state and actions to an `AudioButton` view that is only responsible for rendering whether it is in the "playing" state or not and for emitting (via a listener) "playClicked" and "stopClicked" events.

I worked through this with @lognaturel and one thing we noticed was how many different behaviors of the app that audio playback is part of. Although this isn't finished yet it made sense to get a draft PR up early to get feedback and also have people point out features we might be missing. I'm going to enumerate them below (and tick off what is working in this reimplementation) - shout out if anything is missing:

- [x] Question audio can be played/stopped
- [x] Select option audio can be played/stopped
- [x] Playing one audio button on a screen stops any other playing one
- [x] Question option text is highlight when audio is being played
- [x] Select option text is highlighted when audio is being played
- [x] `MediaPlayer` objects are released when `Activity` is destroyed
- [x] `MediaPlayer` objects are released when `Activity` is paused
- [x] Moving between form screens stops playback
- [x] Audio can be autoplayed for questions that are being **forward** swiped into
- [x] Playing an audio button should stop an `AudioWidget` (and vice versa)
- [x] Audio in an audio widget can be played/paused
- [x] Audio in an audio widget can be seeked with buttons
- [x] Audio in an audio widget can be seeked with seek bar
- [x] Pausing/resuming/rotating Activity doesn't cause any problems (currently playing audio etc)

**Just to note**: this PR does not fix the problems outlined in #2861 - autoplay for select options is not functioning correctly. We'll follow up with a fix for that once this is in.

#### What has been done to verify that this works as intended?

Used several different forms involving `AudioWidget`s, audio in question media and in select options etc etc.

[...]

PR #3326, opened 2019-08-30
Reimplement audio autoplay

### LINKED ISSUE #2861: Autoplay option doesn't work
Collect v1.19.0 and older versions

#### Problem description
During testing #2817 we discovered that `autoplay` option doesn't work.
It crashes on 1.19, 1.18, 1.17
There is no crash but it also doesn't work on 1.16, 1.15 ... 1.8 (we didn't check every version between but we can assume it doesn't work in all cases)

#### Other information 
@yanokwa I said that probably we can remove that option since it seems useless but you didn't agree.
I think I'm right... @mmarciniak90 was testing that option today and it doesn't work on all devices in the way I described above (sometimes crashes).

It crashes on three last versions but we have no crash reports. Today @mmarciniak90 caused firs appearances:
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5c5d46dbf8b88c29631622fc?time=last-seven-days&sessionId=5C5D2E4102EA00013CCDBAA626940E27_DNE_8_v2

https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5c5d47acf8b88c2963163597?time=last-seven-days

additionally, no one complained about it (it doesn't work on 1.16, 1.15 ... 1.8 and probably older versions) so I'm pretty sure it's completely useless. 

The link you attached https://forum.opendatakit.org/search?q=autoplay%20order%3Alatest also shows only one question about that option during last three years.

Even if we want to fix that option for some reason keep it then it would require some discussion.
1. Should that play only media files attached to choices or the one attached to the question too?
2. What if we have multiple questions on one page and we use that option.
etc.

I'm not a fan of that option.

seadowg:
Closes #2861

This reimplements/fixes autoplay audio for labels and select choices. As noted in #2861 the features are:

* Audio on a question will play automatically when a user swipes to a page if the question has audio in the autoplay column
* If the question is a select with audio files attached to the choices they will play in order (after the question) when the question has audio in the autoplay column
* Returning to question will not autoplay audio again (autoplay only happens when moving "forward")

One other notable change is that I've renamed the `MediaLayout` to `QuestionLabel` as this seemed to be confusing myself and others. I've also started moving more classes in the `formentry` package with media specific stuff going in `formentry.media`.

Tests and played around with a few autoplay forms (including one's with select and other media types).

I think the discussion around the approach really happened in #3229 so I'll defer to that.

The only change should be that autoplay now works for select choices. It would be good to test this against different select types and forms with and without selects.


collect_app/src/test/java/org/odk/collect/android/views/helpers/FormAutoplayerTest.java
```diff
@@ -31,7 +31,7 @@
 
     @Test
     public void whenPromptHasAutoplayAudio_returnsTrue() throws Exception {
-        setupMockReference("file://audio.mp3", referenceManager);
+        setupMockReference("file://audio.mp3", "file://reference.mp3", referenceManager);
```
lognaturel: I think it would make these tests easier to understand if the resolved reference looked like the kind of path that the reference manager would actually return (e.g. `/sdcard/odk/instances/my-instance/reference.mp3`).

collect_app/src/test/java/org/odk/collect/android/widgets/SelectWidgetTest.java
```diff
@@ -63,17 +63,21 @@ public void whenQuestionHasAudio_audioButtonUsesIndexAsClipID() throws Exception
...
+        for (int i = 0; i < selectChoices.size(); i++) {
+            selectChoices.get(i).setIndex(i);
+        }
+
         when(formEntryPrompt.getSpecialFormSelectChoiceText(selectChoices.get(0), FormEntryCaption.TEXT_FORM_AUDIO)).thenReturn("file://blah1.mp3");
         when(formEntryPrompt.getSpecialFormSelectChoiceText(selectChoices.get(1), FormEntryCaption.TEXT_FORM_AUDIO)).thenReturn("file://blah2.mp3");
 
-        setupMockReference("file://blah1.mp3", referenceManager);
-        setupMockReference("file://blah2.mp3", referenceManager);
+        setupMockReference("file://blah1.mp3", "ref1", referenceManager);
```
lognaturel: Or even this makes more sense to me than having another `file://` uri.

collect_app/src/main/java/org/odk/collect/android/audio/Clip.java
```diff
@@ -23,4 +24,32 @@ public String getClipID() {
...
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Clip clip = (Clip) o;
+        return getClipID().equals(clip.getClipID()) &&
+                uri.equals(clip.uri);
```
lognaturel: Directly use fields for both for consistency?

collect_app/src/test/java/org/odk/collect/android/support/MockFormEntryPromptBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+import org.javarosa.core.model.IFormElement;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.form.api.FormEntryPrompt;
+
+import java.util.List;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MockFormEntryPromptBuilder {
```
lognaturel, line 15: This is nice.

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -85,7 +84,7 @@
 
     private final int questionFontSize;
     private final FormEntryPrompt formEntryPrompt;
-    private final MediaLayout questionMediaLayout;
+    private final QuestionLabel questionQuestionLabel;
```
lognaturel: Looks like it may be an auto refactor issue. Should be just `questionLabel`, right?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -149,7 +131,7 @@ public boolean onError(MediaPlayer mp, int what, int extra) {
         setGravity(Gravity.TOP);
         setPadding(0, 7, 0, 0);
 
-        questionMediaLayout = createQuestionMediaLayout(prompt);
+        questionQuestionLabel = createQuestionMediaLayout(prompt);
```
lognaturel: `createQuestionMediaLayout` should become `createQuestionLabel`?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -702,8 +665,8 @@ public View getHelpTextLayout() {
         return helpTextLayout;
     }
 
-    public MediaLayout getQuestionMediaLayout() {
-        return questionMediaLayout;
+    public QuestionLabel getQuestionQuestionLabel() {
```
lognaturel: Change to `getQuestionLabel`

collect_app/src/main/java/org/odk/collect/android/audio/Clip.java
```diff
@@ -0,0 +1,55 @@
+package org.odk.collect.android.audio;
+
+import androidx.annotation.NonNull;
+import androidx.core.util.ObjectsCompat;
```
lognaturel, line 4: Ahhhh, I didn't know about this. Should probably replace `ObjectUtils.equals` at some point.

lognaturel (changes requested):
No concerns or even questions about the implementation itself. I marked a handful of small things inline.

I'm not excited about `QuestionLabel` because the name suggests it's only used for prompt labels and not also for select labels. I guess you might argue selects are in the context of a question? `ItemLabel` feels too generic. Maybe `FormEntryLabel` so it's clear it's a label that's shown during form entry? `AudioVideoTextLabel`? 

I guess the big thing is that calling it a label gives more context. I don't feel super strongly and certainly this is not important enough to spend a ton of time finding the perfect name but maybe you like some of the other options I've thrown out or you get another thought from that. @grzesiek2010 might also have good ideas.

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
seadowg: Ah damn. Yeah I'll fix that.

collect_app/src/main/java/org/odk/collect/android/audio/Clip.java
seadowg: Yeah annoyingly Android Studio generates a `hashCode` method that doesn't compile. This took a bit of Googlin'.

collect_app/src/main/java/org/odk/collect/android/audio/Clip.java
seadowg: Yeah regenrated. For some reason Android Studio wasn't recognizing thr URI getter. I've used getters for both now.

collect_app/src/test/java/org/odk/collect/android/widgets/SelectWidgetTest.java
seadowg: Yeah I prefer just having a String like this. I think using a "real" looking URI shows too much knowledge in the test of what the `ReferenceManager` does - we don't strictly care about how it works in these tests.

seadowg: Actually I think what would be nicer after looking at it again would be to return a `String` from `setupMockReference` so that the test could just pass that around instead of us having to trace the value through ourselves when we're reading/writing the test

seadowg:
> `AudioVideoTextLabel`

I like this one the most actually. I'll change it.

lognaturel (approved):
Looks great!

grzesiek2010 (approved):

kkrawczyk123:
@seadowg do you have a sample form for testing?

seadowg:
@kkrawczyk123 there was one posted in the issue I found useful: https://github.com/opendatakit/collect/issues/2861#issuecomment-461999890. A user also posted a demo form here: https://forum.opendatakit.org/t/autoplay-eventually-crashes-survey/20621/6?u=seadowg.

Would be good if possible to have a play around with making a few toy forms and modifyin those examples to see if there are combinations/situations I haven't thought about.

kkrawczyk123:
@seadowg I am trying to. I've already used that one from #2861. I was trying to modify that one 
[mediatest2_test .xml.txt](https://github.com/opendatakit/collect/files/3675446/mediatest2_test.xml.txt)
but it doesn't autoplaying do you have any advices why?

seadowg:
@kkrawczyk123 It looks like that has a field list with two select ones, both with autoplay. Are neither autoplaying?

kkrawczyk123:
Yes, they are not autoplaying and I expected them to be. I think that it is connected t that they both are on one page. With one select question on the page and two answers with separate audios, it works just fine.

seadowg:
@kkrawczyk123 hmmmm I wonder if it's just not working for field lists. Would you be able to send me the form and media in a zip so I can have a play with it?

kkrawczyk123:
@seadowg I have shared via Google Drive as it's quite big.

seadowg:
@kkrawczyk123 ah looking at the code it looks like we only autoplay if there is only one question on the page. I also just remembered I asked a user a question about this case on [the forum](https://forum.opendatakit.org/t/autoplay-eventually-crashes-survey/20621/7?u=seadowg) but we haven't heard back yet.

This is existing behaviour so I think we should stick with it for this PR.

kkrawczyk123:
@seadowg, thanks for the clarification!
I have other questions. I have used the form: 
[test.zip](https://github.com/opendatakit/collect/files/3675786/test.zip)
- When I remove just a.mp3 file - I can see a toast informing me that media is missing on all pages of the form, even on the second page where media a is not used. 
- When a.mp3 file is removed on the last screen where both media are used - the second file is not autoplayed.

- When I remove just b.wav - I can see a toast on the second and the third form pages - I do not see it on the first one (this is exactly what I was expecting)
- When b.wav file is removed on the last screen where both media are used - the first file is autoplayed.

Is it expected behavior?

seadowg:
@kkrawczyk123 Hmmm interesting. I think the error reporting can be improved so it just shows on the relevant page. I can work on that.

In terms of removed files and autoplay I'm interested if you think we should play all the clips we can or if we should not autoplay anything at all if any of the files are missing?

kkrawczyk123:
@seadowg I feel like not deleted audio file should still autoplay and the toast informing of another missing file should display. Are you planning to work on those two things in that pr? If yes please tag me when it's ready for re-testing :)

seadowg:
@kkrawczyk123 that makes sense I agree. I'll have a look at fixing these and get it back to you!

seadowg:
@kkrawczyk123 ok both issues should be fixed now!

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1 and 9.0.

- confirmed that noticed issues have been fixed
- autoplay with text question
- autoplay with select question
- removing 1 media, confirmed that toast informing that the file is missing displayed only on pages where media is used
- removing both media files, confirmed that toast informing that the file is missing displayed only on pages where media is used - toasts are visible one by one on Androids from 4.2 to 7.0. On devices with Android 8.1 and 9.0 toasts are changing so fast that it's barely noticeable. Probably the cause of the behavior is the magic of the Android.
- dark mode and light mode
- minimizing app stops autoplay
- screen rotation autoplay is stopped on Androids 4.2-7.0, autoplay is continued after rotation on devices with Android 8.1 and 9.0
- I have looked for regression on a regular audio widget in All widgets form - nothing suspicious happened 
- I have also run through Birds form and still did not recognized any weird behavior

PR #3327, opened 2019-08-30
Fixed removing repeat groups

### LINKED ISSUE #3033: When repeat with field-list appearance is deleted, the wrong repeat instance is removed
Collect v1.21

#### Problem description
https://forum.opendatakit.org/t/bugs-when-removing-a-repeated-group-that-is-not-the-last-group/4930

> when a "repeat group" is defined with a "field-list" appearance to group several questions on the same screen, trying to remove a group that is not the last one does not remove that instance but the NEXT one.

https://opendatakit-a3b1.kxcdn.com/uploads/default/original/2X/7/7b7f3bff5626bfdd54ae006b5c19369a36851be0.xlsx

#### Other information 
The correct group is deleted when the delete is initiated from the hierarchy view. It would be really good to unify that deletion code as part of this fix so that the behavior is consistent.

Originally reported at #187

grzesiek2010:
Closes #3033

It was a regression, we fixed one issue 2 years ago https://github.com/opendatakit/collect/pull/1034 which introduced this problem so I just combined two solutions to handle both cases.

The change is not very risky, I added automated tests so testing the form attached to this pr would be enough but @mmarciniak90 you can try some more complicated forms of course as well to be sure everything is fine.

The form attached to this pr https://github.com/opendatakit/collect/compare/master...grzesiek2010:COLLECT-3033?expand=1#diff-fe55d3110049e63221b4a8a2a49067cb

No.


collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,282 @@
...
+        FormEntry.clickOnString(R.string.add_another);
+        FormEntry.putText("31");
+        FormEntry.swipeToNextQuestion();
+        FormEntry.putText("32");
+        FormEntry.swipeToNextQuestion();
+        FormEntry.clickOnString(R.string.add_another);
+        FormEntry.putText("41");
+        FormEntry.swipeToNextQuestion();
+        FormEntry.putText("42");
+        FormEntry.swipeToPrevoiusQuestion();
```
lognaturel: Would be good to get this typo fixed! Could be as a commit here or in another PR. (`Prevoius` -> `Previous`)

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,282 @@
...
+            .outerRule(GrantPermissionRule.grant(
+                    Manifest.permission.READ_EXTERNAL_STORAGE,
+                    Manifest.permission.WRITE_EXTERNAL_STORAGE
+                    )
+            )
+            .around(new ResetStateRule())
+            .around(new CopyFormRule(TEST_FORM, null));
+
+    @Test
+    public void testDeletingRepeatGroupsInForm() {
```
lognaturel: I see two high-level issues with these tests:
- they're going to take a long time to run because there's a lot of swiping back and forth
- if one fails, it's going to be hard to know what that means since the cases aren't identified

What are the cases that need to be verified? Seems to me that they are deleting the first repeat, deleting the last repeat and deleting some repeat in the middle. What is the fourth case you have? Is it necessary? 

Why are all of the repeats after the deleted ones verified? Would perhaps just checking the one before and the one after be enough?

Why are there two questions in each repeat? I think it's to visually differentiate the field list and non-field list case but does that actually matter for the test? That is, would the field list test fail before the fix even if there was just a single question in each repeat? I think so and in that case, it'd be simpler to have a single question in each.

Since populating the repeats is not part of the test, could you set the form definition to include them? I mean having something like the following in the form definition `instance` block:
```
<repeatGroup>
  <text>foo</text>
</repeatGroup>
<repeatGroup>
  <text>bar</text>
</repeatGroup>
...
<repeatGroupFieldList>
  <number>12</number>
</repeatGroupFieldList>
<repeatGroupFieldList>
  <number>24</number>
</repeatGroupFieldList>
```

That way you can jump directly to the repeat to delete and avoid a lot of the swiping.

Then it also makes it easier to have independent cases like `requestingDeletionOfFirstRepeat_deletesFirstRepeat` and `requestingDeletionOfMiddleRepeat_deletesMiddleRepeat`.

lognaturel (changes requested):
Fix looks good but the tests will take a long time to run. I've asked some questions and made suggestions inline.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: ok I simplified those tests and divided them into more but smaller.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        onView(withId(R.id.list)).check(matches(RecyclerViewMatcher.withListSize(4)));
+        FormEntry.clickOnText("repeatGroupFieldList > 2");
+        FormEntry.deleteGroup();
+        onView(withId(R.id.list)).check(matches(RecyclerViewMatcher.withListSize(3)));
+        FormEntry.clickOnText("repeatGroupFieldList > 2");
+        FormEntry.checkIsTextDisplayed("3");
+    }
+
+    @Test
+    public void requestingDeletionOfLastRepeatWithFieldListInHierarchy_deletesFirstRepeat() {
```
lognaturel: Should be `deletesLastRepeat`

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        FormEntry.clickGoUpIcon();
+        FormEntry.clickOnText("repeatGroupFieldList");
+        FormEntry.clickOnText("repeatGroupFieldList > 2");
+        FormEntry.clickOnText("number1");
+        FormEntry.deleteGroup("number1");
+        FormEntry.checkIsTextDisplayed("3");
+    }
+
+    @Test
+    public void requestingDeletionOfLastRepeatWithFieldList_deletesFirstRepeat() {
```
lognaturel: Should be `deletesLastRepeat`

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        FormEntry.clickOnText("repeatGroup > 2");
+        FormEntry.checkIsTextDisplayed("2");
+        FormEntry.deleteGroup();
+        onView(withId(R.id.list)).check(matches(RecyclerViewMatcher.withListSize(3)));
+        FormEntry.clickOnText("repeatGroup > 2");
+        FormEntry.checkIsTextDisplayed("3");
+    }
+
+    @Test
+    public void requestingDeletionOfLastRepeatInHierarchy_deletesFirstRepeat() {
```
lognaturel: `deletesLastRepeat`

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Test
+    public void requestingDeletionOfMiddleRepeat_deletesMiddleRepeat() {
+        FormEntry.swipeToNextQuestion();
+        FormEntry.deleteGroup("text1");
+        FormEntry.checkIsTextDisplayed("3");
+    }
+
+    @Test
+    public void requestingDeletionOfLastRepeat_deletesFirstRepeat() {
```
lognaturel: `deletesLastRepeat`

lognaturel (changes requested):
The test split looks good to me. I think on aggregate it probably takes longer to run but I think what's more important is that they're more likely to help us identify and fix an issue this way. 

Some of the cases have incorrect names.

I also can't run the full test suite. I get `java.lang.IllegalStateException: #init was called twice in a row. Make sure to call #release after every #init` and haven't looked any deeper yet.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DeletingRepeatGroupsTest.java
grzesiek2010: Fixed.

grzesiek2010:
> I also can't run the full test suite. I get java.lang.IllegalStateException: #init was called twice in a row. Make sure to call #release after every #init

I haven't noticed anything like that.

lognaturel:
I still see the issue running the tests on Pixel 2 / API27 but if you and @seadowg can't reproduce it's probably something about my state.

seadowg:
@lognaturel hmmm. That's weird. Have could you do a wipe on the emulator and then try?

bognasoldev89:
*Tested with success*

Verified on Android: 
4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 9.0

Verified cases:
- Fixed removing repeat groups
- Rotating phone
- Swiping back and forth
- Blocking phone
- Minimizing phone
- Saving and opening saved form
- Finalizing and opening finalized form
- Sending form to Aggregate

lognaturel (approved):

Issue #2695, opened 2018-10-26
Upgrade osmdroid library version from 5.6.4 to 6+


zestyping:
ODK Collect currently uses the org.osmdroid library version 5.6.4, from February 2017.  The latest version of this library is 6.0.1, released February 2018.

There have been several improvements and bug fixes to osmdroid since 5.6.4 that might be nice for us, such as support for higher zoom levels, fractional zoom levels, and geopackage support.  See https://github.com/osmdroid/osmdroid/wiki/Changelog for detailed release notes.

However, upgrading is more involved than just changing the version number in the `build.gradle` file.  The API has some non-compatible changes; for example, some of the classes for offline tile support are gone.

@zestyping: I took a brief look at this and didn't find obvious documentation on how to migrate from 5.6.4 to 6+.  I think someone will need to take a little time to survey what's changed and make a plan.  Anyone is welcome to look at this, but **please come up with a plan and let's discuss it here** before proceeding to make a pull request.

grzesiek2010:
I tried to update osmdroid working on #3329 but noticed crashes like https://github.com/osmdroid/osmdroid/issues/1299 in the newest version 6.1.0 so let's wait for next release.

grzesiek2010:
Fixed in https://github.com/opendatakit/collect/pull/3728

PR #3329, opened 2019-09-03
Updated dependencies

grzesiek2010:
I tested some functionalities like
- submitting forms to Google Sheets what requires play services
- `auto-send` option

@lognaturel 
I know you have been always afraid of updating google play services because you thought it would require downloading new versions on each device but I tested the functionality with version 15.0 and everything works fine I didn't have to download the newest version. Have you experienced something else?

If we decided to keep updated play services it would be good to do the same I did I mean to test this pr with an older version like v15.0. It would be also good to test `auto-send` option because I had to change some classes after upgrading work manager. Rest if changes we can just test on a daily basis.

No.

No.


seadowg:
On the Play Services versioning: my impression was that they had severed the client library versioning from the actual on device Play Services version. This [documentation](https://developers.google.com/android/guides/versioning) seems to suggest that but doesn't really say it explicitly so I could totally be wrong.

lognaturel:
Thanks, @grzesiek2010, for doing this and for highlighting the Play Services question. Let me do a little reading and digging on that. It would be great to update it if we can without negative user impact.

On another note, could you also please describe your high-level process for figuring out what needs to be updated? For example, #3332 upgraded Robolectric and I was surprised not to see that change here. We also had talked about removing unused dependencies like `rxLifecycle` in [Collect 1.24 planning](https://docs.google.com/document/d/1My9NeINHyHvslpjalub_RR8Aw7jQKBSJzpKdJiaGf68/edit). Dependencies in https://github.com/opendatakit/collect/blob/9b1185f589375fc4e6d6d54581a3714ec43fa8a2/build.gradle should also be updated.

collect_app/src/main/java/org/odk/collect/android/utilities/PlayServicesUtil.java
```diff
@@ -23,10 +23,10 @@ private PlayServicesUtil() {
     /**
      * Check that Google Play services APK is installed and up to date.
      *
-     * @return true if Google Play Services is available and up to
+     * @return true if Google Play Services are available and up to
      * date on this device; false otherwise.
      */
-    public static boolean isGooglePlayServicesAvailable(Context context) {
+    public static boolean areGooglePlayServicesAvailable(Context context) {
```
lognaturel: I can see your rationale for renaming this. Since it's a wrapper on Google's `isGooglePlayServicesAvailable`, though, I would tend to leave it. Don't feel strongly.

lognaturel (changes requested):
Let's get as many updates done as possible done in this PR. I've highlighted several that are missing. When @grzesiek2010 reviews his process he may identify others.

lognaturel:
It used to be that the library versions declared by developers matched the Play Services app version users got but @seadowg is absolutely correct that this is no longer the case. Is it possible that the Play Services app version required is dynamically figured out based on APIs used? If so, we just need to be careful if we use APIs in newer libraries.

After compiling `master`, I can see that `google_play_services_version` in `AndroidManifest.xml` is `12451000`. When I pull this branch and do the same, I get the same Play Services version. This confirms that it's safe. It's not clear to me whether this value is dynamically generated as I described above or pulled from... somewhere? As far as I can tell, the latest Play Services version is 19.0.56 and 12.45.1 would have come out around 2017.

grzesiek2010:
> On another note, could you also please describe your high-level process for figuring out what needs to be updated?

Android Studio highlights dependencies that we can update but not all of them, then I used `./gradlew lint` which shows more of them but looks like also not all hmm... strange

grzesiek2010:
I have just discovered a nice plugin https://github.com/ben-manes/gradle-versions-plugin which shows much much more so I'm going to use it.

collect_app/src/main/java/org/odk/collect/android/utilities/PlayServicesUtil.java
grzesiek2010: Fixed

lognaturel:
Let me know when this is ready to look at again, @grzesiek2010!

grzesiek2010:
I think it is.
I just wasn't able to update: 
[com.journeyapps:zxing-android-embedded:3.5.0](https://github.com/opendatakit/collect/pull/3329/files#diff-fa0f36229667501896e9f29c3a2a4b35R266)
and
[tools.fastlane:screengrab:1.1.0](https://github.com/opendatakit/collect/pull/3329/files#diff-fa0f36229667501896e9f29c3a2a4b35R315)
but I added comments that it would require bumping minSdk.

I didn't also update osmdroid, I tried and was able to fix the code since it requires more changes but the newest version is faulty so decided to keep the current one, we have an issue for that:
https://github.com/opendatakit/collect/issues/2695

collect_app/build.gradle
```diff
@@ -197,76 +197,74 @@ dependencies {
...
+    implementation "com.google.android.gms:play-services-maps:17.0.0"
+    implementation "com.google.android.gms:play-services-location:17.0.0"
 
-    implementation "com.google.firebase:firebase-core:16.0.1"
-    implementation "com.crashlytics.sdk.android:crashlytics:2.9.2"
+    implementation "com.google.firebase:firebase-core:17.2.0"
+    implementation "com.crashlytics.sdk.android:crashlytics:2.10.1"
 
-    implementation 'com.mapbox.mapboxsdk:mapbox-android-sdk:7.3.0'
+    implementation 'com.mapbox.mapboxsdk:mapbox-android-sdk:8.4.0-alpha.1'
```
lognaturel: Why not go with 8.3, here?

collect_app/build.gradle
```diff
@@ -197,76 +197,74 @@ dependencies {
...
     implementation 'com.mapbox.mapboxsdk:mapbox-android-plugin-annotation-v7:0.6.0'
 
-    implementation("com.google.guava:guava:27.1-android")
+    implementation("com.google.guava:guava:28.1-jre")
```
lognaturel: Why is this `-jre` rather than `-android`?

lognaturel (commented):
> the newest version is faulty

Could you please write a note about your experience in #2695?

I did a quick sanity check on versions. In general it looks like the ones that are still alpha/beta have no choice or are test-only dependencies so that looks good to me. I just have two questions inline.

I think that another good QA sanity check would be to try form auto updates since android-job changed a lot.

Finally, I ran the Espresso tests and got failures with `testActivityOpen` and `onlyNextButton_ShouldShowOnFirstScreen`. Not sure whether it's related or not.

collect_app/build.gradle
grzesiek2010: Yeah in this case maybe it would be better.

collect_app/build.gradle
grzesiek2010: Right, my mistake.

grzesiek2010:
> Could you please write a note about your experience in #2695?

done.

> Finally, I ran the Espresso tests and got failures with testActivityOpen and onlyNextButton_ShouldShowOnFirstScreen. Not sure whether it's related or not.

on testLab it looks fine.

lognaturel (approved):

lognaturel:
For @opendatakit/testers, the following sanity checks would be great:
- submission to Google Sheets with Play Services 12.45.1 if you downgrade it (I think you have before)
- form auto send
- form auto update

I'm going to go ahead and merge so the code paths can start being exercised as we do dev work. We can always roll back certain versions if needed.

mmarciniak90:
> submission to Google Sheets with Play Services 12.45.1 if you downgrade it (I think you have before)

@lognaturel Is the exact 12.45.1 version so important? To be honest, I'm not able to find this version. 
Can I use some other version from [this](https://www.apkmirror.com/uploads/page/5/?q=google-play-services) list?


I was not able to downgrade on devices:
* Android 9.0 - Google Play Services v19.2.75
* Android 8.1 - Google Play Services v19.2.75

Downgraded to factory version on devices:
* Android 7.0 - Google Play Services v11.9.51
* Android 6.0 - Google Play Services v9.0.83
* Android 5.1 - Google Play Services v8.3.01

lognaturel:
> Is the exact 12.45.1 version so important?

I don't think so. I believe that's the minimum version required to run all the Collect features currently. It looks like I put the second period in the wrong place and it's probably actually 12.4.51. Something like 12.5.27 would be perfect -- it's close to the minimum version and we should be able to confirm that all functionality that uses Play Services works. As far as I know, that is submission to GSheets, location with GMaps basemap, auto-send and auto-update (anything else, @grzesiek2010?). I think looking at a single device with that version would be sufficient.

mmarciniak90:
**Tested with success** 

**Verified cases:**
- Google Play Services v12.5.27 installed on devices with **Android 7.0, 6.0, 5.1, 4.4, 4.2**
- auto-send to Google Sheet and Aggregate
- form auto-update from Aggregate
- autodelete
- submission with media
- light/dark mode
- English/Spanish language

Cases were verified also on Android 8.1 and 9.0 but with the newest version of Google Play Services

PR #3339, opened 2019-09-11
Add tests for Open Rosa authentication caching

seadowg:
This add tests around authentication caching (in the `OpenRosaServerClientProviderTest`) to ensure our HTTP client doesn't perform authentication handshakes for every single request. There are a few other changes to highlight:

* Moved Open Rosa and OKHttp implementation related classes into the `openrosa` package
* Removed any Android code out of our OkHttp implementations so that we don't have to use Robolectric to test it

Another small change made was to remove a string form the resources that was unused after I inlined it for a Timber log call. Weirdly when removing the string (using the resource editor) Android Studio formatted the line below in every `strings.xml` leaving us with one line out of alignment with everything else. It looks like if you autoformat a strings file with standard settings in Android Studio it moves everything in a couple of spaces. I think for this PR we except it and then follow up with a PR that corrects the formatting on the files.

Ran tests locally and played around with the app with the network profiler just to verify none of the refactoring had broken/changed anything.

 Shouldn't be any changes in behaviour. It would be good to verify the app against different servers with the network profiler however. We probably need to write a some kind of test harness for our open rosa communication that stops us having to be so cautious with every change. I'm sure we'll get there as we keep improving the structure of this part of the code.


collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientProviderTest.java
```diff
@@ -1,5 +1,6 @@
 package org.odk.collect.android.http;
 
+import org.jetbrains.annotations.NotNull;
```
lognaturel: Should we consistently use the androix annotations?

collect_app/src/test/java/org/odk/collect/android/http/support/MockWebServerHelper.java
```diff
@@ -0,0 +1,18 @@
+package org.odk.collect.android.http.support;
+
+import org.jetbrains.annotations.NotNull;
```
lognaturel: androidx?

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaGetRequestTest.java
```diff
@@ -28,22 +29,19 @@
...
     }
 
-    @After
-    public void teardown() throws Exception {
-        mockWebServer.shutdown();
-    }
-
     @Test
     public void makesAGetRequestToUri() throws Exception {
+        MockWebServer mockWebServer = mockWebServerTester.startMockWebServer();
```
lognaturel: Is the benefit of initializing `mockWebServer` locally just that tighter scope is better or is there another benefit I'm not seeing?

lognaturel (approved):
This is great!

As background, I contacted @seadowg privately about an issue with a custom server. It turned out the server had a problem. But @seadowg took the opportunity to make these great improvements. I aspire to be more like @seadowg.

It's particularly nice to have more confidence around the credential caching.

Regarding the string resources, no action needs to be taken. When we next release, we'll pull translations from Transifex and it has slightly different formatting from what Android Studio wants (which is fine).

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaServerClientProviderTest.java
seadowg: Urgh. Annoyling Android Studio is generating these. Bah. I'll fix them.

This code is "pure" Java so we could use the `javax` `@Nonnull` annotation. There is a lot of [confusion](https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use) over what the "standard" is in this case. I'll defer that decision to a time when we modularise this code and it doesn't have access to `androidx` anymore.

collect_app/src/test/java/org/odk/collect/android/http/OpenRosaGetRequestTest.java
seadowg: In a few tests we end up using more than one server so I ended up changing everything to this style. The `XRequestTest` actually only use one server so I could revert back to having field. I think I'll do that.

lognaturel: Ah, great.

bognasoldev89:
*Tested with success*

Verified on Android: 
4.4, 5.1, 6.0, 7.0, 8.1, 9.0

Verified cases:
- Add tests for Open Rosa authentication caching
- Downloading forms from Aggregate (restricted and not)
- Downloading forms from Central
- Downloading forms from Google Drive
- Sending submissions to Aggregate (restricted and not)
- Sending submissions to Central
- Sending submissions to Google Drive
- Android Studio profiler check on Network activity (Android 7.0)
- Rotating device during downloading and submitting
- Minimalizing device during downloading and submitting
- Swiping back and forth through available form lists
- Blocking device during downloading and submitting

It seems like Collect’s behavior didn’t change, I was able to download forms and send submissions on every tested Android - the app didn't crash at any moment.

Network profiler’s Request results seem consistent too, as it is visible on screenshots of examples (master vs pr):

Downloading form from Central

| | |
| --- | --- |
| 3339 master 7 getting form from central | 3339 pr 7 getting form from central |

Sending submission to Aggregate

| | |
| --- | --- |
| 3339 master 7 sending form to aggregate 1 | 3339 pr 7 sending form to aggregate 1 |
| 3339 master 7 sending form to aggregate 2 | 3339 pr 7 sending form to aggregate 2 |

PR #3341, opened 2019-09-14
[Fixes build] Only block finalization if form with missing instanceID also requests encryption

### LINKED ISSUE #3340: Only block finalization of forms without instanceID if those forms also define encryption
#### Problem description
#3265 attempted to block finalization of forms that request encryption but don't define an `instanceID`. Instead, it blocks finalization of any form without an `instanceID`. This was caught by `org.odk.collect.android.regression.FillBlankFormTest.searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName`.

Run test `org.odk.collect.android.regression.FillBlankFormTest.searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName` which incidentally verifies that a form without an `instanceID` is finalized.

#### Expected behavior
That test should pass and in general, the only case that forms without an `instanceID` should not be finalized is when they also require encryption.

lognaturel:
Closes #3340 

Added and ran automated tests; verified that `org.odk.collect.android.regression.FillBlankFormTest.searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName` now passes.

I considered doing a deeper refactor of `getEncryptedFormInformation` but there's no user-facing reason to do so and it's relatively high-risk because there's so much logic there.

I considered adding a targeted test for the case with no `instanceID` and no encryption but because these tests are expensive, I added a comment on `org.odk.collect.android.regression.FillBlankFormTest.searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName` instead. That should help us understand what's going on if it fails again and reminds us that the test has a dual purpose.

I replaced the pure junit test for the missing `instanceID` and encryption test with an Espresso test so that no refactoring of `getEncryptedFormInformation` would be needed. While I was there, I added a test to verify that actually encrypted forms can't be opened.

This brings the code block that checks for missing `instanceID` back to where it used to be so it should be safe. There's again a risk that any form without an `instanceID` or any form with encryption is affected.

See tests.

No


collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
```diff
@@ -0,0 +1,66 @@
...
+        int desiredImageId = InstanceListCursorAdapter.getFormStateImageResourceIdForStatus(desiredStatus);
+
+        onView(allOf(instanceOf(RelativeLayout.class),
+                    hasDescendant(withText(instanceName)),
+                    not(hasDescendant(instanceOf(Toolbar.class)))))
+                .check(matches(hasDescendant(withTagValue(equalTo(desiredImageId)))));
+        return this;
+    }
+
+    public EditSavedFormPage clickSavedFormWithDialog(String instanceName, int dialogMessageId) {
```
seadowg: I think I'd prefer to break this into the actions and assertions - I like "flow" methods for when the steps inside aren't as consequential for the test but I think it's good hear to expose the steps involved in the test itself. We could probably have a standard `Dialog` page that we could do `checkMessage` on. What do you think?

seadowg (changes requested):
Love it! I initially wasn't clear on how Espresso tests would look but I really like what you've come up with. It's also nice that these tests are isolated in their own `EncryptedFormTest`. Having tests that are broken down into features rather than areas of the code can often be easier to work with. It's really easy to build up a codebase that way (basically each feature/epic gets it's own test set) but harder when retrofitting stuff.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
lognaturel: I should have called that out. You already set up that pattern with `ErrorDialog` and `ExitFormDialog`. I think my reasoning was that this felt like a one-off but you're probably right that a generic `Dialog` page would be useful. Will do.

lognaturel: Let me know what you think of 86f6f41. Notice that I identify the dialog by the Ok button because I couldn't find a way to identify just any dialog. See also what you think of `ErrorDialog`.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/ErrorDialog.java
```diff
@@ -5,19 +5,21 @@
 import org.odk.collect.android.R;
 
 public class ErrorDialog extends Page<ErrorDialog> {
+    private final OkDialog okDialog;
```
seadowg: I really like the the `OkDialog`. Seems like the best way to verify "any" dialog at the moment. 

For me, it does feel like the `ErrorDialog` should inherit from the `OkDialog` rather than contain one. What do you think?

lognaturel: I feel like I had a reason but I don't even know anymore.

seadowg (approved):

lognaturel:
@opendatakit/testers please prioritize this so we can hopefully get back to a green build! ✅ ❌

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 6.0, 7.1, 8.1 and 9.0. 
I was able to build app, run tests EncryptedFormTest.java, run FillBlankFormTest.searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName, tested encrypted form with and without instanceId manually.

PR #3127, opened 2019-06-03
Refactored string widgets [removed code duplication + bug fixes]

grzesiek2010:
Closes #3125 

#### What has been done to verify that this works as intended?
I tested the attached forms.

#### Why is this the best possible solution? Were any other approaches considered?
As I mentioned in the issue there was a lot of code duplication. It was a cause of some bugs in ex widgets which we already fixed.
I removed the duplicated code and thanks to that fixed those bugs in ex widgets.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
The change is related to all string widgets (I mentioned them in the issue). All of them should be carefully tested.

#### Do we need any specific form for testing your changes? If so, please attach one.
I used: All widgets form: https://docs.google.com/spreadsheets/d/1af_Sl8A_L8_EULbhRLHVl8OclCfco09Hq2tqb9CslwQ/edit#gid=0
A form with just string widgets:
[stringWidgetsTest.xlsx](https://github.com/opendatakit/collect/files/3249258/stringWidgetsTest.xlsx)
Counter form:  
[counter.xml.txt](https://github.com/opendatakit/collect/files/3249280/counter.xml.txt)

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
No.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #3346, opened 2019-09-17
Refactored string widgets

### LINKED ISSUE #3125: Reduce the code duplication between String and ExString widgets
Collect v1.22.0 and older versions

#### Problem description
String widgets (StringWidget, IntegerWidget, DecimalWidget, StringNumberWidget) and Ex String widgets (ExStringWidget, ExIntegerWidget, ExDecimalWidget) have a lot in common but they don't share their code. That causes a lot of code duplication but also some limitations in ex widgets (things we fixed for normal string widgets like copy/paste option or placing the cursor at the end of the text).

grzesiek2010:
Closes #3125

I tested implemented changes manually using the attached form.

It's a fix for #3125 which does the job using a different approach than the one I tried in #3127.
In #3127 I started changing the hierarchy of widgets and ended up with a messy solution.
In this pr, I almost didn't touch the hierarchy. I've just implemented one small change visible below (`ExStringWidget` extends `StringWidget`):
Untitled Diagram

What I did:
1. I moved the duplicated code to a utility class called TextWidgetUtils. Thanks to that I reduced the code a lot and made widget classes smaller and more readable.
Additionally, TextWidgetUtils should be easier to test.
2. Fixed the bug that it was impossible to use copy/paste in external widgets
3. Fixed the bug that moving the cursor automatically at the end of EditText in external widgets didn't work;
4. Fixed the bug that `thousands-sep` option was available only in normal number widgets (not external).

@lognaturel if it's the approach you accept I'll add some tests as well so let me know.

It implements a lot of refactoring that is related to these widgets:
- StringWIdget
- IntegerWIdget
- DecimalWidget
- StringNumberWIdget
- ExStringWidget
- ExIntegerWidget
- ExDecimalWidget

all of the mentioned widgets should be tested.

[stringWidgets.xlsx](https://github.com/opendatakit/collect/files/3620537/stringWidgets.xlsx)

No.


seadowg:
I think this is a nice step to simplifying these Widgets! Having the major pieces of logic extracted will make them a lot easier to test in isolation as well as making it easier to share between the Widgets and also hopefully let us simplify the hierarchy even more over time. 

One thing I'd suggest is maybe putting the `TextWidgetUtils` class in `widgets.utils` or `widgets.helpers`.

seadowg (changes requested):
Just sticking this into "Request changes" as @grzesiek2010 is going to add some tests

grzesiek2010:
@seadowg 
I think it's ready for final review.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/StringWidgetsTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.odk.collect.android.activities.FormEntryActivity;
+import org.odk.collect.android.espressoutils.pages.FormEntryPage;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.test.FormLoadingUtils;
+
+public class StringWidgetsTest {
```
seadowg: I like the focus of these tests but personally it feels to me like it would be better to have this testing at a JUnit/Robolectric level rather than here in Espresso. It seems to be that each of these pieces of behaviour could be tested in test for the Widget concerned - the behaviours don't rely on many different components and seem to be the responsbility of the Widget alone. Tests for individual components can be faster which is a definite advantge but it can also make it easier to have a component's core behaviours tested in one place that you can be running repeatedly while making changes. I guess what I'm aiming for is that if I make a change to `XWidget` I can run `XWidgetTest` and have a pretty high chance of catching any problems.

How do you feel about changing it?

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/formentry/StringWidgetsTest.java
grzesiek2010: You are right I converted my tests.

collect_app/src/test/java/org/odk/collect/android/widgets/ExIntegerWidgetTest.java
```diff
@@ -34,4 +37,17 @@ public void setUp() throws Exception {
     private int randomInteger() {
         return Math.abs(random.nextInt()) % 1_000_000_000;
     }
+
+    @Test
+    public void digitsAboveLimitOfNineShouldBeTruncatedFromLeft() {
```
seadowg: This might be a "swipe left"/"swipe right" ambiguity problem but should these not all be `TruncatedFromRight`?

seadowg (approved):
Looks great! I have one comment on some test naming just to double check something but happy with this 👍

collect_app/src/test/java/org/odk/collect/android/widgets/ExIntegerWidgetTest.java
grzesiek2010: Good catch it's a test I copied from another one and didn't notice it.

grzesiek2010:
> Looks great! I have one comment on some test naming just to double check something but happy with this

Thanks, so I'm adding `needs-testing` label.

lognaturel:
I'm late to the party but I thought I'd add that I'm happy, too. 😊

mmarciniak90:
**Tested with success**

Verified on Android: 4.2, 4.4, 7.0, 8.1, 9.0

Verified cases:
- `All widgets` and `stringWidgets` forms
- launching an external app like counter
- different types of separators
- light/ dark mode
- English/Spanish language
- the behavior of `Launch` button
- rotating device
- External Integer widget works with thousands separator
- External Decimal widget works with thousands separator
- copy/paste
- cursor automatically at the end of EditText

PR #3352, opened 2019-09-18
Store the server URL that was imported by QR code

### LINKED ISSUE #3043: Store the server URL that was imported by QR code.
Collect v1.21.x, Android vx.x.x, device used...

#### Problem description
Only URLs that were introduced manually in General Settings/Server are stored in the drop-down URL list after using another server URL. User does not have to input the URL again, just choose it from the list, switching servers, when needed, is very easy and comfortable. 
When the server's URL is imported by QR code it is visible in the General Settings/Server right after import. But when you change it to default one or another introduced manually it is not visible on the drop-down URL list and to use it, the user has to input the URL manually again.

1. Introduce server URL on the one device, open Admin Settings, Import/Export Settings
2. On another device open Admin Settings, Import/Export Settings and Scan Code from the first device
3. Open General Settings/Server on the device on which you've just imported the settings, the URL introduced on the second device should be visible.
4. Click on URL, on displayed dialog use the drop-down arrow and change URL to another from the list. 
5. Click on URL again, imported server URL is not visible on the drop-down list.

#### Expected behavior
I think that imported URL should be stored on the URL list, as with Aggregate it is much easier to input manually the URL it is far more complicated with Central (nearly not possible at all). I think it might be a good idea to give users the opportunity to jump between Aggregate and Central without a must to scan the QR code every time especially when they are in the field and they do not have computers. 
Another good question is if the whole list of the URL used on 1st device should be stored on the 2nd or just the one used when QR code was scanned. 
@lognaturel @yanokwa @grzesiek2010 what do you think about that improvement suggestion?

#### Other information 

grzesiek2010:
Closes #3043 

I tested importing settings from QR code and from `collect. settings` file.

It just fixes the bug, there is no other solution than including the history of used urls (max 5) to exported settings.

Tested importing settings from QR code and from `collect. settings` file would be enough. It's not a risky pr.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
@@ -1,0 +1,72 @@
+package org.odk.collect.android.preferences.utielities;
+
+import androidx.test.platform.app.InstrumentationRegistry;
+import androidx.test.runner.AndroidJUnit4;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.R;
+import org.odk.collect.android.preferences.GeneralSharedPreferences;
+
+import java.util.List;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.odk.collect.android.preferences.utielities.KnownUrlListUtils.KNOWN_URL_LIST;
+
+@RunWith(AndroidJUnit4.class)
+public class KnownUrlListUtilsTest {
+
+    @Before
+    public void setUp() {
+        GeneralSharedPreferences.getInstance().reset(KNOWN_URL_LIST);
+    }
+
+    @Test
+    public void urlListShouldContainOnlyDefaultUrlIfNothingHasBeenAdded() {
+        List<String> urlList = KnownUrlListUtils.getUrlList();
+        assertEquals(1, urlList.size());
+        assertEquals(InstrumentationRegistry.getInstrumentation().getTargetContext().getString(R.string.default_server_url), urlList.get(0));
+    }
+
+    @Test
+    public void theLastUrlShouldBeRemovedWhenTheListHasMoreThan5ElementsButNotTheDefaultUrl() {
+        KnownUrlListUtils.addUrlToList("Url1");
+        KnownUrlListUtils.addUrlToList("Url2");
+        KnownUrlListUtils.addUrlToList("Url3");
+        KnownUrlListUtils.addUrlToList("Url4");
+        KnownUrlListUtils.addUrlToList("Url5");
+        KnownUrlListUtils.addUrlToList("Url6");
+        KnownUrlListUtils.addUrlToList("Url7");
+
+        List<String> urlList = KnownUrlListUtils.getUrlList();
+        assertEquals(5, urlList.size());
+
+        assertEquals(InstrumentationRegistry.getInstrumentation().getTargetContext().getString(R.string.default_server_url), urlList.get(4));
+    }
+
+    @Test
+    public void urlListShouldContainUniqueValues() {
+        KnownUrlListUtils.addUrlToList("Url1");
+        KnownUrlListUtils.addUrlToList("Url1");
+        assertEquals(2, KnownUrlListUtils.getUrlList().size());
+    }
+
+    @Test
+    public void urlThatAlreadyExistsShouldBeMovedToTheTopOfTheList() {
+        KnownUrlListUtils.addUrlToList("Url1");
+        KnownUrlListUtils.addUrlToList("Url2");
+        List<String> urlList = KnownUrlListUtils.getUrlList();
+        assertEquals(3, urlList.size());
+        assertEquals("Url2", urlList.get(0));
+        assertEquals("Url1", urlList.get(1));
+        assertEquals(InstrumentationRegistry.getInstrumentation().getTargetContext().getString(R.string.default_server_url), urlList.get(2));
+
+        KnownUrlListUtils.addUrlToList("Url1");
+        urlList = KnownUrlListUtils.getUrlList();
+        assertEquals(3, urlList.size());
+        assertEquals("Url1", urlList.get(0));
+        assertEquals("Url2", urlList.get(1));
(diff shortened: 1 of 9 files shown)


lognaturel:
What this does is make the saved list of servers part of the settings export. That means that if a device has a list of saved servers and then another device is configured from the first device, the second device will get the same saved list of servers.

I don't think that's the desired behavior. Each device should have its own history of saved servers. But if a device is configured by reading a QR code/collect.settings, the server that gets set from those settings should be added to the recent server list.

A concrete example:
- I currently have url1, url2 and url3 in my recent URLs list
- I scan the following QR code which puts a long Central URL in my URL list:
appuser
- I switch to the default server
- I should see the long, imported URL in my recent list

I would also expect url1, url2 and url3 to still be in my recent URLs list. I would expect the settings import to only reset settings that could have been included in the import but it looks like it resets all application settings, is that correct?

lognaturel (changes requested):
See note about expected behavior.

grzesiek2010:
Yes, you are right. I improved the solution and now it works as expected.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
```diff
@@ -0,0 +1,72 @@
+package org.odk.collect.android.preferences.utielities;
```
lognaturel: utielities -> utilities

priusak (approved):

grzesiek2010:
@lognaturel do you think it's ready for testing? Looks as if it's the last pr here which we would need to test for v1.24.

lognaturel:
It's on my list to look at by the end of the day. I only took a quick pass and it looks good. The only thing I've flagged so far is a typo.

lognaturel (commented):
Typo

collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.runner.RunWith;
+import org.odk.collect.android.R;
+import org.odk.collect.android.preferences.GeneralSharedPreferences;
+
+import java.util.List;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.odk.collect.android.preferences.utielities.KnownUrlListUtils.KNOWN_URL_LIST;
+
+@RunWith(AndroidJUnit4.class)
```
lognaturel: Why have this annotation? As far as I know it doesn't do anything and it's deprecated.

collect_app/src/main/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtils.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class KnownUrlListUtils {
+    static final String KNOWN_URL_LIST = "knownUrlList";
+
+    private KnownUrlListUtils() {
+    }
+
+    public static void addUrlToList(String url) {
```
lognaturel: The intended behavior of this method is non-obvious enough that it would be helpful to have a comment that describes its side effects.

collect_app/src/main/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        List<String> urlList = getUrlList();
+
+        if (urlList.contains(url)) {
+            urlList.remove(url);
+        } else if (urlList.size() == 5) {
+            if (urlList.get(4).equals(Collect.getInstance().getString(R.string.default_server_url))) {
```
lognaturel: This ensures that the default server URL is always preserved in the list, right? And that is new behavior? That doesn't seem right to me -- if the default URL hasn't been used, I think it makes sense to replace it.

lognaturel (changes requested):
👍 That's the behavior I had in mind!

I think the tests will be eventually candidates for moving to Robolectric but that will take doing something like adding a setter for the shared preferences in `GeneralSharedPreferences`. I'd be ok with this going in as-is and getting revised later but @seadowg may have some suggestions for making the change now.

Other than that, a couple questions inline.

seadowg:
@lognaturel @grzesiek2010 I think we could potentially have tests at a lower level but I'd want to look at reworking the way these kinds of prefs are stored (like @lognaturel brings up). Right now it's nice to have tests at the Espresso level expressing the feature in terms of what the user sees I think - I'd maybe change the name of the test to reflect that though. We could go with something like `ChangingServerURLTest`?

collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtilsTest.java
grzesiek2010: Right, it's not needed, fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtils.java
grzesiek2010: Ok I added a comment.

collect_app/src/main/java/org/odk/collect/android/preferences/utielities/KnownUrlListUtils.java
grzesiek2010: Yes, I don't know why but I was sure it worked in that way before, fixed.

grzesiek2010:
ok, I think I fixed everything.

lognaturel (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.2, 4.4, 5.1, 6.0, 7.1, 8.1 and 9.0

PR #3353, opened 2019-09-18
Analytics for Audio features

seadowg:
Relies on #3358 

This adds analytics based around audio features within forms. The goal is to be able to understand the usage of audio features in forms so we can make more informed calls in the future as to wheere our focus should be in terms of development and support of these features. The analytics events added are:

* Prompt is viewed that has a audio media attached to it's label (`Prompt`/`AudioLabel`)
* Prompt is viewed that has audio media attached to a select choice (`Prompt`/`AudioChoice`)
* Audio is autoplayed from prompt label (`Prompt`/`AutoplayAudioLabel`)
* Audio is autoplayed from prompt choice (`Prompt`/`AutoplayAudioChoice`)
* Prompt is viewed that has audio attached to a select choice in a `GridWidget` in "no buttons" mode (`Prompt`/`AudioChoiceGrid`)
* Prompt is viewed that has audio attached to a select choice in a `GridMultiWidget` in "no buttons" mode (`Prompt`/`AudioChoiceGrid`)

All events also include the form's analytics ID so that we are able to get numbers around how many forms contain these features rather than just how many times they "occur".

Used local debug logs with Emulator to check events are coming through.

Another solution might have been to have some upfront analysis done on the prompt/form and send events then rather than wait until features are presented in the UI. As we refactor the `FormEntryActivity` that might become more viable but right now it was pretty hard to find the correct place for that code to go.

There shouldn't be any changes in behaviour. It would be good for someone else to double check the events make sense and that we can analyse them properly using Firebase.


collect_app/src/main/java/org/odk/collect/android/formentry/AudioVideoImageTextLabel.java
```diff
@@ -95,8 +95,12 @@ public AudioVideoImageTextLabel(Context context, AttributeSet attrs) {
         ButterKnife.bind(this);
     }
 
-    public void setAVT(TextView labelTextView, String audioURI, String imageURI, String videoURI,
-                       String bigImageURI, ReferenceManager referenceManager, AudioHelper audioHelper) {
+    public void setAudio(String audioURI, AudioHelper audioHelper) {
+        setupAudioButton(audioURI, audioHelper);
+    }
+
+    public void setAVT(TextView labelTextView, String imageURI, String videoURI,
```
lognaturel: `AVT` here refers to "audio, video, text" so I think the method name needs to change.

collect_app/src/main/res/values-hi/strings.xml
```diff
@@ -239,7 +239,7 @@
   <string name="fs_delete_media_path_if_file_error">%sको नहीं मिटाया जा सका. कृपया फाइल को स्वयं हटा दें और फ़ॉर्म को पुनः डाउनलोड करें.</string>
   <string name="fs_create_media_folder_error">\'%s\' मीडिया फ़ोल्डर नहीं बना सका.</string>
   <string name="fs_file_copy_error">\'%2$s\' पर \'%1$s\' की नकल नहीं कर सकता. कारण: %3$s</string>
-  <string name="survey_internal_error">आतंरिक त्रुटि: prompt के स्टेप असफल.</string>
```
lognaturel: All the string changes look like a bad refactor and need to be reverted.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseGridWidget.java
```diff
@@ -73,16 +75,26 @@
...
         this.quickAdvance = quickAdvance;
-        noButtonsMode = WidgetAppearanceUtils.isCompactAppearance(prompt) || WidgetAppearanceUtils.isNoButtonsAppearance(prompt);
+        noButtonsMode = WidgetAppearanceUtils.isCompactAppearance(questionDetails.getPrompt()) || WidgetAppearanceUtils.isNoButtonsAppearance(questionDetails.getPrompt());
         itemViews = new View[items.size()];
 
         setUpItems();
         setUpGridView();
         fillInAnswer();
+
+        for (SelectChoice choice : items) {
```
lognaturel: How about logging once when the first audio file is found rather than for every choice? I don't think logging for every choice adds any info and there could potentially be a lot of choices.

Media is not supported for grid variants with buttons because there's no layout for putting media, right? I don't see that documented in https://github.com/opendatakit/docs/pull/1113/.

lognaturel (changes requested):
I'm not excited to see more reliance on Dagger but at least it's cleaned up and it's good to see the homegrown dependency injection go away.

I'm also not super excited about `QuestionDetails`. I see your point that it could be needed for more things that need to go into the widget hierarchy but right now the formid hash is not very compelling because it doesn't feel like a detail of the question. I might have gone with a middle ground where the `Analytics` class has a method called something like `logRemoteAnalyticsWithCurrentFormId` that makes the call on `Collect.getCurrentFormIdentifierHash` and then mocked `Analytics` for testing so the static call isn't a problem. I understand that's not a particularly elegant design, though, so I won't push it.

I am excited about analytics, though! 🎉

collect_app/src/main/java/org/odk/collect/android/formentry/AudioVideoImageTextLabel.java
seadowg: Great point. Will fix.

collect_app/src/main/res/values-hi/strings.xml
seadowg: Ach sorry that's a bad miss on my part. Well spotted.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseGridWidget.java
seadowg: Yeah you're right we should only log one. More and more I think it would be nicer to have these events happen before we get to the widgets. Going to have a lil think about that and the your point about `QuestionDetails` and see if there is an easy rework. I do for the moment feel like the `formAnalyticsID` is a detail of the question that the Widgets need but I don't 100% feel like they should need it.

seadowg:
@lognaturel replied to you about `QuestionDetails` (mostly in agreement) in a code comment.

On Dagger: I want to have a play with moving these dependencies out as part of the write up of View Guidlines we've discussed. Hopefully that can give us a path away from Dagger inside the Widgets.

seadowg:
@lognaturel good for another look. I'd like to come back to exploring placement of the analytics and having to pass `formAnalyticsID` after this is in. That way I don't hold up the release staring out my window pondering the most perfect implementation.

lognaturel (approved):

grzesiek2010 (approved):

lognaturel:
Will be verified and merged along with #3365 which contains the same commits.

PR #3357, opened 2019-09-20
Bumped targetSdkVersion and compileSdkVersion

### LINKED ISSUE #3189: App updates must target Android 9 (API 28) by November 2019
>Your app currently targets API level 26 and must target at least API level 28 to ensure it is built on the latest APIs optimized for security and performance.
>
>From August 2019, new apps must target at least Android 9.0 (API level 28).
> From November 2019, app updates must target Android 9.0 (API level 28).
> 
> https://developer.android.com/distribute/best-practices/develop/target-sdk.html

grzesiek2010:
Closes #3189 

I investigated the doc:
https://developer.android.com/about/versions/pie/android-9.0-changes-28
https://developer.android.com/about/versions/pie/android-9.0-changes-all
https://developer.android.com/about/versions/10/behavior-changes-10
https://developer.android.com/about/versions/10/behavior-changes-all

also, I tested the app a bit:
- downloading forms
- uploading forms (aggregate and google sheets)
- filling forms

We just have to update targetSdkVersion to 28 so we decided to update to 29 to test everything just once.

I wasn't able to find anything apart from the issue with TimeWidget but it doesn't mean that it's the only problem. Bugs might be everywhere...

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/TimeWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/TimeWidget.java
@@ -218,8 +218,7 @@ public class TimeWidget extends QuestionWidget implements ButtonWidget, TimePick
          */
         @SuppressWarnings("deprecation")
         private void fixSpinner(Context context, int hourOfDay, int minute, boolean is24HourView) {
-            // android:timePickerMode spinner and clock began in Lollipop
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
                 try {
                     // Get the theme's android:timePickerMode
                     final int MODE_SPINNER = 2;
@@ -251,13 +250,7 @@ public class TimeWidget extends QuestionWidget implements ButtonWidget, TimePick
                         Object delegate = delegateField.get(timePicker);
 
                         Class<?> spinnerDelegateClass;
-                        if (Build.VERSION.SDK_INT != Build.VERSION_CODES.LOLLIPOP) {
-                            spinnerDelegateClass = Class.forName("android.widget.TimePickerSpinnerDelegate");
-                        } else {
-                            // TimePickerSpinnerDelegate was initially misnamed in API 21!
-                            spinnerDelegateClass = Class.forName("android.widget.TimePickerClockDelegate");
-                        }
-
+                        spinnerDelegateClass = Class.forName("android.widget.TimePickerSpinnerDelegate");
                         // In 7.0 Nougat for some reason the timePickerMode is ignored and the
                         // delegate is TimePickerClockDelegate
                         if (delegate.getClass() != spinnerDelegateClass) {
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -67,12 +67,12 @@ def googleMapsApiKey = secrets.getProperty('GOOGLE_MAPS_API_KEY', '')
 def mapboxToken = secrets.getProperty('MAPBOX_ACCESS_TOKEN', '')
 
 android {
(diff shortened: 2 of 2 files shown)


grzesiek2010:
There is still a question I asked on slack about `minSdkVersion` now it is 16 and the range is pretty wide 16-29.
Based on firebase I can say that Android 4.x (all versions) is ~8% off all.
Maybe it's time to bump `minSdkVersion` as well? Users with Android 4 still would be able to use v1.23.

collect_app/src/main/java/org/odk/collect/android/widgets/TimeWidget.java
```diff
@@ -218,8 +218,7 @@ public void setTitle(CharSequence title) {
          */
         @SuppressWarnings("deprecation")
         private void fixSpinner(Context context, int hourOfDay, int minute, boolean is24HourView) {
-            // android:timePickerMode spinner and clock began in Lollipop
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
```
lognaturel, line 221: How did you determine this is now only relevant for N?

I think the block in https://github.com/opendatakit/collect/blob/4eb3b2b1d7be3e62ae249e450c526a0673d5b40b/collect_app/src/main/java/org/odk/collect/android/widgets/TimeWidget.java#L253 should also be updated accordingly since that test doesn't make sense anymore.

lognaturel:
I see @opendatakit/testers are in favor of limiting the API level range! It's true that it would be really nice. It's the kind of decision that is major enough that the TSC should consider it and that the community should be told a bit ahead. That means it's definitely not going to be for v1.24.

collect_app/src/main/java/org/odk/collect/android/widgets/TimeWidget.java
grzesiek2010: > How did you determine this is now only relevant for N?

The gist you have used is also updated in the same way: https://gist.github.com/jeffdgr8/6bc5f990bf0c13a7334ce385d482af9f

and also in https://issuetracker.google.com/issues/37119315 you can find that it's fin in API 25

> I think the block in should also be updated accordingly since that test doesn't make sense anymore.

right.

grzesiek2010:
Ahhh I missed one page, maybe even the most important one since is related to our app most: 
https://developer.android.com/about/versions/10/privacy/changes
today I tested emulator with android Q and bumping to 29 will require more changes for devices with Android Q, I know that it's probably not used much now but still.
Two most important changes we will need to implement are:
- [External storage access scoped to app files and media
](https://developer.android.com/about/versions/10/privacy/changes#scoped-storage
)
- [Restriction on non-resettable device identifiers](https://developer.android.com/about/versions/10/privacy/changes#non-resettable-device-ids)

It's not something super simple so I would vote for bumping to 28 (the upcoming release) and then to 29.

lognaturel (approved):

grzesiek2010:
@mmarciniak90 I did a rebase so that it includes the fix for https://github.com/opendatakit/collect/pull/3362

mmarciniak90:
I couldn't find a bug in this PR but I noticed 3 other bugs during testing this issue.  
* #3368 Invalid question is not highlighted in red on the field list on Android 5.1
* #3367 Settings pages have grey text on Android 4.x
* #3366 OpenStreetMap does not display on Androids 5 and higher 

I also confirm that this doesn't mean that fix is without problems so it will be better to merge it and have the opportunity to work with these changes all the time.

Verified cases:
- All widgets form
- Google Maps, Mapbox map 
- Upgrade
- Light/Dark mode
- Get a new form from Google Drive and Aggregate
- Autosend, autodelete - Google Drive and Aggregate - success and fail

PR #3358, opened 2019-09-21
Reimplement audio playback for GridWidget

### LINKED ISSUE #3354: Audio playback not working for `no-buttons` select choices
Collect v1.23.3

#### Problem description
When clicking on a select choice for a prompt that's using the `no-buttons` appearance (previously `compact`) there is an error if the select choice has attached audio.

1. Load [this example form](https://github.com/opendatakit/collect/files/3626538/gridwidget.zip)
2. Click on one of the options in the first question

#### Expected behavior
Previously the app would have played the audio when the choice was made - but only if `no-buttons` or `compact` was used. The easiest fix right now would be to simply remove the old code that's attempting to play the audio in this one case. Alternatively we can implement the new audio playback in `GridWidget` as we've already done for `SelectWidget` and `QuestionWidget`.

seadowg:
Closes #3354. This relies on #3326.

This adds the new audio playback functionality to `GridWidget` and `GridMultiWidget`.

Tested out selects with the relevant appearances (with and without audio) and ran tests.

The heirarchy of `GridWidget`/`GridMultiWidget`/`BaseGridWidget` seemed a little jumbled but I opted not to try and refactoring there as we're still following up with audio analytics after this.

One thing of note is that it seems that the `GridWidget` implements `MultiChoiceWidget` but no-ops the `setChoiceSelected` method and doesn't seemt o have the correct implementation of `getChoiceCount`. Pointing this out for someone with more context. Potentially we need an issue to rework that in some way.

Select one and multiples should now work with the `compact` appearance (and equivelent combinations with `no-buttons`). Would be good to test out these prompts with and without audio.


lognaturel (approved):
Looks good.

This is more about the general audio feature and may not even have been impacted by the recent changes but I notice that having a corrupt audio file now leads to a file missing toast which is misleading. It would be helpful for the message to make it clear that a file is found but that it can't be played. I'll file an issue tomorrow unless someone beats me to it.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1 and 9.0

verified cases:
- audio play for grid select one widget
- audio play for grid select multiple widget 
- the same media file for answers
- different media files for answers
- delete 1 media, confirmed that toast visible when choosing an answer
- delete another media, confirmed that toast visible when choosing an answer
- delete both media files, confirmed that toast visible when choosing both answers
- no differences between light and dark mode 
- answers still highlighted and audio play interrupted after minimizing app and screen rotation
- no regression identified for grid select widgets on All widgets form 

PR #3377, opened 2019-10-08
Use separate (more meaningful) messages when an audio file is missing/invalid

### LINKED ISSUE #3374: Corrupt/invalid audio file results in "file missing" toast
master (seen in https://github.com/opendatakit/collect/pull/3358); haven't tried to reproduce on v1.23

#### Problem description
When a corrupt or invalid audio file is set as a form attachment and there's an attempt to play that file, a toast that says the file is missing is displayed. This is confusing and could lead a user down the wrong troubleshooting path.

1. Use a form with media such as [gridwidget.zip](https://github.com/opendatakit/collect/files/3691137/gridwidget.zip)
1. Instead of the wav in the zip, put a text file with a wav extension in the `-media` folder
1. Try to play audio

#### Expected behavior
The error message should indicate that the file exists but can't be played. I think using the existing `file_invalid` text would be fine.

grzesiek2010:
Closes #3374 

I tested the attached form.

It's the simplest solution that improves `PlaybackFailedException` setting the message based on the cause.

It's not a risky change, testing the attached form should be enough.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/audio/PlaybackFailedException.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/PlaybackFailedException.java
@@ -5,9 +5,11 @@ import androidx.core.util.ObjectsCompat;
 public class PlaybackFailedException extends Exception {
 
     private final String uri;
+    private final int exceptionMsg;
 
-    public PlaybackFailedException(String uri) {
+    public PlaybackFailedException(String uri, int exceptionMsg) {
         this.uri = uri;
+        this.exceptionMsg = exceptionMsg;
     }
 
     public String getURI() {
@@ -36,5 +38,9 @@ public class PlaybackFailedException extends Exception {
         return "PlaybackFailedException{" +
                 "uri='" + uri + '\'' +
                 '}';
+    }
+
+    public int getExceptionMsg() {
+        return exceptionMsg;
     }
 }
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/audio/PlaybackFailedException.java
```diff
@@ -2,18 +2,32 @@
...
         this.uri = uri;
+        this.exceptionMessage = setUpExceptionMessage();
     }
 
     public String getURI() {
         return uri;
     }
 
+    private int setUpExceptionMessage() {
+        return new File(uri).exists() ? R.string.file_invalid : R.string.file_missing;
```
seadowg: I think it would be nicer for this decision to be made inside `AudioPlayerViewModel` where the error occurs. That way we don't need to care about reading the file in `ODKView`. What do you think?

Edit: in that world we'd also be able to write a test around the different scenarios and what error comes out of `AudioPlayerViewModel`.

seadowg (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.2, 4.4, 5.1, 6.0, 7.1 8.1 and 9.0.

verified cases:
- invalid file
- missing file
- one invalid, one missing
- both invalid
- both missing
- select multi with invalid and missing files
- autoplay with invalid and missing files

PR #3408, opened 2019-10-21
Added a listener to change text size immediately.Fix #3356

### LINKED ISSUE #3356: When settings values are changed while filling a form, the new values should take effect immediately
Collect v1.23 and prior though I think this worked as expected at some point.

#### Problem description
When the text font size is changed while a form is being filled, those changes don't take effect until the form is exited and the re-entered.

1. Open a form to fill
1. Tap the overflow menu, settings, user interface, change the font size
1. Tap back three times
1. Notice that the font size has not changed
1. Exit the form
1. Re-enter the form and notice that the font size has changed

#### Expected behavior
Font size setting should take effect immediately.

#### Other information 
Navigation type (swipe vs. buttons) takes effect immediately. I haven't looked in detail at other settings and when they apply.

ShivamPrajapati-dev:
Closes #3356

I had implemented an interface listener to update the form text size immediately.
I considered this approach because we have to reload the form after changing preferences.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -92,9 +92,9 @@ android {
         }
     }
 
-    lintOptions {
-        abortOnError false
-    }
+    //lintOptions {
+       // abortOnError false
+    //}
 
     buildTypes {
         // Release build for all forks
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -92,9 +92,6 @@ android {
         }
     }
 
-    //lintOptions {
-       // abortOnError false
-    //}
 
     buildTypes {
         // Release build for all forks
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -92,6 +92,10 @@ android {
         }
     }
 
+    lintOptions {
+        abortOnError false
+    }
+
     buildTypes {
         // Release build for all forks
         release {
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -113,6 +113,7 @@ import org.odk.collect.android.listeners.FormLoaderListener;
 import org.odk.collect.android.listeners.FormSavedListener;
 import org.odk.collect.android.listeners.PermissionListener;
 import org.odk.collect.android.listeners.SavePointListener;
+import org.odk.collect.android.listeners.TextSizeChangeListener;
 import org.odk.collect.android.listeners.WidgetValueChangedListener;
 import org.odk.collect.android.location.client.GoogleLocationClient;
 import org.odk.collect.android.logic.AuditEvent;
@@ -125,6 +126,7 @@ import org.odk.collect.android.preferences.AdminSharedPreferences;
 import org.odk.collect.android.preferences.GeneralKeys;
 import org.odk.collect.android.preferences.GeneralSharedPreferences;
 import org.odk.collect.android.preferences.PreferencesActivity;
+import org.odk.collect.android.preferences.UserInterfacePreferences;
 import org.odk.collect.android.provider.FormsProviderAPI.FormsColumns;
 import org.odk.collect.android.provider.InstanceProviderAPI.InstanceColumns;
 import org.odk.collect.android.tasks.FormLoaderTask;
@@ -197,6 +199,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         RankingWidgetDialog.RankingListener,
         SaveFormIndexTask.SaveFormIndexListener, FormLoadingDialogFragment.FormLoadingDialogFragmentListener,
         WidgetValueChangedListener,
+        TextSizeChangeListener,
         ScreenContext,
         AudioControllerView.SwipableParent {
 
@@ -295,6 +298,11 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         this.doSwipe = doSwipe;
     }
 
+    @Override
+    public void onTextSizeChange() {
+         loadForm();
+    }
+
     enum AnimationType {
         LEFT, RIGHT, FADE
     }
@@ -324,6 +332,8 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
+        UserInterfacePreferences.textSizeChangeListener = (TextSizeChangeListener) this;
 
         viewModel = ViewModelProviders.of(this, new FormEntryViewModelFactory()).get(FormEntryViewModel.class);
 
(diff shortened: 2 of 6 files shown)


collect_app/build.gradle
```diff
@@ -92,6 +92,10 @@ android {
         }
     }
 
+    lintOptions {
+        abortOnError false
```
seadowg: We want to keep this enabled so we know when our lint failing. This warns us about problems in the code around API level incompatibilities etc. Was it failing for you?

collect_app/src/main/java/org/odk/collect/android/preferences/UserInterfacePreferences.java
```diff
@@ -46,6 +47,7 @@
 public class UserInterfacePreferences extends BasePreferenceFragment {
 
     protected static final int IMAGE_CHOOSER = 0;
+   public static  TextSizeChangeListener textSizeChangeListener;
```
seadowg: Storing an `FormEntryActivity` instance in a static variable here is a problem as it creates a memory leak - the Activity won't be garbage collected and will stay in memory as this variable is still referencing it.

seadowg (changes requested):
Hey @ShivamPrajapati-dev! There are a couple of problems with the implementation I've noted in comments. 

As @lognaturel points out in the issue (#3356) navigation setting do update as soon as they're changed. It looks like that is because the navigation views are setup in the `FormEntryActivity`'s `onResume()`. 

I think for this there are a couple of ways we could implement it:

* We could always reload the form when returning from "General Settings" (or just if anything changes). We could potentially use `onActivityResult` for this.
* We could store the original text size in the `FormEntryActivity` and compare it in `onResume()` to see if we need to reload.

One behavioural question I do have is if the form needs to stay on the same question when returning from "General Settings" or if it's ok for it to start again at the beginning? I'm making the assumption that calling `loadForm()` (or however we "reload" the form) will start us at the beginning. @lognaturel do you have thoughts on this?

collect_app/build.gradle
ShivamPrajapati-dev: Yes, this fails for me even **before** making any changes in the code.

collect_app/src/main/java/org/odk/collect/android/preferences/UserInterfacePreferences.java
ShivamPrajapati-dev: I'll fix that.

collect_app/build.gradle
ShivamPrajapati-dev: But this occurs on my pc.

seadowg: Interesting. Is that from running `./gradlew lint`? I'm not seeing any failures on the current `HEAD` of `master` branch.

ShivamPrajapati-dev: It occurs from running ./gradlew checkcode

seadowg: What errors are you seeing?

ShivamPrajapati-dev: There are 3 errors in Propertymanager.java regarding missing permissions

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -906,6 +906,9 @@ protected void onActivityResult(int requestCode, int resultCode, final Intent in
                     getCurrentViewIfODKView().setBinaryData(bearing);
                 }
                 break;
+                case RequestCodes.FONT_SIZE_MATCHER:
+                    loadForm();
```
seadowg: I'm don't think we should be loading the form when returning here. This is quite intensive and also causes from to reload at the "hierarchy" view so the person filling the form out would lose their place. What we could do is call `refreshCurrentView` instead as that will just re-render the view.

seadowg, line 910: I'm don't think we should be loading the form when returning here. This is quite intensive and also causes from to reload at the "hierarchy" view so the person filling the form out would lose their place. What we could do is call `refreshCurrentView` instead as that will just re-render the view.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesActivity.java
```diff
@@ -62,6 +64,8 @@ public void onBackPressed() {
         } else {
             super.onBackPressed();
         }
+        Intent intent = new Intent();
```
seadowg, line 67: It doesn't right to me to always set a specific result when leaving the `PreferencesActivity`. This could happen in the `UserInterfacePreferences` fragment when the font size itself changes. You can call set result on a Fragment's host Activity by calling `getActivity()`. 

It would also be nice to write a test for the Fragement (probably using Robolectric) to check that it sets the result on it's parent Activity.

collect_app/src/main/java/org/odk/collect/android/utilities/ApplicationConstants.java
```diff
@@ -96,6 +96,7 @@ private ApplicationConstants() {
         public static final int GEOSHAPE_CAPTURE = 20;
         public static final int GEOTRACE_CAPTURE = 21;
         public static final int ARBITRARY_FILE_CHOOSER = 22;
+        public static final int FONT_SIZE_MATCHER = 23;
```
seadowg: Not sure about the name of this. Usually it's nice to have these express "what has happened". Could it be `FONT_SIZE_CHANGED`?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesActivity.java
seadowg: Ah sorry ignore me. Looking at the issue again (#3356) it seems like we want to refresh the form when any setting changes so this actually does feel right.

I still think we need a test for this though!

collect_app/src/main/java/org/odk/collect/android/utilities/ApplicationConstants.java
seadowg: As with my above comment this should be for any setting change so maybe this should be `PREFERENCES` or something similar.

ShivamPrajapati-dev: Ok, I will resolve that.

ShivamPrajapati-dev: I didn't create a feature branch here for the issue. So will I have to close this PR and create a new one from the feature branch?

seadowg: Yeha I think so. As far as I know there isn't a way to edit the branch of a PR.

seadowg:
@ShivamPrajapati-dev hey! Are you planning to continue work on this?

ShivamPrajapati-dev:
> @ShivamPrajapati-dev hey! Are you planning to continue work on this?

Yeah, I will continue to work on this.

seadowg:
@ShivamPrajapati-dev it looks like you've push code that's for #3489.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesActivity.java
ShivamPrajapati-dev: Can you @seadowg sir, please guide what do from here.

seadowg:
Hey @ShivamPrajapati-dev! Are you still planning to continue work on this PR?

ShivamPrajapati-dev:
> Hey @ShivamPrajapati-dev! Are you still planning to continue work on this PR?
No, you can assign this to someone.

Issue #3382, opened 2019-10-09
Revise `AudioVideoImageTextLabel` spacing and layout


seadowg:
This view is the basically the "header" for any question:

![Screenshot_1570551839](https://user-images.githubusercontent.com/556280/66482256-788ed300-ea9a-11e9-8fa8-34ba94a9f38f.png)

The view is also used in the select choices. You can see from the screen shot that there are two cosmetic problems:

1. Spacing around the items is probably too small. You can see the scrollbar actually overlaps the buttons on the right slightly.
2. The label and the audio button are layed out at slightly different heights. This can make it a little hard to differentiate what section owns what label/control.

My suggestion would be to look at trying to look at fixing these problems while also reworking the typography for this view in line with that ongoing effot (https://forum.opendatakit.org/t/reworking-collects-typography/20634/3).

ABHISHEK-AMRUTE:
@opendatakit-bot claim

seadowg:
Hey @ABHISHEK-AMRUTE! Thanks for having a look at this. Just as a heads up I'm working on #425 which may end up having some overlapping work. Probably good for us to push any work early (as draft PRs) so we can keep an eye out for conflicts!

seadowg:
@opendatakit-bot claim

seadowg:
Clean up tasks to get this closed now #3455 is merged:

- [ ] Fix spacing for text widgets

Additional clean up I'll follow up with:

- [ ] Rework `QuestionWidget` so it is defined using XML - right now it's hard to work with layouts for widgets as everything is defined procedurally
- [ ] Rework `ODKView` so it is defined using XML - right now it's hard to work with layouts for widgets as everything is defined procedurally
- [ ] Add an issue for revising each individual widget's typography and spacing

PR #3413, opened 2019-10-22
Improve tappability of select options

### LINKED ISSUE #425: Radio buttons with few options: add more space between response rows
When a radio button question has a small number of answer options: (e.g., male or female) it would be nice to separate them a bit, to decrease the likelihood of accidentally pressing the wrong button. 

Our recent experiment on date data entry errors also showed one error in every 50 entries of sex.  Spacing the responses farther apart might help drive this error rate down where it belongs...well below one error in every 100 entries.

seadowg:
Closes #425 

This adds spacing around the checkbox and text in select one and select multi widgets. It also removes the extra margins on the left/start of widgets that ended up creating an "untappable" space for selects. The result looks a little strange but we'll redefine the spacing/layout for widgets and their "header" in #3382 before the next release.

I've also done some work to reduce the circular dependency between `AbstractSelectListAdapter` and `SelectWidget` which made getting to the right place to make changes confusing. There is still work to do on that front so I've deprecated (with a comment) the offending reference.

We've played around with several screen sizes (as you'll see in the comments).

We also talked about having the size differ for different screen formats but elected to wait and see if we get feedback around this new sizing for selects being too big on smaller devices.

There shouldn't be any actual behaviour changes other than it being easier to tap the select choices. It would be good to play around with a few screen sizes and forms to check that we haven't created any visual regressions (other than the cosmetic problems we've introduced by removing the left margin).


seadowg:
As discussed in the issue I've upped the height of the select options and removed the horizontal padding around widgets. This is where I am at the moment:

Screenshot_1571782624

The "zero" padding on widgets doesn't look great at all but it already looked way to small so tempted to say this a cosmetic hit we can take? I like that this lets us think about the widget designs a little more individually which might give us a risk of inconsistent margins as we add them in but does give us more control for widgets where we need side-to-side spanning controls. I also think keep consistent margins between screens isn't that hard with a well defined set of default values. An actual start to defining some spacing and margins could happen in #3382.

A weirdness here is that I can't seem to tweak the padding on the left of the radio button as its `paddingLeft` controls the distance between the button and the text. I've asked about this on the Material Components discord. An alternative approach is to nest the `RadioButton` and have the parent toggle it but I'm not super into that.

lognaturel:
The vertical spacing feels like a really big change. Tapping will be more accurate which is great but finding the correct choice will require about twice as much scrolling when lists go off the end of the screen. That can also introduce error and certainly slows things down.

It's a very different user experience:
 

This is even worse on a shorter screen like my Alcatel M'Pop which is 3.5" tall.

I'm not sure how to reason about this. Is there a middle ground in terms of spacing that would balance tappability and finding the right option? Is it better to think about short and long lists separately? Do we just trust the Material Design folks?

Full disclosure: I use Gmail, by default the inbox view rows are standard Material Design height like in this PR, and I always use it in compact mode. That is to say, I have a personal bias for fitting more items in vertically. This is a case where it'd be really beneficial to be able to get direct feedback from a variety of field users.

I'm not quite clear on what you're proposing for the horizontal spacing. When you say "but it already looked way to small so tempted to say this a cosmetic hit we can take", do you mean for a release or in the interim? I don't think we can release with that zero padding because it feels distractingly not right.

seadowg:
@lognaturel thoughts:

> I'm not sure how to reason about this. Is there a middle ground in terms of spacing that would balance tappability and finding the right option? Is it better to think about short and long lists separately? Do we just trust the Material Design folks?

I think what might be an option is to look at using different sized selects for different screen sizes. This would be easy enough to achieve with Android resource qualifiers I think. Perhaps a slightly bigger size than before for "phone" screens and this larger size for "tablet" screens. What do you think?

> I'm not quite clear on what you're proposing for the horizontal spacing. When you say "but it already looked way to small so tempted to say this a cosmetic hit we can take", do you mean for a release or in the interim? I don't think we can release with that zero padding because it feels distractingly not right.

Yeah sorry that's not clear. Yes I mean going with this and then revising as part of #3382 before a release.

lognaturel:
> Perhaps a slightly bigger size than before for "phone" screens and this larger size for "tablet" screens

Yes, that makes sense to me.

> going with this and then revising as part of #3382 before a release.

👍 

Side by side from M'Pop:
 

I have to admit that the greater spacing is growing on me. 

The Material Design guidelines don't include horizontal lines between rows. I think they are necessary with the compact layout to provide some visual separation between options but perhaps they're not if the spacing is greater?

seadowg:
> The Material Design guidelines don't include horizontal lines between rows. I think they are necessary with the compact layout to provide some visual separation between options but perhaps they're not if the spacing is greater?

Great point. In fact, looking at the ["Divider" section](https://material.io/components/dividers/#types) of the spec we're specifically told not to use them for single item lists. I'll have a look at removing them.

seadowg:
@lognaturel actually I've realized that our selects aren't always just a single item and actually can contain content that would warrant a divider:

Screenshot_1572272704

It does seem like (again looking at the [Divider docs in Material Design](https://material.io/components/dividers/#types)) that we should be using the "inset" dividers instead of the full bleed ones. I'll change them.

seadowg:
Ok here's the inset dividers:

Screenshot_1572278723

seadowg:
I'm also looking at severing (as much as seems reasonable) the circular dependency between `AbstractSelectListAdapter` and `SelectWidget`. This was based on it being harder than it probably should have been to understand how to make the visual changes I needed to make.

seadowg:
@lognaturel After thinking about this some more I think we should leave the sizing large for every device and then post a "test" build on the forum after #3382 to see if we can get some early feedback on the new look. I'd rather not introduce the complexity of different layouts unless we get feedback around the sizing being a problem. Does that make sense? If so I'll do a quick tidy up and then undraft this.

lognaturel:
Works for me, @seadowg. The sooner it can go out the better.

collect_app/src/main/java/org/odk/collect/android/adapters/SelectOneListAdapter.java
```diff
@@ -101,6 +104,7 @@ RadioButton setUpButton(final int index) {
         adjustButton(radioButton, index);
         radioButton.setOnClickListener(this);
         radioButton.setOnCheckedChangeListener(this);
+        radioButton.setPadding(0, pxFromDp(widget.getContext(), 16), 0, pxFromDp(widget.getContext(), 16));
```
lognaturel: Can/should 16 be a named value?

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
```diff
@@ -120,9 +120,9 @@ protected void publishResults(CharSequence charSequence, FilterResults filterRes
         };
     }
 
-    abstract CompoundButton setUpButton(int index);
+    abstract CompoundButton createButton(int index);
 
-    void adjustButton(TextView button, int index) {
+    void setupButton(TextView button, int index) {
```
lognaturel: Nit: should be `setUp` (phrasal verb). It's pretty consistent across other classes so best to continue.

collect_app/src/main/res/layout/select_multi_item.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.appcompat.widget.AppCompatCheckBox xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:padding="16dp"
```
lognaturel: Same question as above? Is the 16 a named value already? Can it be introduced as one or should that be part of #3382 maybe?

lognaturel (commented):
Looks good to me. I like the `Deprecated` usage in this case.

@grzesiek2010 any questions/comments/concerns?

Again, the idea is to merge this and a first pass at #3382 as soon as possible so we can share a test build with the TSC and beta users. Even if we decide to go with a less aggressive change, the refactors will have value.

collect_app/src/main/java/org/odk/collect/android/adapters/SelectOneListAdapter.java
seadowg: Ah that's from an older commit. But yeah for the other random 16s we have an `@dimen` for that.

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
seadowg: Good catch. Will fix.

lognaturel (approved):

lognaturel:
@opendatakit/testers to reiterate some things from the PR description, you should expect the left and right margins to look bad. Those will be addressed in #3382. Other than that, it should be easier to tap choices for all selects, including with media.

mmarciniak90:
@seadowg I noticed that inset dividers on devices with Android 5.1 and older are red.

| | | |
|---|---|---|
|Screenshot_2019-11-08-15-33-42|Screenshot_2019-11-08-15-40-14|Screenshot_2019-11-08-15-34-11|

seadowg:
@mmarciniak90 huh. Nice find 🎖. I'll take a look at fixing that.

seadowg:
@mmarciniak90 that's fixed now. Basically a way we refer to our  "theme" (light or dark) colors doesn't work before API 23 (Marshmallow) so checking Android 5.1 whenever we make visual changes is a great idea!

mmarciniak90:
@seadowg I can confirm that lines are correct on Android 5.1 but the problem is not fixed for Android 4.4 and 4.2

seadowg:
@mmarciniak90 ach my bad for not checking. Will have a look.

seadowg:
@mmarciniak90 ok apparently the work around we were using didn't work on API 20 and below so we need a workaround for our workaround. Gargh. Fixed now.

mmarciniak90:
@seadowg thanks for so fast reaction. I see progress because the lines are not red but still, there is a difference in Android 4.4 and 4.2.

| Android 5.1 | Android 4.4|
|---|---|
|Screenshot_2019-11-12-15-01-19|Screenshot_2019-11-12-15-01-43|
|Screenshot_2019-11-12-15-09-30|Screenshot_2019-11-12-15-09-56|

seadowg:
@mmarciniak90 ah yeah sorry I saw that too and forgot to mention it - the transparency doesn't work on Android 4.4 and below. Personally I'd vote for leaving this though as I think it's worth avoiding more workarounds that we'll phase out later for little (subjective I know) details like this. What do you think?

mmarciniak90:
@seadowg why do you decide to use transparency instead of grey color? 
I'm a fan of consistency :smile: so let's leave the decision @lognaturel

seadowg:
@mmarciniak90 that's a really good question. It's all to do with the Material Design system and how it interacts with Android. The grey color there is based on a our "On Surface" color (docs around the color system [here](https://material.io/design/color/the-color-system.html#color-theme-creation)). Material Design is a big fan of introducing different emphasis levels for your colors - low, medium and high. The motivation is that it lets designers work with a smaller number of colors buit still gives them a scale to play with in each color for visual items with different levels of importance. On Android we define these emphasis levels using different transparancy levels so if we change the "On Surface" color we don't have to also go and change all it's different "emphasis" colors. This is all great except for the features needed to make it all work of course don't all work in older versions. *sigh*

If we want to work around this we can just define the emphasis as colors along side the "normal" ones. I'd prefer not to to keep the number of different attributes for our style down and also make change easier. That said it is a simpler way of defining them. What say you @lognaturel?

lognaturel:
Thanks for that thorough explanation, @seadowg, and for always pushing us to do things consistently, @mmarciniak90! In this case, I think we should aim for code simplicity. I agree with @seadowg that although the appearance in Android 4.x is not as nice, it's not not terrible (red was). My vote goes for leaving things as they are now.

mmarciniak90:
@lognaturel thanks for voting! :)

**Tested with success**

Verified on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 10.0

Issue #3388, opened 2019-10-11
Collect crashes when use range picker integer widget on fieldlist form


kkrawczyk123:
#### Software and hardware versions 
Collect v1.24.x, Android v5.1, 7.0, other versions probably too, device used...

#### Problem description
Collects crashes after choosing an answer on number picker of range picker integer widget on fieldlist type of form. 
logs:
`2019-10-11 15:51:22.821 27776-27776/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 27776
    java.lang.ClassCastException: android.widget.ImageButton cannot be cast to android.view.ViewGroup
        at androidx.fragment.app.FragmentManager.getFragmentContainer(SourceFile:2612)
        at androidx.fragment.app.FragmentManager.setVisibleRemovingFragment(SourceFile:2596)
        at androidx.fragment.app.FragmentManager.removeFragment(SourceFile:1911)
        at androidx.fragment.app.BackStackRecord.executeOps(SourceFile:408)
        at androidx.fragment.app.FragmentManager.executeOps(SourceFile:2584)
        at androidx.fragment.app.FragmentManager.executeOpsTogether(SourceFile:2372)
        at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(SourceFile:2327)
        at androidx.fragment.app.FragmentManager.execPendingActions(SourceFile:2230)
        at androidx.fragment.app.FragmentManager$3.run(SourceFile:414)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6165)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:778)
2019-10-11 15:51:22.825 1971-5829/? W/ActivityManager:   Force finishing activity org.odk.collect.android/.activities.FormEntryActivity
2019-10-11 15:51:22.831 2745-8399/? D/ProcSpeedReader: Failed to read freq stats from null
[...]

Issue #3384, opened 2019-10-10
Crash when user loads form with rating widgets


mmarciniak90:
#### Software and hardware versions 
Collect master
Crash does not occur on v1.23.3

#### Problem description
Crash when user loads form with rating widgets

#### Steps to reproduce the problem
1.  User loads form
[TestRatingWidget.xml.txt](https://github.com/opendatakit/collect/files/3711553/TestRatingWidget.xml.txt)

#### Expected behavior
* User is able to fill the form

#### Other information 
2019-10-10 11:25:28.039 16517-16517/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 16517
    java.lang.ClassCastException: android.widget.ImageButton cannot be cast to android.view.ViewGroup
        at androidx.fragment.app.FragmentManager.getFragmentContainer(FragmentManager.java:2612)
        at androidx.fragment.app.FragmentManager.setVisibleRemovingFragment(FragmentManager.java:2596)
        at androidx.fragment.app.FragmentManager.removeFragment(FragmentManager.java:1911)
        at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:408)
        at androidx.fragment.app.FragmentManager.executeOps(FragmentManager.java:2584)
        at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2372)
        at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:2327)
        at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:2230)
        at androidx.fragment.app.FragmentManager$3.run(FragmentManager.java:414)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6692)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1468)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1358)

mmarciniak90:
@opendatakit-bot label "regression"

PR #3420, opened 2019-10-28
Downgraded androidx.lifecycle:lifecycle-extensions

### LINKED ISSUE #3418: Fill Blank Form page is always empty in Android 10
This was reported on the forum: https://forum.opendatakit.org/t/nothing-available-to-display-on-fill-blank-form-page/22974/2.

Collect v1.24, Android v10, Installed from Play Store (doesn't appear in debug builds)

#### Problem description

Downloaded forms are not displayed in "Fill Blank Form" page on Android 10.l

On an Android 10 device with Collect installed from the Play Store:
1. Download a form
2. Open "Fill Blank Form"

Form is not shown.

#### Expected behavior

Downloaded forms should be shown.

### Other information

After digging in a little the problem appears to have appeared in https://github.com/opendatakit/collect/commit/237121bf0c5938a65f3d7f04f7285e65ba3c8218. It's only reproducible on builds using ProGuard (`minifyEnabled(true)` in `build.gradle`).

grzesiek2010:
Closes #3418 

I tested the fix using `odkCollectRelease` build.

Seems like a bug in `androidx.lifecycle:lifecycle-extensions:2.2.0-beta01+`. The problem is that [onLoadFinished()](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormChooserListActivity.java#L227) is not called, what interesting it works well in another place FileManagerFragment, the only difference is that there it's used in a Fragment and here it's used in an Activity, so it might be a bug related to activities).

When it comes to loaders they have been deprecated so I filled an issue to replace them https://github.com/opendatakit/collect/issues/3419 I can't be sure but I believe it will allow us to upgrade `androidx.lifecycle:lifecycle-extensions`

To test the issue we need a device with Android 10.
Then we need to update [debug build](https://github.com/opendatakit/collect/blob/master/collect_app/build.gradle#L118):
adding:
```
minifyEnabled(true)
proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
```

Please also confirm that problems with `ClassCastExceptions` are not visible again: https://github.com/opendatakit/collect/pull/3395

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -275,7 +275,7 @@ dependencies {
     odkCollectReleaseImplementation "com.squareup.leakcanary:leakcanary-android-no-op:${rootProject.leakyCanaryVersion}"
 
     // Android Architecture Components:
-    implementation "androidx.lifecycle:lifecycle-extensions:2.2.0-beta01"
+    implementation "androidx.lifecycle:lifecycle-extensions:2.1.0"
 
     // Dagger:
     implementation "com.google.dagger:dagger:${rootProject.daggerVersion}"


lognaturel (approved):

lognaturel:
@grzesiek2010 Can you remind me why we even tried to use alpha and beta releases? I thought we had agreed to only use them when no full release was compatible with everything else. In this case, if `2.1.0` was compatible, we should not have been using pre-release versions.

grzesiek2010:
> I thought we had agreed to only use them when no full release was compatible with everything else

That's true but I think we agreed on it after we upgraded the lib to `2.2.0-alpha04` which was buggy and I bumped it to `2.2.0-beta01` which turned out to be buggy as well. Of course, we could have downgraded it after discovering those bugs with `ClassCastException` but we didn't.

seadowg:
Just marking this as `high priority` on the assumption we want to get it out as a hotfix for 1.24.

mmarciniak90:
**Tested with success**

Tested on **emulator** with Android 10
Verified also on Androids: 9.0, 8.1, 7.0, 6.0, 5.1, 4.4

Verified cases:
* availability the form which was download from Aggregate on `Fill Blank Form` view
* availability the form which was download from Google Drive on `Fill Blank Form` view
* filled `all-widgets-fieldlist-noosm.xml` form  - #3388
* filled `TestRatingWidget.xml` form - #3384

PR #3424, opened 2019-10-29
Added automated tests with repeat groups

mmarciniak90:
### What has been done
* Added `FillBlankFormWithRepeatGroupTest` class which contains test cases for repeat groups
* Created 5 test cases


collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
```diff
@@ -0,0 +1,197 @@
...
+            .around(new CopyFormRule("form6.xml"))
+            .around(new CopyFormRule("form7.xml"))
+            .around(new CopyFormRule("form8.xml"))
+            .around(new CopyFormRule("form9.xml"))
+            .around(new CopyFormRule("RepeatGroupAndGroup.xml"))
+            .around(new CopyFormRule("basic.xml"))
+            .around(new CopyFormRule("repeat_group_form.xml"));
+
+    @Test
+    public void repeatGroupsNotAdded_ShouldNotDoubleLastMessage() {
```
seadowg: Not sure I understand the name of this test. What's the "last message" here?

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        new MainMenuPage(main)
+                .startBlankForm("form9")
+                .closeSoftKeyboard()
+                .swipeToNextQuestion()
+                .clickSaveAndExit();
+    }
+
+    @Test
+    public void repeatGroupNotAdded_ShouldBackButtonBeActive() {
```
seadowg: Again not sure I understand this test name. What are we testing for here?

seadowg (changes requested):
Just some confusion around test names!

seadowg:
Ran through test lab and looks good ✅

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
mmarciniak90: I changed it to 'ShouldNotDoubleLastQuestion'. 
We had a problem where this last question was doubled when no group was added.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
mmarciniak90: I changed the name to 'ShouldBackwardButtonBeClickable'
We had a bug where the backward button was inactive and unclickable when no group was added.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
seadowg: Ah cool. Could it be `whenNoRepeatGroupAdded_doesNotDoubleLastQuestion`? I find adding `when` helps clarify it's the "conditional" in the test. Some people really like `should` but I find it is just noise - we're not hoping ("it should do this") we're stating ("it does do this").

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
seadowg: I think as in the above this could be `whenNoRepeatGroupAdded_backButtonIsClickable`.

seadowg: As with the above I think this could be `whenNoRepeatGroupAdded_backButtonIsClickable`.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .inputText("2")
+                .closeSoftKeyboard()
+                .clickOnGoToIconInForm()
+                .checkIsTextDisplayed("2")
+                .clickJumpEndButton()
+                .clickSaveAndExit();
+    }
+
+    @Test
+    public void openHierarchyPageFromLastView_ShouldNotDisplayError() {
```
seadowg, line 183: This could be something like `openingHierarchyPageFromLastView_doesNotDisplayError`.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .closeSoftKeyboard()
+                .clickBackwardButton()
+                .clickOnDoNotAddGroup()
+                .closeSoftKeyboard()
+                .swipeToNextQuestion()
+                .clickSaveAndExit();
+    }
+
+    @Test
+    public void firstQuestionWithLongLabel_ShouldDisplayBothAnswersInHierarchyPage() {
```
seadowg: How about `whenFirstQuestionHasLongLabel_bothAnswersAreDisplayedInHierarchyPage`?

seadowg (changes requested):
Added a couple more  comments around test naming. Sorry to be so picky with all this! I'll try and write a little guidline piece on test naming so we can all have a bigger conversation about it.

Oh and definitely let me know if you think any of my suggestions are hard to understand. This is all super subjective!

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
mmarciniak90: As I remember, we've determined that the names contain 'Should'

seadowg: Happy to leave `should` in then if that's the case. It doesn't really change the meaning of the test.

collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormWithRepeatGroupTest.java
lognaturel: How do you know I'm not just hoping?! I do see the point about `should`. I've moved away from using it myself. I think it's ok to use or not as long as the meaning is clear.

seadowg (approved):
Looks great thanks 👍

mmarciniak90:
@seadowg is it ready to merge or should I change anything else?

seadowg:
Ah sorry I was waiting to see if CI passed after it failed (due to rotation problems) yesterday but then forgot to merge. Will do now!

PR #3426, opened 2019-10-29
Catch SQLiteExceptions during reading csv files

### LINKED ISSUE #3425: SQLiteException in ItemsetDbAdapter.java:115
Collect v1.24.0 and older versions

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8258efdc52e37fbf543edb7a7804fb3b?time=last-seven-days&sessionId=5DB54C43018100010F51B09C02FB19E0_DNE_0_v2

`Caused by android.database.sqlite.SQLiteException: unrecognized token: ""��l" (code 1 SQLITE_ERROR[1]): , while compiling: create table itemset_9b47b219ac9093cfe8e0071fb338c49e (_id integer primary key autoincrement  , "��list_name" text  , "name" text  , "label::english" text  , "governorate" text  , "district" text  , "label::arabic" text );
       at android.database.sqlite.SQLiteConnection.nativePrepareStatement(SQLiteConnection.java)
       at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1229)
       at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:703)
       at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
       at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:59)
       at android.database.sqlite.SQLiteStatement.<init>(SQLiteStatement.java:31)
       at android.database.sqlite.SQLiteDatabase.executeSql(SQLiteDatabase.java:2227)
       at android.database.sqlite.SQLiteDatabase.execSQL(SQLiteDatabase.java:2155)
       at org.odk.collect.android.database.ItemsetDbAdapter.createTable(ItemsetDbAdapter.java:115)
       at org.odk.collect.android.tasks.FormLoaderTask.readCSV(FormLoaderTask.java:547)
       at org.odk.collect.android.tasks.FormLoaderTask.processItemSets(FormLoaderTask.java:301)
       at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:208)
       at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:68)
       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
       at java.lang.Thread.run(Thread.java:764)`

grzesiek2010:
Closes #3425 

I tested the solution but only hardcoding `throw SQLiteException`.

Catching the exception is the only solution. We can't be sure what characters users put in their csv files, some of them might cause such an exception.

Additionally, I added a warning message which tells users that there was something wrong with their csv file. Unlike `errorMsg`, `warningMsg` cause that only a toast is displayed and a user can continue filling the form.

I wasn't able to reproduce the issue so testing would be difficult. I think testing any form with external itemsets would be enough to confirm there is no regression and that's all. I can also prepare a testing version with hardcoded `throw SQLiteException` if needed.

Generally a form with external itemsets but as I said above I wasn't able to reproduce the issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2182,7 +2182,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 FormController fec = formLoaderTask.getFormController();
                 if (fec != null) {
                     if (!readPhoneStatePermissionRequestNeeded) {
-                        loadingComplete(formLoaderTask, formLoaderTask.getFormDef());
+                        loadingComplete(formLoaderTask, formLoaderTask.getFormDef(), null);
                     }
                 } else {
                     dismissFormLoadingDialogFragment();
@@ -2345,7 +2345,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      * - Initializes background location capture (only if the instance being loaded is a new one)
      */
     @Override
-    public void loadingComplete(FormLoaderTask task, FormDef formDef) {
+    public void loadingComplete(FormLoaderTask task, FormDef formDef, String warningMsg) {
         dismissFormLoadingDialogFragment();
 
         final FormController formController = task.getFormController();
@@ -2478,6 +2478,11 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 activityDisplayed();
 
                 refreshCurrentView();
+
+                if (warningMsg != null) {
+                    ToastUtils.showLongToast(warningMsg);
+                    Timber.w(warningMsg);
+                }
             }
         } else {
             Timber.e("FormController is null");
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
@@ -16,6 +16,7 @@ package org.odk.collect.android.tasks;
 
 import android.content.Intent;
 import android.database.Cursor;
+import android.database.SQLException;
 import android.os.AsyncTask;
 
 import org.javarosa.core.model.FormDef;
@@ -70,6 +71,7 @@ public class FormLoaderTask extends AsyncTask<String, String, FormLoaderTask.FEC
 
     private FormLoaderListener stateListener;
     private String errorMsg;
+    private String warningMsg;
     private String instancePath;
     private final String xpath;
     private final String waitingXPath;
@@ -469,7 +471,7 @@ public class FormLoaderTask extends AsyncTask<String, String, FormLoaderTask.FEC
                     if (wrapper == null) {
                         stateListener.loadingError(errorMsg);
                     } else {
-                        stateListener.loadingComplete(this, formDef);
+                        stateListener.loadingComplete(this, formDef, warningMsg);
                     }
                 }
             } catch (Exception e) {
@@ -559,8 +561,8 @@ public class FormLoaderTask extends AsyncTask<String, String, FormLoaderTask.FEC
                 ida.addRow(pathHash, columnHeaders, nextLine);
 
             }
-        } catch (IOException e) {
-            Timber.e(e, "Exception thrown while reading csv file");
+        } catch (IOException | SQLException e) {
+            warningMsg = e.getMessage();
         } finally {
             if (withinTransaction) {
                 ida.commit();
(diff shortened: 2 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -559,8 +561,9 @@ private void readCSV(File csv, String formHash, String pathHash) {
                 ida.addRow(pathHash, columnHeaders, nextLine);
 
             }
-        } catch (IOException e) {
-            Timber.e(e, "Exception thrown while reading csv file");
+        } catch (IOException | SQLiteException e) {
+            warrningMsg = "Exception thrown while reading csv file";
```
seadowg: Given this is going to get shown to the user, should we not move it into `strings.xml` so we can translate it? Also, Android Studio probably highlights this and is angry about it. It hates Strings.

seadowg (changes requested):
I was wondering if we could have a test for this. It looks like there isn't currently a test for `FormLoaderTask` and we're kind of treating `FormEntryActivity` as "wild west" non testable territory at the moment. Do you think it'd be too much nastiness to get a test in without a large refactor?

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
grzesiek2010: I was thinking about the same so agree.

grzesiek2010:
> I was wondering if we could have a test for this

I think we should do that in a separate pr we can file an issue.

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -70,6 +71,7 @@
 
     private FormLoaderListener stateListener;
     private String errorMsg;
+    private int warrningMsg;
```
lognaturel: warrning -> warning

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2478,6 +2478,12 @@ public void denied() {
                 activityDisplayed();
 
                 refreshCurrentView();
+
+                if (warningMsg != null) {
+                    String msg = getString(warningMsg);
+                    ToastUtils.showLongToast(msg);
+                    Timber.e(msg);
```
lognaturel: Since this is an issue that has to do with the structure of the given CSV file rather than with the code, I don't think it should be logged as an error.

collect_app/src/main/res/values/strings.xml
```diff
@@ -715,4 +715,5 @@
     <string name="qr_code_not_found">QR Code not found in the selected image</string>
     <string name="not_granted_permission">The activity you are trying to start requires a permission which is not granted.</string>
     <string name="basemap_source_unavailable">Sorry, %s basemaps are not available on this device.  Please choose another basemap in General Settings &gt; Maps.</string>
+    <string name="exception_while_reading_csv">Exception thrown while reading csv file</string>
```
lognaturel: I'd tend to leave this untranslated because the issue is (I believe) uncommon and the message is not very informative. If it's going to be translated, I think it should at least help the user figure out what went wrong.

Perhaps using the localized exception message would be a good option? I'm not sure why we don't do that very often. Maybe because a lot of systems don't actually localize the exception text?

lognaturel (changes requested):

collect_app/src/main/res/values/strings.xml
seadowg: We could just change the text to inform the user that there are unsupported characters in their CSV?

lognaturel: That's not the only way to get an `SQLiteException`, right?

seadowg: Ah true. If there is a way to single that out and report it in a more "human" way though it feels useful. It's a problem the user could potentially fix.

grzesiek2010: Sorry I'm not active here but I just don't have any strong preferences. Since it's a rare case that may be caused by different bugs, maybe we should display an error message (`e.getMessage`) without any translations and that's all?

seadowg: Yeah I'm happy with that as a start! 

It seems a lot of our errors are fairly "technical" in their wording so I'm just interested in ways to improve that. That's probably a larger effort down the road though.

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Right, fixed.

lognaturel (approved):

seadowg (approved):

kkrawczyk123:
I've tried very hard to find a regression but I haven't noticed anything suspicious, so
**Tested with success!**

PR #3431, opened 2019-10-31
Implement likert option functionality in Collect

### LINKED ISSUE #1272: Add support for likert appearance
https://github.com/opendatakit/build/issues/173 is adding a likert appearance to select_one. Enketo has support for this, but Collect doesn't. Since most users of Build use Collect, we should probably add it.

We need to have a discussion about what this should look like because there isn't a lot of horizontal space on most devices.

shouryaj:
Closes #1272

We have written espresso tests to verify functionality. We have also extensively stress tested the likert widget with different forms.

We first started out with using a static XML, but soon realized that this approach would require a lot of XML manipulation from Java. Instead we decided to follow the already existing practices used by the other widgets. 

There should be no regression risks since we simply added an extra widget. 

Form is attached. To test images, edit the form with the name of the image you want to test.
[likert_test.xlsx](https://github.com/opendatakit/collect/files/3821394/likert_test.xlsx)

Link to issue: https://github.com/opendatakit/docs/issues/1135


likert.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0"?><h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><h:head><h:title>Likert</h:title><model><instance><likert id="likert" version="2019101701"><list_widget/><meta><instanceID/></meta></likert></instance><bind nodeset="/likert/list_widget" type="select1"/><bind jr:preload="uid" nodeset="/likert/meta/instanceID" readonly="true()" type="string"/></model></h:head><h:body><select1 appearance="likert" ref="/likert/list_widget"><label>Likert widget</label><hint>This is a normal likert widget with (type = select_one, appearance = likert)</hint><item><label>Yes</label><value>yes</value></item><item><label>No</label><value>no</value></item><item><label>Don't Know</label><value>dk</value></item><item><label>Not Applicable</label><value>na</value></item></select1></h:body></h:html>
```
lognaturel: Isn't used -- remove.

lognaturel:
This is really great, @PieterBenjamin, @estberg, @Joseph-Schafer, Andrew, @shouryaj and others who may have contributed. You've had to ramp up on a ton of context and made a lot of really thoughtful decisions along the way.

I haven't looked at the code yet but wanted to start with a few user experience issues I noticed:
- there are white lines to the left and right of the selections. It looks like the lines were made white to have them "disappear" but the form entry activity background is not quite white. This may only be visible on low-brightness screens. You'll want to use a theme color anyway because it's still white in the dark theme (more at https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines).
- select option text is black in the dark theme.
- the buttons are hard to tap. It looks like the tap target is limited to the radio buttons themselves but it should extend to at least 48x48dp according to [material design standards](https://material.io/design/usability/accessibility.html#layout-typography). I think that like with selects, each label/image should be tappable and select the corresponding option. This might be controversial but I also think it would be ideal for the lines in between the options to be tappable as well such that the left half of the line selects the left option and the right side of the line selects the right option.
- likert selections don't update follow-up questions on a single screen. In your test form, the `cascading_select` group has an appearance of `likert`. However, the `likert` appearance shouldn't be applicable to groups so it's just ignored. I changed it to `field-list` to try multiple linked questions on a screen and that is not working. It should be that when I update Level1, Level2 is immediately updated. I think you just need to make sure that `widgetValueChanged()` is called in the right place.
- there's sometimes a tiny gap in the middle of the line separating radio buttons.
device-2019-11-08-124114
I'm guessing it's because there's an odd integer being divided by 2 and truncating or something like that.

These are relatively minor and again, I want to reiterate how great well thought out this is. The testing is good, you've handled a lot of edge cases, etc.

For the issues with the lines between radio buttons, I'm wondering whether there might be a better way to draw them. I'll look at that area of the code when I have a moment. Also consider bouncing ideas around in Slack. In particular, @seadowg may have some good ideas for you.

seadowg:
On the white lines: maybe I'm missing something (and that's likely - it's morning here) but are you not able to just add margins to the leftmost and rightmost buttons rather than drawing an "invisible" line? This would potentially simplify the drawing code you have - you wouldn't need any of the decision making in `addLine`.

andrewsiew2:
@seadowg We actually added the lines because it simplified things more actually. If we were to add margins they would need to be the same length as the lines anyways. The complication with the margins would be getting it to be the right length with different phone sizes while adding a line with the match parent attribute does it for us.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/LikertTest.java
```diff
@@ -0,0 +1,170 @@
...
+import static androidx.test.espresso.matcher.ViewMatchers.isNotChecked;
+import static androidx.test.espresso.matcher.ViewMatchers.withClassName;
+import static androidx.test.espresso.matcher.ViewMatchers.withId;
+import static androidx.test.espresso.matcher.ViewMatchers.withText;
+import static org.hamcrest.Matchers.endsWith;
+import static org.hamcrest.Matchers.startsWith;
+import static org.odk.collect.android.test.CustomMatchers.withIndex;
+
+public class LikertTest {
+    // having trouble getting correct labels, check the xslx and likert_test.xml file
```
lognaturel: These comments no longer look necessary.

collect_app/src/main/java/org/odk/collect/android/widgets/LikertWidget.java
```diff
@@ -0,0 +1,351 @@
...
+            }
+            if (selectedIndex == -1) {
+                return null;
+            }
+            SelectChoice sc = items.get(selectedIndex);
+            return new SelectOneData(new Selection(sc));
+        }
+    }
+
+    /**
```
lognaturel: This comment comes from the `QuestionWidget` implementation and can be removed or replaced with a comment that describes this specific implementation. Please add `@Override` annotation.

collect_app/src/androidTest/assets/forms/likert_test.xml
```diff
@@ -0,0 +1,419 @@
...
+            <item>
+                <label>2</label>
+                <value>choice_2</value>
+            </item>
+            <item>
+                <label>3</label>
+                <value>choice_3</value>
+            </item>
+        </select1>
+        <group appearance="likert" ref="/likert_test/cascading_select">
```
lognaturel: The appearance on the group should be `field-list`, not `likert`.

lognaturel (changes requested):
A little bit of cleanup pointed out inline! Otherwise, this looks ready for QA to me. 🎉

collect_app/src/main/java/org/odk/collect/android/widgets/LikertWidget.java
estberg: I've removed the redundant comment and added the annotation. Don't think the implementation needs a special comment, as the only difference is the [margin](https://github.com/opendatakit/collect/blob/db40fc948a6ea377cefbad3a96217905a3578941/collect_app/src/main/java/org/odk/collect/android/widgets/LikertWidget.java#L120). @andrewsiew2 if you feel there is an appropriate comment, please add it ☺️

lognaturel: Agreed it doesn't seem necessary!

lognaturel (approved):
Looks ready to QA to me!

kkrawczyk123:
I have tested on Androids 4.2 4.4, 5.1, 6.0, 7.1, 8.1, 9.0 and 10 - and I have noticed just one small issue, when I change the language to RTL the scale is reversed in the proper way but the lines are not, they have gaps, see attached screenshots:
Screenshot_20191118-102933 
Screenshot_20191118-103036

andrewsiew2:
@kkrawczyk123 Is there a way to know if an RTL language has been chosen upon rendering?

grzesiek2010:
`QuestionWidget` which your class extends contains `isRTL()` method you can check how we use it.

shouryaj:
@kkrawczyk123 This issue should be fixed now!

grzesiek2010:
@shouryaj 
could you also rebase the branch against master/fix conflict?

lognaturel:
@grzesiek2010 @seadowg @shouryaj @andrewsiew2 @estberg @PieterBenjamin @Joseph-Schafer My thinking was that it would take a fair amount of effort to rework the history for not much benefit. Luckily there are no conflicts and since the feature is relatively stand-alone, I think squashing down to one commit on merge will provide sufficient context. The unfortunate thing is that several people contributed to this effort and the single commit would be attributed to @shouryaj. I was thinking that we could credit everyone in the commit details.

kkrawczyk123:
**Tested with success!**
verified cases:
- light mode
- dark mode
- screen rotation
- app minimizing
- RTL language light mode
- RTL language dark mode 

PR #3434, opened 2019-11-04
Fixed memory leaks in settings

### LINKED ISSUE #3417: Memory leak : LinearLayout when user opens some page under 'User Access Control' on Android >8.0
Collect master
Android 8.1, 9.0

1. User opens Admin settings
2. User goes to `Main Menu Settings` or 'User Settings' or `Form Entry Settings`
3. `LinearLayout` leaked

grzesiek2010:
Closes #3417 

I was able to reproduce the leaks and confirmed they are no more visible after implementing my changes.

The way we created and managed the toolbar in settings was wrong and I refactored the code:
- moving setting the title to resources instead of doing it in fragments
- removing redundant code
- covering `toolbar` and `root` to [local variables](https://github.com/opendatakit/collect/commit/a2a92d3050dd91202ea1ebc9cb513e587dbb3ddf) what eventually fixed the issue.

It should be confirmed that there are no more leaks like those mentioned in the issue, additionally I refactored the way the toolbar (in settings) is created (especially its title) so please confirm there is no regression.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
@@ -21,6 +21,7 @@ import android.os.Bundle;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.Toolbar;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.utilities.LocaleHelper;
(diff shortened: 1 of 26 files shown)


lognaturel:
Looks right to me but @seadowg is more likely to know if there's anything else that should be considered.

seadowg:
I can't see any problems with the changes. 

One thing that stood out to me is the code that sets the toolbar seems overly complicated. Are we not able to simply do `((AppCompatActivity) getActivity()).setSupportActionBar(toolbar)` from the `Fragment`? Maybe that's something we can mark as a problem.

Also, just out of interest, did you work out what the cause of the memory leak was? I couldn't see where `toolbar`/`root` were being leaked.

seadowg (approved):

grzesiek2010:
> One thing that stood out to me is the code that sets the toolbar seems overly complicated

I simplified it.

> Also, just out of interest, did you work out what the cause of the memory leak was?

I'm not 100% sure so I'm not going to explain it but it was related to [root](https://github.com/opendatakit/collect/pull/3434/files#diff-37ab08a05a62c126f15766bfb4e750efL21) field. Probably it was caused by the fact that we were adding views multiple times not removing them https://github.com/opendatakit/collect/pull/3434/files#diff-37ab08a05a62c126f15766bfb4e750efL61

grzesiek2010:
Please add needs testing label if everything is fine.

mmarciniak90:
**Tested with success**

Tested on Androids: 8.1, 9.0
Verified also on Androids: 7.0, 6.0, 5.1, 4.4, 4.2

PR #3438, opened 2019-11-05
Add tests and fixes

kkrawczyk123:
Add new automation tests and small fixes of an existing one.


collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
```diff
@@ -161,4 +168,35 @@ public T checkIfAreaWithKeyIsDisplayed(String key) {
         onData(PreferenceMatchers.withKey(key)).check(matches(isDisplayed()));
         return (T) this;
     }
+
+    public T checkIfStringIsEnabled(int stringID) {
```
grzesiek2010: You don't need this method, you can use checkIfElementWithKeyIsEnabled which you added below.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
```diff
@@ -161,4 +168,35 @@ public T checkIfAreaWithKeyIsDisplayed(String key) {
...
+        onView(withText(getInstrumentation().getTargetContext().getString(stringID))).check(matches(not(isEnabled())));
+        return (T) this;
+    }
+
+    public T scrollToElementWithKey(String key) {
+        onData(PreferenceMatchers.withKey(key)).perform(ViewActions.scrollTo());
+        return (T) this;
+    }
+
+    public T checkIfElementWithKeyIsEnabled(String key) {
```
grzesiek2010: The name is checkIfElementWithKeyIsEnabled but in fact you check if it is disabled.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
```diff
@@ -161,4 +168,35 @@ public T checkIfAreaWithKeyIsDisplayed(String key) {
         onData(PreferenceMatchers.withKey(key)).check(matches(isDisplayed()));
         return (T) this;
     }
+
+    public T checkIfStringIsEnabled(int stringID) {
+        onView(withText(getInstrumentation().getTargetContext().getString(stringID))).check(matches(not(isEnabled())));
+        return (T) this;
+    }
+
+    public T scrollToElementWithKey(String key) {
```
grzesiek2010, line 171: All those methods that use `PreferenceMatchers` like this one are related to settings, so it would be good to have one parent class called eg. `SettingsPage` where we could move them. Such a class should be extended by classes like `AdminSettingsPage`, `GeneralSettingsPage`, `ServerSettingsPage`, `UserAndDeviceIdentitySettingsPage` then.

kkrawczyk123: I think that it would be better to add it in another pr as it is not specifically connected with added tests.

grzesiek2010: Ok.

grzesiek2010 (approved):

PR #3439, opened 2019-11-06
Add identify-user support

### LINKED ISSUE #3435: Form with `identify-user=true` audit attribute adds identifier to audit log events
If `identify-user` is enabled, any time a person opens a form with this feature enabled, they are prompted with a “Enter a user identifier for the audit log” dialog to identify themselves with a plaintext ID or name.

A non-blank identifier will be required to open the form. All audit log events until the form is closed will be attributed to that identifier. The person filling out the form will be responsible for making sure that they exit all forms before handing their device to someone else for editing.

The full spec for this feature is here: https://docs.google.com/document/d/1lVObIqvM75tCSA5w8Iu3WropJDQmpE3Cap3GWOepTOY

Design for feature here: https://github.com/seadowg/collect-design/tree/master/prototypes/identify-user

#### Acceptance

 - [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
Then I am told I need to identify myself

 - [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And I enter my name
And I click "enter"/"go" on the keyboard
Then I can fill in the form normally

- [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And I click "enter"/"go" on the keyboard without entering anything
Then I remain on the same screen

- [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And I go back/click the cross
Then I return to the main menu

 - [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And I enter my name
And I click "enter"/"go" on the keyboard
And I fill in the form
Then the form's audit log includes my name in all events in the `user` column

- [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And I enter my name with commas or quotes ("Call,u"m"")
And I click "enter"/"go" on the keyboard
And I fill in the form
Then the form's audit log includes my name ("Call,u""m""") in all events in the `user` column

 - [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And rotate the device
Then I'm still prompted to enter my identity

 - [x] Given I've downloaded a form with `identify-user=true`
When I click on the form
And background/resume the app
Then I'm still prompted to enter my identity

- [x] Given I've downloaded a form with `identify-user=false`
When I click on the form
Then I am not prompted to enter my name
And I can fill in the form normally

seadowg:
Closes #3435. This implements the feature described (with designs and specs) in that issue.

A couple of notes on implementation:

* I've reworked a little about the structure of audit code so it follows producer-consumer design. I've kept the `AuditEventLogger` (producer) but it now defines an `AuditEventWriter` interface that's implemented with an `AsyncTask`. The goal here was to seperate out the logic that creates the list of events from the writing of those events to a file.
* I've moved code that just deals with auditing into the `audit` package. This is talked about in the State of the Union.
* Showing the dialog had to happen in the `FormEntryActivity` as it controls loading and rendering the form. The code that makes that happen isn't great but I want to follow up with a bit of a refactor of the Activity itself. I didn't feel including that would be the right move as it would be a lot of noise for a reviewer.

Local testing and submitting the form to Aggregate.

There shouldn't be any changes other than the new feature. Would be good to test form entry in  general to make sure we haven't messed up any existing flows.


grzesiek2010:
1. There is no `ok/navigate forward` button?
Screenshot_2019-11-14-16-57-30

It's strange that I have to open the keyboard and click `Done`.

2. The keyboard moves the toolbar up and it's mixed with the status bar:
Screenshot_2019-11-14-17-01-02

3. This white-gray area won't be well visible when a user is outside in a sunny day (especially the `Identity` text and the cursor)
image

4. I think it's strange that once you open a form the cursor is blinking but the keyboard is hidden. Either the cursor and keyboard should be visible or both should be hidden.

5. If you rotate your device when that new view is visible the form is closed then you can open it again bypassing the view :smile: 

6. We should handle commas, if I use a comma entering my identity the answer will end up in two separate columns:
image

7. I'm able to enter just one space char and go further. We shouldn't allow doing this I guess. The same with emoticons.

8. If you minimize the app (seeing the new view) and open it again the view is skipped (similar case to 5 but not the same). 

9. Dark theme requires improvements:
Screenshot_2019-11-14-20-55-14

grzesiek2010:
The keyboard move the toolbar up and it's mixed with the status bar:
Screenshot_2019-11-14-17-01-02

grzesiek2010:
This white-gray area won't be well visible when a user is outside in a sunny day (especially the `Identity` text and the cursor)
image

grzesiek2010:
I think it's strange that once you open a form the cursor is blinking but the keyboard is hidden. Either the cursor and keyboard should be visible or both should be hidden.

grzesiek2010:
If you rotate your device when that new view is visible the form is closed then you can open it again bypassing the view :smile:

seadowg:
@grzesiek2010 all great finds! Can you stick it into "Request changes"? I'll have a look at fixes.

collect_app/src/main/res/menu/menu_ak.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
grzesiek2010: Why do you need this empty menu?

seadowg:
Just questions on a couple of your points @grzesiek2010 (most of them clearly just need fixed):

> There is no ok/navigate forward button?

Intention was just to have the keyboard "done" take care of this but that's fairly useless without the field focusing (as you point out) - I'll fix that. Do you think we still need a button if we have that? I was following the pattern used by password prompts on Android.

> I'm able to enter just one space char and go further. We shouldn't allow doing this I guess. The same with emoticons.

Great point. The [spec](https://docs.google.com/document/d/1lVObIqvM75tCSA5w8Iu3WropJDQmpE3Cap3GWOepTOY/edit#) states a "non-blank identifier" is required. I don't want to add too much validation in but I'd assume not allowing whitespace only would be ok. Do you have any extra contect on this @lognaturel? If not I can dig through the forum discussion.

seadowg:
I've updated the acceptance in #3435 based on @grzesiek2010's findings

grzesiek2010:
> Intention was just to have the keyboard "done" take care of this but that's fairly useless without the field focusing (as you point out) - I'll fix that. Do you think we still need a button if we have that?

I would prefer to have such a button but it's my personal opinion.

collect_app/src/main/res/menu/menu_ak.xml
seadowg: We don't! That's my mistake.

seadowg:
Ok I think I've fixed everything @grzesiek2010 found. Some annoying stuff around Theming and getting the soft keyboard to show held this up a little. I'm unclear on one path into the form that I'll leave a comment on.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2411,28 +2432,21 @@ public void denied() {
...
+                            startFormEntry(formController, warningMsg);
+                        }
+                    });
                 } else {
                     Intent reqIntent = getIntent();
                     boolean showFirst = reqIntent.getBooleanExtra("start", false);
 
                     if (!showFirst) {
                         if (!allowMovingBackwards) {
+                            // How does this happen?
```
seadowg: How does this happen?

grzesiek2010: In `Admin Settings` -> `Form Entry Settings` we have an option called `Moving backwards`, if you uncheck it you can't navigate back in forms.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -136,19 +140,19 @@ private boolean isDuplicatedIntervalEvent(AuditEvent newAuditEvent) {
...
         if (isAuditEnabled()) {
-            manageSavedEvents();
+            manageSavedEvents(formController);
             writeEvents();
         }
     }
 
     // Filter all events and set final parameters of interval events
-    private void manageSavedEvents() {
+    private void manageSavedEvents(FormController formController) {
```
lognaturel: Why have `FormController` come in to the methods if it's also a field? Seems like it would be bad for those not to be the same.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -272,6 +272,10 @@ boolean isAuditEnabled() {
         return locations;
     }
 
+    public Boolean isIdentityRequired() {
```
lognaturel: use primitive type?

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/FillBlankFormPage.java
```diff
@@ -49,4 +49,9 @@ public FillBlankFormPage checkIsFormSubtextDisplayed() {
         onView(withIndex(withId(R.id.form_subtitle2), 0)).check(matches(isDisplayed()));
         return this;
     }
+
+    public FormEntryPage clickOnForm(String formName) {
+        onView(withText(formName)).perform(click());
```
lognaturel: This may require scrolling, I think, depending on what the form list looks like. If that sounds right, might as well address now.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2478,34 +2450,105 @@ public void denied() {
...
+        });
+
+        identityPromptViewModel.isFormEntryCancelled().observe(this, isFormEntryCancelled -> {
+            if (isFormEntryCancelled) {
+                onBackPressed();
+            }
+        });
+    }
+
+    private void startFormEntry(FormController formController, String warningMsg) {
```
lognaturel, line 2498: Start and resume are identical?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -320,16 +319,16 @@ public void allowSwiping(boolean doSwipe) {
      */
     private boolean locationPermissionsPreviouslyGranted;
 
-    FormEntryViewModel viewModel;
+    private FormEntryViewModel formEntryViewModel;
```
lognaturel: Probably this can be renamed to `BackgroundLocationViewModel`, now.

lognaturel (commented):
This is a great example of small refactors in service of your goals! I really like that you added to the "state of the union" in context.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -269,11 +262,26 @@ boolean isAuditEnabled() {
...
 
-    ArrayList<AuditEvent> getAuditEvents() {
-        return auditEvents;
-    }
-
     List<Location> getLocations() {
         return locations;
     }
+
+    public Boolean isUserRequired() {
```
lognaturel: Use primitive `boolean`?

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -269,11 +262,26 @@ boolean isAuditEnabled() {
...
+
+    public String getUser() {
+        return user;
+    }
+
+    public interface AuditEventWriter {
+
+        void writeEvents(List<AuditEvent> auditEvents);
+
+        Boolean isWriting();
```
lognaturel: Use primitive `boolean`?

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
seadowg: Not sure. This is probably something I left around while reworking. Good catch.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/FillBlankFormPage.java
seadowg: Yeah that's a good point. Will add that in.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah I actually noticed that while looking into accounting for the last pathway into the form (comments above). This will get fixed.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
seadowg: This is my Kotlin self taking hold! I guess in Java there is a minor but definite performance hit though.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah totally agree.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Ah thanks. I'll WIP this PR until I take care of that path.

seadowg: Fixed!

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
lognaturel: For me it's not about the performance hit, it's about the problems nulls can cause. I find it especially bad with `Boolean` because they end up with three states which feels very wrong.

lognaturel (approved):
The code looks good to me and the functionality looks like it's all there thanks to @grzesiek2010's initial review! 🙏 

I feel like I'm being a broken record but the fonts really look really small to me. I would expect this to lead to occasional inaccurate identity entry because it's just hard to see what's being typed in. BUT. I understand that they are default and match where the world is going so I think we should go with it and see what user feedback comes in.

Like @grzesiek2010, I instinctively want to swipe forward and had to think a second about tapping "Done". I would tend to have some kind of explicit "Save" button but I'm ok with waiting for user feedback on that part as well.

lognaturel (approved):
The code looks good to me and the functionality looks like it's all there thanks to @grzesiek2010's initial review! 🙏 

I feel like I'm being a broken record but the fonts really look really small to me. I would expect this to lead to occasional inaccurate identity entry because it's just hard to see what's being typed in. BUT. I understand that they are default and match where the world is going so I think we should go with it and see what user feedback comes in.

Like @grzesiek2010, I instinctively want to swipe forward and had to think a second about tapping "Done". I would tend to have some kind of explicit "Save" button but I'm ok with waiting for user feedback on that part as well.

lognaturel (approved):
The code looks good to me and the functionality looks like it's all there thanks to @grzesiek2010's initial review! 🙏 

I feel like I'm being a broken record but the fonts really look really small to me. I would expect this to lead to occasional inaccurate identity entry because it's just hard to see what's being typed in. BUT. I understand that they are default and match where the world is going so I think we should go with it and see what user feedback comes in.

Like @grzesiek2010, I instinctively want to swipe forward and had to think a second about tapping "Done". I would tend to have some kind of explicit "Save" button but I'm ok with waiting for user feedback on that part as well.

grzesiek2010 (approved):

mmarciniak90:
@seadowg Here is what I found
1. The one whitespace is still accepted to go further - the problem reported by Grzegorz is visible

2. The status bar is not visible on Android 4.4 and 4.2

| Anroid 6.0 | Android 4.4 |
|---|---|
||Screenshot_2019-11-25-14-33-07|

3. Copy-paste functionality covers the field. Some extra space is visible near the cursor pointer and 'frame' for Cup/Copy/Clipboard/Share options. Hope, you can see in on screens:

| | |
|---|----|
|Screenshot_20191125-144400|Screenshot_20191125-144338|
|Screenshot_20191125-145017|Screenshot_20191125-145024(1)|

I also have questions:
* Should the font size changes be applied to the identity page?
* I changed the language to Spanish and I still this page in English. I  understand that translations are not updated but I want to make sure that it is not missed.

mmarciniak90:

seadowg:
@mmarciniak90 thanks for these - I'll have a look at fixing these issues. I talked to a form designer today about the feature and they weren't conncerned about whitespace but given it says "non blank" in the spec I think we should block whitespace only entries. On the copy/pate menu covering the field: I'm not sure we have control over that but I'll have a look.

On your questions:

> Should the font size changes be applied to the identity page?

I'm guessing you mean font size changes in the app rather than at the system level? I'm actually not sure where the distinction lies for what should react to the app's font size change. @lognaturel?

> I changed the language to Spanish and I still this page in English. I understand that translations are not updated but I want to make sure that it is not missed.

Good point. What is our normal process for getting translations in place?

lognaturel:
> where the distinction lies for what should react to the app's font size change

It only affects the form entry activity currently. It feels very weird because it doesn't include the hierarchy/jump view or the save/submit screen. Ideally I would think both would be affected because they feel like they're part of the form filling experience. I think that ideally font changes would go to this screen but I don't currently feel strongly about it given the chaotic current state. It's already weird that its fonts don't match the rest of the app (they match Material Design standards which is certainly better than matching nothing).

> What is our normal process for getting translations in place?

https://github.com/opendatakit/collect#contributing-translations outlines the process. Transifex pulls the `strings.xml` file from master once a day so as soon as this is merged, the strings that need translating will be pulled in. The translations will be updated once on ~Dec 4 for a beta and again before release.

Translations are a great thing to check, @mmarciniak90. I believe that it is now impossible to commit untranslatable strings because the build will break. But maybe what you can also do is quickly verify the [`strings.xml`](https://github.com/opendatakit/collect/pull/3439/files#diff-45784e92e63c347573773596b6912771) file to make sure all strings that you expect are there. I also do this during code review.

collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
```diff
@@ -0,0 +1,21 @@
+package org.odk.collect.android.utilities;
+
+import static java.lang.Character.isWhitespace;
+
+public class StringUtils {
```
lognaturel, line 26: What would you think of using the existing `TextUtils` class? It has some other methods that will eventually come from `String`. (`isBlank` is in Java 11 but not yet in Android).

seadowg: Could we rename `TextUtils` to `StringUtils`? Personally would have made that much more discoverable for me.

lognaturel: Works for me. It's kind of a random grab bag of things currently so I could also see an argument for moving `join` to your new `StringUtils` if you prefer.

lognaturel: (Basically, fine to leave as-is, I just wanted to make sure you'd considered `TextUtils`)

seadowg: I'll rename to `StringUtils` and keep everything there.

seadowg:
@mmarciniak90 **1** and **2** are fixed. 

~~**3** is proving amazingly difficult. I'm going to need to some investigation outside our app (in a demo codebase) to work out what's going on. We're going to use the same full screen dialog pattern for the next feature I work ([Track Changes Reason](https://docs.google.com/document/d/1NyBFhASCOAqOSmz70diQObKpymVE59_MuwjOUPCKlR8/edit#heading=h.e0xux2ykt4mp)) on so I'd vote for us to move on and try and fix after that is done. If you agree I'll make an issue once we merge this is.~~

**EDIT:** Scratch that. Had a sudden brainwave and managed to fix 3. Still not 100% sure I understand how exactly but it's looking good!

mmarciniak90:
@seadowg thanks for taking care of the mentioned problems. I can confirm that they are fixed.

I noticed one more thing.
1. User adds identity
2. User rotates device
3. Filled identity disappears

It would be better not to lose this data during rotation, what do you think?

seadowg:
> It would be better not to lose this data during rotation, what do you think?

That definitely shoudn't be happening. I'll take a look.

seadowg:
@mmarciniak90 fixed! Thanks for keeping me on my toes around screen rotations (this goes for @grzesiek2010 too). It's probably very obvious I've managed to work mostly on apps that lock to portrait in the past 😉

mmarciniak90:
**Tested with success**

Tested on Android: 4.2, 4.4, 5.1, 6.0, 6.0, 8.1, 9.0, 10.0

Verified cases:
* Blank identity
* Identity with whitespace  only
* Identity with chars and whitespace
* Status bar
* Copy-paste functionality
* Username in audit.csv file
* Light/dark mode
* Screen rotation

PR #3443, opened 2019-11-12
Fixed NPE while opening MapsPreferences from GeoPointActivity

### LINKED ISSUE #3442: Crash when user clicks on Offline Layers icon on Geo widget
Collect master, all Android versions

#### Problem description
Crash when the user clicks on Offline Layers icon on Geo widget 

1. User opens map on GeoPoint / GeoTrace / GeoShape
2. User clicks on `Offline Layer` icon
3. Collect crashes
Screenshot from 2019-11-12 12-37-40

#### Expected behavior
Dialog with offline layers appears

grzesiek2010:
Closes #3442 

I confirmed that the crash described in the issue is no longer visible would be enough.

We just need a null check here so there is no other option. We display MapsPreferences clicking on the layers icon but it's a dialog so there is no toolbar like in case when we display the same fragment from settings. So the problem was we tried to call `setTitle()` on a `toolbar` which was null.

It's not risky at all because I just added a null check. Confirming that the crash described in the issue is no longer visible would be enough.

Any form with `GeoPointWidget` like `AllWidgets` form for example.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
@@ -87,7 +87,9 @@ public abstract class CollectAbstractActivity extends AppCompatActivity {
 
     public void initToolbar(CharSequence title) {
         Toolbar toolbar = findViewById(R.id.toolbar);
-        toolbar.setTitle(title);
-        setSupportActionBar(toolbar);
+        if (toolbar != null) {
+            toolbar.setTitle(title);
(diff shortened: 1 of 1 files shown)


seadowg (approved):

mmarciniak90:
**Tested with success**

Verified cases:
* GeoPoint / GeoTrace / GeoShape widgets
* available / unavailable offline layers
* light / dark mode

Verified on Android: 4.2, 4.4, 5.1, 7.0, 8.1, 9.0, 10.0

PR #3447, opened 2019-11-13
Add a new MapActivity that shows saved forms on a map.

### LINKED ISSUE #61: Display form submissions on map
<a href="https://github.com/mitchellsundt"></img></a> **Issue by [mitchellsundt](https://github.com/mitchellsundt)**
_Thursday Jul 09, 2015 at 18:40 GMT_
_Originally opened as https://github.com/opendatakit/opendatakit/issues/511 (2 comment(s))_

---

Originally reported on Google Code with ID 510

```
Sorry if I am posting in the wrong location.

I am interested in whether it would be possible to visualize completed forms in a map
on the mobile device (through google earth or any other app).

Cheers,
Jeremy

```

Reported by `jeremy.r.brammer` on 2012-02-07 22:13:46

zestyping:
Closes #61

See the [design document](https://docs.google.com/document/d/10SkgrhQZyhJ1GjgzteaI3i80Qr-SohETayDGttHXwnY/edit?ts=5dc9f351#heading=h.ab030ba7tzz8) for background, details on the design of this feature, the supporting reasoning, and the implementation plan.

Related issues: #61, https://github.com/opendatakit/opendatakit/issues/511
Forum discussion: [Visualize ... the data collected with ODK on a map](https://forum.opendatakit.org/t/visualize-on-the-fly-in-the-mobile-phone-the-data-collected-with-odk-on-a-map/12881)
Roadmap item: [Collect: Map submissions on device](https://github.com/opendatakit/roadmap/projects/1#card-7129946)

The new form chooser list:

form-chooser-with-map-buttons

The new MapActivity with custom icons:

map-activity-zoom-to-bounds

The user-visible behaviour changes in this PR are:
  - Map buttons are shown in the form list for forms with geometry.
  - The map button launches a MapActivity that shows a marker on the map for each form instance.
  - The map markers are small balloons coloured according to form status.
  - The "new instance" button on the map launches the FormEntryActivity.
  - When a form is completed, the app returns to the MapActivity with the newly added point.
  - Clicking a map marker navigates to viewing/editing the saved form instance.

The affected areas of the codebase are:
  - MapFragment and its implementations
  - The form listing displayed by FormChooserListActivity and FormListAdapter
  - The DiskSyncTask, which processes form definitions on load.
  - The SaveToDiskTask, which saves form instances when the user taps "Save".
  - There are new columns in the forms table and the instances table, which means the schemas have been incremented by 1 and a data migration will be automatically performed to the new schema when the app is upgraded from a previous version.

I manually tested each of the behaviour changes shown above.  In particular:
  - I added a few forms that contained geo questions and confirmed that map buttons appeared for these forms in the form list.
  - I saved several instances of the "geo widgets" form and confirmed that I could see the points as icons on the map.
  - I tried changing the status of some instances by finalizing them and submitting them, and confirmed that the icon colours changed as expected.
  
See the [design document](https://docs.google.com/document/d/10SkgrhQZyhJ1GjgzteaI3i80Qr-SohETayDGttHXwnY/edit?ts=5dc9f351#heading=h.ab030ba7tzz8) for the background and rationale.

Yes: https://github.com/opendatakit/docs/issues/1144


zestyping:
I've made the map markers clickable and customized the icons according to the status of each form instance.  (Feels quite satisfying to tap a blue icon, mark finalized, save, and see it turn purple on the map!)

map-activity-custom-icons

I believe this PR is now feature-complete (as defined by the [design document](https://docs.google.com/document/d/10SkgrhQZyhJ1GjgzteaI3i80Qr-SohETayDGttHXwnY/edit?ts=5dc9f351) and ready for review!

zestyping:
Hmm, looks like I can't directly add labels so I can't tag this `needs review`.

zestyping:
Okay, as a bonus, I added a "zoom to bounds" button because it was easy and we're pretty sure we'll need it. 😄 

map-activity-zoom-to-bounds

lognaturel:
This is fantastic. 

> Feels quite satisfying to tap a blue icon, mark finalized, save, and see it turn purple on the map!

The whole thing is super satisfying! 🎉 🗺 Zoom to bounds is a really nice addition. 💯

I'll be diving into the code now but I have a few UI/UX observations:
- I wonder whether the map icon should be black instead of blue. This would match other buttons that represent actions like the search and sort buttons right above and the media buttons while filling the form. @seadowg is the Material Design / UI consistency Tsar now so would be good to get his UI impressions.
- Changing location doesn't get reflected in map. Tap a point to edit, change the location and come back to the map. The location is still the original one.
- Sent and deleted points look the same as sent points. When deleted, instances that have been sent are kept in the db so there's a record of them but their instance XML is deleted. Tapping on deleted point currently opens a new blank form. It should do nothing? Show a toast with the deleted date?
- Point XPath discovery doesn't seem to work from form download (vs. push to sdcard). Probably doesn't matter because it will get revised after #3452 is merged anyway so this is more of a note not to forget that.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
```diff
@@ -105,7 +114,7 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
     public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         try {
             Timber.i("Downgrading database from version %d to %d", oldVersion, newVersion);
-            moveInstancesTableToVersion5(db);
+            upgradeToVersion5(db);
```
lognaturel: I think this should now be `upgradeToVersion6`

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -258,30 +259,44 @@ protected void extractGeometryValues(FormInstance instance, ContentValues values
...
             }
         } catch (XPathException | XPathSyntaxException e) {
             Timber.w(e, "Could not evaluate geometry XPath %s in instance", geometryXpath);
         }
     }
 
-    protected JSONObject toGeoJson(GeoPointData data) {
-        return new JSONObject();
+    protected JSONObject toGeoJson(GeoPointData data) throws JSONException {
+        // The four (nameless, undocumented) fields of a GeoPointData record are
```
lognaturel: Maybe point to https://opendatakit.github.io/xforms-spec/#data-types? Ok, ok, I'll admit it was only added some months ago after your work on other geo so "undocumented" used to be accurate. 🤦🏻‍♀️

collect_app/src/main/java/org/odk/collect/android/adapters/FormListAdapter.java
```diff
@@ -32,22 +32,17 @@
...
 import timber.log.Timber;
 
-/**
- * Implementation of cursor adapter that displays the version of a form if a form has a version.
- *
- * @author mitchellsundt@gmail.com
- */
-public class VersionHidingCursorAdapter extends SimpleCursorAdapter {
-
+/** An adapter for displaying form definitions in a list. */
```
lognaturel, line 35: Thank you.

lognaturel (commented):
Really not much to say here. The commit history is really straightforward to review, thank you.

zestyping:
I'm travelling to DRC very soon, so @lognaturel is taking this PR forward from here.  Thank you so much @lognaturel !

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
zestyping: Sorry about being snarky here—it wasn't helpful. 😓

lognaturel: Haha, no worries, it was earned. 😄

lognaturel:
You've taken us most of the way to this fantastic feature, @zestyping! Safe travels and may the impact of your work continue to grow. For those who may not know, DRC [has an ebola outbreak](https://www.who.int/emergencies/diseases/ebola/drc-2019) and @zestyping is going to be iterating on and deploying some software to support frontline workers there (including a bit of ODK!).

This is what the icon looks like in black (CC @seadowg):
device-2019-11-18-133244

I have a slight preference for this. @smoyth, @cooperka, since you also reviewed Material Design guidelines when you reworked this screen and have some interest in this feature, please consider weighing in. The immediate question is whether the icon should be filled blue or black but feel free to make other suggestions.

cooperka:
Super cool functionality! Latest screenshot LGTM in terms of Material Design. Nice work all.

collect_app/src/main/res/layout/form_chooser_list_item_map_button.xml
```diff
@@ -0,0 +1,19 @@
...
+        android:focusable="false"
+        android:layout_width="40sp"
+        android:layout_height="40sp"
+        android:insetLeft="0sp"
+        android:insetRight="0sp"
+        android:insetTop="0sp"
+        android:insetBottom="0sp"
+        android:padding="8sp"
+        app:icon="@drawable/ic_map"
+        app:iconTint="@color/blue_500" />
```
seadowg: Outlined buttons should really be using `?attr/colorPrimary` for their text/icons ([specs here](https://material.io/components/buttons/#theming)). This is of course what the button would use if we didn't set a tint. The problem we have right now is that our `colorPrimary` is exactly the same as our `colorSurface` so that wouldn't work. Amazingly the Sketch plugin for Material Design (I've attached the PDF output) works this out and it suggests we should be using a black (I'm guessing it just chooses the darkest version of our `colorPrimary`). I think we probably need to make decisions about our `colorPrimary` in the next release as this is going to get messy if we can't just use defaults from our theme. 

For the moment, could you just use `?attr/colorOnSurface` or `?attr/colorOnPrimary` (they're basically the same). I'll quickly bash out a PR after this with a custom `OutlinedButton` style that all the existing ones in the app can use until we can switch our `colorPrimary` to something useful!
 
[collect-styleguide.pdf](https://github.com/opendatakit/collect/files/3863448/collect-styleguide.pdf)

collect_app/src/main/res/layout/instance_map_layout.xml
```diff
@@ -0,0 +1,86 @@
...
+        android:layout_alignParentRight="true"
+        android:layout_below="@+id/form_title"
+        android:layout_above="@+id/new_instance">
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:orientation="vertical">
+
+            <androidx.appcompat.widget.AppCompatImageButton
```
seadowg, line 57: These should really be [Floating Action Buttons](https://material.io/components/buttons-floating-action-button/#anatomy). The Android implementation is documented [here](https://material.io/develop/android/components/floating-action-button/). Using those should give you the correct colors I believe (`colorOnPrimary` and `colorPrimary`).

collect_app/src/main/res/drawable/ic_instance_marker_finalized.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
seadowg: Are these icons coming from Studio's Vector Assests tool? If so could we leave the names the way they are when they are generated? That way it's easier to work out whether we already have an icon in our resources or not.

seadowg (changes requested):
Just a few comments on Material Design/Theming things. The feature looks really cool though!

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
lognaturel: Addressed by 945bc8f

collect_app/src/main/res/layout/instance_map_layout.xml
lognaturel: Agreed. @zestyping and I talked about this and thought that we'd initially match the other map views and then rethink their look and feel together. Does that seem reasonable?

collect_app/src/main/res/drawable/ic_instance_marker_finalized.xml
lognaturel: Going through the Vector Assets tool gives name `ic_room_black_24dp` no matter the tint. The icon already exists without a tint as `ic_place_black` and is used for warning dialogs related to location.

For this feature, there are 4 colored variants needed as you can see in [`getDrawableIdForStatus`](https://github.com/opendatakit/collect/pull/3447/files#diff-8419b5f9069e6719d3038c7dca542854R211). Is there a way to do that other than introducing 4 resources?

collect_app/src/main/res/layout/form_chooser_list_item_map_button.xml
lognaturel: All that sounds great, thank you!

lognaturel:
To do in this PR:
- [x] Update geometry in db when existing instance is edited (*it turns out that this was working for the general case but could fail if there were multiple versions of a form on a device, as I had*)
- [x] ~Tint sent and deleted points differently than sent points~ ~After trying 0e2e6e5 out, I don't think this is necessary~ Realized that location is too sensitive to keep around after delete so went with 0eb45dd
- [x] Show toast when sent and deleted point is tapped
- [x] Make sent forms viewable but not editable
- [x] Consider deleting `geometryType` and `geometry` when instance is deleted to avoid leaking PII
- [x] Consider pulling geometry out of existing instances so those appear on the map, too (doesn't seem worth the complexity/parse time)

To do after #3452 is merged:
- [x] Ensure feature XPath discovery works from form download
- [x] Use the first feature outside of a repeat rather than the first feature overall

To do later:
- [ ] @seadowg will [add an `OutlinedButton` style](https://github.com/opendatakit/collect/pull/3447#discussion_r347867451)
- [ ] Look into changing `setMarkerIcon` API and using "tintable" drawables instead of having icons of different colors (see https://github.com/opendatakit/collect/pull/3447#discussion_r348426013)

collect_app/src/main/res/drawable/ic_instance_marker_finalized.xml
lognaturel: I'll add that the marker gets set by the `setMarkerIcon` method defined by the `MapFragment` interface. I believe that to use a single black icon and tint it, we'd have to introduce a new method there and make sure hat all three mapping engines (Mapbox, OSM, Google Maps) all support tinting across all API levels. Does that seem worth it, @seadowg?

seadowg: Ah I see because it takes a resource `int` I see. Yeah right now that does feel like a bigger disruption. For the moment can we just have them named like `ic_room_red_24dp`, `ic_room_blue_24dp` etc?

Just a not on tinting though: My impression is that we'd be able to use [`DrawableCompat.wrap`](https://developer.android.com/reference/androidx/core/graphics/drawable/DrawableCompat.html?hl=en#wrap(android.graphics.drawable.Drawable)) to create "tintable" drawables back to at least API 14. I guess that it might not work with map engines for some reason but I think we could probably just test at our min API level.

collect_app/src/main/res/layout/instance_map_layout.xml
seadowg: Yeah that does seem right.

lognaturel:
As far as I know, the behavior is now all as expected. I ended up finding one strange bug where the geometry XPath wasn't necessarily taken from the same version of a form that an instance was built from. I only caught this because I happened to have two versions of the same form on my test device. It's not intuitive that we always have to make sure to use the `form_id`, `version` tuple when considering form definitions. `FormsDao` attempts to nudge in that direction by making the default variant of `getFormsCursor` take a both as arguments but that's not a very strong nudge. 

@opendatakit/testers just a note that if we add more to the forms table or make changes to it, we have to be really careful that we update and use the correct version.

I also realized that deleting an instance should by default delete the geometry from the database because folks use deleting instances as a way to make sure sensitive data isn't on devices and location can be very sensitive. We may want to add an option to keep location for deleted instances but that can come later if there's user demand.

As-is, forms that are already on the device won't get a map, only new ones will. I thought it'd be nice to make existing forms/instances visible but upon reflection I don't think it's worth the complexity. Forms can be big and there can be many instances so the parsing tasks could take a long time.

What I'd like tests for:
- [x] The map button is shown if the form definition has a point question
- [x] The map button is not shown if the form definition does not have a point qusetion
- [x] The "new instance" button starts filling a new instance of the form and that new instance is added to the map
- [x] All instances that have geometry are displayed
- [x] Geometry is deleted for deleted instances (so they aren't displayed)
- [x] The marker color matches the instance status
- [x] Tapping on an unfinalized, unencrypted finalized or unencrypted failure to send instance opens the instance for editing
- [x] Tapping on an encrypted finalized or encrypted failure to send instance shows a toast
- [x] Tapping on a sent instance opens the instance for viewing
- [ ] Modifying the status of an instance on edit changes its marker's color
- [ ] Modifying the position of an instance on edit changes its position on the map

Doing all of these as Espresso tests would add quite a bit of time to running the test suite and I'm not very excited about that. I think that by introducing a `ViewModel` layer for `MapActivity` I can bump several of these to lower levels.

lognaturel:
All of the behavior that I could think of now works as expected. There turned out to be a lot of extra considerations:
- the geometry XPath path needs to be taken from the same form version as the instance is built from
- the instance down migration should work regardless of future fields that are added or deleted
- instances that are deleted are kept in the db but their geometry shouldn't be kept because that's sensitive info
- encrypted forms should not attempt to be opened after finalization (that includes when sent or failed to send)
- if the "Edit Saved Form" button is not visible on the main menu, the project administrator does not expect instances to be editable so they should not be editable through the map, either

Please be on the lookout for other such subtleties I may have missed. I went back and forth about what to do if the "View Sent Form" button was not visible from the main menu. Ultimately I decided that organizations mostly remove it to clear clutter and that it's better to show the submissions that are still in the database in the map.

Regarding code structure, I'm not entirely happy with the view model's responsibilities. I talked to @seadowg briefly about this and he rightfully pointed out that the mapping between `featureId` and `databaseId` is really a presentation layer concern as is everything to do with mapping. I agree with that but the `ViewModel` class also serves the purpose of saving state on orientation change. I don't want to have to rebuild the mapping then so it was expedient to put it in the `ViewModel`. Same with the point list. It's not great but at least there is no saved instance of the `MapFragment`. @seadowg if you have an alternate suggestion, I'm open to it. Depending on how dramatic it is, we may consider coming back to it since we will be building on this feature.

I made some potentially controversial changes to the abstractions related to databases. There's a lot of code currently throughout the app that deals directly with cursors that is hard to read and error-prone. I'd like to see us move towards using domain objects as much as possible and I've made the first small steps in that direction here because it really helped with testability.

I'm overall pretty happy with what I've been able to test. Writing the tests helped me identify some issues so it seems they're doing their job. I'd like to have tests around editing an instance and returning to the map to verify that the edits are reflected but I couldn't get reasonable instrumented tests with the map and I don't know how to write unit tests that involve asserting things about the map markers. Ideas here very welcome.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -149,6 +149,9 @@ the specific language governing permissions and limitations under the License.
         <activity
             android:name=".activities.GeoPolyActivity"
             android:configChanges="orientation" />
+        <activity
+            android:name=".activities.FormMapActivity"
+            android:configChanges="orientation" />
```
seadowg: Do you think we need `configChanges` for this Activity? It would be nice to avoid for new ones seeing as it's discouraged.

obdulia-losantos:
Hi!

I was trying this PR with the following xform definition and I could not see the map icon, is there anything else I need to do?

I'm using and Android 6 (API 23) device.

```xml
<?xml version="1.0"?>
<h:html
    xmlns="http://www.w3.org/2002/xforms"
    xmlns:h="http://www.w3.org/1999/xhtml"
    xmlns:jr="http://openrosa.org/javarosa">
    <h:head>
        <h:title>Geo Point</h:title>
        <model>
            <instance>
                <geo id="geo_point" version="v1">
                    <location />
                    <meta>
                        <instanceID />
                    </meta>
                </geo>
            </instance>
            <bind nodeset="/geo/location" type="geopoint" />
            <bind nodeset="/geo/meta/instanceID" readonly="true()" type="string" jr:preload="uid" />
        </model>
    </h:head>
    <h:body>
        <input appearance="placement-map" ref="location">
            <label>Capture the location.</label>
        </input>
    </h:body>
</h:html>
```

image

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
```diff
@@ -0,0 +1,194 @@
...
+    public int getMappedPointCount() {
+        return points.size();
+    }
+
+    /**
+     * Clears the existing features on the given map and places features for the current form's
+     * instances. If features were added to the map and the map view was not previously initialized,
+     * zooms to display all points.
+     */
+    public void mapUpdateRequested(MapFragment map) {
```
seadowg, line 82: I think the ViewModel is serving as a wrapper for a bunch of logic concerned with manipulating `MapFragment`. I'm not sure why we can't have methods like this inside `FormMapActivity`. I really think the mapping between map "features" and instances needs to live somewhere else - it feels like a detail of the presentation implementation rather than "state". What I do like about the ViewModel is that it protects the `FormMapActivity` from where instances actually come from, the `Form` object and also the logic around which instances should end up on the map. For me, the ViewModel could be more like this:

```java
public class FormMapViewModel extends ViewModel {
    /**
     * The form that is mapped.
     */
    private final Form form;

    private final InstancesRepository instancesRepository;

    public FormMapViewModel(Form form, InstancesRepository instancesRepository) {
        this.instancesRepository = instancesRepository;
        this.form = form;
    }

    public String getFormTitle() {
        return form.getDisplayName();
    }

    public String getFormId() {
        return form.getJrFormId();
    }

    public List<GeoFormInstance> getInstances() {
        List<Instance> instances = instancesRepository.getAllBy(form.getJrFormId());
        List<GeoFormInstance> geoInstances = new ArrayList<>();

        for (Instance instance : instances) {
            if (instance.getGeometry() != null) {
                try {
                    JSONObject geometry = new JSONObject(instance.getGeometry());
                    switch (instance.getGeometryType()) {
                        case "Point":
                            JSONArray coordinates = geometry.getJSONArray("coordinates");
                            // In GeoJSON, longitude comes before latitude.
                            double lon = coordinates.getDouble(0);
                            double lat = coordinates.getDouble(1);
                            geoInstances.add(new GeoFormInstance(
                                    instance.getDatabaseId(),
                                    new Pair<>(lat, lon),
                                    getStatusForInstance(instance)
                            ));
                    }
                } catch (JSONException e) {
                    Timber.w("Invalid JSON in instances table: %s", instance.getGeometry());
                }
            }
        }


        return geoInstances;
    }

    private Status getStatusForInstance(Instance instance) {
        if (instance != null) {
            if (instance.getDeletedDate() != null) {
                return Status.DELETED;
            }

            if ((instance.getStatus().equals(InstanceProviderAPI.STATUS_COMPLETE)
                    || instance.getStatus().equals(InstanceProviderAPI.STATUS_SUBMITTED)
                    || instance.getStatus().equals(InstanceProviderAPI.STATUS_SUBMISSION_FAILED))
                    && !instance.canEditWhenComplete()) {
                return Status.NOT_VIEWABLE;
            } else if (instance.getDatabaseId() != null) {
                if (instance.getStatus().equals(InstanceProviderAPI.STATUS_SUBMITTED)) {
                    return Status.VIEW_ONLY;
                }
                return Status.EDITABLE;
            }
        }

        return Status.UNKNOWN;
    }

    public static class GeoFormInstance {

        private final Long instanceID;
        private final Pair<Double, Double> geometry;
        private final Status status;

        public GeoFormInstance(Long instanceID, Pair<Double, Double> geometry, Status status) {
            this.instanceID = instanceID;
            this.geometry = geometry;
            this.status = status;
        }

        public Long getInstanceID() {
            return instanceID;
        }

        public Pair<Double, Double> getGeometry() {
            return geometry;
        }

        public Status getStatus() {
            return status;
        }
    }

    public enum Status {
        DELETED, NOT_VIEWABLE, VIEW_ONLY, EDITABLE, UNKNOWN
    }
}
```

Here the ViewModel is just responsible for "presenting" its client with the data it needs to render. We don't need to create a new object representation for the `Instance` here but I like it as an option as it lets us remove any of the logic around mapping status from one the Instance cares about to one our Activity cares about. In this setup `mapUpdateRequested` would move to the Activity and would just update the features on the map based on the data it's presented:

```java
private void mapUpdateRequested(MapFragment map) {
   points.clear();
   map.clearFeatures();

   List<FormMapViewModel.GeoFormInstance> instances = viewModel.getInstances();
   totalInstanceCount = instances.size();
   for (FormMapViewModel.GeoFormInstance instance : instances) {
        MapPoint point = new MapPoint(instance.getGeometry().first, instance.getGeometry().second);
        int featureId = map.addMarker(point, false);

        int drawableId = getDrawableIdForStatus(instance.getStatus());
        map.setMarkerIcon(featureId, drawableId);

        instancesByFeatureId.put(featureId, instance);
    }

    if (!viewportInitialized && !points.isEmpty()) {
        mapZoomToBoundingBoxRequested(map);
        viewportInitialized = true;
    }
}
```

I think there is something missing from our `MapFragment` interface here that's making this all a little awkward. Maybe we need a way to give features some kind of "meta identifier" like `setTag` on Android `View`. I'm not 100% sure. 

Does any of this make sense or is it silly?

collect_app/src/main/java/org/odk/collect/android/instances/DatabaseInstancesRepository.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Over time, we should consider redefining the responsibility split between
+ * {@link org.odk.collect.android.provider.InstanceProvider}, {@link InstancesRepository} and
+ * {@link InstancesDao}.
+ */
+public final class DatabaseInstancesRepository implements InstancesRepository {
+    private final InstancesDao dao = new InstancesDao();
+
+    @Override
+    public List<Instance> getAllBy(String formId) {
```
seadowg, line 27: ❤

seadowg (changes requested):

lognaturel:
Thanks so much for the review, @seadowg! Will follow up shortly.

Hi @obdulia-losantos! Are you downloading that form from a server by any chance? One current limitation on this PR is that the form definition must be transferred via `adb` to have its geometry detected. That will be updated soon.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
lognaturel: Ah, shoot, I thought I was on board with this but it has the problem that I highlighted above -- the featureId/dbID map, points and viewportInitialized are lost on orientation change. Do you suggest handling saving those manually? I'd also like unit tests for viewport initialization and that's no longer possible.

I think it's unfortunate that Google called their lifecycle-aware container `ViewModel`. I think it really wants to be more of a kind of controller. The first sentence in its description is `The ViewModel class is designed to store and manage UI-related data in a lifecycle conscious way.` To me, `viewportInitialized`, `points`, is UI-related data.

seadowg: But you're rebuilding the featureId/dbID map every time in `onResume` anyway right? Why does that need to persist between rotations? I'm wondering if my comment above about `configChanges` is relevant here and I'm confused about what happens on rotation - I'm assuming `onResume` is called.

Yeah it's not a great distinction. I'd refer to it as "data the UI needs" and that `points` etc are "state about the view".

obdulia-losantos:
Yes, I didn't know that this feature was only available via adb.

Thanks for letting me know, I'll wait for the update.

collect_app/src/main/AndroidManifest.xml
lognaturel: Good catch, thank you. I have removed it. It does indeed block restart on orientation change which we shouldn't need to do, especially since map center and zoom are explicitly saved in `onSaveInstanceState`.

I'm sorry this whole thing is a bit of a mess lifecycle-wise. I evidently didn't carefully go back and review everything Ping had originally done and got myself in a pretty confused state, apparently. I didn't feel confused at the time. "Unknown confusion" is the worst kind of confusion.

lognaturel:
@obdulia-losantos this should now work with downloaded forms.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
lognaturel: > you're rebuilding the featureId/dbID map every time in onResume anyway right?

Sigh. That's too bad.

lognaturel:
I've merged in the latest changes from `master` and updated the point identification code accordingly. I've also gone down most of the path I understood @seadowg describing at https://github.com/opendatakit/collect/pull/3447#discussion_r350645070 in cfafe8a but am out of time to get the tests working because I need to pack for a flight.

@seadowg is that what you had in mind or did I misunderstand? It feels worse to me because there's more code and concepts to take in, there's an extra pass over the instance list which could be long and none of the interesting bits are testable anymore. Granted, the prior tests used mocks so weren't totally satisfying but they still gave quite a bit of confidence.

We are now very squeezed for time since I was planning to be off Weds-Fri. We want to get this right and there's a lot of behavior here so we need to start QA this week. I'll do my best to get some online time to respond to review comments. @seadowg did you review everything else? If you'd like to take things from here in another branch and play around with structure, please do. My tendency would be to revert cfafe8a, focus on code review for the rest of the PR, and come back to the `ViewModel` structure as we get user feedback and know more about where the feature is going.

seadowg:
@lognaturel yeah I had a look throught the whole thing yesterday so happy for you to revert the refactor commit for the moment. I'll do a last pass now on your other changes.

I do like the direction the refactor has gone in. I hear you on it feeling less testable but that should be resolvable via testing at the activity level or by pulling out a use case or maybe even a custom view that we can test. Google have been creating at making sure all their example show ViewModels solving everything but for cases like this it still feels like the majority of the logic is purely display based. Maybe it'd be a fun thing to attempt a pairing session on when we're both back next week!

seadowg:
@lognaturel looks good. Happy to approve after the revert!

lognaturel:
> it still feels like the majority of the logic is purely display based. Maybe it'd be a fun thing to attempt a pairing session on when we're both back next week!

I agree with you conceptually. I'd love to do a pairing session on it! Maybe we can look both at this and reducing duplication/using domain objects for the initial form discovery code as described [here](https://github.com/opendatakit/collect/pull/3452#discussion_r347256054). That would be good clean up to do while it's still fresh and some of what I did here with `Instance` and `Form` will be relevant there.

seadowg (approved):

mmarciniak90:
@lognaturel after today's testing, I have a few thoughts

1. Leaving form always moves to the main view but with these changes, we have different behavior. 
I just want to confirm that we want to change this standard like described on PR, that we want to leave it in that way.
2. I also wonder should this map view be **available by default** to all users. 
Maybe we should think about additional option in settings which turns it on. 
3. **Leaks** appear more frequently
4. User needs to be precise to click on the icon. 
Few times when I wanted to open map, the form opened because I didn't aim at the icon
5. You can say that I'm picky :smile: but the map icon is not centered on all Android versions.
I see that icon is not in the center on Android 9.0, 6.0, 4.4

Screenshot_20191128-143609

lognaturel:
> Leaving form always moves to the main view but with these changes, we have different behavior.

Yes, this is important because users want to be able to perform their data collection in the context of the map view. That is, they may use the map to navigate and to evaluate their progress.

> I also wonder should this map view be available by default to all users.

Can you elaborate on your concerns? Do you think it’s distracting, possibly reveals information, something else? Our reasoning is that it’s broadly useful but also clearly a secondary action so not disruptive. We do have to be careful not to allow users to do things they couldn’t do before, though. For that reason, if Edit Saved Form is made unavailable from the admin menu, instances can’t be edited. Instances that are deleted won’t be mapped. I went back and forth on what to do when the View Sent Form button wasn’t available. Ultimately I decided that it’s reasonable to still make sent instances visible on the map because those instances are still on the device. If the issue is data sensitivity, it’s not a bad thing for users to more clearly see that the data is still on the device so that they can turn on delete after send, for example. 

I will look into 3-5, thank you!

mmarciniak90:
@lognaturel I continued testing and here are results:

> Can you elaborate on your concerns?

You dispelled my doubts :smile: 

--- 
**Cases verified with success:**

* Map buttons are shown in the form list for forms with **geopoint** question  :heavy_check_mark:
* The map button launches a MapActivity that shows a **marker for each form instance** :heavy_check_mark:
* The markers are **small balloons colored** according to form status (blue, purple, red, green) :heavy_check_mark:
* The **"new instance" button** on the map launches the FormEntryActivity :heavy_check_mark:
* When a form is completed, the app **returns to the MapActivity** with the newly added point :heavy_check_mark:
* **Clicking a map marker** navigates to viewing/editing the saved form instance :heavy_check_mark:
* If **Edit Saved Form** is made unavailable from the admin menu, instances can’t be edited :heavy_check_mark:
* If the instance is sent and **deleted** (auto-delete or manually) point disappears from the map  :heavy_check_mark:

---
**Problems with removing response**

Use case:
1. User fills a few forms and adds points. 
2. User opens the new map view
3. User sees a correct number of points shown on map on new map view.
4. User clicks on point on map and removes response for geopoint widget
5. User saves changes. User is moved to map view
6. User sees that removed point is still marked and number of points shown on map is not updated

Expected result
* Points wich were removed are not visible on map
* Number of points does not count removed responses
---

Use case 2:
1. User fills a few forms and adds points.
2. User takes look at the new map and leaves it
3. User goes to saved form and removes response for geopoint widget
4. User takes a look at the new map
5. User sees that removed point is still marked and number of points shown on map is not updated

Expected result
* Points wich were removed are not visible on map
* Number of points does not count removed responses



---
**Problem with moving backward option**

Use case 3:
1. Moving backward is unselected
2. User fills a few forms and adds points.
3. User opens a new map view and clicks on point
4. User is moved to empty hierarchy list with a possibility to open General Settings

Expected result:
* Points are not clickable if moving backward is unselected

|Moving backward is unselected| Edit Saved Form is unselected|
|---|---|
|Moving-backward|edit|

mmarciniak90:

lognaturel:
Thank you so much for the thorough review, @mmarciniak90!

> You dispelled my doubts 😄

Glad to hear it! I do think it's a very valid question and if you come up with any other concerns, please share them.

Todo:
- [ ] Investigate leaks
- [x] Make sure the tap surface extends beyond the icon
- [ ] Icon is not horizontally centered in button on Android 9.0, 6.0, 4.4
- [x] Removing point response should remove map point
- [x] Moving backwards setting should not result in blank hierarchy - this turned out to be a bug introduced in 0a6bb3d. Now if an instance is opened for editing and moving backwards is disabled, the user should be jumped directly to the last question that was viewed. It should also be able to open for viewing just like if moving backwards is enabled. Does that behavior look right to you, @mmarciniak90? You said the points should not be clickable. I can see an argument for that -- the edit saved form button is generally not available when moving backwards is disabled -- but I don't think that's the behavior that users would want or expect. As long as users can't jump around and edit, I think it's ok. It will be important to verify this with the identify user feature.

The two bugs you found should be fixed. 🙏 I haven't made much progress on the other three smaller issues you pointed out. Given that those three are relatively small, I think we could file them as separate issues? Alternately, we can see what I can get done during my Monday.

For the tap surface, I think that adding an [`OutlinedButton` style](https://github.com/opendatakit/collect/pull/3447#discussion_r347867451) will likely address that. Does that sound right, @seadowg? Or maybe the entire vertical space below the button should be for the secondary action?

Another quick thing -- both forms and instance databases have been added to. If you haven't already, please check downgrading.

seadowg:
@lognaturel I'm guessing the problem is that the whole list item is clickable and that surrounds the button? It looks like the style of the button is all good (`Outlined.Icon` is correct). What might be a good idea is to use a layout similar to `AudioViewImageTextLabel` - the right hand buttons live in their own layout (with padding) that is not part of the tappable area for selects etc. The trick is to have the left hand content expand to fill the space if the layout with the buttons (in this case just the map button) is gone so that the whole thing remains clickable when the buttons don't need to be there.

lognaturel:
@obdulia-losantos It'd be great to get your feedback on this initial version of the feature! [This forum post](https://forum.opendatakit.org/t/visualize-on-the-fly-in-the-mobile-phone-the-data-collected-with-odk-on-a-map/12881/19) has a screen capture as well as information on how to download a beta with the functionality.

lognaturel:
9cf5a01 makes all space above and below the map outlined button as well as 16dp to the left and to the right tappable. I'm not proud of the approach but I looked into negative margins and they seem to be tolerated (see e.g. https://stackoverflow.com/a/10673572/137744).

I could not reproduce the horizontal center issue after that change. Perhaps it has been addressed?

The leak that is now happening frequently is the following which I can't figure out:
```
In org.odk.collect.android:v1.25.0-beta.0-78-g9cf5a01b9:3590.
* androidx.lifecycle.ReportFragment has leaked:
* LocationManager$ListenerTransport.!(this$0)!
* ↳ LocationManager.!(mContext)!
* ↳ ContextImpl.!(mAutofillClient)!
* ↳ FormMapActivity.mFragments
* ↳ FragmentController.mHost
* ↳ Activity$HostCallbacks.mFragmentManager
* ↳ FragmentManagerImpl.mAdded
* ↳ ArrayList.elementData
* ↳ array Object[].[0]
* ↳ ReportFragment
```

mmarciniak90:
@lognaturel I focused on upgrade more today and I checked more cases than before.
I did not notice the crash and the update was successful.

I noticed one thing that can be good to improve:
**Use case:**
1. The old version is installed, eg. 1.24
2. Few forms with Geopoint widget are available
3. User upgrade to a version with these changes
4. Forms available before upgrade do not have a new MapActivity. Only forms downloaded after upgrade have new MapActivity. 

**Expected result:**
Forms available before upgrade have new MapActivity after the upgrade. 
User does not need to remove and download again forms to use the new functionality.

lognaturel:
> Forms available before upgrade do not have a new MapActivity. Only forms downloaded after upgrade have new MapActivity.

I agree that this is really annoying. Unfortunately, I think trying to parse existing forms and instances to add the functionality is too risky. This is because form definitions can take a long time to parse depending on their length and the capacity of the device used. There could also potentially be several form definitions. That means we'd have to be sure to provide user feedback that this process is ongoing. We'd probably also want to let the user continue filling out forms. There are then lots of possible failure states to handle and I think it's cleaner to say that only forms downloaded with version 1.25 or later will be mapped. I'm disappointed because it is a big limitation but I don't see a good alternative.

smoyte:
Is this (need to re-parse forms) something that is likely to come up again in the future? What about adding a manual "re-parse forms" action in the settings. It wouldn't be as smooth as just doing it automatically but for folks who don't have the option to easily re-download their forms, it could help. We could potentially/eventually pair this with a "new release notes" modal that lets users know about new features when they are released. Might that be a good practice anyway?

lognaturel:
> Is this (need to re-parse forms) something that is likely to come up again in the future?

That's a good question and concept, @smoyth. I don't see any immediate other need for parsing existing forms and instances in this way but I can imagine that we'll want to similarly cache other things from the form definition in the future. Do you see it as a must-have for this initial release of the map feature or do you think it's something we could add on if we get feedback that it's essential?

> We could potentially/eventually pair this with a "new release notes" modal that lets users know about new features when they are released. Might that be a good practice anyway?

This has come up at various points and I'm not really sure how to reason about it because often the people who need to know about the new features are form designers or trainers rather than end users. Even in the case of a feature like this, it's a bit hard to know what to say because it's highly dependent on the structure of forms that will be put on the device.

smoyte:
Definitely not a must-have for my purposes. I don't have a good sense of what other users might be needing...

mmarciniak90:
**Tested with success**

Verified on Android: 10.0, 9.0, 8.1, 7.0, 6.0, 5.1, 4.4, 4.2


Final test cases:
* upgrade
* Map buttons are shown in the form list for forms with **geopoint** question
* Map buttons are NOT shown in the form list for forms with GeoPoint in **repeat section**
* Area near the map icon is clickable 
* The markers are **small balloons colored** according to form status (blue, purple, red, green)
* The map button launches a MapActivity that shows a **marker for each form instance**
* The **"new instance" button** on the map launches the FormEntryActivity
* When a form is completed, the app **returns to the MapActivity** with the newly added point
* When response for GeoPoint is removed, point disappears from the map
* **Clicking a map marker** navigates to viewing/editing the saved form instance 
* If **Edit Saved Form** is made unavailable from the admin menu, instances can’t be edited 
* If the instance is sent and **deleted** (auto-delete or manually) point disappears from the map
* If **Moving backward** is unselected from the admin menu, instances can’t be edited 

PR #3449, opened 2019-11-15
Provide user-friendly error message when CSV for search() or external itemsets is not on device

### LINKED ISSUE #3235: Provide user-friendly error message when CSV for search()/pulldata is not on device
The error messages shown when a CSV attachment meant to be used with search()/pulldata are not friendly to users. Moreover, they don't provide any real useful information for developers. See https://github.com/opendatakit/collect/issues/3230 for an example.

Ideally the error message would explicitly name the CSV file that was missing. I'm not sure whether that information is available where the error is thrown, though, so something like "Something went wrong when trying to access the columns [column names here] from a CSV file. Was the CSV file attached?" might be enough. Note that presumably this error could have come from something going wrong in the database creation process so we can't say for sure that it was missing.

grzesiek2010:
Closes #3235 

I tested the attached forms and added an automated test.

We already handle such a case in case of external itemsets so I decided to use the same code and display the same message.

It's not a risky change it's just related with two widgets for which I added sample forms below. Testing them would be enough.

[selectOneExternal.zip](https://github.com/opendatakit/collect/files/3851183/selectOneExternal.zip)
[search.zip](https://github.com/opendatakit/collect/files/3851184/search.zip)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
+++ b/collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
@@ -208,7 +208,10 @@ public final class ExternalDataUtil {
             }
             return returnedChoices;
         } catch (Exception e) {
-            String fileName = xpathfuncexpr.args[0].eval(null, null) + ".csv";
+            String fileName = String.valueOf(xpathfuncexpr.args[0].eval(null, null));
+            if (!fileName.endsWith(".csv")) {
+                fileName = fileName + ".csv";
+            }
             FormController formController = Collect.getInstance().getFormController();
             String filePath = fileName;
             if (formController != null) {
--- a/collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
+++ b/collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
@@ -33,7 +33,10 @@ import org.odk.collect.android.R;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.exception.ExternalDataException;
 import org.odk.collect.android.external.handler.ExternalDataHandlerSearch;
+import org.odk.collect.android.logic.FormController;
 
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
```diff
@@ -205,6 +208,16 @@ public static XPathFuncExpr getSearchXPathExpression(String appearance) {
             }
             return returnedChoices;
         } catch (Exception e) {
+            String fileName = xpathfuncexpr.args[0].eval(null, null) + ".csv";
```
lognaturel: At this point the filename hasn't been normalized by `ExternalDataHandlerBase.normalize` so it could include a `.csv` suffix, right? I think you need to check whether it's already there or not before adding it.

lognaturel (commented):
It looks like this doesn't do anything for a form with `pulldata`.

I was imagining a dialog but I think this is probably better because it's always visible and matches what happens when an image/audio/video file is missing. 👍

collect_app/src/main/java/org/odk/collect/android/external/ExternalDataUtil.java
grzesiek2010: Right, I thought in search you can use just a file name without .csv but looks like both versions work:
`search('fruits')` and `search('fruits.csv')`
Fixed.

grzesiek2010:
> It looks like this doesn't do anything for a form with pulldata.

`pulldata` is used in calculations so it's difficult then to tell if there are no elements because it should return no elements or if the file is missing.

lognaturel:
> `pulldata` is used in calculations so it's difficult then to tell if there are no elements because it should return no elements or if the file is missing.

Except currently it crashes if no dataset exists, right? What I had in mind was having a special case for when the SQL exception reason is that there is no table of the requested name. That case would show a nice error instead of a big ugly one with SQL info.

Looking at your solution again, are you sure it works every time the form is loaded? The csv files get renamed on import with something like an `.imported` suffix. I suspect that what you have only works the first time the form is loaded.

grzesiek2010:
When it comes to `search` func the message is always displayed even if at first you have your csv file and then you remove those files that are created from your origins csv.
In case of `pulldata` there is no crash just nothing is displayed (no elements).

lognaturel:
> the message is always displayed even if at first you have your csv file and then you remove those files that are created from your origins csv

I don't really understand this but I see now from looking at the code again that the check for the original CSV is only relevant if there's an exception in the first place so that makes sense.

> In case of pulldata there is no crash just nothing is displayed (no elements).

Ah, that's right, that's what we need to improve. It's different so we can file it as a separate issue. The problem is that if there's user input involved it's hard to tell whether there was no match or the file wasn't there.

lognaturel (approved):

grzesiek2010:
> The problem is that if there's user input involved it's hard to tell whether there was no match or the file wasn't there.

Exactly, moreover in forms like the one attached to [How to pull data from CSV](https://xlsform.org/en/#how-to-pull-data-from-csv) the result is calculated and used in a `note` question so in Collect we don't even now if it's a result of `pulldata()`, so probably it could require changes in javarosa.

kkrawczyk123:
**Tested with success!**

PR #1672, opened 2017-11-30
Only call FileUtils.parseXML once when downloading a form. 

dcbriccetti:
Only call FileUtils.parseXML once when downloading a form.  (Improve @shobhitagarwal1612’s https://github.com/opendatakit/collect/commit/495f8ae4b2685667e512c8c50b24edf13fdada4f.)

Simplify the bad submission URL checking code by moving it inline and not using an exception to communicate a failure.

Closes #1671

#### What has been done to verify that this works as intended?
I tested with two forms, one of which has an invalid submission URL, and causes an appropriate error message to appear in Collect.

#### Why is this the best possible solution? Were any other approaches considered?
It parses once, and uses the results twice. What could be better?

#### Are there any risks to merging this code? If so, what are they?
None that I see.

#### Do we need any specific form for testing your changes? If so, please attach one.
Any form with an invalid submission URL, like this one, can be used to test the path where a submission URL is invalid. Any form can be used to test the other path, but a really large one will show nearly a halving of the download/parsing speed.

[form-with-invalid-submission-url.xml.txt](https://github.com/opendatakit/collect/files/1516894/form-with-invalid-submission-url.xml.txt)

PR #3452, opened 2019-11-15
Build a FormDef when doing initial form discovery

### LINKED ISSUE #3448: Build a FormDef when doing initial form discovery rather than doing a custom parse of the XML
`FormDownloader`/`DiskSyncTask` use `FileUtils.parseXML` to extract metadata from new form definitions. This is redundant with work that JavaRosa already does and makes it hard to use more of the form definition during that discovery. For example, #61 will involve identifying the first geo question in the form definition if it exists and doing this is much easier with a `FormDef` because question type is made available on the `bind`.

One additional benefit of building a `FormDef` on discovery is that we could build the cache file then. That would ensure there is no difference between first form open parsing directly from XML and subsequent form opens reading from cache.

lognaturel:
Closes #3448

I added some tests for the method I changed before changing it to ensure that all behavior stayed the same. I also manually verified the change on a real device to get a feel for the performance difference. For a large form like [nigeria_wards_internal.xml.txt](https://github.com/opendatakit/collect/files/3852792/nigeria_wards_internal.xml.txt) it does take noticeably longer on my Galaxy Tab A when directly compared: about 3 seconds with this change vs. 2 without. I think this is ok because it's still not a huge amount of time, it only happens once, and it sets us up to do really nice things like start building the cache file immediately.

There's a lot of potential for refactor, reuse and performance improvement in this general area so I initially was going for a much more invasive change. I decided that I should instead make this as low-risk as possible so it's easy to verify and @zestyping can pick it up for #61 which is what initially prompted the change. We can then come back and make the other changes.

I think it'd be nice to have the version and form id directly from the `FormDef` which would be adding a couple of getters in JavaRosa. It's not essential to do now, though.

There should be no visible behavior change. However, there will be a slight performance difference because a full `FormDef` is now being built at form discovery time (from disk or download).

There also used to be a logged warning when `uiVersion` was used which is no longer emitted. This was part of a really old form spec and was logging-only so I don't think it will be missed.

Any form will do. It would be good to verify one with a submission URI, for example one that goes to Google Sheets.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
@@ -25,10 +25,8 @@ import android.os.Build;
 import android.os.Environment;
 
 import org.apache.commons.io.IOUtils;
-import org.javarosa.xform.parse.XFormParser;
-import org.kxml2.kdom.Document;
-import org.kxml2.kdom.Element;
-import org.kxml2.kdom.Node;
+import org.javarosa.core.model.FormDef;
+import org.javarosa.xform.util.XFormUtils;
 import org.odk.collect.android.R;
 import org.odk.collect.android.application.Collect;
 
@@ -38,8 +36,6 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.net.FileNameMap;
 import java.net.URLConnection;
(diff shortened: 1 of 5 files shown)


zestyping (approved):
This looks great.  It's straightforward, and things end up nicer and cleaner!

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -261,113 +257,37 @@ private static String actualCopy(File sourceFile, File destFile) {
...
-        if (title != null) {
-            fields.put(TITLE, XFormParser.getXMLText(title, true));
-        }
+    /**
+     * Given a form definition file, return a map containing form metadata. The form ID is required
+     * by the specification and will always be included. Title and version are optionally included.
+     * If the form definition contains a submission block, any or all of submission URI, base 64 RSA
+     * public key, auto-delete and auto-send may be included.
+     */
+    public static HashMap<String, String> getMetadataFromFormDefinition(File formDefinitionXml) {
```
seadowg, line 266: I feel like this might be a nice opportunity to move the meta data representation to a data class rather than a Map.  What do you think?

collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
```diff
@@ -35,4 +39,79 @@ public void mediaDirNameIsCorrect() {
         assertEquals(expected, FileUtils.constructMediaPath("sample-file.123"));
         assertEquals(expected, FileUtils.constructMediaPath("sample-file.docx"));
     }
+
+    @Test public void identifiesFormMetadata_formSimpleForm() throws IOException {
```
seadowg: Could this be `getMetadataFromFormDefinition_withoutSubmission_returnsMetaDataFields()`?

collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
```diff
@@ -35,4 +39,79 @@ public void mediaDirNameIsCorrect() {
...
+
+        HashMap<String, String> metadataFromFormDefinition = FileUtils.getMetadataFromFormDefinition(temp);
+
+        assertThat(metadataFromFormDefinition.get(FileUtils.TITLE), is("My Survey"));
+        assertThat(metadataFromFormDefinition.get(FileUtils.FORMID), is("mysurvey"));
+        assertThat(metadataFromFormDefinition.get(FileUtils.VERSION), is(nullValue()));
+        assertThat(metadataFromFormDefinition.get(FileUtils.BASE64_RSA_PUBLIC_KEY), is(nullValue()));
+    }
+
+    @Test public void identifiesFormMetadata_forFormWithSubmission() throws IOException {
```
seadowg: Could this be `getMetadataFromFormDefinition_withSubmission_returnsMetaDataFields()`?

seadowg (changes requested):
Just a couple of comments around test naming.

This definitely looks a lot better!

collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
lognaturel: Thanks!

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
lognaturel: I agree. This is part of what I ended up backing out of. I considered using the existing `Form` class which is billed as a data transfer object but not used very much. I think that could be appropriate but I'd like to discuss where the code to go from an XML form definition and/or JavaRosa FormDef to one of those should go (e.g. https://github.com/opendatakit/collect/commit/decf3925e7fa28791f5524b66aaad5f252e860fa#diff-824505d34be688f35cf8665de92654e3R245). Some other kind of adapter? Once I started going down that path, I quickly ran into weird inconsistencies between `DiskSyncTask` and `FormDownloader` which both call the current `getMetadataFromFormDefinition`. I think most of the work they do to validate the form properties is not needed but it's different and hard to disentangle. I'm also concerned about things like https://github.com/opendatakit/collect/commit/decf3925e7fa28791f5524b66aaad5f252e860fa#diff-0ceefebf9b2ba619c3af7866f81294fbR285 where it's not entirely clear whether the media path needs to be separately saved (`DiskSyncTask` doesn't do this). I'm worried that we'd end up going back and forth a bunch in review and QA. If you think it's important to do now, let's talk through the high-level strategy before I dive further.

lognaturel: To be clear, there's some urgency around this because it's going to get used for some behavior needed in #3447. On top of that, deeper changes would likely conflict with other open PRs. So what I might suggest is that we go forward with this for now, discuss the refactor strategy, and then I can start playing with it with the goal of getting something mergeable in v1.26. What do you think?

seadowg: Ah cool. I'm happy with an issue then and we can stick it in 1.26 👍. That lets us have a proper dig into it.

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
lognaturel: Filed issue at #3461.

kkrawczyk123:
@lognaturel while testing I have noticed https://github.com/opendatakit/collect/issues/3465 that occurs on master too, so I reported it separately. I am not sure if further testing makes sense until this issue will be fixed and pr rebased to master, what do you think?

lognaturel:
Good catch, @kkrawczyk123. ~That seems like the kind of thing that was introduced previously, I think, right?~ The issue says it's a problem in v1.24.

It seems like you're at least able to confirm that the URI in the form is correctly saved which is what's important for this PR. If possible, it'd be great to verify the rest of the behavior here because some work in #3447 depends on it. Thanks!

kkrawczyk123:
@lognaturel I have noticed an issue. For the form with invalid submission URL there should be toast displayed and that kind of form should not display on Fill Blank Form list. On the pr I am able to open that form and even fill the submission. The form that I have used: 
[form-with-invalid-submission-url.xml.txt](https://github.com/opendatakit/collect/files/3885927/form-with-invalid-submission-url.xml.txt)
And the screenshot of the expected results form master:
Screenshot_20191125-091353

lognaturel:
Thank you, @kkrawczyk123! Do you know where this form came from? This issue was a real puzzle for me because the validation code hasn't changed. The problem turns out to be that the submission block in the form definition specifies a value for the `id` attribute:

`<submission id="submitTest" method="post" action="this is an invalid URL"/>`

This is a JavaRosa feature that is supposed to make it possible to submit a specific subset of a form. However, that feature is not implemented either in Collect or in any of the form builders. This form is especially odd because there is no element with id `submitTest`. This all means that Collect is now doing what it should be doing -- since the submission is for a specific element and there is no support for that, it just ignores it. If you take out the `id` attribute, you should get the correct behavior that you expected.

I think the right thing to do is to modify the test form to remove that `id`. I looked briefly at adding an Espresso test to verify this behavior but unfortunately since the test form discovery is not done with `DiskSyncTask` we can't easily do that for now.

kkrawczyk123:
@lognaturel The form comes from https://github.com/opendatakit/collect/pull/1672
I have removed that `submission id="submitTest"` part from the form and test it on the pr and master - the behavior is compliant - the form is not displayed on Fill Blank Form list in both cases but the information toast is not visible in both cases too. Should we open another issue for that? What do you think? My impression is that it might be misleading for users what is wrong with the form and why it is not visible on the form list.

kkrawczyk123:
I can't see anything suspicious. 

lognaturel:
> @lognaturel The form comes from #1672

Strange! Well, I'm glad we figured it out and fixed the form.

> but the information toast is not visible in both cases

I can't reproduce. It worked for me when I tried it yesterday and I just tried it again and the toast is shown. Can you reset Collect and try again, please?

PR #3454, opened 2019-11-16
test-only changes to fix failing unit tests on Windows

breakbusyloop:
No change to user-facing code, just fixes to unit tests


lognaturel:
Because this is a unit test, it's running in the context of the host machine rather than Android and thus getting filepaths structured for the host machine. It all adds up. Given how long this test has been in the suite, I'm really surprised no one has reported the issue before. 🤔 

Thanks, @OpenDataNerd!

lognaturel (approved):

PR #3455, opened 2019-11-18
Fix `AudioVideoImageTextLabel` spacing

seadowg:
Work towards #3382. I think we should wait for answers to the SO question mentioned below before closing.

This adds margins to the Select One and Select Multiple widgets while maintaining edge to edge tappability. To match these margins have also been added to question labels and group names. Some things to keep in mind about this implementation:

* It only adds margins to the selects for API 21+. I have an [Stack Overflow question](https://stackoverflow.com/questions/58913567/how-to-add-padding-before-checkbox-radiobutton-across-api-levels) open to try and get to an improved version that works across all API levels.
* All the non select widgets still show their contents edge to edge. Personally I haven't seen any where this looks much worse than in 1.24 (the existing margin was only a few pixels) but if anyone can find any particularly egregious examples then we can fix those. My thinking is that each Widget should make it's own decision about the margins it displays in its "answer view". Personally, I'd vote for this work to happen over time (rather than blocking a release) with typography revisals.
* I've prioritised getting this in early so we can test and get it to beta. There are a few things I'm like to follow up with:
    * Define `QuestionWidget` layout with XML rather than procedurally
    * Revise typography for all standard parts of `QuestionWidget`
    * Define `ODKView` layout with XML rather than procedurally
    * Revise typography for `ODKView` elements
    * Revise spacing and typography for end screen
    * Rework audio/image/text label layout
    * Add an issue for revising each individual widget's typography and spacing

Ran through a lot of our "All Widgets" test form to check everything looks ok, tests locally etc

There shouldn't be any behavioural changes. As described above there will be slight cosmetic differences. I think for testing we should try and find any widgets where the lack of margins will be a real problem for a release.


lognaturel:
Selects for API 21+ look really good and are SO easy to hit. Very lovely. I think the left margin makes it look much better.

> My thinking is that each Widget should make its own decision about the margins it displays in its "answer view". Personally, I'd vote for this work to happen over time (rather than blocking a release) with typography revisals.

That sounds fine to me. I'd like to request that before v1.25 goes out (but not necessarily the beta), the text widgets get some margin. They're so common in forms and they look a little silly to me going edge to edge rather than being aligned with the question and hint text. I'm with you in thinking that the rest can wait. Let's see what @opendatakit/testers think since they're the best at keeping us honest. 😊

I think this means you can mark as not draft and we can go to QA with this piece?

kkrawczyk123:
**Tested with success!**
I have gone through All widgets, Brids and other forms with media in both light and dark mode. I have also checked RTL languages - I haven't noticed anything suspicious. 

lognaturel (approved):

PR #3457, opened 2019-11-18
Upgrade to mapbox-android-plugin-annotation-v9:0.8.0.

zestyping:
Upgrading to the newer version provides us some new ways to keep the layers in the right stacking order, so we can simplify how this is done.

See background discussion here: https://github.com/mapbox/mapbox-gl-native/issues/15262

@lognaturel or anyone—please feel free to take this PR forward while I'm away—thanks!

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -220,7 +220,7 @@ dependencies {
     implementation "com.crashlytics.sdk.android:crashlytics:2.10.1"
 
     implementation 'com.mapbox.mapboxsdk:mapbox-android-sdk:9.0.0'
-    implementation 'com.mapbox.mapboxsdk:mapbox-android-plugin-annotation-v9:0.7.0'
+    implementation 'com.mapbox.mapboxsdk:mapbox-android-plugin-annotation-v9:0.8.0'
 
     implementation("com.google.guava:guava:28.1-android")
 
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -220,7 +220,7 @@ dependencies {
     implementation "com.crashlytics.sdk.android:crashlytics:2.10.1"
 
     implementation 'com.mapbox.mapboxsdk:mapbox-android-sdk:9.0.0'
-    implementation 'com.mapbox.mapboxsdk:mapbox-android-plugin-annotation-v7:0.6.0'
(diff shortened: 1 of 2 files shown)


lognaturel:
I've rebased agains master and force pushed so that now the layers dialog can be launched without crashing. Unfortunately, changing layers still results in the the overlay hiding everything else (#3515). As far as I can tell, that's what is left to address here.

lognaturel:
`applyConfig` gets called when the layers are changed via the dialog and that implementation at https://github.com/opendatakit/collect/pull/3457/files#diff-fdd039fc806c853bc4ba175e0ab623edR224 fundamentally doesn't make sense. It sets the style to a brand new style with just the basemap and then adds the reference layer on top. The problem is not that the reference layer is on top of other layers, it's that those other layers are destroyed.

seadowg (changes requested):
Just sticking this to Request Changes based on @lognaturel's comments above

zestyping:
Ah!  My bad here.  When the map is reconstituted in `applyConfig()`, the new map doesn't get a symbol layer and line layer added to it!  I've fixed that in the latest commit.

This is not the tidiest fix (there's a little bit of code duplication), but I've tested it and it appears to work as expected.

lognaturel:
I'm surprised that this change works! It seems like it creates new, empty managers that wouldn't have any contents. When I try it out, symbols disappear. It looks like lines remain which I think is because the new `LineManager` is built off the old, unchanged style. But this is bad because the line is orphaned. That is, if new points are added, they're not connected to the existing line.

This change means that changing the reference layer implies rebuilding the entire style. Do you think this is necessary or desired? If so, then I wonder whether things should be restructured to remove the placeholder layer. That is, if the style always needs to be rebuilt, then the layers can just be added in order.

lognaturel:
This doesn't fix the bug in #3515 but it doesn't make things worse either. Let's merge now so we have fresh dependencies at the beginning of the release cycle and treat #3515 separately.

seadowg (approved):

mmarciniak90:
**Tested with success**

I didn't notice regression

Verified cases:
* GeoShape, GeoTrace, Geopoint
* Tree input method
* Mapbox map styles
* Offline layers
* Buttons on map view
* Removed response
* Display form submissions on map
* light/dark mode

@kkrawczyk123 now it's your turn :smiley:

kkrawczyk123:
Reporting no regression too!

PR #3458, opened 2019-11-18
Add imports for FormsColumns and InstanceColumns in every class that uses them.

zestyping:
This is purely code cleanup with no visible behaviour changes.


lognaturel:
I've brought this up to date. I think it's a fine change that makes things easier to read. It's going to be hard to get in while development is ongoing because there are bound to be conflicts. Right before regression testing would be a good time to merge. I think it's very low risk but it would be good for someone else to read through and verify the changes.

seadowg (approved):

PR #3459, opened 2019-11-18
Add convenience methods to get content URIs for form and instance items.

zestyping:
There are a variety of ways that these URIs are constructed, some a bit clunky.  This PR provides a `getItemUri` method so there's one concise way to do it.

This is purely code cleanup with no behaviour changes.


lognaturel:
This has accumulated a lot of conflicts so let's close for now.

I think that adding a static method doesn't really help us from a testing or design perspective. I think a lot of these calls will be eliminated as we move more towards using model objects and corresponding repositories (https://github.com/opendatakit/collect/pull/3447/commits/b8deb55b738882b8b275eea926e2a07461d7283c#diff-9fd4ad299862fe660384d73e32be2929R10 and https://github.com/opendatakit/collect/pull/3447/commits/36c2a0b1d263718c1ba335b5a6d75077517c00e6). Let's try to move in that direction and see where it takes us.

PR #3460, opened 2019-11-18
Prepare dbs for upgrading

grzesiek2010:
I added automated tests and tested downgrading manually.

We agreed in #3444 that in order to take care of our users and backwards compatibility we should release a version with onDowngrade() methods that would be able to handle path changes first, then we can migrate them.

In order to test this pr please:
1. Install the app with these changes
2. Open and save the attached form.
3. Copy `instances.db` and `itemsets.db` files from `odk/metadata` to your desktop
4. Open them using a browser like https://sqlitebrowser.org/
5. Bump database versions and change paths to start with `../` instead of `/storage/emulated/0/odk/`
6. Paste the edited databases to `odk/metadata`
7. Test the saved forms.

That would be enough when it comes to testing.

Please test a form with external itemsets like:
[select_one_external.zip](https://github.com/opendatakit/collect/files/3861179/select_one_external.zip)

No.


collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
```diff
@@ -156,6 +161,27 @@ private void upgradeToVersion4(SQLiteDatabase db) {
...
     }
 
+    private void downgradeToVersion5(SQLiteDatabase db) {
+        Cursor cursor = db.query(INSTANCES_TABLE_NAME, null, null, null, null, null, null);
+        if (cursor != null && cursor.moveToFirst()) {
+            List<Instance> instances = new InstancesDao().getInstancesFromCursor(cursor);
+            for (Instance instance : instances) {
+                if (instance.getInstanceFilePath().startsWith("../")) {
+                    String where = INSTANCE_FILE_PATH + "=?";
+                    String[] whereArgs = {instance.getInstanceFilePath()};
```
lognaturel: Can we use instance ID and index it? I can't believe there isn't already an index for it.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
```diff
@@ -43,14 +47,13 @@ public void restoreRealDb() {
         FileUtils.copyFile(new File(DATABASE_PATH + TEMPORARY_EXTENSION), new File(DATABASE_PATH));
     }
 
-    @Parameterized.Parameters(name = "{0}")
+    @Parameterized.Parameters(name = "{1}")
     public static Iterable<Object[]> data() {
         return Arrays.asList(new Object[][] {
-                {"Downgrading from version with extra column drops that column", "instances_v7000_added_fakeColumn.db"},
```
lognaturel, line 49: Please add the 7000 tables and tests back in. They're there to guarantee that downgrade with added and removed columns are always tested no matter the current target database version.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
```diff
@@ -43,14 +47,13 @@ public void restoreRealDb() {
...
-                {"Downgrading from version with missing column adds that column", "instances_v7000_removed_jrVersion.db"},
+                {DOWNGRADE, "Downgrading from 5+ to version 5 should updated database paths", "instances_v6.db"},
```
lognaturel: How about making this a separate test case? The other assertions aren't relevant to it and it's exercising particular behavior that will need to be removed in a few versions when downgrade doesn't touch the paths anymore.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
```diff
@@ -156,6 +161,27 @@ private void upgradeToVersion4(SQLiteDatabase db) {
         }
     }
 
+    private void downgradeToVersion5(SQLiteDatabase db) {
+        Cursor cursor = db.query(INSTANCES_TABLE_NAME, null, null, null, null, null, null);
+        if (cursor != null && cursor.moveToFirst()) {
+            List<Instance> instances = new InstancesDao().getInstancesFromCursor(cursor);
+            for (Instance instance : instances) {
+                if (instance.getInstanceFilePath().startsWith("../")) {
```
lognaturel: Can you build a `File` and use `!.isAbsolute`? That better captures the semantics of what you're trying to do.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
```diff
@@ -156,6 +161,27 @@ private void upgradeToVersion4(SQLiteDatabase db) {
...
+                    String where = INSTANCE_FILE_PATH + "=?";
+                    String[] whereArgs = {instance.getInstanceFilePath()};
+
+                    ContentValues values = new ContentValues();
+                    values.put(INSTANCE_FILE_PATH, Collect.ODK_ROOT + instance.getInstanceFilePath().substring(2));
```
lognaturel: Once you have a `File` object as described above, can you use https://developer.android.com/reference/java/io/File#getCanonicalPath()?

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
```diff
@@ -107,7 +111,8 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
     public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         try {
             Timber.i("Downgrading database from version %d to %d", oldVersion, newVersion);
-            moveInstancesTableToVersion5(db);
+
+            downgradeToVersion5(db);
```
lognaturel, line 117: The idea is for this migration to always work no matter what columns may be added or removed in the future. That's why there are the v7000 tests. This down migration should always end up with the correct columns for v5.

I had to rearrange this for introducing a v6 and you can pick up most of that change from https://github.com/opendatakit/collect/pull/3447/commits/945bc8fdab815687a83f4e2f7e2bc1f0aa19b666. Let's merge this first and then I'll rebase and fix conflicts over there.

lognaturel (changes requested):
I didn't answer the question at https://github.com/opendatakit/collect/pull/3444#issuecomment-555014658 about what the relative path prefixes should be because I wanted to think about it but I'm still really not sure.

This builds paths that start with `../` which makes sense given that the database files themselves are located at `<odk root>/metadata/`. That seems like a reasonable approach but I'm not feeling very confident about it. There's something that makes me uncomfortable about having a `../` in there but I can't put my finger on it.

My initial thought was to only store the unique part of the path so something like `All widgets_2019-07-29_18-19-08/All widgets_2019-07-29_18-19-08.xml` and let Collect (or whatever downstream app is using the db if someone copies it off the sdcard) prefix it as it wants.

@seadowg, any thoughts on what the paths should be relative to? Any best practices we should know of?

lognaturel:
Wait a minute. Does this make sense? The end goal is to have files that won't be rooted at `/sdcard/` anymore so they'll still be inaccessible even when going through this down migration, no? Is there any benefit to switching to relative paths before switching the root? This is hard to reason about. It would help to have an exhaustive list of the changes that need to be made, the order that they need to be in and what happens when upgrading and downgrading between versions.

grzesiek2010:
> Wait a minute. Does this make sense? The end goal is to have files that won't be rooted at /sdcard/ anymore so they'll still be inaccessible even when going through this down migration, no?

Yes but we wanted to migrate paths first and then in the future migrate files. Let's assume that we release:
- 1.25 with this pr
- 1.26 which would use only relative paths
- 1.27 which would use scoped storage (files won't be stored on sdcard)

downgrading from 1.26->1.25 will work thanks to this pr
downgrading from 1.27->1.26 - only new files added in 1.27 won't be available but we don't have to remove sdcard/odk dir after migrating file we can leave it so if someone install 1.27 and run into a big bug then they will be possible to downgrade and use files added by previous versions.
We can also prepare a doc how to copy newly added forms from the internal storage to sdcard in case of downgrading.

seadowg:
@lognaturel I haven't been as involved in this stuff so potentially lacking context but yeah it doesn't feel right to have paths like `../` in the database. For me, Collect should mostly be in control of constructing the path to a file. 

An example would be like storing an image in S3 in a web app. If I just store the filename the code gets to dictate where that file lives. That feels easier to change as I don't need to migrate a bunch of data in the database if I change services or structure.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
grzesiek2010: Yes, done.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
grzesiek2010: Yes, done.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
grzesiek2010: Yes, done.

collect_app/src/main/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelper.java
grzesiek2010: Ahhh right because it's called `moveInstancesTableToVersion5` i missed that it's a universal way for the future. Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
grzesiek2010: My bad, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
grzesiek2010: Done.

grzesiek2010:
@lognaturel 

My idea I tried to explain during the meeting is to copy files to sd card in addition to what I've already done in this pr. So onDowngrade() would do:

1. Copy files from odk directory on internal storage to sd card
2. Convert relative paths to absolute paths (what this pr does)

The full scenario would be:
1. We release v1.25 with this pr + changes to copy files to sd card back
2. We release v1.26 with scoped storage (files from sd card are moved to internal storage), the directory on sd card is cleared.

downgrade v1.26->v1.25 the app will copy files from internal storage to sd card back and convert database paths.

Does that sound right?

grzesiek2010:
@lognaturel @seadowg 
can we make a final decision what to do with it?
We have to start using scope storage, that's for sure.
We need to migrate paths we store in databases, that's for sure as well.

The decision we need to make is whether we need to provide backward compatibility or it's not thta important.
My first opinion was yeah it would be good to do that but we would need to implement `onDowngrade()` methods to change paths again during downgrading (what I did in this pr) and also try to copy files from our app dir to sd card.
It will never be a perfect solution because we can do that now and ensure downgrading to v1.26 will work but we can't fix previous versions.
So the question is whether it's worth the effort at all.
I know we have a lot of time till August (when bumping `targetSdkVersion` will be required) but time flies it's been almost a month since I created this pr and we haven't made any decision.

seadowg:
Ach sorry @grzesiek2010 this totally skipped my mind. So by dropping "backwards compatibility" you mean that someone wouldn't be able to downgrade a device from 1.26 to 1.25? I might have this wrong in my head.

lognaturel:
The question for me has been more about *when* we put in the down migration. That is, do we put it in one version before the scoped storage change just to have a path out or do we try to get it in earlier (e.g. v1.26) so that users don't have to think about what version to downgrade to.

But now that I think about this again, another issue came to mind. Along with the scoped storage change, all data will be deleted on app uninstall. Downgrading will require manually installing an APK. Does this APK install lead to the previously-installed (newer) version being uninstalled and thus its data being deleted? If so, there's no point in having a down migration.

Another option we do have is documenting how to move the `/sdcard/odk/` folder to the new location. If folks have to install an APK anyway to downgrade, they'd likely be able to do this.

grzesiek2010:
> Downgrading will require manually installing an APK. Does this APK install lead to the previously-installed (newer) version being uninstalled and thus its data being deleted?

You are right I forgot about it because on a daily basis I work with Android Studio and using it I can downgrade without any problems but yeahh if it's an apk file you need to uninstall the first version.

Ok we have one answer we don't need to handle copying files onDowngrad() (from the app dir on internal storage to sd card).

> So by dropping "backwards compatibility" you mean that someone wouldn't be able to downgrade a device from 1.26 to 1.25?

downgrading will be possible but files won't be available because as @lognaturel pointed out they will be deleted (because you need to uninstall the existed version first) and new version will use relative paths so if you downgrade you will en up wit a database files with relative paths and old code not able to use them.Of course we can handle the second case (what I did in this pr) but taking into account we can't copy files back to sd card it doesn't make much sense.

lognaturel:
> if it's an apk file you need to uninstall the first version.

It is in fact possible to install without removing the files by doing e.g. `adb install -r ODK-Collect-v1.23.3.apk`. See http://adbcommand.com/adbshell/pm for documentation.

But someone would probably need documentation to find that out and the documentation we provide might as well tell them to copy files for safekeeping and then put them back in `/sdcard` if they need to downgrade.

I'm starting to feel like this almost warrants a major version bump to indicate that there isn't really any going back without files being cleared. I hope not too many users routinely downgrade production devices.

As far as this PR goes, I now think we do nothing with the files or database until we absolutely have to make the switch to scoped storage. Does that sound right?

grzesiek2010:
>  I now think we do nothing with the files or database until we absolutely have to make the switch to scoped storage. Does that sound right?

it sound right if we don't want to provide any facilities with downgrading (downgrading databases, copying files back to sd card), because otherwise we need to do that earlier (in previous versions).
Now I'm more in favor of just migrating database paths and that's all.

seadowg:
@lognaturel agreed with the major version bump. It does feel like it's good to create a little drama so people know to the files are now tied to their app install rather than just sitting on device.

grzesiek2010:
Ok so we can close this pr and get #3444 ready (because we can use relative paths even now).
Then we need to warn our users about upcoming changes and switch to scope storage. Am I right?

lognaturel:
> because we can use relative paths even now

You mean updating the code to handle either relative or absolute paths in the db for now but not actually changing the database, right? That'd basically be #3444 without the db changes. We have determined that the db changes won't help because the db files will be removed on uninstall anyway once we move to the new storage location. Yes, as far as I can tell, that sounds right. Hopefully we've thought of most things now. 😫

grzesiek2010:
> You mean updating the code to handle either relative or absolute paths in the db for now but not actually changing the database, right?

I'm not sure now, I thought you prefer having the code that deals just with relative paths and updating existing paths in databases, to avoid complex code with if/else statements (checking if it's a relative or absolute path). Did you change your mind?

lognaturel:
I see this. I think need to do the full analysis with all the information we now have so that hopefully we won't need more back and forth. Hoping I can do that next week.

lognaturel:
There have been some exploration of approaches to getting the scoped storage migration started at https://github.com/opendatakit/collect/pull/3444, https://github.com/opendatakit/collect/pull/3460. I want to summarize at a high-level and hopefully figure out a direction.

> The only thing we need is add android:installLocation="auto" to the Manifest file because by default apps are not movable.

I read more into [app install location](https://developer.android.com/guide/topics/data/install-location) and I now think it's not relevant to us. I thought that moving the app would also move its data files but it does not. I think we can ignore this.

Some other questions that have come up for me:
- [ ] Is there really a benefit to using `Context#getExternalFilesDir` over [`Context#getFilesDir`](https://developer.android.com/reference/android/content/Context.html#getFilesDir())? Is it really the case that files from `getFilesDir` can't be accessed from adb? The docs state that the differences are that (1) `getExternalFilesDir` may not always be available since it could be removed by the user. This seems scary and undesirable. (2) that other apps can write to the directory if they have the `Manifest.permission.WRITE_EXTERNAL_STORAGE` permission. This seems useful but not a must-have, I think.

The #1 priority is that users can't lose saved and filled forms when they upgrade to the scoped storage version. This is non-negotiable. To achieve this, we need to move everything from `/sdcard/odk/` to `Context#getExternalFilesDir`/odk. This needs to happen BEFORE we are forced to switch to scoped storage to be in the Play Store because at that point we won't be able to access `/sdcard/odk/` from Collect. While we'd like to delay this change because it will be inconvenient to users, **we can't delay it too much because we then risk not being able to do it at all**.

"Moving everything" is tricky because it involves both form files and database files that refer to those form files. Currently the references in the database use absolute paths. If those paths were relative, then we could move the whole directory on upgrade and everything should just work. This is where the idea of changing paths in the databases to be relative came from. However, there are two problems:
- downgrading from versions with relative paths in the db to versions with absolute paths in the db would lead to broken behavior because old code can't handle relative paths.
- users don't have to install every version so in the version where we move the directory, we'd also need to first make sure that all paths in the databases are relative. This means that changing paths in a prior version is not very useful.

This leads me to think that for the upgrade path, we need to do the following all in the same new version:
- change paths in the database to be relative
- move all files currently in `/sdcard/odk/` to `Context#getExternalFilesDir`/odk
- use `Context#getExternalFilesDir` to root paths in the database

Sadly I no longer see any way to break this up. I also realize that we probably need to design a user experience around this because we will need to block the UI while the file copy happens and there could potentially be large media files involved.

@grzesiek2010's exploration has focused around what to do if a user wants to downgrade from the Collect version that requires scoped storage version. Unfortunately, [we've realized](https://github.com/opendatakit/collect/pull/3460#issuecomment-564747636) that because one of the side-effects of using `Context#getExternalFilesDir` is that those files get deleted on uninstall (without using the `-r` flag), there's not much point to having a downgrade path. This makes it especially important to get the user messaging right. We've talked about pairing this change with a major version bump and visual refresh. I still think that's probably the way to go but we need to be very careful because if we make overly dramatic or controversial changes, users will want to downgrade.

@grzesiek2010 can you please review this analysis carefully and see if I've overlooked anything?

PR #3464, opened 2019-11-21
Add margins back into Text Widgets

### LINKED ISSUE #3382: Revise `AudioVideoImageTextLabel` spacing and layout

seadowg:
Closes #3382.

Adds horizonatl margins to `StringWidget` that was removed by #3413 

Ran through the "All Widgets" form to check everything looked ok.

This is a bit of a hacky solution. There is some more work to do towards #3382 to detail some larger changes around widget layouts but wanted to get the margins fixed and add a simple way of doing the same for any other widgets we realize need the margins back in.

Should just effect the the Text Widgets but good to check everything looks as expected.


collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
```diff
@@ -45,7 +55,7 @@ protected StringWidget(Context context, QuestionDetails questionDetails, boolean
 
     protected void setUpLayout() {
         setDisplayValueFromModel();
-        addAnswerView(answerText);
+        addAnswerView(answerText, 12);
```
lognaturel: Seems `12` should be a named value?

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
```diff
@@ -114,4 +124,66 @@ public void setDisplayValueFromModel() {
             Selection.setSelection(answerText.getText(), answerText.getText().toString().length());
         }
     }
+
+    private EditText getAnswerEditText(boolean readOnly, FormEntryPrompt prompt) {
```
lognaturel, line 141: Huh. How was that over there?! Mysteries abound.

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
seadowg: Yeah it's just temporary value - we'll hopefully rework the layout so we don't have old school weird padding on the `EditText` (we're making it look like `16dp` padding). But yeah naming it/commenting it would have been a nice thing to do that I will do now 😇

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
seadowg: Haha I was confused about this. Had to run all the tests just to convince myself moving it wasn't a bad idea.

lognaturel (approved):

mmarciniak90:
@seadowg is this PR closes #3382? #3382 looks like connected to other problems.
Sorry, I'm a bit confused

I see that StringWidget and number widgets look better. 

> Should just effect the Text Widgets but good to check everything looks as expected.

What plan we have for other widgets? Eg. the text field in selects widgets with autocomplete or buttons are without margin.

| pr | v 1.24.1|
|---|---|
|Screenshot_20191129-163222|Screenshot_20191129-164025|
|Screenshot_20191129-163236|Screenshot_20191129-164040|
|Screenshot_20191129-163204|Screenshot_20191129-164012|
|Screenshot_20191129-163912|Screenshot_20191129-164053|
|Screenshot_20191129-163917|Screenshot_20191129-164106|

|1.24.1||
|---|---|
|1.25 beta ||
|pr/3464||

seadowg:
@mmarciniak90 sorry I should have linked to #3455 where there was a discussion around it just being the text widgets we needed to fix for this to be releaseable. I didn't want to close #3382 until we had that clean up done. In the issue (#3382) you'll see the plan is to start revising the widgets one by one (probably for 1.26). The goal here would be to improve the usability and look/feel of all the widgets. 

Totally get that for some widgets the current state does feel like a visual regression however. Are there any you feel are too bad to go out with 1.25? If so I think we should fix in this PR.

mmarciniak90:
> Are there any you feel are too bad to go out with 1.25? If so I think we should fix in this PR.

This is a question to @lognaturel more than to me. 
Using these widgets does not change so that is good. Visual aspects are important but correct operation is of greater value

lognaturel:
@seadowg, I somehow didn't notice that vertical spacing was changed not just for selects but across the board. In particular, it looks like question and hint text has more vertical margin/padding which isn't good. I can't remember -- was this an explicit decision in #3413? An inevitability given how the spacing was done? I think we should really try to get that back down. That's the part that makes https://forum.opendatakit.org/t/odk-collect-v1-25-beta-preview-of-increased-spacing-for-selects/23439/2 look a lot worse since the questions have multiple paragraphs. There's no tapping involved in the question label/hint so the spacing only serves visual separation purposes.

Aside from that, I agree with @mmarciniak90 that if it's not a big deal to add the same spacing for the edittext used for the autocomplete/search appearance on selects, that would make things look more consistent.

lognaturel:
Hmm, I went back to look at the QA screenshots at https://github.com/opendatakit/collect/pull/3413 (very helpful, @mmarciniak90), and it doesn't look like vertical spacing was added there. So it may have come in somewhere else?

seadowg:
@lognaturel @mmarciniak90 sounds good. I can get those margins in there. 

As per our discussion on Slack we'll look at adding a quick fix for the vertical margins as part of this work and then look at a larger change in #3483.

mmarciniak90:

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
```diff
@@ -40,12 +59,16 @@ protected StringWidget(Context context, QuestionDetails questionDetails, boolean
...
+        setUpLayout(context);
     }
 
-    protected void setUpLayout() {
+    protected void setUpLayout(Context context) {
         setDisplayValueFromModel();
-        addAnswerView(answerText);
+
+        Resources resources = context.getResources();
+        int marginStandard = dpFromPx(context, resources.getDimension(R.dimen.margin_standard));
```
lognaturel: Perhaps using `getDimensionPixelSize` would make it more obvious that pixels are returned. I remembered you rightfully complained about this method in Slack.

lognaturel (approved):

lognaturel:
I took the vertical spacing on at https://github.com/opendatakit/collect/pull/3492 because I wanted to really think through it and make sure we had time for review before freeze. I think it makes sense to treat the two separately. Please fix conflict and mark as `needs testing`, @seadowg!

seadowg:
@lognaturel thanks so much for taking that on 🙏

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
seadowg: That is at least more obvious!

mmarciniak90:
Margin is visible for widgets:
- String
- String number
- Integer
- Integer with thousands separators
- Decimal
- Select one with search appearance
- Select one with autocomplete appearance
- Select multi with autocomplete appearance

Margin is NOT visible for widgets:
- Ex string when a value is added manually
- Ex integer when a value is added manually
- Ex decimal when a value is added manually
- Bearing when a value is added manually

@seadowg it's hard for me to define on which widgets I should see margin, that's why I created the list above. Let me know if my insights on changes correspond to your intentions. If yes, we can merge it.

Verified on Android: 10.0, 9.0, 8.1, 7.0, 6.0, 5.1, 4.4, 4.2

seadowg:
@mmarciniak90 yup had only intended to change the ones listed. If you feel the others need margins in for release I'm happy to change. Otherwise we'll review them for next release!

mmarciniak90:
@seadowg I think that all the mentioned widgets need margin but in my opinion, they can wait for the next version. 
I'm going to create an issue for this.
I hope that @lognaturel agrees with me. I do not want to delay code freeze because this is not a big deal.

@opendatakit-bot label "behavior verify"

mmarciniak90:

lognaturel:
Agree 100%, @mmarciniak90, thank you!

PR #3473, opened 2019-11-26
Support case insensitive header lookups (#3068)

### LINKED ISSUE #3068: Request headers are treated as case sensitive
Collect v1.21.2, Android v8.0.0, Asus Zenfone 3 Zoom (ZE553KL)

#### Problem description
The [RFC2616 section 4.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) dictates that request / response headers are supposed to be case insensitive.

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

But from what I could tell, it seems the way collect currently works is by putting the headers into a `HashMap<String, String>`, which seems to have the effect of making the headers case sensitive.

In my specific case, I was trying to raise the content length limit by sending an "X-OpenRosa-Accept-Content-Length" header in the HEAD response, but found it didn't work, even though I could see from the source code below that it did check for it's presence and set the `contentLength` accordingly:
https://github.com/opendatakit/collect/blob/8f65795c613a7759abfff0ba557c261fe3c5fa8e/collect_app/src/main/java/org/odk/collect/android/upload/InstanceServerUploader.java#L100-L107

These `responseHeaders` I believe came from here:
https://github.com/opendatakit/collect/blob/8f7a3e6abb9ac18ee3ce8a1eb81dfa8e78b00aa6/collect_app/src/main/java/org/odk/collect/android/http/HttpClientConnection.java#L241-L262

Which justs put the result of `head.getName()` in the `HashMap`.

Technically, it is stated [in the Form Submission API docs](https://docs.opendatakit.org/openrosa-form-submission/#content) "Successful server responses MUST include an X-OpenRosa-Accept-Content-Length header in addition to the required OpenRosa Request response headers.", so I guess that may be interpreted as the burden of sending the correct header being with the server, but since we're talking HTTP requests here, I believe it makes sense that the collect app doesn't depend on that casing.

Also, that problem exists in other places, like for example here:
https://github.com/opendatakit/collect/blob/8f65795c613a7759abfff0ba557c261fe3c5fa8e/collect_app/src/main/java/org/odk/collect/android/upload/InstanceServerUploader.java#L123

it will fail to see a header "Location" if it isn't written exactly like that, even though the Form Submission API states only "If the server is RESTful, the server MAY return an HTTP URI (using the standard HTTP Location header) where the form can be found." (I believe "standard HTTP Location header" meaning that it follows the HTTP standard and so that would be case insensitive).

1. Setup an OpenRosa compliant server except that it sends headers as all uppercase, all downcase or any other combination besides the ones specifically used in the code
2. Configure the server to send a custom content length limit through the X-OpenRosa-Accept-Content-Length header
3. Send a form with a body larger than 10mb
4. See it being sent as multiple requests to the server

#### Expected behavior
Collect respecting the limit on the X-OpenRosa-Accept-Content-Length header no matter in which case it is sent.

#### Other information 
From what I could see, a seemingly common way to make a case insensitive Map is by using a TreeMap with the String.CASE_INSENSITIVE_ORDER comparator being passed as argument to the constructor.

(Kinda like this: `new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER)`).

But my Java is so rusty that I have no confidence in affirming that that will be an easy or good solution.

Also, I confess that I didn't check if the underlying HttpClient will keep the original formatting of headers or will normalize them in some way. If it normalizes them, then this bug will actually be about the "X-OpenRosa-Accept-Content-Length" probably being normalized to "X-Openrosa-Accept-Content-Length" (which is still missed in the check).

breakbusyloop:
Closes #3068

gradlew testDebug
gradlew connectedAndroidTest
NOTE: I have NOT performed any manual testing yet against a server configured for lower case. I have only tested for regressions up to this point, and I will separately attempt to test the new use case manually. This change includes unit test coverage and a mock-based integration test.

We needed an implementation of a Headers data structure that supports case insensitive lookups. We already have such functionality in the OkHttp implementation. Rather than use that directly (and propagate a data structure that would couple ourselves to a third party), I wrapped it in a simple interface so we can change the implementation at a later time without affecting our downstream code.

Users with OpenRosa servers that are configured in non-standard ways (e.g. lower-case only headers) will be able to work with large files (the specific issue reported) and may experience other functionality improvements. All other users should not be affected, provided there were no latent bugs within the ODK code that previously (incorrectly) depended on failed header lookups due to case differences.

#### Does this change require updates to documentation? If so, please file an issue [here]
No change.

- [ X ] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [ n/a ] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [ n/a ] verified that any new UI elements use theme colors. [UI Components Style guidelines]


collect_app/src/main/java/org/odk/collect/android/http/CaseInsensitiveHeaders.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+public interface CaseInsensitiveHeaders {
+    @Nullable Set<String> getHeaders();
+
+    boolean containsHeader(String header);
+
+    @Nullable String getAnyValue(String header);
```
breakbusyloop, line 12: Rather than 'getFirstValue', the naming choice of 'Any' allows for the implementation to have its own approach for rapidly looking up a single value. Most client calls will only care about the case of a single value. If having multiple values is a concern, the client is responsible for navigating the whole collection.

collect_app/src/test/java/org/odk/collect/android/http/OkHttpCaseInsensitiveHeadersTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Test;
+import org.junit.Assert;
+import org.odk.collect.android.http.openrosa.okhttp.OkHttpCaseInsensitiveHeaders;
+
+import java.util.Set;
+import java.util.TreeSet;
+
+import okhttp3.Headers;
+
+public class OkHttpCaseInsensitiveHeadersTest {
```
seadowg, line 13: I was initially interested in if this would be a nice case for deploying a "contract testing" style like we use for `OpenRosaHttpInterface`. However it feel it might not be right as the "empty" implementation and this really only share one "contract" - if they are empty they return false. Interested if you have any thoughts on this however?

seadowg (approved):
Looks great! I like that you ended up with the reponsibility inside our "open rosa" world and definitely support the choice of adding to the abstractio rather than exposing an OkHttp object. This app is old enough to have gone through at least one http client switch already and little decisions like that will be helpful for the next one 👍

lognaturel:
I've marked this as `needs testing` but @opendatakit/testers I think this can probably be merged as-is given the tests and how low risk it is. Any major issues would show up when doing any request like getting a blank form or making a submission. You're about to do a full regression pass so that will let you try it with other servers.

collect_app/src/test/java/org/odk/collect/android/http/OkHttpCaseInsensitiveHeadersTest.java
breakbusyloop: I'm a bottom-up kind of dev. I'd rather have functional tests first, and I loathe Mock testing except as a last resort. Contract-based testing seems best suited to the boundary of external services or APIs, but internal interfaces should just be simple tests that rely on no shared state, can run in isolation and potentially even run in parallel.

mmarciniak90:
@lognaturel sounds good to us. Thanks for marking this as a needs testing. Thanks to that the PR was not overlooked by us. 

I looked at basics cases on Android: 7.0, 6.0, 4.4
- download form from Aggregate as an anonymous user
- download form from Aggregate as a logged user
- downloading form from GS

collect_app/src/test/java/org/odk/collect/android/http/OkHttpCaseInsensitiveHeadersTest.java
seadowg: Right. With our OpenRosaHttpInterface we are drawing a boundary between our "App" and how it communicates with OpenRosa and I think you're right that for this case there isn't really that boundary.

Issue #3150, opened 2019-06-11
Double checkbox is visible on `Reset application` dialog on Android 4.x


mmarciniak90:
#### Software and hardware versions 
Collect v1.22.0,
Android v4.x

#### Problem description
Double checkbox is visible on `Reset application` dialog on Android 4.x

#### Steps to reproduce the problem
1. The user goes to `Admin Settings` > `Reset application...`
2. The user sees the dialog with two checkboxes

![Screenshot_2019-06-11-14-49-43](https://user-images.githubusercontent.com/3907554/59269893-5a31db80-8c50-11e9-953e-b8d4a59cde07.png)

#### Expected behavior
One checkbox is visible

grzesiek2010:
We haven't changed the related code recently so I guess the reason is migrating to Android X. I tried but I can't see any easy way to fix it so maybe we could stop forcing checkboxes to be at the right side? It would look like:
![Screenshot_1560258668](https://user-images.githubusercontent.com/3276264/59278599-98d09180-8c62-11e9-9d27-46970292bc38.png)
![Screenshot_2019-06-11-16-09-44](https://user-images.githubusercontent.com/3276264/59278613-9e2ddc00-8c62-11e9-8951-ce179bc211da.png)

and it would fix the problem with doubled checkboxes.

PR #3474, opened 2019-11-27
Fixed distorted buttons

### LINKED ISSUE #3472: RadioButtons/CheckBoxes are distorted
Current master branch

#### Problem description
This is what I see on Android 5.1 Samsung Galaxy S4:
Screenshot_2019-11-26-15-07-25
Screenshot_2019-11-26-15-07-38

grzesiek2010:
Closes #3472 

I tested the solution on the same device I noticed the problem. Also I used emulators to check it on different android versions.

It's a trick but it seems to fix the problem. I can't be sure if it's a perfect solution but the current one which turned out to be faulty is also a trick so we should test it on as many devices as possible and merge this one if it turns out to be better.

We should just test questions like:
- select_one
- select_multiple
- trigger

nothing else should be affected.

Any form with select_one/select_multiple/trigger questions so All widgets form for example.

No.


grzesiek2010:
@seadowg what do you think?

seadowg (changes requested):
I was initially so excited that this worked. I had tried drawable padding but I don't think I'd tried explicitly set the drawableLeft/Start. Annoyingly on API 19 and below I'm seeing this:

Screenshot_1574876255

I still like it more as a solution though (it doesn't cause the ripple effect to be off center) so I think it'd make sense to still use this instead but just only for API 21+. 

Of course if you can get rid of that double radio button that would be even better! I had a look at removing `drawableStart` and switching the button back to a `RadioButton` (rather than `AppCompatRadioButton`) but no joy. I don't understand why the `android:button="@null"` isn't working.

grzesiek2010:
Ahh it is the same problem like #3150 I will try to fix it

grzesiek2010:
@seadowg the problem with Android 4 is fixed.

seadowg:
Urgh. Using `buttonCompat` seems so obvious with the benefit of hindsight. Amazing!

I think we should QA this on a few different Android versions just to check we've got this nailed down.

seadowg (approved):

kkrawczyk123:
**Tested withs success**
Verified on Androids: 4.2, 4.4, 5.1, 6.0, 7.1, 8.1, 9..0 and 10.

PR #3477, opened 2019-11-27
Add track-changes-reason support

### LINKED ISSUE #3471: Form with `track-changes-reason=on-form-edit` adds reason for change to audit log
On form exit, Collect will show an “Enter reason for change” dialog if all the following are true:
* `track-changes-reasons` is set to `on-form-edit` in the form definition
* The `odk-instance-first-load` event was not fired at the beginning of this form session
* The user has explicitly changed (e.g., we ignore calculates) one or more values at some point in this form session (even if all changes were subsequently reverted)
* The user has tapped “Save Form and Exit” or tapped the back button and “Save Changes”
* Logic re-computation (including constraint checking if required) has completed successfully

The dialog has two buttons: Cancel and Ok.
* When Cancel is tapped, the dialog is dismissed
* When Ok is tapped, the reason for change is written to the audit log, the dialog is dismissed, encryption is performed if required, and the form is closed.

Spec: https://docs.google.com/document/d/1NyBFhASCOAqOSmz70diQObKpymVE59_MuwjOUPCKlR8

Design for this feature: https://github.com/seadowg/collect-design/tree/master/prototypes/track-changes-reason

**Acceptance**

- [x] Given I've downloaded and finalized a form with `track-changes-reasons=on-form-edit`
When I open the form via "Edit Saved Form"
And I change a value in the form
And I navigate to the end of the form and click "Save Form and Exit"
Then I am told I need to provide a reason for the changes I've made

- [x] Given I've downloaded and finalized a form with `track-changes-reasons=on-form-edit`
When I open the form via "Edit Saved Form"
And I change a value in the form
And I press back
And I click "Save changes"
Then I am told I need to provide a reason for the changes I've made

- [x] Given I've downloaded and finalized a form with `track-changes-reasons=on-form-edit`
When I open the form via "Edit Saved Form"
And I change a value in the form
And I press back
And I click "Ignore changes"
Then I return to the main menu

- [x] Given I've downloaded and finalized a form with `track-changes-reasons=on-form-edit`
When I open the form via "Edit Saved Form"
And I don't change any values in the form
And I press back
And I click "Save changes"
Then I return to the main menu

- [x] Given I've downloaded and finalized a form with `track-changes-reasons=on-form-edit`
When I open the form via "Edit Saved Form"
And I don't change any values in the form
And I navigate to the end of the form and click "Save Form and Exit"
Then I return to the main menu

- [x] Given I've been told I need to provide a reason for the changes I've made
When I enter a reason
And click "Save"
Then I return to the main menu
And a `change reason` event is present in the form's audit log
And a `change-reason` column containing my reason is exists for the event

- [x] Given I've been told I need to provide a reason for the changes I've made
When I enter a blank (or whitespace) reason
And click "Save"
Then I remain on the reason for changes prompt

- [x] Given I've been told I need to provide a reason for the changes I've made
When I enter a reason with commas or quotes (`"Callum" told me, to do it`)
And click "Save"
Then I return to the main menu
And a `change-reason` event is present in the form's audit log
And a `change reason` column containing my reason (`"""Callum"" told me, to do it"`) is exists for the event

- [x] Given I've been told I need to provide a reason for the changes I've made
When I enter a reason
And I press back/click the cross
Then I return to where I was in the form
And the form is not saved

- [x] Given I've been told I need to provide a reason for the changes I've made
When rotate the device
Then I am still prompted to provide a reason

- [x] Given I've been told I need to provide a reason for the changes I've made
When background/resume the app
Then I am still prompted to provide a reason

- [x] Given I've downloaded and finalized a form without `track-changes-reasons`
When I open the form via "Edit Saved Form"
And I change a value in the form
And I navigate to the end of the form and click "Save Form and Exit"
Then I return to the main menu

- [x] Given I've downloaded and finalized a form without `track-changes-reasons`
When I open the form via "Edit Saved Form"
And I change a value in the form
And I press back
And I click "Save changes"
Then I return to the main menu

seadowg:
Closes #3471. This implements the feature described (with designs and specs) in that issue.

Some notes on the implementation:

* I seperated all knowlege of audit confiuguration out from the `AuditEvent` as I was finding hard to keep track of all the places this needed passed. There is now a helper dedicated to converting an `AuditEvent` to a CSV line as a consequence of that. This also makes the `AuditEvent` more of a data object which I think makes sense given how it's used.
* The implementation uses a ViewModel to encapsualte the logic around when to show the Changes Reason prompt. This requires "notification" style messages to be sent to the ViewModel from the `FormEntryActivity` (`savingForm` etc). The smell here is that both the ViewModel and the Activity need to know a lot about the loading/saving lifecycle of a form. It's feel like a better and better idea to have a `FormEntryViewModel` that deals with loading and saving forms. This would let us move loading/saving logic out of `FormEntryActivity` and have it bind to LiveData values t let it know when it should display the form, show a dialog etc. I'm keen to give this refactor a shot for 1.26.
* I've bumped the font size up for both this new Dialog and the Identify User one. They know use `Body1` for their description text rather than `Body2`. It seems that we're all feeling like the font sizes are "too damn small" and I've also receieved that feedback from at least one real life designer.
* I've extracted a `MaterialFullScreenDialogFragment` we can use for flows like this in the future. Hopefully we can replace this with a Material Components implementation in the future.

Ran throught the acceptance in #3471 using a form I built with the [XLSForm part of this feature](https://github.com/XLSForm/pyxform/pull/400).

There shouldn't be any changes other than the new feature. Would be good to test form entry in general to make sure we haven't messed up any existing flows.

The code has an example form used for testing (`track-changes-reason-on-edit`) that can be used.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
@@ -63,16 +63,16 @@ public class EditSavedFormPage extends Page<EditSavedFormPage> {
     }
 
     public IdentifyUserPromptPage clickOnFormWithIdentityPrompt(String formName) {
-        clickOnFormButton(formName);
+        scrollToAndClickOnForm(formName);
         return new IdentifyUserPromptPage(formName, rule).assertOnPage();
     }
 
     public FormEntryPage clickOnForm(String formName) {
-        clickOnFormButton(formName);
+        scrollToAndClickOnForm(formName);
         return new FormEntryPage(formName, rule);
     }
 
-    private void clickOnFormButton(String formName) {
+    private void scrollToAndClickOnForm(String formName) {
         onView(withText(formName)).perform(scrollTo(), click());
     }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
@@ -26,6 +26,7 @@ import org.odk.collect.android.adapters.InstanceListCursorAdapter;
 
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.action.ViewActions.click;
+import static androidx.test.espresso.action.ViewActions.scrollTo;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
 import static androidx.test.espresso.matcher.ViewMatchers.hasDescendant;
 import static androidx.test.espresso.matcher.ViewMatchers.withTagValue;
@@ -56,13 +57,22 @@ public class EditSavedFormPage extends Page<EditSavedFormPage> {
         return this;
     }
 
-    public OkDialog clickSavedFormWithDialog(String instanceName) {
-        onView(withText(instanceName)).perform(click());
+    public OkDialog clickOnFormWithDialog(String instanceName) {
+        clickOnForm(instanceName);
         return new OkDialog(rule).assertOnPage();
     }
 
     public IdentifyUserPromptPage clickOnFormWithIdentityPrompt(String formName) {
-        onView(withText(formName)).perform(click());
+        clickOnFormButton(formName);
         return new IdentifyUserPromptPage(formName, rule).assertOnPage();
+    }
+
+    public FormEntryPage clickOnForm(String formName) {
+        clickOnFormButton(formName);
+        return new FormEntryPage(formName, rule);
+    }
+
+    private void clickOnFormButton(String formName) {
+        onView(withText(formName)).perform(scrollTo(), click());
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
@@ -23,12 +23,9 @@ public class QuitFormDialog {
     }
 
     public static AlertDialog show(Context context, FormController formController, Listener listener) {
-        String title;
-        {
-            title = (formController == null) ? null : formController.getFormTitle();
-            if (title == null) {
-                title = context.getString(R.string.no_form_loaded);
-            }
+        String title = (formController == null) ? null : formController.getFormTitle();
+        if (title == null) {
+            title = context.getString(R.string.no_form_loaded);
         }
 
         List<IconMenuItem> items;
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
@@ -1,0 +1,72 @@
+package org.odk.collect.android.formentry;
+
+import android.content.Context;
+import android.widget.ListView;
+
+import androidx.appcompat.app.AlertDialog;
+
+import com.google.common.collect.ImmutableList;
+
+import org.odk.collect.android.R;
+import org.odk.collect.android.adapters.IconMenuListAdapter;
+import org.odk.collect.android.adapters.model.IconMenuItem;
+import org.odk.collect.android.logic.FormController;
+import org.odk.collect.android.preferences.AdminKeys;
+import org.odk.collect.android.preferences.AdminSharedPreferences;
+import org.odk.collect.android.utilities.DialogUtils;
+
+import java.util.List;
+
+public class QuitFormDialog {
+
+    private QuitFormDialog() {
+    }
+
+    public static AlertDialog show(Context context, FormController formController, Listener listener) {
+        String title;
+        {
+            title = (formController == null) ? null : formController.getFormTitle();
+            if (title == null) {
+                title = context.getString(R.string.no_form_loaded);
+            }
+        }
+
+        List<IconMenuItem> items;
+        if ((boolean) AdminSharedPreferences.getInstance().get(AdminKeys.KEY_SAVE_MID)) {
+            items = ImmutableList.of(new IconMenuItem(R.drawable.ic_save, R.string.keep_changes),
+                    new IconMenuItem(R.drawable.ic_delete, R.string.do_not_save));
+        } else {
+            items = ImmutableList.of(new IconMenuItem(R.drawable.ic_delete, R.string.do_not_save));
+        }
+
+        ListView listView = DialogUtils.createActionListView(context);
+
+        final IconMenuListAdapter adapter = new IconMenuListAdapter(context, items);
+        listView.setAdapter(adapter);
+        listView.setOnItemClickListener((parent, view, position, id) -> {
+            IconMenuItem item = (IconMenuItem) adapter.getItem(position);
+            if (item.getTextResId() == R.string.keep_changes) {
+                listener.onSaveChangedClicked();
+            } else {
+                listener.onIgnoreChangesClicked();
+            }
+        });
+
+        AlertDialog alertDialog = new AlertDialog.Builder(context)
+                .setTitle(
+                        context.getString(R.string.quit_application, title))
+                .setPositiveButton(context.getString(R.string.do_not_exit), (dialog, id) -> {
+                    dialog.cancel();
+                })
+                .setView(listView)
+                .create();
+        alertDialog.show();
+        return alertDialog;
+    }
+
+    public interface Listener {
+        void onSaveChangedClicked();
+
+        void onIgnoreChangesClicked();
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
@@ -112,4 +112,47 @@ public class AuditConfig {
     public boolean isTrackChangesReasonEnabled() {
         return isTrackChangesReasonEnabled;
     }
+
+    public static class Builder {
+        private String mode;
+        private String locationMinInterval;
+        private String locationMaxAge;
+        private boolean isTrackingChangesEnabled;
+        private boolean isIdentifyUserEnabled;
+        private boolean isTrackChangesReasonEnabled;
+
+        public Builder setMode(String mode) {
+            this.mode = mode;
+            return this;
+        }
+
+        public Builder setLocationMinInterval(String locationMinInterval) {
+            this.locationMinInterval = locationMinInterval;
+            return this;
+        }
+
+        public Builder setLocationMaxAge(String locationMaxAge) {
+            this.locationMaxAge = locationMaxAge;
+            return this;
+        }
+
+        public Builder setIsTrackingChangesEnabled(boolean isTrackingChangesEnabled) {
+            this.isTrackingChangesEnabled = isTrackingChangesEnabled;
+            return this;
+        }
+
+        public Builder setIsIdentifyUserEnabled(boolean isIdentifyUserEnabled) {
+            this.isIdentifyUserEnabled = isIdentifyUserEnabled;
+            return this;
+        }
+
+        public Builder setIsTrackChangesReasonEnabled(boolean isTrackChangesReasonEnabled) {
+            this.isTrackChangesReasonEnabled = isTrackChangesReasonEnabled;
+            return this;
+        }
+
+        public AuditConfig createAuditConfig() {
+            return new AuditConfig(mode, locationMinInterval, locationMaxAge, isTrackingChangesEnabled, isIdentifyUserEnabled, isTrackChangesReasonEnabled);
+        }
+    }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
@@ -52,13 +52,15 @@ public class AuditConfig {
     private final boolean isTrackingChangesEnabled;
 
     private final boolean isIdentifyUserEnabled;
+    private final boolean isTrackChangesReasonEnabled;
 
-    public AuditConfig(String mode, String locationMinInterval, String locationMaxAge, boolean isTrackingChangesEnabled, boolean isIdentifyUserEnabled) {
+    public AuditConfig(String mode, String locationMinInterval, String locationMaxAge, boolean isTrackingChangesEnabled, boolean isIdentifyUserEnabled, boolean isTrackChangesReasonEnabled) {
         this.locationPriority = mode != null ? getMode(mode) : null;
         this.locationMinInterval = locationMinInterval != null ? Long.parseLong(locationMinInterval) * 1000 : null;
         this.locationMaxAge = locationMaxAge != null ? Long.parseLong(locationMaxAge) * 1000 : null;
         this.isTrackingChangesEnabled = isTrackingChangesEnabled;
         this.isIdentifyUserEnabled = isIdentifyUserEnabled;
+        this.isTrackChangesReasonEnabled = isTrackChangesReasonEnabled;
     }
 
     private LocationClient.Priority getMode(@NonNull String mode) {
@@ -105,5 +107,9 @@ public class AuditConfig {
 
     public boolean isIdentifyUserEnabled() {
         return isIdentifyUserEnabled;
+    }
+
+    public boolean isTrackChangesReasonEnabled() {
+        return isTrackChangesReasonEnabled;
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
@@ -19,7 +19,7 @@ import org.odk.collect.android.R;
  * (https://material.io/components/dialogs/#full-screen-dialog) as no implementation currently
  * exists in the Material Components framework
  */
-public class MaterialFullScreenDialogFragment extends DialogFragment {
+public abstract class MaterialFullScreenDialogFragment extends DialogFragment {
 
     private Toolbar toolbar;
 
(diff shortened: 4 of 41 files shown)


lognaturel:
I've unfortunately only been able to take a pretty quick cruise through this. I don't have any concerns from what I've seen. I like the high-level decisions made, particularly around disentangling `AuditEvent` and `AuditConfig`. I'm glad you backed out of reorganizing a bunch of things in `FormEntryActivity` given where we are with v1.25. I do think coming back to it soon would be valuable. I think formalizing a form lifecycle concept would be a good direction to move in.

@grzesiek2010, it would be really great to get a deeper review during your Monday with the goal of getting to QA as soon as possible since this is relatively high risk.

grzesiek2010:
1. It doesn't work when `Don't keep activities` option is enabled (the dialog is not shown).
2. The toolbar has like two shadows?
Screenshot_2019-12-02-13-42-17
3. When I rotate a device the behavior is strange (from portrait lo landscape), I see the EditText for a sec then it disappears but the keyboard is still visible:
Screenshot_2019-12-02-13-45-38
Screenshot_2019-12-02-13-46-04
4. The way it checks for changes is not perfect for example using the form you attached for tests:
- save a form with answer `my answer`
- edit the saved form
- remove the last char `r` and navigate forward
- navigate backward and add the removed char back
- navigate forward and try to save the form.

The dialog shouldn't be visible but it is.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
```diff
@@ -52,13 +52,15 @@
     private final boolean isTrackingChangesEnabled;
 
     private final boolean isIdentifyUserEnabled;
+    private final boolean isTrackChangesReasonEnabled;
 
-    public AuditConfig(String mode, String locationMinInterval, String locationMaxAge, boolean isTrackingChangesEnabled, boolean isIdentifyUserEnabled) {
+    public AuditConfig(String mode, String locationMinInterval, String locationMaxAge, boolean isTrackingChangesEnabled, boolean isIdentifyUserEnabled, boolean isTrackChangesReasonEnabled) {
```
grzesiek2010, line 57: We have more and more parameters maybe it would be good to use the builder pattern?

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+public class QuitFormDialog {
+
+    private QuitFormDialog() {
+    }
+
+    public static AlertDialog show(Context context, FormController formController, Listener listener) {
+        String title;
+        {
```
grzesiek2010: Why did you add this block?

grzesiek2010: Ahh ok it was there, but still we could get rid of it.

collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
```diff
@@ -0,0 +1,82 @@
...
+import androidx.fragment.app.DialogFragment;
+
+import org.odk.collect.android.R;
+
+/**
+ * Provides an implementation of Material's "Full Screen Dialog"
+ * (https://material.io/components/dialogs/#full-screen-dialog) as no implementation currently
+ * exists in the Material Components framework
+ */
+public class MaterialFullScreenDialogFragment extends DialogFragment {
```
grzesiek2010: Could be abstract right? since it's just used to create other classes.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
```diff
@@ -56,13 +57,22 @@ public EditSavedFormPage checkInstanceState(String instanceName, String desiredS
...
+        clickOnFormButton(formName);
         return new IdentifyUserPromptPage(formName, rule).assertOnPage();
     }
+
+    public FormEntryPage clickOnForm(String formName) {
+        clickOnFormButton(formName);
+        return new FormEntryPage(formName, rule);
+    }
+
+    private void clickOnFormButton(String formName) {
```
grzesiek2010: What do you mean by form button here?

seadowg:
Thanks for checking everything out @grzesiek2010. Some thoughts:

1. Great catch! I'll look into this. A little bit confused as to why this isn't working...
2. Ah looks like the code for adding the status bar in for lower API levels (in `MaterialFullScreenDialog`) is using the wrong API level to switch on. Did you see this in 21?
3. Yeah that behaviour is standard for`EditText` as far as I know - it full screens the keyboard. You can disable it I believe but it also happens for our `StringWidget` and it feels like it's a nice UX improvement for phone sized screens.
4. Ah interesting. I'll look into this. I think I might have got confused between the "exit view" and "exit form" events when getting this in. I agree with you that in your case the Change Reason dialog should be shown.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
seadowg: Yeah "simple button" felt confusing but this isn't actually better. Let me have another go.

collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
seadowg: Yeah totally. Will add that.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
seadowg: Yeah good shout. I'll generate one.

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialog.java
seadowg: Yeah I didn't notice that somehow!

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditConfig.java
grzesiek2010: Kotlin would solve it right? But we use Java.

seadowg: Haha yes it would! The [`data` classes](https://kotlinlang.org/docs/reference/data-classes.html) automatically get a `copy` function that works just like a builder because of the named/optional parameters.

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/EditSavedFormPage.java
seadowg: Yeah that's not a great name. Will rethink.

seadowg:
@grzesiek2010 I can't seem to reproduce the problem you saw for 4). Could you double check and send me the form?

grzesiek2010:
I checked again and I'm able to reproduce it. I used your form: https://github.com/opendatakit/collect/pull/3477/files#diff-27e4fee6a35136cb874044a89ae00d32

ezgif com-video-to-gif

seadowg:
@grzesiek2010 I completely misread the problem. That's intentional. In [the spec](https://docs.google.com/document/d/1NyBFhASCOAqOSmz70diQObKpymVE59_MuwjOUPCKlR8/edit) for the feature it's mentioned that we should prompt for a reason for change "even if all changes were subsequently reverted". I'm assuming this was so the `track-changes` and `track-changes-reasons` events line up. Does that sound right? Good to get your read on this as well @lognaturel.

grzesiek2010:
It's a bit strange because if you revert such a change immediately, then you won't see the dialog. It's displayed only if you navigate forward/backward between changing something and then reverting the change.

seadowg:
@grzesiek2010 yeah that's true. The audit log only registers "changes" when you leave the view. I could go either way on this - on one hand it feels inconsistent but on the other it makes sense as the user isn't finished "editing" until they navigate away from the question.

Maybe because this effects both `track-changes` and `track-changes-reasons` (you'll see the changes in `question` events as well) we could get feedback on whether it's an edge case we want to change and then look at improving it in the future?

grzesiek2010:
ok

grzesiek2010:
Seems like the is still a problem with the keyboard (I described in in point 3). The behavior is slightly different, now when I rotate the device the `EditText` is not focused but the keyboard is visible.

Since it's not a serious problem I am going to add `needs testing` label, but if you have time please take a look.

grzesiek2010 (approved):

grzesiek2010:
Ahh accidentally I noticed something else...
If you are on the last screen and you rotate your device the dialog appears automatically (if it had been visible before - I mean if you open it and close an then rotate your device).
It's also not a serious problem so I'm going to keep the `needs testing` label.

seadowg:
@grzesiek2010 is the keyboard behaviour different from what you see if you rotate while the keyboard is up for a standard text (`StringWidget`) question. 

Urgh that last screen issue could have something to do with the fix for "don't keep activities". The `FormEntryActivity` loading/resuming lifecycle is pretty nasty (as I'm sure you've felt before). I'll have a look at this but I agree it's good to get QA on this feature to see if we can find anything else 👍

grzesiek2010:
> is the keyboard behaviour different from what you see if you rotate while the keyboard is up for a standard text (StringWidget) question.

yes in other cases I have the keyboard and the EditText in landscape mode:
Screenshot_2019-12-03-16-51-40

here the edittext is not focused but  the keyboard is visible.

seadowg:
@grzesiek2010 hmmm I can't seem to reproduce (the unfocused field in landscape). If you type in that state does it type in the field? What API level is that device?

seadowg:
@grzesiek2010 just added in a `requestFocus` that I had in the "Identify User" dialog but not this one. Didn't really change anything for me on the emulator but maybe helps for the device you were seeing the rotation issues on.

grzesiek2010:
I used:
Samsung Galaxy S4 
Android 5.0.1
Don't keep activities - doesn't matter

grzesiek2010:
Another thing I noticed:
if you rotate your device x times you need to click the x icon (or the back button) x times to close the dialog

seadowg:
Ok so looks like the new problems reported are:

1. Samsung Galaxy S4 doesn't focus on the reason input field when the device is rotated
2. Every time the device is rotated a new Change Reason prompt is opened (so the user has to hit the cross/press back n times if they want to get back to the form)
3. Rotating the device on the last page of the form (the Save and Exit page), once the Change Reason prompt has been opened and closed, reopens the prompt

grzesiek2010:
Right.

seadowg:
@grzesiek2010 I've fixed 1 and 2 of the last set of issues. 3 is actually a little complex. I'm running into some really nastiness with needing to block the `FormEntryActivity` but it also being responsible for saving the form.

grzesiek2010:
The third issue seems like something rare and not important, it doesn't cause any problems apart from the fact it's weird, so I would say please file a separate issue if you are not able to fix it by tomorrow morning.

seadowg:
I'm pretty sure I've got a nicer solution so just running tests and cleaning up. I'll remove the "needs testing" for the moment but if someone has already started I can just revert new commits as don't want to force multiple QA runs.

kkrawczyk123:
@seadowg The QA round has already started a little bit, hope that info will be valuable for you:

- An empty instance folder is created when the form is opened not saved in each try - not big issue but it doesn't happen with other forms and it doesn't make much sense 

- Like @grzesiek2010  has already mentioned when rotate screen on Save and Exit view - after screen rotation Change Reason page is visible I have also noticed another similar case: Edit form, Go to Save and Exit page, swipe back to question, rotate device - Change Reason page is visible too

- When: Edit question, click on the hierarchy button, open question again from hierarchy view - Change Reason page displays except for question page. Exactly the same happens when: Edit question, Open General Settings from form view, click back button.

- There are some ways of saving the edited form without input a change reason: 
1. Edit question, click on Save button from form view, Form saved toast is visible, click on back button and ignore changes.
2. Edit question, click on Save button from form view, Form saved toast is visible, kill an app
3. Edit question, go to Save and Exit, click Save and Exit, Change Reason page is visible, kill an app 
The form is saved with edit without reason change - changes in audit file are visible but change reason column is empty.

seadowg:
@kkrawczyk123 amazing thanks 👍. I'll hold off pushing any changes until you're done testing. I'll have a look at addressing the issues you've found and let me know when you're done and if there are any more!

kkrawczyk123:
@seadowg Actually, I have dropped testing right know as you removed the label and want to add changes, so I don't think it's necessary to wait for me. Just tag me and @mmarciniak90 when you are ready with the new approach and we will start all over again.

seadowg:
@kkrawczyk123 sounds good will do

seadowg:
@kkrawczyk123 @mmarciniak90 those issues should be fixed now. A couple appeard to be existing (I've put details below). One user flow the spec doesn't mention is:

1. changing a value
1. hitting save (in the toolbar)
1. entering a reason for change
1. not changing a value
1. hitting save again

Right now you'll see another reason for changes prompt as we consider this form entry "session" to still have changes. I could go either way on whether this is the right behaviour or not. Any thoughts (cc @lognaturel @grzesiek2010)?

> An empty instance folder is created when the form is opened not saved in each try - not big issue but it doesn't happen with other forms and it doesn't make much sense

I'm actually see this for any form with audit enabled in the current beta. It is weird. Would be good to double check but looks like we should have a seperate issue for this.

> 3. Edit question, go to Save and Exit, click Save and Exit, Change Reason page is visible, kill an app
The form is saved with edit without reason change - changes in audit file are visible but change reason column is empty.

This one is really strange! Great catch. To me it looks like this is an existing issue. If you edit a form with (change a value and swipe to the next/previous question) and then quit the app the audit log will record the change (if it has `track-changes` enabled) but if you look at the actual instance file you'll see it will still have the original value (although it looks like it has changed in the app). Reopening the instance to edit and then doing an "ignore changes" will then make the app and the instance file mirror each other again. Are you able to double check I'm not getting myself confused here? Again, if I'm right this should be a seperate issue I think.

lognaturel:
Yikes, when writing the specification, we didn't consider that the user could use the toolbar save button to cheat. Good catch, @kkrawczyk123. 

> Right now you'll see another reason for changes prompt as we consider this form entry "session" to still have changes.

I think this is ok, @seadowg. We need to make sure that the documentation mentions that behavior. It should probably encourage survey managers to train enumerators not to use the save button. This leads me to notice that we don't have documentation on savepoints and recovery after app crash which we should. Then we can make it clear that enumerators don't need to worry about saving progress. In fact, perhaps we should consider removing that toolbar save button entirely.

> An empty instance folder is created when the form is opened not saved in each try - not big issue but it doesn't happen with other forms and it doesn't make much sense

I believe this is inevitable when there's an audit configured. The audit is written out immediately to disk so that no events risk being lost if the app crashes. This means that the containing folder has to be created. This seems fine to me.

> Edit question, go to Save and Exit, click Save and Exit, Change Reason page is visible, kill an app
The form is saved with edit without reason change - changes in audit file are visible but change reason column is empty.

This I don't think we can really do anything about. In general, if the app is killed when the change reason page is shown, we don't have any way of detecting that or doing anything before the kill goes through. This should be documented and it should be clear that if folks are worried about cheating, they should make sure they have plenty of RAM and install some kind of system locking app.

> Reopening the instance to edit and then doing an "ignore changes" will then make the app and the instance file mirror each other again.

I can't reproduce this. When I open the instance for edit, I see the changes that I made between saving and killing. When I tap "ignore changes" and then come back to edit the instance, the changes that I made between saving and killing the app are lost.

All this savepoint behavior is implicit and pretty confusing. I've filed #3494 to add it to the list of things to redesign.

seadowg:
> I can't reproduce this. When I open the instance for edit, I see the changes that I made between saving and killing. When I tap "ignore changes" and then come back to edit the instance, the changes that I made between saving and killing the app are lost.

I think (🤞) we're seeing the same thing but that it's just really hard to explain. By "mirror" I meant that the you never see those changes (the ones between saving and killing) in the instance file. I was a little over ambitious in not writing out full repro steps there sorry!

kkrawczyk123:
@seadowg I have noticed one more cheating scenario: Edit submission, go to General Settings from form view, change some of User interface option like - language/mode and go back through Edit Saved form to submission - you can save without input reason code.

kkrawczyk123:
@seadowg I can see another issue, reproduction steps:
1. Open Edit form
2. Do not change anything
3. Click on Save button on form view
4. Input a change
5. Click Save button again
The form is successfully saved, toast is visible - you can swipe and click on ave and Exit button. The form is Saved and Reason Change Page is not visible at all.

seadowg:
@kkrawczyk123 wow nice find. I'll have a look at this.

seadowg:
@kkrawczyk123 The first issue (changing a setting that pops you back to the main menu) is part of #3494 in my mind and an existing problem for normal `track-changes`. An easier way to reproduce is:

1. Open form to edit
2. Change a value
3. Kill app
4. Open form to edit again
5. Save

In this case you end up saving the edits but you won't see a `question` event (with the changes).

The second one is also an existing issue. Again we can reproduce with `track-changes`:

1. Open a form to edit
2. Save form
3. Make a change
4. Save

If you check the audit there is no `question` event with the change. I'll add the details of these to the existing issue (although we might want to break into several small ones).

I don't want to belittle these problems by the way - they're great finds and genuine problems with the way our audit log works. It feels like we've found a whole new work stream in testing this feature! I don't think they should block merging this however, as the implementation of this feature is heavily tied to the mechanism for `track-changes`. 

Let me know if you find anything else!

**EDIT**: Actually ended up filing a second issue for the save button at #3498

kkrawczyk123:
@seadowg I can't see more suspicious things. So I think it's ready for merging if @lognaturel agrees with me too.

kkrawczyk123:

lognaturel:
Good finds, @kkrawczyk123. Clearly there are some issues around the savepoint behavior. Luckily I don't think any of it is urgent for v1.25.

PR #3480, opened 2019-11-28
Setup reference manager when copying forms in tests

seadowg:
This fixes the failing tests we were seeing (`ExternalSecondaryInstancesTest`). 

Ran tests locally.

I think we should use as much of the "real" form loading code as possible in tests as these problem will just keep coming up otherwise. This change is hopefully a small step in the right direction as it extracts some code from `FromLoaderTask` to the new `forms` package (which @lognaturel created in another PR). I'd like to see form loading extracted from the task as much as possible but definitely didn't want to carry out that refactor in the red.

Shouldn't be any change in behaviour. Some code was moved but probably not worth a QA pass.


seadowg:
Marking this as "high-priority" as it fixes the build

grzesiek2010 (approved):

PR #3482, opened 2019-11-29
Fixed crash in Mapbox

### LINKED ISSUE #3479: Collect crashes when the offline layer is changed to 'None'
Collect master

#### Problem description
Collect crashes when the offline layer is changed to 'None'

Preconditions:
1. User uses Mapbox
2. The offline layer is set

Use case 2:
1. User fills the form with some Geo widget with map
2. User opens layers dialog
3. User selects `None`
4. Collect crashes 

#### Expected behavior
The offline layer is disabled 

#### Other information 
Noticed during testing #3447 
It's hard to define stacktrace. 
Here is possible stacktrace but I'm not sure

Screenshot from 2019-11-28 14-22-41

grzesiek2010:
Closes #3479 

I reproduced the crash and confirmed the pr fixes the issue. I also debugged the code to confirm that the code I removes was redundant.

Just to clarify: the issue is not related to changing the layer to `None` but generally to changing any layers.

The crash was caused [here](https://github.com/opendatakit/collect/compare/master...grzesiek2010:COLLECT-3479?expand=1#diff-fdd039fc806c853bc4ba175e0ab623edL585) because we tried to remove a layer that doesn't exist. It didn't cause a problem when we had mapbxox v8.3.0 but since we updated to 8.4.0 it's been crashing.

I removed the code because as I said above it always tried to remove a layer that didn't exist. The code was redundant. It was called just in `loadReferenceOverlay()` which is called in two places:
https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java#L208
https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java#L161
in both cases after we call: `map.setStyle` what clears layers we then again tried to remove.

Testing changing layers in Mapbox would be enough. Nothing else should be affected.

Any form with geo widgets.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
@@ -104,7 +104,6 @@ public class MapboxMapFragment extends org.odk.collect.android.geo.mapboxsdk.Map
     private boolean isDragging;
     private String styleUrl = Style.MAPBOX_STREETS;
     private File referenceLayerFile;
-    private final List<Layer> overlayLayers = new ArrayList<>();
     private final List<Source> overlaySources = new ArrayList<>();
     private static String lastLocationProvider;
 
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
I think that this was added in preparation for multiple user-selected layers. When that does happen, there will need to be a list of layers to manage but until then I agree it makes sense to remove.

mmarciniak90:
**Tested with success**

Verified on Android: 10.0, 9.0, 7.0, 6.0, 5.1, 4.4, 4.2

Verified cases:
- changing layers on GeoPoint, GeoTrace, and GeoShape widgets
- Light/dark mode
- English/Spanish language

PR #3483, opened 2019-12-02
Rework layouts for QuestionWidget and ODKView

seadowg:
This should provide a better solution to #3504 (than #3505).

The core change in this PR was to define the `ODKView` and `QuestionWidget` custom views using XML layouts. This should make them easier to understand and change as we'll be able to use Android Studio's design tools with them. 

Some other changes I made along the way were:

* Add `setText` to `AudioVideoImageTextLabel` that lets you just set text rather than a setting up a `TextView`. This simplifies the majority of places the class is used.
* Rework typography for `ODKView` and `QuestionWidget`. Biggest change here visually is that the guidance and help text are no longer italic and use medium emphasis colors. Screenshots for the changes below.
* Introduce a `QuestionTextSizeHelper` for calculating the size of for a `TextView` based on it's named text appearance.
* Move shared helpers in `QuestionWidget` for creating views in widgets to a new `WidgetViewUtils` class. It might be possible to convert many of this view creation to shared styles in the future.

| Before        | After          |
|------------|------------|
| Screenshot_1576600352 | Screenshot_1576602463 |

Ran tests locally and eyeballed the AllWidgets form. Also double checked original bug hadn't regressed.

As above this lets us use design tooling when working with these views which should make our lives a lot easier and hopefully prevent us missing things when changing things.

There should be some visual changes but hopefully just improvements. As this effects the class every widget inherits from it would be good to check them all in as many configurations as possible.

It would also be good time to double check that using markdown/html as the question label/hint/guidance is working correctly (when we have audio/video as well) as there has been a lot of work in this area over the last couple of releases.


seadowg:
Closing this for the moment.

seadowg:
Reopened this now we're at code freeze. Closing was honestly just a piece of mental self help but will definitely be coming back for this now!

lognaturel:
Ooh, I completely lost track of this, sorry. Will try to review soon so it can stop accumulating conflicts.

seadowg:
@lognaturel conflicts fixed!

collect_app/src/main/res/layout/activity_capture_selfie_new_api.xml
```diff
@@ -18,7 +18,7 @@ limitations under the License.
     android:layout_height="match_parent" >
 
     <FrameLayout
-        android:id="@+id/container"
+        android:id="@+id/question_widget_container"
```
lognaturel: I think this is unintentional.

collect_app/src/main/res/layout/activity_capture_selfie_video_new_api.xml
```diff
@@ -14,7 +14,7 @@ limitations under the License.
 -->
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/container"
+    android:id="@+id/question_widget_container"
```
lognaturel: I think this is unintentional.

collect_app/src/main/res/values/refs.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
```
lognaturel: remove?

lognaturel (changes requested):
Couple of tiny catches that don't affect behavior.

The medium emphasis color for the hint text is not high-enough contrast, I think.

It also looks like something happened to the question text. Did the font change? It looks "more bold" and that makes it harder to read to me.

I'm a bit disappointed by the rendered views in Android Studio since one of the stated goals for this change is to make the tooling more useful. I thought this might get to a point where we could see something that looks like a question. It's certainly a step in the right direction but do you think there's more that can be done (at another time)? `odk_view` just shows the hierarchy label because the widgets and dividers are still added programmatically. `audio_video_image_text_label` shows the label and the media placeholders and I'd expect that to show up in `question_widget` but it doesn't.

seadowg:
> The medium emphasis color for the hint text is not high-enough contrast, I think.

I'll have a look at this. It's the recommended our surface color being grey doesn't work well this I think.

> It also looks like something happened to the question text. Did the font change? It looks "more bold" and that makes it harder to read to me.

Yeah it's always been bold but now it's using our Headline 6 style **and** is bold which I think pushed the letter spacing to close. Generally it seems like bolding isn't as much a thing in the Material typography. I'm glad you noticed this as I agree (but didn't want to bring it up and have everyone obsess over it). I'll look at taking the bold out!

seadowg:
@lognaturel great catch on the medium emphasis text having too low a contrast! Basically our `colorSurface` attribute was set to `#DE000000` which is black with an alpha of 0.87. This is the correct color for ["High Emphasis" in Material](https://material.io/design/color/text-legibility.html#text-backgrounds) but what I hadn't realized is that when we apply our "filter" (alpha of 0.6) to this color we get an alpha of 0.6 * 0.87 rather than an alpha of 0.6. 

Really the `colorOnSurface` should just be` #000000` (for light theme) anyway as it will be used for other things than text (like icons) so I've set it to that and introduced a "high emphasis" color in the same way that we had "medium emphasis" before and just used that as the default color for the typography styles.

lognaturel (dismissed):
Thanks for making those last subtle improvements!

lognaturel:
Wait a minute. The hint text is now the same size as the group text. There can be quite a lot of very useful text there so that seems too small.

seadowg:
Screenshot_1579700016

@lognaturel you're talking about the "This is optional, of course" right? It doesn't seem to small to me necessarily and it's the standard body size (16sp). Do you think the group text should be smaller or do you think the hint needs bumped up?

lognaturel:
> Do you think the group text should be smaller or do you think the hint needs bumped up?

The previous hint size seemed about right to me. But I know you'll tell me it's non-standard. I'll run it by @yanokwa to see what he thinks.

seadowg:
@lognaturel yeah good to get some other opinions. The current size is the standard body text but I could see that the text should be more important than that.

lognaturel:
Sounds great. QA can let us know their opinion and start looking for other regressions while we decide.

**EDIT**: @yanokwa doesn't feel strongly about the text size. He didn't notice it had changed when I showed it to him and he finds it comfortable. He likes that it is standard text size.

I didn't mention it initially because it felt small but everything feels small to me so I figured it was the same as before. It wasn't until I was looking at it again that I thought to compare it with what it was before. I also think it's slightly easier to read non-italicized text so the change didn't jump out as much to me.

I'm ok with proceeding.

I wonder whether we should shuffle a few planned things around and consider making a few more changes to the widget typography for this release. The button text or "No value" text looks really big, now.

seadowg:
@lognaturel we've got that work coming up (new widget framework and then changes to actual widgets). Maybe once we've got the new framework in we can pritoritize which widgets/elements we think should get changed first?

mmarciniak90:
@seadowg I set RTL language and I see one potentially unexpected appearance
Screenshots are from Android 7.0 and 8.1

| v1.25.1 | pr/3483 |
|---|---|
|Screenshot_20200129-154041 |Screenshot_20200129-154037|

seadowg:
@mmarciniak90 hmm I feel like we've run into this before. Let me take a look at it!

seadowg:
@mmarciniak90 I've fixed the layout. It doesn't really feel like the previous one is great but I think we should keep it unchanged for the moment.

lognaturel:
I think that the modified layout was actually more accurately RTL than the 1.25.1 one. Question labels showing to the right of the options seems like it would be the expected behavior in a RTL language.

That said, I think it's also fine to leave until a RTL user provides specific requirements for that layout.

> It doesn't really feel like the previous one is great 

These grids are inherently inappropriate for narrow devices and that's ok. Is that what you're referring to? It would look good on e.g. a tablet. We actually made a change to wrap hints at one point and immediately got the feedback that this was not what users wanted.

seadowg:
> These grids are inherently inappropriate for narrow devices and that's ok. Is that what you're referring to? It would look good on e.g. a tablet. We actually made a change to wrap hints at one point and immediately got the feedback that this was not what users wanted.

Huh. Interesting. Yeah I've not personally seen any complaints about the widget so it doesn't make sense to change it. I think our example is just a little weird probably!

mmarciniak90:
@lognaturel, @seadowg thanks!

**Tested with success**

Verified on Android 4.4, 6.0, 7.0, 8.1

kkrawczyk123:
*Tested with success!*
Verified on 4.2, 5.1, 9.0 and 10

I haven't noticed anything suspicious.

Verified cases:
All widgets in light mode
All widgets in dark mode
All widgets in RTL language (There are always weird things on Android 4.2)
All widgets in different font sizes
autoplay widget
grid widgets with autoplay
select one and multi: all kinds, no appearance, compact, column-flex
form with questions on OnePage
fieldlist form.

lognaturel (approved):

PR #3487, opened 2019-12-03
Support items from search function in RankWidget

### LINKED ISSUE #3486: The app crashes if items used in Rank widget come from search function
Collect v1.24.1 and older versions

#### Problem description
The app crashes if items used in Rank widget come from search function.

Try the attached form:
[search_and_select.zip](https://github.com/opendatakit/collect/files/3913242/search_and_select.zip)

grzesiek2010:
Closes #3486 

I tested the ranking widget manually and also added an automated test.

I refactored the code to allow using a list of `SelectChoices` in Ranking widget like other item widgets do. That's the simplest solution which is also consistent with other widgets.

This change is related just to the Ranking widget. We need to test it with items from the choices sheet (the one i All widgets form), and with items from an external csv file (the one attached below). That would be enough, nothing else should be affected.

[ranking_widget.zip](https://github.com/opendatakit/collect/files/3916480/ranking_widget.zip)

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
@@ -78,6 +78,13 @@ abstract class Page<T extends Page<T>> {
         return (T) this;
     }
 
+    public T checkIsTextDisplayed(String...  text) {
+        for (String s : text) {
+            onView(withText(s)).check(matches(isDisplayed()));
+        }
+        return (T) this;
+    }
+
     public T checkIsTranslationDisplayed(String... text) {
         for (String s : text) {
             try {
--- a/collect_app/src/androidTest/assets/forms/ranking_widget.xml
+++ b/collect_app/src/androidTest/assets/forms/ranking_widget.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0"?>
+<h:html
+	xmlns="http://www.w3.org/2002/xforms"
+	xmlns:ev="http://www.w3.org/2001/xml-events"
+	xmlns:h="http://www.w3.org/1999/xhtml"
+	xmlns:jr="http://openrosa.org/javarosa"
+	xmlns:odk="http://www.opendatakit.org/xforms"
+	xmlns:orx="http://openrosa.org/xforms"
+	xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+	<h:head>
+		<h:title>ranking_widget</h:title>
+		<model>
+			<instance>
+				<ranking_widget id="ranking_widget">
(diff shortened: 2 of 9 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
```diff
@@ -72,6 +72,17 @@ public T checkIsTextDisplayed(String text) {
         return (T) this;
     }
 
+    public T checkIsTextDisplayed(String...  text) {
+        for (String s : text) {
+            try {
+                onView(withText(s)).check(matches(isDisplayed()));
+            } catch (NoMatchingViewException e) {
+                Timber.i(e);
```
seadowg: This means the test won't fail if it can't find one of the items passed in right? Maybe we could rename the method to something like `checkIsTextDisplayedLoosley` and add a comment to make sure that's clear.

seadowg (changes requested):
Just one comment on a test helper. Looks good otherwise!

collect_app/src/androidTest/java/org/odk/collect/android/espressoutils/pages/Page.java
grzesiek2010: It should fail, my mistake, fixed.

seadowg: 👍

seadowg (approved):

lognaturel:
To be explicit here, I think this is definitely good to do but not critical for v1.25 because it hasn't been reported by users. Using external choices is usually done in a case where the choice list is long and that doesn't lend itself very well to ranking. Of course, the list can be filtered down by prior questions' values so it's still possible.

mmarciniak90:
@grzesiek2010 I noticed that filled answer is not visible if I open saved form. 

The second thing is crash when I try to fill form [rank-field-list.xml.txt](https://github.com/opendatakit/collect/files/3983213/rank-field-list.xml.txt). Exactly when I press 'Rank item' button

grzesiek2010:
@mmarciniak90 
I have fixed the bugs you have mentioned.

mmarciniak90:
@grzesiek2010 an answer is still not visible for saved submission for this particular form [filter-rank.xml.txt](https://github.com/opendatakit/collect/files/3987939/filter-rank.xml.txt)

EDIT!
I checked this case on 1.24.1 and also there the problem is visible so I'm gonna create a separate issue for this form

mmarciniak90:
**Tested with success**

Verified cases:
- rank widget on field-list
- read-only
- light dark mode
- English language
- RTL language
- saved form

PR #3490, opened 2019-12-04
[WIP] Likert analytics added

shouryaj:
Addition to issue #1272

Registers an event every time likert is used in a blank form. Event has the label "Likert" and its action is the form's analytics ID. 

Marking this as WIP because I need some help with actually making sure the event is logged! I could not see it in the console. 

Will be verified on review. 

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/LikertWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/LikertWidget.java
@@ -58,6 +58,8 @@ public class LikertWidget extends ItemsWidget {
         setButtonListener();
         setSavedButton();
         addAnswerView(view);
+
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel:
Looks good to me. Leaving unmerged in case @grzesiek2010 or @opendatakit/testers want to take a look first.

Here are a few things I wrote in Slack to explain the values passed:
> This leads me to realize that the category/action/label setup is better suited for the old Google Analytics and doesn't lead to very useful default reports in Firebase Analytics.

> I think you should use the category/action variant and log "Likert" as the category with the form identifier hash as the action.

> In Google Analytics we would have put the form identifier hash as a label but since there is no hierarchical relationship between action and label anymore, that's hard to analyze in Firebase.

grzesiek2010 (approved):

PR #3492, opened 2019-12-04
Reduce select and question header padding

### LINKED ISSUE #3491: Vertical spacing around question components and select labels is excessive
Collect v1.25.0-beta.0

#### Problem description
[In the v1.25 beta forum thread](https://forum.opendatakit.org/t/odk-collect-v1-25-beta-preview-of-increased-spacing-for-selects/23439), several users have shared that they find the new spacing excessive.

I see two options:
1) Reduce the default spacing
2) Introduce a setting to get compact spacing

lognaturel:
Closes #3491

Visual inspection and manual verification only. Options feel very easy to tap accurately.

| v1.25.0-beta.0 	| this PR     	|
|----------------	|-------------	|
|device-2019-12-03-190705 | device-2019-12-03-183621

I have done a lot of experimenting and research before settling on this proposed approach. @yanokwa has reviewed and is on board. I considered adding a setting to get a compact display but I'd like to keep that as a last resort. Adding settings adds to the maintenance burden because we then have more behavior combinations to verify and keep consistent. In this case, we'd be maintaining additional layouts. I'm hoping that this balanced approach will make users more comfortable.

When I did initial looking at other apps for #3413, I found 1-line lists with consistent 48dp heights. A lot of these lists have an icon at the left. Looking at other apps again, I see greater variation when it comes to select lists with radiobuttons/checkboxes on the left. That is not specifically called out on the [Material Design Lists page](https://material.io/components/lists/). In fact, somewhat ironically, there's an interactive demo at the top of that page with a radiobutton list with minimal vertical padding. That list style does not seem to be specified anywhere.

Settings dialogs in apps (including Collect -- see the font size setting for example) consistently have shorter rows. That is the spacing I have matched in this PR. It also matches the [material-ui radio button component](https://material-ui.com/components/radio-buttons/) which is very widely used.

I have also reduced the vertical padding for group labels, question labels and hint text. #3455 stated "To match these margins have also been added to question labels and group names" but unfortunately I did not notice that text nor the change because I was focused on getting the beta out. I think that seeing user screenshots, it's clear that the extra padding there is disruptive. There are no tappability concerns there so reducing the spacing back to how it was previously seems like the best course of action. I have made the spacing above group names a little bit larger than before (from extra_small to small) to balance things out, however.

While here, I increased the right padding for quick select icons for symmetry since that bothered me in one of the user screenshots.

I don't think there's any risk here other than users still finding the padding too large. If that happens, we can add a setting.

Any form with selects.

No.


seadowg:
I'm a fan of this compromise - it's definitely better than what we had in 1.24. I think it throws us a little away from the Material Specs which is frustrating. However, comparing what we have in the List Specs it actually makes a lot of sense to reduce the padding given our (standard) font size is much bigger than that used in the specs (16 vs 21). 

As we review the typography for widgets we can potentially reduce the font size which acutally lets us get back to standard margins. However, as we've discussed the standard Material fonts are "too damn small" a lot of the time so that might not work out either.

Apologies for not getting a chance to come back and dig into this and many thanks on doing it for me!

seadowg (approved):

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.2, 4.4, 5.1 6.0, 7.0, 8.1, 9.0 and 10.

lognaturel:
> I think it throws us a little away from the Material Specs which is frustrating. However, comparing what we have in the List Specs it actually makes a lot of sense to reduce the padding given our (standard) font size is much bigger than that used in the specs (16 vs 21).

What do you think about what I've pointed out about uses of lists with radiobuttons and checkboxes that have less padding? I don't actually think that the font size explains the difference. I haven't busted out any fancy rulers but looking at https://material-ui.com/components/radio-buttons/ it looks like less padding than what the list spec shows. Anyway, we can revisit. Luckily it's an easy enough change.

> Apologies for not getting a chance to come back and dig into this and many thanks on doing it for me!

No worries at all. You ended up with a lot on your plate and I'm glad I could get this resolved. I'm eager to see what users think.

seadowg:
> What do you think about what I've pointed out about uses of lists with radiobuttons and checkboxes that have less padding?

Oh I agree with you! I just mean that's it's frustrating to not be able to use the Material "Specs" section as our source of truth here. It definitely looks like they use a reduced padding version on the web. 

And to clarify, I think our font size is the difference between our lists and the spec lists (given we were following the same margin/padding size).

lognaturel:
> I think our font size is the difference between our lists and the spec lists (given we were following the same margin/padding size)

Ah, yes, agreed. But the spec list has excessive padding. 😄😭

PR #3503, opened 2019-12-06
Don't allow selecting basemap that is unavailable

### LINKED ISSUE #3497: Don't select Mapbox as a basemap source if it is unavailable
#### Problem description
When Mapbox is unavailable, a toast is shown to that effect. However, Mapbox is still selected as the source which is confusing. See https://forum.opendatakit.org/t/confused-about-current-state-of-offline-basemaps/23694.

1. Use an emulator
1. Change the map source to Mapbox and see the toast error that Mapbox is unavailable
1. Notice that Mapbox was still selected

#### Expected behavior
Attempting to change the source to Mapbox should fail and the previously-selected source should remain selected.

grzesiek2010:
Closes #3497 

I reproduced the issue and confirmed this pr fixes it.

It was strange that we could select a basemap that is unavailable.

Now if a basemap is unavailable we won't be able to select it.
The only thing we should test is changing basemaps in General Settings -> Maps.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/MapsPreferences.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/MapsPreferences.java
@@ -90,8 +90,14 @@ public class MapsPreferences extends BasePreferenceFragment {
         );
         onBasemapSourceChanged(MapProvider.getConfigurator());
         basemapSourcePref.setOnPreferenceChangeListener((pref, value) -> {
-            onBasemapSourceChanged(MapProvider.getConfigurator(value.toString()));
-            return true;
+            MapConfigurator cftor = MapProvider.getConfigurator(value.toString());
+            if (!cftor.isAvailable(context)) {
+                cftor.showUnavailableMessage(context);
+                return false;
+            } else {
+                onBasemapSourceChanged(cftor);
+                return true;
+            }
         });
     }
 
@@ -101,10 +107,7 @@ public class MapsPreferences extends BasePreferenceFragment {
         PreferenceCategory baseCategory = (PreferenceCategory) findPreference(CATEGORY_BASEMAP);
         baseCategory.removeAll();
         baseCategory.addPreference(basemapSourcePref);
-        if (!cftor.isAvailable(context)) {
-            cftor.showUnavailableMessage(context);
-            return;
-        }
+
         for (Preference pref : cftor.createPrefs(context)) {
             baseCategory.addPreference(pref);
         }


lognaturel (approved):

lognaturel:
I'm comfortable with this going in without a test because detangling to write non-integration tests is a big endeavor, this is currently part of QA and easy to verify, and I'm worried about diminishing returns in our integration test suite as it slows down.

@seadowg @OpenDataNerd This is related to the conversation y'all were having in Slack. Am I lacking imagination in how a unit test could be introduced? Do you feel strongly about having an integration test for now?

@grzesiek2010 Did you think about an automated test and what did you conclude?

breakbusyloop:
@ln it seems that the struggle lies in the singletons. Since much of the code accesses global state, we can't cover any of it with isolation tests. We would need to break out the logic into small helper methods that take a MapProvider implementation; within app logic that implementation is still the singleton, but for testing purposes we could write a test double 'UnavailableBaseMapProvider' that yields the weird behavior needed for this repro. We also need to break out the logic to determine what to do (show a message, handle the base map change) from the actual work. At some point you wind up with a model of the work to be done (very testable) and a controller for implementing that model.... which requires a whole lot of infrastructure code! I'm not very well versed in MVVM type of separations but I think in general those help with testing.

lognaturel:
I agree that what you've described would be an ideal state.

> which requires a whole lot of infrastructure code!

Exactly. I don't think we can do all that now as we're doing a bug fix. So I think our immediate options are add an integration test or keep relying on QA for now.

seadowg:
Totally agee with @OpenDataNerd. Understand if we have something high priority to fix we can strategically skip a rework like that but I think it'd be good to create an issue (with the "testing" label) before merge that either the PR creator or someone else can pick up. This, in my ever swayable opinion, seems to be a "confusing at worst" issue though so I'm not sure I'd see it at that level of priority. I do understand it is tempting to sneak this stuff in to a release however.

While thinking about the testing I was wondering if we could solve this in a different way. Would it not be possible to only show basemap options that are available? Maybe I'm wrong but it looks like `MapProvider` could filter within `getLableIds()` and `getIds()`.

lognaturel:
> create an issue (with the "testing" label) before merge

That works for me.

> seems to be a "confusing at worst" issue though so I'm not sure I'd see it at that level of priority

I did go back and forth on when it should go in. What ultimately made me feel like it should be treated as a bug fix is that a real user hit and report it. That usually means other folks are silently having the issue.

In this case, how many people actually use the emulator/an x86 device, etc, etc. If y'all would rather it wait to v1.26, I'm ok with that too.

> Would it not be possible to only show basemap options that are available?

I think we do want to be explicit about the fact that the option exists but is not available on the **current device**. We could add a more permanent warning/notice in the UI but I think a toast is a good balance of easy to implement and informative enough for something that won't be hit super frequently.

lognaturel:
Given the other high priority items that we have, let's bump this. Thanks for bringing up that it might not be so urgent, @seadowg.

I'd still tend to accept that this goes untested because it feels less dangerous than some of the other parts of the code base that aren't currently tested and I'd rather devote our limited capacity to those. What @OpenDataNerd has described at https://github.com/opendatakit/collect/pull/3503#issuecomment-562784929 is not a small task. But do push back if you disagree.

breakbusyloop:
@lognaturel I agree this can go in untested.

mmarciniak90:
I noticed one case, I think unlikely.

1. A user uses v1.25 on a device where Mapbox is not supported
2. Despite this, the user selects Mapbox 
3. User upgrades to a version with these changes
4. Nothing happens - crash if the user clicks on the new map icon on the `Fill Blank Form` list, toast is visible on GeoShape, GeoTrace widgets

@grzesiek2010 do you think that we can do something with this case? Is it worth?

grzesiek2010:
> do you think that we can do something with this case? Is it worth?

it's something doable but it would require adding code that practically would be used just once while updating from v1.25 to v1.26. I would vote for leaving it as is since in my opinion the game is not worth the :candle: at all :smiley:

lognaturel:
Agreed with @grzesiek2010! I'm reworking the `FormMapViewModel` based on review feedback and I think I may end up addressing that case, anyway so let's not do anything here.

mmarciniak90:
Thanks for your feedback.

**Tested with success**

Tested on an emulator and additionally on devices with Android 4.4, 5.1, 7.0, 9.0 to check possible regression

Verified cases
* set Mapbox from General Settings
* set Mapbox during filling a form
* English/Spanish language
* [upgrade](https://github.com/opendatakit/collect/pull/3503#issuecomment-567511721) - which does not work perfectly

PR #3511, opened 2019-12-09
Clear geometry after encryption

### LINKED ISSUE #3506: Geopoint for encrypted form is kept in cleartext in the instance database
Form encryption ensures that no one can get access to sensitive form data while it's on the device. Currently, the first geopoint from an encrypted form is stored in the instance database. When encryption happens on form finalization, that point should be removed.

### LINKED ISSUE #3510: Instance geometry is not saved if user has manually tapped save button before capturing it
#### Problem description
Discovered while addressing #3506. Geopoint is not saved and instance is not mapped if user has manually tapped save button before capturing point.

1. Open a form with a geopoint outside of any repeat (e.g. All Widgets)
1. Tap the save button
1. Collect a geopoint
1. Save and exit
1. Open the map view

The instance won't appear because only the first save attempted to read geometry.

#### Expected behavior
The instance should appear on the map.

lognaturel:
Closes #3506 
Closes #3510

I addressed these together because the "manually tap the save button" case described in #3510 requires knowing which instance row to save the geometry to and the solution to #3506 ended up providing that. 

Manually verified with [encrypted-geopoint.xml.txt](https://github.com/opendatakit/collect/files/3941409/encrypted-geopoint.xml.txt) that:
- when new instances are saved but not finalized, they appear on the map
- when instances are finalized, they disappear from the map and the display count updates accordingly

Manually verified with All Widgets that new instances that are saved manually with steps described in #3510 are updated on the map.

I really had to hold back from refactoring `updateInstanceDatabase` more significantly. I think this is the best given that we are working towards a release and really want to minimize risk.

I feel badly about not including a test. The only test I can think to write at this point would be an integration test over the encrypted-geopoint form linked above. I'd rather come back and do a lower-level test along with refactors if reviewers think that's ok.

This should only fix the two bugs described in #3506 and #3510. However, it does touch how every form instance which poses a risk. I tried to limit risk as much as possible by leaving existing code alone and adding to it. 

One particular type of risk is that if I didn't address error states, there could be additional crashes introduced (see https://github.com/opendatakit/collect/pull/3511#discussion_r355626021 for an example of one I almost didn't catch). There's a lot of state involved so it's hard to tell where there might e.g. be nulls.

[encrypted-geopoint.xml.txt](https://github.com/opendatakit/collect/files/3941409/encrypted-geopoint.xml.txt) 

It will be described as part of https://github.com/opendatakit/docs/issues/1144 which I'm working on.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
@@ -41,6 +41,9 @@ import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.dao.FormsDao;
 import org.odk.collect.android.dao.InstancesDao;
 import org.odk.collect.android.exception.EncryptionException;
+import org.odk.collect.android.instances.DatabaseInstancesRepository;
+import org.odk.collect.android.instances.Instance;
+import org.odk.collect.android.instances.InstancesRepository;
 import org.odk.collect.android.listeners.FormSavedListener;
 import org.odk.collect.android.logic.FormController;
 import org.odk.collect.android.provider.FormsProviderAPI.FormsColumns;
@@ -159,6 +162,17 @@ public class SaveToDiskTask extends AsyncTask<Void, String, SaveResult> {
         return saveResult;
     }
 
+    /**
+     * Updates the status and editability for the database row corresponding to the instance that is
+     * currently managed by the {@link FormController}. There are three cases:
+     * - the instance was opened for edit so its database row already exists
+     * - a new instance was just created so its database row doesn't exist and needs to be created
+     * - a new instance was created at the start of this editing session but the user has already
+     * saved it so its database row already exists
+     *
+     * Post-condition: the uri field is set to the URI of the instance database row that matches
+     * the instance currently managed by the {@link FormController}.
+     */
     private void updateInstanceDatabase(boolean incomplete, boolean canEditAfterCompleted) {
         ContentValues values = new ContentValues();
         if (instanceName != null) {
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -436,6 +453,20 @@ private void exportData(boolean markCompleted) throws IOException, EncryptionExc
...
+                // branches and because it has no knowledge of encryption status.
+                ContentValues values = new ContentValues();
+                values.put(InstanceColumns.GEOMETRY, (String) null);
+                values.put(InstanceColumns.GEOMETRY_TYPE, (String) null);
+                try {
+                    int updated = Collect.getInstance().getContentResolver().update(uri, values, null, null);
+                    if (updated < 1) {
+                        Timber.w("Instance geometry not cleared after encryption");
+                    }
+                } catch (IllegalArgumentException e) {
```
lognaturel, line 473: This could happen if the `uri` wasn't properly set to the instance's in one of the two cases where `FormEntryActivity` was called with a URI that represents a form definition (from Fill Blank Form, e.g., as opposed to Edit Saved Form).

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -183,28 +197,31 @@ private void updateInstanceDatabase(boolean incomplete, boolean canEditAfterComp
...
             String where = InstanceColumns.INSTANCE_FILE_PATH + "=?";
-            String[] whereArgs = {
-                    instancePath
-            };
-            int updated = new InstancesDao().updateInstance(values, where, whereArgs);
+            int updated = new InstancesDao().updateInstance(values, where, new String[] {instancePath});
+
+            // Set uri to handle encrypted case (see exportData)
+            InstancesRepository instances = new DatabaseInstancesRepository();
+            Instance instance = instances.getByPath(instancePath);
```
lognaturel, line 219: I decided to introduce `getByPath` here because we'll definitely need it and it allowed me to abstract away the logic related to cursors and not getting exactly one result back (though I still have to handle the null case here).

lognaturel:
@grzesiek2010, @seadowg If you have changes you'd like to see here, please consider pushing to this branch and working together to get in a state you're happy with so it can be complete during your day. Unless something else comes back from regression testing, I think this is highest priority.

seadowg (approved):
I can't see anything wrong with this but it'd be good for @grzesiek2010 to have a look as well as it does seem pretty sensitive. 

@lognaturel Would you be able to file an issue around the refactoring and testing? The entire save process is pretty hard to get tests around at the moment. I think this has overlap with other `FormEntryActivity` related refactors.

grzesiek2010 (approved):

mmarciniak90:
**Tested with success**

Verified on Android: 4.2, 4.4, 6.0, 7.0, 8.1, 9.0, 10.0

Verified cases:
* Case 1:
    1. Open a form with a geopoint outside of any repeat (e.g. All Widgets)
    2. Tap the save button
    3. Collect a geopoint
    4. Save and exit
    5. Open the map view
* Case 2:
    1. When new instances are saved but not finalized, they appear on the map
    2. When instances are finalized, they disappear from the map and the display count updates accordingly
* changing device orientation
* correct value on 'Saved Forms' bar
* all form statuses presented as markers

lognaturel:
Thanks so much, all! I'm really glad we caught the encryption issue now and aren't leaking sensitive data. I noticed it while writing documentation and thinking through what a user might do so that's an extra reason to write documentation before a release.

> Would you be able to file an issue around the refactoring and testing? The entire save process is pretty hard to get tests around at the moment.

Absolutely. I made a note in https://github.com/opendatakit/collect/issues/3461 for now. I'd like to discuss high level strategy and then file more specific issues if needed. I'm not exactly sure of the scope of what we can/should do here.  Some things I've thought of quickly turn into "and now we rewrite everything."

Issue #3335, opened 2019-09-06
The search()/pulldata() handler consumes files meant to be used as external secondary instances


lognaturel:
To reproduce this issue, #3334 first needs to be addressed.

Then, try a form like [this one](https://docs.google.com/spreadsheets/d/1xZDS2XqwbMcgqa5nvi5_Ug03WkijogcQh4yPNexqaE8/edit#gid=0) that attempts to use a CSV file as an external secondary instance. It works the first time the form is opened but fails the second time because the CSV has been imported to a db by the search()/pulldata() machinery and renamed.

Some options I see to address this:
- instead of renaming CSVs to mark them as ingested, leave the files untouched and use a file hash to detect a change
- instead of greedily putting all CSVs in a DB, only read in the ones used for search()/pulldata() (or not used as external secondary instance)

breakbusyloop:
I barely have a sense of what's going on here, but I see enough to have a strong opinion. Using a secondary characteristic (file name change) as a source of primary information has the same pitfalls as the use of sentinels, and will forever be a bug factory.

I really prefer the first option:
 * instead of renaming CSVs to mark them as ingested, leave the files untouched and use a file hash to detect a change

lognaturel:
> Using a secondary characteristic (file name change) as a source of primary information has the same pitfalls as the use of sentinels, and will forever be a bug factory.

➕➕➕

I agree that the first option would be much better.

You're getting into some really head-scratching territory here. You'll see that search()/pulldata() is a different way of accessing data in files external to the form definition as what is described in #3334.

https://forum.opendatakit.org/t/external-data-current-state-of-affairs should hopefully provide some context.

breakbusyloop:
It looks like the CSV file renaming happens in ExternalDataReaderImpl.doImport. The file will be reimported if it shows up again with the same name. After we stop renaming files, we could reimport based on a file hash (which requires reading the whole file, right?), or we could reimport based on simpler attributes like the file timestamp.

lognaturel:
Yes, the hash does require reading through the whole file. See `FileUtils.getMd5Hash` for how that's currently done. I think it's ok to do that work since there isn't particularly a need for high performance when downloading a new form version. But I can't see a reason why file timestamp wouldn't be sufficient in this case.

breakbusyloop:
[...]

PR #3518, opened 2019-12-10
Enable external CSV secondary instances (issue #3334) 

### LINKED ISSUE #3334: CSV external secondary instances don't work because the jr://file-csv scheme is undefined
Add `file-csv` as a session root translator in `FormLoaderTask`. It needs to go before `file` because `file` is a substring.

https://github.com/opendatakit/javarosa/files/3095508/external_csv.zip is a form for testing. See https://docs.opendatakit.org/collect-forms#loading-form-media for loading form media. If pushing the form directly to a device, a folder with name `<form_id>-media` needs to be created and populated with the media.

breakbusyloop:
Closes #3334

gradlew testDebug
gradlew connectedAndroidTest

The only solution to the issue was to add an extra host string to handle "file-csv". The tests seemed to require a fair bit of state (CopyFormRule, global singletons) so they are mostly integration tests. I'm open to discussing ways to make the tests more narrow, but the implementation itself is not and would require more refactoring to support true unit tests.

This enables users to work with secondary instances defined by CSV files. The regression risk is clear: it exposes issue #3335.

Same form as is listed in issue #3334: https://github.com/opendatakit/javarosa/files/3095508/external_csv.zip

.... maybe?

- [ X ] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [ X ] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [ n/a ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
@@ -4,7 +4,6 @@ import android.Manifest;
 
 import androidx.test.rule.GrantPermissionRule;
 
-import org.javarosa.core.reference.RootTranslator;
 import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
@@ -12,11 +11,9 @@ import org.junit.rules.RuleChain;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.support.CopyFormRule;
 import org.odk.collect.android.support.ResetStateRule;
-import org.odk.collect.android.utilities.FileUtils;
 
 import java.io.File;
 import java.util.Arrays;
-import java.util.List;
 
 public class FormLoaderTaskTest {
     private static final String EXTERNAL_CSV_FORM = "external_csv_form.xml";
--- a/collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
@@ -19,7 +19,6 @@ import java.util.Arrays;
 import java.util.List;
 
 public class FormLoaderTaskTest {
-    private static final String BASIC_FORM = "basic.xml";
     private static final String EXTERNAL_CSV_FORM = "external_csv_form.xml";
 
     @Rule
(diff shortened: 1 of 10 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
```diff
@@ -0,0 +1,38 @@
...
+            .around(new ResetStateRule())
+            .around(new CopyFormRule(EXTERNAL_CSV_FORM,
+                    Arrays.asList("external_csv_cities.csv", "external_csv_countries.csv", "external_csv_neighbourhoods.csv")));
+
+    @Test
+    public void loadFormWithSecondaryCSV() throws Exception {
+        final String formPath = Collect.FORMS_PATH + File.separator + EXTERNAL_CSV_FORM;
+        FormLoaderTask formLoaderTask = new FormLoaderTask(formPath, null, null);
+        FormLoaderTask.FECWrapper wrapper = formLoaderTask.execute(formPath).get();
+        Assert.assertNotNull(wrapper);
```
lognaturel, line 36: We've generally written these kinds of Espresso form loading tests to display the form and assert something on the form. See `ExternalSecondaryInstancesTest` for an example with an XML external secondary instance. That makes it more end-to-end for that form definition.

I think you were (rightfully) looking for a way to make the test more targeted, right? The wrapper being null is a proxy for the task failing?

I wonder whether since we don't have any tests for secondary CSV instances and instrumented tests are expensive anyway we should go all the way and verify that the form entry can be done correctly.

lognaturel (commented):
I am left wondering what I saw that made me think there was an ordering issue with the roots. I remember trying it before I wrote the issue and getting a failure but I guess it must have been related to something else. Maybe I forgot to install when I put the `file-csv` last so I was actually getting the behavior without the root definition at all. Thanks for adding tests!

Question inline about whether perhaps `FormLoaderTaskTest` should become `ExternalCSVSecondaryInstanceTest` and test the form display too.

@seadowg or @grzesiek2010 should also review the CSV header row change.

collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
breakbusyloop: Yes, I was trying to make a targeted test but I didn't want to make big refactoring changes to FormLoaderTask to do so. With tests like this in place we can proceed to make the refactoring changes needed to make the test even smaller. I'd rather go that route than make the test bigger to verify form entry. But I understand how hijacking an intermediate loading stage for this testing purpose is a bit deviant :)

breakbusyloop:
> @seadowg or @grzesiek2010 should also review the CSV header row change.

Note that change addresses a recently introduced failure on Windows that was unrelated to my CSV work, but blocked me. Two issues, actually: 1) use File.separator instead of hard coded forward slash in "/temporaryAudit.csv"  2) instead of populating a temporary file and moving it on top of the original, I rename the original and populate a new file with the same name as the original.

In the future should I file an issue for such breakages and target my fix just to that issue?

collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
lognaturel: > a bit deviant :)

Was trying not to say it that way but yes... ;)

lognaturel:
> In the future should I file an issue for such breakages and target my fix just to that issue?

Yes, that would be preferable in a case like this where there are no conflicts between the two because then the review and QA can be more targeted. Not a big deal.

lognaturel (approved):

mmarciniak90:
@OpenDataNerd I need your help because I'm not sure what result is expected. 

I checked the `external_csv_form` form and media used in an automated test. 

Result for a version without these changes is 


When I opened the same form on a version with these changes result is:


I see a difference in the message but is it the main change?

The second thing is a problem with an attached form for testing. I made small modifications, I created a folder for media and renamed form.
[external_csv.zip](https://github.com/opendatakit/collect/files/4038942/external_csv.zip) 
I copied it to `forms` directory and when I opened the `Fill Blank Form` page I saw an error:
Screenshot (Jan 9, 2020 07_32_40)

I checked also without name modification and the result is the same
[external_csv_v2.zip](https://github.com/opendatakit/collect/files/4038986/external_csv_v2.zip)

breakbusyloop:
@mmarciniak90 those errors are expected until the following issue is addressed:
https://github.com/opendatakit/collect/issues/3335

I will post a PR for that fix once this one has been merged.

lognaturel:
Ah yes, we should have been more clear about the expected QA results. I'm going to merge this because the change in the path in the error is good and because it will be much easier to QA all of the new behavior with @OpenDataNerd's PR for #3335.

mmarciniak90:

PR #3523, opened 2019-12-12
Set up the reference manager when doing form discovery parse

### LINKED ISSUE #3522: Forms with external secondary instances fail to parse on copy or download
Collect v1.25

#### Problem description
Forms with external secondary instances fail to parse on copy or download. Discovered while reviewing #3518. This issue has come up because the XForm is now parsed on ingestion and parsing a form def with an external secondary instance requires that external secondary instance file to actually exist as described in https://github.com/opendatakit/javarosa/issues/444. In an ideal world, we wouldn't include it.

1. push or download a form with an external secondary instance such as 
[external-csv.zip](https://github.com/opendatakit/collect/files/3952973/external-csv.zip). This one requires CSVs so will need #3518.
1. If downloading, notice error dialog.
1. If pushing, go to Fill Blank Form and notice error toast

#### Expected behavior
Form is ingested as before the change to parse the form on ingestion.

lognaturel:
Closes #3522

Copied and downloaded forms with external secondary instances and verified they could be filled.

Tried with external XML instance ([external-select-xml.zip](https://github.com/opendatakit/collect/files/3953809/external-select-xml.zip)) and CSV instance ([external-csv.zip](https://github.com/opendatakit/collect/files/3952973/external-csv.zip)).

The true fix would be not requiring the presence of those secondary instance files to build a `FormDef` because in this case there's no point to parsing them and it just slows things down. However, that would be a JR change with sweeping consequences. The change in this PR is safe and it's easy to verify that it does nothing in the common case so it's the best change for now.

This should only make forms with external secondary instances work and should have no other effect.

external XML instance ([external-select-xml.zip](https://github.com/opendatakit/collect/files/3953809/external-select-xml.zip)) and CSV instance ([external-csv.zip](https://github.com/opendatakit/collect/files/3952973/external-csv.zip))

No.

CC @OpenDataNerd

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/DiskSyncTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/DiskSyncTask.java
@@ -20,6 +20,7 @@ import android.database.SQLException;
 import android.net.Uri;
 import android.os.AsyncTask;
 
+import org.javarosa.core.reference.ReferenceManager;
 import org.odk.collect.android.R;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.dao.FormsDao;
@@ -36,6 +37,8 @@ import java.util.LinkedList;
 import java.util.List;
 
 import timber.log.Timber;
+
+import static org.odk.collect.android.forms.FormUtils.setupReferenceManagerForForm;
 
 /**
  * Background task for adding to the forms content provider, any forms that have been added to the
@@ -257,6 +260,10 @@ public class DiskSyncTask extends AsyncTask<Void, String, String> {
 
         HashMap<String, String> fields;
         try {
+            // If the form definition includes external secondary instances, they need to be resolved
+            final File formMediaDir = FileUtils.getFormMediaDir(formDefFile);
+            setupReferenceManagerForForm(ReferenceManager.instance(), formMediaDir);
+
             fields = FileUtils.getMetadataFromFormDefinition(formDefFile);
         } catch (RuntimeException e) {
             throw new IllegalArgumentException(formDefFile.getName() + " :: " + e.toString());
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
@@ -20,6 +20,7 @@ import android.content.ContentValues;
 import android.database.Cursor;
 import android.net.Uri;
 
+import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.xform.parse.XFormParser;
 import org.kxml2.kdom.Element;
 import org.odk.collect.android.R;
@@ -45,6 +46,8 @@ import java.util.Map;
 import javax.inject.Inject;
 
 import timber.log.Timber;
+
+import static org.odk.collect.android.forms.FormUtils.setupReferenceManagerForForm;
 
 public class FormDownloader {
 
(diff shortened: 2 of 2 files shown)


lognaturel:
This all definitely doesn't make me feel warm and toasty inside and really illustrates the challenge of needing to handle such a broad range of form definitions. I would never have thought to try or introduce automated tests for the combination of form discovery and external secondary instances, it was just luck that I noticed the problem when reviewing #3518. Conceptually, we don't want anything to do with secondary instances in that initial discovery parse so it doesn't make sense to do anything special with them. But because of https://github.com/opendatakit/javarosa/issues/444, we can't get a `FormDef` without them. Luckily for this issue the stakes were low -- someone working on a new form with an external secondary instance would have hit the problem and hopefully reported it and there was no risk of data loss.

seadowg (approved):
This looks like the right fix to me. 

We actually do have tests for external secondary instances but I don't they caught this because we manually add forms to the database as part of test setup. Because of that we recently had to add this `setupReferenceManager` to that form copying code. I'll have a look at our test setup code and see if I can get to a place where we would have had a correctly failing test (rather than one that failed at the setup level) to hopefully follow this up with.

mmarciniak90:
@lognaturel please look at my results. I'm not sure that everything is ok

| Collect version | form external-select-xml | form external-csv|
| --- |---|---|
|v 1.24.1 | I'm able to fill the form | error when form is loaded Screenshot_20191212-125454|
|v 1.25.0 beta.2 |Screenshot_20191212-125611|Screenshot_20191212-125634|
|version from pr/3523| I'm able to fill the form | Screenshot_20191212-125905|

seadowg:
@mmarciniak90 was just playing around with it and got the same thing. The error looks strange though. Going to check I've definitely got the files in the right place and report back after lunch.

breakbusyloop:
@seadowg @mmarciniak90 can you examine your ODK folder on the device (or emulator)? Please let me know if the missing CSV file is in the form '-media' folder but has been renamed with ".imported". If so, then I think the error message may come from issue 3335.

mmarciniak90:
@OpenDataNerd here's what my files look like

breakbusyloop:
@mmarciniak90 thanks, it appears that the error you were seeing is not related to issue 3335 after all.

seadowg:
@OpenDataNerd yeah mine is fine as well.

From the message it looks to me like it's looking in the wrong place for the file (`odk-csv`). I'm confused as I just managed to get a red/green test cycle working using `ExternalSecondaryInstancesTest` and @lognaturel's changes.

lognaturel:
I’m sorry, I explained in the issue that the CSV form requires adding the file-csv root but didn’t repeat here. That error is expected. The XML one should work without changes.

lognaturel:
> it looks to me like it's looking in the wrong place for the file (odk-csv)

That's because it's matching the `file` prefix in the absence of the `file-csv` prefix which will get added by #3518. I discovered the issue on that PR which is why I confused things with CSVs.

> Because of that we recently had to add this setupReferenceManager to that form copying code.

Sadly that failure was a real failure that pointed to this very problem but I didn't notice it go by, somehow! I think my local tests were failing for other reasons and then when master broke, https://github.com/opendatakit/collect/pull/3480 went by and I didn't see it. Let's make sure in the future that if a PR breaks the build that PR's author is involved in the fix!

> as I just managed to get a red/green test cycle working using ExternalSecondaryInstancesTest and @lognaturel's changes

That one only looks at XML external secondary instances so all is working as expected.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success**
I have confirmed the behavior on Androids: 4.2, 4.4, 5.1, 6.0, 7.1, 8.1 and 10.

PR #3536, opened 2019-12-13
Use renamed isAncestorOf() method

grzesiek2010:
Tested collect with javarosa's master branch.

in https://github.com/opendatakit/javarosa/pull/515/files#diff-0c84c5364a094f70c8aa2c2ca3376a1a
the method isParentOf() has been renamed to isAncestorOf(). We use it in Collect so **we need to merge this pr one we add javarosa with the mentioned fix**.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
@@ -560,7 +560,7 @@ public class FormHierarchyActivity extends CollectAbstractActivity {
                 // retrieve the current group
                 TreeReference curGroup = (visibleGroupRef == null) ? contextGroupRef : visibleGroupRef;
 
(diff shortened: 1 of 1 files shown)


seadowg (approved):
@grzesiek2010 am I right in thinking we can't merge this until the next JavaRosa release but that if we don't our app will fail to compile (as `isParent` won't exist)?

lognaturel:
> we can't merge this until the next JavaRosa release

Correct. I wonder whether it might be easier to re-introduce this when the JR update happens so we don't risk merging something broken.

> our app will fail to compile (as isParent won't exist)

This was an accidental API change and `isParentOf` will be re-introduced as deprecated: https://github.com/opendatakit/javarosa/issues/519. Using `isAncestorOf` will certainly be preferable.

grzesiek2010:
> I wonder whether it might be easier to re-introduce this when the JR update happens so we don't risk merging something broken.

Ok so I'm closing for now. I added it to my private todo list so I won't forget to reopen.

PR #3537, opened 2019-12-13
On form download, only set up the reference manager if the form has media

### LINKED ISSUE #3535: Unable to download forms without media attachments
Collect current master.x, Android v7, 9 probably all.x.x, device used...

#### Problem description
The error message is displayed when try to download form from Central:
Screenshot_20191213-161532

1. Import settings via QR code from Central to Collect.
2. Go to Get Blank Form
3. Select form from the displayed list
4. Click on Get Selected

#### Expected behavior
The form should be successfully downloaded.

#### Other information 
!! The issue does not occur on beta, only on current master

lognaturel:
Closes #3535

Ran the automated tests. Downloaded forms with and without media from the default Aggregate server.

The fix itself is a null check and I think that's the only possible option.

The tests I don't love but I couldn't come up with any other option without risky refactoring. They use Robolectric because of `FormDownloader`'s reliance on the `Collect` singleton for dependency injection, for looking up strings and for database access (this class, like many in this code base, has all the concerns). I mocked the actual download of files and tried to make it really clear in the test comments what the tests target.

I made the mistake of not thinking about this question carefully enough in #3523. This code is involved in the downloading of all forms from any server so that's where the risk is. There are different branches based on whether there is or isn't media attached and both should be verified.

Forms on https://opendatakit.appspot.com

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
@@ -204,7 +204,7 @@ public class FormDownloader {
         return submission == null || Validator.isUrlValid(submission);
     }
 
-    private boolean installEverything(String tempMediaPath, FileResult fileResult, Map<String, String> parsedFields) {
+    boolean installEverything(String tempMediaPath, FileResult fileResult, Map<String, String> parsedFields) {
         UriResult uriResult = null;
         try {
             uriResult = findExistingOrCreateNewUri(fileResult.file, parsedFields);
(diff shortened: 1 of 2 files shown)


lognaturel:
CC @OpenDataNerd, you may have some thoughts on the tests and whether there might be alternate ways to structure them.

seadowg (approved):
There are some things in the tests I'm not a fan of (that you've pointed out) - the spy usage and this having to use Robolectric. However, I think the test is at the right "level" here and that these nasty parts are just exposing the problems in the objects design. I think this is good to go and is a great candidate for a little clean up work. I'll add an issue!

breakbusyloop:
> you may have some thoughts on the tests and whether there might be alternate ways to structure them.

I added my thoughts to #3538

grzesiek2010 (approved):

mmarciniak90:
**Tested with success**

Central:
- download form without media
- download form with media
- send submission with attachments
- send submission without attachments
- downloading a few forms during at once

Local Aggregate v2.0:
- download form without media
- download form with media
- verified Aggregate which accepts anonymous user and does not accept

Default Aggregate v1.7.4:
- download all available forms
- send submission with attachments
- send submission without attachments

PR #3541, opened 2019-12-14
Rework regex with tests

### LINKED ISSUE #477: Put regex in meaningfully-named methods and introduce unit tests
See the discussion below for the suggested approach. The goal is to give meaning to the regex and test them.

breakbusyloop:
Closes #477

ran tests

Following the commentary in #477, the regex was not relocated away from the relevant code unless it was either 1) necessary for testing purposes, or 2) exposed useful logic for other parts of the code. The method names exposed for testing attempt to convey the regex matching criteria.

No change

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
@@ -31,6 +31,7 @@ import org.odk.collect.android.database.helpers.FormsDatabaseHelper;
 import org.odk.collect.android.provider.FormsProviderAPI.FormsColumns;
 import org.odk.collect.android.utilities.FileUtils;
 import org.odk.collect.android.utilities.MediaUtils;
+import org.odk.collect.android.utilities.UriUtils;
 
 import java.io.File;
 import java.util.HashMap;
@@ -530,10 +531,10 @@ public class FormsProvider extends ContentProvider {
     // https://developer.android.com/reference/android/content/UriMatcher
     static {
         URI_MATCHER = new UriMatcher(UriMatcher.NO_MATCH);
-        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, FormsColumns.CONTENT_URI.getPath().replaceAll("^/+", ""), FORMS);
-        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, FormsColumns.CONTENT_URI.getPath().replaceAll("^/+", "") + "/#", FORM_ID);
+        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, UriUtils.stripLeadingUriSlashes(FormsColumns.CONTENT_URI.getPath()), FORMS);
+        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, UriUtils.stripLeadingUriSlashes(FormsColumns.CONTENT_URI.getPath()) + "/#", FORM_ID);
         // Only available for query and type
-        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, FormsColumns.CONTENT_NEWEST_FORMS_BY_FORMID_URI.getPath().replaceAll("^/+", ""), NEWEST_FORMS_BY_FORM_ID);
+        URI_MATCHER.addURI(FormsProviderAPI.AUTHORITY, UriUtils.stripLeadingUriSlashes(FormsColumns.CONTENT_NEWEST_FORMS_BY_FORMID_URI.getPath()), NEWEST_FORMS_BY_FORM_ID);
 
         sFormsProjectionMap = new HashMap<>();
         sFormsProjectionMap.put(FormsColumns._ID, FormsColumns._ID);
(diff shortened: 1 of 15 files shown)


collect_app/src/test/java/org/odk/collect/android/utilities/FormNameUtilsTest.java
```diff
@@ -40,4 +40,21 @@ public void normalizeFormNameTest() {
         assertEquals("Lorem  ipsum", normalizeFormName("Lorem\n\nipsum", true));
         assertEquals(" Lorem ipsum ", normalizeFormName("\nLorem\nipsum\n", true));
     }
+
+    @Test
+    public void formatFilenameFromFormNameTest() {
+        assertNull(FormNameUtils.formatFilenameFromFormName(null));
```
lognaturel: I've typically favored hamcrest `assertThat` with matchers so assertions read more like sentences and I believe @seadowg and @grzesiek2010 typically do the same:
- `assertThat(FormNameUtils.formatFilenameFromFormName(null), is(nullValue()));`
- `assertThat(FormNameUtils.formatFilenameFromFormName("simple"), is("simple"));`

I don't feel strongly about imposing a standard here because these aren't particularly hard to read but at least wanted to make sure you knew about those tools.

collect_app/src/test/java/org/odk/collect/android/utilities/FormNameUtilsTest.java
```diff
@@ -20,9 +20,9 @@
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
-import static org.odk.collect.android.utilities.RegexUtils.normalizeFormName;
+import static org.odk.collect.android.utilities.FormNameUtils.normalizeFormName;
```
lognaturel, line 24: You might also consider a static import for `formatFilenameFromFormName` to cut down the noise.

collect_app/src/main/java/org/odk/collect/android/widgets/BarcodeWidget.java
```diff
@@ -88,14 +89,20 @@ public IAnswerData getAnswer() {
...
-            response = response.replaceAll("\\p{C}", "");
-        }
-        stringAnswer.setText(response);
+        stringAnswer.setText(stripInvalidCharacters(response));
 
         widgetValueChanged();
     }
 
+    // Remove control characters, invisible characters and unused code points.
+    @Contract("null -> null; !null -> !null")
```
lognaturel, line 98: Ooh, nice!

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -530,10 +531,10 @@ public int update(Uri uri, ContentValues values, String where,
     // https://developer.android.com/reference/android/content/UriMatcher
```
lognaturel, line 531: Remove comment from here?

lognaturel (approved):
Small take-it-or-leave-it comments inline. Thanks!

breakbusyloop:
Thanks, I'll address these comments and post another change.

collect_app/src/main/java/org/odk/collect/android/widgets/BarcodeWidget.java
breakbusyloop: The IDE recommended that. I'm curious to see how it helps with code cleanliness - maybe it can eliminate some null pointer checks?

collect_app/src/test/java/org/odk/collect/android/utilities/FormNameUtilsTest.java
breakbusyloop: Thanks for the guidance. I initially tripped myself up on the JUnit 'assertThat'. Switching to hamcrest gave me what I need.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
breakbusyloop: Or maybe I should remove the comment from UriUtils. The method name explains 'what' we are trying to do, but the comment explains 'why'. The 'why' part seems like it should be kept closer to the decision point, i.e. in FormsProvider. So should I remove the identical comment from UriUtils instead?

lognaturel: Ah, yes, I see. I agree that what you suggest makes more sense. 👍

collect_app/src/test/java/org/odk/collect/android/utilities/FormNameUtilsTest.java
lognaturel: Aha! Ok, I think this might be worth a note in the contributor guide. I forgot there was that ambiguity.

breakbusyloop: I can add that note, but it seems more appropriate for the Coding Guidelines doc.

lognaturel: Good point! It just got split out so I forgot. 😊

lognaturel: Apparently I forget a lot of things...

breakbusyloop: That's why we document them, so we can forget!

lognaturel (approved):
I extra approve.

PR #3542, opened 2019-12-16
Update December translations

lognaturel:
Routine translation update before release.


seadowg (approved):

seadowg:
All looked good to me. Just checked that all the same strings looked like they had been translated.

PR #3544, opened 2019-12-17
Qr code

### LINKED ISSUE #2993: User should be informed then scanned QR code does not contain settings
#### Problem description
User should be informed then scanned QR code does not contain settings 

1. User goes to `Admin Settings`
2. User goes to `Import/Export settings`
3. User clicks on `Scan code from other device`
4. User scans QR code which does not contain ODK settings

#### Expected behavior
* User is informed that QR code does not contain settings
* When user selects an incorrect file from SD card, toast is visible
Screenshot_20190404-103532

ShivamPrajapati-dev:
Closes #2993 

Checked with different QR-codes.
If we scan the wrong QR-code the two types of exceptions are thrown one is DataFormatException and the second is IllegalArgumentException. DataFormatException while scanning wrong QR-code and with valid Base64 encoding and IllegalArgumentException with invalid encoding or arguments to the method.
No.
No.


seadowg:
@ShivamPrajapati-dev it looks like this contains commits from #3408. It's a good idea to start a new branch (off `master`) for each PR. 

As far as I can see there aren't any changes to code to fix the issue (just some test method signature updates). I'll close this for the moment but let me know if I've missed something!

PR #3545, opened 2019-12-17
Add guideline for testing using Hamcrest

breakbusyloop:
@ln small documentation update per your suggestion to test using Hamcrest matchers.


lognaturel (approved):
🎉

lognaturel:
Merged now even though we're still frozen since it's docs-only.

PR #3549, opened 2019-12-18
Handle the case when a geopoint question is not relevant

### LINKED ISSUE #3548: Invalid index 0, size is 0 exception is thrown if a user tries to save a form with a geopoint question which is not releavant
Collect v1.25.0

#### Problem description
Invalid index 0, size is 0 exception is thrown if a user tries to save a form with a geopoint question which is not relevant.

Originally reported on the forum: https://forum.opendatakit.org/t/problem-odk-collect-v1-25/23963

1. Open the attached form: 
[geoRelevant.xlsx](https://github.com/opendatakit/collect/files/3980009/geoRelevant.xlsx)
[geoRelevant.xml.txt](https://github.com/opendatakit/collect/files/3980013/geoRelevant.xml.txt)
It consists of two question:
- select _one yes/no
- geopoint question which is relevant only if a user select yes in the first question

2. Select no in the first question.
3. Try to save the form.

#### Expected behavior
The form should be successfully saved. The geo question is not relevant so related columns in instances.db should not be filled (geometry and geometryType).

grzesiek2010:
Closes #3548 

I was able to reproduce the issue using the form attached to the issue and confirm that this pr fixes the problem.

We just need to handle a case when a geopoint question is not relevant and that's all that's the only solution.

It's a safe change. We just need to test the form attached to the issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
@@ -292,6 +292,11 @@ public class SaveToDiskTask extends AsyncTask<Void, String, SaveResult> {
             Object result = expr.eval(instance, context);
             if (result instanceof XPathNodeset) {
                 XPathNodeset nodes = (XPathNodeset) result;
+                if (nodes.size() == 0) {
+                    Timber.i("TreeElement is missing for xpath %s!, probably it's just not relevant", xpath);
+                    return null;
+                }
+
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Since it's a very safe change, we have clear reproduction steps, and we have a user who is blocked, I'm inclined to merge and do a point release now. Any objections?

lognaturel:
I'm doing some refactors around some of the `FormMap` feature and will look at adding some testing around this behavior.

PR #3550, opened 2019-12-19
Fixed IndexOutOfBoundsException in FormMapActivity.java:65

### LINKED ISSUE #3546: IndexOutOfBoundsException in FormMapActivity.java:65
Collect v1.25.0

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/22688dd6b5390b11013bfb38148dedcd?time=last-seven-days&sessionId=5DF8EBF4000700012C1E0AD753D1A008_DNE_1_v2

`Caused by java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
       at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
       at java.util.ArrayList.get(ArrayList.java:308)
       at org.odk.collect.android.activities.FormMapActivity.onCreate(FormMapActivity.java:65)
       at android.app.Activity.performCreate(Activity.java:6609)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1134)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3104)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3261)
       at android.app.ActivityThread.access$1000(ActivityThread.java:219)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1735)
       at android.os.Looper.loop(Looper.java:145)`

grzesiek2010:
Closes #3546

Tested FormMapActivity manually.

I wasn't able to reproduce the issue so I just implemented changes to handle such a case, hopefully it's something rare and random.

It's a safe change that is related just to the `FormMapActivity`. So starting the activity and checking if saved forms are visible on the map should be enough.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -30,8 +30,6 @@ import androidx.lifecycle.ViewModelProviders;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.activities.viewmodels.FormMapViewModel;
-import org.odk.collect.android.dao.FormsDao;
-import org.odk.collect.android.forms.Form;
 import org.odk.collect.android.geo.MapFragment;
 import org.odk.collect.android.geo.MapPoint;
 import org.odk.collect.android.geo.MapProvider;
@@ -61,10 +59,8 @@ public class FormMapActivity extends BaseGeoMapActivity {
     @Override public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        FormsDao dao = new FormsDao();
-        Form form = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData())).get(0);
         viewModel = ViewModelProviders.of(this,
-                new FormMapActivity.FormMapViewModelFactory(form, new DatabaseInstancesRepository())).get(FormMapViewModel.class);
+                new FormMapActivity.FormMapViewModelFactory(getIntent().getData(), new DatabaseInstancesRepository())).get(FormMapViewModel.class);
         Timber.i("Starting FormMapActivity for form \"%s\" (jrFormId = \"%s\")",
                 viewModel.getFormTitle(),
                 viewModel.getFormId());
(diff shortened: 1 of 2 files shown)


lognaturel (changes requested):
This breaks the tests, can you please tell me more about why you checked the first item in the PR checklist? It seems to happen relatively frequently that folks check this but haven't run the tests so I'm curious to know more about how it happens.

Given that the only activity that can launch `FormMapActivity` by intent is `FormChooserListActivity` and that it has to have accessed the form record this seems very unlikely to be common. If it is common there's something we don't understand that we need to understand. For those reasons, I'd tend to monitor this one and not do anything for now.

We definitely don't want a `FormsDao` object being constructed in the view model constructor because that creates a dependency on the Android framework. We could pass in a `FormDao` like we do an `InstancesRepository` but I made a deliberate decision to do the resolution of a form URI to a form object outside of the view model to separate concerns. If you feel strongly about getting this in even though we've only seen the crash once for now, I think that we should do `dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData()))` and check the length before getting the first item. Closing the activity if the size is not exactly 1 is fine as we try to figure out what's going on.

If you're not sure why constructing the `FormsDao` in the constructor is a problem, I highly recommend doing a little reading on the dependency injection design pattern. [The Wikipedia article](https://en.wikipedia.org/wiki/Dependency_injection) is a pretty reasonable place to start. In order to have faster, more granular, lower-level tests, we need to move towards isolated abstractions that take their dependencies in rather than building them. You can see that at work with `InstancesRepository` which is an interface. Because `FormMapViewModel` takes in a generic repository, it's possible to easily substitute a test version in. For the most part, since we don't have a hugely complex domain and large numbers of dependencies, we can generally inject dependencies through constructors which is easy to follow and test. Where it gets tricky is with `Activity` classes that can't rely on constructor injection. Dagger has sometimes been used to overcome that. Right now classes don't have a ton of dependencies and there aren't long dependency chains so it's pretty heavy-handed. But it may become more necessary as things are broken down into smaller classes.

grzesiek2010:
Thanks for that exhaustive explanation. That's my bad. I'm aware of MVVM and DI of course. I wanted to move that code (reading from database) out of the activity but of course I should have made another step and do that in the same way we do with `InstancesRepository`. Sorry, I saw your [comment](https://github.com/opendatakit/collect/pull/3552#discussion_r360228512) and agree that we can close this pr.

lognaturel:
> I'm aware of MVVM and DI of course.

I was pretty sure but wanted to make sure it was all tied together! Apologies if my explanation felt way too basic. 😊 Definitely good concepts to keep studying. 

> move that code (reading from database) out of the activity

I feel you. I am puzzled about how to do that because the URI that comes in via intent itself feels like an implementation detail.

PR #3551, opened 2019-12-20
Added tests to confirm that read-only option works as expected

### LINKED ISSUE #3106: Add espresso tests to verify the read-only option
#### Problem description
I have just fixed the issue related to the read-only option: #3103 
It's an issue that appears from time to time so it would be good to have automated tests to check it.

grzesiek2010:
Closes #3106

I ran all tests in `widgets` directory.

The problem with read-only option appears from time to time so I decided to add automated tests. I did that using Robolectric tests to make the work fast (compared to Espresso tests).

Since I most of the changes are in tests and we are just after releasing v1.25 I think it doesn't require much testing we could even merge it after reviewing and test on a daily basis.

All widgets form.

No.


collect_app/src/test/java/org/odk/collect/android/widgets/BooleanWidgetTest.java
```diff
@@ -79,4 +79,8 @@ public void changingTheCheckboxShouldChangeTheAnswer() {
         assertTrue(widget.isChecked());
         assertTrue((Boolean) widget.getAnswer().getValue());
     }
+
+    @Test
+    public void usingReadOnlyOptionShouldMakeAllClickableElementsDisabled() {
```
seadowg: Is there just not anything to test here? If so I think we should just have a comment in the method body.

seadowg (changes requested):
Looks good. Just one wee comment!

collect_app/src/test/java/org/odk/collect/android/widgets/BooleanWidgetTest.java
grzesiek2010: This widget is unused, I tried to use it (it works like TriggerWidget) but it crashes. @lognaturel can we get rid of it at all?

seadowg: Ah that makes sense!

lognaturel: Yes, I think we should get rid of it! At some point we explored what it would take to have a more true boolean type but we ran into semantics issues and I guess we never fully backed out.

grzesiek2010: Ok so I removed it.

seadowg (approved):

PR #3552, opened 2019-12-20
Pull display code out of FormMapViewModel and add Activity-level tests

### LINKED ISSUE #3546: IndexOutOfBoundsException in FormMapActivity.java:65

lognaturel:
In response to @seadowg's feedback at https://github.com/opendatakit/collect/pull/3447#discussion_r350645070, I have pulled out display-related code from the `FormMapViewModel`.

Incidentally closes #3546 

I have added automated tests and clicked around a map for All Widgets for basic manual verification.

I decided to write my own test double for the `MapFragment`. I took a stateful spy approach since these maps are inherently stateful. I'm pretty happy with how it lets me test behavior without being too tightly coupled to the implementation.

I'm not sure I'm super happy with the `map` field of `FormMapActivity` being public and set by the test but I couldn't come up with a great alternative other than using Dagger which y'all know I really try to avoid. The unfortunate thing about this setup is that I can't test configuration changes as far as I can tell. I've put the test in so we don't forget about it but ignored it.

This change is isolated to the submission map feature so that's where the risk is. Everything for that feature gets moved around but it should all be supported by tests. The only change to cross-cutting code is adding methods so nothing else is at risk.

Any with a geopoint.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -30,6 +30,8 @@ import androidx.lifecycle.ViewModelProviders;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.activities.viewmodels.FormMapViewModel;
+import org.odk.collect.android.activities.viewmodels.FormMapViewModel.ClickAction;
+import org.odk.collect.android.activities.viewmodels.FormMapViewModel.MappableFormInstance;
 import org.odk.collect.android.dao.FormsDao;
 import org.odk.collect.android.forms.Form;
 import org.odk.collect.android.geo.MapFragment;
@@ -40,12 +42,17 @@ import org.odk.collect.android.instances.InstancesRepository;
 import org.odk.collect.android.preferences.AdminKeys;
 import org.odk.collect.android.preferences.AdminSharedPreferences;
 import org.odk.collect.android.preferences.MapsPreferences;
+import org.odk.collect.android.provider.InstanceProviderAPI;
 import org.odk.collect.android.provider.InstanceProviderAPI.InstanceColumns;
 import org.odk.collect.android.utilities.ApplicationConstants;
 import org.odk.collect.android.utilities.ToastUtils;
 
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
 import timber.log.Timber;
 
@@ -57,6 +64,22 @@ public class FormMapActivity extends BaseGeoMapActivity {
     private FormMapViewModel viewModel;
 
     private MapFragment map;
+
+    /**
+     * Quick lookup of instance objects from map feature IDs.
+     */
+    private final Map<Integer, FormMapViewModel.MappableFormInstance> instancesByFeatureId = new HashMap<>();
+
+    /**
+     * Points to be mapped. Note: kept separately from {@link #instancesByFeatureId} so we can
+     * quickly zoom to bounding box.
+     */
+    private final List<MapPoint> points = new ArrayList<>();
+
+    /**
+     * True if the map viewport has been initialized, false otherwise.
+     */
+    private boolean viewportInitialized;
 
     @Override public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -108,7 +131,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
             map.zoomToPoint(map.getGpsLocation(), true));
 
         findViewById(R.id.zoom_to_bounds).setOnClickListener(v ->
-            viewModel.mapZoomToBoundingBoxRequested(map));
+            map.zoomToBoundingBox(points, 0.8, false));
 
         findViewById(R.id.layer_menu).setOnClickListener(v -> {
             MapsPreferences.showReferenceLayerDialog(this);
@@ -129,41 +152,76 @@ public class FormMapActivity extends BaseGeoMapActivity {
         updateInstanceGeometry();
     }
 
-    protected void updateInstanceGeometry() {
+    private void updateInstanceGeometry() {
         if (map == null) {
             return;
         }
 
-        viewModel.mapUpdateRequested(map);
+        updateMapFeatures();
+
+        if (!viewportInitialized && !points.isEmpty()) {
+            map.zoomToBoundingBox(points, 0.8, false);
+            viewportInitialized = true;
+        }
 
         TextView statusView = findViewById(R.id.geometry_status);
-        statusView.setText(getString(R.string.geometry_status, viewModel.getInstanceCount(), viewModel.getMappedPointCount()));
+        statusView.setText(getString(R.string.geometry_status, viewModel.getTotalInstanceCount(), points.size()));
     }
 
+    /**
+     * Clears the existing features on the map and places features for the current form's instances.
+     */
+    private void updateMapFeatures() {
+        points.clear();
+        map.clearFeatures();
+
+        List<MappableFormInstance> instances = viewModel.getMappableFormInstances();
+        for (MappableFormInstance instance : instances) {
+            MapPoint point = new MapPoint(instance.getLatitude(), instance.getLongitude());
+            int featureId = map.addMarker(point, false);
+
+            int drawableId = getDrawableIdForStatus(instance.getStatus());
+            map.setMarkerIcon(featureId, drawableId);
+
+            instancesByFeatureId.put(featureId, instance);
+            points.add(point);
+        }
+    }
+
+    /**
+     * Zooms the map to the new location if the map viewport hasn't been initialized yet.
+     */
     public void onLocationChanged(MapPoint point) {
-        viewModel.locationChanged(point, map);
+        if (!viewportInitialized) {
+            map.zoomToPoint(point, true);
+            viewportInitialized = true;
+        }
     }
 
+    /**
+     * Reacts to a tap on a feature by showing a toast or switching activities to view or edit a form.
+     */
     public void onFeatureClicked(int featureId) {
-        FormMapViewModel.FeatureStatus status = viewModel.getStatusOfClickedFeature(featureId);
+        MappableFormInstance instance = instancesByFeatureId.get(featureId);
+        ClickAction clickAction = instance == null ? FormMapViewModel.ClickAction.NONE : instance.getClickAction();
 
         boolean canEditSaved = (Boolean) AdminSharedPreferences.getInstance().get(AdminKeys.KEY_EDIT_SAVED);
 
-        switch (status) {
-            case DELETED:
+        switch (clickAction) {
+            case DELETED_TOAST:
                 String deletedTime = getString(R.string.deleted_on_date_at_time);
                 String disabledMessage = new SimpleDateFormat(deletedTime,
-                        Locale.getDefault()).format(viewModel.getDeletedDateOf(featureId));
+                        Locale.getDefault()).format(viewModel.getDeletedDateOf(instance.getDatabaseId()));
 
                 ToastUtils.showLongToast(disabledMessage);
                 break;
-            case NOT_VIEWABLE:
+            case NOT_VIEWABLE_TOAST:
                 ToastUtils.showLongToast(R.string.cannot_edit_completed_form);
                 break;
-            case VIEW_ONLY:
+            case OPEN_READ_ONLY:
                 startActivity(getViewOnlyFormInstanceIntentFor(featureId));
                 break;
-            case EDITABLE:
+            case OPEN_EDIT:
                 if (canEditSaved) {
                     startActivity(getEditFormInstanceIntentFor(featureId));
                 } else {
@@ -174,7 +232,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
     }
 
     private Intent getEditFormInstanceIntentFor(int featureId) {
-        Uri uri = ContentUris.withAppendedId(InstanceColumns.CONTENT_URI, viewModel.getDatabaseIdOf(featureId));
+        Uri uri = ContentUris.withAppendedId(InstanceColumns.CONTENT_URI, instancesByFeatureId.get(featureId).getDatabaseId());
         return new Intent(Intent.ACTION_EDIT, uri);
     }
 
@@ -189,6 +247,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
         double mapZoom = state.getDouble(MAP_ZOOM_KEY);
         if (mapCenter != null) {
             map.zoomToPoint(mapCenter, mapZoom, false);
+            viewportInitialized = true; // avoid recentering as soon as location is received
         }
     }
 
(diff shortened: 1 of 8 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -56,15 +64,45 @@
...
+    @VisibleForTesting public ViewModelProvider.Factory viewModelFactory;
 
     @Override public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         FormsDao dao = new FormsDao();
-        Form form = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData())).get(0);
-        viewModel = ViewModelProviders.of(this,
-                new FormMapActivity.FormMapViewModelFactory(form, new DatabaseInstancesRepository())).get(FormMapViewModel.class);
+        Form form = null;
```
lognaturel, line 91: @grzesiek2010 this is incidentally a fix for #3546 and the one I would recommend if you feel strongly about addressing it before v1.26 goes out with this change.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -75,9 +113,16 @@
...
-            .addTo(this, R.id.map_container, this::initMap, this::finish);
+        if (map == null) { // tests set their maps directly
+            Context context = getApplicationContext();
+            map = MapProvider.createMapFragment(context);
+        }
+
+        if (map != null) {
+            map.addTo(this, R.id.map_container, this::initMap, this::finish);
+        } else {
+            finish(); // The configured map provider is not available
```
lognaturel, line 124: I haven't tested/tried it yet but I think this will address @mmarciniak90's find at https://github.com/opendatakit/collect/pull/3503#issuecomment-567511721.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
```diff
@@ -66,129 +46,126 @@ public String getFormId() {
...
 
+    private void initializeFormInstances() {
         List<Instance> instances = instancesRepository.getAllBy(form.getJrFormId());
-        totalInstanceCount = instances.size();
-        for (Instance instance : instances) {
+
+        // Note: there is currently no way to delete instances from FormMapActivity so this works
+        // because a change in size means a refresh is needed. Compromise because we don't currently
+        // have an easy way to observe database changes.
+        if (mappableFormInstances == null || instances.size() != totalInstanceCount) {
```
lognaturel, line 63: Kind of shady. Any alternative option I looked into -- getting some kind of callback when there's a DB change -- seemed terribly complex for not much gain. @seadowg you may have some ideas.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -56,15 +64,45 @@
 
     private FormMapViewModel viewModel;
 
-    private MapFragment map;
+    @VisibleForTesting public MapFragment map;
```
seadowg, line 67: I think it would be nice to use a factory for injection here rather the `MapFragment` itself. Google talked about introducing `FragmentFactory` in the near future (https://www.notion.so/Fragments-Past-Present-and-Future-f21afdfc5e5346f9bba1c89707ff3358) so this feels in keeping with where the framework is going as well.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -56,15 +64,45 @@
...
+        List<Form> forms = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData()));
+        if (forms.size() == 1) {
+            form = forms.get(0);
+        } else {
+            // This shouldn't be possible because the instance URI had to be used in the calling
+            // activity but it has been logged to Crashlytics.
+            finish();
+        }
+
+        if (viewModelFactory == null) { // tests set their factories directly
```
seadowg, line 101: Is it possible to do this at field initialization or do we need to be in `onCreate` for some things to work?

collect_app/src/test/java/org/odk/collect/android/activities/FormMapActivityTest.java
```diff
@@ -0,0 +1,198 @@
...
+    @Test public void startingFormMap_doesNotZoom_ifThereAreNoInstanceMarkers_andLocationIsUnavailable() {
+        ActivityController controller = RobolectricHelpers.buildThemedActivity(FormMapActivity.class);
+        FormMapActivity activity = (FormMapActivity) controller.get();
+
+        TestInstancesRepository testInstancesRepository = new TestInstancesRepository(new ArrayList<>());
+        FormMapViewModel viewModel = new FormMapViewModel(FormMapViewModelTest.testForm1, testInstancesRepository);
+        activity.viewModelFactory = new TestFactory(viewModel);
+        activity.map = new TestMapFragment();
+
+        controller.setup().visible().start().resume();
```
seadowg: `setup` runs all the following method calls here (unless you're meaning to go through the lifecycle twice). I wrote the method 😉.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: I should perhaps have given a bit more detail here about what I was running into. There's already a static `MapProvider` which uses general settings to build a configurator which can itself build a `MapFragment`. I think that ideally the `MapProvider` wouldn't be static but that will require (some) rework of all the geo components as far as I can see. Is that what you have in mind? Do you think it's appropriate to do here/now?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: The content provider URI of the form is passed via intent to the activity. That feels like way too low-level of a concern for `FormMapViewModel` to have which is why I resolved it to a `Form` in `onCreate`. If the `Form` is passed in to the factory constructor, then the factory can't be built at the field initialization. I suppose I could use a setter on the factory? Or do you think the view model should get the form uri?

collect_app/src/test/java/org/odk/collect/android/activities/FormMapActivityTest.java
lognaturel: Yikes, thank you! Did I get confused with `create`? I guess this is exactly why `ActivityScenario` is being pushed. Too may newbs like me putting together "creative" lifecycle calls.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
seadowg: Ah actually it looks [`FragmentFactory`](https://developer.android.com/reference/androidx/fragment/app/FragmentFactory) already exists and can be set on the `FragmentManager`. I'm not too in to the design of it - you basically create a `FragmentFactory` implementation that handles all the Fragments in your app.

I had a quick play with using this approach but I think you're right in that because we have multiple of the way the `MapFragment` is designed (it's not a fragment) this approach wouldn't work. Happy to go with your approach right now I think. Alternatively if you're up for it you could wrap the call to `MapProvider.createMapFragment` in a simple Factory that you can swap out in the test.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
seadowg: Nah I think I like the way you have it then. I haven't found a nice way to set these up in `onCreate`. In `FormEntryActivity` I've ended up with `setupViewModels` method to try and seperate it out from all the other `onCreate` logic.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
seadowg: I'm probable being dumb but I can work out how you end up with the instance number changing during the lifecycle of this View Model. Does the user not have to leave the map view to delete/add something and then when they come back they'll have a new View Model?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: > it's not a fragment

I guess you mean that the concrete classes are `Fragment`s but the interface itself doesn't include anything `Fragment`-like? That is quite odd now that you mention it. And I guess there's no way to require that subclasses be `Fragment`s since there is no `Fragment` interface? Yikes. Not sure what to do about this. It seems like semantically we want something like `MapFragment extends Fragment`. Or maybe `MapFragment` should be more of a fragment container?

> you could wrap the call to MapProvider.createMapFragment in a simple Factory that you can swap out in the test.

That makes sense. Feels a little inception-like (a factory for a provider of factories) but will play with it!

seadowg: > It seems like semantically we want something like `MapFragment extends Fragment`. Or maybe `MapFragment` should be more of a fragment container?

Yeah my first though is that `MapFragment` should extend `Fragment` and should have a `setMap`/`getMap` that takes the interface we currently call `MapFragment` (The interface formerly known as...)

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
lognaturel: This surprised me but it appears that the `ViewModel` is retained when the user goes to the form entry activity. I've done some digging just now and am even more confused. Do you have a reference that describes what you think should happen?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: > `MapFragment` should extend `Fragment`

But that's not possible because there's no `Fragment` interface, right? Or am I missing something...

seadowg: I mean we make a MapFragment extends the Android Fragment class and that it can support any of the mapping engines (maybe by using a Fragment container). That's future us's problem though - not for this PR.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
seadowg: Ah sorry you can click on an instance and then come back to this page right? That makes sense. The ViewModel will be retained while the Activity is in the back stack.

lognaturel: Do you have a reference for this? I couldn't find it anywhere. The `Activity` itself gets killed, right?

seadowg: After chatting on Slack I think @lognaturel is right. We can get back to this screen after changing the instances so do need to think about reloads!

lognaturel: What I was confused about was just the normal behavior for activities on the back stack. [They live on in a paused state](https://developer.android.com/guide/components/activities/activity-lifecycle) so all this makes sense. I'm not sure what I was getting confused about. Perhaps too much usage of "Don't keep activities."

What @seadowg was unsure about is the flow back to the map because typically exiting `FormEntryActivity` ignores the back stack and goes straight back to the main activity. We made a decision to return to the map on save because we want to enable a map-centric workflow where an enumerator uses the map as their landing place to fill and edit forms.

So I think we're all on the same page, now!

collect_app/src/main/java/org/odk/collect/android/instances/DatabaseInstancesRepository.java
```diff
@@ -25,7 +25,7 @@
 
     @Override
     public Instance getBy(long databaseId) {
-        Cursor c = dao.getInstancesCursorForId("" + databaseId);
```
seadowg: This was such a cheeky line of code 😁. Took me a second to understand what was happening.

collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
```diff
@@ -31,87 +31,87 @@
...
         List<FormMapViewModel.MappableFormInstance> mappableFormInstances = viewModel.getMappableFormInstances();
 
         assertThat(mappableFormInstances.size(), is(0));
     }
 
     @Test public void getDeletedDateOf_returnsDeletedDate() {
-        FormMapViewModel viewModel = new FormMapViewModel(testForm1, testInstancesRepository);
+        FormMapViewModel viewModel = new FormMapViewModel(TEST_FORM_1, testInstancesRepository);
 
         assertThat(viewModel.getDeletedDateOf(0L), is(1487782554846L));
```
seadowg: Minor nitpick: could the assertion be that it's equal to `testInstance[0].getDeletedDate()` instead. Just a bit hard to track where this value is coming from.

collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
```diff
@@ -31,87 +31,87 @@
     }
 
     @Test public void getFormTitle_returnsFormTitle() {
```
seadowg, line 33: I'm wondering if we can have a data object that encapsulates the form title and id here. I've been feeling like something along those lines would be helpful for decoupling other `ViewModel`s from the `FormController`. I've been strugging to find a simple place to start with this but this feels like it might be it?

seadowg (approved):
Looking good! Just a nitpick and a suggestion but good to go to QA.

collect_app/src/main/java/org/odk/collect/android/instances/DatabaseInstancesRepository.java
lognaturel: Yeah, sorry, not my best work. 😄

collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
lognaturel: Thanks, done.

collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
lognaturel: I struggle a bit with where those kinds of limited façades into domain objects are simplifying vs. obfuscating. In this case, we don't really need the id, it's just being used in a log that perhaps should be cleaned up. The `activity` knows about the `Form` that is being mapped so it might even make more sense to remove those two methods from the view model and have the activity directly get the values it needs from the `Form` object. What do you think?

seadowg: @lognaturel yeah I like that!

mmarciniak90:
**Tested with success**

Verified on device: 4.4, 5.1, 6.0, 7.0, 8.1, 9.0, 10.0
I  didn't notice regression. I wasn't able to reproduce a crash from #3546 but also I didn't notice crash on apk with these changes.

**Verified cases:**
* test cases from regression issue NODK-217 GeoPoint widget
* test cases from regression issue NODK-203 Location widgets
* light/dark mode

PR #3557, opened 2019-12-26
Remove custom handling for file types handled by android itself

### LINKED ISSUE #3173: Remove custom handling for file types

chidauri:
Closes #3173

It passes `./gradlew testDebug ` 
Failes `./gradlew checkAll` with these errors :(
```
> Task :collect:collect:collect_app:lintDebug FAILED
> Task :collect:checkCode FAILED
```

These are the mappings that are present in android's MimeTypeMap class:
```
XML("xml",  "text/xml"),
_3GPP("3gpp", "audio/3gpp"),
_3GP("3gp",  "video/3gpp"),
AVI("avi",  "video/avi"),
CSV("csv",  "text/csv"),
JPG("jpg",  "image/jpeg"),
MP3("mp3",  "audio/mp3"),
MP4("mp4",  "video/mp4"),
OGG("ogg",  "audio/ogg"),
WAV("wav",  "audio/wav"),
XLS("xls",  "application/vnd.ms-excel");
```
Some of the extensions above have multiple mappings, for example "3gpp", i have currently deleted them too.

This change should not effect users.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
@@ -7,6 +7,11 @@ import androidx.annotation.NonNull;
 import org.odk.collect.android.http.openrosa.OpenRosaHttpInterface;
 import org.odk.collect.android.utilities.FileUtils;
 
+/**
+ * This covers types not included in Android's MimeTypeMap
+ * Reference https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java
+ */
+
 public class CollectThenSystemContentTypeMapper implements OpenRosaHttpInterface.FileToContentTypeMapper {
 
     private final MimeTypeMap androidTypeMap;
--- a/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
@@ -33,21 +33,10 @@ public class CollectThenSystemContentTypeMapper implements OpenRosaHttpInterface
     }
 
     private enum CollectContentTypeMappings {
-        XML("xml",  "text/xml"),
-        _3GPP("3gpp", "audio/3gpp"),
-        _3GP("3gp",  "video/3gpp"),
-        AVI("avi",  "video/avi"),
         AMR("amr",  "audio/amr"),
-        CSV("csv",  "text/csv"),
-        JPG("jpg",  "image/jpeg"),
-        MP3("mp3",  "audio/mp3"),
-        MP4("mp4",  "video/mp4"),
         OGA("oga",  "audio/ogg"),
-        OGG("ogg",  "audio/ogg"),
         OGV("ogv",  "video/ogg"),
-        WAV("wav",  "audio/wav"),
-        WEBM("webm", "video/webm"),
-        XLS("xls",  "application/vnd.ms-excel");
+        WEBM("webm", "video/webm");
 
         private String extension;
         private String contentType;
--- a/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
@@ -11,7 +11,6 @@ import org.odk.collect.android.utilities.FileUtils;
  * This covers types not included in Android's MimeTypeMap
  * Reference https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java
  */
-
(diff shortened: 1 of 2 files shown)


Rezachow:
How can i join?

collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
```diff
@@ -33,21 +33,10 @@ public String map(String fileName) {
     }
 
     private enum CollectContentTypeMappings {
-        XML("xml",  "text/xml"),
```
seadowg, line 36: Would you be able to link to what docs/code helped you make the decision to remove these content types?

seadowg (changes requested):
@chidauri I'm not seeing a failure from `lintDebug` here. Can you have a look at the report and see what the problem is?

**EDIT**: I just double checked `checkAll` and I'm not seeing any failures.

seadowg:
@Rezachow it's generally easier top just have one person working on a PR at a time. If you're looking for something to work on have a look at our [good first issue tag](https://github.com/opendatakit/collect/labels/good%20first%20issue).

collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
chidauri: [this one](https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java)

chidauri:
@seadowg Actually i think the error is due to my phone or my installation. How do i grant these permissions? 
Screenshot from 2019-12-28 19-49-33

seadowg:
@chidauri that error occurs if the Android tooling thinks something in our code is incorrect. Here it looks like it thinks we need extra permissions for the code in `PropertyManager` to work. I'm not seeing this on my machine and it's also passing on CI 😕.

Could you make sure you're local SDK tooling is up to date and also that uyou haven't got any changes in `PropertyManager` and check this again?

collect_app/src/main/java/org/odk/collect/android/http/CollectThenSystemContentTypeMapper.java
seadowg: Could you add that reference as a comment to the top of the file. It would be good to have a note that state that "This covers types not included in Android's `MimeTypeMap`" with a link so we know where to look if something goes wrong!

chidauri: Done

chidauri:
@seadowg i checked the lint report, here it shows missing _android.permission.READ_PRIVILEGED_PHONE_STATE_
Screenshot from 2019-12-28 21-15-24

which is required from android 10 onwards , and i am using Android 10 SDK platform currently.

Ps. I checked, All SDK Tools are up to date.

lognaturel:
Please compile against a lower SDK. We are indeed not ready to compile with the Android 10 SDK. We will delay making this change as long as we can (likely through summer 2020) because users currently do rely on the fixed IDs that we will no longer have access to.

chidauri:
@lognaturel @seadowg All errors are gone(after changing to android 9 SDK) :)

seadowg (dismissed):

seadowg (changes requested):
@chidauri one thing I'm feeling unclear on (maybe just paranoid) is whether the `MimeTypeMap` is consistent between API levels. Would you be able to have a play/do some research into that? Our minimum API level is 16 so my main concern is that older devices might not have the same mapping as newer ones.

chidauri:
@seadowg The documentation says all methods were added in api 1 and i have used git blame [here](https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java) , and all changes correspond to 2009.

seadowg (approved):

seadowg:
For QA I think we should test forms with files as part of their submission to behaviour is still the same. We're probably ok - if we were writing this code for the first time now we'd probably rely on Android's own content type mapping without giving it a thought. Moving back to it from custom code is a little scary however.

mmarciniak90:
**Tested with success.**
I did not notice regression.

Verified cases:
- media formats: 
   - image widget
       - png, jpg, jpeg, gif
   - audio widget
      - ogg, m4a, mp3
   - video widget
       - 3gp, mp4
    - file widget
        - jpg, jpeg, png, gif
        - amr, 3gp, ogg, m4a, mp3, mp4, wav
        - csv, xml, doc, docx, pdf, rtf, txt, xls, xlsx
        - zip, rar
- media loaded from Google Drive
- media loaded from device
- media loaded from `download` directory
- new media
- form sent to Aggregate 2.0
- form sent to Aggregate 1.7
- form sent to Central
- form sent to Google Sheet 

PR #3568, opened 2019-12-30
Fixes Issue#3493: Fixed margin for widgets with buttons when value is added manually

### LINKED ISSUE #3493: Margin is not noticeable for widgets with buttons when a value is added manually
Collect v1.25

#### Problem description

Margin is NOT visible for widgets:
* Ex string when a value is added manually
* Ex integer when a value is added manually
* Ex decimal when a value is added manually
* Bearing when a value is added manually

1. User opens one on mentioned widgets
2. User clicks on the button
3. Text field appears
4. Margin is not visible for the text field

#### Expected behavior
Margin is noticeable 

#### Other information 
Noticed during testing #3464

SaumiaSinghal:
Closes #3493 

I tested it on Android 9.0.

Before adding margin           |     After adding margin
:-------------------------:|:-------------------------:
Screenshot_20191230_034615  |  Screenshot_20191230_033317

   

I used the default margin value, that is being used in the other widgets, for uniformity. It adds the margin to the answer layout, where the answer field will appear. 

I don't think there are any regression risks.

Allwidges or any form that has any of the three widgets:

- Ex-integer widget
- Ex-decimal widget
- Ex-string widget 

no

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/AbstractSelectOneWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/AbstractSelectOneWidget.java
@@ -16,11 +16,10 @@ package org.odk.collect.android.widgets;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.widget.RadioButton;
+
 import androidx.annotation.Nullable;
 import androidx.recyclerview.widget.RecyclerView;
-
-import android.content.res.Resources;
-import android.widget.RadioButton;
 
 import org.javarosa.core.model.FormIndex;
 import org.javarosa.core.model.SelectChoice;
@@ -28,7 +27,6 @@ import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.SelectOneData;
 import org.javarosa.core.model.data.StringData;
 import org.javarosa.core.model.data.helper.Selection;
-import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.SelectOneListAdapter;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.exception.JavaRosaException;
@@ -38,9 +36,6 @@ import org.odk.collect.android.logic.FormController;
 import org.odk.collect.android.widgets.interfaces.MultiChoiceWidget;
 
 import timber.log.Timber;
-
-import static org.odk.collect.android.utilities.ViewUtils.dpFromPx;
-import static org.odk.collect.android.widgets.StringWidget.FIELD_HORIZONTAL_MARGIN_MODIFIER;
 
 /**
  * SelectOneWidgets handles select-one fields using radio buttons.
@@ -104,11 +99,7 @@ public abstract class AbstractSelectOneWidget extends SelectTextWidget implement
             recyclerView.setAdapter(adapter);
             answerLayout.addView(recyclerView);
             adjustRecyclerViewSize(adapter, recyclerView);
-
-            Resources resources = context.getResources();
-            int marginStandard = dpFromPx(context, resources.getDimensionPixelSize(R.dimen.margin_standard));
-            int margin = marginStandard - FIELD_HORIZONTAL_MARGIN_MODIFIER;
-            addAnswerView(answerLayout, margin);
+            addAnswerView(answerLayout);
         }
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/AbstractSelectOneWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/AbstractSelectOneWidget.java
@@ -18,6 +18,8 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import androidx.annotation.Nullable;
 import androidx.recyclerview.widget.RecyclerView;
+
+import android.content.res.Resources;
 import android.widget.RadioButton;
 
 import org.javarosa.core.model.FormIndex;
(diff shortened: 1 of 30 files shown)


grzesiek2010 (changes requested):
The issue is related to all widgets that contain buttons but you implemented changes only in ExStringWidget class, that's not enough. Please investigate [All Widgets form](https://docs.google.com/spreadsheets/d/1af_Sl8A_L8_EULbhRLHVl8OclCfco09Hq2tqb9CslwQ/edit#gid=0) to get to know more about different types of widgets and complete your solution.

grzesiek2010:
Apart from the fact this solution is not complete @seadowg isn't it something that you are working on as a part of another issue/pr? I'm asking just to avoid conflicts.

seadowg:
@grzesiek2010 @SaumiaSinghal I think we'd want to revise the layout of these widgets after we look at changing up the widget framework. That'll be after #3483 (which doesn't look like it conflicts with this) so I think it's ok to go ahead with an earlier fix as long as it's not introducing any large changes.

SaumiaSinghal:
Hie @grzesiek2010, I fixed it for all the widgets, wherever I found this problem. Can you review it again?

collect_app/src/main/java/org/odk/collect/android/widgets/AbstractSelectOneWidget.java
```diff
@@ -90,15 +96,19 @@ public IAnswerData getAnswer() {
...
             RecyclerView recyclerView = setUpRecyclerView();
             recyclerView.setAdapter(adapter);
             answerLayout.addView(recyclerView);
             adjustRecyclerViewSize(adapter, recyclerView);
-            addAnswerView(answerLayout);
+
+            Resources resources = context.getResources();
+            int marginStandard = dpFromPx(context, resources.getDimensionPixelSize(R.dimen.margin_standard));
+            int margin = marginStandard - FIELD_HORIZONTAL_MARGIN_MODIFIER;
+            addAnswerView(answerLayout, margin);
```
seadowg: We don't want to any of the select widgets as we need the select options to spread from edge to edge. This change should be reverted.

collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
```diff
@@ -63,7 +66,11 @@ public AnnotateWidget(Context context, QuestionDetails prompt) {
         imageCaptureHandler = new ImageCaptureHandler();
         setUpLayout();
         addCurrentImageToLayout();
-        addAnswerView(answerLayout);
+
+        Resources resources = context.getResources();
```
seadowg, line 70: It'd be good to pull this repeated code into `WidgetViewUtils`. Could be a method like `getStandardMargin()`.

**EDIT**: `WidgetViewUtils is actually not in the code yet as it is introduced in #3483. I think we should block this PR on that one as we may end up with conflicts. It should hopefully be merged soon 🤞.

collect_app/src/main/java/org/odk/collect/android/widgets/ArbitraryFileWidget.java
```diff
@@ -171,7 +175,10 @@ private void setUpLayout() {
         widgetLayout.addView(chooseFileButton);
         widgetLayout.addView(answerLayout);
 
-        addAnswerView(widgetLayout);
+        Resources resources = this.getResources();
+        int marginStandard = dpFromPx(getContext(), resources.getDimensionPixelSize(R.dimen.margin_standard));
+        int margin = marginStandard - FIELD_HORIZONTAL_MARGIN_MODIFIER;
```
seadowg: This `FIELD_HORIZONTAL_MARGIN_MODIFIER` should only really be used on the `StringWidget` as it will make the margin smaller than it should be. Could you make sure to only modify the standard margin on the widgets that you feel need it? If it's not just the `StringWidget` that needs it we should move the constant somewhere else.

collect_app/src/main/java/org/odk/collect/android/widgets/ExStringWidget.java
```diff
@@ -116,7 +118,11 @@ protected void setUpLayout(Context context) {
         answerLayout.setOrientation(LinearLayout.VERTICAL);
         answerLayout.addView(launchIntentButton);
         answerLayout.addView(answerText);
-        addAnswerView(answerLayout);
+        
+        Resources resources = context.getResources();
+        int marginStandard = dpFromPx(context, resources.getDimensionPixelSize(R.dimen.margin_standard));
+        int margin = marginStandard - FIELD_HORIZONTAL_MARGIN_MODIFIER;
```
seadowg, line 124: This margin ends up being too big as `QuestionWidget#getSimpleButton` is adding a margin on for us. Your changes should allow us to delete the code that adds the margin there so you should be able to do that! Again though, I'd wait for #3483 to be merged as this method moves to `WidgetViewUtils`.

seadowg (changes requested):
A few changes needed. I'd wait for #3483 to be merged before going ahead though as that will give you some help.

collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
SaumiaSinghal: Yes, @seadowg, I also think that this bit of code is repetitive, and better should be specified in a separate class. It would be nice to have the PR merged with `WidgetViewUtils`, after that I can shift it there.

SaumiaSinghal:
Hie @seadowg, yes it'd be better to wait for #3483 to be merged then, we can decide which all widgets need the margin and if we are extending the margin not only to the string widgets, but also to others as well, we can specify the constant `FIELD_HORIZONTAL_MARGIN_MODIFIER` somewhere else. Once, #3483 is merged, I will rework on the margin issue.

seadowg:
@SaumiaSinghal #3483 is merged now 👍

SaumiaSinghal:
Hie @seadowg! I shifted the reused code to `WidgetViewUtils` and added margins using  `WidgetViewUtils.getStandardMargin()` to the widgets, where I found it necessary. I'm closing this PR and opening a new one.

PR #3569, opened 2020-01-03
Modified checkstyle EmptyLineSeparator rule allowing empty lines between static blocks

### LINKED ISSUE #3075: Turn off the Checkstyle rule requiring a blank line before static { } blocks.
This is a developer-facing issue only.

Checkstyle currently enforces the following rule:

```
'STATIC_INIT' should be separated from previous statement.
```

This forces us to put a blank line before any `static { }` declaration.  For example, Checkstyle breaks the build for the following code:

```
    protected static final Map<String, String> STYLE_URLS = new HashMap<>();
    static {
        STYLE_URLS.put(GeneralKeys.MAPBOX_MAP_STREETS, Style.MAPBOX_STREETS);
        STYLE_URLS.put(GeneralKeys.MAPBOX_MAP_LIGHT, Style.LIGHT);
    }
```

It rejects the code because there is no blank line between the first line and the second line.

I think we should turn this off, because a `static { }` block commonly has the purpose of initializing a declared member, just as in this example where the purpose is to declare a constant `HashMap` populated with some constant values.  In this case, the `static { }` block should be grouped together with the declaration, because the declaration and initialization of this member should be treated as a unit.  It's unfortunate that Checkstyle would force us to write this code in a way that is slightly less clear as to its purpose.

SaumiaSinghal:
Closes #3075 

I tested it by running `./gradlew checkstyle` for the example mentioned in the issue.

I added another property to EmprtyLineSeparator rule, which does not allow empty lines between static fields. 

No

No

No

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -99,6 +99,7 @@
             <property name="allowMultipleEmptyLines" value="false"/>
             <property name="allowMultipleEmptyLinesInsideClassMembers" value="false"/>
             <property name="allowNoEmptyLineBetweenFields" value="true" />
+            <property name="tokens" value="STATIC_INIT"/>
         </module>
         <module name="SeparatorWrap">
             <property name="id" value="SeparatorWrapDot" />


grzesiek2010 (approved):

grzesiek2010:
Thanks!

Issue #2700, opened 2018-10-29
Fix ImpliedQuantity lint


yanokwa:
Filing comment from https://github.com/opendatakit/collect/pull/2698 so we don't forget. 

> I followed the process described here. I did make a one manual change which was to ignore the ImpliedQuantity lint for French and Ukrainian. Now that I see an example of a failure in a language I know (French), I considered disabling the lint altogether. I thought this might be better to do after the release.

lognaturel:
To provide a bit more context, while some languages like English only have singular (1 item) and plural (multiple items), some languages have more complex grammar around quantities. From the lint docs:

```
Plural strings should generally include a %s or %d formatting argument. In
locales like English, the one quantity only applies to a single value, 1, but
that's not true everywhere. For example, in Slovene, the one quantity will
apply to 1, 101, 201, 301, and so on. Similarly, there are locales where
multiple values match the zero and two quantities.

In these locales, it is usually an error to have a message which does not
include a formatting argument (such as '%d'), since it will not be clear from
the grammar what quantity the quantity string is describing.
```

The SMS sending string has two variants:
- `Submitting form as a single message.`
- `Submitting form, sending message %1$d of %2$d`

We get a lint failure for slavic languages (Slovene, Ukrainian, for example) for this string because the translators don't use a value in the `one` translation. I've talked to them on Transifex and they assure me that makes sense. We get a failure for French because apparently the `one` translation is usable for 0 or 1. This is irrelevant because if there are 0 messages the message isn't used.

I see three options:
1) keep manually suppressing the lint check locally when there's a translation update
2) disable the lint check entirely
3) change the SMS sending message so there's only one variant

I'm leaning towards 3. I think "sending message 1 of 1" is explicit and easy to understand. There isn't much benefit to having the `one` variant and it's causing confusion.

PR #3572, opened 2020-01-03
adding the code for disable the lint

maharishi-01:
In the Reference to #2700 .
On going through the code , i choose 2nd option.Actually this is my first PR, but i'm taking a look alongside to try to fix it.

if it is wrong please guide me.....

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -82,6 +82,13 @@ android {
         archivesBaseName = 'collect'
     }
 
+    android {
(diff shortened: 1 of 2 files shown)


grzesiek2010:
Hi @akatsuki619 
thanks for your contribution but we don't want to disable the lint at all what you did. We want to fix those warnings that should be fixed (have higher priorities) and probably just leave those less important. If you have a question please ask in the issue.

PR #3574, opened 2020-01-05
Fix unexpected prompt to add not relevant repeat instances

### LINKED ISSUE #2992: Unexpected prompt to add not-relevant repeat instances
Collect v1.21

#### Problem description
As [discussed in PR #2990](https://github.com/opendatakit/collect/pull/2990#issuecomment-479600700), using the [hiddenRepeat.xml](https://github.com/opendatakit/collect/files/3040187/hiddenRepeat.xml.txt) form, stepping forward and backward results in unexpected prompts to add new instances of groups that are designated as `relevant="false()"`.

1. Step to a repeatable group that's not relevant
2. You're prompted to add a new instance

#### Expected behavior
You should not be prompted to add a new instance.

chidauri:
Closes #2992 

I have tested the change with these forms:
[hiddenRepeat1.xlsx](https://github.com/opendatakit/collect/files/4023810/hiddenRepeat1.xlsx)
[hiddenRepeat2.xlsx](https://github.com/opendatakit/collect/files/4023811/hiddenRepeat2.xlsx)
[hiddenRepeat3.xlsx](https://github.com/opendatakit/collect/files/4023814/hiddenRepeat3.xlsx)
[hiddenRepeat4.xlsx](https://github.com/opendatakit/collect/files/4023816/hiddenRepeat4.xlsx)

The relevance of a repeat group can be found by the relevance of its child indices i.e., child indices are not relevant if the group itself is not relevant, so before prompting to the user i create one dummy instance, check the relevance with the help of child indices and then delete that instance. 
I am open to discussion about any another approach.

If the user adds a repeat group that does not contain any relevant child indices though the group itself is relevant, the group will not be considered relevant and user will not be shown the prompt to add this type of group. 

No i don't think so.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/logic/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/logic/FormController.java
@@ -645,7 +645,6 @@ public class FormController {
         return isindexrelev;
     }
 
-
     /**
      * Move the current form index to the index of the next question in the form.
      * Stop if we should ask to create a new repeat group or if we reach the end of the form.
@@ -664,7 +663,7 @@ public class FormController {
                         case FormEntryController.EVENT_END_OF_FORM:
                             break group_skip;
                         case FormEntryController.EVENT_PROMPT_NEW_REPEAT:
-                            if(!isRepeatRelevant()) {
+                            if (!isRepeatRelevant()) {
                                 // skip Irrelevant repeat prompt
                                 return stepToNextScreenEvent();
                             }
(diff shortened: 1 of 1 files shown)


grzesiek2010:
@chidauri 
Thanks for your effort I'm sure it wasn't easy to investigate and understand that part of code.
Unfortunately it's not a solution we would expect...
First of all the mentioned issue is not something that would cause problems or something it's just a strange behavior we could handle in a better way. If doing that requires such code (adding new repeats checking elements inside and then removing) I'm not sure if it's worth the effort and it's not too risky.
On the other hand I'm also not sure if it should be fixed in Collect at all... maybe in Javarosa?
Actually there is a method in Javarosa called [isRepeatRelevant()](https://github.com/opendatakit/javarosa/blob/master/src/main/java/org/javarosa/core/model/FormDef.java#L536) and probably that method should be improved to cover cases described in the issue.

I'm closing this pr. If you want to investigate it on the Javarosa side please go ahead.

PR #3575, opened 2020-01-07
show a warning Toast if trouble getting start-geopoint

### LINKED ISSUE #3427: Display warning in toast if background location is requested but location is turned off in settings
Let the enumerator know if no location provider is available for a background location request. [Forum user request](https://forum.opendatakit.org/t/auto-gps-implementation-on-odk-collect/10635/38).

`BackgroundLocationManager.locationPermissionGranted` does this for audit location requests but not for background location requests. This was not done for background location requests because `locationClient.isLocationAvailable` docs say `Must be called <b>after</b> Location updates have been requested.` In the case of background location requests, the `setgeopoint` action manages its own location client so `BackgroundLocationManager` doesn't request location updates. However, I don't think that the order is strictly necessary and we could call `locationClient.isLocationAvailable`. I believe the reason that requirement is there is to ensure the priority level is set before. It may not be relevant in this case.

chidauri:
Closes #3427 

Tested by turning location off and on, android 9.
I have not translated the warning Toast yet for different languages, i wanted to ask what is the go-to way for that.

I agree with @lognaturel that since `setgeopoint` manages its own location client we don't need to set the priority before asking `isLocationAvailable()` here.

If the user enters a form that does not audit location but needs start geo-point, a warning toast will appear that says "Trouble getting start geo-point..." if the location is turned off in settings.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2906,7 +2906,13 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         }
 
         if (backgroundLocationMessage == BackgroundLocationManager.BackgroundLocationMessage.PROVIDERS_DISABLED) {
-            new LocationProvidersDisabledDialog().show(getSupportFragmentManager(), LocationProvidersDisabledDialog.LOCATION_PROVIDERS_DISABLED_DIALOG_TAG);
+            // If current form doesn't audit location and we can't get background location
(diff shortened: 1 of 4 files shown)


lognaturel (changes requested):
We discussed in the issue and decided to show the same dialog as in the background audit location case for simplicity.

lognaturel (approved):

lognaturel:
Thank you, @chidauri! 🎉

kkrawczyk123:
**Tested with success!**
Verified on Androids 4.2, 4.4, 5.1, 7.1, 8.1, 9.0 and 10. 

PR #3576, opened 2020-01-08
Move form saving to ViewModel

### LINKED ISSUE #3521: Form saving should move out of `FormEntryActivity`
Right now `FormEntryActivity` creates and executes a `SaveToDiskTask` and adds itself as a listener. It also has control of the "lifecycle" of saving a form - validation, encryption etc. These details are all things the `FormEntryActivity` shouldn't really need to care out. Ideally it would just be able to say "save now" and get told when to show UI changes to the user around saving (error dialogs, toasts etc).  This would make it easier to test and work with and modify the saving process as we wouldn't have to deal with the machinery of the Activity itself. 

In my mind the best course of action is to look at repurposing the `ChangeReasonPromptViewModel` to be responsible for saving forms. This leads us down a path where we could have a `FormEntryViewModel` that also encapsulates form loading.

seadowg:
Closes #3521.

This PR moves the majority of (non-UI) decision making, async work and audit logging concerned with form saving out of `FormEntryActivity` and into the new `FormEntryViewModel`. This ViewModel used to be `ChangesReasonPromptViewModel`. Before this only handled one part of form saving and needed updates from `FormEntryActivity`. This new setup is hopefully easier to understand and should make it easier to add/change form saving in the future.

There are a follow up tasks I think we should look at after this PR:

* Use a `DialogFragment` for saving dialog rather than the deprecated `showDialog`/`dismissDialog` - I wanted to include this work in this PR but have run into some weird problems so going to continue that after this.
* Move core form saving logic from `SaveToDiskTask` to `DiskFormSaver` - add tests to DiskFormSaver to drive moving the logic
* Move public static methods in `SaveToDiskTask` to shared user case/helpers or to private methods where they are used - again adding test coverage if it's not obvious here is good
* Work out what exactly the purpose of `autoSaved` in `FormEntryActivity` is

Played around with a few different forms, ran tests locally and on test lab.

This way forward was driven out by discussion after #3477. Would be great to get feedback on the approach however!

There shouldn't be any changes in behaviour. I think testing form saving with and without encryption, `track-changes-reasons` audit attribute and with different constraints would be best.


collect_app/src/main/java/org/odk/collect/utilities/Clock.java
```diff
@@ -0,0 +1,10 @@
+package org.odk.collect.utilities;
```
seadowg, line 1: This is potentially controversial. It feels like it might be nice to start pulling objects like this out of the `android` package so we can have a distinction in our code what things actually relate to the Android world. This is useful for drawing a line around components and trying not to muddle them with Android stuff. The main advantage other than seperation of concerns and simiplifcation is to let us write tests for these kinds of things without using Robolectric.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/FormEntryViewModel.java
```diff
@@ -91,27 +94,41 @@ private boolean requiresReasonToSave() {
 
     public static class SaveRequest {
 
-        private final boolean complete;
+        private final boolean formComplete;
```
lognaturel: This represents whether the user has checked the checkbox for this form to be finalized, right? I think it would be useful to add a comment here and that it would be clearer to use "finalized". "Complete" is used in the codebase for historical reasons but there is no such user-facing concept so it's really confusing. This is really something I should have mentioned in the original feature implementation but I didn't do as deep of a review. I also wonder whether it would better capture the semantics to call it `shouldBeFinalized` -- it's a user request, right?

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -347,7 +347,7 @@ private JSONObject toGeoJson(GeoPointData data) throws JSONException {
      */
     static File getSavepointFile(String instanceName) {
         File tempDir = new File(Collect.CACHE_PATH);
-        return new File(tempDir, instanceName + ".save");
+        return new File(tempDir, instanceName + ".saveAndExit");
```
lognaturel: This change shouldn't have happened and should be reverted. Leaving it in would mean orphaned savepoint files and possible loss of savepoint data right after the change was deployed.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -39,22 +55,92 @@ public void setAuditEventLogger(@Nullable AuditEventLogger auditEventLogger) {
...
     }
 
-    public void saveToDiskTaskComplete(SaveResult saveResult, long currentTime) {
+    public void editingForm() {
+        if (auditEventLogger != null) {
+            auditEventLogger.setEditing(true);
+        }
+    }
+
+    public void promptDismissed() {
```
lognaturel: Since this is now a more general view model, `changeReasonPromptDismissed` would be a more helpful name, I think.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -39,22 +55,92 @@ public void setAuditEventLogger(@Nullable AuditEventLogger auditEventLogger) {
...
+
+    private void setSaveRequestState(org.odk.collect.android.tasks.SaveResult saveResult, SaveResult.State finalizeError) {
+        this.saveResult.setValue(new SaveResult(finalizeError, saveResult.getSaveErrorMessage()));
+    }
+
+    @SuppressLint("StaticFieldLeak")
+    private void saveToDisk(Uri uri, boolean complete, String updatedSaveName, boolean viewExiting) {
+        saveResult.setValue(new SaveResult(SaveResult.State.SAVING));
+
+        task = new AsyncTask<Void, Void, org.odk.collect.android.tasks.SaveResult>() {
```
lognaturel: This feels really strange. It's a task that in its `doInBackground` method calls a method that builds another task. I don't think there's actually a problem with it but I think it'd be worth adding a comment on why this structure is advantageous.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -100,97 +186,34 @@ public void saveToDiskTaskComplete(SaveResult saveResult, long currentTime) {
                     auditEventLogger.logEvent(AuditEvent.AuditEventType.CONSTRAINT_ERROR, true, currentTime);
                 }
 
-                setSaveRequestState(saveResult, SaveRequest.State.CONSTRAINT_ERROR);
+                setSaveRequestState(saveResult, SaveResult.State.CONSTRAINT_ERROR);
```
lognaturel: Should this be called `setSaveResultState`?

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -69,9 +69,8 @@
  * @author Carl Hartung (carlhartung@gmail.com)
  * @author Yaw Anokwa (yanokwa@gmail.com)
  */
-public class SaveToDiskTask extends AsyncTask<Void, String, SaveResult> {
+public class SaveToDiskTask {
```
lognaturel: Do you think the suffix `Task` is still appropriate? I feel like in the Android world it implies that it's an `AsyncTask` and that had me confused for a bit. I'm not sure it's important but wanted to point it out for consideration.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -81,8 +86,10 @@ public boolean saveReason() {
             auditEventLogger.logEvent(AuditEvent.AuditEventType.CHANGE_REASON, null, true, null, clock.getCurrentTime(), reason);
         }
 
-        if (saveResult != null) {
-            saveToDisk(lastSaveRequest.uri, lastSaveRequest.formComplete, lastSaveRequest.updatedSaveName, lastSaveRequest.viewExiting);
+        if (saveResult.getValue() != null) {
```
lognaturel: `null` here would mean that there was an attempt to save a change reason without a form save first having been triggered, right? Mostly making sure I'm following the flow.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -92,109 +99,99 @@ public String getReason() {
...
-                    saveResult.postValue(new SaveResult(SaveResult.State.SAVING, progress));
-                });
+            public void onProgressPublished(String progress) {
+                saveResult.postValue(new SaveResult(SaveResult.State.SAVING, saveRequest, progress));
             }
 
             @Override
-            protected void onPostExecute(org.odk.collect.android.tasks.SaveResult saveResult) {
-                saveToDiskTaskComplete(saveResult, clock.getCurrentTime());
+            public void onSaveComplete(SaveToDiskTaskResult saveToDiskTaskResult) {
```
lognaturel: The save isn't complete if there was an error? Rename to something like `onSaveTaskComplete`?

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -100,18 +100,18 @@ public SaveResult saveForm(FormSaver.ProgressListener progressListener) {
...
-                saveResult.setSaveResult(validateStatus, markCompleted);
-                return saveResult;
+                saveToDiskTaskResult.setSaveResult(validateStatus, markCompleted);
+                return saveToDiskTaskResult;
             }
         } catch (Exception e) {
             Timber.e(e);
 
             // SCTO-825
             // that means that we have a bad design
```
lognaturel: I don't think this comment adds anything. It's (now) clear that the error message is being saved and that it's a catch-all case. The ticket number is from a forking organization's private bugtracker. Might be good to remove while you're cleaning up.

collect_app/src/main/java/org/odk/collect/utilities/Clock.java
lognaturel: Sounds great to me.

collect_app/src/main/java/org/odk/collect/utilities/Clock.java
```diff
@@ -0,0 +1,10 @@
...
+
+/**
+ * An object that exposes the current time to it's client. Useful for decoupling
```
lognaturel: Your good friend "its" asks you not to replace it with "it is". 😂 (sorry, pet peeve)

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -240,12 +238,12 @@ boolean isViewExiting() {
         private final SaveRequest saveRequest;
         private final FormSaver formSaver;
 
-        private final WeakReference<Listener> listenerRef;
+        private final Listener listener;
```
lognaturel: Add some @NonNull annotations?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -317,7 +314,7 @@ public void allowSwiping(boolean doSwipe) {
 
     private BackgroundLocationViewModel backgroundLocationViewModel;
     private IdentityPromptViewModel identityPromptViewModel;
-    private ChangesReasonPromptViewModel changesReasonPromptViewModel;
+    private FormEntryViewModel formEntryViewModel;
```
lognaturel: Given that there are still multiple view models for this activity, does it make sense for one of them to match the activity name? Is this more of a formSaveViewModel?

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    private static class SaveRequest {
+
+        private final boolean formComplete;
+        private final boolean viewExiting;
+        private final String updatedSaveName;
+        private final Uri uri;
+
+        SaveRequest(boolean formComplete, boolean viewExiting, String updatedSaveName, Uri uri) {
```
lognaturel: This may seem petty but I find the order of parameters here make it hard to understand what's important. I'd rename `uri` to `instanceContentUri` and put it first since that's what is requested to be saved.

lognaturel (changes requested):
A few small things for your consideration inline. Overall, a nice change.

For 8bb379c, it would have been helpful to have some of the straightforward moves/renames split out so they didn't distract from the meaningful changes. I'm pretty confident that I followed everything but that particular commit was hard to review.

There are a lot of cases. The viewmodel tests do a nice job of documenting and testing state changes but it's a little scary that there's no end-to-end coverage for any of them.

Some things I think should be QAed:
- fill in some answers, kill Collect, make sure that recovery from savepoint works as expected
- change constraint behavior setting to at end rather than on swipe, violate some constraints, try to finalize
- constraint error, required question blank, encryption error with and without changes reasons
- save a large form, cancel part way

Longer-term, I wonder whether there are some reasonable Robolectric tests that could be added for some of these since it's such critical behavior.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
lognaturel: Ah, I was getting confused by `SaveToDiskTask` no longer actually being an `AsyncTask`.

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
seadowg: Ach sometimes I hate how smart IntelliJ is!

collect_app/src/main/java/org/odk/collect/android/formentry/audit/FormEntryViewModel.java
seadowg: Yeah that's much clearer. I'll change.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: This was eventually deleted.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Yuuuup. Will fix.

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
seadowg: Yeah I really should have renamed that at that point. Will take care of that.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Yeah. It doesn't make sense for that to happen but we get a warning in the code otherwise as it is technically possible.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Yeah really could just be `onComplete` given the listener is specific to the task. I'll change it.

collect_app/src/main/java/org/odk/collect/android/tasks/SaveToDiskTask.java
seadowg: Will do

collect_app/src/main/java/org/odk/collect/utilities/Clock.java
seadowg: ARRRRGGGGHHHH. I feel like I spend so much of my life thinking about this but still get it wrong.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Sounds good.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah I couldn't find a name I was happy with but I think I'll go with your suggestion.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: No I agree actually. Will change.

seadowg:
Thanks for the comments @lognaturel. Responded to them all and will do some cleanup to address them.

Sorry about the chunky commit. I tried to take it as step-by-step as possible but ended up with one big change I had to make to keep from having commits with failing build/tests (which is usually how I define a viable commit). Should have left you a warning not to go commit by commit because of that though.

collect_app/src/main/java/org/odk/collect/utilities/Clock.java
lognaturel: I always say "it is"/"they are"/"we are" in my head before picking. But don't worry, I'll get great joy out of continuing to point them out if need be. 😉

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Actually just inlining that method.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Actually not sure we need these here especially. It would be nice to have them everywhere but also not something I see as practical (unless we had a lint for it). This is also a private class so isn't really a "public" API. Did you have a specific reason for the suggestion?

lognaturel: Listeners can be optional but not in this case so it feels like a nice safeguard if someone needs to eventually be in this code. It seems like I've usually seen null checks around listener usage.

Also, when you had the `WeakReference` wrapper, there were clear null checks. So the diff that took that wrapper away made it very apparent that having a null listener would be a problem.

Don't feel strongly about it, though.

lognaturel (approved):

lognaturel:
Thanks! @seadowg has some good QA ideas in the PR description and I added some additional ones at https://github.com/opendatakit/collect/pull/3576#pullrequestreview-344955540.

lognaturel:
@opendatakit/testers please prioritize this one! It's big, it's risky, it's likely to conflict and it has another PR that depends on it already.

mmarciniak90:
@seadowg, @kkrawczyk123 and I are working on testing this PR and we noticed a serious problem. It's so critic so we inform about it immediately and we will continue our testing.

The problem concerns the status of forms.

Repro steps:
1. Start filling form
2. Click on device back button 
3. Choose `Save Changes`
4. Status of saved form is **finalized**

Repro steps 2:
1. Start filling form
2. Go to end
3. **Unselect** `Mark form as finalized`
4. Click `Save Form and Exit`
5. Status of saved form is **finalized**

Repro steps 3:
1. Go to `General Settings` -> `Form management` -> unselect `Default to finalized`
2. Start filling form
3. Go to end
4. `Mark form as finalized` is unselected
5. Click `Save Form and Exit`
6. Status of saved form is **finalized**

If the auto-send option is active, unfinalized forms are sent because of incorrect status.

seadowg:
@mmarciniak90 great find! It looks like we don't have any test coverage for these cases so I'll get working on that and a fix.

seadowg:
@mmarciniak90 I've fixed the finalized problem

collect_app/src/androidTest/java/org/odk/collect/android/formentry/FormFinalizingTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.activities.MainMenuActivity;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.android.support.pages.SaveOrIgnoreDialog;
+
+@RunWith(AndroidJUnit4.class)
+public class FormFinalizingTest {
```
lognaturel, line 20: @opendatakit/testers as you consider more regression checks to add to your list and/or the automated suite, please make sure you check these out. 😊

mmarciniak90:
@seadowg I noticed unexpected behavior in case:

1. Start form with **required** question
2. Do not fill the required question
3. Open hierarchy view and click `Go To End`
4. Click `Save Form and Exit`
5. The required question appears and it is marked on red
6. Click on the `Save` icon

The `Save` icon closes the form. User is moved to the main menu. 
Correct behavior is to save the form but stay in it.

We are continuing our testing.

seadowg:
@mmarciniak90 that case should be fixed now. Great find. There were problems with any case where you "save and exit", it doesn't work for some reason and then you try and save with the disk icon.

kkrawczyk123:
Verified on Androids: 4.4, 5.1, 8.1 and 10. 

Verified cases: 
- finalizing form
- saving form
- default to finalize from form and from settings
- finalizing and saving for encrypted form
- finalizing and saving for encrypted form without instanceID
- autosend form settings Google sheet/Aggregate
- autosend form form Google sheet/Aggregate
- auto delete from settings
- auto delete from form
- required question
- identify user screen
- tranck reason on changes 
- validate upon swiping
- validate until finalizing
- form with submission url
- preloading questions
- form with remembring previously entered values feature
- encrypted form with previously entered values feature

mmarciniak90:
**Tested with success**

Verified on Android: 4.2, 6.0, 7.0, 9.0

@lognaturel we used all our ideas for test cases. Even if we missed some case I feel that it is better to merge it and have the opportunity to use the app with this change testing other PRs.

lognaturel:
Thank you, @mmarciniak90 and @kkrawczyk123 for the thorough QA and @seadowg for being so quick to fix issues that came up. 👍 

I agree, @mmarciniak90, let's keep exercising it.

PR #3579, opened 2020-01-10
Support default image answers

### LINKED ISSUE #84: Annotate Image widget with a fixed image instead of an image capture.
<a href="https://github.com/mitchellsundt"></img></a> **Issue by [mitchellsundt](https://github.com/mitchellsundt)**
_Thursday Jul 09, 2015 at 19:09 GMT_
_Originally opened as https://github.com/opendatakit/opendatakit/issues/727 (0 comment(s))_

---

Originally reported on Google Code with ID 726

```
Enhance the annotate widget so that it can use a fixed image (when itext map is used)
as the basis for the annotation.

Change would be to add a new appearance ('markup') and itext form ('markup-image')
for use in the 'upload' question rendering.  The 'markup-image' is copied and used
as the basis for the markup image:

<text id="cardetail">
    <value form="long">Please mark up the image below.</value>
    <value form="short">Car damage</value>
    <value form="markup-image">jr://images/cardetail.png</value>
</text>
...
<h:body>     
  <upload ref="my_damage" appearance="markup"  mediatype="image/*">
    <label ref="jr:itext('cardetail')"/>
  </upload>
</h:body>

------------
Annotate widget would be changed to detect the 'markup' appearance
and suppress the choose and take picture buttons. It would then use:

  FormEntryPrompt p;
  ...
  String markupURI = p.getSpecialFormQuestionText("markup-image");
  String markupImageFilename = ReferenceManager._()
                     .DeriveReference(markupURI).getLocalURI();
  File markupImage = new File(markupImageFilename );

To get the markupImage to copy into the instance folder and use as the underlying image
for the markup.





```

Reported by `mitchellsundt` on 2012-11-26 21:07:35

grzesiek2010:
Closes #84

I tested the implemented changes manually and added an espresso tests.

It's the only solution we agreed in #84 and https://forum.opendatakit.org/t/proposal-specify-a-fixed-image-to-be-annotated-or-signed/14954. I can't come up with anything simpler or better.

This change is only related to Image widgets so we can test only those. It just allows to use default image values. I don't find this change very risky so testing the attached form should be enough.

[defaultImage.zip](https://github.com/opendatakit/collect/files/4045793/defaultImage.zip)

No.


collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.rules.RuleChain;
+import org.odk.collect.android.activities.FormEntryActivity;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.pages.FormEntryPage;
+import org.odk.collect.android.test.FormLoadingUtils;
+
+import java.util.Collections;
+
+public class DefaultImageAnswerTest {
```
seadowg, line 19: I think we could drive these changes out in tests for the `Widget`s (Robolectric most likely) themselves rather than at an Espresso level. This would also let us test that the correct default image is loaded, as well as testing the `ImageView` appears. What do you think?

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
grzesiek2010: I was thinking about it. We could add a test for every single widget like I did for [readOnly](https://github.com/opendatakit/collect/pull/3551) option. The problem is that we always get answers using` getAnswerText()` (from Javarosa) so we don't know if such an answer comes from a default value or it's just a previously saved answer. So in fact such a test (in every single widget) would be for saved answers not default values. The only way to make sure we are testing default values is using a form.

seadowg: I'm not meaning that we test the whole of the default value behaviour - like you say that's something that JavaRosa is partly responsible for. I mean that we could have a test for driving out that the image is shown (in `BaseImageWidget`) if there is a default (`getAnswerText` returns an image reference). In pseudocode I'm thinking:

1. Create temp image file
2. Set up mock reference to default image (`Helpers.createMockReference`)
3. Set up prompt with reference returned from `getAnswerText()` (`MockFormEntryPromptBuilder`)
4. Create `BaseImageWidget` with prompt
5. Check the widget's image view has loaded the bitmap - I think you can do this using Robolectric's `ShadowBitmapDrawable` (`shadowOf((BitmapDrawable) imageView.getDrawable()).getPath()`)

As far as I can see this would drive out the changes without needing Espresso. I might be misunderstanding something though.

lognaturel: I believe that @grzesiek2010 is correct that what you've described, @seadowg, would not specifically test having an image default but would just check that an image is loaded when there's an image in the current form instance. I think it's a worthwhile test but it's not testing the new behavior here. In other words, I would expect it to pass on `master`.

The new behavior is doing something with a `jr://images/` URI and I think step 2 of what @seadowg has described wipes that out.

@seadowg's test would need to be augmented with some tests that fail today. I think that ideally we could add a component to the test that lets us verify that either a relative (local to the submission folder) URI or an absolute (`jr://images/` which means relative to the form definition folder) URI can be resolved and passed on to the image display machinery.

Also, I think the implementation is incorrect because it copies an image with a `jr://images/` URI to the submission folder which it shouldn't do.

seadowg: Hmmm I must be misunderstanding how it shiould work. My thinking was that kind of test would drive out the code in `getDefaultFilPath` that deals with `jr://images/`. Either way I think agree with how you think it should go!

lognaturel: Well, wait a minute, maybe I'm the one not getting it. How are you thinking a relative URI and `jr://images/` URI would be differentiated in what you described?

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
```diff
@@ -154,27 +157,25 @@ protected void addCurrentImageToLayout() {
             int screenHeight = metrics.heightPixels;
 
             File f = new File(getInstanceFolder() + File.separator + binaryName);
+            if (!f.exists()) {
+                f = addDefaultImageIfExists(f);
```
lognaturel: I think that here you can just try to use the reference manager to resolve the ref and go with that if it works. You don't want to copy the image and you don't want to change what value is in the instance (`binaryName`). The instance value should only change if the image is replaced (including if an image previously from `jr://images/` is annotated).

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
seadowg: I think what I'd missed is that it did different things on whether the URI started with `jr://images/ ` rather than only setting `defaultFilePath` if the URI started with `jr://images/ `. I was seeing an `if` rather than an `if-else`. I'm guessing you're saying we need two tests (one for each kind of URI) to drive out the new behaviour?

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
```diff
@@ -313,4 +314,32 @@ protected void launchActivityForResult(Intent intent, final int resourceCode, fi
...
+        FormController formController = Collect.getInstance().getFormController();
+        if (formController != null) {
+            if (binaryName.startsWith("jr://images/")) {
+                try {
+                    defaultFilePath = referenceManager.deriveReference(binaryName).getLocalURI();
+                } catch (InvalidReferenceException e) {
+                    Timber.w(e);
+                }
+            } else {
+                defaultFilePath = formController.getMediaFolder() + File.separator + binaryName;
```
lognaturel: This means you're also trying to manually get things from the media folder if a file has no prefix? Why? Also consider that if a `jr://file/` URI is set, for example, you'd end up with a path that looks like `/path/to/media/folder/jr://file/name`

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
lognaturel: > I'm guessing you're saying we need two tests (one for each kind of URI) to drive out the new behaviour?

Yes. And on `master`, the test using a `jr://images/` URI should fail.

I added some comments around the implementation which I don't think is quite right. There really is no "default" concept at the widget level since we can't tell the difference between a value from a form definition vs a submission that's being re-entered.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
grzesiek2010: If I don't copy the image it won't be saved in instance dir and it won't be uploaded to a server.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
grzesiek2010: Ok I assumed that we should support both with and without prefix. I'll get rid of it.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
lognaturel: That is the intent! You can read more about that in the forum thread. You'll see I wasn't on board initially but I was convinced.

grzesiek2010: Ok I fixed it, now it should be ok.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
grzesiek2010: I added some unite tests for each image widget but still I think it would be good to keep this espresso tests as well.

seadowg: I put together an example of how I **think** we could drive out the changes to the `ImageWidget`s using just Robolectric. It requires an update to our `createMockReference` method so it can use real files.

https://gist.github.com/seadowg/a9dd682666b2ec7b8f9d242f6553e9ea

I'm very interested in seeing how far we can go with testing Widgets in isolation as I think it would let us write a lot fewer Espresso tests. Can totally see that it might just feel easier/better to stick with the Espresso test here though.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
```diff
@@ -309,7 +309,7 @@ protected void launchActivityForResult(Intent intent, final int resourceCode, fi
         }
     }
 
-    private File getFile() {
+    File getFile() {
```
lognaturel, line 312: I think I've said something like this before before -- having to change the visibility of methods to add tests seems like a bit of a red flag. That probably means either the design isn't right or the tests aren't trying out a realistic path through the app.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/DefaultImageAnswerTest.java
lognaturel: @seadowg, when you say "drive out", you mean that in a test-driven development sense, right? As in write a test that describes the desired behavior which initially fails and use that test to get to a good code design? I just want to make sure that we're all on the same page. I do think that's a powerful approach that would be good to use more as opposed to writing some code and then fitting tests around it.

I think that both the tests that @grzesiek2010 added and the one @seadowg is proposing basically check to make sure that the `ReferenceManager` gets used to resolve a `jr://images` URI. @grzesiek2010's very much rely on the current structure of the code because the unit they test is the method that goes from an answer value in an instance to an actual path. What I like about @seadowg's in contrast is that it is at a higher level and verifies what the user ultimately cares about -- that the image that will be shown is the one they expect. It requires exposing a getter for the `imageView` but that change seems acceptable to me. The test remains high-level enough that it could support a refactor of the functionality it tests. @grzesiek2010, would a test like that make you feel comfortable dropping the Espresso test?

As I mentioned before, ideally we'd have an analogous test for an image without a `jr://images` prefix. I believe that would take a non-trivial refactor because of the dependency that `addCurrentImageToLayout` has on `FormController`.

seadowg: > when you say "drive out", you mean that in a test-driven development sense, right? As in write a test that describes the desired behavior which initially fails and use that test to get to a good code design?

Yes exactly. I find a nice trick for writing these after the fact is to comment out the changes in the code, write the tests for the behaviour, verify they they're red/fail and then bring the code back (uncomment) until you're green again.

> It requires exposing a getter for the imageView but that change seems acceptable to me.

Yeah this can seem weird given you'd usually avoid making something public just to test it - you want to test the "interface" and so as you were pointing out elsewhere it can be a sign that you need to change your design. In this case the `imageView` is "visible" to the user so actually is part of the "interface".

grzesiek2010: > would a test like that make you feel comfortable dropping the Espresso test?

yes I think so but if it's only about test I would vote for testing it and merging as is and then working on tests in a separate pr.

lognaturel: Works for me. Maybe @seadowg can take the test he's written, polish it up and send it in, also removing the Espresso test. Perhaps it will also be a reasonable time to pull out the `FormController` dependency and do the test for non-`jr://images` files?

seadowg: Sounds good to me!

kkrawczyk123:
**Tested with success:**

When I was testing, I have recognized separate issues:
https://github.com/opendatakit/collect/issues/3634
https://github.com/opendatakit/aggregate/issues/482
https://github.com/opendatakit/central-backend/issues/250

seadowg (approved):

PR #3580, opened 2020-01-10
Focus on textfield on ServerUrl, Username, Password and shows keyboard.

### LINKED ISSUE #3502: No focus on text field on Server URL / Username / Password dialogs, keyborad does not appear - Android 9.0
Collect v1.25.0 beta 2, 
Android v9.0, v10.0

#### Problem description
No focus on text field on Server URL / Username / Password dialogs, keyboard does not appear

1. User goes to General Settings -> Server
2. User clicks on URL / Username /  Password
3. Dialog appears but without the focus on text field and the keyboard does not appear

#### Expected behavior
Keyboard appears

be-wake:
Closes #3502 

The issue was that when clicked on preference(ServerUrl) it didn't set focus on edit text and show the keyboard
**Before:**
prb1
prb2

**After:**
sl1
sl2

I tried other methods(implementing Preference.OnPreferenceClickListener() and directly setting focus ) but only this works.
#### How does this change affect users? Describe intentional changes to behaviour and behaviour that could have accidentally been affected by code changes. In other words, what are the regression risks?

Now it will automatically set focus on edit text so user doesn't need to click again.

No, it does not need any specific testing.

No, it does not require any change in the document.


grzesiek2010:
Thanks for your effort @Vivever!

You have fixed three EditTextPreferences but we have some more in general settings so the solution is not complete.
I also don't like such tricks so now I'm wondering whether it make sense at all... Did you check maybe how EditTextPreferences work in other apps is it only our problem or maybe it's a default Android behavior for newer APIs? If it's a default behavior maybe we shouldn't change it.

grzesiek2010:
Thanks for your effort @Vivever!

You have fixed three EditTextPreferences but we have some more in general settings so the solution is not complete.
I also don't like such tricks so now I'm wondering whether it make sense at all... Did you check maybe how EditTextPreferences work in other apps is it only our problem or maybe it's a default Android behavior for newer APIs? If it's a default behavior maybe we shouldn't change it.

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
```diff
@@ -490,4 +490,21 @@ private void continueOnBackPressed() {
     public void doBack() {
         runGoogleAccountValidation();
     }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        serverUrlPreference.setOnPreferenceClickListener(preference -> {
```
seadowg: ~~This code looks right to me but I think these listeners should be setup in `addAggregatePreferences` along with the rest of the setup for each `Preference`.~~

I handn't noticed @grzesiek2010's comment. I think he's right that we should investigate the default behaviour here and understand if this is a problem just for our app or not. If we do go with this solution though I think it's good to move this listener setup to `addAggregatePreferences` along with the rest of the setup for each `Preference`.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
be-wake: @grzesiek2010 and @seadowg, Yes I had checked that  before pull request and that is the default behaviour

img

So, Should I remove the code?

seadowg: Yuck that's annoying. I'm not seeing that problem in Android settings but maybe they work around it as well. 

Personally I think it makes sense to keep these changes as it's definitely a better experience (in the most basic sense it cuts down on taps). What do you think @grzesiek2010?

grzesiek2010: I'm not sure but also not strongly against so you can proceed @Vivever but as I said we have more such `EditTextPreferences` to fix.

be-wake: @grzesiek2010, I will find them and fix them.

be-wake: @grzesiek2010 @seadowg  There was only one extra EditTextPreference (Admin Password in Admin Settings) which I have fixed. 
Other preferences are either checkbox or radio buttons.

seadowg: Thanks for looking at that @Vivever! As far as I can see there are still a few fields in "User and device identity" > "Form metadata".

As I said before it'd be better not to call `setPreferenceClickListener` in `onResume`. These should probably go with the rest of the preference setup in `ServerPreferencesFragment#addAggregatePreferences`. In other fragments it can just be in `onCreate`. You shouldn't need to `setPreferenceClickListener(null)` either as the preference object won't (as far as I know) outlive the Fragment in memory.

seadowg (changes requested):

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android: 10.0, 9.0. Verified also on Android 8.1, 7.0, 4.4

Verified cases
- all text fields when Aggregate was set as a server
- all text fields when Other was set as a server
- all text fields when Google Drive was set as a server
- rotate device orientation
- light/dark mode

grzesiek2010:
Thanks @Vivever

PR #3581, opened 2020-01-10
Restrict sending forms if error occurred

### LINKED ISSUE #1939: Collect tries to send all forms even if error occurred precluding to continue sending
Collect v1.13 and earlier

#### Problem description
Collect tries to send all forms even if error occurred precluding to continue sending

1. User sends a lot of forms
2. Error occurred (need to check few error types)
3. User is informed about error after sending all forms

#### Example of real use case:
1. User uses ODK Aggregate server
2. User sends 1000 forms which have declared google sheets `<submission action="https://docs.google.com/spreadsheets/d/***/edit?usp=sharing" method="form-data-post"/>`
3. Error occurred 
screenshot_2018-02-26-19-43-40

4. User need to wait wait a few minutes because Collect tries to send all forms

#### Expected behavior
* Sending process should be stopped after the first error

chidauri:
Closes #1939 

[backup1249.zip](https://github.com/opendatakit/collect/files/4047210/backup1249.zip) provided by @mmarciniak90 

The output message will contain a particular failure only once even if that occurs for different instances, unlike earlier when a particular failure would show up multiple times for multiple instances.
**And because we don't try to send all forms , all forms don't show the message `Sending failed...`, only the first form shows error, i will change this behaviour if desired.**
This is how it looks currently if try to send all those 1006 forms + sample all widgets form:

Screenshot_20200111-000821

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/InstanceServerUploaderTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/InstanceServerUploaderTask.java
@@ -25,6 +25,7 @@ import org.odk.collect.android.upload.UploadException;
 import org.odk.collect.android.utilities.WebCredentialsUtils;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -54,6 +55,7 @@ public class InstanceServerUploaderTask extends InstanceUploaderTask {
     @Override
     protected Outcome doInBackground(Long... instanceIdsToUpload) {
         Outcome outcome = new Outcome();
+        HashSet<String> badUrls = new HashSet<>();
 
         InstanceServerUploader uploader = new InstanceServerUploader(httpInterface, webCredentialsUtils, new HashMap<>());
         List<Instance> instancesToUpload = uploader.getInstancesFromIds(instanceIdsToUpload);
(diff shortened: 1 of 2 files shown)


grzesiek2010:
Hmmm I'm not in favor of fixing that issue...
It assumes that we have a lot of submissions of the same form then if we have the same exception for at least two it should stop. What if we have mixed list of forms that we want to upload?
Anyone finds such a change to be valuable? @lognaturel @seadowg

seadowg:
Hmmm yeah I do see your point @grzesiek2010. It also doesn't look like we've had any users complain about this in the wild right?

grzesiek2010:
> It also doesn't look like we've had any users complain about this in the wild right?

right

chidauri:
> Hmmm I'm not in favor of fixing that issue...
> It assumes that we have a lot of submissions of the same form then if we have the same exception for at least two it should stop. What if we have mixed list of forms that we want to upload?
> Anyone finds such a change to be valuable? @lognaturel @seadowg

I agree that mixed list would be a problem that's why i have proposed a solution which will just look for those errors which are solely due to the destination url set(whether it is from intent, or in the form, or in the settings) regardless of the form information, and we only stop uploading those which contain these type of urls. Though if users don't complain about this, it can be ignored.

grzesiek2010:
Yes but it's possible to specify a different [submission url](https://xlsform.org/en/#settings-worksheet) for each form, you don't need to use the one from server settings, then this solution which checks previous submission urls won't work as well.

chidauri:
Hmm then i think this check would only be useful for server settings and we wouldn't need this as this is not where users will often make mistake.

grzesiek2010:
There are many potential causes of `UploadException` supporting only [one](https://github.com/opendatakit/collect/pull/3581/files#diff-9e1980840a34e15c6a2e5845b9ccbe0dR94) and assuming that all selected forms have the same submission url doesn't make sense to me at all.
I would vote for closing both the issue and this pr. Any thoughts @lognaturel @seadowg?

seadowg:
@grzesiek2010 I don't see a problem with that

grzesiek2010:
> I don't see a problem with that

You mean you I don't see a problem with closing both or merging this pr?

seadowg:
@grzesiek2010 sorry that was massively ambiguous. I meant closing both.

grzesiek2010:
Ok so sorry @chidauri but we need to close this pr nevertheless, we appreciate your engagement and encourage you to work on other issues.

PR #3585, opened 2020-01-11
Used Number Picker to change line width in Annotate and Draw Widget

### LINKED ISSUE #3089: tool to change the line width in the annotate and draw widget
Originally reported on Google Code with ID 971

```
What tool are you using?
Annotate widget & Draw widget

What is frustrating, troubling, annoying or lacking?
I'd like to see an option to change the line width or at least have the line width
default to a thinner line.

How would you like it to work?
next to set color have an option for change line width.


NOTE: the core team has limited ability to make software enhancements.
We rely heavily on community-contributed enhancements to the tools. If this
is something you very much need, consider hiring a developer to make the
changes with the stipulation that they work with the core team to specify
the details of the change and contribute the change back to the project
under an Apache 2 license.

```

Reported by `oconnorholmesbailey` on 2014-02-05 03:14:44

SaumiaSinghal:
Closes #3089 

 I tested it on Android 9.0.

Issue#3089

I added a `floatingActionButton`, that allows the user to change stroke width and choose any value from 1 to 100, using the standard NumberPicker android widget. 

No regression risks, as of now.

AllWidgetsForm or any form having draw or annotate widget.

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/DrawActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/DrawActivity.java
@@ -31,6 +31,7 @@ import android.view.animation.AccelerateDecelerateInterpolator;
 import android.view.animation.OvershootInterpolator;
 import android.widget.AdapterView;
 import android.widget.ListView;
+import android.widget.NumberPicker;
 
 import com.google.common.collect.ImmutableList;
 import com.rarepebble.colorpicker.ColorPickerView;
@@ -104,6 +105,8 @@ public class DrawActivity extends CollectAbstractActivity {
         fabActions = findViewById(R.id.fab_actions);
         final FloatingActionButton fabSetColor = findViewById(R.id.fab_set_color);
         final CardView cardViewSetColor = findViewById(R.id.cv_set_color);
+        final FloatingActionButton fabSetLineWidth = findViewById(R.id.fab_set_line_width);
+        final CardView cardViewSetLineWidth = findViewById(R.id.cv_set_line_width);
         final FloatingActionButton fabSaveAndClose = findViewById(R.id.fab_save_and_close);
         final CardView cardViewSaveAndClose = findViewById(R.id.cv_save_and_close);
         final FloatingActionButton fabClear = findViewById(R.id.fab_clear);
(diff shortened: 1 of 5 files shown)


seadowg:
@SaumiaSinghal thanks for looking at this! 

I think the general flow in this solution makes sense but I do think we should alter the design a bit. 

I think the issue (#3089) really should have had a `needs discussion` label on it so we could have discussed designs before you did any work so apologies about that.

P.S. I don't really think we should be using a FAB for the list of options here as they are generally meant to represent the primary action on any given page (like composing a new email on an inbox screen for example). Given that is already in the app I'll add it to our list of [potential design changes](https://docs.google.com/document/d/1X_YYYziaW58tacGSHIpNbDlbRjAtGNzduJ6X1pyw19M/edit).

seadowg:
@SaumiaSinghal thanks for looking at this! 

I think the general flow in this solution makes sense but I do think we should alter the design a bit. For me, the problem here is that it will be hard for an enumerator to get a sense of line width they are selecting based on the number (How big is 100? How small is 10?). I think a way to solve this is to give the enumerator a limited number of options and name them Small, Medium, Large. These could still be presented in a dialog but we could use Confirmation Dialog with selection (the Material examples are [here](https://material.io/components/dialogs/#confirmation-dialog)).

I also think it would be good for a new feature like this to have tests driving it out. Personally I'd look at testing `DrawView` to drive out the flow and checking selecting an option changes the line width.

I think the issue (#3089) really should have had a `needs discussion` label on it so we could have discussed designs before you did any work so apologies about that.

P.S. I don't really think we should be using a FAB for the list of options here as they are generally meant to represent the primary action on any given page (like composing a new email on an inbox screen for example). Given that is already in the app I'll add it to our list of [potential design changes](https://docs.google.com/document/d/1X_YYYziaW58tacGSHIpNbDlbRjAtGNzduJ6X1pyw19M/edit).

seadowg (changes requested):
As above.

SaumiaSinghal:
Thanks @seadowg for reviewing the PR. Yes, it is not convenient to get the sense of line width by selecting from an enumerator. How about I look into how we can implement this feature in the app, keeping in mind your suggestion of providing the user a number of options like Small, Medium, Large and so on, through a confirmation dialog, and present the UI here. And then, after finalizing, maybe I can finally implement it in the app. 

I also agree with your idea of not using FAB for the mentioned reason, but that can be addressed later in a separate PR.

seadowg:
@SaumiaSinghal that sounds great!

grzesiek2010:
Don't you think it would be good to have color picker and line weight picker as a one dialog?
The lib we use https://github.com/martin-stone/hsv-alpha-color-picker-android doesn't seem to provide such an option but maybe there is a different one we could use, or maybe we could file an issue there to add it?

seadowg:
@grzesiek2010 that's a good idea. I think looking into existing solutions like that makes sense as well @SaumiaSinghal. Once you have some ideas about the design/flow we can all discuss!

seadowg:
@SaumiaSinghal just to keep everything organized I'm going to close this PR. Let's move any more discussion to #3089 and I'll add the "needs discussion" label there.

PR #3589, opened 2020-01-14
main menu: change delete form button text

eeshugerman:
Changes button text from "Delete Saved Form" to "Delete Form". This button is used to delete both "saved" and "blank" forms, so it's confusing that the button text specifies "saved".

----
There is no GitHub issue for this. Should I open one first?

---

I'll be honest: I don't have a development environment set up for this app. However, this is the only place in the source code where `"Delete Saved Form"` appears.

I can't think of a better value for this string than "Delete Form".

N/A

No

No

### DIFF
--- a/collect_app/src/main/res/values/strings.xml
+++ b/collect_app/src/main/res/values/strings.xml
@@ -95,7 +95,7 @@
     <string name="location_accuracy">Accuracy is %1$s m</string>
     <string name="main_menu">Main Menu</string>
     <string name="main_menu_details">Data collection made easier…</string>
-    <string name="manage_files">Delete Saved Form</string>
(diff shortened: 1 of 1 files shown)


grzesiek2010:
Hi @eeshugerman 

>There is no GitHub issue for this. Should I open one first?

Yes please do that first. If you had done that first We would have let you know that we manage the translations using [Transifex](https://www.transifex.com/opendatakit/collect/) so even if we need this change it probably should be changed then not in our code.

More about translations you can read here https://github.com/opendatakit/collect#contributing-translations

I'm closing for now.

eeshugerman:
https://github.com/opendatakit/collect/issues/3591

PR #3590, opened 2020-01-14
PMD: Excluded Singular Field Rule

### LINKED ISSUE #2932: PMD: Exclude the SingularField rule.
This issue is developer-facing only.

I propose that we stop enforcing the SingularField rule.  This rule suggests that if a field is only used in one method, it could be a local variable inside the method instead of a field of the class instance.

#### Rationale

1. The message for this rule looks like:

    > Perhaps 'activity' could be replaced by a local variable.

    It doesn't make sense for a "Perhaps" suggestion to be enforced so hard that it breaks the build.

2. The field might be defined on the class because the author intended it to be meaningful as a member of the class, and so it makes the most sense to keep it there.

3. This rule can generate needless work.  It typically comes into force when a field is used in two methods, say A and B, and then method A is edited or deleted so that the field is used in only method B.  But it is quite possible that A will be edited again in such a way that it uses the field, or we might add another method C that needs to use the field.  In this situation, the rule has forced us to do extra work that only has the effect of making a future edit also require more work.  This situation arises, for example, in test classes as test methods are added and removed.

In some situations, this rule may motivate a helpful edit.  But overall it doesn't seem to generate much value and may even have negative net value, so let's consider turning it off.

SaumiaSinghal:
Closes #2932

I tested it by running `./gradlew pmd`. 

I excluded the Singular Field rule for the reasons stated by @zestyping.

No regression risks

Developer-facing issue

No

### DIFF
--- a/config/pmd-ruleset.xml
+++ b/config/pmd-ruleset.xml
@@ -81,6 +81,7 @@
         <exclude name="NcssCount" />
         <exclude name="NPathComplexity" />
         <exclude name="SignatureDeclareThrowsException" />
+        <exclude name="SingularField" />
         <exclude name="SimplifyConditional" />
         <exclude name="SwitchDensity" />
         <exclude name="TooManyFields" />


seadowg:
@SaumiaSinghal sorry about this but I'm not sure I actually support the underlying issue. I'll tag @lognaturel to see if she disagrees with me (happy to be outvoted) but otherwise I think we should close this and have a discussion in #2932.

lognaturel:
Thanks for helping us clean up some of our undecided issues, @saumiasinghal! I agree with @seadowg and have added my reasoning on the issue.

SaumiaSinghal:
That's okay @seadowg, @lognaturel, It was not much of a task. :)

PR #3592, opened 2020-01-15
Stop repeat dialog closing when app is backgrounded

### LINKED ISSUE #3322: Repeat group dialog disappears after locking phone
This started on the [forum](https://forum.opendatakit.org/t/add-new-prompt-desappears-after-unlocking-the-screen/21564).

Collect v1.23.3

#### Problem description

When prompted to add a repeat group, locking your phone (or navigating to another app) and then returning to the app shows a blank screen without the dialog prompt. You can swipe back and forward to be reprompted.

1. Start filling out a blank form with a repeat (an example: [why.xlsx](https://github.com/opendatakit/collect/files/3546228/why.xlsx))
2. When the repeat group prompt ("Add one more group?") appears press the home button or lock the phone
3. Return to the app

You'll just see a blank screen.

#### Expected behavior

I'd expect to still be prompted to add a repeat group (or not). The current situation could be very confusing, especially if you had locked the phone to deal with something else and then come back to it a few minutes later.

#### Other information 

It looks like the offending line of code is [here](https://github.com/opendatakit/collect/blob/4a8d2990cf5788e0b471455d5b192ca1cf68c377/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L2104). It would be intersting to know why we dismiss dialogs and if this weirdness could potentially be addressed with a redesign (I've noticed a discussion happened around this in #809).

seadowg:
Closes #3322. Depends on #3576.

This removes code in `FormEntryActivity` that was dismissing dialogs in `onPause`. As far as I can see there was no reason for that to happen but if someone needed that code speak now!

As well as fixing the issues this PR:
* Removes the use of `showDialog`/`dismissDialog` from `FormEntryActivity`
* Use our `ProgressDialogFragment` for form loading and saving

Played around with repeats and backgrounding. I wasn't able to write a test for this as it requires backgrounding the app which we can't do with Espresso and a Robolectric test for `FormEntryActivity` isn't really feasable right now.

As long as the code I deleted wasn't needed in someway then it's certainly the simplest solution.

Should just fix the issue. Good to test repeats, form loading and saving as the dialogs have all had code changes in someway. Would also be good to investigate any other dialog involving scenarios during form entry to check that the code I've removed wasn't needed for something.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -479,7 +479,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 } else {
                     Timber.w("Reloading form and restoring state.");
                     formLoaderTask = new FormLoaderTask(instancePath, startingXPath, waitingXPath);
-                    showFormLoadingDialogFragment();
+                    DialogUtils.showIfNotShowing(FormLoadingDialogFragment.newInstance(), getSupportFragmentManager());
                     formLoaderTask.execute(formPath);
                 }
                 return;
@@ -614,7 +614,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         }
 
         formLoaderTask = new FormLoaderTask(instancePath, null, null);
-        showFormLoadingDialogFragment();
+        DialogUtils.showIfNotShowing(FormLoadingDialogFragment.newInstance(), getSupportFragmentManager());
         formLoaderTask.execute(formPath);
     }
 
@@ -853,8 +853,9 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                  * Android 1.6) we want to handle images the audio and video
                  */
 
-                ProgressDialogFragment.newInstance(null, getString(R.string.please_wait))
-                        .show(getSupportFragmentManager(), ProgressDialogFragment.COLLECT_PROGRESS_DIALOG_TAG);
+                ProgressDialogFragment progressDialog = new ProgressDialogFragment();
+                progressDialog.setMessage(getString(R.string.please_wait));
+                progressDialog.show(getSupportFragmentManager(), ProgressDialogFragment.COLLECT_PROGRESS_DIALOG_TAG);
 
                 mediaLoadingFragment.beginMediaLoadingTask(intent.getData());
                 break;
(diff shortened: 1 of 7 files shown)


lognaturel:
Now that #3576 is merged, is this ready for review?

seadowg:
@lognaturel I need to rebase to kill the extra commits as it was dependent on #3576

seadowg:
@lognaturel ok that's good to go now!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1640,18 +1636,6 @@ public void showView(View next, AnimationType from) {
         }
     }
 
-    // Hopefully someday we can use managed dialogs when the bugs are fixed
-    /*
-     * Ideally, we'd like to use Android to manage dialogs with onCreateDialog()
-     * and onPrepareDialog(), but dialogs with dynamic content are broken in 1.5
```
lognaturel, line 1648: Mm. Nice to see this comment gone.

lognaturel (approved):
I can't reproduce the issue on `master` anymore so I'm not sure that the `onPause` code was the actual culprit.

That said, it's good that this pulls code out of `FormEntryActivity`, that it moves towards standardizing how dialogs are used in that activity, and that it adds tests. I also can't think of a reason to dismiss all dialogs on pause and like I said in the issue, I suspect it's boilerplate.

@opendatakit/testers this affects any dialog that is shown while filling a form. It would be good to try weird things like backgrounding the app while a large form is loading and coming back to it.

seadowg:
Fixed conflicts!

kkrawczyk123:
**Tested with success!** 
Verified on Androids: 4.2, 4.4, 5.1, 7.1, 9.0 and 10.

**Verified cases:**

add group dialog on a few different forms with repeats:
- extinguish the screen
- minimize an app
- devices rotation
- don't keep activities

remove group dialogs:
- extinguish the screen
- minimize an app

remove this response dialog:
- extinguish the screen
- minimize an app

exit form dialog:
- extinguish the screen
- minimize an app

large form loading dialog:
- minimize an app: dialog restored from the background, the form was able to load
- extinguish the screen: the same result, the form was able to load

spinner widget, select multiple dialog:
- extinguish the screen
- minimize an app

bearing widget dialog:
- extinguish the screen
- minimize an app

range picker integer widget dialog:
- extinguish the screen
- minimize an app

date widgets dialogs:
- extinguish the screen
- minimize an app

rank widget: I have noticed crashes but also on master so I have reported as a separate issue in #3615 . 

sorry, location providers are disabled dialog:
- extinguish the screen
- minimize an app

PR #3593, opened 2020-01-15
Changed User-Agent string

### LINKED ISSUE #3253: User-Agent string lists system before product, contrary to RFC 7231 recommendation

SaumiaSinghal:
Closes #3253 

I confirmed by checking the Timber output.

I changed the order of product identifiers, in decreasing order, as suggested by  [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.5.3), in User String.

No regression risks.

No specific form required.

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -213,10 +213,12 @@ public class Collect extends Application {
      * This deviates from the recommended format as described in https://github.com/opendatakit/collect/issues/3253.
      */
     public String getUserAgentString() {
-        return String.format("%s %s/%s",
-                System.getProperty("http.agent"),
+        String[] httpAgent = System.getProperty("http.agent").split("\\(");
+        return String.format("%s %s/%s %s",
+                httpAgent[0],
                 BuildConfig.APPLICATION_ID,
-                BuildConfig.VERSION_NAME);
+                BuildConfig.VERSION_NAME,
+                "(" + httpAgent[1]);
     }
 
     public boolean isNetworkAvailable() {
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -61,6 +61,7 @@ import org.odk.collect.android.preferences.GeneralSharedPreferences;
 import org.odk.collect.android.preferences.PrefMigrator;
 import org.odk.collect.android.tasks.sms.SmsNotificationReceiver;
 import org.odk.collect.android.tasks.sms.SmsSentBroadcastReceiver;
+import org.odk.collect.android.utilities.AndroidUserAgent;
 import org.odk.collect.android.utilities.FileUtils;
 import org.odk.collect.android.utilities.LocaleHelper;
 import org.odk.collect.android.utilities.NotificationUtils;
@@ -212,12 +213,6 @@ public class Collect extends Application {
      *
      * This deviates from the recommended format as described in https://github.com/opendatakit/collect/issues/3253.
      */
-    public String getUserAgentString() {
-        return String.format("%s/%s %s",
-                BuildConfig.APPLICATION_ID,
-                BuildConfig.VERSION_NAME,
-                System.getProperty("http.agent"));
-    }
 
     public boolean isNetworkAvailable() {
         ConnectivityManager manager = (ConnectivityManager) getInstance()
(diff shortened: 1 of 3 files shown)


seadowg:
@SaumiaSinghal just seen that the unit tests are failing. Can you confirm you definitely ran `./gradlew checkAll` for your most recent changes?

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -213,10 +213,12 @@ public String getVersionedAppName() {
      * This deviates from the recommended format as described in https://github.com/opendatakit/collect/issues/3253.
      */
     public String getUserAgentString() {
```
seadowg: I think it'd be really nice to pull this method into it's own helper/use case object (`AndroidUserAgent` for instance that implements a `UserAgent` interface). That would let us pass this object to where it's needed and easily fake/mock it in cases where we might not have all the static state set up (that might be the fix for the broken tests). 

I'd considered also having it be tested but I think the test and the code would end up basically being the same formatting code.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
SaumiaSinghal: Hie @seadowg Thanks for the suggestion and sorry for late replying. I think the tests failed because I tried to split the host identifier. But yes we can have this method in a separate helper class `AndroidUserAgent` with a `User Agent` interface. Should I go about it?

seadowg: Yeah I think it's a nice bit of restructuring we can do as part of this work.

seadowg (approved):

seadowg:
For testing I think we just want to verify the user agent looks correct.

mmarciniak90:
@seadowg I need more instructions. I really don't know how to test it manually. Is it even possible? Maybe someone should create an automated test for this new method `getUserAgent`

seadowg:
@mmarciniak90 ah sorry I thought we'd done tests on HTTP requests before but I'm probably wrong! For this lets just test form downloading/submission on a couple of different devices to make sure we don't run into any problems. I can do verification of the user agent itself.

mmarciniak90:
**Tested with success**

I did not notice regression.
Verified on Android: 10.0, 9.0, 8.1, 7.0, 6.0, 5.1, 4.4, 4.2(only cases with Aggregate and Central)

Verified cases:
- downloading from Aggregate
- sending to Aggregate
- downloading from Google Drive
- sending to Google Sheet
- downloading from Central
- sending to Central

seadowg:
@SaumiaSinghal it looks like these changes have broken our integration tests and so our CI is now red. We don't run `./gradlew connectedDebugAndroidTest` against PRs but we trust that you've done it if you check the box in the PR description. Please make sure to run it against your latest changes in future!

SaumiaSinghal:
Hie @seadowg! yes, the tests failed locally, but when I pushed the commit and thought of discussing it here, the circle ci tests passed and I thought that maybe, there's a problem locally. Yes, I should have reported that. I will remember that in future.

PR #3597, opened 2020-01-18
Manifest cleanup in preparation for FileProvider changes

lognaturel:
I am looking at designing the API to support external applications that currently directly read from or write to `/sdcard/odk` and won't be able to after #3360. There are currently several Android Studio checks failing and those make it hard to see clearly in the manifest. I'd like to make these quick cleanup changes before starting work.

There should be no behavioral changes. I ran all tests and also made sure that the app actually launches.

I split each change into its own commit to explicitly explain why they're being made.

There should be no behavior change. The riskiest part is probably 8573897 but the documentation is very explicit about not needing this block and devices that don't support OpenGL 2 must be quite rare by now.

No.

I have filed https://github.com/opendatakit/docs/issues/1168 to document backup behavior since it seems like something that users would like to explicitly know about.

### DIFF
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -9,7 +9,8 @@ the specific language governing permissions and limitations under the License.
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    package="org.odk.collect.android">
+    package="org.odk.collect.android"
+    tools:ignore="GoogleAppIndexingWarning">
 
     <uses-feature
         android:name="android.hardware.location"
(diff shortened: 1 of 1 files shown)


collect_app/src/main/AndroidManifest.xml
```diff
@@ -87,7 +78,8 @@ the specific language governing permissions and limitations under the License.
         android:label="@string/app_name"
         android:largeHeap="true"
         android:supportsRtl="true"
-        android:usesCleartextTraffic="true">
+        android:usesCleartextTraffic="true"
+        android:allowBackup="true">
```
seadowg, line 82: I was initially confused why this was added but then I remembered it's true by default. I'm guessing the intention is just to make it explicit here?

seadowg (approved):
Looks good. Just one question tiny question.

collect_app/src/main/AndroidManifest.xml
lognaturel: Yes, exactly. There's an Android Studio check for it and I hadn't actually fully understood the ramifications so I think it's helpful to see it explicitly listed.

seadowg:
I feel like we can merge this without QA. Will do when the build is green again!

PR #3601, opened 2020-01-21
Return error message immediately when an exception occurs during downloading forms

### LINKED ISSUE #3600: Misleading message when downloading a form fails
Collect v1.25.1 and older versions

#### Problem description
Downloading a form we catch a few exceptions:
[TaskCancelledException](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java#L154)
[Exception](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java#L160)
[RuntimeException](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java#L183)

However only in case of `TaskCancelledException` we throw that exception and stop running the rest of code in the method. That leads to misleading messages like the one described on the forum https://forum.opendatakit.org/t/blank-surveys-wont-download-please-help/24371
The problem was not related to media files at all but the message says it's a cause.

Try to download a form from the server mentioned in the topic: https://forum.opendatakit.org/t/blank-surveys-wont-download-please-help/24371

grzesiek2010:
Closes #3600 

I tested the changes with the server mentioned in https://forum.opendatakit.org/t/blank-surveys-wont-download-please-help/24371/1

Once an exception occurs we should stop downloading a form and performing other operations in that method.

It should improve the messaged displayed for a user when downloading a form fails. As I described in the issue the part related to media files was misleading and not related to the problem at all.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FormDownloader.java
@@ -158,7 +158,7 @@ public class FormDownloader {
             // do not download additional forms.
             throw e;
         } catch (Exception e) {
-            message += getExceptionMessage(e);
+            return message + getExceptionMessage(e);
         }
 
         if (stateListener != null && stateListener.isTaskCanceled()) {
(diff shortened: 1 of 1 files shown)


seadowg (approved):

kkrawczyk123:
**Tested with success**

checked that media failing message is not visible, compared with master.

PR #3602, opened 2020-01-21
Update repeat language

seadowg:
This updates the language used in the repeat dialog to be in line with the prototype found [here](https://github.com/seadowg/collect-design/tree/master/prototypes/repeat-tweaks).

I've also made it so the repeat dialog text will not be different based on the number of previous repeats - it will always be `Add "<repeat-name>"?`. Before it would read as "Add new..." or "Add another..." depending on if a repeat has already been added. One of the pieces of feedback (during [research](https://github.com/seadowg/collect-design/blob/master/research/repeats.md)) was that "new" had confused enumerators as it made them think they had actually already added one. As this is feels like a bit more of an experimental change and I still think we should have the other versions (`Add new...` and `Add another...`) translated I've put this behind a feature flag that will make it a one line change to revert if we get negative feedback on it during beta.

Ran through a couple of different repeat forms to check nothing has broken.

As before this was driven out by a research round.

Should only effect the repeat dialog but good to have play with repeat forms ("Birds" for example).


collect_app/src/main/java/org/odk/collect/android/application/FeatureFlags.java
```diff
@@ -0,0 +1,10 @@
+package org.odk.collect.android.application;
+
+public class FeatureFlags {
+
+    public static final boolean STATIC_ADD_REPEAT_LANGUAGE = true;
```
seadowg: This is just using a very simple static feature flag. If we were looking at having tests for both behaviours I'd generally look at having a `FeatureFlag` object that could be passed to objects that need it.

lognaturel:
Apologies for the bikeshedding. Trying to balance getting this through and making sure we are thoughtful. As you know I can lean too far in the latter direction. 😫

Can you say more about your yes/no choice? That seems unusual for alert dialogs. I don't think Material Design guidelines explicitly say anything about action naming but examples include a verb for the primary action. See https://material.io/components/dialogs/#usage. I'm not seeing anything explicit from Apple either but again, examples have verbs for the primary action: https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/dialogs/. Same for this very box that I am typing in right now!

I think that it's common not to carefully read dialog text so it's helpful when the buttons provide all the needed information. In this case, `DO NOT ADD / ADD` would also mean an easier transition from prior versions (`DO NOT ADD / ADD GROUP`). I also wonder whether some languages have different words for yes/no depending on context.

I very strongly agree that the "group" part has to go. But what about `NO / YES` do you prefer to `DO NOT ADD / ADD`? Was there specific feedback on those button labels other than about "group"?

More support for using verbs in dialogs [here](https://usabilitypost.com/2008/08/30/usability-tip-use-verbs-as-labels-on-buttons/), [here](http://ignorethecode.net/blog/2008/10/31/nobody-reads/), [here](https://medium.com/@wutronic/which-to-choose-cancel-ok-ok-cancel-or-yes-no-ec119861eb9b).

seadowg:
@lognaturel hmmm yeah I actually agree with you. It feels to me like it should `Cancel` / `Add`. How does that feel?

I don't think this is bike shedding! Language is super important as we saw from the research. I would like to push these conversations further back into when we look at a mockup however. Lets look at having a quick feedback session next time we run through this kind of process. Also maybe not sending people the mockup right before Christmas would help 🎄

lognaturel:
> It feels to me like it should Cancel / Add. 

That wouldn’t be right. “Cancel” would mean staying at the end of the last repeat instance with no action. What the negative button does is move out of the current repeat. 

> I would like to push these conversations further back into when we look at a mockup however.

Agreed that would have been ideal. I did not see any problems and it wasn’t until I got into high concentration review mode that I saw challenges. I don’t think you explicitly called attention to the button label changes or why you did them. That might have helped.

seadowg:
@lognaturel I've updated the button text. I've also removed the feature flagging concept as we agreed that the change isn't as risky as I'd initially thought.

I've realized there other places that the word "Group" appears (like in the hierarchy screen). I think we should review those and follow up with changes rather than lumping everything into 1 PR as we didn't really get much feedback about the language in that area.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddNewRepeatDialog.java
```diff
@@ -0,0 +1,37 @@
...
+        this.repeatName = repeatName;
+    }
+
+    @Override
+    public AddNewRepeatDialog assertOnPage() {
+        onView(withText(getTranslatedString(R.string.add_repeat_question, repeatName))).check(matches(isDisplayed()));
+        return this;
+    }
+
+    public FormEntryPage clickOnAddGroup(FormEntryPage destination) {
```
lognaturel: `clickOnAddRepeat` or just `clickOnAdd`?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddNewRepeatDialog.java
```diff
@@ -0,0 +1,37 @@
...
+        clickOnString(R.string.add_repeat);
+        return destination;
+    }
+
+    public FormEntryPage clickOnDoNotAddGroup(FormEntryPage destination) {
```
lognaturel: Same as above.

lognaturel (approved):
There's a test-only naming change that I think would be mildly helpful but would be good to go to QA ASAP so we can get translation going.

grzesiek2010 (approved):

lognaturel:
@opendatakit/testers This is relatively low risk and would be great to get to translators as soon as possible. If you want to take a break from #3576 or are done with it, this is a great next one to look at! 😊

It would be good to verify forms with field lists, repeats and nested repeats, a repeat with field-list appearance...

mmarciniak90:
> I very strongly agree that the "group" part has to go.

@seadowg, I wonder that I understand it correctly. 
I see that the question part has removed this word. 
New line ` <string name="add_repeat_question">Add \"%s\"?</string>` is added and it does not contain "group". 
But what with options? English version has `<string name="add_repeat">Add</string>`
but eg. Spanish `<string name="add_repeat">Agregar Grupo</string>`
So the "group" word is included.


---

Verified on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 9.0 and 10.0

|v1.25.1|PR|
|---|---|
|Screenshot_20200126-155310|Screenshot_20200126-155224|



Verified cases
- 8 languages, including one RTL
- `+` icon on hierarchy view
- `next back` navigation buttons
- light/dark mode
- rotating on dialog
- removing group

Verified forms:

lognaturel:
> But what with options? English version has <string name="add_repeat">Add</string>
but eg. Spanish <string name="add_repeat">Agregar Grupo</string>
So the "group" word is included.

That is confusing, @mmarciniak90, thanks for bringing it up. In the case of `add_repeat_question` the name of the resource has changed so that's why it's going through to all languages. In the case of the affirmative action, the resource name is still `add_repeat` so the old translations with the word "group" have remained across languages. That's ok -- when the strings are pushed to Transifex, the Transifex system will notice that the English text has changed and request a translation in other languages. This PR is time sensitive because we want to make sure translators have time to get to it before the release.

mmarciniak90:
@lognaturel thanks for the explanation.

PR #3604, opened 2020-01-21
Fix rendering of google map labels on top of offline layers

### LINKED ISSUE #3004: When Google Maps is used as mapping engine, don't render labels on top of offline layers
Collect v1.21

#### Problem description
When offline tiles are used with the Google Maps engine, the default basemap labels are rendered on top which may conflict with elements in a custom basemap or layer.

See discussion at https://forum.opendatakit.org/t/behavior-of-offline-tiles-with-osm-vs-google-maps/18704

1. Set mapping engine to Google Maps in Settings
1. Add offline tiles - https://docs.opendatakit.org/collect-offline-maps
1. Note that labels are rendered on top of the offline tiles

#### Expected behavior
Labels should not be rendered on top of offline tiles.

chidauri:
Closes #3004 

Tested on android 9, with 
[countries-raster.zip](https://github.com/opendatakit/collect/files/4092637/countries-raster.zip)

As discussed on the forums, google maps labels are rendered dynamically so our only option is to turn them off completely in case of offline layers.

No

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
@@ -463,10 +463,16 @@ public class GoogleMapFragment extends SupportMapFragment implements
             referenceOverlay = this.map.addTileOverlay(new TileOverlayOptions().tileProvider(
                 new GoogleMapsMapBoxOfflineTileProvider(referenceLayerFile)
             ));
-            map.setMapStyle(new MapStyleOptions(getResources().getString(R.string.google_map_labels_switch, "off")));
+            switch_google_map_labels("off");
         } else {
-            map.setMapStyle(new MapStyleOptions(getResources().getString(R.string.google_map_labels_switch, "on")));
+            switch_google_map_labels("on");
         }
+    }
+
+    // Turns off google map labels for offline layers to work correctly
+    private void switch_google_map_labels(String state) {
+        String style = String.format(" [ { featureType: all, elementType: labels, stylers: [ { visibility: %s } ] } ]", state);
+        map.setMapStyle(new MapStyleOptions(style));
     }
 
     private LatLngBounds expandBounds(LatLngBounds bounds, double factor) {
--- a/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
@@ -33,6 +33,7 @@ import com.google.android.gms.maps.model.Circle;
 import com.google.android.gms.maps.model.CircleOptions;
 import com.google.android.gms.maps.model.LatLng;
 import com.google.android.gms.maps.model.LatLngBounds;
+import com.google.android.gms.maps.model.MapStyleOptions;
 import com.google.android.gms.maps.model.Marker;
 import com.google.android.gms.maps.model.MarkerOptions;
 import com.google.android.gms.maps.model.Polyline;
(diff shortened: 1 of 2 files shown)


grzesiek2010 (changes requested):
The solution seems to work so that's great thanks! However keeping that google_map_labels_switch in strings that are translatable is not a good idea. You could move that to `untraslated.xml` but since it's used just in one place you can use local variables. You can also consider moving the code to a separate method with a meaningful name to make the code clearer.

chidauri:
Is it okay now?

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -462,9 +463,18 @@ private void loadReferenceOverlay() {
...
                 new GoogleMapsMapBoxOfflineTileProvider(referenceLayerFile)
             ));
+            switch_google_map_labels("off");
+        } else {
+            switch_google_map_labels("on");
         }
     }
 
+    // Turns off google map labels for offline layers to work correctly
+    private void switch_google_map_labels(String state) {
```
grzesiek2010: Generally we should avoid using comments. Comments should be used if it's a tricky case that requires an explanation. In 99% cases you can just come up with a better method name. In this case it would be for example `setLabelsVisibility()`; ahh and we don't use _ in method names.

grzesiek2010 (changes requested):

grzesiek2010 (approved):

mmarciniak90:
@chidauri @lognaturel @grzesiek2010 
I noticed one behavior I want to discuss.

I applied the offline layer and I see that labels are not visible. 
But when I zoom to the current location or look at more area I do not see any labels



The same area without offline layer looks like:


I'm not sure that losing this information is expected behavior. Tell me what you think, please

grzesiek2010:
Yeah it would be better to enable labels back in such a case but I'm not sure if there is an easy way to handle it. @chidauri Could you try to solve it?

chidauri:
@grzesiek2010 yes i didn't find any option in the api reference to do that, we would've to do manually, like getting zoom levels of tile overlays and then processing accordingly.
(There is a method [`isVisible()`](https://developers.google.com/android/reference/com/google/android/gms/maps/model/TileOverlay.html#public-boolean-isvisible-), which says **this does not return whether the tile overlay is actually within the screen's viewport** so i think it is of no use in this case.)

mmarciniak90:
@chidauri I noticed that labels are still visible on `Terrain` and `Hybrid` Google map styles.

Use case:
1. User goes to `General Settings` -> `Maps`
2. User sets Google map style as `Terrain` or `Hybrid`
3. User sets some offline layer
4. User goes to form and opens Geowidget with map

chidauri:
Hmm.. it is written [here](https://developers.google.com/maps/documentation/android-sdk/hiding-features) that styling works only on `normal` map type, @grzesiek2010 what should we do then?

lognaturel:
> @grzesiek2010 yes i didn't find any option in the api reference to do that, we would've to do manually, like getting zoom levels of tile overlays and then processing accordingly.

Thanks for looking into it, @chidauri! I think that this case is ok. The test offline layer is a bit strange since it only exists when very zoomed out. In real usage we would expect offline layers that are specific to an area where data collectors are and so exist at higher zoom levels.

> it is written here that styling works only on normal map type

How about we just document that users should be on the street style if they want to use offline layers with Google maps.

mmarciniak90:
@lognaturel, sounds good!

**Tested with success**

Tested on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1, 9.0, 10.0

Labels are completely not visible only for `Streets` and `Satellite` style.

PR #3605, opened 2020-01-22
Add AdminSettingsTest, AboutPageTest and update ResetApplicationTest

kkrawczyk123:
@grzesiek2010  I have pushed AboutPageTest but the second case currently does not work as the icons from AboutListAdapter.java don't have tags. Could you please add it?

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/regression/AboutPageTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/regression/AboutPageTest.java
@@ -1,0 +1,111 @@
+package org.odk.collect.android.regression;
+
+import androidx.test.runner.AndroidJUnit4;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.R;
+import org.odk.collect.android.support.pages.AboutPage;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.assertion.ViewAssertions.matches;
+import static androidx.test.espresso.matcher.ViewMatchers.withText;
+import static org.odk.collect.android.support.matchers.DrawableMatcher.withImageDrawable;
+import static org.odk.collect.android.support.matchers.RecyclerViewMatcher.withRecyclerView;
+
+//Issue NODK-234
+@RunWith(AndroidJUnit4.class)
+public class AboutPageTest extends BaseRegressionTest {
+
+    @Test
+    public void when_rotateScreenOnAboutPage_should_notCrash() {
+        //TestCase1
+        new MainMenuPage(rule)
+                .clickOnMenu()
+                .clickAbout()
+                .rotateToLandscape(new AboutPage(rule))
+                .assertOnPage()
+                .scrollToOpenSourceLibrariesLicenses();
+    }
+
+    @Test
+    public void when_openAboutPage_should_iconsBeVisible() {
+        //TestCase2
+        new MainMenuPage(rule)
+                .clickOnMenu()
+                .clickAbout()
+                .assertOnPage();
+
+        onView(withRecyclerView(R.id.recyclerView)
+                .atPositionOnView(0, R.id.title))
+                .check(matches(withText(R.string.odk_website)));
+
+        onView(withRecyclerView(R.id.recyclerView)
+                .atPositionOnView(0, R.id.summary))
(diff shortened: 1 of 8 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/regression/AboutPageTest.java
```diff
@@ -0,0 +1,111 @@
...
+        new MainMenuPage(rule)
+                .clickOnMenu()
+                .clickAbout()
+                .rotateToLandscape(new AboutPage(rule))
+                .assertOnPage()
+                .scrollToOpenSourceLibrariesLicenses();
+    }
+
+    @Test
+    public void when_openAboutPage_should_iconsBeVisible() {
```
seadowg, line 33: I'm not sure we need this as an automated test. In my opinion, it's unlikely that the icons would dissapear and if they did it wouldn't a huge problem. Do you think we could just delete this test?

seadowg (changes requested):
Just one comment. Let me know what you think!

collect_app/src/androidTest/java/org/odk/collect/android/regression/AboutPageTest.java
kkrawczyk123: I think that automation of all cases that are performed manually in the regression cycle is valuable and can shorten the time of performing the whole cycle. It is something that we are checking anyway so why not check that automatically instead?

seadowg: Right that's a good point. I think we probably need to start thinking about seperating out `regression` from the other Espresso tests so we can have one faster set to run locally as devs but keep the value of having all these regression test cases automated - potentially running then before PR merge or once a day or something.

seadowg (approved):

seadowg:
@kkrawczyk123 I'm just going to check these test on Test Lab before merging.

collect_app/src/androidTest/java/org/odk/collect/android/regression/AboutPageTest.java
lognaturel: > I think we probably need to start thinking about seperating out regression from the other Espresso tests so we can have one faster set to run locally

I think this is a really good idea. Is introducing a new Gradle task that runs the non-regression tests a reasonable way to do this?

seadowg: @lognaturel I think there is a way to filter on annotations but not sure about packages. I'll have a look into this.

seadowg:
@kkrawczyk123 I'm seeing failures in `when_openAboutPage_should_iconsBeVisible` on Test Lab:

```java
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
at org.odk.collect.android.support.matchers.DrawableMatcher$1.matchesSafely(DrawableMatcher.java:25)
at org.odk.collect.android.support.matchers.DrawableMatcher$1.matchesSafely(DrawableMatcher.java:17)
at androidx.test.espresso.matcher.BoundedMatcher.matches(BoundedMatcher.java:73)
at androidx.test.espresso.matcher.ViewMatchers.assertThat(ViewMatchers.java:527)
at androidx.test.espresso.assertion.ViewAssertions$MatchesViewAssertion.check(ViewAssertions.java:105)
at androidx.test.espresso.ViewInteraction$SingleExecutionViewAssertion.check(ViewInteraction.java:425)
at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:288)
at androidx.test.espresso.ViewInteraction$2.call(ViewInteraction.java:272)
at android.os.Handler.handleCallback(Handler.java:790)
at android.app.ActivityThread.main(ActivityThread.java:6494)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

Everything fine locally?

grzesiek2010:
> Everything fine locally?

No, that's why @kkrawczyk123 asked me to implement required changes in the code:

> @grzesiek2010 I have pushed AboutPageTest but the second case currently does not work as the icons from AboutListAdapter.java don't have tags. Could you please add it?

What I've just done in https://github.com/opendatakit/collect/pull/3605/commits/0e09c7da48e8fb965f18c94a2c1e0fa37a340046

PR #3606, opened 2020-01-23
Implemented classes to handle database operations for both sdcard and scoped storage

grzesiek2010:
This pr is supposed to be the first step towards Scoped Storage.
- I created `StorageManger` class where I moved operations on the storage. I know the class name itself is not good and generally we should avoid names like manger, handler etc. but I'm not good at it so if you can come up with something better please help. Of course I could move to this class much more code from other classes but I wanted to avoid making this pr gigantic so I refactored only the most important code.
- For reviewers probably @lognaturel @seadowg most of the changes I made using refactoring tools from Android Studio so I find them pretty safe, I mean commits from the first one to 
d4180db
The most dangerous commits are: cec6ac68f4385f107e5c382dd5163742026dda60, 5256e8df8a7fb4290723a6a51c09bea5624300d1, 88ab3a63f4f346f23d4f0ac1d440967584340fa3, 34feac0579bb82d7a43a983af99c2a5215f47fac, 89f5d23b4b548db9ae8f235537e36f9b1fc0eeb9
it's because they read/write database paths, for now absolute paths are used but the code is ready to use relative paths and absolute paths depending on whether migration to scoped storage took place or not. That makes the code a bit complicated (we will be able to make it better once we switch to scoped storage forever).

- tried to be super careful especially implementing changes for five mentioned above commits.
- added automated tests for StorageManager class
- ran instrumented tests on firebase test lab
- and finally I tested the changes manually especially checked if paths saved in databases are correct

First of all we didn't have any class responsible for operations on storage. A lot of the code was placed in Collect class (my impressions is that contains a mix of different code and it has been getting bigger and bigger because we treated it kind of like FormEntryActivity)
Second of all we need to handle somehow scoped storage and start using relative paths once we switch to it, StorageManager is responsible for that.

At this moment it shouldn't course any changes, everything should work like before. However it's a risky pr especially in terms of regression in managing file paths. There is no one functionality we would need to test, everything might be affected. Please make sure that database paths in forms, instances, itemsets databases are saved properly.

Since we need to test paths saved in databases any form and instance would allow us to do that, at least to test forms and instances databases but there is also itemsets.db and to test it we need a form with external itemsets like:
[select_one_external.zip](https://github.com/opendatakit/collect/files/4105155/select_one_external.zip)

No.


collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -68,4 +67,8 @@ public static String getMetadataPath() {
     public static String getCachePath() {
         return getOdkRoot() + File.separator + ".cache";
     }
+
+    public static String getOfflineLayers() {
```
lognaturel: Should be `getOfflineLayersPath` to match the others.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -77,4 +75,8 @@ public static String getSETTINGS() {
     public static String getTmpfilePath() {
```
lognaturel: Do these really belong here?

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -40,50 +40,50 @@ public static void createODKDirs() throws RuntimeException {
...
+                getFormsDirPath(),
+                getInstancesDirPath(),
+                getCacheDirPath(),
+                getMetadataDirPath(),
+                getOfflineLayersDirPath()
             };
     }
 
-    public static String getOdkRoot() {
+    public static String getMainODKDirPath() {
```
lognaturel: By naming conventions, it should be `Odk`. How about `getStorageRootPath`?

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -54,8 +54,26 @@ public static void createODKDirs() throws RuntimeException {
             };
     }
 
+    private static String getStoragePath() {
+        return isScopedStorageUsed()
+                ? getPrimaryExternalStorageFilePath()
```
lognaturel: I don't think primary/secondary is easy to understand. How about:
* `getScopedExternalFilesDir` or `getAppExternalFilesDir` 
* `getUnscopedExternalFilesDir` or `getRootExternalFilesDir`

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -43,6 +42,14 @@ public static void createODKDirs() throws RuntimeException {
         }
     }
 
+    private static boolean isStorageAvailable() {
+        return getStorageState().equals(Environment.MEDIA_MOUNTED);
+    }
+
+    private static String getStorageState() {
+        return Environment.getExternalStorageState();
```
lognaturel: Inlining this in `isStorageAvailable` would be clearer, I think.

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -258,7 +259,7 @@ private void processItemSets(File formMediaDir) {
             ida.open();
             // get the database entry (if exists) for this itemsets.csv, based
             // on the path
-            final Cursor c = ida.getItemsets(csv.getAbsolutePath());
+            final Cursor c = ida.getItemsets(relativeCsvFilePath);
```
lognaturel: This looks like it would have different behavior from what was there before. If this is based on the path, I'd expect needing to go through getFormFilePath (the method that builds database paths based on whether or not scoped storage is used).

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -508,7 +508,7 @@ public void setActivityResult(int requestCode, int resultCode, Intent intent) {
         this.intent = intent;
     }
 
-    private void readCSV(File csv, String formHash, String pathHash) {
+    private void readCSV(File csv, String formHash, String pathHash, String relativeCsvFilePath) {
```
lognaturel: `relativeCsvFilePath` and `csv` are redundant. Can you do a substring? Generally not understanding 59b6e64.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2027,7 +2027,7 @@ public void onClick(DialogInterface dialog,
                                         languages[whichButton]);
                                 String selection = FormsColumns.FORM_FILE_PATH
                                         + "=?";
-                                String[] selectArgs = {formPath};
+                                String[] selectArgs = {StorageManager.getFormFilePath(StorageManager.getRelativeFormFilePath(formPath))};
```
lognaturel: Why does it have to be relativized first? Can't `getFormFilePath` deal with it?

collect_app/src/test/java/org/odk/collect/android/storage/StorageManagerTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private void mockUsingSdCard() {
+        doReturn(false).when(storageManager).isScopedStorageUsed();
+    }
+
+    @Test
+    public void when_scopedStorageNotUsed_should_sdCardPathsBeUsed() {
+        mockUsingSdCard();
+
+        assertEquals("/storage/emulated/0/odk", storageManager.getMainODKDirPath());
```
lognaturel: Quick stylistic note you don't need to do anything about but I wanted to share for your consideration:

`assertThat(storageManager.getMainODKDirPath(), is("/storage/emulated/0/odk"))` reads more like English. Here we have to go to the end of the assertions to find out what code is under test and then look at what the expected value is.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -124,7 +124,7 @@ public void recordMigrationToScopedStorage() {
     }
 
     // TODO the method should be removed once using Scoped storage became required
-    public String getCacheFilePath(String relativePath) {
+    public String getCacheFilePathToStoreInDatabaseBasingOnRelativePath(String relativePath) {
```
lognaturel: The naming here is really tricky. It's something I had flagged when the methods were first introduced so I'm glad to see you looking for something more specific. I must say that even for me who loves words, these are a bit long. 😊 Here are some other ideas:
- `getCacheFilePathForDatabase`
- `getDatabasePathFromRelativePath(Enum fileType (e.g. CACHE, FORM, INSTANCE), String relativePath)`
- `relativePathToDatabasePath(String relativePath, Enum fileType)`

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
```diff
@@ -0,0 +1,186 @@
...
+                ? getPrimaryExternalStorageFilePath()
+                : getSecondaryExternalStorageFilePath();
+    }
+
+    String getPrimaryExternalStorageFilePath() {
+        File primaryStorageFile = Collect.getInstance().getExternalFilesDir(null);
+        if (primaryStorageFile != null) {
+            return primaryStorageFile.getAbsolutePath();
+        }
+        return "";
```
lognaturel: This seems like it would be catastrophic. It means no media is available, right? Would having an empty string here just cause things to crash downstream?

lognaturel (changes requested):
I think this is really great to do now as its own PR. It will be really nice to have all the pre vs post migration logic in a single place. 👍

It's good that the new class is instantiable rather than being a collection of static methods. This PR currently creates it at points of usage so it doesn't help reduce the dependencies that client classes manage. That is, where there was previously a dependency on `Collect`, now there's a dependency on the new class rather than the dependency being passed in. I think that's ok -- at least we're set up to build the dependency externally and pass it in as the client code gets restructured and tested so it's a step in the right direction.

I agree that the `StorageManager` name isn't ideal because it doesn't set an expectation about what the class does. I wonder whether it might be hard to name because it does too many things. If the `createODKDirs` method were separated out, I think it would feel more cohesive. So this new class could just be responsible for building paths and its name could be something like `PathCreator`, `StoragePathCreator`, `PathResolver` or `PathProvider`. The `createODKDirs` method could go into a class with a name like `DirectoriesCreator` or `StorageInitializer` that has a `StoragePathCreator` (or whatever it's named) constructor dependency.

I also wonder whether the structure of having a group of methods for each subdirectory/type of file is hard to understand and maintain. I feel responsible for this because I remember thinking for a moment about alternatives when we were speaking and saying it'd probably be better to have separate methods. Seeing it now, the redundancy seems problematic. I tend to prefer a more domain-driven structure and I'd find it ideal to see all of this functionality in domain objects like `Instance` and `Form` so that you could ask a `Form` instance for paths to its files (probably by delegating out to something like a `StoragePathCreator`). That is a big refactor on the caller side and I don't think it's worth doing now. For now, I wonder whether we could make this general structure easier to follow by introducing something like a `Subdirectory` enum. I had trouble describing it so I took a quick stab at what that might look like at https://github.com/opendatakit/collect/compare/master...lognaturel:hm-storage. I also tried some renames there and broke down the commits so that my reasoning would be very clear.

To collapse the methods, I first did the extract parameter refactor on all variants. Then I used the change signature refactor on each variant to put parameters in the same order and give them all the same name.

Before you change anything, I think it would be good to see what @seadowg thinks of the high-level structure and whether he prefers what you already have or has an alternative idea that addresses the redundancy.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -325,21 +324,20 @@ public int delete(@NonNull Uri uri, String where, String[] whereArgs) {
...
+                                        .getColumnIndex(FormsColumns.FORM_FILE_PATH))));
+                                deleteFileOrDir(storageManager.getAbsoluteFormFilePath(c.getString(c
+                                        .getColumnIndex(FormsColumns.FORM_MEDIA_PATH))));
 
                                 try {
                                     // get rid of the old tables
                                     ItemsetDbAdapter ida = new ItemsetDbAdapter();
                                     ida.open();
-                                    ida.delete(c.getString(c
+                                    ida.delete(storageManager.getRelativeFormFilePath(c.getString(c
```
lognaturel: This is going to do a database query, right? So it seems like it should match whatever is in the database? I'd expect no transformation needed here.

seadowg:
> So this new class could just be responsible for building paths and its name could be something like PathCreator, StoragePathCreator, PathResolver or PathProvider. The createODKDirs method could go into a class with a name like DirectoriesCreator or StorageInitializer that has a StoragePathCreator (or whatever it's named) constructor dependency.

Yeah I'd agree with this. I'd vote for `StoragePathProvider` and `StorageInitializer`. That structure would be nice to move to in this PR I think. In that setup the `StorageInitializer` could actually depend on the `StoragePathProvider`.

I think down the line I'd love to see us not having to deal with as much file reading and writing in our UI objects (like our Activity objects). Potentially there would be some kind of `FormStore`/`FormRepo` interface (and similar for other domain objects) that would abstract the I/O involved. This refactor makes that kind of world a lot more possible though so this a great step!

seadowg:
> For now, I wonder whether we could make this general structure easier to follow by introducing something like a Subdirectory enum

Hmmm. Yeah I'm not sure. It's definitely cleaner now but I'm having a hard time predicting where the model around all this is going to go exactly and if that will be better or worse for the future. Maybe go for it  but if it's a huge amount of work it might not be worth it?

lognaturel:
> I'd vote for StoragePathProvider and StorageInitializer.

Sounds great to me. I like the repeated `Storage` prefix that carves out what part of the universe this functionality refers to.

> I think down the line I'd love to see us not having to deal with as much file reading and writing in our UI objects (like our Activity objects). Potentially there would be some kind of FormStore/FormRepo interface (and similar for other domain objects) that would abstract the I/O involved. This refactor makes that kind of world a lot more possible though so this a great step!

Absolutely. I think `FormsRepository`/`InstanceRepository` starts that and we have a note in our state of the union so it just needs to happen as we work on the client code (the callers of the methods here is what I mean).

> I'm having a hard time predicting where the model around all this is going to go exactly and if that will be better or worse for the future. Maybe go for it but if it's a huge amount of work it might not be worth it?

That's a good point. I don't feel strongly about it either way. My two concerns with the current structure are:
1. Given the redundancy, it'd be possible to introduce really subtle bugs. For example, let's say that the method for getting a database path for cache files changes ever so slightly and the form/instance ones stay the same. We might not notice cache files aren't correctly being resolved. I don't think that this class will need to change much so it's theoretical but if we keep the redundant methods we have to be very careful about modifications.
1. The names of the methods to get db paths currently look like `getCacheFilePathToStoreInDatabaseBasingOnRelativePath` which is an accurate name.  That suggests to me that more parameters would clarify the meaning.

That said, I'm fine with leaving as-is with slightly improved names.

Over to you, @grzesiek2010! Consider pulling me or @seadowg into a pairing session if that would help finalize things. 🚀

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
lognaturel: If you don't want to go the enum route (totally fine), what about `getCacheDbPath`/`getInstanceDbPath`/etc? I think that given the parameter name, it's not critical to state in the name that the incoming path must be relative. But `getCacheDbPathFromRelativePath` also works for me.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: It is called getOfflineLayersDirPath() (which matches all the others) in the final version so probably you looked at a commit where I moved the static filed from Collect class here and created a method everything using Android Studio so it wasn't even a name that I created.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: I have a method called `getStoragePath()` which returns a path to sd card/scoped storage so `getStorageRootPath` would be too similar... wee need to emphasize it's odk dir.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: In my opinion yes but you can try to change it :)

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: Primary and secondary are words used across Android doc. I feel like Scoped Storage is something more, it represents the whole change which forces use to use the primary external storage so to be honest I prefer my naming. What do you think @seadowg?

collect_app/src/test/java/org/odk/collect/android/storage/StorageManagerTest.java
grzesiek2010: Yes sometimes I use `assertThat` sometimes `assertEquals` I should choose one approach (`assertThat`).

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
lognaturel: Can you link out to the Android doc you're referring to? I think that in Android primary would refer to internal storage and secondary to the (first) physical sdcard.

I think that means the primary/secondary naming is inaccurate if it intends to match Android so it's important to change:
* [`Environment.getExternalStorageDirectory()`](https://developer.android.com/reference/android/os/Environment.html#getExternalStorageDirectory()): "Return the primary shared/external storage directory."
* [`Context.getExternalFilesDir()`](https://developer.android.com/reference/android/content/Context.html#getExternalFilesDir(java.lang.String)): "Returns the absolute path to the directory on the primary shared/external storage device where the application can place persistent files it owns"

So the paths are on the same media. The difference is whether the storage area is scoped or not / belongs to the app or not.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: I agree this method is not needed.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: Yeah I agree that it could be improved. I'll rethink it.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Because `getFormFilePath()` (final name `getFormFilePathToStoreInDatabaseBasingOnRelativePath()`) expects a relative path as a parameter. Theoretically I could have a method in which as a passed value I would accept both relative and absolute paths and then deal with it but I'm not sure it's better, somehow I decided to do that in this way. I'll look at it again.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: Ahh ok i think I read this answer https://stackoverflow.com/a/29404440/5479029 which is obviously wrong. Sorry I think you are right and the names you proposed make sense.

collect_app/src/main/java/org/odk/collect/android/storage/StorageManager.java
grzesiek2010: Yes, I hardcoded it for testing and it returns:
Screenshot_2020-01-28-16-55-32
https://github.com/opendatakit/collect/pull/3606/files#diff-294c525d328b50b03236414b1f1ef7a7R33

Probably we should improve the message?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Fixed, now one method deals with it.

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
grzesiek2010: I changed the approach a bit so now those methods responsible for returning database paths deal with both relative and absolute paths so those changes are not needed.

collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
grzesiek2010: Like below. I changed the approach a bit so now those methods responsible for returning database paths deal with both relative and absolute paths so those changes are not needed.

grzesiek2010:
@lognaturel @seadowg 
I implemented required changes.
I needed to take a step back and one of the things I changed was improving methods responsible for returning database paths. In the previous version those methods accepted relative paths whatforced me in some cases to use something like `storagePathProvider.getFormDbPathFromRelativePath(storagePathProvider.getRelativeFormPath(formfilepath));`
now those methods deal with both relative and absolute paths.
I also decided to use your approach with Subdirectory enum class, I wasn't sure at first but after implementing those changes I mentioned above I ended up with a lot of duplicated code which I was able to remove thanks to your approach.

I think you can take a look at this pr again.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
grzesiek2010: Right, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/dao/FormsDaoTest.java
```diff
@@ -126,7 +128,7 @@ public void getFormsCursorTest() {
 
     @Test
     public void getFormsCursorForFormFilePathTest() {
-        Cursor cursor = formsDao.getFormsCursorForFormFilePath(Collect.FORMS_PATH + "/Miramare.xml");
+        Cursor cursor = formsDao.getFormsCursorForFormFilePath(storagePathProvider.getDirPath(StorageSubdirectory.FORMS) + "/Miramare.xml");
```
lognaturel, line 131: This is going to do a db request. Shouldn't it use the `getDbPath` method? As written, this code will always build an absolute path but when we flip the switch to scoped storage, there won't be absolute paths in the db to query against.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -200,19 +203,19 @@ public synchronized Uri insert(@NonNull Uri uri, ContentValues initialValues) {
             values.put(FormsColumns.MD5_HASH, md5);
 
             if (!values.containsKey(FormsColumns.JRCACHE_FILE_PATH)) {
-                String cachePath = Collect.CACHE_PATH + File.separator + md5
+                String cachePath = storagePathProvider.getDirPath(StorageSubdirectory.CACHE) + File.separator + md5
```
lognaturel: This should be a db path, right?

lognaturel (changes requested):
The changes look great, thank you! I think there are some issues with how the tests get paths for db requests. I just marked one in `FormsDaoTest` but it looked like several tests had this issue. I think I also caught one similar problem in `FormsProvider`. 

Other than that, I think we can proceed unless @seadowg has additional changes he'd like to see now.

collect_app/src/androidTest/java/org/odk/collect/android/dao/FormsDaoTest.java
grzesiek2010: getFormsCursorForFormFilePath() used here deals with it.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
grzesiek2010: Ohh yes generaly that line was redundant at all because it was just:
```java
String cachePath = storagePathProvider.getDirPath(StorageSubdirectory.CACHE) + File.separator + md5 + ".formdef";
values.put(FormsColumns.JRCACHE_FILE_PATH, storagePathProvider.getCacheDbPath(cachePath));
```

so first I an absolute path was created and then used in getCacheDbPath() which deals with relative paths as well.
That means we can just use:

```java
values.put(FormsColumns.JRCACHE_FILE_PATH, storagePathProvider.getCacheDbPath(md5 + ".formdef"));
```

Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/StorageInitializer.java
```diff
@@ -0,0 +1,41 @@
...
+import timber.log.Timber;
+
+public class StorageInitializer {
+    /**
+     * Creates required directories on the SDCard (or other external storage)
+     *
+     * @throws RuntimeException if there is no SDCard or the directory exists as a non directory
+     */
+    public void createODKDirs() throws RuntimeException {
+        if (!new StorageStateProvider().isStorageMounted()) {
```
seadowg: I think it'd be nice to pull this out as a constructor argument. For me that helps to signal that this is a dependency for the object and I think we'll want to move to having these objects constructed in Dagger to allow us to mock/fake them in tests. For the moment you could also just have a no arg constructor as well like you have in `StoragePathProvider`.

grzesiek2010: Done.

lognaturel (approved):

lognaturel:
Another risky and far-reaching one for you, @opendatakit/testers. I think that for now it's sufficient to make sure that no current behavior has regressed. The behavior when scoped storage is used can be verified once the migration code is in place.

Lots of it is very quick to verify -- if forms and instances show up in the interface, they paths are correctly stored. The more subtle things would be verifying whether cache files are properly accessed and whether image and annotation questions are correctly saved.

mmarciniak90:
**Tested with success**

@mmarciniak90 tested on Android: 4.2, 4.4, 7.0, 8.1
@kkrawczyk123 tested on Android 5.1, 6.0, 9.0, 10.0

Verified cases
- fresh install
- downgrade (remove old version without data, install new version)
- upgrade
- reset settings
- delete db from metadata
- downloading from Aggregate
- downloading from GD
- filled 4 forms with external itemset frim Aggregate
- filled 7 forms with external itemset from GD
- Ignore editing 
- all form statuses
- delete some previously sent form
- loading from cache Nigeria Wards form 
- Sample form - Preloading form
- Search and select form
- Static items predicatel
- Encrypted geopoint
- auto-send / auto-delete set in form
- manual copy offline layer
- form with remembring previously entered values 
- encrypted form with remembring previously entered values 
- automated regression tests on Android 6.0

PR #3607, opened 2020-01-24
Move `CollectServerClient` and `DownloadFormListTask` tests to Robolectric

seadowg:
These tests were previously instrumentation tests. And, of course, moving them into Robolectric left some irresistible refactors in front of me:

* Moved focus of task test to `DownloadFormListUtils` which was actually where the work was happening
* Rename `DownloadFormListUtils` to `FormListDownloader` in the spirit of it being a "usecase" or "noun-verber".
* Rename `CollectServerClient` to `OpenRosaAPIClient` as it seems like it's trying to represent the OpenRosa API as opposed to our `OpenRosaHttpInterface` (in the way an SDK would) which is concerned with the lower level HTTP requests. It's not quite there but I've added a note to `STATE.md` about where it should go.
* Reorganize the `http` package as it now feels like (as above) we have a top level package that abstracts our communication with OpenRosa

There is definitely a lot more that I could have done (starting to properly flesh out the `OpenRosaAPIClient` would have been my next step) but I wanted to timebox this so I didn't end up down a rabbit hole.

Ran tests and played with the app against Aggregate.

We'd previously discussed moving non UI tests like this to Robolectric and decided it did make sense as the start up time is far smaller for running individual tests classes.

Hopefully there aren't any changes in behaviour. As with any changes to the HTTP stack we should play around with the app against a variety of servers as it still feels like a blindspot in our automated test suite.


collect_app/src/androidTest/java/org/odk/collect/android/ServerNameIndicationTest.java
```diff
@@ -1,74 +0,0 @@
...
-/**
- * An on-device test for TLS server name indication support.
- *
- * @see <a href="https://github.com/opendatakit/collect/issues/199">COLLECT-199</a>
- */
-@Suppress
-@RunWith(AndroidJUnit4.class)
-public class ServerNameIndicationTest {
-
-    private static final String SNI_URI = "https://sni.velox.ch/";
```
lognaturel, line 34: Good cleanup. This test wouldn't work anyway because this domain no longer exists. As you say, it's not something that really needs to be tested anymore so I don't think any such services are around.

lognaturel (approved):
I was initially confused by 83f35e0. I had to dig through a bit to realize that `OpenRosaServerClientProviderTest` has tests to verify that the expected headers are sent, as do the head/post/get request tests. I also verified that the other removed test cases were either replaced or already covered in lower-level tests.

kkrawczyk123:
@seadowg How we can test that pr? (as the label has been added)

lognaturel:
@seadowg wrote "we should play around with the app against a variety of servers as it still feels like a blindspot in our automated test suite." I think doing downloads and uploads to Aggregate and Central on a single Android version is sufficient. For Aggregate, it would be good to verify with an authenticated user.

kkrawczyk123:
**Tested with success!**

PR #3608, opened 2020-01-24
Update font size immediately to reflect changed settings while filling a form

### LINKED ISSUE #3356: When settings values are changed while filling a form, the new values should take effect immediately

chidauri:
Closes #3356 

Tested on Android 9, on different forms and different questions.

This is a simple way and i didn't find any risks in it.

No

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -314,6 +314,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      * detection of location permissions changes while the activity is in the background.
      */
     private boolean locationPermissionsPreviouslyGranted;
+    private int questionFontSize;
 
     private BackgroundLocationViewModel backgroundLocationViewModel;
     private IdentityPromptViewModel identityPromptViewModel;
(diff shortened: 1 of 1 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2157,6 +2160,11 @@ protected void onResume() {
             return;
         }
 
+        if (questionFontSize != Collect.getQuestionFontsize()) {
```
grzesiek2010: It would be good to factor it out to a separate method, `onResume()` is big enough... our goal should be to have small methods with good names to read the code like a story. I know it might be difficult to find such an approach especially looking at `ForEntryActivity` class but that's a good practice.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: The name could be `adjustFontSize` for example.

chidauri: Done. Thanks :)

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**

kkrawczyk123:

PR #3610, opened 2020-01-27
Update tests for new repeat helpers

seadowg:
This fixes the build. My PR changed test helper methods that were being used in new changes the PR hadn't been updated with. These are test only changes so should be able to be merged straight in after review.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/OpenSourceLicensesPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/OpenSourceLicensesPage.java
@@ -1,0 +1,17 @@
+package org.odk.collect.android.support.pages;
+
+import androidx.test.rule.ActivityTestRule;
+
+class OpenSourceLicensesPage extends Page<OpenSourceLicensesPage> {
+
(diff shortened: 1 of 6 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/OpenSourceLicensesPage.java
```diff
@@ -0,0 +1,17 @@
...
+
+class OpenSourceLicensesPage extends Page<OpenSourceLicensesPage> {
+
+    OpenSourceLicensesPage(ActivityTestRule rule) {
+        super(rule);
+    }
+
+    @Override
+    public OpenSourceLicensesPage assertOnPage() {
+        waitForText("Open Source Licenses");
```
seadowg, line 13: This was failing for me locally. Looks like the page takes a second or two to load but that the standard Espresso wait syncs (on background threads etc) don't work. Probably because of the `WebView` involved I'd imagine.

lognaturel (approved):

lognaturel:
Thanks for getting right on that, @seadowg. I agree that this kind of situation where two PRs affecting the same code come in at roughly the same time is bound to happen sometimes and doesn't require a process change.

PR #3611, opened 2020-01-27
Fixes Issue#3054: Months in Myanmar calendar displayed in English on devices running API level lower than 21

### LINKED ISSUE #3054: Months in Myanmar calendar are not displayed on some devices
Current master branch

#### Problem description
The problem has been reported here https://github.com/opendatakit/collect/pull/2662#issuecomment-491258283

Months in Myanmar calendar are not displayed on some devices. The only device we are able to use for reproduction is Infinix Race Bolt Q X451 device with Android 4.2.

57524338-d1a9dd80-7327-11e9-8b05-fa0a48b396a1

SaumiaSinghal:
Closes #3054 

I tested it on Android devices running API level 16 onwards,  to confirm the API level from which the device start supporting Burmese language. 

Screenshot from 2020-01-27 13-05-48

I found that for lower Android versions, the rendering does not take place for some script languages, even on browsers, so, for those devices (running Kitkat or lower android versions), the best is to display the names of Myanmar months in english language itself.  

No regression risks

A from having Myanmar date widget, eg. All Widgets form.

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
@@ -70,7 +70,9 @@ public class MyanmarDateUtils {
     public static LanguageCatalog getLanguageCatalog() {
         if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             return new LanguageCatalog(Language.MYANMAR);
-        } else return new LanguageCatalog(Language.ENGLISH);
+        } else {
+            return new LanguageCatalog(Language.ENGLISH);
+        }
     }
 
     public static int getFirstMonthDay(MyanmarDate myanmarDate) {
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
@@ -68,7 +68,7 @@ public class MyanmarDateUtils {
     }
 
     public static LanguageCatalog getLanguageCatalog() {
-        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if (android.os.Build.VERSION.SDK_INT >= 4.4) {
             return new LanguageCatalog(Language.MYANMAR);
         } else {
             return new LanguageCatalog(Language.ENGLISH);
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
@@ -16,8 +16,6 @@
 
 package org.odk.collect.android.utilities;
 
-import android.os.Build;
-
 import org.joda.time.LocalDateTime;
 
 import mmcalendar.Language;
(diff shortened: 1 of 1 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
```diff
@@ -59,10 +61,17 @@ public static MyanmarDate createMyanmarDate(int myanmarYear, int myanmarMonthInd
...
 
     public static String[] getMyanmarMonthsArray(int myanmarYear) {
-        return MyanmarDateKernel
-                .getMyanmarMonth(myanmarYear, 1)
-                .getMonthNameList(new LanguageCatalog(Language.MYANMAR))
-                .toArray(new String[0]);
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            return MyanmarDateKernel
+                    .getMyanmarMonth(myanmarYear, 1)
+                    .getMonthNameList(new LanguageCatalog(Language.MYANMAR))
```
grzesiek2010: Since only this part depends on the os version (`Language.MYANMAR` or `Language.ENGLISH`) I would factor out a method that would return `Language` based on os version and call it here.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/utilities/MyanmarDateUtils.java
SaumiaSinghal: Hie @grzesiek2010, I got your point, writing the same return statement twice just for a change in a single variable seems a bit cumbersome to me as well. How about, if I create a separate bool variable `higherVersion` to check the version, and then use the ternary operator to get the correct language on the system.

grzesiek2010: Ternary operator would bake that line long I would prefer having a separate method it could return Language or even LanguageCatalog and then the line would be:
getMonthNameList(getLanguageCatalog())

grzesiek2010 (approved):

mmarciniak90:
@SaumiaSinghal I see that English version is visible on Android v4.2 and v4.1 :+1: 
But I also noticed regression on Android 4.4. 
The problem does not applicable Android 4.4 but the fix caused that the English version is visible instead of original

| v1.25.1| pr/3611|
|---|---|
|Screenshot_2020-01-28-16-21-14|Screenshot_2020-01-28-16-18-23|

SaumiaSinghal:
Hie @mmarciniak90! Thanks for reporting this issue. I'll see if there's some way I can identify whether native characters of a language are supported in Android version or not, and, if they are not, then we can switch to the English version, I think that will be a better solution.

mmarciniak90:
@SaumiaSinghal unfortunately, last changes caused a regression on Android 4.1 and 4.2. 
Months are not displayed in the Myanmar calendar.



SaumiaSinghal:
Ohh I'm really sorry @mmarciniak90, and thanks for reporting it. I'm away right now but I will fix it by the end of the day.

lognaturel:
I have good and bad news. The good news is that the [Technical Steering Committee](https://opendatakit.org/community/governance/#tsc-1) discussed Collect Android version support today and agreed to drop support for Android <5. This is great news because Android 4.x was really different and supporting it has resulted in lots of special cases like this one. It's bad news because it means that @SaumiaSinghal's excellent work here is no longer needed.

Sorry about that, @SaumiaSinghal! I thought we wouldn't be dropping support until the following release but the more we discussed it, the clearer it became that there would be little benefit to continuing 4.x support in this release.

If you would like, you could be involved in helping raise the `minSdkVersion` to 21 and removing special cases that are no longer needed. https://github.com/opendatakit/collect/issues/3624

SaumiaSinghal:
Oh that's great news @lognaturel, yeah I would love to work on it. Thanks!

PR #3613, opened 2020-01-29
Disable custom swipe animation when animations are off

seadowg:
This makes sure that the custom swipe animation we use in `FormEntryActivity` is disabled when the emulator/device has animations off. This should give some speed up for UI tests that deal with form entry.

Ran tests locally and on test lab.

We could have used the custom disabling we already had but I think it's better to be able to use Android's own animation scale to determine when we have animations turned off or not. Means we have less code to maintain on our side!

Only really changes to tests so should be good to merge aftert approval.


lognaturel (approved):
NO! I did a literal 🤦🏻‍♀️. I remember a long time ago going "huh, there are a lot of animations when animations are off" and then not worrying about it. Yet another reminder of "see something, say something." Thanks so much for catching this, @seadowg.

PR #3616, opened 2020-01-30
Fix how change tracking works with save button

### LINKED ISSUE #3498: Form with `track-changes` missing changes for question if edited after save
Collect v1.24.x

#### Problem description

If a form is saved while viewing a question (using the 💾 icon) and then a change is made the `question` event will not have its `old-value` and `new-value` populated.

1. Fill in the [attached form](https://github.com/opendatakit/collect/files/3927316/track-changes.zip).
2. Open the form to edit
3. Go to the question
4. Click 💾
5. Make a change
6. Press back
7. Press "Save changes"

The `question` event will be in the logs but will not have the `old-value` and `new-value` populated.

#### Expected behavior

There are a few things we could do. The two that come to the top of my head are:

1. `question` event is somehow updated with the `old-value`/`new- value` on the second save
2. A second `question` event appears after `form-save` with the `old-value`/`new- value`

Personally the second feels more correct to me.

seadowg:
Closes #3498. We decided in the discussion in the issue that the solve would be for a second question event to appear after form-save with the old-value/new- value.

Used a form with change tracking enabled to check the issue was solved. I also added tests to drive out the new (fixed) behaviour.

We discussed in #3498. In terms of the structure I think it makes sense to move audit logging into our `ViewModel`s as much as possible. This lets us abstract the details of all the interactions we need to make every time there is some 'action' performed in the UI. It does feel like we might end up with a lot of logic in the `ViewModel`s however so it probably would be good to start abstracting this to a more high level component responsible for handling logging logic (`EnumeratorBehaviourLogger` for example) before we go any further. I don't think we need an issue for this but it might be a good idea for the future.

It'd be good to try a few forms with change tracking and alos try repeats, groups and single questions as they all output "question" audit events. The new changes are tested so it feels less risky than a lot of changes to form entry.

#### Do we need any specific form for testing your changes? If so, please attach one

Any form with change tracking would be good to use but there aren't any other specific features we need to look at.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/DiskFormSaver.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/DiskFormSaver.java
@@ -2,14 +2,15 @@ package org.odk.collect.android.formentry;
 
 import android.net.Uri;
 
+import org.odk.collect.android.logic.FormController;
 import org.odk.collect.android.tasks.SaveFormToDisk;
 import org.odk.collect.android.tasks.SaveToDiskResult;
 
 public class DiskFormSaver implements FormSaver {
 
     @Override
-    public SaveToDiskResult save(Uri instanceContentURI, boolean shouldFinalize, String updatedSaveName, boolean exitAfter, ProgressListener progressListener) {
-        SaveFormToDisk saveFormToDisk = new SaveFormToDisk(instanceContentURI, exitAfter, shouldFinalize, updatedSaveName);
+    public SaveToDiskResult save(FormController formController, boolean shouldFinalize, String updatedSaveName, boolean exitAfter, ProgressListener progressListener, Uri instanceContentURI) {
+        SaveFormToDisk saveFormToDisk = new SaveFormToDisk(formController, exitAfter, shouldFinalize, updatedSaveName, instanceContentURI);
         return saveFormToDisk.saveForm(progressListener);
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
@@ -97,6 +97,16 @@ public class AuditEventLogger {
         }
     }
 
+    /*
+     * Finalizes and writes events
+     */
+    public void flush() {
+        if (isAuditEnabled()) {
+            finalizeEvents();
+            writeEvents();
+        }
+    }
+
     private void addLocationCoordinatesToAuditEvent(AuditEvent auditEvent, long currentTime) {
         Location location = getMostAccurateLocation(currentTime);
         String latitude = location != null ? Double.toString(location.getLatitude()) : "";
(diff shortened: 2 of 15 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -97,6 +97,16 @@ public void logEvent(AuditEvent.AuditEventType eventType, FormIndex formIndex,
         }
     }
 
+    /*
+     * Finalizes and writes events
+     */
+    public void flush() {
```
lognaturel, line 103: Ah yes, good name!

collect_app/src/main/java/org/odk/collect/android/formentry/DiskFormSaver.java
```diff
@@ -2,14 +2,15 @@
...
+import org.odk.collect.android.logic.FormController;
 import org.odk.collect.android.tasks.SaveFormToDisk;
 import org.odk.collect.android.tasks.SaveToDiskResult;
 
 public class DiskFormSaver implements FormSaver {
 
     @Override
-    public SaveToDiskResult save(Uri instanceContentURI, boolean shouldFinalize, String updatedSaveName, boolean exitAfter, ProgressListener progressListener) {
-        SaveFormToDisk saveFormToDisk = new SaveFormToDisk(instanceContentURI, exitAfter, shouldFinalize, updatedSaveName);
+    public SaveToDiskResult save(FormController formController, boolean shouldFinalize, String updatedSaveName, boolean exitAfter, ProgressListener progressListener, Uri instanceContentURI) {
```
lognaturel: This is a small thing but I'd keep the `Uri` first since that's what you're saving.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.javarosa.form.api.FormEntryController.EVENT_QUESTION;
+import static org.javarosa.form.api.FormEntryController.EVENT_REPEAT;
+
+public class AuditUtils {
+
+    private AuditUtils() {
+
+    }
+
+    public static void logCurrentEvent(FormController formController, AuditEventLogger auditEventLogger, long currentTime) {
```
lognaturel: This is more like `logCurrentScreen` or `logCurrentEvents`, I think?

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
seadowg: Heh finally!

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditUtils.java
seadowg: In my head it's logging the "current event" of the `FormController` but I see  now that is confusing. I'll make that explicit.

mmarciniak90:
**Tested with success**

Verified on Android: 4.4, 5.1, 6.0, 7.0, 8.1, 9.0, 10.0

Analyzed all audit files.

Verified cases:
* simple form
* form with groups
* form with nested repeats
* form with a field-list group
* edited previously saved form
* adding a new answer
* returning to the previous answer
* changing device orientation


|old version | version with fix|
|---|---|
|old|pr|

mmarciniak90:

PR #3617, opened 2020-01-30
Add scoped storage strings for translation

lognaturel:
This PR adds the banner text for letting users know about the scoped storage migration. This has been discussed at https://forum.opendatakit.org/t/let-users-know-of-required-storage-location-change-and-let-them-opt-in-before-august-2020/24444/. The context for it is that we are going to be letting translators know about the change to repeat strings and we'd like to alert them to these at the same time even though the feature is not complete. Both of these are important because it would be particularly confusing for enumerators to come across text in a language they don't understand (English).

One deviation from the discussion there is to make the button text "Learn more and migrate." This is because we have decided not to make the banner dismissible as a first pass. I think the longer text is helpful but let me know if it feels too long. I can also take it back to the forum thread if needed.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -278,6 +278,10 @@ public class Collect extends Application {
         setupRemoteAnalytics();
         setupLeakCanary();
         setupOSMDroid();
+
+        // Force inclusion of scoped storage strings so they can be translated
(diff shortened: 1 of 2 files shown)


yanokwa (approved):
Clever use of Timber.

lognaturel:
I'm going to merge this so Transifex can pull them next time it refreshes. I'll wait for a thumbs up from @seadowg before sending a message to translators.

PR #3623, opened 2020-02-05
Changes for allowing files migration

grzesiek2010:
Migration steps are:
1. Clear app dir on scoped storage (just in case)
2. Check if form uploader is running in the background if it is stop
3. Check if form downloader is running in the background if it is stop
4. Check if there is enough space to copy files, if not stop
5. Try to copy files, if any exception happens stop
6. Migrate database paths, if any exception happens stop.
7. Update offline map layer path (if selected in map preferences)
8. Reset ReferenceManager to clear old paths
9. Remove itemsets.db
10. Clear cache
11. Remove odk dir from sdcard
12. Use scoped storage

I tested implemented changes manually and added some automated tests.

It's what we need because of Google policy and what we agreed to do.

The code responsible for migration is pretty isolated that means fortunately I wouldn't expect that it might affect anything not related to this option but in the migration itself you can expect unexpected of course.
In order to use the AdminPassworDialog in migration I needed to factor it out (before it was used just in the Main Menu), so please make sure there is no regression there.

No.

Probably we need to document the new scoped storage but it's not just related with this pr.


seadowg:
@grzesiek2010 hmmm. Yeah for this kind of work I think you're right that `WorkManager` isn't the right thing. What I'm worried about with `AsyncTask` is:

* The docs state "AsyncTasks should ideally be used for short operations (a few seconds at the most.)". I'm worried this might mean we run into problems if the migration takes longer than that which in my mind it could. That said, I can never find anything around any hard limits on run time for `AsyncTask` and I've never actually run into a "timeout" with them
* We're not protected from process death. If the user backgrounds the app and it is then killed the migration will be killed in the  middle (as far as I understand)
* `AsyncTask` is being deprecated. Maybe not a big problem as we won't need this migrator for too long hopefully.

It feels to me like the right course of action might be a "Foreground Service" which would mean we display a notifcation while the migration is happening. This would signal to Android that the app shouldn't be killed as we're ["doing something important"](https://developer.android.com/guide/background/#fg-svc). 

It might be nice to just use the `AsyncTask` for now and play around with it on a bunch of devices in low battery, lots of apps running etc to check how it behaves. I think it'd be a good idea to make sure all the migration logic lives in a seperate object so it's testing and behaviour isn't tied to `AsyncTask`. That way it won't be a large effort to change "how" it's executed.

grzesiek2010:
> It feels to me like the right course of action might be a "Foreground Service" which would mean we display a notifcation while the migration is happening.

but we didn't want to display a notification and we wanted to block the UI.

> It might be nice to just use the AsyncTask for now and play around with it on a bunch of devices in low battery, lots of apps running etc to check how it behaves.

I agree. I think it's time to involve the QA team to check for example how much time such a migration might take.

@seadowg any comments related to the rest of implementation? maybe you can think of something I missed performing the migration?

seadowg:
@grzesiek2010 sorry I didn't take a deeper look. Thought you were just flagging the `AsyncTask` up as the PR's still a draft. I'll have a look now!

collect_app/src/main/java/org/odk/collect/android/storage/StorageEraser.java
```diff
@@ -0,0 +1,24 @@
+package org.odk.collect.android.storage;
+
+import java.io.File;
+
+class StorageEraser {
```
seadowg: I like that this is pulled out as its own object! I think you could pass the `StoragePathProvider` in as a constructor arg rather than have it passed in to both methods as it feels like a dependency.

collect_app/src/main/java/org/odk/collect/android/storage/StorageMigrator.java
```diff
@@ -0,0 +1,201 @@
...
+        this(new StoragePathProvider(), new StorageStateProvider(), new StorageEraser());
+    }
+
+    private StorageMigrator(StoragePathProvider storagePathProvider, StorageStateProvider storageStateProvider, StorageEraser storageEraser) {
+        this.storagePathProvider = storagePathProvider;
+        this.storageStateProvider = storageStateProvider;
+        this.storageEraser = storageEraser;
+    }
+
+    StorageMigrationResult performStorageMigration() {
```
seadowg: Big fan of this returning an enum/result object for each case 👍

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -640,4 +643,14 @@ private void disableSmsIfNeeded() {
...
     }
+
+    private void setStorageMigrationBannerVisibility() {
+        if (!new StorageStateProvider().isScopedStorageUsed()) {
+            findViewById(R.id.storage_migration_banner).setVisibility(View.VISIBLE);
+        }
+    }
+
+    public void learnMoreAndMigrate(View view) {
+        new StorageMigratorTask(this).execute();
```
seadowg: I think it would be nice to have the execution encapsulated in a `ViewModel`. An example of this would be how `FormSaveViewModel#saveForm` works. Something like this for the `ViewModel`:

```java
class StorageMigrationViewModel {
    public LiveData<StorageMigratorResult> migrate() {
        MutableLiveData<StorageMigratorResult> result = new MutableLiveData<>();
        new StorageMigratorTask((migratorResult) -> result.setValue(migratorResult)).execute();
        return result;
    }
}
```

You'll see an example of a task with a listener like this in `FormSaveViewModel`.

In can then deal with the dialog/errors in the Activity:

```java
storageMigrationViewModel.migrate().observe(this, (result) -> {
    if (result != null) {
        // handle result
    }
});
```

Something like that lets us seperate out our UI logic from the migration logic and lets us avoid using `WeakReference` or worrying about `AsyncTask` memory leaks. Totally up for discussing if that approach doesn't feel right but it'd be nice for us to find a pattern we like for this kind of thing.

collect_app/src/main/java/org/odk/collect/android/storage/StorageMigrator.java
```diff
@@ -0,0 +1,201 @@
...
+import timber.log.Timber;
+
+import static android.provider.BaseColumns._ID;
+import static org.odk.collect.android.database.ItemsetDbAdapter.KEY_PATH;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.FORM_FILE_PATH;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.FORM_MEDIA_PATH;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.JRCACHE_FILE_PATH;
+import static org.odk.collect.android.provider.InstanceProviderAPI.InstanceColumns.INSTANCE_FILE_PATH;
+
+public class StorageMigrator {
```
seadowg: What are you thinking for tests for this? Feels to me like testing the `StorageMigrator` in a all the possible scenarios (each result) and then maybe an Espresso test for the "happy case" (where everything goes right)?

collect_app/src/main/res/layout/storage_migration_banner.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
seadowg, line 2: It would be cool to pull this out as a `MaterialBanner` component but I think we can do that another time. I can look at doing thta in a similar way as we did for the `MaterialFullScreenDialog`.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: I think it's a good idea I implemented it, thanks.

mmarciniak90:
We checked the time needed to migrate files

Steps:
1. Install app
2. Add files
3. Migrate

Test Data: [3623_test_data.zip](https://drive.google.com/file/d/1G7xE8gPrn0xlNjXJOkfkSlIKkSEJfPvb/view?usp=sharing)


|Files/Android|10 forms, ~1k filled submissions, without layers|10 forms, ~1k filled submissions, 2 layers with ~9MB per one|10 forms, 40 filled submissions, without layers| 10 forms, 40 filled submissions,   2 layers with ~9MB per one|
|---|---|---|---|---|
|Android 5.1|54s|56s|5s|6s|
|Android 6.0|1m 18s|1m 19s|7s|7s|
|Android 7.0|57s|59s|5s|4s|
|Android 8.1|52s|50s|4s|4s|
|Andorid 9.0|13s|14s|~1s|~2s|
|Android 10.0|15s|15,s|~2s|~2s|

grzesiek2010:
That helps a lot!
According to the doc `AsyncTask` should be used for operations that take a few seconds at the most. Maybe that would be ok in most cases (10 forms, 40 submissions) but we can't rule out such cases where users have tons of forms right?
@lognaturel @seadowg  what do you think?

kkrawczyk123:
@grzesiek2010 when testing on Android 5.1 on Huawei Y560-L01 the migration of 10 forms and 1000 submissions (both cases with layers and without) the app crashed right after migrating all the files.

collect_app/src/main/java/org/odk/collect/android/storage/StorageMigrator.java
grzesiek2010: Ok I added such a test.
When it comes to Espresso we need to have the UI implemented first.

collect_app/src/main/java/org/odk/collect/android/storage/StorageEraser.java
grzesiek2010: Done.

grzesiek2010:
> @grzesiek2010 when testing on Android 5.1 on Huawei Y560-L01 the migration of 10 forms and 1000 submissions (both cases with layers and without) the app crashed right after migrating all the files.

Ok thanks, I'll look at it once we decide which component should be used to perform migration.

grzesiek2010:
Ok so to sum up the current state:
1. @seadowg I solved all things you pointed out. If you have anything else let me know.
2. The UI is not finished I implemented the banner but it starts migration immediately @lognaturel let me know if your proposal is accepted by TSC and I can go ahead.
3. The QA team performed tests to figure out the amount of time needed to perform migration, results probably indicate it's to much for `AsyncTask` @lognaturel @seadowg  if you agree I think `Service` is what we should use.

lognaturel:
Thanks so much for getting ahead of the on-device verification, @opendatakit/testers and @grzesiek2010. It's great to have some concrete info to base the component selection on. I agree with the conclusion that `AsyncTask` won't work here.

To summarize the properties of what we're dealing with:
- the user triggers the migration and it must start right away (non-deferrable)
- the user must not be able to use Collect while the migration is ongoing
- the user may navigate away from Collect while the migration is ongoing
- the migration may take several seconds
- the user must get some kind of feedback about a successful migration

@grzesiek2010 you're correct that `WorkManager` definitely won't work because of the its deferrable property. The [Android background work flowchart](https://developer.android.com/guide/background) points pretty clearly to a foreground service.

> Foreground services must display a Notification.

This seems fine and even good to me. That way if it does take a moment and the user navigates away, they can see the progress at a glance. I've found the Android docs for `Service`s extremely confusing. 

I found [this article](https://proandroiddev.com/deep-dive-into-android-services-4830b8c9a09) pretty helpful. From that, I believe we want a started service that periodically broadcasts status information. I don't believe there's any reason for the service to be bound because the activity that launches it will not need to communicate with the service after creation.

Keeping all of the actual migration work in its own class and then writing a really thin `Service` class seems like the way to go.

> let me know if your proposal is accepted by TSC and I can go ahead.

After the last TSC meeting, we have the go-ahead to do a full-screen dialog with some explanatory text, a button to migrate, and another to cancel. There may be some slight changes in the text wording but that should just be changing strings. You can go ahead with creating the rest of the user interface. Consider just putting some placeholder text on the full screen dialog. Thanks!

grzesiek2010:
The current migration looks like:
ezgif com-video-to-gif (2)

what do you think?

grzesiek2010:
Ok so I replaced an `AsyncTask` with a `Service` but I didn't use notifications. I find it redundant since we need to block the UI either way and display a progress bar. Of course if a user closes the app while migration is being performed and open it again then there will be no message (that the migration is done or something went wrong) so in this case a notification would make some sense but still I think that it would be better to update the banner. For example:

If everything went well we could change the message to let a user know and make the banner dissemble otherwise we can add a not that a user needs to try again because migration failed.
What do you think?

lognaturel:
It's looking really good. I see your point about notifications and your suggestion seems reasonable. What happens when the user navigates away from the app? Is the service killed immediately or just when the `Activity` is killed?

I'm afraid I got side tracked by the regression at https://github.com/opendatakit/collect/issues/3635 today and didn't have a chance for a deeper review yet. Is this ready to go out of draft state? Would be great @seadowg if you have a chance to deep dive during your day. Consider an interactive review if you two are up for it.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
```diff
@@ -0,0 +1,51 @@
...
+import android.content.Intent;
+
+import androidx.lifecycle.LifecycleService;
+import androidx.localbroadcastmanager.content.LocalBroadcastManager;
+
+import org.jetbrains.annotations.NotNull;
+import org.odk.collect.android.storage.StoragePathProvider;
+import org.odk.collect.android.storage.StorageStateProvider;
+
+public class StorageMigrationService extends LifecycleService {
```
seadowg: I think we should be using an `IntentService` here (https://developer.android.com/guide/components/services#ExtendingIntentService). I've usually used them in the past as they take care of creating a worker threads and also protect us from multiple requests being exectued simultaneously etc. This would potentially mean the service could take care of the lock in `StorageMigrator` (`.isMigrationBeingPerformed`). Also if we do use a plain ol' service I think it needs to call `stopSelf()` when we're done executing - again an `IntentService` just takes care of that for us. 

Sorry if this is research you already did and made the decision for some reason I'm missing!

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationDialog.java
```diff
@@ -0,0 +1,69 @@
...
+import android.widget.TextView;
+
+import androidx.annotation.Nullable;
+import androidx.appcompat.widget.Toolbar;
+import androidx.fragment.app.DialogFragment;
+
+import org.jetbrains.annotations.NotNull;
+import org.odk.collect.android.R;
+
+public class StorageMigrationDialog extends DialogFragment {
```
seadowg: You should be able to use the `MaterialFullScreenDialogFragment` for this I think

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
```diff
@@ -0,0 +1,163 @@
...
+    private StorageEraser storageEraser;
+    private final StoragePathProvider storagePathProvider = spy(StoragePathProvider.class);
+    private final StorageStateProvider storageStateProvider = mock(StorageStateProvider.class);
+
+    @Before
+    public void setup() {
+        storageEraser = spy(new StorageEraser(storagePathProvider));
+        storageMigrator = spy(new StorageMigrator(storagePathProvider, storageStateProvider, storageEraser));
+
+        doNothing().when(storageMigrator).reopenDatabases();
```
seadowg, line 35: I think we need to break this object up a bit so we're not having to fake/mock methods on the object under test. It looks like you've had to do this because of a bunch of static methods though. Maybe using a similar strategy to the `BackgroundLocationHelper` that @lognaturel created? This worked at the time as it allowed her to wrap a bunch of static methods/singletons that the new object she was build needed without having to go and refactor/unwrap all of them. This interface could also include `isFormUploaderRunning()` and `isFormDownloaderRunning()`.

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
```diff
@@ -0,0 +1,163 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class StorageMigratorTest {
+
+    private StorageMigrator storageMigrator;
+    private StorageEraser storageEraser;
+    private final StoragePathProvider storagePathProvider = spy(StoragePathProvider.class);
```
seadowg, line 24: Can this just be a mock rather than a spy? Personally I find spies make tests harder to read and understand as the boundaries of the test (i.e. what is faked/stubbed) become unclear.

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
```diff
@@ -0,0 +1,163 @@
...
+    private final StorageStateProvider storageStateProvider = mock(StorageStateProvider.class);
+
+    @Before
+    public void setup() {
+        storageEraser = spy(new StorageEraser(storagePathProvider));
```
seadowg: Again would be nice for this to be a mock/fake rather than a spy.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationDialog.java
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
grzesiek2010: I was thinking about using broadcast receivers vs livedata... could you take a look at the last commit whwre I replaced broadcast receivers with livedata and share you thought?

grzesiek2010:
> What happens when the user navigates away from the app? Is the service killed immediately or just when the Activity is killed?

If I close the app it's not killed the same when the activity is killed (Don't keep activities option) but it will be killed if you navigate to settings and force stop the app.

collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
```diff
@@ -39,8 +39,10 @@ public void onStart() {
             int height = ViewGroup.LayoutParams.MATCH_PARENT;
             dialog.getWindow().setLayout(width, height);
 
-            // Make sure soft keyboard shows for focused field - annoyingly needed
-            dialog.getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_VISIBLE);
+            if (shouldShowSoftKeyboard()) {
```
seadowg, line 41: Oh wow was this making it pop open the keyboard even though there aren't any fields? That's annoying. Feel free to reverse it so the default is `false` and the other dialogs override.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
seadowg: Using `LiveData` does seem a lot cleaner in the observers (the `Activity` and the `DialogFragment`). I'd want to avoid having the `LiveData` object being static and also avoid having the `StorageMigrator` have to deal with `LiveData` though. This is to avoid us running into problems with tests polluting each other  and also just avoids complicating the `StorageMigrator` with reactive stuff.

What I might do is look at having a seperate `MigrationStatusRepository` like this:

```java
class MigrationStatusRepository {

    private MutableLiveData<StorageMigrationResult> migrationStatus = new MutableLiveData<>();

    public LiveData<StorageMigrationResult> getStatus() {
        return migrationStatus;
    }

    public void setStatus(StorageMigrationResult storageMigrationResult) {
        migrationStatus.setValue(storageMigrationResult);
    }
}
```

The Activity/Fragment would call `getStatus()` to observe and the Service could write the status when `StorageMigrator` completes using `setStatus`. This can be set up as a singleton (Dagger's `@Singleton`) and injected into the Activity, Fragment and Service. That way it can be easily swapped out in tests for any of those objects.

How does that feel? cc/ @lognaturel

grzesiek2010: But the service doesn't know about every single step it just starts migration (one method) sot it might receive the result but not every single status which changes after every step. Looking at your code where you mixed words `status` and `result` I think you didn't notice that we need both: one general result and a few statuses to update current progress text.

grzesiek2010: So I would need to 
- inject such repository to StorageMigrator class
- or divide the migration into a few methods and call all of them in Service one by one (and then deal with statuses in Service)

seadowg: Ah you're absolutely right I missed that somehow. Sorry about that.

Yeah I'd still think about about looking at something lilke the `MigrationStatusRespository`. It could injected into the `StorageMigrator` (probably just as a constructor argument). It could still just have one set/get pair as the "results" could also just be a statuses i.e. the status could change from `MIGRATING_DATABASES` to `SUCCESS` or  from `PREPARING_SCOPED_STORAGE` to `NOT_ENOUGH_SPACE`.

collect_app/src/main/java/org/odk/collect/android/material/MaterialFullScreenDialogFragment.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
grzesiek2010: Ok I pushed new changes and now I use `LiveData` and a repository class for both results and statuses. I think those statuses is more like my idea so here is a question to @lognaturel because maybe you just want to display one message like `Please wait. it might take a few seconds...` and then we can get rid of it at all otherwise I think we need both.

collect_app/src/main/java/org/odk/collect/android/storage/StorageStateProvider.java
```diff
@@ -1,21 +1,78 @@
...
 
+    public void disableUsingScopedStorage() {
+        GeneralSharedPreferences.getInstance().save(KEY_SCOPED_STORAGE_USED, false);
+    }
+
     boolean isStorageMounted() {
         return Environment.getExternalStorageState().equals(Environment.MEDIA_MOUNTED);
     }
+
+    public boolean isEnoughSpaceToPerformMigartion(StoragePathProvider storagePathProvider) {
```
lognaturel: Migartion -> Migration

collect_app/src/test/java/org/odk/collect/android/storage/migration/StorageMigratorTest.java
lognaturel: Agreed that sounds like a promising direction.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
lognaturel: My initial reaction was that I really like the statuses because it gives users a sense of progress. That said, it's not clear that the information is actionable, the migration is going to be really quick for most people and it's a lot of translation effort for a one-time migration. Let's go with "Migration in progress.\nThis may take a few seconds..." I do like the approach you'd come up with for the statuses, though @grzesiek2010. Sorry to see it go.

lognaturel:
I've only done a first pass through but things are looking good. I agree with @seadowg that we should only use spies if we have existing code that is risky to break apart.

I tried it out and things work well but I think something may be wrong with the instance geometry database column migration. I had several instances of All Widgets that were showing up on the form map, then I migrated, and then none of those instances were on the map. If I open an existing instance and save it again, it then shows up on the map.

collect_app/src/main/java/org/odk/collect/android/storage/StorageStateProvider.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationService.java
grzesiek2010: Ok, done.

grzesiek2010:
> I think something may be wrong with the instance geometry database column migration

I'll take a look at it.

@lognaturel you said you like my proposal about updating the banner instead of displaying a notification/toast or whatever so here are some screenshots so that you can decide if it's what we want:

| Screenshot_2020-02-14-15-05-06  | Screenshot_2020-02-14-15-05-11 | Screenshot_2020-02-14-15-05-14 | Screenshot_2020-02-14-15-05-22 | Screenshot_2020-02-14-15-08-39 |
| ------------- | ------------- | ------------- | ------------- | ------------- |

One more question... what is the final decision about https://docs.google.com/document/d/1RALXguNg6VHLAfdQl_CWDxx-VkL_y5v0_ljVbvHSVaw/edit?disco=AAAAI5EsOlk?

grzesiek2010:
> I think something may be wrong with the instance geometry database column migration

Fixed. Actually it was an existing bug in `InstancesDao.getValuesFromInstanceObject()` method.

lognaturel:
Updating the banner looks good to me. I don't think the red color is necessary or advised by Material Design. How about you wire it with regular text for now and we can see whether any kind of highlight is needed. Perhaps obvious but to make it explicit, I don't think these status messages should survive the activity being killed.

The decision is not to require making submissions. I'll work on getting you all the strings in the next couple days.

collect_app/src/main/java/org/odk/collect/android/dao/InstancesDao.java
```diff
@@ -363,9 +363,8 @@ public ContentValues getValuesFromInstanceObject(Instance instance) {
         values.put(InstanceColumns.STATUS, instance.getStatus());
         values.put(InstanceColumns.LAST_STATUS_CHANGE_DATE, instance.getLastStatusChangeDate());
         values.put(InstanceColumns.DELETED_DATE, instance.getDeletedDate());
+        values.put(InstanceColumns.GEOMETRY, instance.getGeometry());
```
lognaturel, line 366: Interesting! Does this mean that the `getValuesFromInstanceObject` method wasn't used in that context before?

collect_app/src/androidTest/java/org/odk/collect/android/dao/InstancesDaoTest.java
```diff
@@ -195,6 +196,35 @@ public void updateInstanceTest() {
         assertThat(formWithGeopointInstance.getGeometry(), is(nullValue()));
     }
 
+    @Test
+    public void getValuesFromInstanceObjectTest() {
```
lognaturel: Is your thinking that this test prevents someone accidentally removing one of the db fields from the method? Is this likely? It seems like it doesn't help prevent the kind of problem we saw here -- a new db field not being represented in the method.

Does it have to be a connected test? Seems it could be Robolectric-based?

collect_app/src/main/java/org/odk/collect/android/dao/InstancesDao.java
grzesiek2010: Yes it was used in some tests but didn't verify those new columns.

grzesiek2010:
> Updating the banner looks good to me. I don't think the red color is necessary or advised by Material Design. How about you wire it with regular text for now and we can see whether any kind of highlight is needed.

Done.

> Perhaps obvious but to make it explicit, I don't think these status messages should survive the activity being killed.

Could you elaborate. The current implementation saves the last revived result in shared preferences so that the banner will always contain a message about the last failed attempt. Is that bad?

lognaturel:
> The current implementation saves the last revived result in shared preferences so that the banner will always contain a message about the last failed attempt.

What I meant is that I don't think they should be saved in shared preferences! The reason for that is that most reasons for failure will be short-lived. For example, a download task may have been running on first attempt but that won't be relevant if the user opens the app again two days later. Those status messages feel transient to me.

I just noticed that files are now stored in `/sdcard/Android/data/org.odk.collect.android/files/odk`. I would expect simply `/sdcard/Android/data/org.odk.collect.android/files`. Is there a reason to maintain an additional enclosing folder?

One requirement that I see missing from [the spec document](https://docs.google.com/document/d/1RALXguNg6VHLAfdQl_CWDxx-VkL_y5v0_ljVbvHSVaw/edit#) is that if there is an administrator password specified, the user should be prompted for it when attempting  the migration.

After performing the migration, I was on auto-pilot and pushed a form definition to `/sdcard/odk/forms`. This created my form as a file named `forms`. This is a pretty confusing state so I do think we should create a `/sdcard/odk` file to make sure that a user trying something like that gets an error they can follow up on. I've added that to [the spec document](https://docs.google.com/document/d/1RALXguNg6VHLAfdQl_CWDxx-VkL_y5v0_ljVbvHSVaw/edit#).

When trying this out on the emulator, I wanted to take a screenshot of the migration ongoing screen. I did Apple+S and tapped on the emulator screen right as the migration was starting and the migration kept going forever. I had to force close the application and when I reopened the banner said that the migration failed with no more detail. Not sure how to reproduce but something to look out for @opendatakit/testers.

I've updated strings in [the spec document](https://docs.google.com/document/d/1RALXguNg6VHLAfdQl_CWDxx-VkL_y5v0_ljVbvHSVaw/edit#) and will run them by folks in the forum thread one more time. Please also comment there if you see any issues.

One last thing that I feel is missing is a way for users to tell what their storage directory is. I don't see a clear place to put this but will think about it some more.

collect_app/src/main/java/org/odk/collect/android/preferences/GeneralKeys.java
```diff
@@ -146,7 +148,8 @@
             KEY_METADATA_MIGRATED,
             KEY_AUTOSEND_WIFI,
             KEY_AUTOSEND_NETWORK,
-            KEY_SCOPED_STORAGE_USED
+            KEY_SCOPED_STORAGE_USED,
+            KEY_SCOPED_STORAGE_BANNER_DISMISSED
```
lognaturel: Do you really need this? Like the failure message, the success message feels transient. So I think it's good for it to keep existing while someone is in a single "app session" but I don't think it still needs to be shown if the user closes and then reopens the app. That is, Collect going out of memory would implicitly dismiss the dialog.

collect_app/src/main/java/org/odk/collect/android/storage/StorageStateProvider.java
```diff
@@ -1,21 +1,78 @@
...
+    private long getAvailableScopedStorageSize(StoragePathProvider storagePathProvider) {
+        String scopedStoragePath = storagePathProvider.getScopedExternalFilesDirPath();
+        if (scopedStoragePath.isEmpty()) {
+            return 0;
+        }
+
+        StatFs stat = new StatFs(scopedStoragePath);
+        long blockSize;
+        long availableBlocks;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
```
lognaturel: I think you can pre-emptively go ahead and remove this one since we're sure we're changing the minSdk in this release.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationDialog.java
```diff
@@ -0,0 +1,128 @@
...
+    private void setUpMessageAboutUnsetSubmissions() {
+        if (unsentInstancesNumber > 0) {
+            messageText2.setVisibility(View.VISIBLE);
+            messageText2.setText(getString(R.string.storage_migration_dialog_message2, unsentInstancesNumber));
+        }
+    }
+
+    private void showMoreDetails() {
+        Intent intent = new Intent(getContext(), WebViewActivity.class);
+        intent.putExtra("url", "https://forum.opendatakit.org/t/24159");
```
lognaturel, line 141: I think we'll introduce a more user-facing post but this is a great placeholder for now.

collect_app/src/main/res/values/strings.xml
```diff
@@ -725,4 +725,21 @@
...
     <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage by August 2020 to comply with Android requirements.</string>
     <string name="scoped_storage_learn_more">Learn more and migrate</string>
+    <string name="storage_migration_dialog_message1">Android requirements have changed and files used by Collect must be migrated by August 2020.</string>
+    <string name="storage_migration_dialog_message2">You have %d unsent submissions. You are encouraged to submit all data before migrating.</string>
+    <string name="storage_migration_dialog_message3">If you use an external application integration such as OpenMapKit, please do not migrate yet.</string>
+    <string name="storage_migration_more_details">More details</string>
+
+
+
+    <string name="storage_migration_not_enough_space">You have not enough available space on your internal storage to finish data migration. Please free up some space and try again later.</string>
```
lognaturel: "You have not enough available space" -> "You do not have enough space"

"to finish data migration" -> "to migrate your data"

"try again later" -> "try again"

grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/StorageStateProvider.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/GeneralKeys.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+    void markMigrationStart() {
+        isMigrationBeingPerformed = true;
+    }
+
+    void markMigrationEnd() {
+        isMigrationBeingPerformed = false;
+    }
+
+    public void consumeResult() {
```
seadowg: I'm missing why we need this. Is there any problem in just keeping the result?

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -79,9 +84,9 @@
  * @author Carl Hartung (carlhartung@gmail.com)
  * @author Yaw Anokwa (yanokwa@gmail.com)
  */
-public class MainMenuActivity extends CollectAbstractActivity {
+public class MainMenuActivity extends CollectAbstractActivity implements AdminPasswordDialog.AdminPasswordDialogListener {
 
-    private static final int PASSWORD_DIALOG = 1;
+    private static final String KEEP_BANNER_WITH_SUCCESS_MSG_VISIBLE = "keepBannerWithSuccessMsgVisible";
```
seadowg: Maybe I'm misinterpreting what @lognaturel was getting at before but I don't think we need any state around the banner at all. If the activity is recreated (rotation/out of memory) I think the `LiveData` observation on the result will take care of continuing to show the dialog (as the result is stored in a singlton). This is assuming we get rid of the `consumeResult` call.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -585,6 +559,26 @@ private boolean loadSharedPreferencesFromFile(File src) {
...
     }
 
+    @Override
+    public void onCorrectAdminPassword(AdminPasswordDialog.Action action) {
+        switch (action) {
+            case ADMIN_SETTINGS:
+                startActivity(new Intent(this, AdminPreferencesActivity.class));
+                break;
+            case STORAGE_MIGRATION:
+                StorageMigrationDialog storageMigrationDialog = (StorageMigrationDialog) DialogUtils.getDialogFragment(StorageMigrationDialog.class, getSupportFragmentManager());
```
seadowg: Could this just use `showIfNotShowing` instead? and call `startStorageMigration` on the returned dialog?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/AdminPasswordDialog.java
```diff
@@ -0,0 +1,93 @@
...
+    private AdminPasswordDialog(AdminPasswordProvider adminPasswordProvider, Action action) {
+        this.adminPasswordProvider = adminPasswordProvider;
+        this.action = action;
+    }
+
+    @Override
+    public void onAttach(@NotNull Context context) {
+        super.onAttach(context);
+        try {
+            listener = (AdminPasswordDialogListener) getActivity();
```
seadowg: I feel like it might make more sense to use a `instanceof` check here rather than caching the `ClassCastException`. This means we don't get warnings if we're using the `Fragment` in contexts where we're not interested in the listener (tests for instance).

collect_app/src/main/res/layout/storage_migration_banner.xml
```diff
@@ -0,0 +1,60 @@
...
+        android:paddingTop="@dimen/margin_standard">
+
+        <TextView
+            android:id="@+id/storageMigrationBannerText"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textColor="?colorOnSurface">
+        </TextView>
+
+        <Button
```
seadowg: I think it would be good for us to start using `MaterialButton` instead for new views like this. Should mean that you don't have to set `textAllCaps`.

collect_app/src/main/res/layout/storage_migration_banner.xml
```diff
@@ -0,0 +1,60 @@
...
+            android:id="@+id/storageMigrationBannerDismissButton"
+            style="@style/Widget.MaterialComponents.Button.TextButton"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="end"
+            android:text="@string/scoped_storage_dismiss"
+            android:textColor="?colorAccent"
```
seadowg: Lets use `colorSecondary` for this. They're the same but `colorSecondary` is the "official" name: https://material.io/develop/android/theming/color/

collect_app/src/main/res/layout/storage_migration_banner.xml
```diff
@@ -0,0 +1,60 @@
...
+            android:paddingTop="12dp"
+            android:paddingBottom="@dimen/margin_small"
+            android:textStyle="bold"
```
seadowg: I think we should avoid adding bold and just use the standard style here. If people disagree maybe we should look at changing the button text style (`textAppearanceButton`) for everything.

collect_app/src/main/res/layout/storage_migration_banner.xml
```diff
@@ -0,0 +1,60 @@
...
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        android:paddingStart="@dimen/margin_standard"
+        android:paddingEnd="@dimen/margin_standard"
+        android:paddingTop="@dimen/margin_standard">
+
+        <TextView
+            android:id="@+id/storageMigrationBannerText"
```
seadowg, line 19: I believe this should be using `TextAppearance.Collect.Body2` as the style looking at the [specs](https://material.io/components/banners/banners.html#specs) and the [style guide](https://github.com/seadowg/collect-design/releases/download/0.2.0/collect-styleguide.pdf) I generated from Sketch.

collect_app/src/main/res/layout/storage_migration_dialog.xml
```diff
@@ -0,0 +1,137 @@
...
+                android:layout_height="wrap_content"
+                android:layout_marginTop="@dimen/margin_small"
+                android:text="@string/storage_migration_dialog_message3" />
+
+            <TextView
+                android:id="@+id/moreDetailsButton"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="@string/storage_migration_more_details"
+                android:textColor="?colorAccent"
```
seadowg: Again we should be using `colorSecondary` instead of `colorAccent`

collect_app/src/main/res/layout/storage_migration_dialog.xml
```diff
@@ -0,0 +1,137 @@
...
+
+            <TextView
+                android:id="@+id/messageText3"
+                style="@style/TextAppearance.Collect.Body1"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="@dimen/margin_small"
+                android:text="@string/storage_migration_dialog_message3" />
+
+            <TextView
```
seadowg: This can be a `MaterialButton` with the `TextButton` style right?

collect_app/src/main/res/layout/storage_migration_dialog.xml
```diff
@@ -0,0 +1,137 @@
...
+                    android:layout_height="wrap_content"
+                    android:src="@drawable/shadow_up" />
+
+                <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal"
+                    android:gravity="end">
+
+                    <Button
```
seadowg: Again lets do `MaterialButtons` for these and see if we can get rid of the `bold` and `textAllCaps`

seadowg (dismissed):
Commented with some suggested tweaks. I was also wondering about our test coverage here. It seems to me that we also need (at least) tests to cover:
* the banner showing/hiding
* the dialogs more details, cancel and migrate buttons
* the `AutoSendWorker` and `ServerPollingJob` not running if the migration is

I think for the services these could be Robolectric tests that fake out the `StorageMigrationRepository`. For the UI stuff I could see it a couple of Espresso tests that run though the different flows or Robolectric tests for the `MainMenuActivity` and `StorageMigrationDialogFragment`.

collect_app/src/androidTest/java/org/odk/collect/android/ResetAppStateTest.java
```diff
@@ -166,7 +166,7 @@ private void setupTestSettings() throws IOException {
 
         assertTrue(new File(storagePathProvider.getDirPath(StorageSubdirectory.SETTINGS)).exists() || new File(storagePathProvider.getDirPath(StorageSubdirectory.SETTINGS)).mkdir());
         assertTrue(new File(storagePathProvider.getDirPath(StorageSubdirectory.SETTINGS) + "/collect.settings").createNewFile());
-        assertTrue(new File(storagePathProvider.getDirPath(StorageSubdirectory.ODK) + "/collect.settings").createNewFile());
+        assertTrue(new File(storagePathProvider.getStorageRootDirPath() + "/collect.settings").createNewFile());
```
lognaturel, line 170: Ooh, yeah, the odk subdirectory was from the StoragePathProvider PR. Sorry I didn't think through that before! Looks great now.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
lognaturel: I think there are two reasons that it's still here. We want the success banner to be user-dismissible so this currently is what keeps track of whether the user has dismissed it or not. I think an alternative would be to clear/consume/nuke the livedata value on dismissal. I think the other is that the success banner should only hold the success text where as error banners keep notice test above the error. I think that could be done by making the case based on the livedata observed value.

collect_app/src/main/java/org/odk/collect/android/utilities/AdminPasswordProvider.java
```diff
@@ -0,0 +1,22 @@
+package org.odk.collect.android.utilities;
+
+import org.odk.collect.android.preferences.AdminSharedPreferences;
+
+import static org.odk.collect.android.preferences.AdminKeys.KEY_ADMIN_PW;
+
+public class AdminPasswordProvider {
```
lognaturel, line 7: Nice.

lognaturel:
Thanks for all the hard work on this, @grzesiek2010. It’s been an opportunity to have really good architectural discussions. I wish we had split it up a bit but without a dedicated PM and a finished spec to work from I think we did the best we could. 

I think @seadowg’s various small comments are good ones. Maybe you can change the minSdk here so there isn’t the lint failure.

I too would like to see some additional automated tests but I think that shouldn’t stop QA from happening. So as soon as you’ve fixed conflicts (sorry) and addressed the last few small things, please hand over to @opendatakit/testers!

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/storage_migration_dialog.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/storage_migration_dialog.xml
grzesiek2010: Done, but bold is still needed.

collect_app/src/main/res/layout/storage_migration_dialog.xml
grzesiek2010: Yes but such a text button always has it's own padding and it doesn't look good because it's not aligned with other elements (`TextViews`).

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: I don't think so then it would be like:
Screenshot_2020-02-21-09-19-12
but it should be like:
image

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: I prefer having it bold otherwise it's like:
Screenshot_2020-02-21-09-53-41

the sample from material design is with bold:
image

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/AdminPasswordDialog.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationRepository.java
grzesiek2010: Now I use it in `MainMenuActivity.onDestroy()` otherwise if you close the app and open it again the banner with success message will be visible because the status persist. We agreed that the banner in such a case should be gone.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: I improved the code a bit removing that field and using just livedata + clearing the result in onDestroy();

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: Uuuups I don't understand... `showIfNotShowing` what? A user should be asked for password when they clicks on `MIGRATE` button from `StorageMigrationDialog` not `LEARN MORE AND MIGRATE` from the `Main Menu`. Am I missing something or maybe it's you :smile:

grzesiek2010:
> Maybe you can change the minSdk here so there isn’t the lint failure.

It would cause other errors so I added the os version check back for now.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
seadowg: hah yeah sorry I meant `DialogUtils.showIfNotShowing`. Does that make sense? I wouldn't be surprised if I have the flow wrong in my head here!

collect_app/src/main/res/layout/storage_migration_banner.xml
seadowg: @grzesiek2010 are you sure that's `Body2`? It looks like it might be the `Button` style.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: I know you meant `DialogUtils.showIfNotShowing` but there is nothing to show here.

seadowg: I guess my head it's a nice way of "fetching" while also ensuring it's definitely on screen. Here it looks like you do expect the dialog to be showing. Is that right? If so there is a case here where if the dialog for some reason **isn't** showing then this is just a dead end (as `storageMigrationDialog != null` wouldn't pass). The code could just be:

```java
case STORAGE_MIGRATION:
    DialogUtils.showIfNotShowing(new StorageMigrationDialog(), getSupportFragmentManager()).startStorageMigration();
```

If we did that we should probably change `showIfNotShowing` to take a class rather than an instance (to not waste allocations).

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: Right i did something wrong. Fixed.

collect_app/src/main/res/layout/storage_migration_dialog.xml
seadowg: Ah because of the left/right padding? Personally I think I'd rather we use the component and style and then customize the padding or margins around it to get it to look right. That way most of the styling still comes from the Material Components lib.

collect_app/src/main/res/layout/storage_migration_banner.xml
seadowg: Hmmm yeah it feels to me like the button examples they have are using a heavier font than I can seem to get access to using the Material Component styles. We're looking at changing the Main Menu buttons to be Material Buttons which means us thinking about these styles so I'm happy to just go with bold on this for now if you prefer it and we can play with different options then.

So this is easier for us to play with globally could you make a new style in `typography.xml` for the buttons (`TextAppearance.Collect.Button`) that inherits from `TextAppearance.MaterialComponents.Button` with the bold `textStyle`? If we set it as the `textAppearanceButton` in our themes that will make the bold style apply to the `MaterialButton` objects (with the `TextButton` style here of course).

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: Ok, now I understand what you mean, fixed. Thanks!

seadowg: Sorry I absolutely should have just put the example in at first!

collect_app/src/main/res/layout/storage_migration_banner.xml
grzesiek2010: You mean like https://github.com/opendatakit/collect/pull/3623/commits/713f6865a307681891c93ef04c816a260ecb51ba?
Is this ok?

seadowg: Not exactly. On the button itself (the `MaterialButton`) you still have:

```xml
style="@style/Widget.MaterialComponents.Button.TextButton"
```

Then in `typography.xml` we add:

```xml
<style name="TextAppearance.Collect.Button" parent="TextAppearance.MaterialComponents.Button">
        <item name="android:textStyle">bold</item>
</style>
```

Then in `light_theme.xml` and `dark_theme.xml` we change the `textAppearanceButton` attribute so that we have:

```xml
<item name="textAppearanceButton">@style/TextAppearance.Collect.Button</item>
```

The Material Component styles use the `textAppearanceButton` theme attribute automatically.

grzesiek2010: Ok done.

collect_app/src/main/res/layout/storage_migration_dialog.xml
grzesiek2010: Done.

lognaturel:
> It would cause other errors so I added the os version check back for now.

Ack, ok, sorry about the bad idea!

lognaturel:
I talked to @grzesiek2010 and @opendatakit/testers today. The plan is to get #3625 verified and merged as soon as possible so @grzesiek2010 can fix any conflicts and start exercising those changes. @grzesiek2010 and @opendatakit/testers will then work side by side to verify all of the behavior here.

@seadowg I dismissed your review as I think everything you've brought up has been addressed. One of us can do a last review pass once QA changes have been made and before merging.

mmarciniak90:
I noticed that error on a small screen in horizontal orientation is unreadable. Part of the text appears on the bottom bar and `CANCEL` / `TRY AGAIN` buttons. The rest is not visible at all. 
@grzesiek2010 suggested that we can move them up. 
We can also try with scroll.
Or maybe you think it's not worth the attention.

@lognaturel let us know what you think about this.

| | | |
|---|---|---|
|6835D9B7-7A67-4047-B1E0-14C205CADD57|88977999-7083-4815-82D2-76BC4A07228D|20BD0C92-5D49-402A-95FD-001AA0A34C7B|

lognaturel:
I'm surprised that the text is stacking in that way. Does this mean that there will never be any scrolling and that the bottom bar is not actually on top of the main part of the full screen dialog? I think that this particular case isn't a big deal but there are devices that have much smaller screens and I wonder how the message even without an error would render. Ideally text would scroll when it reached the top of the bottom bar.

grzesiek2010:
Yes it should contain scroll that for sure, I'll fix it but what about moving that error message to the top to be sure it's visible without scrolling?

lognaturel:
> what about moving that error message to the top to be sure it's visible without scrolling?

I don't have strong expectations about where it should go and I think you can do whatever is easiest to make sure it is visible. That could be putting it at the bottom and scrolling so it's always on screen or putting it at the top. Probably the ideal would be for it to go in a separate visual element (banner?) but this is one-time thing so let's not go down that path.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationBannerTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.odk.collect.android.support.StorageMigrationNotPerformedRule;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+public class StorageMigrationBannerTest {
+
+    @Rule
+    public ActivityTestRule<MainMenuActivity> main = new ActivityTestRule<>(MainMenuActivity.class);
+
+    @Rule
+    public StorageMigrationNotPerformedRule storageMigrationNotPerformedRule = new StorageMigrationNotPerformedRule();
```
seadowg: Really liking this use of `@Rule`! I think it'd be good to still have the `ResetStateRule` and standard permssion rules for these tests just to make sure they are starting fresh.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationBannerTest.java
```diff
@@ -0,0 +1,36 @@
...
+                .assertStorageMigrationBannerIsDisplayed()
+                .rotateToPortrait(new MainMenuPage(main))
+                .assertStorageMigrationBannerIsDisplayed()
+                .reopenApp()
+                .assertStorageMigrationBannerIsDisplayed();
+    }
+
+    @Test
+    public void when_learMoreButtonClicked_should_storageMigrationDialogAppear() {
+        new MainMenuPage(main)
```
seadowg: I think you'd be fine to combine this with the previous test as one flow. I think we can also move rotation/reopen testing out of Espresso for these cases but I'll have a quick think about that and get back to you on that.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationDialogTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.odk.collect.android.activities.MainMenuActivity;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.StorageMigrationNotPerformedRule;
+import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.android.support.pages.StorageMigrationDialogPage;
+
+public class StorageMigrationDialogTest {
```
seadowg: I'm wondering if it might make sense to combine these tests with the `StorageMigrationBannerTest` as one file that tests the migration flow.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+                    Manifest.permission.READ_EXTERNAL_STORAGE,
+                    Manifest.permission.WRITE_EXTERNAL_STORAGE
+            ))
+            .around(new StorageMigrationNotPerformedRule())
+            .around(new ResetStateRule())
+            .around(new CopyFormRule("basic.xml", true));
+
+    @Test
+    public void when_migrationIsFinishedWIthSuccess_should_storageMigrationDialogDisappear() {
+        new MainMenuPage(main)
```
seadowg, line 41: Same again: feels like we might be able to just combine these with the other migration tests

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationBannerTest.java
seadowg: Yeah you should be able to use the `ActivityScenario` in [AndroidX testing](http://robolectric.org/androidx_test/) and just call `recreate` to test recreation (otation is really just that). We're using it for a fragment test in `FormMetadataFragmentTest`. We should be able to use this kind of pattern to test the `MainMenuActivity` directly rather than having to worry about rotation/recreation at the Espresso level.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigratorTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import timber.log.Timber;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.core.IsNot.not;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+
+public class StorageMigratorTest {
```
seadowg, line 34: I'm wondering if there is a way we could test that the StorageMigrator does the right thing when these jobs are runnign without actually having to start them up (and also have tests that run in the JVM). As far as I can see these tests are non deterministic due to the waits and background work. 

I know Google released some testing around `WorkManager` so I can have a look and see if there is anything we can use there. Ideally we can fake the state of work or have it run paused or something. This wouldn't solve for the `JobManager` stuff though, Alternatively, for the moment we could hide the logic around looking up the statuses of these pieces of background work behind an interface that we can sub out for the `StorageMigrator`'s unit test (probably as a constructor arg):

```java
public interface WorkStatusProvider {
    public boolean isFormDownloaderRunning();
    public boolean isFormUploaderRunning();
}
```

This would also mean that objects like the `StorageMigrator` wouldn't have to care about the details involved of fetching those statuses.

grzesiek2010:
@lognaturel 

We discovered one serious problem with external itemsets like [sample form](https://docs.google.com/spreadsheets/d/12qZL34kuHSZGWDv0BBJ1qf7dSmml-d2VnMWH0Vtg-O4/edit#gid=683809272).

The problem is that tables in itemsets db are created using md5 hashes like here:
https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/dao/ItemsetDao.java#L40

the parameters is `itemsetFile.getAbsolutePath()`

that means a table crated using unscoped storage is different compared to scoped storage and then it can't be found.
I think the easiest way would be to remove `itemsets.db` and let it create again it won't be a problem.
What do you think?

lognaturel:
> I think the easiest way would be to remove itemsets.db and let it create again it 

I don’t recall the details of that import off the top of my head but if indeed the files are still there in their original form then that sounds fine. I can spend more time looking at it later if needed but wanted to give a first impression in case you want to make the change now. 

As for the testing improvements suggested by @seadowg, I think it’s worth doing some of the quick ones for clarity. The more involved ones could go to another PR so we can merge and start addressing conflicts with other PRs.

grzesiek2010:
> I don’t recall the details of that import off the top of my head but if indeed the files are still there in their original form then that sounds fine. I can spend more time looking at it later if needed but wanted to give a first impression in case you want to make the change now.

I have a fix but I'll add it later and add tests to confirm form with various external files work well.
Either way we need Monday for testing.

lognaturel:
> I have a fix but I'll add it later and add tests to confirm form with various external files work well. Either way we need Monday for testing.

Sounds great. There are going to be conflicts to fix across a few different PRs once this gets merged so I'd say bias towards merging Monday if you can. Even if there are some small things still pending, those can be filed as separate issues.

grzesiek2010:
> There are going to be conflicts to fix across a few different PRs

yeah but not much I guess. Most of my changes are isolated in clases like StorageMigration[Something].

lognaturel:
True. #3637 in particular also uses the admin password so it will need to pull in your changes for that. #3659 is already on top of this PR but will need to pull any new changes before it can go to QA. So it's not huge but the more time there is to do it right, the better.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationDialogTest.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationTest.java
grzesiek2010: Done.

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationBannerTest.java
grzesiek2010: Done, but when it comes to the new rule I removed it because it was called too late. I needed to override `beforeActivityLaunched` like here https://github.com/opendatakit/collect/pull/3623/commits/f6073fa3ecd523166fd21dbace8654398e2b6379#diff-031d2e263a9e3d4064f608116ef282e4R34

collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationBannerTest.java
grzesiek2010: > think you'd be fine to combine this with the previous test as one flow.

Done.

when it comes to screen rotation I used recreate in some cases for example where I wanted to simulate that the app is [reopened](https://github.com/opendatakit/collect/pull/3623/files#diff-cb71153b5f29ea480658a710c445028bR85). However in this case rotation and recreation differ. In MainMenuActivity we use
`android:configChanges="locale|orientation|screenSize"`

that means `onDestroy` won't be called if you rotate your device so I expect a different behavior compared to recreation. For example if we have the banner with success msg it should disappear after reopening the app but should be still visible after simple rotation.

kkrawczyk123:
Me and @mmarciniak90 have verified cases: 
- try to migrate when autosend is running
- successful migration after autosend finished
- successful migration after auto-send and auto-delete finished
- try to migrate when automatic form downloader enabled
- successful migration after automatic downloading finished
- try to migrate when not enough space for migration (only on Android 5.1)
- try to break migration by: screen rotation, minimizing, extinguish screen, kill app, minimizing app with don’t keep activities enabled, kill app with don’t keep activities enabled, turning off device, call to device with ongoing migration, making a screenshot, clicking on devices screen when process ongoing
- errors in horizontal orientation
- admin password dialog
- path to offline layers after migration updated
- migrate in dark mode
- migrate when the language changed
- migrate when changed settings in General Settings
- migrate when changes settings in Admin Settings
- multiclick on migrate button
- multiclick on more details button
- success dialog not restored after dismissing it
- splash image
- forms with last save, the save restored after migration
- encrypted form with last save, the save and the status restored after migration
- encrypted form
- encrypted form without instanceId
- external CSV files are migrating
- form with external CSV useable after migration, ability to fill, edit submission
- external secondary instances files are migrating
- form with external secondary instances useable after migration, ability to fill, edit submission
- reset settings - migration banner/success banner not restored
- invalid forms are migrated, they are still invalid after migration
- media files are migrated (checked different types of media)
- upgrade
- audit files are migrating
- events, value changes, background location, reason changes, identify user are visible in audits after migration  

Tested on Android 5.1, 6.0, 7.0, 8.1, 9.0 and 10. 
@grzesiek2010 @lognaturel what do you think about that? Is there anything more you have on your mind that we can verify?

lognaturel:
This looks really thorough, @kkrawczyk123, @mmarciniak90, thanks so much. I propose we merge now so that we can get the other PRs ready for your day tomorrow. You thought of some interesting things I hadn't considered so I'll also spend a little time during my day seeing if your list sparks any other ideas. @seadowg, do you want to do the merge since you've done the bulk of the review? If there are addition action items we think of, we can use issues to keep track of them.

seadowg (approved):
I think there are a bunch of improvements we want to make to tests but it does make sense to merge now (mainly to avoid conflicts moving forward) and build towards those. I can look at adding issues around any test refactors or extra coverage.

PR #3627, opened 2020-02-06
Add v1.26 beta translations

lognaturel:
Add translations for early v1.26 beta so that translators can see translated repeat text in action.


yanokwa (approved):
LGTM

PR #3628, opened 2020-02-08
Move .settings file out of Import/Export and rename to Configure via QR Code

### LINKED ISSUE #3618: Move saving .settings file out of Import/Export settings and rename Import/Export settings to QR code configuration
I was talking to @andrewsiew2 about the QR code settings feature and he pointed out that having QR code configuration and the `.settings` file export on the same screen is confusing. We'd rather use a name like `QR code configuration` instead of `Import/Export settings`. We can't currently because the screen doesn't just do QR code configuration.

I propose pulling the saving to `.settings` functionality out of the current screen and putting it directly in Admin Settings right above the `User Access Control` section with name `Save legacy .settings file to disk". The details text should say "Will be removed in a future version. Please use the QR code above.". There should be analytics on this settings option so we can track usage.

Once that is done, the existing Import/Export Settings screen can be renamed to `QR code configuration`.

I chose not to put this on the forum because it seems like a relatively minor change. I think that if @andrewsiew2 agrees with this proposal and no one objects after a few days, then we can remove the `needs discussion` tag and proceed.

hypercubestart:
Closes #3618 

This change moves the .settings file out of Import/Export Settings and moves it under a new header in Admin Settings called App Configuration with
- Reset application...
- Configure via QR code
- Save configuration to file
    - Will be removed in a future version

as proposed by @andrewsiew2 and @lognaturel and thanks to suggestions from @yanokwa . This will reduce confusion of having QR Code and the .settings save file on the same screen. Changing `Import/Export Settings` to `Configure via QR Code` will also be more intuitive for users.


seadowg (approved):

seadowg:
@hypercubestart this looks all good to me! It looks the build is failing as you need to delete some the translated versions of the deleted `save_preferences` string. Once that's fixed we should be good to hand this to QA.

hypercubestart:
@seadowg thanks for the review! Just fixed the build tests. How do we hand this to QA?

seadowg:
@hypercubestart don't worry that's my job!

mmarciniak90:
Hi @hypercubestart!

I'm testing this PR and I noticed different text than in issue.
I expected 

> Save configuration to file
>  Will be removed in a future version

But I see 

> Save legacy .settings file to disk
> Will be removed in a future version. Please use the QR code above.

It's not a big difference but was it determined?

seadowg:
@mmarciniak90 ah it looks like @yanokwa and @lognaturel had different versions of the text. I'd say if you're happy with everything else let's label this "behaviour verified" as we can tweak the language easily after a merge if there are strong feelings.

mmarciniak90:
@seadowg yes, everything else seems to be good.

Verified cases:
* save collect.settings file via a new option
* save icon removed
* new option split
* reset
* QR code recreation
* failed file with QR code
* successful load QR code from the file
* successful scan QR code with settings 
* light/dark mode

PR #3629, opened 2020-02-09
Parallelizing Instrumentation Tests using flank

### LINKED ISSUE #3578: Look at parallelizing instrumentation tests
Right now our tests take 20m+ to run. This is causing timeout issues with CI and I don't think we want to have to keep bumping our timeout values every time the test suite grows. We do also need an audit of test coverage we can move from Espresso to Robolectric/JUnit but I think we expect this suite to stay at least this big in the long term. 

It'd be good to look into parallelizing our tests. I think the focus for this should be for CI (with Firebase Test Lab). Parallelizing locally is potentially a bad idea due to the hit on people's dev environments. Locally I think continuing to break tests into concentrated domain areas will help devs have smaller sets of tests to run while working on features/refactors.

chidauri:
Closes #3578 

Currently, i am using my service account for testing and i've verified that the changes work as intended by running the build on CircleCi.
Current setting creates 5 different matrices which run in parallel. One or two(max.) of the matrices fail sometimes due to a single test(but not the same set of tests everytime) being flaky. I've fixed this currently by using `num-flaky-test-attempts: 1` and it happens the test which fails in the first run, passes the next time.
Currently parametrized tests are not being run, because flank uses orchestrator by default, which is useful for parallel tests, but we can turn it off if we want to run parametrized tests too.

Results:
**5 shards:** approx 10 mins
**3 shards:** approx 17 mins

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,26 +159,37 @@ jobs:
     steps:
       - *attach_debug_workspace
       - run:
-          name: Authorize gcloud
+          name: Prepare wget
           command: |
-            if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" ]]; then \
-              gcloud config set project api-project-322300403941
-              echo $GCLOUD_SERVICE_KEY | base64 --decode > client-secret.json
-              gcloud auth activate-service-account --key-file client-secret.json
-            fi
+            apt-get update
+            apt install python2.7 python-pip -y
+            apt-get install wget
+      - run:
+          name: Authorize service account
+          command: |
+            mkdir -p $HOME/.config/gcloud/
+            echo $GCLOUD_KEY | base64 -di > $HOME/.config/gcloud/application_default_credentials.json
       - run:
           name: Run integration tests
           command: |
-            if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" ]]; then \
-              echo "y" | gcloud firebase test android run \
-              --type instrumentation \
-              --app collect_app/build/outputs/apk/debug/*.apk \
-              --test collect_app/build/outputs/apk/androidTest/debug/*.apk \
-              --device model=Pixel2,version=27,locale=en,orientation=portrait \
-              --results-bucket opendatakit-collect-test-results \
-              --environment-variables coverage=true,coverageFile=/sdcard/coverage.ec \
-              --directories-to-pull /sdcard --timeout 30m
-            fi
+            wget --quiet https://github.com/TestArmada/flank/releases/download/v8.1.0/flank.jar -O ./flank.jar
+            cat <<- 'EOF' > ./flank.yml
+            gcloud:
+              timeout: 30m
+              num-flaky-test-attempts: 1
+              app: collect_app/build/outputs/apk/debug/*.apk
+              test: collect_app/build/outputs/apk/androidTest/debug/*.apk
+              device:
+                - model: Pixel2
+                  version: 27
+                  locale: en
+                  orientation: portrait
+            flank:
+              max-test-shards: 5
+              num-test-runs: 1
+              project: propane-analogy-263606
+            EOF
+            java -jar ./flank.jar android run
           no_output_timeout: 60m
       - run:
           name: Copy integration test results
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ToastMatcher.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ToastMatcher.java
@@ -1,0 +1,27 @@
+package org.odk.collect.android.support.matchers;
+
+import android.os.IBinder;
+import android.view.WindowManager;
+
+import androidx.test.espresso.Root;
+
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
+
+public class ToastMatcher extends TypeSafeMatcher<Root> {
+    @Override
+    public void describeTo(Description description) {
+        description.appendText("is toast");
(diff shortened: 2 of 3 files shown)


chidauri:
@seadowg does this look okay? Then i'll change it to use ODK service account as before.

seadowg:
@chidauri I think it would be good for us to investigate the flakey tests as I'm not comfortable with the idea that some are failing some of the time. If you get a chance would you be able to list out the tests that are failing?

Apologies as I imagine this is a bit of slow/annoying issue to work on! Thanks for keeping at it.

chidauri:
@seadowg no worries :) I'll update this comment with flakey tests as i encounter them.

Flakey Tests:
**1. [searchExpression_ShouldDisplayWhenItContainsOtherAppearanceName](https://github.com/opendatakit/collect/blob/e13082ad9c0c4ec13912f5bb652351fc199cd7ea/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java#L234)** 
[log1.txt](https://github.com/opendatakit/collect/files/4185760/log1.txt)

**2. [formsWithDate_ShouldSaveFormsWithSuccess](https://github.com/opendatakit/collect/blob/e13082ad9c0c4ec13912f5bb652351fc199cd7ea/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java#L137)**
[log2.txt](https://github.com/opendatakit/collect/files/4185764/log2.txt)

**3. [savedAndBlankForms_shouldBeReset](https://github.com/opendatakit/collect/blob/46768d010f136b389a2c3e446df32d3bf564a013/collect_app/src/androidTest/java/org/odk/collect/android/regression/ResetApplicationTest.java#L52)**
[log3.txt](https://github.com/opendatakit/collect/files/4207760/log3.txt)

**4. [when_chooseAnswer_should_beVisibleInNextQuestion](https://github.com/opendatakit/collect/blob/46768d010f136b389a2c3e446df32d3bf564a013/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java#L624)**
[log4.txt](https://github.com/opendatakit/collect/files/4207761/log4.txt)

**5. [app_ShouldNotCrash](https://github.com/opendatakit/collect/blob/46768d010f136b389a2c3e446df32d3bf564a013/collect_app/src/androidTest/java/org/odk/collect/android/regression/FillBlankFormTest.java#L398)**
[log5.txt](https://github.com/opendatakit/collect/files/4208432/log5.txt)

lognaturel:
Thanks so much for taking on this super valuable change, @chidauri!

I just was trying to run `InstanceServerUploaderTaskTest` in isolation and had to add storage read/write permissions to the `GrantPermissionRule`. That means this test would only work if another test that granted those permissions happened to be run before. I'm sure it's not the only such issue left over.

chidauri:
@lognaturel thanks for your support :)
I am not able to understand the behaviour why the flaky tests which fail in the first run, pass in the second run everytime (even though only the matrix which fails is made to run again).

EDIT: [This](https://engineering.shopify.com/blogs/engineering/unreasonable-effectiveness-test-retries-android-monorepo-case-study) article states simply by retrying pass rate can increase significantly, so maybe this is the reason they pass the second time.

lognaturel:
> even though only the matrix which fails is made to run again

Is the order always guaranteed to be the same? If not, the permissions could still be the explanation.

chidauri:
@lognaturel I checked the log videos, the order is the same for both runs. Also same set of tests are part of a matrix everytime in the same order.

seadowg:
> EDIT: This article states simply by retrying pass rate can increase significantly, so maybe this is the reason they pass the second time.

That's an interesting article. Short answer: I think I'd like to avoid retries as I'm worried introducing the mechanism will mean that flakey tests will remain in the codebase and will cause us to lose trust in our build. I'm interested in if @lognaturel's theory around permission could be the fix.

As an aside because I love chatting about this stuff: the article's premise that they are aiming for a high "pass rate" (percentage of the time the build passes) really feels off to me. The build going red (failing) isn't a bad thing - it's an indication that's something wrong and exactly the reason we have it. If the build fails "erroneously" because a test is flakey or wrong then that is something we need to fix or we lose the value of the build failing and it meaning something. Of course the Collect suite isn't quite there yet but I really want to get it to that place 😉

lognaturel:
> the order is the same for both runs. Also same set of tests are part of a matrix everytime in the same order.

Good to know. That certainly makes the permissions hypothesis a lot less likely. Still, it would be a great thing for this PR to fix as well. 

I just got a local failure of `SpinnerWidgetTest.spinnerList_ShouldDisplay` that went away on re-run. I'm guessing that's an animation timing issue. Are the two at https://github.com/opendatakit/collect/pull/3629#issuecomment-584138789 still the only ones that have failed for you so far?

chidauri:
> >
>Are the two at [#3629 (comment)](https://github.com/opendatakit/collect/pull/3629#issuecomment-584138789) still the only ones that have failed for you so far?

Yup! Also these have been encountered multiple times, so something is wrong with these two.

chidauri:
@lognaturel @grzesiek2010 should i stop working on this PR as #3679  accomplished this?
And don't we face flaky tests that way?

seadowg:
@chidauri yeah it looks like Firebase have released a feature that means we don't need Flank. It's still in beta though. 

It looks like you added a new `ToastMatcher` though so that would probably still be good to merge if you found it worked better! Maybe for the moment you could open a  new PR with just that? I'll close this for the moment but we might need to come back to it if the Firebase solution doesn't work.

PR #3632, opened 2020-02-10
Display new Install ID in Form Metadata preferences

### LINKED ISSUE #3595: Generate install-specific uuid and display it in form metadata settings
Starting in fall 2020, we will be required to target SDK 29 which means we will no longer be able to use the IMEI as `deviceId`. See https://forum.opendatakit.org/t/collect-will-need-to-stop-using-imei-as-deviceid-and-making-simserial-and-subscriberid-available/24157 and https://github.com/opendatakit/collect/issues/3360 for context.

We have agreed to replace it by generating and managing a uuid to identify the current install. The UUID should be a RFC 4122 Version 4 UUID generated by https://developer.android.com/reference/java/util/UUID.html#randomUUID() encoded to base64. That should bring it to 22 characters. It should be prefixed by `uuid:`.

That UUID should be lazily created the first time that it is requested and stored in general preferences. It should be made visible in the Form Metadata settings below `Device ID` with the label: "Install ID - will replace Device ID in August 2020"

It should be possible to copy this ID to the device clipboard.

Marking as `needs discussion` as long as the forum discussion is still running.

seadowg:
Closes #3595

This displays the newly generated "Install ID" in the Form Metadata preferences screen. Right now there is no way to copy the ID but we'll break this into a seperate issue.

One thing to note is that I moved a couple of the `regression` tests that dealt with this screen out into a new test "feature" test (tests that drive out the core of a feature without checking edge cases etc) focused on the metadata screen.

Added new tests for the Form Metadata preferences and drove this out wtih feature/unit tests. Also played around on device.

There were many discussions around the best way to generate/implement the install ID. See the issue for that.

Should just add the new install ID. I did touch the `FormMetadataFragment` so double checking that screen is probably the only thing worth looking at.


lognaturel (approved):

seadowg:
Just ran this on Test Lab and saw we have some failure in the new tests as we get `null` back from `PropertyManager` for some device details there. I'll sort that out so the tests can cope with that. Won't block QA but don't merge until I fix!

kkrawczyk123:
@seadowg I have noticed that both Device ID and Install ID content disappears after screen rotation and is not restored after rotating back to portrait. However, it has not been introduced in this pr and it is also visible for Device ID on the master and release version. Do you think it is something that you could fix in this pr or it would be better to open a new issue for that?

seadowg:
@kkrawczyk123 Great find. I'm reworking this screen to add copying to the Install ID so will fix as part of that in another PR.

kkrawczyk123:
@seadowg I have another question - after resetting settings new install id is generated - is it accidental or on purpose?

seadowg:
@kkrawczyk123 yeah that makes sense as doing that clears out app data. We could work around it but I'd say we can have that as another issue if we think we need it.

kkrawczyk123:
**Tested with success!**

Verified cases: 
- install ID added to form metadata
- info about replacing it in August added - text is wrapped
- light mode
- dark mode
- instal id not exported via qr code
- new install id in each install process
- no changes after killing an app
- new install id after settings reset, on lower Androids: 4.2 and 4.4 install id content disappears after the reset - probably it's not important as we stop support Android<5 soon.
- rotation is causing the issue but already mentioned above. 

PR #3637, opened 2020-02-11
Scan Qr code main menu button

andrewsiew2:
Based on https://forum.opendatakit.org/t/add-button-to-import-collect-settings-on-collect-main-menu/23876/7 feature request.

This feature adds a Scan QR code button in the overflow menu that can be hidden in the main menu settings. This feature was modified based on the feedback received the forum post comments.


seadowg:
@andrewsiew2 heyo! Would you be able to clean up the history so it doesn't include all the reverts and merges. Also it looks `pmd` is failing on CI for this PR. Could you fix these and then make sure to run `./gradlew checkAll` on your changes to check everything is all good. It's on our checklist in the PR template which you can check off rather deleting so we know you've made sure the PR is safe to merge in! I'll edit the PR description so it includes the template questions and todo list again for you to fill in.

seadowg (changes requested):
As above

andrewsiew2:
@seadowg . Hey I just fixed the gradle issues. As for cleaning the history, its a bit complicated to do it on my existing branch but we can do a squash merge when we merge it into the main branch.

seadowg:
@andrewsiew2 you can use interactive rebase (`git rebase -i`) to squash commits together on a branch. Some reviewers like to go commit by commit for reviews so getting into a state where that's possible is often a good idea. I'm happy to look at the changes as is though so can squash as part of the merge as you suggest!

seadowg (changes requested):
I'd like to make sure that we have tests for this change to minize the risk that the feature regresses in the future. I'd probably look at their being at least two tests:

1. A test checking that when you click the new "Scan QR code" menu item it navigates to the scanner
2 A test checking that the "Scan QR code" item does not appear when the preference is disabled/unchecked

I think these tests could either be Espresso tests or Robolectric tests for the `MainMenuActivity`. Happy to help out if you haven't used those test frameworks before!

I'm also wondering if it makes sense to pull out the QR code scanning into a seperate activity that both `ShowQRCodeFragment` and `MainMenuActivity` can use rather than having to add the Intent flags to `AdminPreferencesActivity`. What do you think? I'll pull in @lognaturel to see if she agrees or not.

collect_app/src/main/res/values/strings.xml
```diff
@@ -506,7 +507,7 @@
     <string name="qrcode_without_passwords">This QR code does not contain your admin or server passwords. \n(Tap to add the passwords)</string>
     <string name="qrcode_with_admin_password">This QR code includes your admin password. Please treat it carefully. \n(Tap to remove the password)</string>
     <string name="qrcode_with_server_password">This QR code includes your server password. \n(Tap to remove the passwords)</string>
-    <string name="import_export_settings">Configure via QR Code</string>
+    <string name="import_export_settings">Configure via QR code</string>
```
lognaturel: You can remove this resource and use `import_settings` instead. In fact, I recommend renaming that resource to `configure_via_qr_code`.

lognaturel:
I agree that this change is fine to review as one diff so no need to modify history. It is a good thing to get to know how to do, though! [This howto](https://www.internalpointers.com/post/squash-commits-into-one-git) goes step by step on how to squash all existing commits.

Agreed that tests should be added. I think they should both be possible with Robolectric. https://github.com/opendatakit/collect/pull/3552/files#diff-e5569bbccd110c47e7ca03420c4fb7a9R195 shows an example of checking that when something is tapped, the correct intent is passed to the system. See the `@Before` block for setting up the activity.

> I'm also wondering if it makes sense to pull out the QR code scanning into a seperate activity that both ShowQRCodeFragment and MainMenuActivity can use rather than having to add the Intent flags to AdminPreferencesActivity

Yes, I very much agree that having to launch the admin preferences from that overflow menu button is strange and hard to follow. It's also unusual to fake a user button click to get certain functionality.

The challenge is that you want to use `ShowQRCodeFragment` to process the result of the scan so you can't directly launch the xzing scanning, right? Could you just add `ShowQRCodeFragment` to the backstack? Is there a reason I'm missing why you didn't do it that way? To go straight to the camera, you could use [`setArguments`](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#setArguments(android.os.Bundle))

That said, I think it would also be fine to just start `ShowQRCodeFragment` as-is for now since you're also working on improvements to that screen.

In general, please be sure to fill out the entire pull request template. The question about why this is the best solution gives you an opportunity to discuss challenges you ran into and code structure alternatives that you considered.

You also want to be sure to spend some time thinking about the risk that your PR poses. Is it modifying any code that could have unintended consequences?

hypercubestart:
@lognaturel yes you are right, our challenge was to directly open the `ScannerWithFlashlightActivity` but we needed the `ShowQRCodeFragment` to process the result. Our main concerns when implementing this feature were:
- didn't want to duplicate code for processing the qr code
- wanted `onBack` press from camera to go straight back to the main activity, rather than to `ShowQRCodeFragment`
- wanted to keep changes localized to the specific activities/fragments

I am not too familiar with directly manipulating the backstack as per your suggestion and it seems like a good idea. My only concern would be that it still feels a little unnatural, because the main activity would need to instantiate the `AdminPreferencesActivity`, which would need to first commit the `ShowQRCodeFragment`. The main activity would then need to instantiate the `ScannerWithFlashlightActivity` on the top of the stack, and affiliate the `ScannerWithFlashlightActivity` with the `ShowQRCodeFragment` in a way consistent with `onActivityResult` so that `ShowQRCodeFragment` would be able to handle the result. I may be missing something, but this would be my understanding to how it would work.

In this sense, I feel like creating a separate activity that both `MainMenuAcitivity` and `ShowQrFragment` can use like @seadowg suggested may be a better idea. However, this may limit the future design over the ShowQRFragment, because ShowQRFragment will always need to rely on a separately activity to scan qr code.

lognaturel:
> directly manipulating the backstack

To be clear, I'm just suggesting the normal way that `Fragment`s are added with fragment transactions.

> because the main activity would need to instantiate the AdminPreferencesActivity

I'm not clear on why `AdminPreferencesActivity` would need to be involved?

hypercubestart:
> To be clear, I'm just suggesting the normal way that Fragments are added with fragment transactions.

Sorry, I misunderstood the backstack to be the activity stack. 

So are you suggesting that we add `ShowQRCodeFragment` to the backstack after `AdminPreferencesFragment` is transacted? Currently, we are only adding the `ShowQRCodeFragment` to the stack so that onBackPressed will take the user directly back to the main menu activity rather than to `AdminPreferencesFragment`. This way, we don't have to handle any special case when the user presses the back button.


> I'm not clear on why AdminPreferencesActivity would need to be involved?

Since `ShowQRCodeFragment` is a fragment, it needs to be hosted by an activity mentioned in [here](https://developer.android.com/guide/components/fragments). AdminPreferencesActivity needs to be involved to host `ShowQRCodeFragment`.

> The challenge is that you want to use `ShowQRCodeFragment` to process the result of the scan so you can't directly launch the xzing scanning, right? Could you just add `ShowQRCodeFragment` to the backstack? Is there a reason I'm missing why you didn't do it that way? To go straight to the camera, you could use [`setArguments`](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#setArguments(android.os.Bundle))

Doesn't the xzing scanner need to be launched directly from `ShowQRCodeFragment` because `ShowQRCodeFragment` needs to be able process the result?

seadowg:
> To be clear, I'm just suggesting the normal way that Fragments are added with fragment transactions.

Are you meaning from the `MainMenuActivity` here @lognaturel? For that we'd need to add a container view to the Activity and also have it switch between it's main view and that when we add the Fragment. I'm totally onboard with avoiding using the `AdminPreferencesActivity` though. Two ways to avoid it come to mind:

1. Create a new Activity (`ShowQRCodeActivity`) that launches when you click "Scan QR code..." (in either the Admin Preferences or Main Menu) that just hosts the `ShowQRCodeFragment`. This would still need to be able to support the "straight to camera" mode with Intent flags.
2. Create a "proxy" Activity (`ScanQRCodeActivity`) with no view that encapsulates the scanning logic i.e. its `onCreate` is basically just `ShowQRCodeFragment#scanButtonClicked` and it handles the `onActivityResult` step. This could be started from the Main Menu "Scan QR code..." and the ShowQRCodeFragment when the scan button is clicked.

I'd tend towards 2) as it lets us have one componenet that knows about scanning that can be used in both places as is rather than having a shared component with variable behaviour.

lognaturel:
> For that we'd need to add a container view to the Activity and also have it switch between its main view and that when we add the Fragment.

Ah, yes, I see. That's what I had in mind but I'm liking your option 2 better as well!

hypercubestart:
alright, if everyone is okay with option 2, should I go ahead and create a new branch to start implementing? Reverting the changes on this current branch is also an option, but may be messier.

seadowg:
@andrewsiew2 happy to continue on this branch so we can keep the thread of our conversation

lognaturel:
@andrewsiew2 consider hopping into Slack (http://slack.opendatakit.org/) if you want to discuss some architectural or git history decisions live. I'd tend to revert all changes on your current branch (`git reset upstream/master`) and then work from there. You can cherry-pick commits that are still relevant or just copy them in. Then when you're ready to share, you can `git push --force` and we'll see your new commits right here.

collect_app/src/androidTest/java/org/odk/collect/android/settings/MainMenuSettingsTest.java
```diff
@@ -0,0 +1,49 @@
...
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class MainMenuSettingsTest {
+    @Rule
+    public ActivityTestRule<MainMenuActivity> activityRule =
+            new ActivityTestRule<>(MainMenuActivity.class);
+
+    @Test
+    public void configureWithQRCanBeDisabled() {
+        openActionBarOverflowOrOptionsMenu(getInstrumentation().getTargetContext());
```
lognaturel: We have some nice abstractions that you can use in `MainMenuPage` that will make this test even cleaner.

lognaturel (commented):
The structure is looking great. I'll let @seadowg do a final code review since he had the good architecture ideas.

As I try this, once thing I notice is that it's not very clear what the qr code scanning will do. How about using the same "Configure via QR code" string in that menu? It's long but doesn't look too long to me. If we go that route, I think we should put a comment above the string resource for translators that says something like "This phrase will be displayed in the main screen options menu. Please try to keep it as short as possible."

Going straight to the scanning interface doesn't help with that either. Hopefully once the QR code screen [is redesigned](https://forum.opendatakit.org/t/qr-code-screen-redesign/24996), the overflow menu item can go to that new screen and provide new context.

lognaturel (changes requested):
I just realized that there's a piece of functionality missing. It's possible to [set an admin password](https://docs.opendatakit.org/collect-settings/#admin-settings). The goal of that admin password is to restrict access to certain functionality. When a user taps on the overflow menu button you've added, they should be prompted for the admin password if one is set. This should be the same as when the user attempts to open the admin settings screen.

hypercubestart:
I'm having an error in the Robolectric test where 
https://github.com/opendatakit/collect/blob/9a856959b9cf4e44e6ef7264c133c74e05a89b06/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java#L375-L376

`adminPreferences` is null in the above code. It doesn't look like a regression because although a previous test is also failing for the same reason, previously, the code was structured differently such that this code was never executed by the Robolectric test. 

https://github.com/opendatakit/collect/pull/3637/commits/a6b1bc87d9992f4e8aa73fb2df8d7ab3f6311fa1 should fix this error, I couldn't figure out why `adminPreferences` was not being initialized correctly by Robolectric

hypercubestart:
@seadowg can you please re-review? thanks!

collect_app/src/androidTest/java/org/odk/collect/android/settings/MainMenuSettingsTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.odk.collect.android.activities.MainMenuActivity;
+import org.odk.collect.android.support.pages.AdminSettingsPage;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+import androidx.test.filters.LargeTest;
+import androidx.test.rule.ActivityTestRule;
+import androidx.test.runner.AndroidJUnit4;
+
+@RunWith(AndroidJUnit4.class)
+@LargeTest
```
seadowg: You can delete this annotation. Google always uses in it examples but we don't split our tests by size so it's not somthing we need!

collect_app/src/androidTest/java/org/odk/collect/android/settings/MainMenuSettingsTest.java
```diff
@@ -0,0 +1,34 @@
...
+public class MainMenuSettingsTest {
```
seadowg: Pretty much all our tests need permissions granted to run on fresh devices and so the test doesn't get polluted by others (another test bmight disable the QR code configuration for instance) you need to clear state before running the test. To do so you can add this rule chain:

```java
@Rule
public RuleChain copyFormChain = RuleChain
    .outerRule(GrantPermissionRule.grant(
        Manifest.permission.READ_EXTERNAL_STORAGE,
        Manifest.permission.WRITE_EXTERNAL_STORAGE,
        Manifest.permission.READ_PHONE_STATE
    ))
    .around(new ResetStateRule());
```

seadowg: Also I think we should call this test `MainMenuTest` for now

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -357,21 +356,38 @@ protected void onPause() {
...
 
+    @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        qrcodeScannerMenuItem.setVisible(this.getSharedPreferences(AdminPreferencesActivity.ADMIN_PREFERENCES, 0).getBoolean(AdminKeys.KEY_QR_CODE_SCANNER, true));
+        return super.onPrepareOptionsMenu(menu);
+    }
+
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
+        SharedPreferences adminPreferences = this.getSharedPreferences(
```
seadowg: This should really be initialized in `onCreate` as `onOptionsItemSelected` won't be called until someone clicks on a menu item. This means `adminPreferences` will have a high likelyhood of being `null` if someone tries to call it somewhere else.

I'm guessing this was causing problems in your test? I've got some ideas around fixing that but I'll leave them on the test 😁

collect_app/src/test/java/org/odk/collect/android/activities/MainActivityTest.java
```diff
@@ -197,4 +200,18 @@ public void manageFilesButtonTest() throws Exception {
         assertEquals(FileManagerTabs.class.getName(),
                 shadowIntent.getIntentClass().getName());
     }
+
+    /**
```
seadowg: I had never seen this test file! Well done on finding it. I think it's good that you've followed the style set by the file but it's a little unusual compared to the tests we'd usually write. Generally we try and be as descriptive as possible in our test names rather than using JavaDoc. For example I would have named your test `pressingConfigureQRCode_launchesScanQRCodeActivity`. I like the test itself.

It also looks like test doesn't set up the application for the `MainMenuActivity` correctly - this is why you were getting a `null` `adminSharedPreferences`. The activity's `onCreate` has this code:

```java
try {
    new StorageInitializer().createODKDirs();
} catch (RuntimeException e) {
    createErrorDialog(e.getMessage(), EXIT);
    return;
}
```

For this to work the app's "storage state" has to be setup correctly. It wasn't so the `catch` branch was firing before `adminSharedPreferences` was initialized. I think it would nice to rework this whole test (which is something I can take a look) so for the moment could you create a new `MainMenuActivityTest` with your test (iwith a new name and without the javadoc comment). The `@Before` you need to make sure storage state is setup is:

```java
@Before
public void setup() {
    ShadowEnvironment.setExternalStorageState(MEDIA_MOUNTED);
    mainMenuActivity = Robolectric.setupActivity(MainMenuActivity.class);
}
```

collect_app/src/main/java/org/odk/collect/android/activities/ScanQRCodeActivity.java
```diff
@@ -0,0 +1,98 @@
...
+import org.odk.collect.android.utilities.PermissionUtils;
+import org.odk.collect.android.utilities.ToastUtils;
+
+import java.io.IOException;
+import java.util.zip.DataFormatException;
+
+import androidx.annotation.Nullable;
+import timber.log.Timber;
+
+// ScanQRCodeActivity uses the decorator pattern to process the results from ScannerWithFlashlightActivity
```
seadowg: I like the comment but the "decorator" part confused me. Generally I'd think of a decorator adding functionality to an interface. I could totally just be missing something!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/ScanQRCodeActivity.java
hypercubestart: I guess I was thinking of ScanQRCodeActivity as adding functionality on top of the ScannerWithFlashlightActivity and removing the need to interact with ScannerWithFlashlightActivity directly anymore. 

Would it be more appropriate as Proxy design pattern because ScanQRCodeActivity is wrapping the ScannerWithFlashlightActivity?

lognaturel: I don't think you need to explicitly call out a pattern. I'd change the comment to `Initiates barcode scanning and processes its results` or something like that.

hypercubestart (approved):

collect_app/src/test/java/org/odk/collect/android/activities/MainMenuActivityTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.robolectric.RuntimeEnvironment;
+import org.robolectric.fakes.RoboMenuItem;
+import org.robolectric.shadows.ShadowEnvironment;
+
+import static android.os.Environment.MEDIA_MOUNTED;
+import static junit.framework.Assert.assertEquals;
+import static org.robolectric.Shadows.shadowOf;
+
+/**
+ * Unit test for checking {@link ScanQRCodeActivity} launched correctly from {@link MainMenuActivity}
```
seadowg: We probably don't need this comment. We'll most likely add tests to it so it'll get out of date fast!

seadowg (dismissed):
Just one thing to change for me but it's looking good 👍

hypercubestart (approved):

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -445,61 +463,71 @@ public void onClick(DialogInterface dialog, int i) {
         alertDialog.show();
     }
 
-    @Override
-    protected Dialog onCreateDialog(int id) {
-        switch (id) {
-            case PASSWORD_DIALOG:
+    private Dialog createPasswordDialog(Intent intent) {
```
lognaturel: #3623 adds some nice abstractions for this admin dialog (https://github.com/opendatakit/collect/pull/3623/files#diff-587fa336330bb907832356ee77ed08d1R405). I wasn't sure when each of these two PRs would land and how the conflict would best be dealt with. As luck would have it, they're ready at the same time. Since it's going through quality assurance right now, let's plan on merging it first and then you'll unfortunately have conflicts to fix throughout this class.

lognaturel:
The analytics look good, @hypercubestart! The scoped storage PR has been merged so you're ready to bring in those changes here. Let us know in Slack if you have any questions about how to do that merge.

lognaturel (approved):
@hypercubestart and I chatted on Slack and agreed I'd fix merge conflicts because his push attempt was in an odd state. I verified the behavior and noticed that the call on `onAttach(Context)` wouldn't work for API < 23. We really should move off of the original Fragment library for settings to androidx Fragments to avoid these kinds of issues.

mmarciniak90:
**Tested with success**

Verified cases:
- Scan QR code button in the overflow menu
- Hidden option in the main menu 
- Denied camera permission
- Scan invalid QR code
- Scan random QR codes with invalid chars what then throw exceptions (pr/3422)
- Scan QR code with set dark mode, Spanish language, admin password
- Scan QR code with username and password to Aggregate and set User-defined detains in Form metadata
- Reset settings when Scan QR code button is hidden

lognaturel:
Thanks so much, @hypercubestart! 🎉

Issue #3538, opened 2019-12-14
Rework `FormDownloader`


seadowg:
We recently added tests to the `FormDownloader` (#3537) and doing so revealved some awkwardness in it's design. I think we should look at using the things we don't like about the tests to drive out a refactor here:

1. This object shoudn't require Robolectric to be tested. To remove it we'd most likely need to remove/invert it's dependencies on the `Collect.getInstance()` singlton.
2. We shouldn't need to mock a method (using `spy`) on the object under test. We might need to break out another object or be passing a mock/fake of one of the object's dependencies.

breakbusyloop:
I like to follow this rough pattern for refactoring code to expand test coverage:

1. Split out the obvious 'helper' snippets of code, especially code that is free of side effects but is adjacent to code with side effects. E.g. `downloadXform` needs helpers for `sanitizeFormName` (the regex), `generateFilenameFromFormName` (with file I/O, but still testable), and `proposeFilename` (helper to generateFilenameFromFormName, but without I/O).
2. Separate the decisions from the actions. Pull out fragments of code that examine variables to make a decision about what action to perform. Write tests for the decision logic. E.g. `downloadManifestAndMediaFiles` should be split to have the parsing logic isolated from the download action so we can test parsing support for a malformed manifest document.
3. Locate dependencies that can be passed in instead of accessed as global singletons or member variables representing complex objects. Hoist up those dependencies to make them the responsibility of the caller. E.g. `installEverything` could take some kind of simple wrapper Resolver class that handles the calls to `findExistingOrCreateNewUri` and `Collect.getInstance().getContentResolver().delete`. Under test we would call installEverything with a test double of Resolver that would not require access to singletons.
4. Relocate code to a separate file if it is not closely related to the rest of the logic. E.g. `downloadFile` is the only part of FormDownloader.java that operates on byte[], file streams and direct access to the HTTP input stream.

This process concentrates the integration points to a few methods. We would still need mock testing to handle HTTP requests, form DAO method calls, singleton access, etc. But it keeps our integration tests isolated to just the code that has integration points, while unit tests cover the rest of the code. Eventually we could refactor to inject all dependencies, using test doubles to provide implementations that don't access singletons or perform side effects. That final stage is much more work than the other stages and provides less value, so it is likely less important than continuing the steps above for other files in the code base.

breakbusyloop:
[...]

PR #3639, opened 2020-02-12
Fix error when downloading forms with references to external secondary instances including last-saved

### LINKED ISSUE #3635: Error when form with reference to last-saved instance is downloaded
Collect v1.25.1

#### Problem description
When a form with a reference to the last-saved instance is downloaded, the last-saved instance isn't created so there's a crash when Collect tries to build a FormDef to get the geo coordinate for mapping.

See [user report](https://forum.opendatakit.org/t/xlsform-spec-proposal-add-syntax-to-make-it-easy-to-use-a-value-from-the-last-saved-instance/18814/37?u=ln).

1. Create a form with a reference to the last saved instance such as 
[neighborhood-last-saved.xlsx](https://github.com/opendatakit/collect/files/4187385/neighborhood-last-saved.xlsx)
1. Upload it to a server
1. Download it from Collect

#### Expected behavior
Downloading form with last-saved works.

lognaturel:
Closes #3635

Reviewers: I recommend running and understanding the tests in the first commit first since there's a lot going on here.

I added and ran automated tests. I also configured a device with the following Central QR code to try forms with CSV attachments (but remember that https://github.com/opendatakit/collect/issues/3335 means it won't load more than once) and last-saved: 
central-app-user-external-instances

I tried a form with a XML attachment locally but both Aggregate and Central seem to currently choke on those.

The changes are to always create a temporary media directory whether or not a form definition comes with media, to always create a temporary last-saved.xml file in that directory, and to configure the global `ReferenceManager` to resolve `jr://file` and `jr://file-csv` URIs to that directory.

I tried to be as minimal and surgical as possible here because we're going to put this out as a point release. I also considered removing the temporary media directory concept because I don't see an advantage to it. It seems as easy to clean up files out of the final directory that they will belong in as it is from a temp directory. However, that felt too aggressive for now. Hopefully the test coverage will allow us to be more ambitious with refactors if this code needs to be modified again.

I also really had to hold myself back from trying to redesign how `last-saved` works. It doesn't make sense that the `last-saved.xml` file is stored with form media. Ideally, a whole new process would be introduced to handle virtual `jr://instance` endpoints. Again, though, now is not the time for that.

As described in https://github.com/opendatakit/collect/pull/3537, I don't love the use of a spy in the test but I think it's the best we can do now and it gives us further coverage for any future refactor we may want to do.

The intentional change is the fix to #3635.

One risky thing that I didn't know how to test is the resetting of the global `ReferenceManager`. The risk would be that under certain circumstances form media can't be resolved. I think resetting it is fine because the root session translators have to be configured each time a new form definition is loaded and I've verified that this is done in `FormLoaderTask` but it does make me a bit nervous.

Download of any form goes through this code path so special cases around form downloads should be considered. One I thought of and added a test for is a `last-saved.xml` attachment.

See server configured by QR code above and `FormDownloadTest`.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Certain functions in ODK Collect depend on cloud services that require API keys 
   - Follow [these instructions to generate a signing certificate fingerprint and register the application with the Google API Console](https://developers.google.com/drive/android/auth#generate_the_signing_certificate_fingerprint_and_register_your_application).
   - [Enable the Google Drive API](https://console.developers.google.com/apis/api/drive.googleapis.com).
   - [Enable the Google Sheets API](https://console.developers.google.com/apis/api/sheets.googleapis.com).
+  - Copy the oauth clientID from [the developer console](https://console.developers.google.com/apis/credentials) into the `client_id` of `collect_app/google-services.json` file
 
 **Google Maps API**: When the "Google Maps SDK" option is selected in the "User interface" settings, ODK Collect uses the Google Maps API for displaying maps in the geospatial widgets (GeoPoint, GeoTrace, and GeoShape).  To enable this API:
   - [Get a Google Maps API key](https://developers.google.com/maps/documentation/android-api/signup).  Note that this requires a credit card number, though the card will not be charged immediately; some free API usage is permitted.  You should carefully read the terms before providing a credit card number.
--- a/collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
@@ -163,7 +163,7 @@ public abstract class AbstractSelectListAdapter extends RecyclerView.Adapter<Abs
         String errorMsg = null;
         if (imageURI != null) {
             try {
-                final File imageFile = new File(ReferenceManager.instance().DeriveReference(imageURI).getLocalURI());
+                final File imageFile = new File(ReferenceManager.instance().deriveReference(imageURI).getLocalURI());
                 if (imageFile.exists()) {
                     Bitmap bitmap = FileUtils.getBitmap(imageFile.getPath(), new BitmapFactory.Options());
 
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
@@ -227,7 +227,7 @@ public class AudioVideoImageTextLabel extends RelativeLayout implements View.OnC
 
     private void openImage() {
         try {
-            File bigImage = new File(referenceManager.DeriveReference(bigImageURI).getLocalURI());
+            File bigImage = new File(referenceManager.deriveReference(bigImageURI).getLocalURI());
             Intent intent = new Intent("android.intent.action.VIEW");
             Uri uri =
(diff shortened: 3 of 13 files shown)


seadowg (approved):
As far as I can see this looks safe and fixes the issue and I think it's good to get this to QA. 

I'm not a fan at all of the `FormDownloader` (I know you're with me on this) and the tests having spies make it really hard to comprehend exactly what's going on and what is being tested. I think it's good to have tests around this stuff obivously but I'm worried if we keep going like this we're really just locking in the bad design and making harder to unpick. This part of the code has created regressions in a couple of releases now so I think we need to look at #3538 as soon as possible. Is there any chance you have capacity to look at that next, having just been in the code, or should we just make sure to prioritize for the next release?

lognaturel:
> I'm not a fan at all of the FormDownloader (I know you're with me on this)

Yes. 😣

Thanks, @seadowg. I've given more thought to what went wrong and wrote a post-mortem at https://github.com/opendatakit/collect/issues/3538#issuecomment-585929951. I'm interested to see whether you agree with some of the root causes I've identified.

kkrawczyk123:
@lognaturel I have verified downloading form attached to the issue from Aggregate, Central and Google Drive. It works fine for Aggregate and Central but not for Google Drive. I can't use that form when it is downloaded form drive, error message displays and form does not shows up on the fill blank form list. See the screenshots:
| form attached to issue | other forms with last-save |
| -------- | -------- |
Screenshot_20200214-085628 | Screenshot_2020-02-14-09-35-59

I have checked that on the master too, and I was able to reproduce the issue when downloading the form from Aggregate but on master - downloading from Drive works well. It might be a regression. 

I am continuing my testing.

kkrawczyk123:
@lognaturel I have used attached Central QR code to import settings:

I downloaded Neighbourhood last saved form, I was able to fill it, edit saved form, fill another submission and also able to send a filled submission. I was able to upload it to Aggregate too and I have obtained the same results as for Central. 

I was able to download Forms with external CSV files from Central too. I filled just one submission. There is no way to fill another or edit already filled one, a screenshot of the displayed error:
Screenshot_20200214-100113
 Sending one filled form works well. I wasn't able to upload it to Aggregate 2.0 and do the same checks.

lognaturel:
Thank you, @kkrawczyk123. You'd think that I would have checked to make sure that the push-to-device case still worked but I didn't. The Google Drive integration uses that same code path. I don't even know whether to laugh or cry. At least it was a really quick fix once I figured out how to get Google Drive configured! Please make sure to also check push to device as you do your last verification.

I don't know how to put any kinds of test around that. In a separate PR, I'm going to see whether I can reduce some of the duplication between `DiskSyncTask` and `FormDownloader`. As described in https://github.com/opendatakit/collect/issues/3538#issuecomment-585929951, that redundancy is a significant problem.

The CSV behavior on second load is expected because of #3335.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 4.4, 5.1, 6.0, 7.1, 8.1, 9.0 and 10.

Verified cases:
- download form with last save, form with last save and media from Aggregate
- fill form, send form
- download form with last save, form with last save and media from Google Drive
- fill form, send form
-  download form with last save, form with last save and media from Central
- fill form, send form
- download form with external secondary instances (from attached central qr code) from Central
- fill form, send form
- download form with external secondary instances (from attached central qr code) from Google Drive
- fill form, unable to send form
- unable to add it to Aggregate, so I wasn't able to verify that case with Aggregate 
- downloaded and filled other form with last save
- downloaded and filled encrypted form with last save
- downloaded and filled other forms with external secondary instances 

PR #3640, opened 2020-02-14
Add copy on long press to device details in Form Metdata

### LINKED ISSUE #3633: Allow copying of Install ID
This is a follow up to #3595.

As @lognaturel suggests in the original feature:

> It should be possible to copy this ID to the device clipboard.

This was good to split out as copying preferences to the clipboard means we need to switch to using the new androidx.preferences dependencies rather than the deperecated one's we are using now. It would be good to limit this to just the Form Metadata screen if possible so then we can covert the other screens later which breaks the risk into chunks.

## Acceptance

Given I've navigated to the "Form Metadata" screen
When I long press on "Install ID"
And click "Copy"
Then I see a toast saying "\<installID\> copied to clipboard"
And I can paste the Install ID

### LINKED ISSUE #3646: Install ID should not be reset when resetting settings from admin settings
#3632 adds an install ID that will soon replace IMEI which we can no longer read after targeting SDK 29. That install ID is stored in shared preferences and is currently reset when a user chooses to reset settings from the Admin Settings inside of Collect. The goal for the setting reset is that it reset settings that are user-facing. The install ID is more system-facing and we only want it reset if the app is reinstalled or the user clears settings from Android.

There's a `KEYS_WE_SHOULD_NOT_RESET` collection that exists for this reason. The install ID key should be added to it.

seadowg:
Closes #3633
Closes #3646

This makes it possible to copy any of the device defined metadata values on the "Form Metadata" screen by long pressinging and then hitting copy. 

There are a couple of code changes that went in as well:
* Added `androidx.test` runner and `androidx.test.fragment` testing utils dependencies. These let us use `FragmentScenario` to write tests for individual `Fragment` objects. Behind the scenes this uses Robolectric.
* Moved `SharedPreferencesTest` from `androidTest` to `test`
* Deprecate `GeneralSharedPreferences.getInstance()` and `AdminSharedPreferences.getInstance()`. Static singletons like this cause problems in tests as the same instance is retained between tests. The singleton is now managed in Dagger (which has a dependency module attached to the lifecycle of the `Application`, not the process).
 
Getting this to work meant adding in the new `androix.preference` framework. This is replacing the deprecated `android.preference` framework which didn't include the copy feature. This definitely felt like the right solution vs hacking the feature in ourselves but it does mean we needed to create a new activity (`FormMetadataPreferencesActivity`) that works with this framework. I think we can slowly move to using this for every preference screen - probably just by updating them as we touch them.

Only thing I'd be worried about is regressions on the Form Metadata screen itself so good to have a play with that.


collect_app/src/main/AndroidManifest.xml
```diff
@@ -285,5 +285,6 @@ the specific language governing permissions and limitations under the License.
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
+        <activity android:name=".preferences.FormMetadataPreferencesActivity" />
```
lognaturel: This should be above the `External API` comment. I'd tend to group it with others in the preferences package.

collect_app/src/main/res/values/strings.xml
```diff
@@ -484,8 +484,8 @@
     <string name="form_metadata_summary">These values will be added to forms that have username, email and/or telephone preloaded fields to identify the person submitting data.</string>
     <string name="phone_number">Phone number</string>
     <string name="email">Email address</string>
-    <string name="user_defined">User-defined</string>
-    <string name="device_defined">Device-defined</string>
+    <string name="user_defined">User defined</string>
```
lognaturel: Does the hyphen seem wrong to you? These are compound words so the hyphen is required at least by [APA guidelines](https://apastyle.apa.org/learn/faqs/when-use-hyphen). Maybe it's a regional thing?

Changing these strings means they'll be pushed out to translators. If translators don't re-translate them before the release then they'll show up in English instead.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -166,7 +165,7 @@ public ActivityAvailability providesActivityAvailability(Context context) {
 
     @Provides
     InstallIDProvider providesInstallIDProvider(Context context) {
-        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences preferences = context.getSharedPreferences("meta", Context.MODE_PRIVATE);
```
lognaturel: I think that this is good but I'm concerned that it will get lost and forgotten. How about marking `KEYS_WE_SHOULD_NOT_RESET` as deprecated and describing this strategy? Maybe `meta` should be a constant somewhere?

lognaturel (commented):
Couple of small things that I think should be addressed. Please mark for QA after!

lognaturel (changes requested):
Should have marked as `request changes` even if it's small things.

collect_app/src/main/AndroidManifest.xml
seadowg: Ah sorry yes this was just auto generated. But good catch as that's how these files get so messy!

collect_app/src/main/res/values/strings.xml
seadowg: Hmmm yeah it feels wrong to me as a title (rather than in a sentence). Are we using that as a language guide for English? Would be good to reference that.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
seadowg: Yeah I was in two minds about this as I'm not a fan of the existing shared prefs abstractions and the keys classes. I do agree that it would be nice to have something that makes the `meta` prefs more discoverable though so they get used for other things. I'll have a think.

seadowg: As a note I'd like to keep the test with a hardcoded/inline as this prevents someone changing the constant (in the code) and the test still passing. It's checking configuration as well as behaviour.

seadowg: I'm going to go with a provider for the `meta` prefs for now. I didn't want to start building a whole `MetaSharedPrefences` to mimic the other ones. I like that we have abstraction for prefs storage though so I might follow up with a refactor that pulls out the shared logic around shared prefs and maybe improves the interfaces a little.

mmarciniak90:
@seadowg I noticed unexpected behavior. 

1. Click on `Username` or `Phone number` or `Email address` to open dialog
2. Close the dialog
3. `SIM serial number` or `Subscriber ID` become inactive

|master | first run on PR| after first opening some dialog | after the second opening dialog|
|---|----|----|---|
|Screenshot_20200221-122544|Screenshot_20200221-123134|Screenshot_20200221-123142|Screenshot_20200221-123146|

It is not reproducible in 100% but it is visible very often. It also depends on the device. For me, reproducing this case on Android 7 is the easiest.

Rotating or reopening this page restore default appearance.

Is it possible that these changes cause an impact on changing the language on Android 7?
I cannot change it to e.g. Spanish but I can't reproduce it on the master branch. 

mmarciniak90:

seadowg:
@mmarciniak90 hmmm. Yeah it could be that the new Preferences framework causes problems with language changes as the way we do that is not "officially" supported. Probably good to investigate that now. I'll have a look on Monday!

seadowg:
@mmarciniak90 I've fixed the langauge change problem. It was related to a broken Android dependency that got pulled in again (I've added an issue at #3649). 

I'm not sure I can work out a fix for the subscriber id/sim serial changing styles. It seems to be a wierd piece of behaviour in the Android preferences library that occurs only when the preference value is blank (in this case because the phone doesn't have a sim inserted). If you do think it'll case problems for people we could have those fields display "No sim inserted" or "Not set" when there isn't a value to display. An alternative to that would just be hiding the fields when we can't get them.

mmarciniak90:
@seadowg I completely agree! Displaying this information doesn't affect the user experience but allows us to avoid unexpected behavior.
I really like this idea but the final decision belongs to @lognaturel

lognaturel:
Using `Not set` wherever a value is unavailable sounds great to me. I think there are a bunch of reasons we might not be able to get the values (permissions, hardware, etc) so I wouldn't want to get too specific. I think it's nice to still display the fields so users know what values could potentially be available.

seadowg:
@lognaturel sounds good to me. I'll stick that in this PR

seadowg:
@lognaturel that's fixed now. I've also made it so only the "Install ID" has the copy on long press behaviour. Initially I was a little unsure how best to deal with the fact that people might end up copying "Not set" somewhere but then I realized we probably want to discourage folks using those other IDs anyway so making "Install ID" the easiest to use seemed like a nice move.

lognaturel:
That makes sense. Sorry, just thought of something. I was thinking that "Not set" came from somewhere Android but it doesn't look like it. I'm wondering whether "Not available" would actually make more sense?

seadowg:
The standard behaviour (which uses SimpleSummaryProvider) will display "No set" but be a use these preferences are just displayed and not actually stored we're not using that behaviour.

I do like "Not available" a whole lot more though so I'll change it!

lognaturel (approved):

seadowg:
@lognaturel ok that's changed now!

mmarciniak90:
@seadowg I noticed a small issue. I compared PR with the master branch and I noticed that keyboard with alphabet appears instead of a numeric keyboard when I click on `Phone number`

| master| PR|
|---|---|
||Screenshot_2020-02-25-08-49-20-963_org odk collect android|

The second thing that I want to present is a comparison of the PRs appearance. Adding subtext caused the main text is grey out in the Device-defined part. I just wanted to mention this visual change.

|master| PR|
|---|---|
|Screenshot_20200225-085108|Screenshot_20200225-085105|

seadowg:
@mmarciniak90 huh interesting on the keyboard - must be a change about how we specificy it's a "number" preference. I'll look at that.

On they grey text, yeah it seem to be the stye that non editable/selectable are disabled in the new preference framework. We might be able to look at changing that in the future but I think we can live with it for this screen.

seadowg:
@mmarciniak90 fixed! Looks like the new preferences framework had made that less easy to do. Phone number will be using a numeric keyboard now.

mmarciniak90:
**Tested with success**

Verified on devices with and without a SIM card.

Verified cases:
- Copy Install ID
- Toast
- Reset
- Light/ dark mode
- English, Spanish, RTL language
- Rotating device
- Numeric keyboard for `Phone number`

PR #3642, opened 2020-02-18
Replace Espresso tests for default image with Widget tests

seadowg:
Follow up to #3579

This adds tests for widgets that do and don't support default images. These tests are at the individual widget level and allow us to remove the Espresso test that was covering this feature previously. We'd initially discussed also adding tests for ther non default image flow but I think this is enough to merge for now (I've create an issue for the extra work at #3643).

To do something like this generally the process I follow is:
1. Run the Espresso test I think we can get rid of - verify it's green
2. Comment out code for feature until the Espresso test is red
3. Write tests that drive out the feature at a different level - verify they're red
4. Make the new tests go green
5. Verify the Espresso test is also now green - delete it

We could of course keep the tests at an Espresso level instead but I feel like a widget/Robolectric test is better as it allows more granularity around the assertions (the image loaded for instance) and also provides faster and more "connected" feedback - it's easier to know what tests to run when working on a specific widget.

No changes to non-test code so probably good to merge this on acceptance.


config/checkstyle.xml
```diff
@@ -186,7 +186,7 @@
             <property name="allowedAbbreviationLength" value="1"/>
         </module> -->
         <module name="OverloadMethodsDeclarationOrder" />
-        <module name="VariableDeclarationUsageDistance" />
+        <!-- <module name="VariableDeclarationUsageDistance" /> -->
```
seadowg, line 189: This is potentially contraversial so happy to reenable the check and fix the problems if people disagree with me here. 

The reason for disabling the check is that I think it can make writing tests with setup data hard. For instance:

```java
String testData1 = // setup test data
String testData2 = // setup test data
String testData3 = // setup tests data

testObject.doAThing(testData1);
testObject.doAThing(testData2);
testObject.doSomethingElse();

assertThat(testObject.getSomething(), notNullValue());
assertThat(testObject.getSomething().anotherThing(), equalTo(testData3));
```

This is a pretty common pattern for a test - we're setting up our inputs and outputs, carrying out the steps we're testing and then making an assertion on something we set up at the beginning of the test. This can make tests more readable and also helps if you need to mutate input/output values. This would fail our `checkstyle` build however as the `testData3` declaration would be too far away from it's use.

lognaturel: I agree that this split is reasonable for tests. I do't feel strongly one way or the other but I will say that I have had that checkstyle rule fail on code I was writing and was glad for it. Maybe it's something that would have been flagged in review anyway.

lognaturel (approved):
Looks great to me, thanks.

@grzesiek2010 Please take a look and do highlight any questions or concerns you may have. The difference in tests is somewhat subtle. The key thing is that by removing `doesSupportDefaultValues` and `getFile` from the tests and checking the loaded image, we now have a test that's closer to what the user actually experiences and we're free to refactor some of the implementation details.

lognaturel:
`BaseImageWidget`'s `getFile` and `getDefaultFilePath` methods should now be private.

seadowg:
@lognaturel good catch! I'll fix that.

lognaturel (approved):

PR #3645, opened 2020-02-18
Define new Widget "framework"

seadowg:
The core motivation here was to make it easier (and encouraged) to define a widget's 
"answer" view as declaratively as possible. Previously widgets would build this view procedurally (and often entirely in code rather than XML) using the `addAnswerView` method. This makes it hard to adjust styles as there isn't a simple way to preview the widget. It also just makes the widget harder to understand as generally reading XML layouts (especially layout constraints etc) is easier than view setup code.

I've added new docs around the pattern I think we should be adopting and used it in `TriggerWidget` (and it's tests) as an example.

This does make one visual change which is to use the `Headline6` text size for answer text. This is slightly smaller than the text was before. The thinking behind this was that the widget felt strange to look at given the question and answer texts were different sizes. If we think this new size is too small I'd suggest we bump them both up.

Ran tests and also manually verified that the `TriggerWidget` hadn't broken in anyway.

The team has had a bunch of discussions around this but I'm definitely interested in feedback around how this new approach feels to folks.

Shouldn't be any regression risks other than to `TriggerWidget`.


collect_app/src/test/java/org/odk/collect/android/widgets/TriggerWidgetTest.java
```diff
@@ -1,67 +1,98 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.nullValue;
+import static org.mockito.Mockito.verify;
+import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.mockValueChangedListener;
+import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.promptWithAnswer;
+import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.promptWithReadOnly;
+import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.widgetTestActivity;
+
+@RunWith(RobolectricTestRunner.class)
+public class TriggerWidgetTest {
```
seadowg, line 24: I've chosen to not extend the `QuestionWidgetTest`. From looking at a lot of our tests, although this pattern initially makes sense to me, it seems we run into problems with widgets having different behaviours for some of the common methods. I thought trying to build out helpers `QuestionWidgetHelpers` would be a more flexible approach - make the tests easy to write rather than making the tests hard to change.

Additionally it feels like the hierarchical nature of the widgets has only really caused problems (unpicking `StringWidget` common code for example) and I think using test extension here feeds into that problem as well.

Interested to see what others think of this however!

collect_app/src/test/java/org/odk/collect/android/support/MockFormEntryPromptBuilder.java
```diff
@@ -91,4 +91,11 @@ public MockFormEntryPromptBuilder withAnswerDisplayText(String text) {
 
         return this;
     }
+
+    public MockFormEntryPromptBuilder withAnswer(IAnswerData answer) {
+        when(prompt.getAnswerValue()).thenReturn(answer);
+        when(prompt.getAnswerText()).thenCallRealMethod();
```
lognaturel, line 98: Huh, I didn't know this was possible. I guess this makes the resulting form entry prompt more of a spy.

collect_app/src/test/java/org/odk/collect/android/widgets/TriggerWidgetTest.java
lognaturel: On one hand it's too bad that we'll end up with a bunch of trivial duplicated tests. But on the other they're really easy to write and there is value to having each widget considered independently so that they can more easily be refactored to suit their specific needs as you say. We just have to be careful to make sure all the same edge cases are considered.

collect_app/src/test/java/org/odk/collect/android/widgets/TriggerWidgetTest.java
```diff
@@ -1,67 +1,98 @@
...
+    }
+
+    @Test
+    public void checkingCheckbox_setsAnswer() {
+        TriggerWidget widget = createWidget(promptWithAnswer(null));
+        CheckBox triggerButton = widget.getCheckBox();
 
         triggerButton.setChecked(true);
-        triggerButton.callOnClick();
+        assertThat(widget.getAnswer().getDisplayText(), equalTo("OK"));
```
lognaturel, line 79: This makes me so angry. Why "OK"?! Everything about this trigger widget is objectionable starting from the fact that `trigger` is a [well-defined concept in W3C XForms](https://en.wikibooks.org/wiki/XForms/Trigger) and that this is something else entirely.

No action to take, just venting. 😾

lognaturel (approved):

collect_app/src/test/java/org/odk/collect/android/support/MockFormEntryPromptBuilder.java
seadowg: Yeah I'm not a fan of this. Ideally this would just setup a real `FormEntryPrompt` but it's really hard to build one. The `getAnswerText` method definitely feels like a "decorator" that shouldn't really be on the `FormEntryPrompt` as well.

collect_app/src/test/java/org/odk/collect/android/widgets/TriggerWidgetTest.java
seadowg: 🔥

seadowg:
@lognaturel on the duplicated test thing: Yeah I'm super tempted to turn this into a "contract" style testing world but I honestly feel like from the way it's gone in the current suite that just writing as many helpers/tools for writing these tests is the way to go.

mmarciniak90:
> This does make one visual change which is to use the Headline6 text size for answer text. This is slightly smaller than the text was before

I noticed on all devices change mentioned by @seadowg.

I didn't notice regression on `TriggerWidget`.
I'm not sure what else I could test. If you have any suggestions, please let me know.

Verified cases:
- TriggerWidget 
- light/dark mode
- English/Spanish language
- RTL language
- minimize
- rotation
- hierarchy view
- new/saved form


|master|Screenshot_20200304-125202|Screenshot_20200304-132433|
|---|---|---|
|**PR**|Screenshot_20200304-125536|Screenshot_20200304-132319|

seadowg:
@mmarciniak90 no that's all good!

mmarciniak90:

PR #3652, opened 2020-02-25
Fix the form map crash by waiting until the map fragment is initialized to save it in the map field

### LINKED ISSUE #3651: Collect crashes when the map button is clicked if at least one point is set
Collect master

#### Problem description
Collect crashes when the map button is clicked if at least one point is set

1. User fills `All widgets` form 
2. User adds a location to Geopoint question 
3. User leaves jobs (completed or not)
4. User goes to `Fill Blank Form` and clicks on the map button 
5. Collect crashes

#### Expected behavior
The map view is opened and added point is visible

lognaturel:
Closes #3651 

In e3af24b, we let tests set their own map fragment value through the `map` field in `FormMapActivity`. We also immediately set the `map` field for a map created by the real activity which was a change in behavior. Following that change, when `onResume` called `updateInstanceGeometry`, the map wasn't `null` so there was an attempt to map features even though the map fragment was not guaranteed to be fully initialized. We need the geometry update in `onResume` for the case where a user triggers form filling from this activity and then returns to it having added a new feature.

It's not clear to me how this wasn't detected before we merged #3552. Maybe it didn't happen with every map provider or device?

Verified that all tests still pass (though they weren't helpful in this case) and manually verified opening a form map with mappable instances and Google maps and OSM fragments. I also tried geopoint and geotrace with Google Maps and OSM basemaps.

This makes `MapProvider` into a Dagger-injected singleton. This means that tests can mock `MapProvider` and provide a specific `MapFragment` implementation. This removes the risk of this particular problem resurfacing because there is never a case where the `map` field is set before initialization. `MapProvider` still feels a little awkward since it has several static methods that are involved in setting up map configuration preferences but I don't think it's so bad. I don't think it needs to be a singleton because I believe that the (Java) maps it keeps only are used during the existence of a single map fragment but I decided to stay close to the original design for now.

The biggest code change is in the `FormMapActivity` constructor. This now touches the way that any map interface is initialized so there is a regression risk to every question type that involves a map in addition to the form map.

Any form with a geopoint like All Widgets.

No.


collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -113,13 +112,14 @@
         TextView titleView = findViewById(R.id.form_title);
         titleView.setText(viewModel.getFormTitle());
 
-        if (map == null) { // tests set their maps directly
-            Context context = getApplicationContext();
-            map = MapProvider.createMapFragment(context);
+        MapFragment mapToAdd = map; // tests will have a value for the map field
+
+        if (mapToAdd == null) {
+            mapToAdd = MapProvider.createMapFragment(getApplicationContext());
```
seadowg: Yeah I think we should use Dagger for this. It feels like we're adding machinery in to mimic what it would do for us here.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: Sigh, ok, will do. 😔

lognaturel: I should have referred back to https://github.com/opendatakit/collect/pull/3552#discussion_r360364242 before mentioning Dagger. We can't inject the `MapFragment` in `onCreate` for exactly the thing that this bug is related to -- we don't want to set the field until the asynchronous initialization has completed. Alternately, we could keep extra state to know whether the initialization has completed or not but I don't think that's very clean. So I went the route of injecting a `MapProvider` singleton. I hope you will find this reasonable, @seadowg. I had chosen not do that in the previous PR because it would touch every geo component but I suppose it's not particularly risky.

lognaturel: For the record and so that perhaps someone can push back on me, some things that upset me about Dagger/our Dagger setup:
- the magic. We end up with a lot of generated code that feels brittle.
- the mental overhead. I thought I could inject into `BaseGeoMapActivity` not realizing that I couldn't because it's abstract. I had to track down that I was getting a no-op inject method and find my way to https://github.com/google/dagger/issues/489 since there's no warning. It might just be a familiarity thing but I feel like I end up with some kind of irritation like this every time I'm using Dagger.
- we wax poetic about separation of concerns and then we have `AppDependencyModule` and `AppDependencyComponent` which feel like really ugly grab-bags of stuff. Part of that is that the app itself isn't currently divided up into modules.

collect_app/src/main/java/org/odk/collect/android/geo/MapProvider.java
```diff
@@ -31,13 +33,12 @@
 import static org.odk.collect.android.preferences.GeneralKeys.KEY_USGS_MAP_STYLE;
 
 /**
- * A static class that obtains a MapFragment according to the user's preferences.
+ * Obtains a MapFragment according to the user's preferences.
  * This is the top-level class that should be used by the rest of the application.
  * The available options on the Maps preferences screen are also defined here.
  */
+@Singleton
```
seadowg: You don't need the `@Singleton` here as you have it on the provider. It can be useful as documentation but I'm not sure the class needs to know it's going to be used as a singleton in this case.

docs/CODE-GUIDELINES.md
```diff
@@ -78,10 +78,10 @@ While it's important to read the Dagger [documentation](https://google.github.io
 
 ### Providing dependencies
 
-To declare a new dependency that objects can inject add a `@Provider` method to the `AppDepedencyModule`:
+To declare a new dependency that objects can inject add a `@Provides` method to the `AppDependencyModule`:
```
seadowg, line 81: Woh good catch!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/geo/MapProvider.java
lognaturel: I went back and forth on this and it seemed helpful to identify that the map fields are an ongoing concern. How do you decide whether it’s useful or not?

seadowg: Hmmm. I guess for me I like that all the configuration around how a dependency is provided lives in the Dagger files (`AppDependencyModule` for instance) and thatt really the object itelf shouldn't need to know about it's own lifecycle. 

A more concrete criticism would be that having `@Singleton` in both places is confusing for newbies to Dagger (but then what about Dagger isn't confusing).

lognaturel: > thatt really the object itelf shouldn't need to know about it's own lifecycle.

I see what you're getting at. I feel like singletons often have a pretty different structure and the assumption that a single instance exists is baked in but perhaps that shouldn't be the case. Here, like I wrote in the PR description, it could pretty easily be restructured to not be a singleton. Anyway, I don't feel strongly about it so I've removed it.

> having @Singleton in both places is confusing for newbies to Dagger

I do agree with that.

seadowg (approved):

mmarciniak90:
@lognaturel I'm still able to reproduce a crash when I'm using Mapbox.

03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err: java.io.FileNotFoundException: /jacoco.exec: open failed: EROFS (Read-only file system)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at libcore.io.IoBridge.open(IoBridge.java:452)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at java.io.FileOutputStream.<init>(FileOutputStream.java:87)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.jacoco.agent.rt.internal_8ff85ea.output.FileOutput.openFile(FileOutput.java:67)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.jacoco.agent.rt.internal_8ff85ea.output.FileOutput.startup(FileOutput.java:49)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.jacoco.agent.rt.internal_8ff85ea.Agent.startup(Agent.java:122)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.jacoco.agent.rt.internal_8ff85ea.Agent.getInstance(Agent.java:50)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.jacoco.agent.rt.internal_8ff85ea.Offline.<clinit>(Offline.java:31)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.odk.collect.android.application.Collect.$jacocoInit(Collect.java)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at org.odk.collect.android.application.Collect.<init>(Collect.java)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at java.lang.Class.newInstance(Native Method)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at android.app.Instrumentation.newApplication(Instrumentation.java:997)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at android.app.Instrumentation.newApplication(Instrumentation.java:982)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at android.app.LoadedApk.makeApplication(LoadedApk.java:577)
03-03 10:47:05.973 28562-28562/org.odk.collect.android W/System.err:     at android.app.ActivityThread.handleBindApplication(ActivityThread.java:4779)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at android.app.ActivityThread.access$1600(ActivityThread.java:154)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1452)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:102)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at android.os.Looper.loop(Looper.java:234)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:5526)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err: Caused by: android.system.ErrnoException: open failed: EROFS (Read-only file system)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at libcore.io.Posix.open(Native Method)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at libcore.io.BlockGuardOs.open(BlockGuardOs.java:186)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err:     at libcore.io.IoBridge.open(IoBridge.java:438)
03-03 10:47:05.974 28562-28562/org.odk.collect.android W/System.err: 	... 21 more

lognaturel:
@mmarciniak90 Did the app actually crash or is this something you saw in the logs? The stack trace looks related to `jacoco` which is the library we use to compute code test coverage. It looks like a crash that can safely be ignored.

I've successfully added points on the Mapbox street style. I tried adding a reference layer and that also seems to work fine. I'm on Android 5.1.

mmarciniak90:
@lognaturel 
I check again with last changes `Merge remote-tracking branch 'upstream/master' into issue-3651` and I'm still able to reproduce the crash. I attached a video showing it on Android 5.1. Sorry for the bad quality of this video.

lognaturel:
Thanks! Could you please look to see if there’s another stack trace? Maybe @grzesiek2010 could help identify the meaningful one? Also knowing whether you can reproduce across devices would be helpful or anything else you can tell me about the state of instances at the time of crash.

grzesiek2010:
There is no stacktrace unfortunately

grzesiek2010:
Or maybe this:
`03-04 17:34:22.927 1003-1237/? E/NativeCrashListener: Exception dealing with report
    android.system.ErrnoException: read failed: EAGAIN (Try again)
        at libcore.io.Posix.readBytes(Native Method)
        at libcore.io.Posix.read(Posix.java:147)
        at libcore.io.BlockGuardOs.read(BlockGuardOs.java:230)
        at android.system.Os.read(Os.java:364)
        at com.android.server.am.NativeCrashListener.consumeNativeCrashData(NativeCrashListener.java:240)
        at com.android.server.am.NativeCrashListener.run(NativeCrashListener.java:138)`

Strange, I was able to reproduce it like 5 times in a row the crash took place just after opening the map and now I have to open and wait about 20s.

grzesiek2010:
Yes it might be the cause take a look here https://github.com/mapbox/mapbox-gl-native-android/issues/123

grzesiek2010:
Seems like updating mapbox to 9.0.0 fixes the issue. It doesn't mean the newest version doesn't contain 10 other bugs of course so be careful :)

lognaturel:
Thanks so much, @grzesiek2010. That gives us some good leads to work from.

lognaturel:
I've done more experimentation with the two devices I have and I really can't reproduce. That means I won't be able to confirm a fix.

I think it makes sense for QA to finish verifying this one and to merge it if no other defects are found. This crash can then be treated as a separate issue. Since @grzesiek2010 has a device he can reproduce on, he's in the best position to take it. I suggest first trying to upgrade the mapbox library to 8.6.2 since others have reported that the point release fixed the issue for them.

mmarciniak90:
**Tested with success**

Verified on Android: 5.1, 6.0, 7.0, 8.1, 10.0
Crash on Mapbox will be fixed in a separate issue.

Verified cases:
- saved, finalized, failed point statuses
- Google Maps, all map styles
- OSM
- offline layers

PR #3655, opened 2020-02-26
fix: Remove analytics for null form controller #3621

### LINKED ISSUE #3621: Remove analytics for null form controller
The `getFormController` methods in `FormEntryActivity` currently log analytics when the `formController` is `null`. However, the flow through that activity relies on the `formController` being null before the form is loaded. Those analytics have not been helpful and should be removed.

shubhamkumar1739:
Closes #3621 

I removed the analytics logging for null form controller.
This was recommended in the issue
Since, logging is removed from null form controller,  the data won't be logged in now.
No
No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -475,7 +475,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             formLoaderTask = (FormLoaderTask) data;
         } else if (data == null) {
             if (!newForm) {
-                if (getFormController(true) != null) {
+                if (getFormController() != null) {
                     FormController formController = getFormController();
                     identityPromptViewModel.setAuditEventLogger(formController.getAuditEventLogger());
                     formSaveViewModel.setFormController(formController);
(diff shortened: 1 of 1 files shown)


seadowg (approved):
This looks good to me and nice catch to remove the overloaded `getFormController`. @lognaturel do we also want to remove the logging `FormHierarchyActivity` or is that something you're interested in keeping?

lognaturel:
Hi and welcome, @shubhamkumar1739. Thanks for taking this on.

Please be sure to take a moment to read [the contribution guide](https://github.com/opendatakit/collect/blob/master/docs/CONTRIBUTING.md) to learn about claiming issues. In general, if someone has expressed interest in an issue it is best to look for another one.

> do we also want to remove the logging FormHierarchyActivity or is that something you're interested in keeping

Our current explanation for that case is that someone would have left their phone on the hierarchy screen and that the app would have gone out of memory. It happens at a pretty high frequency, though, which doesn't feel right. Because all three null form controller cases are logged together, it's hard to pull any information out of the noise caused by the two removed here. So I'd like to keep the hierarchy screen one for at least another release and see if we can get any more useful information when it's logged alone.

shubhamkumar1739:
Ok thanks. @lognaturel I will keep this in mind.

PR #3656, opened 2020-02-27
Add button for creating inline repeat

### LINKED ISSUE #3644: Enumerator can add another repeated question while in one
This feature was driven out by research [here](https://github.com/seadowg/collect-design/blob/master/research/repeats.md) and is mocked up [here](https://github.com/seadowg/collect-design/tree/master/prototypes/repeat-tweaks).

### Acceptance
- [x] Given I'm filling out `birds.xml`
And I'm already adding an observation
When I click the ➕ icon in the top right
And I click "Add"
Then I can fill out another observation

- [x] Given I'm filling out `birds.xml`
And I'm already adding an observation
When I click the ➕ icon in the top right
And I click "Do not add"
Then I return to where I was

- [x] Given I'm filling out `birds.xml`
And I've already added 2 observations
And I'm in the first observation
When I click the ➕ icon in the top right
Then I'm filling out a third observation

- [x] Given I'm filling out `birds.xml`
And I'm already adding an observation
When I click the ➕ icon in the top right
And click "Add"
Then an analytics event is sent ("AddRepeat", "Inline")

- [x] Given I'm filling out `birds.xml`
And I'm already adding an observation
When I click the ➕ icon in the top right
And click "Do not add"
Then an analytics event is sent ("AddRepeat", "InclineDecline")

- [x] Given I'm filling out `birds.xml`
And I'm already adding an observation
When I swipe top the end of the observation
And I click "Add"
Then an analytics event is sent ("AddRepeat", "Prompt")

- [x] Given I'm filling out `birds.xml`
When I go to the observations hierarchy view
When I click the ➕ icon in the top right
Then an analytics event is sent ("AddRepeat", "Hierarchy")

seadowg:
Closes #3644 

Other than implenenting the feature as described this PR:

* Moves code that "wraps" JavaRosa into a `javarosawrapper` package so it's called out
* Pulls the options menu logic partially out into a `FormEntryMenuDelegate` (using [delegate](https://en.wikipedia.org/wiki/Delegation_pattern) pattern) that can be iterated on and tested seperately from the `FormEntryActivity`
* Extract an interface (`MenuDelegate`) so the above pattern can be used consistently elsewhere
* Adds a component (`FormIndedAnimationHandler`) for determing whether how we should animate to the "new" screen of a form

Follow ups I think we should do (unless others think they should go in this PR):

* Move any remaining menu logic in `onOptionsItemSelected` into `FormEntryMenuDelegate`
* Use the `FormIndexAnimationHandler` for all the animations in `FormEntryActivity`
* Move newly added `FormController` methods and `FormIndexUtils` to JavaRosa itself
* Extract a view holder object from `FormEntryActivity` that allows it not have to contain the `showView` logic

Played around with a bunch of differnet repeat froms and also added feature tests for add repeats in the context of this feature and generally.

Lots that could be discussed here. I think comments on the list of "other" changes would be good if people are uncomfortable with them.

Should just add the feature but good to tests repeat forms to make sure there are no regressions. It's a good opportunity to try and tighten up our confidence in our automated tests of this feature.

Any repeat forms are good. Nested repeats and other weirdness would be great to play with.


lognaturel:
Could you please update to get in sync with `master`? This is currently based off of a revision that had a weird forms directory so it's harder to verify.

On my Android 5.1 device, the button does not show up when I first enter a repeat if the repeat is not the first question. If I tap the overflow menu or rotate the screen then it does appear. Similarly, if I exit a repeat, the add button doesn't disappear until I force the activity to refresh. I'm surprised the Robolectric tests don't catch that.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public void onCreate(MenuInflater menuInflater, Menu menu) {
+        menuInflater.inflate(R.menu.form_menu, menu);
+    }
+
+    public void onPrepare(Menu menu) {
+        FormController formController = formControllerProvider.getFormController();
+
+        boolean useability;
```
lognaturel, line 53: This is an exceptionally bad name. I guess we just leave it for now since it's not really related to the current changes?

collect_app/src/androidTest/java/org/odk/collect/android/FormNavigationTestCase.java
```diff
@@ -22,7 +22,7 @@
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import org.odk.collect.android.listeners.FormLoaderListener;
-import org.odk.collect.android.logic.FormController;
+import org.odk.collect.android.formentry.javarosawrapper.FormController;
```
lognaturel: I'm ok with a `javarosawrapper` package but I'm not sure it belongs in `formentry` since `FormController` is also used for form discovery. Maybe the package should be top-level? Or maybe it's fine in `formentry` since that's the bulk of what it's for?

@grzesiek2010?

collect_app/src/test/java/org/odk/collect/android/formentry/javarosawrapper/FormControllerTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.javarosa.xform.util.XFormUtils;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class FormControllerTest {
```
lognaturel: It'd be great to get a comment here saying that the implementation and tests really should move to JavaRosa.

I really don't like that these tests are based on the `toString`. Maybe we can discuss alternatives when we make the move. I'm thinking we can add test-only functionality to build a `FormIndex` from a string representation (similar to `Scenario.getRef`) and do the comparison with that.

collect_app/src/main/java/org/odk/collect/android/formentry/javarosawrapper/FormIndexUtils.java
```diff
@@ -11,7 +14,7 @@ private FormIndexUtils() {
     }
 
     /**
-     * used to go up one level in the formIndex. That is, if you're at 5_0, 1 (the second question
+     * Used to find one level up from the formIndex. That is, if you're at 5_0, 1 (the second question
      * in a repeating group), this method will return a FormIndex of 5_0 (the start of the repeating
      * group). If your at index 16 or 5_0, this will return null;
```
lognaturel: While improving this comment, your -> you're and lose the last ;

seadowg:
> On my Android 5.1 device, the button does not show up when I first enter a repeat if the repeat is not the first question. If I tap the overflow menu or rotate the screen then it does appear. Similarly, if I exit a repeat, the add button doesn't disappear until I force the activity to refresh. I'm surprised the Robolectric tests don't catch that.

Are you only seeing this on 5.1? I'm wondering if it could be something to do with the options menu not being invalidated as I expected on different versions. Will check this on other versions myself later.

collect_app/src/main/java/org/odk/collect/android/formentry/javarosawrapper/FormController.java
```diff
@@ -892,6 +879,21 @@ public int jumpToIndex(FormIndex index) {
         return formEntryController.jumpToIndex(index);
     }
 
+    /**
+     * Jumps to the next prompt for a repeated instance of the group referenced by the current FormIndex.
+     */
+    public void jumpToNewRepeatPrompt() {
+        FormIndex repeatGroupIndex = getRepeatGroupIndex(getFormIndex(), getFormDef());
+        String repeatRef = repeatGroupIndex.getReference().toString(false);
```
lognaturel: Would be better not to use `toString` here. Can you compare actual references?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1446,38 +1444,42 @@ private void showNextView() {
...
+            default:
+                Timber.w("JavaRosa added a new EVENT type and didn't tell us... shame on them.");
+                break;
+        }
+    }
+
+    private boolean saveBeforeNextView(FormController formController) {
+        formController.getAuditEventLogger().flush();    // Close events waiting for an end time
+
+        if (formController.currentPromptIsQuestion()) {
```
lognaturel, line 1489: Why lose the null check here? It shouldn't be possible for the form controller to be null here, is that why? I think it may still somehow happen and it seems better to fail by nothing happening than by crashing?

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -50,9 +50,7 @@ public void formLoaded(FormController formController) {
     }
 
     public void editingForm() {
-        if (formController.getAuditEventLogger() != null) {
-            formController.getAuditEventLogger().setEditing(true);
-        }
+        formController.getAuditEventLogger().setEditing(true);
```
lognaturel, line 65: Why lose the null checks here?

collect_app/src/main/java/org/odk/collect/android/formentry/FormIndexAnimationHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public FormIndexAnimationHandler(Listener listener) {
+        this.listener = listener;
+    }
+
+    public void handle(FormIndex index) {
+        if (lastIndex == null) {
+            listener.refreshCurrentView();
+        } else {
+            if (index.compareTo(lastIndex) > 0) {
```
lognaturel, line 28: This is a little overly paranoid but did you do a quick check that `compareTo` does what you want/expect? It would be its first time being used. I just opened it and looked and it seems fine but I think it would be good for one of us to do a slightly deeper think-through.

collect_app/src/main/java/org/odk/collect/android/formentry/repeats/AddRepeatDialog.java
```diff
@@ -15,21 +14,21 @@
 
     private AddRepeatDialog() {}
 
-    public static void show(Context context, FormController formController, Listener listener) {
+    public static void show(Context context, String groupName, Listener listener) {
```
lognaturel: `groupLabel`?

lognaturel:
> Are you only seeing this on 5.1?

Shoot, the problem was that I was locally looking at `seadowg/inline-repeat` but this pull request is actually off `upstream/inline-repeat` which is ahead. I see the commits that fix this, now, and I can confirm that it works as expected.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/AddRepeatTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.pages.EndOfFormPage;
+import org.odk.collect.android.support.pages.FormEntryPage;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+@RunWith(AndroidJUnit4.class)
+public class AddRepeatTest {
+
+    private static final String FORM = "one-question-repeat.xml";
```
lognaturel: How about making the form have a question before the repeat and testing that there's no button on that first question and that it appears on swipe? Or do you think the Robolectric test with mock that sets whether or not the current question is a repeat is sufficient for that?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -965,72 +988,52 @@ private void saveFileAnswer(Object media) {
...
         return true;
     }
 
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
+        if (optionsMenuDelegate.onOptionsItemSelected(item)) {
+            return true;
+        }
+
+        // These actions should move into the `FormEntryMenuDelegate`
```
lognaturel, line 1026: File a good first issue?

collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
```diff
@@ -1,9 +1,17 @@
...
 
 public class AnalyticsEvents {
+
     private AnalyticsEvents() {
+
     }
 
+    /**
+     * Used to measure how popular different methods of adding repeats are. The "action" should
+     * describe the way the repeat was added.
```
lognaturel: How about also logging a form hash so we can differentiate between one organization training staff to use the button vs. lots of people trying it out.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
```diff
@@ -0,0 +1,125 @@
...
+import org.odk.collect.android.preferences.PreferencesActivity;
+import org.odk.collect.android.utilities.MenuDelegate;
+import org.odk.collect.android.utilities.PlayServicesUtil;
+
+import static org.odk.collect.android.preferences.GeneralKeys.KEY_BACKGROUND_LOCATION;
+
+public class FormEntryMenuDelegate implements MenuDelegate {
+
+    private final AppCompatActivity context;
+    private final FormControllerProvider formControllerProvider;
```
lognaturel: You need this indirection here because the controller may not be set when this is instantiated?

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -0,0 +1,110 @@
...
+            analytics.logEvent(ADD_REPEAT, "Inline");
+        } else if (fromPrompt) {
+            analytics.logEvent(ADD_REPEAT, "Prompt");
+        } else {
+            analytics.logEvent(ADD_REPEAT, "Hierarchy");
+        }
+
+        getFormController().newRepeat();
+
+        if (!getFormController().indexIsInFieldList()) {
```
lognaturel: I see you have Robolectric tests around a repeat being a field list. Did you try it with a real form? I can't even remember what the intended behavior is. Does the current behavior make sense? How about adding a connected test with a real form for this one. It's the kind of thing that's not super common and that I think we could easily break and not notice.

lognaturel (changes requested):
There's a lot going on here. A few questions and light suggestions inline.

> Moves code that "wraps" JavaRosa into a javarosawrapper package so it's called out

Sounds fine to me but I don't know where it should go as I called out inline.

> Pulls the options menu logic partially out into a FormEntryMenuDelegate (using delegate pattern) that can be iterated on and tested seperately from the FormEntryActivity

Good.

> Extract an interface (MenuDelegate) so the above pattern can be used consistently elsewhere

Sure.

> Adds a component (FormIndedAnimationHandler) for determing whether how we should animate to the "new" screen of a form

I'm eager to see more animation code move out but it's good to do as a separate PR as there's a lot going on here already.

> Move newly added FormController methods and FormIndexUtils to JavaRosa itself

Can't wait.

lognaturel:
With the new button, the eye is really drawn to the upper right corner of the screen. I'm wondering whether we may want to prioritize for v1.27 a review of the size and weight of the menu bar icons and the size of the form name text. I know this will be totally shocking since I always find everything too small but they look really big to me in the context of everything else.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
seadowg: haha yeah

collect_app/src/androidTest/java/org/odk/collect/android/FormNavigationTestCase.java
seadowg: I'm happy with top level if everyone agrees?

collect_app/src/test/java/org/odk/collect/android/formentry/javarosawrapper/FormControllerTest.java
seadowg: I could create an issue instead?

collect_app/src/main/java/org/odk/collect/android/formentry/javarosawrapper/FormController.java
seadowg: Will give this a try

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah I think I need to have a look through these. I really don't like that so many of these methods have to deal with the nullability though. What I might do is try and push that decision as early as I can and use @NonNull and @Nullable to make sure it's clear when we have to check.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
seadowg: Same as the comment before. I don't think this method should be able to be called before the `formController` is set but I'll look into it.

collect_app/src/main/java/org/odk/collect/android/formentry/FormIndexAnimationHandler.java
seadowg: I mean it seems like it works given all the things I've thrown at it but I think we should play around with it in QA a bunch.

collect_app/src/androidTest/java/org/odk/collect/android/formentry/AddRepeatTest.java
seadowg: Might see if I can get that into the `FormController` tests instead as I'd like to avoid setting up anything other than the simplest cases in the "feature tests". I'll review the coverage here for sure.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah I'll follow up with issues. I put a list in the description.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
seadowg: Yup!

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Yeah that might be worth a connected test. Again, I'll review the coverage and see how I feel.

seadowg:
@lognaturel all great comments. Thanks! Will run through a few things again and have a think. Been a couple weeks since I left this so need to refresh a bit!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: Yes, I agree with that general approach. Perhaps it doesn't need to be done here but then I might suggest keeping the existing null checks. It would be really ideal if after the form is expected to be loaded there was no way for the form controller to be null. Even better, as we've previously discussed, would be if the form controller could never be null and instead its functionality changed whether or not a form was active but that's further out.

collect_app/src/main/java/org/odk/collect/android/formentry/javarosawrapper/FormController.java
seadowg: Hmm couldn't get it working with references but worked out a solution using depth that passed all the tests. Makes sense to me but maybe there is an edge case I haven't though of...

collect_app/src/androidTest/java/org/odk/collect/android/formentry/AddRepeatTest.java
seadowg: Sorry I completely misinterpreted this when I read it on Friday. Basically you're suggesting a test for checking the options menu is setup correctly (and whether `invalidateOptionsMenu` is getting called in the right places). I'll review the coverage and maybe do some mutation testing to see if we're got some blind spots.

seadowg: Yeah after playing around with it feels like we're pretty well covered. I've realised looking at it that we had extra calls to `invalidateOptionsMenu` as we really should just call it every time we call `showView`. 

Also looking at this made me realize it would be nice to pull out some a view (holder) that can encapsulate `createView` and `showView` as it would remove a lot of logic from `FormEntryAcitivty` and would let us test things like "every time we should a view we do X". I'll add this to the list of follow ups that I'll make issues for.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: So for the moment I'm just going to go with the `FormController` being `@Nullable` in all the `ViewModel` classes that depend on it and have null check guards on all the actions. There is definitely something better on the tip of my tongue - maybe the ViewModels shouldn't be created until after the `FormController` is available or maybe all the ViewModels should talk to some loader/repo that gives access to the `FormController` as a `Future`. I think the right™️ solution is going to need us to move "form loading" away from the `FormEntryActivity` so best to leave it right now.

The `@Nullable` annotation gives us some safety here but ideally we'd write tests for ALL the null cases. I'm very tempted to jump to Kotlin here if I'm being frank as we'd get compile checks and also much nicer sugar for dealing with the nulls here. That would definitely be a different PR but I think it's worth bringing up to get the discussion going.

seadowg:
@lognaturel the merge for this is pretty horrific. A bit stuck but hopefully be done this afternoon!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: All sound good.

collect_app/src/androidTest/assets/forms/field-list-repeat.xml
```diff
@@ -0,0 +1,26 @@
...
+                    <person>
+                        <age/>
+                    </person>
+                </data>
+            </instance>
+            <bind nodeset="age" type="int"/>
+        </model>
+    </h:head>
+    <h:body>
+        <group ref="/data/person" appearance="field-list">
```
lognaturel: A field-list with a single question is the same as a single question so I don't think this form adds anything. It also looks like `field-list` on a group that encloses a `repeat` doesn't have any effect. I'm not sure whether this is intentional or not and will need to dig deeper. This is what I had in mind:
```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <h:head>
        <h:title>Field-List Repeat</h:title>
        <model>
            <instance>
                <data id="field_list_repeat">
                    <person>
                        <name/>
                        <age/>
                    </person>
                </data>
            </instance>
            <bind nodeset="/data/person/name" type="int"/>
            <bind nodeset="/data/person/age" type="int"/>
        </model>
    </h:head>
    <h:body>
        <group ref="/data/person">
            <label>Person</label>
            <repeat nodeset="/data/person" appearance="field-list">
                <input ref="/data/person/name">
                    <label>Name</label>
                </input>
                <input ref="/data/person/age">
                    <label>What is their age?</label>
                </input>
            </repeat>
        </group>
    </h:body>
</h:html>
```

The checks would be that the + is correctly shown and that both questions are on the same screen.

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -884,13 +884,12 @@ public int jumpToIndex(FormIndex index) {
      */
     public void jumpToNewRepeatPrompt() {
         FormIndex repeatGroupIndex = getRepeatGroupIndex(getFormIndex(), getFormDef());
-        String repeatRef = repeatGroupIndex.getReference().toString(false);
-        String testRef = "";
+        Integer depth = repeatGroupIndex.getDepth();
```
lognaturel, line 899: I like the use of depth here! I can really visualize how you go into a deeper part of the form within the repeat, maybe even deeper into nested repeats and then you reach the same depth at the new repeat event.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -71,14 +71,14 @@ public void addRepeat(boolean fromPrompt) {
     public void cancelRepeatPrompt() {
         analytics.logEvent(ADD_REPEAT, "InlineDecline");
 
-        FormController formController = getFormController();
+        FormController formController = this.formController;
```
lognaturel: Not necessary?

lognaturel (changes requested):
The smallest of comments and then let's get this to QA!

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: haha yeah. I'd imagine AS probably flagged that to me as well!

collect_app/src/androidTest/assets/forms/field-list-repeat.xml
seadowg: Ah right. This doesn't really feel like it has much to do with adding a repeat but more to do with just how repeats are displayed. I think it's good to have a test but I might put it in a different file.

seadowg: Scratch that. Disagree with myself. Adding repeats is basically the core of repeats are. Ignore me.

seadowg: Oh wow it actually did need the code in `FormEntryViewModel`. Who'd have thunk it.

lognaturel (approved):

seadowg:
I'll fix conflicts!

seadowg:
@opendatakit/testers once I've fixed conflicts (and the JavaRosa change is tested) it'd be really helpful if this could tested next before any newer PRs. It's really hard work to keep coming back and fix conflicts on a PR this old as every day I'm less and less in context on the changes 🤯

seadowg:
Urgh something has gone really wrong with this. I'll look at this in the morning. Might need reflog to fix.

mmarciniak90:
@seadowg I noticed one behavior and I'm not sure if this is expected.

When I add new repeat by new + button I see a dialog with buttons DO NOT ADD / ADD.
But when I open the hierarchy view and click on +, the dialog is not visible and the group is added.

Before these changes, the behavior was different. The dialog was visible when + was clicked on the hierarchy view.

Take a look:
ezgif com-video-to-gif

Here is a form used in the example: [nested-repeats.xml.txt](https://github.com/getodk/collect/files/4504449/nested-repeats.xml.txt)

Let me know if this is an intentional change?

seadowg:
@mmarciniak90 good catch. Yes that is intentional. Sorry I added acceptance around analytics for the group being added without the dialog but should have also added acceptance for the visible behaviour change! The change came out of a discussion you'll see in the comments of the issue between @lognaturel and myself.

mmarciniak90:
@seadowg thanks for the explanation.

Now I noticed something in the form [testRepeatCount.xml.txt](https://github.com/getodk/collect/files/4509116/testRepeatCount.xml.txt)
The number of repeats is declared at the beginning and there is no option to add one more - but `+` icon is visible. When I click on `+`, I'm moved to view with `Save Form and Exit`.
The hierarchy view has no `+` sign.

seadowg:
@mmarciniak90 oh yeah great catch! We don't have a test for that scenario. I'll look at fixing that.

seadowg:
@mmarciniak90 I think it'd be good to continue with QA on this though as we could potentially merge this and treat fixed count repeats as a seperate feature. I don't think we want to leave this unmerged much longer.

kkrawczyk123:
@seadowg I have noticed a similar problem that @mmarciniak90 described above but the result is an app freeze and then crash not moving to save and exit view. I have used form: 
[nested-indexed-repeat.xml.txt](https://github.com/getodk/collect/files/4509690/nested-indexed-repeat.xml.txt)

reproduction steps:
1. Add group
2. define a number of animals
3. swipe to next question
4. click on newly added + (should not be available as the number has been defined)
App freezes and you have to wait a little for a crash. 

stacktrace: 
`2020-04-21 13:18:39.548 1410-1715/? E/ActivityManager: ANR in org.odk.collect.android (org.odk.collect.android/.activities.FormEntryActivity)
    PID: 5745
    Reason: Input dispatching timed out (Waiting to send non-key event because the touched window has not finished processing certain input events that were delivered to it over 500.0ms ago.  Wait queue length: 5.  Wait queue head age: 53067.0ms.)
    Parent: org.odk.collect.android/.activities.FormEntryActivity
    Load: 1.82 / 1.61 / 1.4
    CPU usage from 155306ms to 0ms ago (2020-04-21 13:16:00.576 to 2020-04-21 13:18:35.882):
      48% 5745/org.odk.collect.android: 47% user + 0.4% kernel / faults: 49765 minor
      5.1% 1410/system_server: 2.5% user + 2.5% kernel / faults: 16061 minor 560 major
      2.2% 10098/com.google.android.gms.persistent: 1.6% user + 0.5% kernel / faults: 22909 minor 492 major
      2% 630/surfaceflinger: 1.2% user + 0.7% kernel / faults: 448 minor
      0.1% 966/android.hardware.biometrics.fingerprint@2.1-service.fpc: 0% user + 0.1% kernel / faults: 1 minor
      1.3% 1833/com.android.systemui: 0.9% user + 0.3% kernel / faults: 6090 minor 606 major
      0.9% 10126/com.google.android.gms: 0.5% user + 0.3% kernel / faults: 24440 minor 16823 major
      0.8% 5669/kworker/u16:7: 0% user + 0.8% kernel
      0.5% 5668/kworker/u16:3: 0% user + 0.5% kernel
      0.6% 632/android.hardware.graphics.composer@2.2-service: 0.3% user + 0.2% kernel / faults: 36 minor
      0.6% 310/crtc_commit:97: 0% user + 0.6% kernel
      0.4% 5121/kworker/u16:0: 0% user + 0.4% kernel
      0% 29997/com.google.android.apps.wellbeing: 0% user + 0% kernel / faults: 24789 minor 76 major
      0.4% 878/audioserver: 0.2% user + 0.2% kernel / faults: 287 minor 48 major
      0.3% 2150/com.android.phone: 0.2% user + 0.1% kernel / faults: 268 minor 86 major
      0.3% 858/android.hardware.sensors@1.0-service: 0.1% user + 0.1% kernel / faults: 82 minor
      0.3% 4662/kworker/u16:6: 0% user + 0.3% kernel
      0.3% 10708/com.google.android.inputmethod.latin: 0.2% user + 0% kernel / faults: 3073 minor 2 major
      0.2% 563/ueventd: 0.2% user + 0% kernel
      0.1% 5145/com.google.android.apps.messaging: 0.1% user + 0% kernel / faults: 5728 minor 3279 major
      0.2% 10184/com.google.android.googlequicksearchbox:search: 0.1% user + 0.1% kernel / faults: 7824 minor 7595 major
      0.2% 880/lmkd: 0% user + 0.2% kernel
      0% 836/android.hardware.audio@2.0-service: 0% user + 0% kernel
      0.2% 581/logd: 0.1% user + 0.1% kernel / faults: 6 minor
      0.2% 876/msm_irqbalance: 0% user + 0.1% kernel
      0.2% 3674/com.google.android.apps.scone: 0.1% user + 0% kernel / faults: 441 minor
      0.2% 877/ashmemd: 0% user + 0.1% kernel
      0% 1026/adbd: 0% user + 0% kernel / faults: 424 minor
      0.1% 865/vendor.google.wifi_ext@1.0-service-vendor: 0% user + 0.1% kernel / faults: 244 minor
      0.1% 2512/irq/144-1436400: 0% user + 0.1% kernel
      0.1% 257/kgsl_worker_thr: 0% user + 0.1% kernel
      0.1% 438/mmc-cmdqd/0: 0% user + 0.1% kernel
      0.1% 1727/cds_mc_thread: 0% user + 0.1% kernel
      0% 850/android.hardware.light@2.0-service: 0% user + 0% kernel
      0.1% 590/sugov:0: 0% user + 0.1% kernel
      0% 2649/com.google.process.gservices: 0% user + 0% kernel / faults: 1355 minor 1174 major
      0.1% 924/mediaserver: 0% user + 0% kernel / faults: 396 minor 8 major
      0% 934/media.swcodec: 0% user + 0% kernel / faults: 236 minor 1 major
      0.1% 108/system: 0% user + 0.1% kernel
      0.1% 926/statsd: 0% user + 0% kernel / faults: 37 minor
      0.1% 7/rcu_preempt: 0% user + 0.1% kernel
      0.1% 5554/irq/272-synapti: 0% user + 0.1% kernel
      0% 933/qcrild: 0% user + 0% kernel / faults: 29 minor
      0% 4709/com.android.vending:instant_app_installer: 0% user + 0% kernel / faults: 131 minor 62 major
      0% 89/lpass_smem_glin: 0% user + 0% kernel
      0% 582/servicemanager: 0% user + 0% kernel
      0% 583/hwservicemanager: 0% user + 0% kernel
      0% 2949/com.google.android.as: 0% user + 0% kernel / faults: 540 minor 268 major
      0% 32/rcuop/2: 0% user + 0% kernel
      0% 50/rcuop/4: 0% user + 0% kernel
      0% 312/crtc_event:97: 0% user + 0% kernel
      0% 635/vendor.qti.hardware.display.allocator@1.0-service: 0% user + 0% kernel / faults: 360 minor
      0% 928/wificond: 0% user`

I was able to reproduce on both Android 5.1 and 10.

kkrawczyk123:
@seadowg I have also noticed a weird behavior in the form: 
[repeatTest.xml.txt](https://github.com/getodk/collect/files/4510446/repeatTest.xml.txt)

reproduction steps:
1. click add group on the first dialog
2. swipe to next question
3. click on do not add
I can see + on the displayed screen when I use it I am moved to Save and Exit view
when on the same view I choose hierachy button - there is no + button there at all so it gives the impression that + is displayed accidentally here. 

gif:
https://drive.google.com/file/d/1HAWmetVkUA8Vsq1SYLN15ZmU6Mpbhbho/view

seadowg:
@kkrawczyk123 @mmarciniak90 I think that change I just pushed should fix all those cases. As far as I can see it's just that we shouldn't be showing the ➕ button when the repeat has a defined count.

mmarciniak90:
**Tested with success**

I can confirm that the mentioned issues are fixed.
Also, I can confirm that I didn't notice other issues.
I tested a bunch of forms with repeats.

lognaturel:
Great catches! Will be really great to have this merged.

mmarciniak90:

PR #3659, opened 2020-02-28
Revise main menu layout/design

### LINKED ISSUE #3650: Move app version to toolbar and remove tagline
Described by @seadowg at https://github.com/seadowg/collect-design/tree/master/prototypes/revised-main-menu. For now, no changes would be made to the buttons.

Additionally, the git commit hash needs to be added for builds that don't come from a tagged commit. This will only matter to developers so I don't really mind where it goes. I propose putting it right above the buttons and below other UI components such as banners.

seadowg:
Closes #3650. ~~Dependent on/blocked by #3623.~~

Screenshot_1583173638

This removes the title and subtitle from the main menu. The app name and version our now displayed in the app bar and the commit sha/beta tag appear at the bottom of the screen for dev and beta builds.

The app bar color has also changed. This component (along with the `MaterialButton`s that are now being used) on the main menu use our theme's `colorPrimary` which has now been changed to reflect the button colors. Down the line we should look at making this something a little less Windows '95.

Ran through a couple of forms and the UI just to check nothing had changed (in a bad way).

We could have continued to just use `Button` instead of `MaterialButton` but I think this was the moment to pull the tirgger on the change. Next step is to change the rest of the buttons in the app. This will then let us use the `MaterialComponents` theme rather than `AppCompat`. Otherwise the theme automatically changes all our buttons to `MaterialButtons`.

There is a risk that some views now look weird in light or dark mode given the primary color has changed. Good to have a general run through and check but as far as I can see we are ok.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
@@ -1,5 +1,6 @@
 package org.odk.collect.android.activities.viewmodels;
 
+import androidx.annotation.Nullable;
 import androidx.lifecycle.ViewModel;
 
 import org.odk.collect.android.BuildConfig;
@@ -7,6 +8,21 @@ import org.odk.collect.android.BuildConfig;
 public class MainMenuViewModel extends ViewModel {
 
     public String getVersion() {
-        return BuildConfig.VERSION_NAME.split("v")[1].split("-")[0];
+        return versionNameWithoutPrefix().split("-")[0];
+    }
+
+    @Nullable
+    public String getVersionSHA() {
+        String[] split = versionNameWithoutPrefix().split("-", 2);
+
+        if (split.length > 1) {
+            return split[1];
+        } else {
+            return null;
+        }
+    }
+
+    private String versionNameWithoutPrefix() {
+        return BuildConfig.VERSION_NAME.split("v")[1];
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
@@ -1,0 +1,12 @@
+package org.odk.collect.android.activities.viewmodels;
+
+import androidx.lifecycle.ViewModel;
+
+import org.odk.collect.android.BuildConfig;
+
+public class MainMenuViewModel extends ViewModel {
+
+    public String getVersion() {
+        return BuildConfig.VERSION_NAME.split("v")[1].split("-")[0];
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
@@ -1,28 +1,78 @@
 package org.odk.collect.android.activities.viewmodels;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.lifecycle.ViewModel;
+import androidx.lifecycle.ViewModelProvider;
 
+import org.jetbrains.annotations.NotNull;
 import org.odk.collect.android.BuildConfig;
 
 public class MainMenuViewModel extends ViewModel {
 
-    public String getVersion() {
-        return versionNameWithoutPrefix().split("-")[0];
+    private final VersionDescriptionProvider versionDescriptionProvider;
+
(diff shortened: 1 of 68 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+        if (split.length > 1) {
+            return split[1];
+        } else {
+            return null;
+        }
+    }
+
+    private String versionNameWithoutPrefix() {
+        return BuildConfig.VERSION_NAME.split("v")[1];
```
lognaturel: How did you decide to remove the `v`? We have the `v` prefix everywhere that we show a Collect version including GitHub releases, the forum, the Play Store release notes, etc. That's the case across all ODK tools. I like the consistency and would personally keep the `v` in.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class MainMenuViewModel extends ViewModel {
+
+    public String getVersion() {
+        return versionNameWithoutPrefix().split("-")[0];
+    }
+
+    @Nullable
+    public String getVersionSHA() {
+        String[] split = versionNameWithoutPrefix().split("-", 2);
```
lognaturel: This means that for beta versions the beta will be at the bottom of the screen. I think of the beta as being part of the version (e.g. we put v1.26.0-beta.0 in the release notes) and I think it's helpful to have up top.

lognaturel (commented):
Couple of quick discussion points inline.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
seadowg: Yeah no strong feeling really. Will change this.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
seadowg: Make sense. I'll change it!

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModelTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+@RunWith(RobolectricTestRunner.class)
+public class MainMenuViewModelTest {
+
+    @Test
+    public void version_whenRelease_returnsSemanticVersionWithPrefix() {
+        MainMenuViewModel viewModel = new MainMenuViewModel(() -> "v1.1.7");
```
lognaturel, line 15: Nice! These tests end up looking really nice.

I probably would have pushed all the methods now in the view model into the version provider class and kept the view model super thin so that tests wouldn't need `Robolectric` and to limit the dependence on Android-specific components for paranoia reasons but I do like how all this ends up looking.

lognaturel (approved):

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModelTest.java
seadowg: Yeah another step would be to extract a `VersionDescriptionParser` to deal with all this (and move the tests there) but I don't think that's worth it until the `ViewModel` is doing other things.

lognaturel: Makes sense!

kkrawczyk123:
@seadowg, is changing buttons color in dark mode an intentional here? probably it is, but I have to ask as I am not convinced about making it darker:
| Collect 1.25.2 | Pr's build |
| ------- | ---- |
| Screenshot_2020-03-03-10-55-58 | Screenshot_20200303-105530 |

seadowg:
@kkrawczyk123 oooo I double checked this but must have changed something after. In this case I think the app bar and the buttons should be the same color as the buttons in 1.25. I'll fix.

kkrawczyk123:
**Tested with success!**

Verified cases:
- light mode
- dark mode
- rotate screen
- header with migration banner and without
- change language
- go through other headers in the app 

seadowg:
Getting a test failure on test lab that looks like a scrolling issue so lets hold off on merging

PR #3661, opened 2020-02-29
Show that form updates are available with Google Drive as server

### LINKED ISSUE #3048: Show that form updates are available with Google Drive integration
Collect v1.21.2 and prior

#### Problem description
When the server is set to Google Drive, there is no indication that form updates are available.

Upload a form to Google Drive, upload it or its media, go to Get Blank Form and see that there is no indication of updates.

#### Expected behavior
The "This is an update to a form you have" message should show when updates are available.

#### Other information 
We should really consider abstracting away commonalities between the two download lists at some point.

chidauri:
Closes #3048 

Tested on Android 9, with multiple forms.

Since manifest info. or any other useful info is not available, **last modified time** seemed to be the most reasonable one. And i am using java file system's last modified value to compare because the forms database doesn't update if the md5 hasn't changed, so even if we download the new form, the database's last modified time doesn't update.

No


lognaturel:
Thanks so much for taking this on, @chidauri! The high-level approach sounds great and a quick peek at the code is also looking very positive.

As I'm sure you've seen, we're working towards a code freeze this week for v1.26 and have a number of critical PRs to get into that. I think we'll have to come back to this one after the release is out and it can be the first improvement in v1.27.

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
```diff
@@ -103,4 +115,15 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
         holder.onBind(items.get(position));
         return view;
     }
+
+    private boolean isNewVersion(DriveListItem item) {
+        FormsDao formsDao = new FormsDao();
+        Cursor cursor = formsDao.getFormsCursorForFormFilePath(Collect.FORMS_PATH + File.separator + item.getName());
+        if (cursor != null && cursor.getCount() > 0) {
+            Long lastModifiedLocal = new File(Collect.FORMS_PATH + File.separator + item.getName()).lastModified();
```
grzesiek2010: If you already have cursor it's better to read lastModifiedLocal from its DATE column.

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
```diff
@@ -103,4 +115,15 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
         holder.onBind(items.get(position));
         return view;
     }
+
+    private boolean isNewVersion(DriveListItem item) {
+        FormsDao formsDao = new FormsDao();
+        Cursor cursor = formsDao.getFormsCursorForFormFilePath(Collect.FORMS_PATH + File.separator + item.getName());
+        if (cursor != null && cursor.getCount() > 0) {
```
grzesiek2010: I think it's better to use `cursor != null && cursor.moveToFirst()` like we do in other places. `cursor.moveToFirst()` will return false if `cursor.getCount() <= 0` and make sure you are at the first position.

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
```diff
@@ -76,10 +82,16 @@ void onBind(DriveListItem item) {
...
+                if (isNewVersion(item)) {
+                    formUpdateAlert.setVisibility(View.VISIBLE);
+                } else {
+                    formUpdateAlert.setVisibility(View.GONE);
+                }
             } else {
                 Drawable d = ContextCompat.getDrawable(getContext(), R.drawable.ic_folder);
                 imageView.setImageDrawable(d);
                 checkBox.setVisibility(View.GONE);
+                formUpdateAlert.setVisibility(View.GONE);
```
grzesiek2010, line 86: The view is gone by default but we need to set it here to avoid problem while scrolling when elements are reused am I right?

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
```diff
@@ -76,10 +82,16 @@ void onBind(DriveListItem item) {
                 Drawable d = ContextCompat.getDrawable(getContext(), R.drawable.form_state_blank);
                 imageView.setImageDrawable(d);
                 checkBox.setVisibility(View.VISIBLE);
+                if (isNewVersion(item)) {
+                    formUpdateAlert.setVisibility(View.VISIBLE);
+                } else {
+                    formUpdateAlert.setVisibility(View.GONE);
```
grzesiek2010: If gone is needed at all here (depends on my question below since it;s the same case) I would prefer using ternary operator: `formUpdateAlert.setVisibility(isNewVersion(item) ? View.VISIBLE : View.GONE);`

collect_app/src/main/res/layout/form_chooser_list_item_text.xml
```diff
@@ -34,7 +34,7 @@
 
     <TextView
         android:id="@+id/form_update_alert"
-        android:layout_width="wrap_content"
+        android:layout_width="match_parent"
```
grzesiek2010, line 37: Why do we need this change?

grzesiek2010 (changes requested):
I added some comments but the main question is:
Have you considered comparing media files as well. If we use a server like ODK Aggergate the message is displayed even if only a single media file has changed and the form (xml files) is still the same?

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
chidauri: I tried using that but that poses a problem. Since in this case we do not have access to md5 or any other form specific data it is possible that the form available on google drive is essentially the same form(not an update!) but it would show as an update to the user because we're comparing modified time on gdrive and so if the user chooses to download that form, the database doesn't update(because we have md5 check there) so the DATE column also doesn't update, so the user will still see that the update is available any number of times he downloads the form.

lognaturel:
@chidauri Do you think you'll have a chance to get back to this at some point in the next couple of weeks and incorporate @grzesiek2010's feedback? It'd be a great improvement for v1.27. Thanks!

chidauri:
> @chidauri Do you think you'll have a chance to get back to this at some point in the next couple of weeks and incorporate @grzesiek2010's feedback? It'd be a great improvement for v1.27. Thanks!

Sure :)

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
chidauri: yep, absolutely right!

collect_app/src/main/res/layout/form_chooser_list_item_text.xml
chidauri: Some proportion of the text was cut because of this.

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
chidauri: @grzesiek2010 can we ever face the problem i mentioned above or we're fine using DATE column??

lognaturel: @chidauri this would happen if a user did something like change the form definition and then change their mind and change it back, right? That seems like an uncommon enough scenario that I wouldn’t worry about it. Is there another case that would be affected?

There is possible weirdness if the user’s device has the wrong date/time, around daylight savings boundaries and if the device is in a time zone other than the one Google Dive is in but I think these are ok to document and not worry about.

chidauri: @lognaturel this would also happen if we upload the form we already have to google drive. Also, is it fine to use last modified time for media files too?

collect_app/google-services.json
```diff
@@ -15,7 +15,7 @@
       },
       "oauth_client": [
         {
-          "client_id": "640412404956-ag8do5nic02k70qr2k1a0mv3du8222hv.apps.googleusercontent.com",
+          "client_id": "329060319550-7dpmv9fnoon5ton9r1b1iqmgjr41okgt.apps.googleusercontent.com",
```
grzesiek2010: Please revert this change.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/adapters/FileArrayAdapter.java
grzesiek2010: Oh I see thanks for the explanation. In this case leaving this as is seems fine to me.

grzesiek2010:
> Have you considered comparing media files as well. If we use a server like ODK Aggergate the message is displayed even if only a single media file has changed and the form (xml files) is still the same?

I think you didn't answer my question or maybe I have missed something?

chidauri:
@grzesiek2010 i have now added checks for media files too, can be reviewed now.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -559,7 +562,7 @@ public void onClick(View v) {
                     if (folderIdStack.empty()) {
                         parentId = ROOT_KEY;
                     } else {
-                        parentId = folderIdStack.pop();
+                        parentId = folderIdStack.peek();
```
grzesiek2010, line 568: What is it for? I think it's not related to this issue.

chidauri: This fixes the back button behaviour, before this change if we clicked on some folder and returned back to parent folder, the update messages on the forms were gone.

chidauri: Though i wonder why navigating up and down the hierarchy of folders was fine.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -729,47 +739,107 @@ protected void onPostExecute(HashMap<String, Object> results) {
...
-            Collections.sort(forms);
-            driveList.addAll(dirs);
-            driveList.addAll(forms);
-            updateAdapter();
         }
     }
 
+    private boolean isNewerFormVersionAvailable(DriveListItem item) {
+        FormsDao formsDao = new FormsDao();
+        Cursor cursor = formsDao.getFormsCursorForFormFilePath(new StoragePathProvider().getDirPath(StorageSubdirectory.FORMS) + File.separator + item.getName());
```
grzesiek2010: The cursor should be closed/ You can use [try-with-resources](https://docs.oracle.com/javase/8/docs/technotes/guides/language/try-with-resources.html) for that.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -729,47 +739,107 @@ protected void onPostExecute(HashMap<String, Object> results) {
...
+        if (cursor != null && cursor.moveToFirst()) {
+            Long lastModifiedLocal = new File(new StoragePathProvider().getDirPath(StorageSubdirectory.FORMS) + File.separator + item.getName()).lastModified();
```
grzesiek2010: I think it;s time to use injected `StoragePathProvider` because in many places we use `new StoragePathProvider()`. It's already available in [AppDependencyModule](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java#L256) and this activity is also ready to handle injected objects https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java#L127 so you can just use it like we do with [GoogleAccountsManager](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java#L110) for example.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -729,47 +739,107 @@ protected void onPostExecute(HashMap<String, Object> results) {
...
+        Collections.sort(forms);
+        driveList.addAll(dirs);
+        driveList.addAll(forms);
+        checkFormUpdates();
+    }
+
+    private void checkFormUpdates() {
+        for (DriveListItem item: driveList) {
+            if (item.getType() == DriveListItem.FILE) {
+                if (isNewerFormVersionAvailable(item) || areNewerMediaFilesAvailable(item)) {
```
grzesiek2010: Please merge those two nested if statements. It can be just one.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -729,47 +739,107 @@ protected void onPostExecute(HashMap<String, Object> results) {
...
+                        Long lastModifiedLocal = localMediaFile.lastModified();
+                        Long lastModifiedServer = mediaFile.getModifiedTime().getValue();
+                        if (lastModifiedServer > lastModifiedLocal) {
+                            return true;
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            Timber.e(e);
```
grzesiek2010: I would log it as a warning not an error so `Timber.w(e);` Since it's not something very important and on the other hand not necessarily our fault.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -701,7 +704,8 @@ protected void onPreExecute() {
                         nextPage.put(PARENT_ID_KEY, parentId);
                         nextPage.put(CURRENT_ID_KEY, currentDir);
                         nextPage.put(FILE_LIST_KEY, driveFileListPage);
-                        publishProgress(nextPage);
+                        filterForms(nextPage);
+                        publishProgress();
```
grzesiek2010, line 711: Could you clarify this change? Why did you factor out `filterForms()` which contains generally the same code as `publishProgress` before and now call both?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
chidauri: The filtering which was being done in `publishProgress` , i needed it for calculation of media folders and all these should run on background thread because of being computationally expensive, Whereas `publishProgress` runs on UI thread that's why.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
```diff
@@ -795,18 +867,15 @@ void setGoogleDriveFormDownloadListener(GoogleDriveFormDownloadListener gl) {
...
                         results.put(fileItem.getName(), getString(R.string.multiple_media_folders_detected_notification));
                         return results;
                     }
 
-                    if (folderId != null) {
-                        List<com.google.api.services.drive.model.File> mediaFileList;
-                        mediaFileList = driveHelper.getFilesFromDrive(null, folderId);
-
+                    if (mediaFileList != null) {
                         FileUtils.createFolder(new StoragePathProvider().getDirPath(StorageSubdirectory.FORMS) + File.separator + mediaDirName);
```
grzesiek2010: Please also check the entire file and use injected storagePathProvider instead of `new StoragePathProvider()` like here.

grzesiek2010 (changes requested):
I think it's the last small thing and we can start testing.

collect_app/src/main/java/org/odk/collect/android/activities/GoogleDriveActivity.java
chidauri: Done.

grzesiek2010 (approved):

mmarciniak90:
@chidauri QA team noticed a few issues in this PR

1. The time needed to load a list increased diametrically. We run some tests and we noticed that it increased from ~1sec to ~23 sec. A longer list of forms is loaded up about 4 minutes what makes the app unusable. This also makes testing difficult.
2. Forms have information about the update even if there are no downloaded forms at all



cc: @lognaturel 

grzesiek2010:
@chidauri 
are you able to figure out what exactly consumes the time? Is it something we can fix?

chidauri:
@grzesiek2010 @lognaturel 
Currently every form whether it is downloaded or not is being checked so that is to be fixed, but even after it is fixed, the current approach of checking every media file in the cloud with local file will always be time consuming(when number of downloaded forms are large). 

We can try another approach where we only compare the latest(or last) modified media file with its local counterpart, if that has been changed then we can conclude that this is an update otherwise conclude that other files are unchanged too. This should save us time.

OR Maybe we can skip checking media files for google drive or introduce a "check for updates" button so we don't make the app unusable. what do you think?

grzesiek2010:
> Currently every form whether it is downloaded or not is being checked so that is to be fixed, but even after it is fixed, the current approach of checking every media file in the cloud with local file will always be time consuming(when number of downloaded forms are large).

Ok so please fix this bug first. You don't download those media files but just the list of them with their metadata so it shouldn't be that time consuming.

@mmarciniak90 do you have a lot media files attached to those forms? what would be the number more or less?

grzesiek2010:
Now as I investigated files metadata on GD I realized there is `getMd5Checksum()` why don't you use instead of `lastModyfiedTime`?

chidauri:
> Now as I investigated files metadata on GD I realized there is `getMd5Checksum()` why don't you use instead of `lastModyfiedTime`?

I tried using that but it returned empty values or there was some other issue i don't remember exactly, but i am sure it doesn't work.

grzesiek2010:
> I tried using that but it returned empty values or there was some other issue i don't remember exactly, but i am sure it doesn't work.

ok I checked it and it returns null because according to the doc:

> The MD5 checksum for the content of the file. This is only applicable to files with binary content in Google Drive

could you add a comment above `checkFormUpdates()` method? otherwise we will forget about it soon and will wonder why we decided to use dates.

When it comes to checking all media files hmm... I'm not sure it would be good to have it but yeah it's slower, @lognaturel what do you think?

grzesiek2010:
@mmarciniak90 

how many forms and media files did you have on your drive while testing? (more or less of course)

mmarciniak90:
@grzesiek2010, In one directory which we use the most - ~180 forms + directories, which also contains forms.

grzesiek2010:
> @grzesiek2010, In one directory which we use the most - ~180 forms + directories, which also contains forms.

That's why. I think it's a pretty unusual case and if we tested about 10 forms the behavior shouldn't be that bad.
We could think of an option to disable checking updates for those who have a lot of forms or poor connection.

lognaturel:
Thanks again for all the great follow-up here. The changes you've come up with @grzesiek2010 and @chidauri seem reasonable to me. Let's do another round of QA with a smaller number of forms and media files and see where we are. ~10 seems like a good number.

grzesiek2010:
@chidauri 
I think you removed comparing media files could you add the code back so that we can test it again?

lognaturel:
Ah, that's right, I hadn't fully understood 2d01066. Yes, it would be good to revert that commit.

grzesiek2010 (approved):

mmarciniak90:
@chidauri @lognaturel 
I have to admit that performance is better. The form list loads smoother. I do appreciate that.

I noticed a few behaviors which are not working as I imagined. Let me know what do you think about them.
**1st:** The automatic download of the latest version does not work for Google Drive
**2nd:** I downloaded a few versions of the same form. When I unassigned option `Hide old form versions` I saw only the latest
**3rd:** I noticed this behavior only on Android 7.0.
   1. I downloaded form
   2. I added another version to GD and I saw the update message but I didn't download it.
   3. I migrated data
   4. I opened the form list to download the latest version but the message was not available.

lognaturel:
> The automatic download of the latest version does not work for Google Drive

That's right and I think it's ok to treat that as a follow-on feature. 

> I downloaded a few versions of the same form. When I unassigned option Hide old form versions I saw only the latest

I don't think this is related to this PR so we can file a separate issue. It was just harder to notice before. On `master`, I downloaded a version of a form, updated that same form, downloaded the new version, and I verified that if I go to `Delete Saved Form` -> `Blank Form`, I only see the most recently downloaded.

> I migrated data

Are you sure you only saw this on Android 7? This PR uses the file last modified timestamp and compares it to the server timestamp. Doing a data migration updates the local timestamp so unfortunately I would expect that it always has the effect of masking updates that were made available before the migration but weren't downloaded at that time. It's a temporary issue that will go away in a couple of months once the migration is forced so my preference would be to leave things as they are and document the limitation. Do let me know if you think I misunderstood the nature of the issue, though.

kkrawczyk123:
@lognaturel I was not able to reproduce the migration case described by @mmarciniak90 above on Android 7.1.

I have noticed some other bothering me cases, I followed scenario:
1. Downloaded 1st version from Google Drive
2. Uploaded 2nd version to another folder in Google Drive
3. Checked for an update - It is not visible.
4. Downloaded form - I can see that a new version has been downloaded.
gif with the steps:
ezgif com-video-to-gif (8)

Then I've tried a case, using only the same folder:
1. Downloaded 1st version
2. Updated the form in the same folder with version 2nd
3. Checked for update - I saw information about form update
4. I downloaded the form.
5. Deleted all blank forms
6. Repeated the same scenario - Form update information is not visible anymore.
It might be confusing when the user actually might have that form previously downloaded but doesn't remember that at all. It makes me feel like the update is one-time useable.

chidauri:
@kkrawczyk123 This happens because the first version is downloaded after the second version has been uploaded, which makes local modified time later than server's, this is a limitation unfortunately, in this approach.

lognaturel:
> The automatic download of the latest version does not work for Google Drive

> That's right and I think it's ok to treat that as a follow-on feature.

I talked to @seadowg about this since he's touching auto-update as part of adding the "match exactly" update mode. We agreed that he will disable all auto-update options if the server type is Google Sheets. This will help prevent confusion and I think given this we shouldn't file the issue yet and instead let users ask for it. @chidauri's addition here takes care of the most important missing behavior. So @mmarciniak90 don't worry about an issue for this.

I agree with @chidauri that the cases @kkrawczyk123 explored illustrate limitations of using timestamps to detect form updates. We don't have a great alternative and those scenarios don't seem very common so I think documentation is sufficient to explain what's going on.

mmarciniak90:
Based on earlier comments, @kkrawczyk123  and I can close this issue as tested. 
We know the limitations of the approach based only on timestamps and all of them will be documented. 

Verified on Android versions: 5.1, 7.0, 7.1, 8.1, 9.0, 10.0

Verified cases:
- updated XML
- updated media
- added additional media
- different form versions in different GD directories
- update XML from GD when the first form was downloaded from Aggregate

PR #3665, opened 2020-03-02
Add analytics for features we want to know more about and consolidate all analytics event names

lognaturel:
Ran all tests and verified that the app starts as intended. I have also verified that events show up in Firebase Analytics.

@seadowg pointed out that it would be helpful to tell at a glance what analytics are being tracked and why. He introduced the `AnalyticsEvents` class to do that at https://github.com/opendatakit/collect/pull/3656/files#diff-6993ffa5c4789574e3f416fabbf4137f and I'm using that same approach.

I considered only adding the new analytics but I think this is a logical time to pull out all the constants. I separated out the commits to hopefully make review easier and hopefully make this very low risk.

I considered removing the `logRemoteAnalytics` method and injecting the `Analytics` singleton as needed but that made the change bigger than I wanted at this point in a release cycle.

There should be no change in behavior as far as users are concerned. We'll get more analytics which should helps us make better decisions. I tried to make this very low risk overall but if anything looks risky let me know. Every class with analytics has been touched. I think the highest risk is to preferences where there were several changes.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -177,6 +177,7 @@ import timber.log.Timber;
 import static android.content.DialogInterface.BUTTON_NEGATIVE;
 import static android.content.DialogInterface.BUTTON_POSITIVE;
 import static android.view.animation.AnimationUtils.loadAnimation;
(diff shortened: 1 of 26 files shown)


lognaturel:
Conflicts will need to be resolved when #3668 is merged.

grzesiek2010 (approved):

PR #3666, opened 2020-03-03
Use an ingestion timestamp to detect whether the search/pulldata handler should re-process new data file

### LINKED ISSUE #3335: The search()/pulldata() handler consumes files meant to be used as external secondary instances

lognaturel:
Closes #3335

I cherry picked @OpenDataNerd's excellent commits from https://github.com/OpenDataNerd/collect/tree/csv_external_data. The XLSForm documentation has been updated at http://xlsform.org/en/#multiple-choice-from-file to recommend using CSV external instances so it's now feeling more urgent to get this fixed.

Ran automated tests. Opened a form with external secondary CSV instance multiple times and one with search() multiple times. I also tried a form with an `itemsets.csv` file even though there should be no interaction with that feature.

The issue discusses alternatives. In particular, I had suggested using a file hash to identify changes. I think the time stamp solution is simpler and just as effective. We also could have considered storing the time stamps in an auxiliary file or something like that but adding a metadata table keeps things well-organized. 

Regression risk is around everything having to do with external data. 
Config for a Central project with two forms: one with a CSV external instance and one with a search() CSV:
csv

No. 


seadowg:
I think in the interest of getting this through it would be good for @opendatakit/testers to look at this in parallel to review so I'll mark this as "needs testing" now

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.odk.collect.android.utilities.CustomSQLiteQueryBuilder;
+import org.odk.collect.android.utilities.FileUtils;
+import org.odk.collect.android.utilities.SQLiteUtils;
+
+import java.io.File;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ExternalDataTest {
```
seadowg: Is this really `ExternalDataReaderTest`?

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @Before
+    public void setUp() {
+        File formPath = new File(new StoragePathProvider().getDirPath(StorageSubdirectory.FORMS) + File.separator + SIMPLE_SEARCH_EXTERNAL_CSV_FORM_FILENAME);
+        File mediaDir = FileUtils.getFormMediaDir(formPath);
+        csvFile = new File(mediaDir + File.separator + SIMPLE_SEARCH_EXTERNAL_CSV_FILENAME);
+        dbFile = new File(mediaDir + File.separator + SIMPLE_SEARCH_EXTERNAL_DB_FILENAME);
+    }
+
+    @Rule
```
seadowg: Could this run in Robolectric instead?

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
```diff
@@ -0,0 +1,199 @@
...
+            .around(new CopyFormRule(SIMPLE_SEARCH_EXTERNAL_CSV_FORM_FILENAME, Collections.singletonList(SIMPLE_SEARCH_EXTERNAL_CSV_FILENAME)));
+
+    private static Map<String, File> makeExternalDataMap(File csvFile) {
+        Map<String, File> externalDataMap = new HashMap<>();
+        externalDataMap.put(SIMPLE_SEARCH_EXTERNAL_CSV_NAME, csvFile);
+        return externalDataMap;
+    }
+
+    @Test
+    public void testCreateDBonImportCSV() {
```
seadowg: These tests are pretty dense and I'm having a hard time understanding what they are driving at. I think using a different style for the names would be helpful. This test could be `doImport_createsDataAndMetaDatabaseTables`. That way we know what the subject is (`ExternalDataReaderImpl#doImport`) and what our expectation is pretty clearly.

collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
```diff
@@ -14,13 +15,19 @@
 
 import java.io.File;
 import java.util.Arrays;
+import java.util.Collections;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.notNullValue;
 
 public class FormLoaderTaskTest {
```
seadowg, line 23: I'd love to say "this should be Robolectric" but I think that would involve picking the task apart. The `XTaskTest` classes are pretty much the prime candidates for moving out of Espresso so I'm confident that will happen later.

seadowg (changes requested):
I found some of the tests hard to follow so I've left comments about that for ways that could hopefully be improved.

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
lognaturel: Yes, that sounds right. Will change!

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
lognaturel: I think so, good call. Let me explore that...

collect_app/src/androidTest/java/org/odk/collect/android/external/ExternalDataTest.java
lognaturel: Yes, sounds good.

collect_app/src/androidTest/assets/forms/simple-search-external-csv.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lognaturel, line 1: It's too bad to have duplicates of these test files in `androidTest` and `test`. I think it's the right thing to do the Robolectric tests with real form files, though.

This might actually be a good reason for all tests to live in the same source set, as the androidx libs want.

collect_app/src/androidTest/java/org/odk/collect/android/tasks/FormLoaderTaskTest.java
lognaturel: Yeah, agreed that coming back to this sounds good. 

I think the new Robolectric tests may actually offer sufficient coverage. There is something comforting about knowing that these are even more end-to-end, though.

collect_app/src/test/java/org/odk/collect/android/external/ExternalDataReaderTest.java
```diff
@@ -0,0 +1,225 @@
...
+    private static File dbFile;
+
+    private static Map<String, File> formDefToCsvMedia;
+
+    @Before
+    public void setUp() throws IOException {
+        ShadowEnvironment.setExternalStorageState(Environment.MEDIA_MOUNTED);
+        new StorageInitializer().createOdkDirsOnStorage();
+
+        File formFile = new File(new StoragePathProvider().getDirPath(StorageSubdirectory.FORMS) + File.separator + SIMPLE_SEARCH_EXTERNAL_CSV_FORM_FILENAME);
```
lognaturel, line 59: This work here is basically exactly what `CopyFormRule` and the methods in `FormLoadingUtils` do but they're in different sourcesets. Again, maybe a reason to have all tests in one. Or if there is an elegant way to share that code, please feel free to modify, @seadowg.

collect_app/src/test/java/org/odk/collect/android/external/ExternalDataReaderTest.java
```diff
@@ -0,0 +1,225 @@
...
+        externalDataReader = new ExternalDataReaderImpl(null);
+        externalDataReader.doImport(formDefToCsvMedia);
+        assertThat(dbFile.exists(), is(true));
+        db = SQLiteDatabase.openDatabase(dbFile.getAbsolutePath(), null, SQLiteDatabase.OPEN_READWRITE);
+        assertThat("metadata table should be recreated", SQLiteUtils.doesTableExist(db, EXTERNAL_METADATA_TABLE_NAME));
+        db.close();
+    }
+
+    @Test
+    public void doImport_reimportsCsvIfFileIsUpdated() throws IOException, InterruptedException {
```
lognaturel, line 161: Note that this test has changed. It previously faked an update by changing the db timestamp. Now it actually tests a real file change.

collect_app/src/androidTest/assets/forms/simple-search-external-csv.xml
seadowg: I think it's fine to duplicate. A nice solution is to bundle the test forms in gradle submodule if it gets annoying. I have thoughts on everything living in the same set but they can be a fun discussion another time!

collect_app/src/test/java/org/odk/collect/android/external/ExternalDataReaderTest.java
seadowg: I'm happy with this for now!

collect_app/src/test/java/org/odk/collect/android/external/ExternalDataReaderTest.java
```diff
@@ -0,0 +1,225 @@
...
+        assertThat(dbFile.exists(), is(true));
+
+        SQLiteDatabase db = SQLiteDatabase.openDatabase(dbFile.getAbsolutePath(), null, SQLiteDatabase.OPEN_READWRITE);
+        assertThat(db.rawQuery(SELECT_ALL_DATA_QUERY, null).getCount(), is(3));
+
+        long originalTimestamp = csvFile.lastModified();
+        long metadataLastModified = ExternalSQLiteOpenHelper.getLastImportTimestamp(db, EXTERNAL_METADATA_TABLE_NAME, csvFile);
+        assertThat(metadataLastModified, is(originalTimestamp));
+
+        Thread.sleep(1000); // this test is so fast that we need to guarantee the modified time changes
```
seadowg, line 174: I think there is a way to fake this with fake file layer. Can't remember the name of the framework though. We can stick with this right now as I'm pretty sure it's still deterministic.

seadowg (approved):

kkrawczyk123:
@lognaturel I have identified one weird scenario using that form as it is visible on master too I reported it separately in #3682. 
The pr was verified on Androids: 5.1, 6.0, 7.0, 8.1, 9.0 and 10.
I have cooperated with @mmarciniak90 and we have verified cases:
- download attached forms from Central
- fill forms
- edit filled forms
- send finalized forms
- migrate and again fill, edit, send both forms
- download the same forms from Google drive: fill, edit, send, migrate, again fill, edit
- download form with external CSV and last save: fill, edit, send, migrate, again fill, edit
- check other forms with external CSV files: fill, edit, send, migrate, again fill, edit

PR #3667, opened 2020-03-03
Fixed bug with radioButtons/CheckBoxes on Android 6

### LINKED ISSUE #3664: Answer in select one/multi widget is not marked when the user moves back - Android 6.0
Collect v1.25.0 
Android 6.0

#### Problem description
Answer in select one widget is not marked when the user moves back

1. User starts a form ([forms.zip.txt](https://github.com/opendatakit/collect/files/4275922/forms.zip.txt) / `All widgets`)
2. User puts answers
3.  User moves back
4. The selected answer is unselected
5. User opens hierarchy view - all answers are visible
6. The selection is back after changing the device orientation
and6

#### Expected behavior
Selected answer is marked

grzesiek2010:
Closes #3664 

I tested the fix on Android 6.

Looks as if it's a bug on Android 6. The selected answer is saved only radioButtons/CheckBoxes don't show it. It was introduced in #3474 where I wanted to fix another bug but visible on Android 5.
Since the problem on android 5 was less serious **and** 5 probably is less popular os version than 6 **and** as I remember the problem from #3474 wasn't even visible on every device with Android 5
I think the best way is juts revert that change what I did here.

It just fixes the problem. Nothing else should be affected.

Any form with `select_one`/`select_multiple` questions.

No.

### DIFF
--- a/collect_app/src/main/res/drawable/inset_check_box.xml
+++ b/collect_app/src/main/res/drawable/inset_check_box.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<inset xmlns:android="http://schemas.android.com/apk/res/android"
+    android:drawable="?android:attr/listChoiceIndicatorMultiple"
+    android:insetLeft="@dimen/margin_standard" />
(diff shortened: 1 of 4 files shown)


lognaturel (approved):
Thanks, @grzesiek2010. Too bad we can't also get the fix for 5 but I agree with your analysis.

kkrawczyk123:
**Tested with success!**

PR #3670, opened 2020-03-04
Open user-facing post with scoped storage migration details in browser

lognaturel:
The scoped storage text currently links to a developer-facing forum post. I want to make sure we simplify this for users and provide a dedicated thread for them to ask questions. I have added a post in the scheduled category that won't be visible to folks who aren't moderators. Here is a screenshot that includes the URL:
Screen Shot 2020-03-03 at 9 22 53 PM

This PR also changes the way the forum post is presented. I had originally specified launching the page in a browser but I didn't explain why. The reason is that folks may want to log into the forum and using a browser means they'll have access to password managers, for example. We also want to encourage them to browse the forum and that will be easier to do if the link opens in a browser that can stay open concurrently with Collect.

Ran tests and manually verified that the correct link was opened in a browser.

I considered leaving the developer-facing post but I think the content is not right for users.

This should just change the explanatory page that's opened and what browser it's opened in. I can't think of a regression risk.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/storage/StorageMigrationTest.java
@@ -121,7 +121,7 @@ public class StorageMigrationTest {
         new MainMenuPage(main)
                 .clickLearnMoreButton()
                 .clickMoreDetails()
-                .assertWebViewOpen();
+                .assertForumPostOpen();
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


grzesiek2010 (approved):

grzesiek2010:
@mmarciniak90 @kkrawczyk123 this should be quick to verify just test the `More details` button.

kkrawczyk123:
@lognaturel when I open the link I see: 
Screenshot_20200304-113820
I am able to log in to the forum using Log in button. The "Oops!" message is a little bit misleading but I understand that it is just a temporary state?  
When I log to the forum and click on More details again I can see:
Screenshot_20200304-114813
I can't see my photo in the right corner on that page or sign up/log in buttons. When I open one of the threads I can see my photo so I am logged in - I do not completely understand why it works like that. I obtained the same results for Chome and Firefox so it is not connected to browsers. 
Please just let me know if it's the expected behavior.

grzesiek2010:
> Please just let me know if it's the expected behavior.

It is since as mentioned above:

> I have added a post in the scheduled category that won't be visible to folks who aren't moderators. Here is a screenshot that includes the URL:

PR #3672, opened 2020-03-04
Run tests in parallel on 50 separate devices

grzesiek2010:
Nothing.

I use this option manually running test on Firebase Test Lab and it works much faster.
https://cloud.google.com/sdk/gcloud/reference/beta/firebase/test/android/run#--num-uniform-shards

> Specifies the number of shards into which you want to evenly distribute test cases. The shards are run in parallel on separate devices. For example, if your test execution contains 20 test cases and you specify four shards, each shard executes five test cases.
> The number of shards should be less than the total number of test cases. The number of shards specified must be >= 1 and <= 50.

It doesn't require testing.

No.

No.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,6 +172,7 @@ jobs:
             if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" ]]; then \
               echo "y" | gcloud firebase test android run \
               --type instrumentation \
+              --num-uniform-shards=50 \
               --app collect_app/build/outputs/apk/debug/*.apk \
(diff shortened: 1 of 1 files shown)


seadowg:
Oh wow great find. I'll give this a try myself as a review!

seadowg:
@lognaturel are there any cost reasons that might block this approach?

lognaturel:
> are there any cost reasons that might block this approach?

I'm not sure. Google's billing is notoriously opaque. Our Firebase TestLab costs over the past couple of days have been much higher than usual ($10/day vs. $3/month) but it might just be because we had a lot of runs to do. Let's try it and monitor?

lognaturel (approved):

PR #3675, opened 2020-03-04
Fix flakey tests

seadowg:
This moves one of the tests in `StorageMigrationCompletedBannerTest` to Robolectric and introduces an `AsyncTask` where a `Thread` was being used. For the former this stabilizes the test as for some reason recreating an `Activity` in Espresso appears to be flakey. For the latter this means that Espresso is in sync with the background work (it only knows about `AsyncTask`).

Ran locally and on Test Lab.

I did consider using standard Robolectric for the `MainMenuActivityTest` but the `ActivityScenario` felt right as it lets us worry less about what exact lifecycle methods get called in what scenario (versus the `ActivityController`) which is pretty nice. I like this blend of `androidx` lifecycle control but not using `onView` etc.

Should really only effect tests. The changes to reset feature feel pretty covered by the test that was flakey so I think I'd be ok merging this without QA.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/activities/MainMenuActivityTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/activities/MainMenuActivityTest.java
@@ -4,6 +4,9 @@ import android.content.Intent;
 import android.view.MenuItem;
 import android.view.View;
 
+import androidx.test.core.app.ActivityScenario;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -13,10 +16,7 @@ import org.odk.collect.android.storage.StorageStateProvider;
 import org.odk.collect.android.storage.migration.StorageMigrationRepository;
 import org.odk.collect.android.storage.migration.StorageMigrationResult;
 import org.odk.collect.android.support.RobolectricHelpers;
-import org.robolectric.Robolectric;
-import org.robolectric.RobolectricTestRunner;
 import org.robolectric.RuntimeEnvironment;
-import org.robolectric.android.controller.ActivityController;
 import org.robolectric.annotation.LooperMode;
 import org.robolectric.fakes.RoboMenuItem;
 import org.robolectric.shadows.ShadowEnvironment;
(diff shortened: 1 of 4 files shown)


collect_app/src/test/java/org/odk/collect/android/activities/MainMenuActivityTest.java
```diff
@@ -26,12 +26,13 @@
...
 
 import static android.os.Environment.MEDIA_MOUNTED;
+import static androidx.lifecycle.Lifecycle.State.DESTROYED;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.robolectric.Shadows.shadowOf;
 import static org.robolectric.annotation.LooperMode.Mode.PAUSED;
 
-@RunWith(RobolectricTestRunner.class)
+@RunWith(AndroidJUnit4.class)
```
lognaturel, line 35: So is this a no-op or am I missing something?

seadowg: I think so. The docs seem to make a distinction http://robolectric.org/androidx_test/. I guess again for me I find it useful to know that this is an "Android X" test. Do you think it's confusing?

lognaturel: It's not a big deal but I'd like to have a sense of when I should explicitly specify a runner or not.

lognaturel (approved):
I agree with the analysis that this is safe to merge without QA. FYI @opendatakit/testers if you want to take a look.

collect_app/src/test/java/org/odk/collect/android/activities/MainMenuActivityTest.java
seadowg: I guess I think for these ones there should always be a runner to show that we're not just using JUnit. It's a little wierd because anything in `androidTest` automatically uses this runner but that's a really strange concept for anyone used to JUnit (not even Spring does that).

lognaturel: To summarize: for tests in `test` that don't use Robolectric, then it's just JUnit and not specifying an explicit runner makes sense. But if they're in `androidTest` and use the androidx `AndroidJUnit4` runner, we should make that explicit even though it's the default because it's a weird default?

seadowg: My gut tells me that's the right thing yeah.

lognaturel: I appreciate your gut for sharing a clear rule to follow. 👍

PR #3679, opened 2020-03-05
Run tests in parallel on 50 separate devices

### LINKED ISSUE #3578: Look at parallelizing instrumentation tests

grzesiek2010:
Closes #3578 

We decided to revert #3672 because it turned out that that `--num-uniform-shards` requires `gcloud alpha firebase test android run` so since we are just about to release a new version of the app it's important to have the build green and we can play with this improvement later.

Nothing.

I use this option manually running test on Firebase Test Lab and it works much faster.
https://cloud.google.com/sdk/gcloud/reference/beta/firebase/test/android/run#--num-uniform-shards

> Specifies the number of shards into which you want to evenly distribute test cases. The shards are run in parallel on separate devices. For example, if your test execution contains 20 test cases and you specify four shards, each shard executes five test cases.
> The number of shards should be less than the total number of test cases. The number of shards specified must be >= 1 and <= 50.

It doesn't require testing.

No.

No.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,8 +170,9 @@ jobs:
           name: Run integration tests
           command: |
             if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" ]]; then \
-              echo "y" | gcloud firebase test android run \
+              echo "y" | gcloud alpha firebase test android run \
               --type instrumentation \
+              --num-uniform-shards=50 \
               --app collect_app/build/outputs/apk/debug/*.apk \
               --test collect_app/build/outputs/apk/androidTest/debug/*.apk \
               --device model=Pixel2,version=27,locale=en,orientation=portrait \
--- a/collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/database/helpers/InstancesDatabaseHelperTest.java
@@ -4,6 +4,7 @@ import android.database.sqlite.SQLiteDatabase;
 
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
@@ -52,7 +53,7 @@ public class InstancesDatabaseHelperTest extends SqlLiteHelperTest {
 
     @Parameterized.Parameters(name = "{0}")
     public static Iterable<Object[]> data() {
-        return Arrays.asList(new Object[][] {
+        return Arrays.asList(new Object[][]{
                 {"Downgrading from version with extra column drops that column", "instances_v7000_added_fakeColumn.db"},
                 {"Downgrading from version with missing column adds that column", "instances_v7000_removed_jrVersion.db"},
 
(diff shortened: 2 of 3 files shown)


seadowg:
I gave this a go and found we're getting 3 failures: https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/8628208877531572155.

I was initially only getting one failure in `FormDatabaseHelperTest`. It looked like it didn't have permissions or our reset rule so I added those to `SqliteHelperTest`. This annoyingly caused `InstancesDatabaseHelperTest` to fail. I think we should seperately added fixes for dodgy tests (as other PRs) until we can get this green. Hopefully only a couple of tweaks required!

seadowg:
I keep getting this failure: https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/4733938542694164299/executions/bs.1827bf05fa566136/testcases/2/test-cases. Can't work out a way to reproduce locally. @lognaturel / @grzesiek2010 any ideas around what could be happening?

seadowg (dismissed):
Sticking this into "Request changes" until it's working

grzesiek2010:
I tried today but have no idea why that one test is failing and how multiple shards might impact it... maybe we could comment it out temporarily and merge this pr to check how costs are going to change because if it's to expensive for us it also doesn't make sense to spend time on failing test?

lognaturel:
That sounds reasonable to me. We can add an issue to bring back that coverage in some way if deemed necessary so we don't forget. @seadowg? And I think we could merge this one before the repeat add PR.

shobhitagarwal1612:
I was wondering if it's possible to run tests incrementally. What if gradle only ran tests for the classes that were compiled? I'm certain gradle build is incremental but not the test task. Maybe we can create one? Has anyone seen that before?

seadowg:
@grzesiek2010 good idea. That's probably a test we can move to Robolectric anyway. I'll try ignoring it and running a few times.

@shobhitagarwal1612 we're more focused on our full suite run here but being able to run subsets of tests locally is pretty amazing. Reorganizing tests into domain slices so it's clearer what to run and also breaking the code up into modules so the build time is faster for individual modules are things we're looking at.

seadowg:
@grzesiek2010 I'm still seeing the occasional test flake but I think it'll be easier to fix those as we see them come up as part of our actual workflow. If we are really stumped it's an easy revert. Merging this feels like a good step forward so happy for you to go ahead when you the checks pass!

grzesiek2010:
Ok so let's merge. I can't do that since it's my pr.

lognaturel (approved):

lognaturel:
I knew there was a risk this would break the build but it did so in a very unexpected way: “ (gcloud.alpha.firebase.test.android.run) unrecognized arguments: --num-uniform-shards=50”

seadowg:
I'm guessing from the error we probably need to update the GCloud tools we use in Circle CI. It looks like we use the `cloud-sdk` Docker image. I'll make a quick PR that updates that to the latest.

PR #3680, opened 2020-03-05
Use one app bar layout across the app

grzesiek2010:
I reviewed the screens that I updated.

It's a refactoring that allows removing some duplicated code. Now we have one toolbar layout we can use across the app.
It also makes the layouts consistent because without this pr we use different shadows for example in different screens.

It requires reviewing all screens to confirm that there is everything fine with their toolbars.

No.

No.


collect_app/src/main/res/layout/form_chooser_list.xml
```diff
@@ -18,19 +18,13 @@ the License.
...
-            android:layout_height="wrap_content"
-            android:layout_below="@id/toolbar"/>
+    <include layout="@layout/app_bar_layout"/>
 
     <LinearLayout
             android:id="@+id/llParent"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:layout_below="@id/toolbar"
+            android:layout_below="@id/appBarLayout"
```
seadowg, line 27: We don't actually have this written down anywhere but generally my opinion is that layout IDs should be snakecased (`app_bar_layout`). I've never actually been able to find any good arbitrary opinions on this. Thoughts @lognaturel?

seadowg (approved):
In addition to what @grzesiek2010 pointed out it's probably good to check this at our lowest Android levels just to make sure nothing weird happens.

seadowg:
@grzesiek2010 needs a conflict fix before moving to `needs testing` I think!

mmarciniak90:
@grzesiek2010 I noticed that there are no changes on the `About` page. Could you look at this?

|Option | without fix | with fix|
|---|---|---|
| Admin Settings - unnecessary shadow removed | Screenshot_2020-03-23-09-27-41-364_org odk collect android|Screenshot_2020-03-23-08-47-07-414_org odk collect android|
|Form management - unnecessary shadow removed| Screenshot_2020-03-23-09-27-59-269_org odk collect android|Screenshot_20200323-105420|
|About - shadow is still visible| Screenshot_2020-03-23-09-27-27-658_org odk collect android|Screenshot_2020-03-23-08-46-38-714_org odk collect android|

grzesiek2010:
There is a change I use the same app bar which contains thinner shadow (difficult to notice in dark mode). The margin you probably are talking about is not related to my changes in this case because it exists in about_layout.xml file but ok I'll fix it as well since it's one line change.

mmarciniak90:
Thanks, @grzesiek2010!

**Tested with success**

Verified cases:
* headers
* shadows under the header
* light/dark mode

@kkrawczyk123 test this on Android 5.1 and 10.0. Layout issues can be Android and screen size sensitive so it's worth to check

kkrawczyk123:
There is a slightly visible shadow on Android 5 but I am not sure if fixing it is worth the effort:
Screenshot_2020-03-24-09-53-24
so still verified with success for me!

PR #3685, opened 2020-03-05
Fix version parsing for beta releases and dirty releases

lognaturel:
Run automated tests, manually tried `v1.26.0-beta.1` and `v1.26.0-beta.1-dirty` tags.

I think this is now pretty difficult to understand. I considered restructuring the code so that the beta tag suffix is dealt with before any of the rest of the decision making. I also considered using a regular expression to identified the different components of the version string. Ultimately, I figured that it would be easier to review a simpler diff for now since my goal is to get a beta out today.

I also considered catching exceptions and falling back in case we change our tag structure and/or a fork wants to use different tags but I figure that can be dealt with when it comes up. It is a little strange that a tag can now prevent app launch.

I also considered allowing beta numbers above 9 or documenting that with a test because it's also an unexpected limitation but again, I figured it was safer to keep the diff small for now.

It prevents the application from crashing on launch on a version tagged with our tagging scheme. There's a chance I didn't think of all missing cases and that there are still possible crashes.

No.

I have added a line to the README to document the mandatory tag structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,8 @@ RELEASE_KEY_PASSWORD=secure-alias-password
 
 Maintainers also have a `google-services.json` file in the `collect_app/src/odkCollectRelease` folder. The contents of the file are similar to the contents of `collect_app/src/google-services.json`.
 
+When ready to generate a build for the Play Store, maintainers tag the build by [adding a release](releases). Tags for full releases must have the format `vX.X.X`. Tags for beta releases must have the format `vX.X.X-beta.X`.
+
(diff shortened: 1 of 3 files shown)


yanokwa (approved):

lognaturel:
@seadowg I wanted to get a quick patch in but what do you think about changing this to use a regex like `-beta\.(\d+)` to get the beta number, replacing that with an empty string, and then doing a split on `-` for what remains and if there are two items, putting the second one below the buttons? That seems like it'd be easier to think about. I also think that no matter what the tag structure is, there should never be a hard crash.

seadowg:
I'm totally flexible on it. Using a regex would definitely be better for the beta number and I'm indifferent to what goes at the bottom as I personally never use it. Including everything after the `-` is definitely a lot simpler!

PR #3690, opened 2020-03-06
Fixed Save button in Reason for changes page does not work after rotating device orientation

### LINKED ISSUE #3689: Save button in Reason for changes page does not work after rotating device orientation
Collect v1.26.0 beta
Works correctly on v1.25.2

#### Problem description
Save button in Reason for changes page does not work after rotating device orientation

1. User fills the form [track-changes-reason-on-edit.xml.txt](https://github.com/opendatakit/collect/files/4298052/track-changes-reason-on-edit.xml.txt)
2. User goes to `Edit Saved Form `
3. User edits an answer, go to `Save and Exit` and click on it.
4. User inputs a reason
5. User rotates the screen
6. User clicks on Save button
7. **User is moved to view with the `Save and Exit` button, instead of to the main menu**

#### Expected behavior
The form is saved and user is moved to the main menu

grzesiek2010:
Closes #3689

I tested the changes manually and improved one of the automated tests.

It's the only solution. The previous one was wrong because once `onDestroy()` was called the observation was killed, now it's resumed.

This change should just fix the problem and that's all. It shouldn't affect anything else. I think the risk here is not big. Testing the attached form seems enough.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
@@ -123,7 +123,9 @@ public class TrackChangesReasonTest {
                 .clickSaveAndExitWithChangesReasonPrompt()
                 .enterReason("Something")
                 .rotateToLandscape(new ChangesReasonPromptPage("Track Changes Reason", rule))
-                .assertText("Something");
+                .assertText("Something")
+                .rotateToPortrait(new ChangesReasonPromptPage("Track Changes Reason", rule))
+                .clickSave();
     }
 
     @Test
--- a/collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
@@ -124,7 +124,7 @@ public class TrackChangesReasonTest {
                 .enterReason("Something")
                 .rotateToLandscape(new ChangesReasonPromptPage("Track Changes Reason", rule))
                 .assertText("Something")
-                .rotateToPortrait(new ChangesReasonPromptPage("Track Changes Reason", rule))
+                .closeSoftKeyboard()
                 .clickSave();
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1868,7 +1868,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
                     finishAndReturnInstance();
                 }
-
+                formSaveViewModel.consumeSavedResult();
                 break;
 
             case SAVE_ERROR:
@@ -1883,6 +1883,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 }
 
                 showLongToast(message);
+                formSaveViewModel.consumeSavedResult();
                 break;
 
             case FINALIZE_ERROR:
@@ -1890,6 +1891,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 showLongToast(String.format(getString(R.string.encryption_error_message),
                         result.getMessage()));
                 finishAndReturnInstance();
+                formSaveViewModel.consumeSavedResult();
                 break;
 
             case CONSTRAINT_ERROR: {
@@ -1909,11 +1911,10 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     // otherwise, we can get the proper toast(s) by saving with constraint check
                     saveAnswersForCurrentScreen(EVALUATE_CONSTRAINTS);
                 }
-
+                formSaveViewModel.consumeSavedResult();
                 break;
             }
         }
-        formSaveViewModel.consumeSavedResult();
     }
 
     /**
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -425,6 +425,8 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         formSaveViewModel = ViewModelProviders
                 .of(this, changesReasonViewModelFactory)
                 .get(FormSaveViewModel.class);
+
+        formSaveViewModel.getSavedResult().observe(this, this::handleSaveResult);
     }
 
     private void setupFields(Bundle savedInstanceState) {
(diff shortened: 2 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1821,91 +1823,97 @@ private boolean saveForm(boolean exit, boolean complete, String updatedSaveName,
...
+                    saveAnswersForCurrentScreen(EVALUATE_CONSTRAINTS);
                 }
-            }
-        });
 
-        return true;
+                break;
+            }
+        }
+        formSaveViewModel.consumeSavedResult();
```
seadowg: I think we should call this `resumeFormEntry` seeing as that's the "view action" that's occuring. I'm also thinking this should only really happen for results that end the same process (like `SAVED` and the errors) but not for `SAVING` or `CHANGE_REASON_REQUIRED`.

seadowg:
We could actually add test coverage to do this by adding these two lines to the bottom of `openingAFormToEdit_andChangingAValue_andClickingSaveAndExit_andRotating_remainsOnPrompt`:

```java
.closeSoftkeyboard()
.clickSave();
```

If you add that in without your changes you should see that test go red and then your fix will hopefully make it go green!

collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
```diff
@@ -123,7 +123,9 @@ public void openingAFormToEdit_andChangingAValue_andClickingSaveAndExit_andRotat
                 .clickSaveAndExitWithChangesReasonPrompt()
                 .enterReason("Something")
                 .rotateToLandscape(new ChangesReasonPromptPage("Track Changes Reason", rule))
-                .assertText("Something");
+                .assertText("Something")
+                .rotateToPortrait(new ChangesReasonPromptPage("Track Changes Reason", rule))
```
seadowg: I think this should be `closeSoftkeyboard()` instead just so we are checking the full "rotated" flow.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -68,38 +68,44 @@ public void setup() {
...
+        assertThat(saveResult1.getState(), equalTo(SAVING));
 
-        LiveData<FormSaveViewModel.SaveResult> saveResult2 = viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        FormSaveViewModel.SaveResult saveResult2 = viewModel.getSavedResult().getValue();
         assertThat(saveResult2, not(equalTo(saveResult1)));
     }
 
     @Test
     public void saveForm_wontRunMultipleSavesAtOnce() {
```
seadowg, line 76: I don't know if this makes sense anymore? Maybe it's best just to return when we're already saving rather than setting the result to `ALREADY_SAVING` just seems a little weird for the same `SaveResult` LiveData to go from the `SAVING` to `ALREADY_SAVING` state.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -68,38 +68,44 @@ public void setup() {
 
     @Test
     public void saveForm_returnsNewSaveResult_inSavingState() {
-        LiveData<FormSaveViewModel.SaveResult> saveResult1 = viewModel.saveForm(Uri.parse("file://form"), true, "", false);
```
seadowg, line 71: Yeah I think this test also needs to change as the `LiveData` will be the same between calls. Should just be that `saveForm_returnsSaveResult_inSavingState` and it only needs to make one call to `save`.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -68,38 +68,44 @@ public void setup() {
...
-        assertThat(saveResult2.getValue().getState(), equalTo(ALREADY_SAVING));
     }
 
     @Test
     public void saveForm_whenReasonRequiredToSave_returnsSaveResult_inChangeReasonRequiredState() {
         whenReasonRequiredToSave();
 
-        LiveData<FormSaveViewModel.SaveResult> saveResult = viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        LiveData<FormSaveViewModel.SaveResult> saveResult = viewModel.getSavedResult();
```
seadowg, line 91: I think having the `saveResult` declarations before the call to `save` would be good because it ensures that the same `LiveData` instance is being used before/after the call. This mimics how you would use the `ViewModel` where you make an observation before you call an action.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Right, fixed.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -181,6 +179,14 @@ private void handleTaskResult(SaveToDiskResult taskResult, SaveRequest saveReque
         }
     }
 
+    public LiveData<SaveResult> getSavedResult() {
+        return saveResult;
+    }
+
+    public void resumeFormEntry() {
+        saveResult = new MutableLiveData<>(null);
```
seadowg: In my head this needs to be `saveResult.setValue(null)` instead. Otherwise saving multiple times won't work for observations made in `onCreate` of an Activity/Fragment as they'll be observing a different `LiveData` value. Maybe I'm reading this wrong!

We can verify with a test for the `ViewModel`:

```java
@Test
public void resumeFormEntry_clearsSaveResult() {
    LiveData<FormSaveViewModel.SaveResult> saveResult = viewModel.getSavedResult();
    viewModel.saveForm(Uri.parse("file://form"), true, "", false);
    viewModel.resumeFormEntry();
    assertThat(saveResult.getValue(), equalTo(null));
}
```

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/formentry/audit/TrackChangesReasonTest.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -67,39 +66,36 @@ public void setup() {
...
+        viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        FormSaveViewModel.SaveResult saveResult1 = viewModel.getSavedResult().getValue();
+        assertThat(saveResult1.getState(), equalTo(SAVING));
 
-        whenFormSaverFinishes(SaveFormToDisk.SAVED);
-        assertThat(saveResult1.getValue().getState(), equalTo(SAVED));
-        assertThat(saveResult2.getValue().getState(), equalTo(ALREADY_SAVING));
+        viewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        FormSaveViewModel.SaveResult saveResult2 = viewModel.getSavedResult().getValue();
+        assertThat(saveResult2.getState(), equalTo(ALREADY_SAVING));
```
seadowg: Sorry I don't think what I meant was clear here: I think we should get rid of the `ALREADY_SAVING` state entirely. It's unlikely that it would happen but if it did and then the user rotated the `LiveData` would be in the `ALREADY_SAVING` state rather than `SAVING` when the observation is setup again.

I think we can just return in `saveForm` if a save is in progress.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -344,6 +346,14 @@ public void saveReason_whenReasonIsNotValid_returnsFalse() {
         assertThat(viewModel.saveReason(), equalTo(false));
     }
 
+    @Test
+    public void resumeFormEntry_clearsSaveResult() {
```
seadowg, line 351: 👍

seadowg (changes requested):
Just one wee change I think we need but looks great!

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
grzesiek2010: So you mean to remove the last three lines at all? Then the test would be equal to the one above.

seadowg: Yeah I don't think we need this test because my feeling is that we should remove the line that this drives out in `FormSaveViewModel` (`saveResult.setValue(new SaveResult(SaveResult.State.ALREADY_SAVING, null))`). Really we should change it so that the test checks we don't call `formSaver#save` again - thinking about it the original test should really have had that check as well.

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android: 6.0, 7.0, 8.1, 9.0, 10.0

Verified cases:
- track-changes-reason-on-edit.xml form
- rotating device

PR #3698, opened 2020-03-09
Rebuild the mappable instance list when the map view is shown to reflect deleted geopoint

### LINKED ISSUE #3696: Removed Geopoint does not disappear from the MapActivity view immediately
Collect v1.26.0 beta

#### Problem description
Removed Geopoint does not disappear from the MapActivity view immediately

1. User opens new MapActivity
2. User sees marker for each form instance
3. User clicks on some marker
4. User removes response for Geopoint widget
5. User leaves the form and saves it
6. User is moved to MapActivity
7. **Removed point is visible**
8. User leaves map and opens it again -> point is not visible

#### Expected behavior
* When response for GeoPoint is removed, point disappears from the map
* Bottom bar shows the correct number of points and filled submissions

map

@opendatakit-bot label "regression"

lognaturel:
Closes #3696

Manually verified case from issue. Added and ran automated test.

I went to the trouble of commenting that we could skip rebuilding the mappable instance list unless the instance count changed but that is not accurate. When geometry is removed, the instance count is the same. I considered trying to do something more clever like keeping track of which instance was opened since that's the only one that could change but I don't think it's worth it for now given the complexity it would add. If folks have a lot of submissions, this code will become a performance problem, I think, since it has to do some json parsing for each point. We can revisit it if we have evidence of users having problems.

This now rebuilds the mappable instance list each time the form map activity is displayed. The intentional change is fixing the bug. There's also likely a performance hit for very large numbers of submissions. Risk is only in the map view.

Any form with a geopoint such as All Widgets.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
@@ -133,6 +133,30 @@ public class FormMapViewModelTest {
         assertThat(instances.get(6).getClickAction(), is(FormMapViewModel.ClickAction.OPEN_EDIT));
     }
 
+    @Test public void clearingInstanceGeometry_isReflectedInInstanceCountsAndList() {
+        FormMapViewModel viewModel = new FormMapViewModel(TEST_FORM_1, testInstancesRepository);
+
+        List<FormMapViewModel.MappableFormInstance> mappableInstances = viewModel.getMappableFormInstances();
+        assertThat(viewModel.getTotalInstanceCount(), is(7));
+        assertThat(mappableInstances.size(), is(6));
+
+        assertThat(mappableInstances.get(5).getClickAction(), is(FormMapViewModel.ClickAction.NOT_VIEWABLE_TOAST));
+        ((TestInstancesRepository) testInstancesRepository).removeInstanceById(6L);
+
+        ((TestInstancesRepository) testInstancesRepository).addInstance(new Instance.Builder().databaseId(6L)
+                .jrFormId("formId1")
+                .jrVersion("2019103101")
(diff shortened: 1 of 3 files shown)


collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
```diff
@@ -133,6 +133,30 @@
         assertThat(instances.get(6).getClickAction(), is(FormMapViewModel.ClickAction.OPEN_EDIT));
     }
 
+    @Test public void clearingInstanceGeometry_isReflectedInInstanceCountsAndList() {
```
seadowg, line 136: I'm wondering if this test could be made a little simpler if it was just `getMappableFormInstances_alwaysReturnsUpToDateInstances()`. If the `TestInstancesRepository` had a method that let you switch out the list of instances you could change instances in the repo that way and then check `getMappableFormInstances`

seadowg (approved):

collect_app/src/test/java/org/odk/collect/android/activities/FormMapViewModelTest.java
lognaturel: Ah yes, I am far from mastering the art of writing as narrow of a test as possible. I tend to go with what matches the user experience where possible. I unfortunately won’t get a chance to come back to this in the next couple of days but if someone wants to push that change to my branch they are welcome to.

mmarciniak90:
**Tested with success**

Tested on Android: 5.1, 6.0, 7.0, 8.1, 9.0, 10.0

Verified cases:
- removing a point from finished submission
- removing a point from saved submission
- removing a point from submission with `sanding failed` status
- verified number of points

PR #3701, opened 2020-03-10
Disable Markup button if default image couldn't be loaded

### LINKED ISSUE #3634: Markup Image button should be disabled when there is no media or path to default image is invalid.
Collect v1.26x, Android v4-10, device used...

#### Problem description
For Annotate widget with default image described Markup Image button should be enabled only when path to default image is described in a correct way. Currently, when delete media and open form Markup button is enabled. The same issue is visible when the default image path is described in incorrect way. Screenshots attached:
| invalid | media deleted |
| -------- | -------- |
Screenshot from 2020-02-11 11-19-30 | Screenshot from 2020-02-11 11-19-12 

Just open Annotate widget with default image described in the wrong way or delete media. I have used form: 
[defaultImage.zip](https://github.com/opendatakit/collect/files/4185248/defaultImage.zip)

#### Expected behavior
Markup Image button should be active only when the picture is added/ path to default image is described correctly.

chidauri:
Closes #3634 

Tested on android 9 device

This is a simple solution. Another option might be to call `setUpLayout()` after `addCurrentImageToLayout()` .

None.

The form mentioned in the issue.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
@@ -65,7 +65,7 @@ public class AnnotateWidget extends BaseImageWidget {
         imageCaptureHandler = new ImageCaptureHandler();
         setUpLayout();
         addCurrentImageToLayout();
-        if (errorLoadingImage) {
+        if (binaryName == null || imageView == null || imageView.getVisibility() == GONE) {
             annotateButton.setEnabled(false);
         }
         addAnswerView(answerLayout, WidgetViewUtils.getStandardMargin(context));
@@ -79,9 +79,7 @@ public class AnnotateWidget extends BaseImageWidget {
         chooseButton = createSimpleButton(getContext(), R.id.choose_image, getFormEntryPrompt().isReadOnly(), getContext().getString(R.string.choose_image), getAnswerFontSize(), this);
 
         annotateButton = createSimpleButton(getContext(), R.id.markup_image, getFormEntryPrompt().isReadOnly(), getContext().getString(R.string.markup_image), getAnswerFontSize(), this);
-        if (binaryName == null) {
-            annotateButton.setEnabled(false);
-        }
+
         annotateButton.setOnClickListener(v -> imageClickHandler.clickImage("annotateButton"));
 
         answerLayout.addView(captureButton);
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
```diff
@@ -65,6 +65,9 @@ public AnnotateWidget(Context context, QuestionDetails prompt) {
         imageCaptureHandler = new ImageCaptureHandler();
         setUpLayout();
         addCurrentImageToLayout();
+        if (errorLoadingImage) {
```
grzesiek2010: I think you don't need this at all... what about a check like?
```
        if (binaryName == null || imageView == null || imageView.getVisibility() == GONE) {
            annotateButton.setEnabled(false);
        }
```

there is a similar check:
```
if (binaryName == null) {
            annotateButton.setEnabled(false);
        }
```

in `setUpLayout()` method so you could remove it there.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
chidauri: Done.

collect_app/src/main/java/org/odk/collect/android/widgets/AnnotateWidget.java
```diff
@@ -65,6 +65,9 @@ public AnnotateWidget(Context context, QuestionDetails prompt) {
         imageCaptureHandler = new ImageCaptureHandler();
         setUpLayout();
         addCurrentImageToLayout();
+        if (binaryName == null || imageView == null || imageView.getVisibility() == GONE) {
```
grzesiek2010: It would be good to factor it out to a separate method like adjustAnnotateButtonaAailability(0 for example.

grzesiek2010 (changes requested):
It looks good just one possible improvement I mentioned.
Could you also add an automated test? Ideally in `AnnotateWidgetTest` class if possible. If not it can be also an espresso test.

chidauri:
@grzesiek2010 does the test look good? also sorry for so much delay.

grzesiek2010 (approved):

grzesiek2010:
> @grzesiek2010 does the test look good? also sorry for so much delay.

yes it looks great now. Thanks!

mmarciniak90:
**Tested with success**

Verified on an emulator with Android 5.1.
`Markup Image` button is inactive

Verified cases:
* form with default image described in the wrong way
* form with a correct default image
* Adding/ removing response

PR #3702, opened 2020-03-11
Fixed canceling form loading

### LINKED ISSUE #3638: User is able to leave form by clicking on header immediately when the form is loaded
Collect v1.26.0 beta.0

#### Problem description
User is able to leave form by clicking on header immediately when the form is loaded

1. User goes to `Fill Blank Form`
2. User clicks on some form
3. User immediately clicks on the header (with icons save / jump to )
4. User is moved to main menu

#### Expected behavior
Clicking on the header does not cause leaving form

grzesiek2010:
Closes #3638

I tested canceling form loading and confirmed that now it's possible only by clicking on the `cancel` button.

The problem was introduced in https://github.com/opendatakit/collect/pull/3592/ by mistake I guess.
Prior to that pr we were able to cancel form loading only by clicking the `cancel` button. The pr changed it and now we are able to do that clicking the area out of the dialog as well what might be misleading. It's not a problem but I think if there is a button it should be responsible for that.

It's a regression but I don't think it's that important to block the release and add it to v1.26.

The risk is low here. Testing canceling form loading would be enough. A big form and maybe an older device would help here because it will give more time to cancel the loading process.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/FormLoadingDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/FormLoadingDialogFragment.java
@@ -45,6 +45,7 @@ public class FormLoadingDialogFragment extends ProgressDialogFragment {
 
         setTitle(getString(R.string.loading_form));
         setMessage(getString(R.string.please_wait));
+        setCancelable(false);
 
         if (context instanceof FormLoadingDialogFragmentListener) {
             listener = (FormLoadingDialogFragmentListener) context;
--- a/collect_app/src/test/java/org/odk/collect/android/formentry/FormLoadingDialogFragmentTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formentry/FormLoadingDialogFragmentTest.java
@@ -1,0 +1,34 @@
+package org.odk.collect.android.formentry;
+
+import androidx.fragment.app.FragmentActivity;
+import androidx.fragment.app.FragmentManager;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.Robolectric;
+import org.robolectric.RobolectricTestRunner;
+
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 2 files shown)


seadowg (approved):

seadowg:
I think even small change like this would be good to test (altough not worth not accepting the PR right now). We could call `setCancellable` in many different places in the `DialogFragment` (given all the lifecycle) and some would be wrong and others right. I'd probably write a test like this (before making the fix):

```java
public class FormLoadingDialogFragmentTest {

    private FragmentManager fragmentManager;

    @Before
    public void setup() {
        FragmentActivity activity = Robolectric.setupActivity(FragmentActivity.class);
        fragmentManager = activity.getSupportFragmentManager();
    }

    @Test
    public void dialogIsNotCancellable() {
        FormLoadingDialogFragment fragment = new FormLoadingDialogFragment();
        fragment.show(fragmentManager, "TAG");

        assertThat(fragment.isCancelable(), equalTo(true));
    }
}
```

This test would be red which lets us know our fix works for a "shown" fragment. Happy to push this in here or make another PR but wanted to make sure we discussed first! I'm sure there are other ways that we could create a red test.

grzesiek2010:
If it's a working test feel free to push your commit. I don't want to just copy your code.

mmarciniak90:
**Tested with success**

Verified cases:
- click many times on area
- click many times on header
- vertical / horizontal device orientation
- light / dark mode

seadowg:
@mmarciniak90 is this good to have `behaviour verified` on it now?

kkrawczyk123:
The same cases verified successfully on Android 5.1 and 10. 

@seadowg Now, it is!

seadowg:
@kkrawczyk123 ah sorry being impatient!

mmarciniak90:
@seadowg No worries, it could be confusing.
We work remotely and I execute tests on 4 devices, Kasia does the same on 2 other devices.
In the current situation, we needed to divide available devices between us.

seadowg:
@mmarciniak90 makes perfect sense. I'll make sure to pay attention to the "Verified on" section.

PR #3699, opened 2020-03-10
Factored out MultiClickGuard class

grzesiek2010:
#### What has been done to verify that this works as intended?
Just reviewed implemented changes.

#### Why is this the best possible solution? Were any other approaches considered?
We keep to much in the application class. This is just a refactoring that factors out the code responsible for blocking multi-clicks which is not related to the application class so should exist somewhere else.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
I would say it doesn't require testing. Reviewing the code would be enough.

#### Do we need any specific form for testing your changes? If so, please attach one.
No.

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
No.

#### Before submitting this PR, please make sure you have:
- [ ] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [ ] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [ ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #3719, opened 2020-03-17
Avoid multiclicking in Admin and General settings

### LINKED ISSUE #3713: `Admin Password` dialog can be opened two times
Collect v1.25
Older verions were not verified

#### Problem description
`Admin Password` dialog can be opened two times

1. User opens `Admin Settings`
2. User clicks multi times on `Admin password`
3. 'Admin password' dialog is opened two times

#### Expected behavior
The dialog is not opened two times

#### Other information 
Discovered during testing #3699 

EDIT
There could be more pages with mentioned issue
Noticed example:
1. `General Settings` -> `User and device identity` -> clicks multi times on `Form metadata`

grzesiek2010:
Closes #3713 

I built this pr on top of #3699 to test the newly factored `MultiClickGuard` class.

I tested the implemented changes manually.

We just need to use our MultiClickGuard class to avoid such bugs since android doesn't do that for us.

I didn't only fix the case described in the issue with the Admin Password dialog. I decided to add the check for all options in Admin Settings and even in General Settings. I was able for example to reproduce the same issue with General Settings option that is in Admin Settings so I guess the same is possible with other options as well so I decided to handle all of them just in case. Please pay attention to that option (General Settings from Admin Settings) since in order to handle it as well I needed to add some more refactoring.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
@@ -1,0 +1,27 @@
+package org.odk.collect.android.utilities;
+
+import android.os.SystemClock;
+
+public class MultiClickGuard {
+    private static final int CLICK_DEBOUNCE_MS = 1000;
+
+    private static long lastClickTime;
+    private static String lastClickName;
+
+    private MultiClickGuard() {
+    }
+
+    // Debounce multiple clicks within the same screen
+    public static boolean allowClick(String className) {
+        long elapsedRealtime = SystemClock.elapsedRealtime();
+        boolean isSameClass = className.equals(lastClickName);
+        boolean isBeyondThreshold = elapsedRealtime - lastClickTime > CLICK_DEBOUNCE_MS;
+        boolean isBeyondTestThreshold = lastClickTime == 0 || lastClickTime == elapsedRealtime; // just for tests
+        boolean allowClick = !isSameClass || isBeyondThreshold || isBeyondTestThreshold;
+        if (allowClick) {
+            lastClickTime = elapsedRealtime;
+            lastClickName = className;
+        }
+        return allowClick;
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/activities/AboutActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/AboutActivity.java
@@ -31,6 +31,7 @@ import android.widget.Toast;
 import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.AboutListAdapter;
 import org.odk.collect.android.application.Collect;
+import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.utilities.CustomTabHelper;
 
 import java.util.List;
@@ -85,7 +86,7 @@ public class AboutActivity extends CollectAbstractActivity implements
 
     @Override
     public void onClick(int position) {
-        if (Collect.allowClick(getClass().getName())) {
+        if (MultiClickGuard.allowClick(getClass().getName())) {
             switch (position) {
                 case 0:
                     websiteTabHelper.openUri(this, websiteUri);
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormChooserListActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormChooserListActivity.java
@@ -27,7 +27,6 @@ import android.widget.AdapterView;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.FormListAdapter;
-import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.dao.FormsDao;
 import org.odk.collect.android.listeners.DiskSyncListener;
 import org.odk.collect.android.listeners.PermissionListener;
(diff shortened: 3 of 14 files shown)


seadowg:
Two things I'm thinking about: for this whole set of changes:

1. I'm wondering if we could have the helper let us setup click listeners rather than us having to always check in the click listener. In my head it would be nice if we could do something like this:

    ```java
    MultiClickGuard.setOnClickListener(button, view -> {
        // handle click            
    });
    ```

2. I'm a little worried this could create (or might have already created) problems for tests where multiple clicks might happen very quickly. I think we can ease that worry easily by disabling it somehow in both Espresso and Robolectric.

seadowg (commented):
As above!

seadowg (changes requested):
Just sticking this into "Request changes" as leaving a "Comment" review apparently doesn't remove the "Review required" state annoyingly. Feel free to re-request review if you disagree with my comments!

grzesiek2010:
>'m wondering if we could have the helper let us setup click listeners rather than us having to always check in the click listener. In my head it would be nice if we could do something like this:
MultiClickGuard.setOnClickListener(button, view -> {
    // handle click            
});

I think it's not possible because we use MultiClickGuard in too many different contexts I mean:
- onClick();
- onPreferenceClick();
- clickImage();
- onItemClick();

In many cases we don't even set the listener but just implements and overrides like [here](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormChooserListActivity.java#L123) for example. So it would require a lot of changes and the code won't be perfect either way.

>I'm a little worried this could create (or might have already created) problems for tests where multiple clicks might happen very quickly. I think we can ease that worry easily by disabling it somehow in both Espresso and Robolectric.

I think it is disabled https://github.com/opendatakit/collect/pull/3719/files#diff-b4b1a1c24bed33b17e60ef5611e6b38dR19

collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private MultiClickGuard() {
+    }
+
+    // Debounce multiple clicks within the same screen
+    public static boolean allowClick(String className) {
+        long elapsedRealtime = SystemClock.elapsedRealtime();
+        boolean isSameClass = className.equals(lastClickName);
+        boolean isBeyondThreshold = elapsedRealtime - lastClickTime > CLICK_DEBOUNCE_MS;
+        boolean isBeyondTestThreshold = lastClickTime == 0 || lastClickTime == elapsedRealtime; // just for tests
```
seadowg, line 19: I'm not sure I understand this. Does this assume no time passes between clicks in tests? As far as I can see this is `true` if not time has passed since the last click. In both espresso and normal junit it's very possible for a millisecond to pass between clicks. Maybe I just need some more caffeine ☕️.

seadowg:
> I think it's not possible because we use MultiClickGuard in too many different contexts

That's a great point! I agree we just go with the way it is for now.

collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
grzesiek2010: Yes, someone has added it on purpose so I guess it fixed the problem he had run into.

grzesiek2010: Generally it shouldn't be a problem in which test a situation like that could happen that you need to click on the same element twice quickly?

seadowg: Is doesn't need to be the same element though right? It just needs to be in the same class (generally we're passing the Activity). Personlly the solution confuses me and I'm sure there are simpler ways to work out if we're in tests or not. I can look at later seeing as I have a personal vendetta against the line 😁. Let's just make sure to run `connectedAndroidTest` to check Espresso doesn't run into any problems with all the new guards.

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
grzesiek2010: > Is doesn't need to be the same element though right?

right.

kkrawczyk123:
@grzesiek2010 I was able to open twice Form Metadata in User and device identity but only just once on Android 7.1. @mmarciniak90, please keep it in mind while testing as you have currently more devices than I do. 

Verified cases:
- multiclick on all settings in admin settings 
- multiclick on all settings in general settings 
- light mode/dark mode

grzesiek2010:
Yeah actually I didn't dig that deep and handled just options from the main screens not their children... I'll take a look to check if it make sense.

mmarciniak90:
> I was able to open twice Form Metadata in User and device identity but only just once on Android 7.1

I had more luck because I was able to reproduce it on Android 6.0, 7.0, 8.1, 9.0

grzesiek2010:
> I was able to open twice Form Metadata in User and device identity

ok I fixed that case too but I think it doesn't make sense to handle every single option (it's especially if there is no one place where I could do that), so that's the last case I fixed if you find something else what doesn't cause any problems apart from the fact it opens something twice please ignore.

mmarciniak90:
**Tested with success**

Verified cases:
* multi-click on all settings in admin settings
* multi-click on all settings in general settings
* multi-click on Form Metadata in User and device identity
* light mode/dark mode

PR #3727, opened 2020-03-19
Update translations

lognaturel:
Translation update.

We typically don't update translations with point releases because if master has new features with translation strings it can get messy but since we kept master frozen there is no such risk.

Because of the repeat dialog change and the confusing storage migration I think it's more important this release to have up-to-date translations.


seadowg (approved):
Good to go as soon as the build passes!

PR #3728, opened 2020-03-20
Update dependencies

grzesiek2010:
Reviewed implemented changes, tested the app manually (OSM maps)

It's something we do after each release to keep the dependencies updated.

Usually we don't need those updates but this time I think it would be good to test OSM maps.

No.

No.


collect_app/build.gradle
```diff
@@ -262,15 +262,10 @@ dependencies {
     implementation "net.danlew:android.joda:2.10.3"
 
     // Real LeakCanary for debug builds only: notifications, analysis, etc
-    debugImplementation "com.squareup.leakcanary:leakcanary-android:${rootProject.leakyCanaryVersion}"
-    // No-Op version of LeakCanary for release builds: no notifications, no analysis, nothing
-    releaseImplementation "com.squareup.leakcanary:leakcanary-android-no-op:${rootProject.leakyCanaryVersion}"
-    testImplementation "com.squareup.leakcanary:leakcanary-android-no-op:${rootProject.leakyCanaryVersion}"
-    androidTestImplementation "com.squareup.leakcanary:leakcanary-android-no-op:${rootProject.leakyCanaryVersion}"
-    odkCollectReleaseImplementation "com.squareup.leakcanary:leakcanary-android-no-op:${rootProject.leakyCanaryVersion}"
```
grzesiek2010, line 270: According to https://square.github.io/leakcanary/upgrading-to-leakcanary-2.0/ we no longer need no-op artifacts

collect_app/src/main/java/org/odk/collect/android/application/TestCollect.java
```diff
@@ -12,12 +10,6 @@
 
 public class TestCollect extends Collect {
 
-    @Override
-    protected RefWatcher setupLeakCanary() {
```
seadowg, line 16: Huh so LeakCanary won't be running in any test build automatically now?

grzesiek2010:
@seadowg please take a quick look if you want and add `needs testing` label

seadowg (dismissed):

collect_app/src/main/java/org/odk/collect/android/application/TestCollect.java
grzesiek2010: That's what I understand https://square.github.io/leakcanary/recipes/#running-leakcanary-in-instrumentation-tests

seadowg: So much better!

seadowg:
Just removing "needs testing" and my review as I'm going to a push a fix for the gradle plugin problem.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,9 +1,8 @@
-#Wed May 08 10:57:16 CEST 2019
+#Tue Mar 24 09:13:30 GMT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
 android.enableD8.desugaring=true
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.3-all.zip
-# Android Studio Gradle wrapper >= 3.1.2 doesn't support configure on demand
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```
grzesiek2010: Why not the latest 6.2.2?

seadowg: Huh that was auto generated. Probably just game me the minimum for that plugin version. I'll update it.

seadowg: Wow this is causing all kinds of problems. Did you try and update spotbugs at all as part of this? It looks like we might be using a very old version.

grzesiek2010: ahh spotbugs... I think I was able to update it to version 2.0.0 but not the latest 4.0.4 because it was related somehow with gradle version

build.gradle
```diff
@@ -23,7 +23,6 @@ allprojects {
         jcenter()
         maven { url 'https://oss.sonatype.org/content/groups/public' }
         maven { url 'https://jitpack.io' }
-        maven { url 'https://staging.dev.medicmobile.org/_couch/maven-repo' }
```
seadowg, line 30: This repo link doesn't seem to be working any longer as they don't provide a `.pom` for the dependency. Given its a pretty stagnant dependency I think using a local jar is fine.

lognaturel (approved):

lognaturel:
Lots of good stuff unlocked by the minSDK change. 👍 Agreed that checking OSM would be good. Since there's now also a change to the native architecture stubs, it would be good to also quickly verify Mapbox. I'm going to see if I can also get https://github.com/opendatakit/collect/pull/3457 complete shortly so we can be fully up to date. That will also require Mapbox checking so really here it can just be to sanity check that it works on more than one device.

build.gradle
grzesiek2010: I think such a trick would work:
```
maven {
            url 'https://staging.dev.medicmobile.org/_couch/maven-repo'
            metadataSources { artifact() }
        }
```
and you won't need anything more.

seadowg: I was looking around for something like this as it really surprised me that there wasn't a way to get it working with just a JAR. Will give this a go tomorrow!

lognaturel: Great if we can track the Medic repo but not super important I don’t think.

grzesiek2010: Yes I was right. Just added a commit on my own, so that you don't need to play with it just take a look.

collect_app/build.gradle
```diff
@@ -194,12 +196,6 @@ configurations.all {
...
 dependencies {
-    // To make ODK Collect is installable on all devices even though the Mapbox
-    // native library is missing for some ABIs, we include an .aar dependency
-    // named "nbi-stubs", which causes there to be at least one .so library
-    // present for every ABI (armeabi, armeabi-v7a, arm64-v8a, x86, x86_64).
-    implementation fileTree(dir: 'libs', include: ['*.jar', '*.aar'])
-
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.browser:browser:1.2.0'
     implementation 'androidx.appcompat:appcompat:1.1.0-rc01' // Shouldn't be upgraded until https://issuetracker.google.com/issues/141869006 is fixed
```
grzesiek2010: It's marked as `won't fix`. We can leave it as if for now but in this case we can't block it for ever.

seadowg: We actully have our own issue for tracking this with notes on the possible workaround. I'll update this so it references that rather than Google's issue.

mmarciniak90:
Verified with success on Android: 6.0, 7.0, 8.1, 9.0

I didn't notice regression.
New design of zoom in / out buttons visible.


**Verified cases:**
* Carto, Stamen, USGS, OSM
* Mapbox
* OSM on view showing submissions on map
* Message at the bottom of the map
* Geopoint, GeoTrace, Geoshape
* 3 input methods 
* zoom in / out
* offline layer
* toast with not enough number of point to create Polyline/Polygon
* English/Spanish language

@kkrawczyk123 take a look at this PR, especially on Android 5.1

kkrawczyk123:
I haven't noticed any regression too. 

PR #3734, opened 2020-03-23
Revise `DialogUtils.showIfNotShowing` helper

seadowg:
This slightly revises our `DialogUtils.showIfNotShowing` helper based on some recent crashes related to it (fixed in another PR). The goal here was to make the helper a `void` method rather than returning a `DialogFragment`. This was becasue the returned dialog didn't make any sense in the case where a dialog wasn't actually shown (because the `FragmentManager` is in the wrong state). Changing it so the dialog instance isn't accessible to the caller has a couple of other advantages:

* It's stops there being any confusion around which dialog instance is actually shown
* It stops us having to create a new dialog instance for no reason

I've added a `getDialog` helper that returns the shown dialog for cases where we really need to call something on the dialog but as much as possible we should be using arguments (for static data the dialog needs), Dagger (for dependencies) or `ViewModel` (for non static data) to get things into fragments so as to avoid crashes or weirdness when they are recreated.

Added some new tests for cases that weren't covered and played around with some dialogs.

Shouldn't be any change in behaviour. I'd recommend testing:

* The storage migration flow (with and without admin password set)
* Form loading and saving

### DIFF
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -93,7 +93,7 @@ the specific language governing permissions and limitations under the License.
 
         <activity
             android:name=".activities.MainMenuActivity"
-            android:configChanges="locale|orientation|screenSize" />
+            android:configChanges="locale" />
         <activity
             android:name=".activities.ScannerWithFlashlightActivity"
             android:screenOrientation="portrait"
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -93,7 +93,7 @@ the specific language governing permissions and limitations under the License.
 
         <activity
             android:name=".activities.MainMenuActivity"
-            android:configChanges="locale" />
+            android:configChanges="locale|orientation|screenSize" />
         <activity
             android:name=".activities.ScannerWithFlashlightActivity"
             android:screenOrientation="portrait"
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
@@ -5,6 +5,8 @@ import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
 import androidx.lifecycle.ViewModel;
 
+import org.odk.collect.android.logic.FormController;
+
 import static org.odk.collect.android.utilities.StringUtils.isBlank;
 
 public class IdentityPromptViewModel extends ViewModel {
@@ -15,6 +17,7 @@ public class IdentityPromptViewModel extends ViewModel {
     @Nullable
     private AuditEventLogger auditEventLogger;
     private String identity = "";
+    private String formName;
 
     public IdentityPromptViewModel() {
         updateRequiresIdentity();
@@ -32,8 +35,9 @@ public class IdentityPromptViewModel extends ViewModel {
         return identity;
     }
 
-    public void setAuditEventLogger(AuditEventLogger auditEventLogger) {
-        this.auditEventLogger = auditEventLogger;
+    public void setFormController(FormController formController) {
+        this.formName = formController.getFormTitle();
+        this.auditEventLogger = formController.getAuditEventLogger();
         updateRequiresIdentity();
     }
 
@@ -58,5 +62,9 @@ public class IdentityPromptViewModel extends ViewModel {
 
     private static boolean userIsValid(String user) {
         return user != null && !user.isEmpty() && !isBlank(user);
+    }
+
+    public String getFormName() {
+        return formName;
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
@@ -64,7 +64,7 @@ public class IdentityPromptViewModel extends ViewModel {
         return user != null && !user.isEmpty() && !isBlank(user);
     }
 
-    public String getFormName() {
+    public String getFormTitle() {
         return formName;
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/DialogUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/DialogUtils.java
@@ -154,13 +154,18 @@ public final class DialogUtils {
         return alertDialog;
     }
 
-    public static <T extends DialogFragment> T showIfNotShowing(Class<T> dialogClass, FragmentManager fragmentManager) {
-        return showIfNotShowing(dialogClass, null, fragmentManager);
+    @Nullable
+    public static <T extends DialogFragment> T getDialog(Class<T> dialogClass, FragmentManager fragmentManager) {
+        return (T) fragmentManager.findFragmentByTag(dialogClass.getName());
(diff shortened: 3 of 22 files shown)


collect_app/src/main/AndroidManifest.xml
```diff
@@ -93,7 +93,7 @@ the specific language governing permissions and limitations under the License.
 
         <activity
             android:name=".activities.MainMenuActivity"
-            android:configChanges="locale|orientation|screenSize" />
+            android:configChanges="locale" />
```
lognaturel: Are you leaving `locale` because you think it's needed or just because it's not related to what you're doing here?

collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
```diff
@@ -59,4 +63,8 @@ private void updateRequiresIdentity() {
     private static boolean userIsValid(String user) {
         return user != null && !user.isEmpty() && !isBlank(user);
     }
+
+    public String getFormName() {
```
lognaturel: Super small thing but I'd recommend using `title` rather than `name`. I think `title` is very clearly from the `title` tag but `name` feels more ambiguous to me.

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
```diff
@@ -0,0 +1,108 @@
...
+            return true;
+        }
+
+        @Override
+        public String getAdminPassword() {
+            return "password";
+        }
+    }
+
+    private static class MockAdminPasswordDialogCallback extends FragmentActivity implements AdminPasswordDialogFragment.AdminPasswordDialogCallback  {
```
lognaturel: This feels more like a spy to me since it's keeping extra state and verifying it rather than holding an expectation of what it's supposed to receive. Not that important but would be good to be somewhat aligned. I'm referring to https://martinfowler.com/articles/mocksArentStubs.html.

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        activityScenario.recreate();
+        activityScenario.onActivity(activity -> {
+            AdminPasswordDialogFragment fragment = (AdminPasswordDialogFragment) activity.getSupportFragmentManager().findFragmentByTag("tag");
+
+            fragment.getInput().setText("password");
+            ((AlertDialog) fragment.getDialog()).getButton(AlertDialog.BUTTON_POSITIVE).performClick();
+            shadowOf(getMainLooper()).idle();
+
+            assertThat(activity.onCorrectAdminPasswordCalledWith, equalTo(STORAGE_MIGRATION));
```
lognaturel, line 92: I can't wrap my brain around why you can call a private `MockAdminPasswordDialogCallback` field here?

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
```diff
@@ -0,0 +1,116 @@
...
+        });
+    }
+
```
lognaturel, line 96: While you're here, would you consider testing that the correct callback method is called if the password is incorrect?

collect_app/src/main/java/org/odk/collect/android/utilities/DialogUtils.java
```diff
@@ -154,13 +154,18 @@ public static Dialog createErrorDialog(@NonNull Activity activity, String errorM
         return alertDialog;
     }
 
-    public static <T extends DialogFragment> T showIfNotShowing(Class<T> dialogClass, FragmentManager fragmentManager) {
-        return showIfNotShowing(dialogClass, null, fragmentManager);
+    @Nullable
+    public static <T extends DialogFragment> T getDialog(Class<T> dialogClass, FragmentManager fragmentManager) {
```
lognaturel, line 165: How about adding the notice from the PR here? ("as much as possible we should be using arguments (for static data the dialog needs), Dagger (for dependencies) or ViewModel (for non static data) to get things into fragments so as to avoid crashes or weirdness when they are recreated.")

lognaturel (changes requested):
A few questions and light suggestions inline.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/IdentityPromptViewModel.java
seadowg: Seems fair! Will change

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
seadowg: Ach my fav article. Hoisted with my own petard! Yeah you're dead on. In my head was using for "verification" but you're right that it's doing "recording". I'll fix this!

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
seadowg: Because in `onAttach` we have:

```java
if (context instanceof AdminPasswordDialogCallback) {
    callback = (AdminPasswordDialogCallback) context;
}
```

The dialog assumes expects its host activity to implement the listener. We could decouple this but I didn't want to get into that. I'm not a fan of the pattern as it's awkward for testing (which is always a sign that the design is a little awkward) but it is actually how the Android [docs suggest this should be done](https://developer.android.com/guide/components/fragments#EventCallbacks). Using an actual listener object would be a problem as it could cause weirdness with Fragment recreation.

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
seadowg: Yeah sounds good

collect_app/src/main/java/org/odk/collect/android/utilities/DialogUtils.java
seadowg: Ah I actually meant to do this but it slipped my mind. Will do!

collect_app/src/test/java/org/odk/collect/android/preferences/AdminPasswordDialogFragmentTest.java
lognaturel: > Hoisted with my own petard!

What a lovely expression. 💥 @grzesiek2010 do you know this one?!

lognaturel (approved):

mmarciniak90:
Tested with success on Android 6.0, 7.0, 8.1, 9.0

Verified cases:
* The storage migration flow (with and without admin password set)
   * used package prepared to release testing 
   * scanned QR code with admin password
* Form loading 
   * 85 forms on blank forms list 
   * 1027 submissions on saved forms list
   * loading big form: Nigeria
   * loading from cache
* Form saving
   * cases with save button
   * audit with group
* Admin password dialog
   * no crash after rotation, minimization, lock the screen

@kkrawczyk123 your turn :)

kkrawczyk123:
Nothing weird from my and my Androids 5.1 and 10 sides.

PR #3738, opened 2020-03-24
Get rid of configChanges flags in the manifest file

### LINKED ISSUE #799: Use onSaveInstanceState to retain activity state instead of using configChanges
#### Problem description
Using `configChanges` is a bad practice for retaining the state of the activity. This has been used in [GoogleSheetsUploaderActivity](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/AndroidManifest.xml#L155), [GoogleDriveActivity](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/AndroidManifest.xml#L117), etc.

#### Expected behavior
This should be replaced by using `onSaveInstanceState` to retain the state of the activity on destroy and recreation.

grzesiek2010:
I investigated how `orientation|screenSize` work together and separately and reviewed implemented changes.

We agreed that blocking activity recreations is a bad approach so we decided to get rid of it.

The only risky change is the one made in `MainMenuActivity` take a look at this comment for example: https://github.com/opendatakit/collect/pull/3738#issuecomment-603193682
in other cases I removed single `orientation` flags and as we already agreed 

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/storage/StorageMigrationCompletedBannerTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/storage/StorageMigrationCompletedBannerTest.java
@@ -44,18 +44,10 @@ public class StorageMigrationCompletedBannerTest {
             .around(rule);
 
     @Test
-    public void when_storageMigrationCompleted_should_bannerBeVisibleAndPersistScreenRotation() {
+    public void when_storageMigrationCompleted_should_bannerBeVisibleAndDisappearsAfterScreenRotation() {
         new MainMenuPage(rule)
                 .assertStorageMigrationCompletedBannerIsDisplayed()
                 .rotateToLandscape(new MainMenuPage(rule))
-                .assertStorageMigrationCompletedBannerIsDisplayed()
-                .rotateToPortrait(new MainMenuPage(rule))
-                .assertStorageMigrationCompletedBannerIsDisplayed()
-                .clickDismissButton()
-                .assertStorageMigrationCompletedBannerIsNotDisplayed()
-                .rotateToLandscape(new MainMenuPage(rule))
-                .assertStorageMigrationCompletedBannerIsNotDisplayed()
-                .rotateToPortrait(new MainMenuPage(rule))
                 .assertStorageMigrationCompletedBannerIsNotDisplayed();
     }
 
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -104,7 +104,6 @@ the specific language governing permissions and limitations under the License.
             />
         <activity
             android:name=".activities.FormEntryActivity"
-            android:configChanges="orientation"
             android:windowSoftInputMode="adjustResize" />
         <activity
             android:name=".activities.NotificationActivity"
@@ -139,12 +138,8 @@ the specific language governing permissions and limitations under the License.
         <activity android:name=".activities.FormHierarchyActivity" />
         <activity android:name=".activities.ViewOnlyFormHierarchyActivity" />
         <activity android:name=".activities.GeoPointActivity" />
-        <activity
-            android:name=".activities.GeoPointMapActivity"
-            android:configChanges="orientation" />
-        <activity
-            android:name=".activities.GeoPolyActivity"
-            android:configChanges="orientation" />
+        <activity android:name=".activities.GeoPointMapActivity" />
+        <activity android:name=".activities.GeoPolyActivity" />
         <activity android:name=".activities.FormMapActivity" />
         <activity android:name=".activities.BearingActivity" />
         <activity
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -91,9 +91,7 @@ the specific language governing permissions and limitations under the License.
                 android:resource="@xml/provider_paths" />
         </provider>
 
-        <activity
-            android:name=".activities.MainMenuActivity"
-            android:configChanges="locale|orientation|screenSize" />
+        <activity android:name=".activities.MainMenuActivity" />
         <activity
             android:name=".activities.ScannerWithFlashlightActivity"
             android:screenOrientation="portrait"


grzesiek2010:
@seadowg 
> t looks like orientation on its own doesn't change the behaviour on rotation (the Activity is still recreated) but that using orientation|screenSize does. The MainMenuActivity for instance doesn't call onCreate when it is rotated.

I can confirm that `orientation` alone doesn't work it must be combined with `screenSize`. So seems as if we can remove all `android:configChanges` which contains only one flag.
If we have both `orientation|screenSize` it prevents loosing data on rotation but if we put the app in the background and open again (with `Don't keep activities`) the data will be lost either way so not a good solution.

in https://github.com/opendatakit/collect/pull/3738/commits/60ae23f8d5e63321d2fffd916e34d3d82afd6b78 I removed `android:configChanges` from activities where only `orientation` flag was used because obviously it didn't make sense.

in https://github.com/opendatakit/collect/pull/3738/commits/41294fcf18dd9c8b22a4f47aad1b36afa2f087d5 I removed it from the `MainMenuActivity` there is nothing to save there so it's safe. Actually it has been introduced not long ago in https://github.com/opendatakit/collect/commit/c256cdf9564be66833e2f758987d05c832c47e1b because @lognaturel asked for it. @lognaturel do you maybe remember why? I tried setting languages without that and everything seems fine. I guess we might have fixed the issue (whatever issue it was) accidentally later because I remember some other pr witch changes in the same area but I can't be sure to it's better to ask.

There are four other places where we still have `android:configChanges="orientation|screenSize"`:
- InstanceUploaderListActivity
- GoogleSheetsUploaderActivity
- GoogleDriveActivity
- FileManagerTabs

in all of them I think the reason was (as I guess) to deal with AsyncTask easily (all of them have their own AsyncTasks). 
In such a case the good solution would be to use ViewModels right?
@seadowg @lognaturel any thoughts?

seadowg:
> In such a case the good solution would be to use ViewModels right?
@seadowg @lognaturel any thoughts?

Yeah I think so!

> I removed it from the MainMenuActivity there is nothing to save there so it's safe. 

I actually found that the `AdminPasswordDialog` didn't work without the flag but that's fixed in #3734. Fine as long as we merge that PR first and otherwise it's noted here so QA knows!

grzesiek2010:
> I actually found that the AdminPasswordDialog didn't work without the flag

ahh tricky. Ok we can wait for #3734 but it also depends on what @lognaturel  say about https://github.com/opendatakit/collect/commit/c256cdf9564be66833e2f758987d05c832c47e1b

lognaturel:
Good stuff! I don't know why we thought there had to be some data saved on rotation with the app-level language option and I don't now believe it's necessary. Maybe it was for the admin password dialog but I don't see evidence of that. As far as using the `configChanges` flag instead of `onConfigurationChanged`, I was likely just matching the pattern established by the app without realizing it wasn't a good idea. I think your change is good (with #3734).

grzesiek2010:
I started working on those four activities that still use `android:configChanges="orientation|screenSize"` but I stopped because I think such fixes don't make sense in the long term. Let me explain why.
**InstanceUploaderListActivity** and **GoogleSheetsUploaderActivity** are responsible for sending finalized forms. In both cases those activities are just holders for asyncTasks that do uploading. That's why `android:configChanges="orientation|screenSize"` is used there - it's the easiest way to deal with asyncTasks. In the future it would be good to refactor that code because:
- asyncTasks are not for such long operations like uploading forms
- we could support uploading form in the background using [foreground service](https://developer.android.com/guide/components/services) for example
at this moment since as I said both those activities are just holders for asyncTasks and display just a dialog they can use `android:configChanges="orientation|screenSize"` there is no value in removing it because it would only require risky changes. Those activities could even use one orienattion for example portrait since they just display a dialog and nothing more.

**GoogleDriveActivity** and  **FileManagerTabs** in this case removing `android:configChanges="orientation|screenSize"` makes sense and I even started working on the code but then I ran into tones of duplicated code:
- ListActivities like `GoogleDriveActivity`, `FormChooserListActivity`, `InstanceChooserList` etc extends `AppListActivity` but `FileManagerTabs` contains two fragments `DataManagerList` and `FormManagerList` both extends `AppListFragment`
[AppListActivity](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/AppListActivity.java) and [AppListFragment](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/fragments/AppListFragment.java) are almost identical 
- then those classes are extended by `FormListActivity`, `InstanceListActivity` (`AppListActivity`) and `FormListFragment`, `InstanceListFragment` (`AppListFragment`) where pairs [FormListActivity](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormListActivity.java)-[FormListFragment](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/fragments/FormListFragment.java) and [InstanceListActivity](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/InstanceListActivity.java)-[InstanceListFragment](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/fragments/InstanceListFragment.java) are also identical.
That's tones of duplicated code that is difficult to maintain.

My proposal is to refactor the existed code first so that all places with lists would be fragments and then all classes would extend one `AppListFragment`. For example `FormChooserListActivity` would contain a fragment that would extend `AppListFragment` etc.
Then we can implement viewmodels and handle configuration changes much easier instead of doing it now in various different places.

@lognaturel 
The first case with **InstanceUploaderListActivity** and **GoogleSheetsUploaderActivity** is not important for now we can leave it as is and get back to it in he future.
For now we should fix the places where w display lists. I started investigating it but I stopped because it would be a much bigger change so I need a green light from you.

seadowg:
@grzesiek2010 definitely seems like a good route to look at pulling out duplicated code between all those Activity classes. The one risk is that we end up creating a nasty (deep) OO hierarchy but as long as we keep in mind to avoid that I think we'll be fine.

Maybe this PR could just be the `MainMenuActivity` changes and we could put a checklist of the activities that need changed in the issue given there will be a few PRs.

grzesiek2010:
> Maybe this PR could just be the MainMenuActivity changes and we could put a checklist of the activities that need changed in the issue given there will be a few PRs.

I agree. i'm going to make this pr ready for review as is and then I'll create an issue for the refactoring i mentioned.

seadowg (approved):
@opendatakit/testers there is a crash on rotation that will happen if the Admin Password dialog is showing but this is fixed in #3734. Good to go with testing on this one but just letting you know!

lognaturel:
Merging this, keeping #799 open, adding a comment referring to the findings above and modifying the original issue to add a checklist sounds great to me. @grzesiek2010 you should be able to edit it. 

The refactors are lower priority than other issues for this release but would be nice to do.

mmarciniak90:
@grzesiek2010 I noticed a difference between this PR and released v1.26.1.
Banner with information `Storage migration completed!` disappears after rotating the device.

Here are the exact steps:
1. `Don't keep activities` can be turned ON or OFF - the same result
2. User finishes storage migration
3. Banner `Storage migration completed!` is visible
4. User changes device orientation
5. Banner disappears 

@lognaturel what is your opinion about this difference?

lognaturel:
@grzesiek2010, @mmarciniak90 this feels fine to me because we intend for that banner to be short-lived. Basically we want to be fairly sure that the user sees it but even if they didn't it wouldn't really be a big deal. Does anyone feel strongly that it's a problem?

mmarciniak90:
Thanks, @lognaturel. It's a good point. I can live with it  :stuck_out_tongue_winking_eye: 

But today I noticed similar behavior for the expandable menu :confused:
It disappears after the rotation.

I can compare this with the expandable menu on the started form. In this case, the menu doesn't disappear after the rotation.

@lognaturel, @grzesiek2010 I have also one idea. When we accept this PR, we could merge this PR and #3734. Then @kkrawczyk123 and I can verify both PRs on the master branch because they are related. In this case, I see the added value of this kind of end to end testing.

seadowg:
@mmarciniak90 happy to go with that if that works for you folks!

grzesiek2010:
> When we accept this PR, we could merge this PR and #3734

you can test that pr first and after merging it I can rebase this one the result will be the same.

> It disappears after the rotation.
I can compare this with the expandable menu on the started form. In this case, the menu doesn't disappear after the rotation.

That's how it should behave if we don't block activity recreation, weird... I can't see anything in `FormEntryActivity` that would be responsible for that neither in `MainMenuActivity`.
Maybe a separate issue for hardcores who like such things?

grzesiek2010:
The branch has been rebased onto master so now it contains #3734

mmarciniak90:
I tested this PR on Android 6.0, 7.0, 8.1, 9.0

**Verified cases**
- `Don't keep activities` can be turned ON or OFF
- Migration banner
- `Storage migration completed!` banner
- Admin password - crash fixed in #3734
- rotating the device 
- locking the screen
- minimizing the app

**Mentioned regressions:**
- disappearing `Storage migration completed!` banner
- disappearing expandable menu after the rotation

grzesiek2010:
Is testing finished here? We are aware of those two cases:
> disappearing Storage migration completed! banner
disappearing expandable menu after the rotation

and agreed that it's not a big problem so maybe we can merge.

mmarciniak90:
Let's give @kkrawczyk123 a chance to test it on Monday :grin:

kkrawczyk123:
Verified on Androids 5.1 and 10. I haven't noticed any issues except those already mentioned regressions. 

PR #3622, opened 2020-02-04
Support scanning inverted qr codes

grzesiek2010:
Closes #3614 

#### What has been done to verify that this works as intended?
I tested barcode widget and scanning qr code with settings.

#### Why is this the best possible solution? Were any other approaches considered?
Inverted qr code it's not something common but if there is a request from a real user and we can do that why not?

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
It's not a risky change because it is related just to barcode widget and scanning setting. If we test those two functionalities everything should be fine.

#### Do we need any specific form for testing your changes? If so, please attach one.
Any form with barcode widget.

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
No.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #3743, opened 2020-03-24
Improved QRCode design

hypercubestart:
See https://forum.opendatakit.org/t/qr-code-screen-redesign/24996/13

I have been working with a team to improve UI/UX the QR code design. Please see the above link for screenshots and prior discussion.

We introduce the `QRCodeTabs` activity which contains a viewpager in its layout. The viewpager has two tabs: an embedded qr code scanner fragment and a show qr code fragment. We chose this design because other apps we referenced used a similar layout. 

This PR also removes the need of a decided `ScanQRCodeActivity`, since putting the Scan QR Fragment first in the ViewPager allows the `Configure via QR Code` menu item to send the user directly to the newly created `QRCodeTabs` activity.

- [x ] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).


seadowg:
Hey @hypercubestart. Could you make sure to check off the todo list items so we know if this is ready for review or not?

hypercubestart:
@seadowg yep, ready to review

lognaturel:
As I was trying it out, I noticed that switching tabs in landscape crashes:
```
    java.lang.NullPointerException: Attempt to invoke virtual method 'void com.journeyapps.barcodescanner.DecoratedBarcodeView.pause()' on a null object reference
        at org.odk.collect.android.fragments.QRScannerFragment.onPauseFragment(QRScannerFragment.java:142)
        at org.odk.collect.android.preferences.QRCodeTabs$1.onPageSelected(QRCodeTabs.java:68)
        at androidx.viewpager.widget.ViewPager.dispatchOnPageSelected(ViewPager.java:1947)
        at androidx.viewpager.widget.ViewPager.scrollToItem(ViewPager.java:686)
        at androidx.viewpager.widget.ViewPager.setCurrentItemInternal(ViewPager.java:670)
        at androidx.viewpager.widget.ViewPager.setCurrentItemInternal(ViewPager.java:631)
        at androidx.viewpager.widget.ViewPager.setCurrentItem(ViewPager.java:612)
        at com.google.android.material.tabs.TabLayout$ViewPagerOnTabSelectedListener.onTabSelected(TabLayout.java:3278)
        at com.google.android.material.tabs.TabLayout.dispatchTabSelected(TabLayout.java:1793)
        at com.google.android.material.tabs.TabLayout.selectTab(TabLayout.java:1786)
        at com.google.android.material.tabs.TabLayout.selectTab(TabLayout.java:1746)
        at com.google.android.material.tabs.TabLayout$Tab.select(TabLayout.java:2139)
        at com.google.android.material.tabs.TabLayout$TabView.performClick(TabLayout.java:2327)
        at android.view.View$PerformClick.run(View.java:21256)
```

hypercubestart:
@lognaturel Thanks for taking a look! Bug is now fixed.
PR changing documentation can be found here: https://github.com/opendatakit/docs/pull/1204

I would appreciate it if you could review this PR again!

collect_app/src/main/java/org/odk/collect/android/adapters/TabAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentManager;
+import androidx.fragment.app.FragmentPagerAdapter;
+
+import org.odk.collect.android.fragments.QRScannerFragment;
+import org.odk.collect.android.fragments.ShowQRCodeFragment;
+
+public class TabAdapter extends FragmentPagerAdapter  {
```
seadowg: We should really be using [ViewPager2](https://developer.android.com/reference/androidx/viewpager2/widget/ViewPager2) and [FragmentStateAdapter](https://developer.android.com/reference/androidx/viewpager2/adapter/FragmentStateAdapter) here. ViewPager was recently deprecated so adding it in now is queuing up rework in the future unfortunately! Would you be able to have a look and see much work it would be to convert these changes over?

seadowg (changes requested):
Big change around some annoying Android deprecations. Sorry! I'll pause reviewing for you to look at that.

collect_app/src/main/java/org/odk/collect/android/adapters/TabAdapter.java
hypercubestart: yep, updated now! thanks!

collect_app/build.gradle
```diff
@@ -261,7 +261,8 @@ dependencies {
     debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.2'
 
     // Android Architecture Components:
-    implementation "androidx.lifecycle:lifecycle-extensions:2.2.0"
+    implementation "androidx.lifecycle:lifecycle-extensions:2.1.0"
```
seadowg: Why the downgrade here?

collect_app/src/main/java/org/odk/collect/android/adapters/TabAdapter.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public Fragment createFragment(int position) {
+        switch (position) {
+            case 0:
+                return new QRScannerFragment();
+            case 1:
+                return new ShowQRCodeFragment();
+            default:
+                // should never reach here
+                return null;
```
seadowg: I'd be happy with this throwing a `IllegalArgumentException` instead. I'm pretty sure a `null` will result in a crash anyway so might as well have something that's easy to diagnose!

collect_app/src/main/java/org/odk/collect/android/fragments/QRScannerFragment.java
```diff
@@ -0,0 +1,145 @@
...
+public class QRScannerFragment extends Fragment implements DecoratedBarcodeView.TorchListener {
+
+    DecoratedBarcodeView barcodeScannerView;
+    private Button switchFlashlightButton;
+    private BeepManager beepManager;
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setRetainInstance(true);
```
seadowg: Why do we need this? This method has generally always been a little dangerous as it changes how the Fragment lifecycle works and it also looks like it's going to be deprecated: https://developer.android.com/reference/androidx/fragment/app/Fragment#setRetainInstance(boolean)

collect_app/src/main/java/org/odk/collect/android/fragments/QRScannerFragment.java
```diff
@@ -0,0 +1,145 @@
...
+        });
+        new PermissionUtils().requestCameraPermission(getActivity(), new PermissionListener() {
+            @Override
+            public void granted() {
+                barcodeScannerView.decodeSingle(new BarcodeCallback() {
+                    @Override
+                    public void barcodeResult(BarcodeResult result) {
+                        beepManager.playBeepSoundAndVibrate();
+                        try {
+                            QRCodeTabs.applySettings(getActivity(), CompressionUtils.decompress(result.getText()));
```
seadowg: Should this static method move out of the activity now? Seems weird for the logic to live there given it's being called from a Fragment.

collect_app/src/main/java/org/odk/collect/android/fragments/ShowQRCodeFragment.java
```diff
@@ -96,10 +95,11 @@
     @Override
     public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, Bundle savedInstanceState) {
         View view = inflater.inflate(R.layout.show_qrcode_fragment, container, false);
-        ((CollectAbstractActivity) getActivity()).initToolbar(getString(R.string.configure_via_qr_code));
         ButterKnife.bind(this, view);
         setHasOptionsMenu(true);
         setRetainInstance(true);
+        passwordsSet[0] = !((String) AdminSharedPreferences.getInstance().get(KEY_ADMIN_PW)).isEmpty();
```
seadowg: The `getInstance()` methods here are deprecated. You should see warning in Android Studio about this. As the comment on them says it would be better to use [Dagger injection to fetch these](https://github.com/opendatakit/collect/blob/master/docs/CODE-GUIDELINES.md#dependency-injection).

collect_app/src/main/res/values/strings.xml
```diff
@@ -493,18 +493,16 @@
     <string name="no_time_selected">No time selected</string>
     <string name="configure_via_qr_code">Configure via QR code</string>
     <string name="qr_code">QR code configuration</string>
-    <string name="scan_qrcode">Scan code from other device</string>
+    <string name="import_qrcode_sd">Import QRCode</string>
```
seadowg: This should be "Import QR Code" right?

collect_app/src/main/res/layout/fragment_scan.xml
```diff
@@ -0,0 +1,25 @@
...
+    tools:context="org.odk.collect.android.fragments.QRScannerFragment">
+
+
+    <com.journeyapps.barcodescanner.DecoratedBarcodeView
+        android:id="@+id/barcode_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        />
+
+    <Button
```
seadowg: Can we make this a `MaterialButton` and remove the custom padding, background etc? It also looks like it needs a larger margin from what I can see in the design preview. I think using `@dimen/margin_large` would do it.

collect_app/src/main/res/layout/show_qrcode_fragment.xml
```diff
@@ -81,6 +63,10 @@
             android:layout_weight="1"
             android:padding="5dp"
             android:textSize="15sp"
-            android:textStyle="bold" />
+            android:gravity="center_vertical"
+            android:text="@string/qrcode_with_both_passwords"
+            android:textColor="#000000"
```
seadowg: This should be the default text color. Hard coding like this will make the text the wrong color for the dark theme though. Were you running into problems that made you set this?

collect_app/src/main/java/org/odk/collect/android/preferences/QRCodeTabs.java
```diff
@@ -0,0 +1,128 @@
...
+import androidx.appcompat.widget.Toolbar;
+import androidx.core.content.FileProvider;
+import androidx.viewpager2.widget.ViewPager2;
+
+import com.google.android.material.tabs.TabLayout;
+import com.google.android.material.tabs.TabLayoutMediator;
+
+import static org.odk.collect.android.activities.ActivityUtils.startActivityAndCloseAllOthers;
+
+public class QRCodeTabs extends CollectAbstractActivity {
```
seadowg: This should be `QRCodeTabsActivity`

seadowg (changes requested):
A few changes inline.

As a bigger piece of feedback it looks like there isn't really a lot of tests in place for the changes in this PR. Right now I can change `TabAdapter#getItemCount` to just `return 0` and comment out `onCreateOptionsMenu`/`onPrepareOptionsMenu` to basically remove all the features in this PR and then run tests and I won't see any failures.

As far as I can see there are 4 key "flows" in this PR:

* Scan a QR code
* View a QR code
* Import a QR code (from disk)
* Share a QR code with another app

We should make sure we have tests in place to protect these flows and any error cases that could come up. Granted some of this wasn't tested before but as we touch these things we should make sure we don't move on without making sure the new code has a reason to exist. Happy to chat throught the best way to get that testing in place either here or on Slack!

collect_app/src/main/res/values/strings.xml
lognaturel: Yes, and I think that for consistency with the menu on the main screen it should be lowercase "code" -- "Import QR code"

lognaturel (commented):
Great comments by @seadowg. I'm not sure why I wasn't thinking about tests when we last spoke. He's absolutely right that there's logic here that should be covered. Definitely pull someone into a Slack conversations if you want more guidance/ideas/support on that.

collect_app/build.gradle
hypercubestart: sorry this was an accident. thanks for catching this!

collect_app/src/main/java/org/odk/collect/android/adapters/TabAdapter.java
hypercubestart: thanks for review, done!

collect_app/src/main/java/org/odk/collect/android/fragments/QRScannerFragment.java
hypercubestart: I was testing different behaviors, forgot to delete this. Sorry, I will delete this.

collect_app/src/main/java/org/odk/collect/android/fragments/QRScannerFragment.java
hypercubestart: I put the applySettings method in QRCodeTabs activity because both `QRScannerFragment` and `ShowQRCodeFragment` use it. Does this make sense or should the static method be moved somewhere else?

collect_app/src/main/res/values/strings.xml
hypercubestart: absolutely! done, thanks for review!

collect_app/src/main/res/layout/fragment_scan.xml
hypercubestart: yep! @lognaturel and I talked about this before and we were going to update the button in a follow-up PR, but I can change it here

collect_app/src/main/java/org/odk/collect/android/preferences/QRCodeTabs.java
hypercubestart: done! @seadowg should this also be moved to collect/android/activities folder?

collect_app/src/main/java/org/odk/collect/android/fragments/ShowQRCodeFragment.java
hypercubestart: done thanks!

collect_app/src/main/res/layout/show_qrcode_fragment.xml
hypercubestart: with the dark color scheme, white text over a yellowish background: 
https://github.com/hypercubestart/collect/blob/51fbeda54ab3c3510e135114a660c2fe0fde2be8/collect_app/src/main/res/layout/show_qrcode_fragment.xml#L43
did not look well. So, I planned to just keep the color scheme consistent. Is this okay? As a sidenote, should I pull out the color hex values into `colors.xml`

collect_app/src/main/java/org/odk/collect/android/fragments/QRScannerFragment.java
seadowg: Yeah ideally the fragments shouldn't have to know anything about the activity. I'd maybe move it to a new `SettingsUtils` in the `preferences` packages. Saying that actually I realize all of this code should live in the `preferences` package. Would you be able to move it?

collect_app/src/main/res/layout/show_qrcode_fragment.xml
seadowg: Ah right I see. You're right that is not going to look good on the yellow! We really want our colors to as much as possible come from our theme attributes. This makes them easier to change but also just limits the amount of colors floating around in the app. I think for the background color here we should use `?colorError` (a red warning/error color) instead of the yellow and for the text we should use `?colorOnError`.

collect_app/src/main/java/org/odk/collect/android/preferences/QRCodeTabs.java
seadowg: As above I think it moves to `preferences`. Maybe all of it should go in a `preferences.qr` packakge... what do you think? We're trying to organize our code by function/domain rather than component type.

collect_app/src/main/res/layout/show_qrcode_fragment.xml
hypercubestart: i can roll with this, the error icon will also use the `?colorOnError` to match the color of the text

collect_app/src/main/java/org/odk/collect/android/preferences/QRCodeTabs.java
hypercubestart: I like the idea of a `preferences.qr` package. There's a few files in here and it feels a lot cleaner than forcing even more items into the crowded `preferences` package.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
```diff
@@ -0,0 +1,138 @@
+/* Copyright (C) 2017 Shobhit
```
seadowg, line 1: Did this come from a file that had a copyright notice on it? 

@lognaturel should we have this kind of copyright notice in the code? I know copyright headers like this can be useful for relicensing projects (as you can scan for all the copyright holders) but as far as I know we don't have them for every contribution so we'd probably need to get that info from Git anyway.

lognaturel: https://github.com/opendatakit/governance/issues/2 For unresolved governance issue. I go back and forth on what I think is best and so have stopped making any code review comments. I think that yes, git is and should be the ultimate source for authorship and copyright info. I also think that for forks, it's nice to have an ongoing reminder that they're pulling in code from another source. I find the license header more compelling than the copyright notice. I (intend to) always include the header on new files I create.

This specific file was renamed. On one hand, the header does capture original copyright. On the other, the file has changed enough that it's not super meaningful.

So I don't have strong feelings about what happens to this header.

seadowg: Yeah totally agree a license header is a different story. I'd be up for getting rid of author specific copyright notices like this where we see them as they're usually just inaccurate - they don't have copyright for this entire file. I'd generally ask people not to include them unless they feel strongly they want it as part of the PR.

Getting rid of them without lawyers is pretty scary but I'd say we can remove this as it looks autogenerated and doesn't actually reference anyone by email/full name. I'll CC our pal @shobhitagarwal1612 (who I'm guessing is our "shobit" here) and see if he's ok getting rid of it. At least that will be one gone!

shobhitagarwal1612: I have no issues. Please remove it. Happy to see the progress happening to the project 😄 🚀 

I have always been nervous while adding a new file without a copyright. But since that issue is still a gray area, I'd leave the final decision to @lognaturel.

hypercubestart:
@seadowg the latest commits should improve the test quality of View QR Code by introducing Dependency Inversion into how the QR Code is generated. Last problem I am dealing with is how to test `pressShareQRCode` and verify the nested intent. I think this PR is ready to go after this.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static androidx.test.espresso.intent.Intents.intended;
+import static androidx.test.espresso.intent.Intents.intending;
+import static androidx.test.espresso.intent.matcher.IntentMatchers.hasAction;
+import static androidx.test.espresso.intent.matcher.IntentMatchers.hasType;
+import static androidx.test.espresso.intent.matcher.IntentMatchers.isInternal;
+import static org.hamcrest.Matchers.not;
+
+
+@RunWith(AndroidJUnit4.class)
+public class QrCodeActivitiesTest {
```
seadowg: I'd try and name this around the "feature" rather than the components under test (as this test shouldn't really care about them). Maybe `ConfigureWithQRCodeTest`?

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
```diff
@@ -0,0 +1,126 @@
...
+    // drawable resource that will act as "qr code" in this test
+    private final int checkerBackgroundDrawableId = R.drawable.checker_background;
```
seadowg: This should probably be a `static` and use constant syntax (`CHECKER_BACKGROUND_DRAWABLE_ID`).

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
```diff
@@ -0,0 +1,126 @@
...
+            .outerRule(GrantPermissionRule.grant(
+                    Manifest.permission.READ_EXTERNAL_STORAGE,
+                    Manifest.permission.WRITE_EXTERNAL_STORAGE,
+                    Manifest.permission.READ_PHONE_STATE
+            ))
+            .around(new ResetStateRule(new AppDependencyModule() {
+                @Override
+                @Provides
+                public QRCodeGenerator providesQRCodeGenerator() {
+                    return new QRCodeGenerator() {
```
seadowg: Maybe this stub version could be a private static inner class here so you're able to just do `new StubQRCodeGenerator()`.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Before
+    public void stubAllExternalIntents() {
+        // By default Espresso Intents does not stub any Intents. Stubbing needs to be setup before
+        // every test run. In this case all external Intents will be blocked.
+        intending(not(isInternal())).respondWith(new Instrumentation.ActivityResult(Activity.RESULT_OK, null));
+    }
+
+    @Test
+    public void checkQRScannerIsInitiated() {
```
seadowg: Again just in terms of naming lets try and get closer to describing the feature. `onMainMenu_clickConfigureQRCode_andClickingOnScan_opensScanner()` for this as an example. I'd try doing a similar "given-when-then" style for all these tests.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Test
+    public void pressShareQRCode() {
+        new MainMenuPage(rule)
+                .assertOnPage()
+                .clickOnMenu()
+                .clickConfigureQR()
+                .clickOnId(R.id.menu_item_share);
+
+        intended(hasAction(Intent.ACTION_CHOOSER));
```
seadowg: I think you still want another assertion here on the file that's sent right?

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -265,4 +271,14 @@ public BackgroundWorkManager providesBackgroundWorkManager() {
     public NetworkStateProvider providesConnectivityProvider() {
         return new ConnectivityProvider();
     }
+
+    @Provides
+    public QRCodeGenerator providesQRCodeGenerator() {
+        return new QRCodeGenerator() {
+            @Override
+            public Observable<Bitmap> generateQRCode(Collection<String> selectedPasswordKeys) {
+                return QRCodeUtils.getQRCodeGeneratorObservable(selectedPasswordKeys);
```
seadowg: If possible it'd be nicer for `QRCodeUtils.getQRCodeGeneratorObservable`'s logic to move to this implementation (and probably have it in its own class file). Wrapping like this is a nice trick for dealing with statics we don't control but we do control this one!

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
hypercubestart: yes, I'm not quite sure how though, because it may also be helpful to check that the stream is valid

seadowg: I think you want to check that the intent has the URI in it right?

hypercubestart: yes, but this Intent wraps another Intent with contains the URI as an Extra. I'm having trouble testing this inner intent.

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
hypercubestart: done, thanks!

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
hypercubestart: yes you are right, thanks!

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
hypercubestart: done, thanks!

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
hypercubestart: Is it okay to break up QRCodeUtils logic like this? Because it uses certain functions/static constants that aren't accessible outside the `QRCodeUtils` or outside the `utils` package? 

My concern is that If this is the only thing that needs to be moved out, I'm not sure if it's worth it to separate it from the other QRCodeUtils functions which may decrease readability.

seadowg: I do see your point but it seems to be that everything in that utils provides information about a single QR code image (the image itself, it's path etc) other than `decodeFromBitmap`. Could it be broken into a `QRCodeGenerator` and a `QRCodeReader` interface?

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/QrCodeActivitiesTest.java
hypercubestart: nvm, figured it out! can you please take another look? I'm a little worried now that the test is too coupled with the code, especially the `expectedUri` part

https://github.com/hypercubestart/collect/blob/4005fe290c0d61bdd52271c16e0ce4a9f53cf4ba/collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/ConfigureWithQRCodeTest.java#L125
        Uri expected = FileProvider.getUriForFile(ApplicationProvider.getApplicationContext(),
                BuildConfig.APPLICATION_ID + ".provider",
                new File(path));

which I copied over directly

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
hypercubestart: If its okay with you, I went ahead and broke it into `QRCodeGenerator` interface, and left the remaining functions untouched in the `QRCodeUtils.java`

collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/ConfigureWithQRCodeTest.java
```diff
@@ -0,0 +1,192 @@
...
+import static org.junit.Assert.assertTrue;
+
+
+@RunWith(AndroidJUnit4.class)
+public class ConfigureWithQRCodeTest {
+    // drawable resource that will act as "qr code" in this test
+    private static final int CHECKER_BACKGROUND_DRAWABLE_ID = R.drawable.checker_background;
+
+    @Rule
+    public IntentsTestRule<MainMenuActivity> rule = new IntentsTestRule<>(MainMenuActivity.class);
```
seadowg, line 61: We've found recently that we could run into ordering problems with rules so it'd be best if this rule was also part of the `RuleChain` below. Just remove the `@Rule` and add `around(rule)` to the end of the chain.

seadowg (approved):
Looks good! Just one comment but it won't hold up going to QA.

mmarciniak90:
I noticed small shortcoming but it's just appearances issue :smiley: 
Lines under tabs in dark mode have blue color instead of the color which matches the theme.

I added a screenshot from `Delete Saved Form` to compare

|Screenshot_2020-05-11-13-03-20-774_org odk collect android(1)|Screenshot_2020-05-11-13-14-41-372_org odk collect android|
|---|---|

mmarciniak90:
@hypercubestart I also noticed an issue with importing QR code from a file. Even if I select proper QR code, settings aren't applied.

Another thing is scanning inverted QR code which looks like broken.

invert

Additionally, when I scan incorrect QR code I see the message `QR does not contain valid settings` but I'm not moved to the previous view. That cause that scanning is not 'active' - I need to change device orientation to active scanning again.

hypercubestart:
@mmarciniak90 thanks for all the testing! all issues should be fixed now, except the scanning inverted QR Code. Is this a feature that was always there? Is it something that we should support since it doesn't seem the xzing qr scanning library supports inverted qr code scanning?

@lognaturel tagging for your opinion on this

lognaturel:
Good catches and fixes. 👍

@hypercubestart inverted QR code scanning is supported by XZing and had just been added to Collect: https://github.com/getodk/collect/pull/3622/files. See if that helps you bring it back. I don't think it's critical because configuration happens infrequently and usually in controlled environments so we can always file a separate issue if needed.

For the importing a QR code from a file fix, it would have been nice to see a test alongside it! That kind of broken functionality usually lends itself really well to a red-green cycle -- start with a failing test and then write code that fixes the test. Maybe you can still introduce a test now to prevent future regressions? I think it's more important to get this merged so not a big deal if you don't have time.

hypercubestart:
@lognaturel I don't think I'll be able to introduce the changes by the codefreeze, so we should move those into a separate issue. thanks!

mmarciniak90:
New issue with inverted QR code reported #3823

@hypercubestart 
Noticed incorrect case:
1. User set Aggregate as a server
2. Username and **password to Aggregate** are declared
3. **Admin password is NOT set**
4. User opens QR code - information `Contains sensitive information: server password` is visible
5. QR code is scanned by another device

RESULT: Aggregate password is set as a server password as well as **ADMIN password**

Verified cases with success on Android 6.0, 7.0, 8.1, 9.0:
- Lines under tabs in dark mode have the correct color
- scanning QR code after scan code without valid settings
- import QR code from a file
- select a file without QR code as a file to import QR code
- share
- scanned Central's settings 
- scanned Aggregate's settings 
- scanned Google Drive's settings 

hypercubestart:
@mmarciniak90 thanks so much for catching that, I really appreciate it. Issue is fixed now

lognaturel:
Good catch, @mmarciniak90, and thanks for the quick fix, @hypercubestart. We can add more tests around that password toggle behavior in a follow-up PR.

mmarciniak90:
**Re-tested cases with success on Android 6.0, 7.0, 8.1, 9.0:**
* generated QR code with Aggregate password
* applied settings from `collect.settings` file
* lines under tabs in dark mode have the correct color
* scanned QR code after scan code without valid settings
* imported QR code from a file
* selected a file without QR code as a file to import QR code
* shared
* scanned Central's settings
* scanned Aggregate's settings
* scanned Google Drive's settings

kkrawczyk123:
Verified the same cases on Androids 5.1 and 10. I have noticed an issue with saving QR code to drive and attaching to email message for my Android 5.1 device but it was also visible on master and store version so I documented it in #3826 in case any user would encounter similar behavior.

PR #3745, opened 2020-03-25
Support calculations in field-lists

### LINKED ISSUE #3211: Fields with calculations  in field lists are not updated as calculated value changes
Collect v1.22+

#### Problem description
If a field in a field list has a calculation, its value should update as the calculated value changes. It does not.

See failing test at https://github.com/opendatakit/collect/blob/e1a726ef6ca5a9c95b4abd3569a04f97e3193c24/collect_app/src/androidTest/java/org/odk/collect/android/formentry/FieldListUpdateTest.java#L185

Note: the comment says "calculation doesn't seem to be updated whether or not there's a fieldlist" but now it looks to me like it is updated. See [visible_calculate.xlsx](https://github.com/opendatakit/collect/files/3399080/visible_calculate.xlsx)

Open a form such as [visible_calculate.xlsx](https://github.com/opendatakit/collect/files/3399080/visible_calculate.xlsx). Change the value of the "Foo" field.

#### Expected behavior
The "Foo length" field should update as the "Foo" field's value changes. Interestingly, even navigating away from the field list and back to it doesn't update the value.

grzesiek2010:
Closes #3211 

I tested implemented changes manually and there is also an automated test.

I'l try to explain what was wrong with our code:
The form we are using has two questions on the same screen where the answer of the second question is based on the answer of the first question (calculated)
| type  | name | label | appearance | calculation |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| begin group | my-group | My group | field-list | |
| text  | foo  | Foo | | |
| integer | foo-length | Foo length | | string-length(${foo}) |
| end group | my-group | | | |

When we open the form the first answer is `null` and the second answer is `0`. We expect the second answer to be updated every time we edit the first one so 
if we type `q` it should be updated fro `0`->`1`
then if we type `w` (`qw`) it should be updated from `1`->`2`
then if we type `e` (`qwe`) it should be updated from `2`->`3` etc

The problem with the current implementation was that we were trying to save all answers from the screen at the same time so analyzing the same scenario as above:
if we type `q` we are trying to save answers `q` and `0` because it is what we have on the screen,
First we save `q` an the value of the element is changed from `null` to `q` that triggers calculations and the value of the second element is changed from `0` to `1` (it takes place in javarosa automatically) and then we try to save the second question from the screen `0` ignoring that calculations took place what changes the answer back from `1` -> `0`

The only solution I can come up with is to save questions one by one and check whether the answer of the questions we are trying to save changed after saving a previous answer if it changed we do nothing because that mean it was recalculated  otherwise we save the answer from the screen.

We should test field-list especially with calculations/hiding questions to ensure there is no regression.

I'm not sure... was the limitation (bug) mentioned anywhere?

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1092,6 +1092,28 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         return true;
     }
 
+    // The method saves questions one by one in order to support calculations in field-list groups
+    private void saveAnswersForCurrentScreen(FormEntryPrompt[] mutableQuestionsBeforeSave, List<ImmutableDisplayableQuestion> immutableQuestionsBeforeSave) {
+        FormController formController = getFormController();
+        ODKView currentView = getCurrentViewIfODKView();
+        if (formController == null || currentView == null) {
+            return;
+        }
+
+        int index = 0;
+        for (Map.Entry<FormIndex, IAnswerData> answer : currentView.getAnswers().entrySet()) {
+            // Questions with calculates will have their answers updated as the questions they depend on are saved
+            if (!isQuestionRecalculated(mutableQuestionsBeforeSave[index], immutableQuestionsBeforeSave.get(index))) {
+                try {
+                    formController.saveOneScreenAnswer(answer.getKey(), answer.getValue(), false);
+                } catch (JavaRosaException e) {
+                    Timber.w(e);
+                }
+            }
+            index++;
+        }
+    }
+
     /**
      * Clears the answer on the screen.
      */
@@ -2808,7 +2830,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             immutableQuestionsBeforeSave.add(new ImmutableDisplayableQuestion(questionBeforeSave));
         }
 
-        saveAnswersForCurrentScreenOneByOneToSupportCalculations(questionsBeforeSave, immutableQuestionsBeforeSave);
+        saveAnswersForCurrentScreen(questionsBeforeSave, immutableQuestionsBeforeSave);
 
         FormEntryPrompt[] questionsAfterSave = Collect.getInstance().getFormController().getQuestionPrompts();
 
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2888,6 +2886,34 @@ private void updateFieldListQuestions(FormIndex lastChangedIndex) {
         }
     }
 
+    private void saveAnswersForCurrentScreenOneByOneToSupportCalculations(FormEntryPrompt[] mutableQuestionsBeforeSave, List<ImmutableDisplayableQuestion> immutableQuestionsBeforeSave) {
```
lognaturel: I think it's ok to overload `saveAnswersForCurrentScreen` instead of having a long name. You could add a comment to differentiate them.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2888,6 +2886,34 @@ private void updateFieldListQuestions(FormIndex lastChangedIndex) {
...
+        FormController formController = getFormController();
+        ODKView currentView = getCurrentViewIfODKView();
+        if (formController == null || currentView == null) {
+            return;
+        }
+
+        int index = 0;
+        for (Map.Entry<FormIndex, IAnswerData> answer : currentView.getAnswers().entrySet()) {
+            if (isQuestionNotRecalculated(mutableQuestionsBeforeSave[index], immutableQuestionsBeforeSave.get(index))) {
```
lognaturel: I have a hard time with the negative in the method name. In fact, you could just use `!Objects.equals` here, right? I think that would make things easier to understand. Then you could have a comment above the condition that says something like `questions with `calculates` will have their answers updated as the questions they depend on are saved`.

lognaturel (changes requested):
Good explanation. The substance of the fix looks good to me but I have a couple of naming/clarity suggestions.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2888,6 +2908,12 @@ private void updateFieldListQuestions(FormIndex lastChangedIndex) {
         }
     }
 
+    // If an answer has changed after saving one of previous answers that means it has been recalculated automatically
+    private boolean isQuestionRecalculated(FormEntryPrompt mutableQuestionBeforeSave, ImmutableDisplayableQuestion immutableQuestionBeforeSave) {
+        return !(mutableQuestionBeforeSave.getAnswerText() == null && immutableQuestionBeforeSave.getAnswerText() == null
```
lognaturel, line 2887: This is exactly what `Objects.equals` does, right? I was suggesting not even introducing a custom method. I don't feel strongly about it, though, if you like this better.

lognaturel (approved):
This may be slow if the visible fields on the screen each trigger calculations but I don't think there's much of a way around it.

mmarciniak90:
@grzesiek2010 Is it expected behavior that removing response for `Foo length` removes also `Foo` answer?

grzesiek2010:
Of course not! :open_mouth: but I think it's not my bug (at least in this pr) and there is something wrong on master... I reported an [issue](https://github.com/opendatakit/collect/issues/3751) please take a look to make sure I didn't miss anything. If I'm right you can add whatever red flag is available because it's a horrible bug.

mmarciniak90:
@grzesiek2010 you have right. This bug is visible on released v1.26.1.

mmarciniak90:
**Tested with success**

**Verified cases:**
* `visible_calculate` form
* ~20k signs
* add/remove a sign
* remove response of the field with calculation (Foo length) - backs to null, not to `0`
* remove response of calculated field (Foo) - backs to null and field with calculation backs to '0'
* overwrite value when answer added manually
* new lines, spaces, extra signs are counted
* `fieldlist-updates` form with hidden and calculated fields

PR #3746, opened 2020-03-25
Fixed layers directory path

### LINKED ISSUE #3741: The layers directory path is incorrect after storage migration
Collect v1.26.1

#### Problem description
When setting a reference layer, the layers directory path is incorrect after storage migration

1. Make sure the storage migration has completed successfully
1. Go to General Settings > Maps
1. Tap `Layer data file`

#### Expected behavior

It should say `/sdcard/Android/data/org.odk.collect.android/files/layers` but just says `/layers` instead.

grzesiek2010:
Closes #3741

Verified the case described in the issue manually and added automated tests.

It's just a small fix.

Verifying the case described in the issue would be enough, it's not a risky change.

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
@@ -284,5 +284,6 @@ public class FileUtilsTest {
         assertThat(FileUtils.simplifyScopedStoragePath(""), is(""));
         assertThat(FileUtils.simplifyScopedStoragePath("blahblahblah"), is("blahblahblah"));
         assertThat(FileUtils.simplifyScopedStoragePath("/storage/emulated/0/Android/data/org.odk.collect.android/files/layers"), is("/sdcard/Android/data/org.odk.collect.android/files/layers"));
+        assertThat(FileUtils.simplifyScopedStoragePath("/storage/emulated/0/Android/data/org.odk.collect.android/files/layers/countries/countries-raster.mbtiles"), is("/sdcard/Android/data/org.odk.collect.android/files/layers/countries/countries-raster.mbtiles"));
     }
 }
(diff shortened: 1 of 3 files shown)


seadowg (approved):

collect_app/src/test/java/org/odk/collect/android/utilities/FileUtilsTest.java
```diff
@@ -276,4 +276,13 @@ public void mediaDirNameIsCorrect() {
         assertThat(metadataFromFormDefinition.get(FileUtils.FORMID), is("set-geopoint-before"));
         assertThat(metadataFromFormDefinition.get(FileUtils.GEOMETRY_XPATH), is("/data/location1"));
     }
+
+    @Test
+    @SuppressWarnings("PMD.DoNotHardCodeSDCard")
+    public void simplifyScopedStoragePathTest() {
```
seadowg, line 282: 👍. I had always though `is` was equivelant to `==` rather than just an alias of `is(equalTo())`. Will use in future!

mmarciniak90:
@grzesiek2010, please, take a look at this again. There are 2 places where the layers directory path is visible. I see that you fixed it in the lower message. This path is visible also under each name of the layer. Before storage migration, path contained a full directory path in **both places**. 

|Before migration | After migration without changes | After migration with changes|
|---|---|---|
|Screenshot_20200326-134822|Screenshot_20200326-134850|Screenshot_20200326-135052|

grzesiek2010:
Yes but it was shorter I don't know if it makes sense to keep full paths in both places @lognaturel what do you think?

lognaturel:
Good catch, @mmarcianiak90. I think it’s important to put the full path in the message below because that one is a hint about where to put the files if none are available. We could just do /layers under each option but I prefer that it be consistent.

grzesiek2010:
> but I prefer that it be consistent

ok it's done.

mmarciniak90:
**Tested with success**

Verified cases:
* offline layer before migration - correct path under option and in lower message
* offline layer after migration- correct path under option and in lower message
* correct source in lower message
* additional directory in /layer 
* the offline map list in Geo widget
* the offline map list on a view that display form submissions on map
* English/Spanish language
* light/dark mode

|Screenshot_20200327-143421|Screenshot_20200327-143216|
|---|---|

PR #3749, opened 2020-03-30
Keep logging analytics in one class

grzesiek2010:
I reviewed implemented changes and also tested some random places where we log events using adb (according to the documentation: https://firebase.google.com/docs/analytics/get-started?platform=android)

Some time ago we have added `FirebaseAnalytics` class and moved methods responsible for logging events there but we have never cleaned the whole project so we ended up with duplicated code in the Collect class.

I'm not sure if testing is required here. Maybe just reviewing from another contributor would be enough? I don't find this pr very risky.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -430,6 +430,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 .of(this, changesReasonViewModelFactory)
                 .get(FormSaveViewModel.class);
 
+        formSaveViewModel.setAnalytics(analytics);
         formSaveViewModel.getSavedResult().observe(this, this::handleSaveResult);
     }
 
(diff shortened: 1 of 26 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -430,6 +430,7 @@ private void setupViewModels() {
                 .of(this, changesReasonViewModelFactory)
                 .get(FormSaveViewModel.class);
 
+        formSaveViewModel.setAnalytics(analytics);
```
seadowg: This could be constructor dependency right? Then there doesn't need to be any nullability risk in the object.

seadowg (changes requested):
Just one wee change!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Right, done.

seadowg (approved):

PR #3752, opened 2020-04-01
Allow null select one values

### LINKED ISSUE #3142: NullPointerException in SelectOneListAdapter.java:105
Collect v1.22.1 and older versions

#### Problem description
`Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
       at org.odk.collect.android.adapters.SelectOneListAdapter.setUpButton(SelectOneListAdapter.java:105)
       at org.odk.collect.android.adapters.SelectOneListAdapter.setUpButton(SelectOneListAdapter.java:38)
       at org.odk.collect.android.adapters.AbstractSelectListAdapter$ViewHolder.bind(AbstractSelectListAdapter.java:216)
       at org.odk.collect.android.adapters.SelectOneListAdapter$ViewHolder.bind(SelectOneListAdapter.java:90)
       at org.odk.collect.android.adapters.AbstractSelectListAdapter.onBindViewHolder(AbstractSelectListAdapter.java:82)
       at org.odk.collect.android.adapters.AbstractSelectListAdapter.onBindViewHolder(AbstractSelectListAdapter.java:60)`

Try to use this form: 
[selectOne.xml.txt](https://github.com/opendatakit/collect/files/3273143/selectOne.xml.txt)

seadowg:
Closes #3142

This makes it possible to have null select choice values in select one. This not working was causing a crash. While it's not common it is possible for JavaRosa's `SelectChoice#getValue()` to return `null`. 

The behaviour we now expect is as follows:
1. Enumerator selects choice that has no underlying value
2. Choice looks selected
3. Enumerator swipes to another question and swipes back
4. There is no choice selected

This is a bit confusing but really a result of strange/bad form design so we're going to go with it for the moment.

Wrote tests for `null` case for `SelectChoice` values. 

We initially considered hiding the value but it felt like the simplest thing to do was to let this kind of form exist as-is but make sure it wouldn't crash.

Should just fix the crash. Good to have a play around with selects.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/widgets/SelectOneWidgetTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/widgets/SelectOneWidgetTest.java
@@ -36,6 +36,7 @@ import java.util.List;
 
 import static java.util.Arrays.asList;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
@@ -43,12 +44,13 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.odk.collect.android.support.CollectHelpers.setupFakeReferenceManager;
 import static org.odk.collect.android.support.RobolectricHelpers.populateRecyclerView;
+import static org.robolectric.shadows.ShadowView.innerText;
 
 /**
  * @author James Knight
  */
 
-public class SelectOneWidgetTest extends GeneralSelectOneWidgetTest<AbstractSelectOneWidget> {
+public class SelectOneWidgetTest extends GeneralSelectOneWidgetTest<SelectOneWidget> {
 
     @NonNull
     @Override
@@ -85,7 +87,7 @@ public class SelectOneWidgetTest extends GeneralSelectOneWidgetTest<AbstractSele
                 ))
                 .build();
 
-        populateRecyclerView(getActualWidget());
+        populateRecyclerView(getWidget());
         verify(audioHelper).setAudio(any(AudioButton.class), eq(new Clip("i am index 0", REFERENCES.get(0).second)));
         verify(audioHelper).setAudio(any(AudioButton.class), eq(new Clip("i am index 1", REFERENCES.get(1).second)));
     }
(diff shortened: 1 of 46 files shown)


collect_app/src/test/java/org/odk/collect/android/widgets/SelectOneWidgetTest.java
```diff
@@ -104,29 +106,21 @@ public void whenChoicesHaveAudio_logsAudioChoiceEvent() throws Exception {
...
-            public ReferenceManager providesReferenceManager() {
-                return referenceManager;
-            }
-
-            @Override
-            public AudioHelperFactory providesAudioHelperFactory() {
-                return context -> audioHelper;
-            }
+    @Test
+    public void whenAChoiceIsBlank_itIsShownAsBlank() {
```
lognaturel: I don't understand this test! The test that is introduced should fail or crash at this point and this one passes. I don't understand "shown" in its name. This issue is only about the underlying value, not about what is displayed, right? If you were going to test what is displayed, `innerText` on `getChildAt(0)` wouldn't make sense because that gives you the linear layout that contains the audio-video-image-text-label view.

I think you probably want to do something like select the first choice and verify that its value is blank. Again, it should fail or crash in this commit.

collect_app/src/test/java/org/odk/collect/android/widgets/SelectOneWidgetTest.java
```diff
@@ -123,6 +123,22 @@ public void whenAChoiceIsBlank_itIsShownAsBlank() {
...
+
+        formEntryPrompt = new MockFormEntryPromptBuilder()
+                .withSelectChoices(asList(
+                        selectChoice
+                ))
+                .build();
+
+        SelectOneWidget widget = getWidget();
+        populateRecyclerView(widget);
+        assertThat(innerText(widget.getChoicesList().getChildAt(0)), equalTo(""));
```
lognaturel: This test would crash without the code change so that's good. But I still don't think it's asserting something that makes sense.

lognaturel (changes requested):
Code change looks good but I don't think the tests are testing what you want!

collect_app/src/test/java/org/odk/collect/android/widgets/SelectOneWidgetTest.java
lognaturel: I see from looking at the other commit that you seem to have intentionally introduced a test for a blank value that was intended to pass.

seadowg: Yeah I wanted to check the blank and `null` cases. It's calling `widget.getChoicesList().getChildAt(0)` so it is checking what is displayed for that select choice. That being said I agree that isn't really the right test here. I think when I wrote this (a wee while ago) I confused value and label. As you say the "feature" we're actually trying to drive out here is that when you select a blank/`null` value you get a blank/`null` answer in the widget. Rest assured a forehead has been slapped and this will be changed.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Crash does not occur

Verified cases:
- mentioned form filled a few times
- filled a few other forms with eg.:
   - select widget with an autocomplete appearance
   - select widget with a quick appearance
   - select widget - read-only
   - external secondary instance

Issue #3016, opened 2019-04-12
Deleting a repeat does not result in calculates that have count() of the number of repeats


grzesiek2010:
#### Software and hardware versions 
Collect v1.21.0 and older versions

#### Problem description
Deleting a repeat does not result in calculates that have count() of the number of repeats.

The problem has been reported in this issue: https://github.com/opendatakit/collect/issues/2942#issuecomment-480468517

#### Steps to reproduce the problem
1. Open the attached form: 
[countRepeatTest.xml.txt](https://github.com/opendatakit/collect/files/3074617/countRepeatTest.xml.txt)
2. Ad a few groups (let's say 3).
3. Navigate to the last question which should show number 3.
4. Navigate back and remove one group.
5. Navigate forth - the number is still 3.

lognaturel:
@grzesiek2010, can you please give a little bit of commentary when you pick up an older issue that's not part of the milestone so that we can have the context about why it's coming back up and when it should be reviewed?

lognaturel:
I should add. ~For this specific issue, make sure to review https://github.com/getodk/javarosa/blob/master/dag.md#c---a-repeat-instance-is-deleted.~ I now see that the count expression is outside of the repeat so it's not related to that section. It would be good to also check against JavaRosa master.

Based on why this is coming back up, we can discuss when it should be prioritized.

grzesiek2010:
Sorry, I know it's not the best approach but I was blocked with a few other tasks so I went through the list of existed issues.
The description here is not perfect in fact the problem occurs only if we delete the last group because then recalculation is not triggered. I created a draft pr based on my investigation https://github.com/getodk/javarosa/pull/572 so you can take a look.

lognaturel:
Ahh, great. Sorry about the blocking and good to know that's where you're at! Picking up an old task is totally great, it's just helpful to have a little bit of context on why and what initial findings are.

lognaturel:
Turns out this is a good time to take this because I have recently been in that code myself and can provide thoughtful review without needing to load a whole bunch of context back into my brain. 👍

grzesiek2010:
Fixed in https://github.com/getodk/javarosa/pull/572

Issue #498, opened 2017-03-02
Google account options should refresh when 'Google account' is tapped


lognaturel:
#### Software and hardware versions 
Collect v1.4.16

#### Problem description
After adding a Google account to the device, it is not available in the Google account selection menu.

#### Steps to reproduce the problem
1. Set the platform to Google Drive, Google Sheets
1. Tap Configure platform settings
1. Go to the device Settings > Account > Google and add a Google account
1. Go back to ODK Collect
1. Tap Google account and the new account doesn't appear
1. Tap back to return to the general settings
1. Tap Configure platform settings
1. Tap Google account and the new account appears

#### Expected behavior
The account list should be refreshed when Google account is tapped.

rishsethia:
Hi @lognaturel I am trying to solve this !
I will send a PR when solved !
Thanks

Dvik:
it's probably because we should get the accounts in `onResume()` rather than `onCreate()`. 

@rishsethia are u working on it or should i send a PR as i've almost set up the code to do so?

rishsethia:
@Dvik 
You can go ahead since you've set up the code ! I was looking to solve it after finishing my current task !
All the best ! :D

PR #3753, opened 2020-04-01
Update JavaRosa to v2.17.1

lognaturel:
This PR fixes various bugs that were reported by users and recently addressed in JavaRosa. Several of these fixes came in just as the Collect release was getting prepared so it wasn't a good time to update. JavaRosa has gone through a lot of changes so instead of releasing off master, I'm proposing we do a point release with only low-risk bug fixes. 

Addresses opendatakit/javarosa#131, opendatakit/javarosa#509, opendatakit/javarosa#540, opendatakit/javarosa#498, opendatakit/javarosa#547

The branch this PR uses a JavaRosa snapshot. The sequence I propose here is:
- [x] @seadowg or @grzesiek2010 reviews [the JavaRosa diff](https://github.com/opendatakit/javarosa/compare/v2.17.x) with the notes I've made below
- [x] QA verifies the fixes and regression risks
- [x] We iterate on JavaRosa as needed
- [x] We release JR v2.17.1
- [ ] We update this PR to use the final release
- [ ] Someone does a quick sanity check to make sure that JR v2.17.1 matches the last snapshot code that was QAed and merges
- [ ] We do a Collect point release off a v1.26.x branch also including the fix for https://github.com/opendatakit/collect/issues/3751.

opendatakit/javarosa#131 / https://github.com/opendatakit/javarosa/commit/efe171f: exclusively an error message change so should pose no risk. It could help with troubleshooting so I figure it's good to get out.

opendatakit/javarosa#509 / https://github.com/opendatakit/javarosa/commit/5fc1012, https://github.com/opendatakit/javarosa/commit/6cb87e4, https://github.com/opendatakit/javarosa/commit/3f22f7e: only the first commit has a code change and it is limited to the `count-not-empty` function. The other two commits are tests. I see no regression risk with this one.

opendatakit/javarosa#540 / https://github.com/opendatakit/javarosa/commit/8bc3362: I cherry-picked https://github.com/opendatakit/javarosa/commit/26cdfa2838ce1768f301411eb365e880579a01aa and fixed spacing conflicts. Then I verified the diff against master. I didn't bring the tests in because the ones added in https://github.com/opendatakit/javarosa/pull/546 rely on new testing infrastructure. This one is riskier and there's a regression risk around any relative path expression with parenting steps (`../`) and multiplicities. However, I think the logic is straightforward enough to review. The only change is the addition of the `if (i + refLevel <= newRef.size())` condition. This means that the behavior only changes in the case where `i + refLevel > newRef.size()` in which case multiplicities aren't copied. I think the tests document this sufficiently. We know that this only affects forms with nested repeats which are relatively rare. However, this bug is a showstopper for those who want to use complex expressions in nested repeats.

opendatakit/javarosa#498 / https://github.com/opendatakit/javarosa/commit/0fe85cf: I cherry-picked https://github.com/opendatakit/javarosa/commit/eb8cfd1f because the tests added by https://github.com/opendatakit/javarosa/pull/551 rely on new testing infrastructure. It's similar in nature to the one above in that it will only affect relative references and these will only be generated by pyxform for nested repeats. I think the regression risk here is even lower because `contextualize` does nothing with absolute references. The tests are also comprehensive (https://github.com/opendatakit/javarosa/pull/551/files#diff-68d926706ca036914f632b04878da9ca). Still, there's a regression risk to the `indexed-repeat` function and it would be good to verify it in single repeats and nested repeats.

opendatakit/javarosa#547 / https://github.com/opendatakit/javarosa/commit/4fdc453, https://github.com/opendatakit/javarosa/commit/6b28fd5: only the first commit changes code. The second removes a test that relies on updated test infrastructure. This is like the `count-not-empty` change above. It only changes the `digest` function so that should be the only place where there's risk of regression.

I have run all tests and verified that they pass. I verified 
[dynamic-digest.xml 2.zip](https://github.com/opendatakit/collect/files/4417960/dynamic-digest.xml.2.zip) to make sure that the `digest` function works as expected.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -230,7 +230,7 @@ dependencies {
     implementation "com.rarepebble:colorpicker:3.0.1"
     implementation "commons-io:commons-io:2.6"
     implementation "net.sf.opencsv:opencsv:2.4"
-    implementation("org.opendatakit:opendatakit-javarosa:2.17.0") {
+    implementation("org.opendatakit:opendatakit-javarosa:2.17.1-SNAPSHOT") {
         exclude group: 'joda-time'
         exclude group: 'org.slf4j'
     }


seadowg:
@lognaturel I've had a look through and haven't spotted anything that scares me. Looks like unit tests are passing and I'll give this a run through test lab as well.

seadowg:
All green on test lab!

lognaturel:
🎉 @seadowg Are you comfortable with this going to QA, then?

seadowg:
@lognaturel yeah seems good to get there as soon as we can!

kkrawczyk123:
@lognaturel I have started testing this pr on Androids 5.1 and 10 and I have already tested some cases like:
- digest function form
- count-non empty form
- indexed repeats forms, nested repeats forms 
I think I must have misunderstood something or maybe do something wrong as I didn't notice any differences between this pr and master when using the form attached in https://github.com/opendatakit/javarosa/issues/498 Should I change something in the form manually?

I also have a question, how this one https://github.com/opendatakit/javarosa/issues/540 should be verified. 

Definitely testing should be continued here with @mmarciniak90 and her devices :)

lognaturel:
@kkrawczyk123 I've updated the form for #498 at https://github.com/opendatakit/javarosa/issues/498#issuecomment-608754074 so that hopefully it makes a little more sense. If you can, it'd be great for you to build your own form with a nested repeat that uses `indexed-repeat` and/or look at the possible reproduction from the forum. The tricky thing is that you won't be able to use the online form converter to convert an XLSForm with `indexed-repeat` in a nested repeat because it won't pass validation. That's why I had used another expression in the calculation for form conversion and then replaced it manually in the XML. If you already have `pyxform` available locally or if you have Python installed and can do `pip install pyxform`, you could run the conversion with `--skip_validate`. More at https://github.com/XLSForm/pyxform/

As I was getting you a good form to verify opendatakit/javarosa#540, I realized that the fix doesn't get all cases so thanks for your help discovering that. 😊 I have issued https://github.com/opendatakit/javarosa/pull/565. @seadowg since we want to get this out quickly, I've tagged you there for review in case @ggalmazor doesn't have time. I don't think it will require a ton of context to verify.

seadowg (dismissed):
Sorry forgot to mark this when I reviewed it last week!

kkrawczyk123:
@lognaturel thanks for the tips, I have run through it once again. I was able to reproduce the issue with the form that you have attached and also another with nested repeats and indexed-repeat on master. I was not able to reproduce on PR, I haven't noticed any regression too. Please let me know if there is anything more on your mind that should be also verified here.

@mmarciniak90, go ahead!

lognaturel:
A user just reported the additional issue with opendatakit/javarosa#540 at https://forum.opendatakit.org/t/infinite-loop-and-dynamically-delete-already-selected-options-from-list/25902/3 so it's great we're getting this ready to ship soon. The additional fix has been merged and the snapshot updated so please try it out. I encourage you to try the form shared on the forum as well. Thank you!

mmarciniak90:
@lognaturel here are the results of my today's tests. Please take a look at 540 because it's still a challenge for us and maybe you will be able to help us.

---

* opendatakit/javarosa#131

   * form: [internal_empty_select.xml.txt](https://github.com/opendatakit/collect/files/4444574/internal_empty_select.xml.txt)
   * Error contains question name without choices

|master| pull request|
|---|---|
|Screenshot_2020-04-07-15-21-09-132_org odk collect android||

---
* opendatakit/javarosa#509

   * form : [count-non-empty.xml.txt](https://github.com/opendatakit/collect/files/4444113/count-non-empty.xml.txt)
      * Error does not appear
      * User is able to add group
      * On `Other info` question, empty and filled groups are counted
---
* opendatakit/javarosa#540
   
   * form: [indefinite.xml.txt](https://github.com/opendatakit/collect/files/4444124/indefinite.xml.txt)
      * Form from the forum was converted but it's not possible to fill it. 
      * @lognaturel we see this error and we are not sure how to fix it.


---
* opendatakit/javarosa#498

   * form: [indexed-repeat-nested.xml.txt](https://github.com/opendatakit/collect/files/4444115/indexed-repeat-nested.xml.txt)
      * Error does not appear
      * Questions have auto-filled answers.

   * form: [nestedTest.xml.txt](https://github.com/opendatakit/collect/files/4444117/nestedTest.xml.txt)
      * Error does not appear
---
* opendatakit/javarosa#547

   * form: [dynamic-digest.xml.txt](https://github.com/opendatakit/collect/files/4444119/dynamic-digest.xml.txt)
      * Error contains more information
      * I saw two different errors - I was not able to detect why

|master |pull request| pull request|
|---|---|---|
|Screenshot_2020-04-07-14-44-13-637_org odk collect android|||

lognaturel:
As we discussed in Slack, the snapshot had not actually been updated which explains the existing error with opendatakit/javarosa#540. It has been updated now.

The middle digest issue is expected -- it's stating that the 'aa' algorithm isn't supported. I can't reproduce the third error. It's a strange one because it's implying an issue with the `if` function call. Unless you can reproduce it systematically, I don't think there's much we can do about it.

lognaturel:
@mmarciniak90 pointed out that there's still a problem related to opendatakit/javarosa#540. I believe I have fixed it and updated the snapshot. I just tried this branch with the forms from https://forum.opendatakit.org/t/infinite-loop-and-dynamically-delete-already-selected-options-from-list/25902 and https://forum.opendatakit.org/t/error-in-indefinite-repeat-group-inside-another-repeat-group/24729 and I was able to get them to work. More details on how to run these forms, especially the one in French, on Slack. Take a look and see if you can come up with any other creative ways to break it!

mmarciniak90:
Yeah! I was able to open forms!
 * [indefinite-repeat-inside-other-repeat.xml.txt](https://github.com/opendatakit/collect/files/4455516/indefinite-repeat-inside-other-repeat.xml.txt)
* [indefinite.xml.txt](https://github.com/opendatakit/collect/files/4455540/indefinite.xml.txt)

But... I see a crash when I click on the JUMP icon to open the hierarchy view.

`04-09 12:05:55.661 25230-25230/org.odk.collect.android E/ExtMediaPlayer-JNI: env->IsInstanceOf fails
04-09 12:05:55.661 25230-25230/org.odk.collect.android E/MediaPlayer-JNI: JNIMediaPlayerFactory: bIsQCMediaPlayerPresent 0`

@seadowg, @grzesiek2010  any ideas?

seadowg:
Hmmm not sure about that error. Can have a look with the form later if @grzesiek2010 doesn't get to it first.

grzesiek2010:
> if @grzesiek2010 doesn't get to it first.

I've just reproduced it but there is nothing obvious in logs at first glance. You can take a look if you want.

lognaturel:
From Slack: I did some interactive review with @seadowg of the last chunk of work that I thought would fix javarosa#540 and it turns out there are several problems with it. It does work for the sample forms but it's not a general or correct solution, unfortunately.

I'm going to do a point release without the JR update so we can get #3755 out. Then we can do a separate point release with the JR change when ready.

seadowg (changes requested):
Sorry forgot to update the status here!

lognaturel:
The snapshot has now been updated with the latest updates to fix getodk/javarosa#540 reviewed and approved by @seadowg. I have verified this by trying the species form from the forum. I have also run all tests on the v1.17.1 JR branch and here.

Here are some additional test forms to try:
- the original ones from the issue
- [indefinite-repeat-nested-relative.xml.zip](https://github.com/getodk/collect/files/4542193/indefinite-repeat-nested-relative.xml.zip) expected behavior: outer repeat has one instance initially and user is prompted to add more. Inner repeat has one instance initially and another one is added if user types "yes" (all lowercase). Inner repeat exists if any other text is included.
- [predicate-trigger-change.xml.zip](https://github.com/getodk/collect/files/4542216/predicate-trigger-change.xml.zip) expected behavior: specify a number for the first question. On the second screen you should see a sum of numbers 1-6 that are greater than the number from the first question. For example, enter 5 and you should see 6. Only the first outer repeat has pre-populated values 1-6. You can add another outer repeat and specify your own inner repeat values to verify the same behavior.
- [predicate-current.xml.zip](https://github.com/getodk/collect/files/4542226/predicate-current.xml.zip) similar to above but there are always exactly 5 items that come from an external list and the list labels are concatenated. For example, enter in 4 and you should get "Item5" or enter in 2 and you should get "Item2, Item3, Item4, Item5"

These might give you other ideas of form types to try.

mmarciniak90:
@lognaturel I added noticed issues and thoughts.

All cases were verified on Android 6.0, 7.0, 8.1, 9.0

Verified cases:
* forms:
   * _indefinite-repeat-inside-other-repeat.xml_ and _indefinite.xml_
      *  **I noticed a crash mentioned previously - when I click on the JUMP icon to open the hierarchy view.**
      * when I reopen form after the crash I have another crash because the app wants to recover data and open hierarchy view - and it happens again and again. Clearing cache is the way to fix this vicious cycle.
   * _indefinite-repeat-nested-relative.xml_
   * _predicate-trigger-change.xml_
       * **the sum is not re-counted dynamically after removing repeat for repeats added manually**
       
   * _predicate-current.xml_
      * > For example, enter in 4 and you should get "Item5" or enter in 2 and you should get "Item2, Item3, Item4, Item5"
       * I see a small difference but I guess there is misleading in the description - When I enter 2 I see "Item3, Item4, Item5"
* removing repeats
* removing repeats via trash icon in the hierarchy view
* opening hierarchy view

lognaturel:
Thank you, @mmarciniak90!

> I noticed a crash mentioned previously - when I click on the JUMP icon to open the hierarchy view.

Yes, that's expected. The infinite loop happens on finalization as well.

> the sum is not re-counted dynamically after removing repeat for repeats added manually

Is that only the case if you delete the last repeat? I suspect you're experiencing #3016

> When I enter 2 I see "Item3, Item4, Item5"

Whoops, yes, that's exactly right.

mmarciniak90:
> Is that only the case if you delete the last repeat? I suspect you're experiencing #3016

Yes, you are totally right. It's not related to manually added repeat but to the last one! Sorry!

I retested cases described in https://github.com/getodk/collect/pull/3753#issuecomment-610396844
I don't have anything more to add to this PR. 

@lognaturel, I remember you mentioned it's high priority PR because of a point release. Testing on 4 devices from 6 is finished. We can wait for Kasia's confirmation or check it even after the merge (Android 5.1, 10.0). On the other hand, this PR seams to be quite risky.

lognaturel:
Thank you, @mmarciniak90! I wouldn't want to release on a Friday anyway so the earliest we would do the point release is your Monday evening. If @kkrawczyk123 can take a quick look before then that'd be great. There's nothing dependent on Android version here so I think it's more about getting another brain to review the cases you and I have tried and see if any other ideas come up.

Happy May Day!

kkrawczyk123:
I haven't found anything weirder that @mmarciniak90 has mentioned above on Androids 5.1 and 10. I have checked forms attached and also made some regression checks with all widgets and repeats forms - so I guess we have success here!

seadowg (approved):

PR #3754, opened 2020-04-02
Reorder prefs

lognaturel:
I have been spending some time looking at our preference keys while iterating on https://forum.opendatakit.org/t/proposal-publish-a-settings-key-value-standard/25142 and improving documentation at https://docs.opendatakit.org/collect-import-export/#list-of-keys-for-all-settings.

It would be really nice for the key order to be consistent everywhere. Matching the UI seems to make the most sense. That's what this PR attempts to do.

Questions for reviewer:
- is this a sensible order?
- should the order be explicitly documented in a comment or is that just likely to never be read?

I don't think there's any risk because if I removed a key things just wouldn't compile. I used copy/paste for the moves so there's no risk that I changed an underlying value. I think this can just be merged without QA.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/AdminKeys.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/AdminKeys.java
@@ -25,45 +25,41 @@ public final class AdminKeys {
     public static final String KEY_DELETE_SAVED                 = "delete_saved";
     public static final String KEY_QR_CODE_SCANNER              = "qr_code_scanner";
 
-    // form entry
-    public static final String KEY_SAVE_MID                     = "save_mid";
-    public static final String KEY_JUMP_TO                      = "jump_to";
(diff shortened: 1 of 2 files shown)


seadowg (approved):
Personally I'm not sure if having them in any order is better than any other. If I'm creating a new one I'm probably ALT+Entering my way into the class, if I'm looking to see the definition I've probably CMD+Clicked my way in and if I want to find one related to a word I know I'm probably searching in the file. I can see that for your case it certainly made sense to have them in "UI order" and that's the only time that I can think of it being useful to need an order so yeah I'm with you on it being sensible!

Yeah I don't think a comment makes sense - all that IDE tooling is going to prevent people from seeing it. What might be nice is JavaDoc on each key that describes what the preference actually is (similar to what we're doing in `AnalyticsEvents`) but I'm not going to make you do that!

PR #3763, opened 2020-04-03
Handle exceptions that might occur when a user tries to open a browser

grzesiek2010:
Reviewed and tested manually with hardcoded exceptions.

We agreed that the crash when opening the forum from `StorageMigrationDialog` is not serious but since we are about to have a point release and another similar crash came up:
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/84debbdb9ff2c7dfb84c961e3b54bc61?time=last-seven-days&sessionId=5E8451ED02FE000123A7DE9A2D5016BF_DNE_7_v2
I think it's worki merging and adding to v1.26.2

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationDialog.java
@@ -1,6 +1,7 @@
 package org.odk.collect.android.storage.migration;
 
 import android.app.Activity;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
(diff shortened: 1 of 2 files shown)


grzesiek2010:
Just to clarify the `SecurityException` might occur when a user tries to complete `Intent.ACTION_VIEW` not using a browser but another app that might throw it, here is I think the same case like our: https://groups.google.com/forum/#!topic/google-admob-ads-sdk/CvmDj2VTpwA

seadowg (approved):

mmarciniak90:
**Tested with success**

I was able to reproduce the crash on 2 devices (Android 6.0 and 8.1) where I was able to disable the browser.

After the fix, the crash does not occur. The forum page is opened.

PR #3766, opened 2020-04-06
 Make sure to check for failed constraints when saving screen answers 

seadowg:
Before when saving screen answers using `saveAllScreenAnswers` we weren't checking for non-null `FailedConstraint` return values so failures wouldn't show in the UI. This was cause 3 Espresso tests to fail.

I also changed the assertion we use to check for Toasts as the failure message was quite confusing.

Failing tests now pass

Should just fix the issue. Probably good to check forms with constraints to see if there is anything that the tests aren't catching here. I'd lean towards merging this and then doing that QA on master however.


grzesiek2010 (approved):

lognaturel:
Wow, good thing there were tests to catch that. 👍

PR #3767, opened 2020-04-07
Factored out QuestionFontSizeProvider

grzesiek2010:
I reviewed implemented changes and tested changing font size manually.

@seadowg 
In most cases we use object we pass/inject. In this case I just created an utility class with a static method because I don't think it's something we need test (mock) so this approach is easier and we will avoid passing many object via constructors.
...but let me know if your opinion differs.

It shouldn't affect/change anything it's refactoring that should just make the code a bit cleaner. Quick testing changing font size would be enough.

No.

No.


collect_app/src/main/java/org/odk/collect/android/utilities/QuestionFontSizeProvider.java
```diff
@@ -0,0 +1,21 @@
+package org.odk.collect.android.utilities;
+
+import org.odk.collect.android.preferences.GeneralSharedPreferences;
+
+import static org.odk.collect.android.preferences.GeneralKeys.KEY_FONT_SIZE;
+
+public class QuestionFontSizeProvider {
```
seadowg: This should probably just be `QuestionFontSizeUtils` now

seadowg (approved):
I was in two minds about this being static or not. I think if we were testing that a widget is using it (which we really should be) we'd probably want to write something like this:

```java
assertThat(widget.getSomeTextView().getTextSize(), equalTo(QuestionFontSizeUtils.getQuestionFontSize());
```

This seems nicer than faking the font size then asserting on the fake or (worse still) verifying we call `getQuestionFontSize` so the static feels right here.

collect_app/src/main/java/org/odk/collect/android/utilities/QuestionFontSizeProvider.java
grzesiek2010: Right, fixed.

seadowg (approved):

mmarciniak90:
**Tested with success**

> It shouldn't affect/change anything it's refactoring that should just make the code a bit cleaner. Quick testing changing font size would be enough.

I can confirm that I didn't notice any change.

Verified cases:
* change the font size before start filling a form
* change the font size while filling a form
* scan QR code with extra small font size (without emulator)
* range picker reacts on size change
* `All widgets` form

PR #3770, opened 2020-04-08
Don't register DateWidget and TimeWidget separately if both are part of DateTimeWidget

### LINKED ISSUE #3761: Remove response option is doubled on Date time widget

#### Problem description
Remove response option is doubled on Date time widget

1. User fills `Date time widget`, in eg. All widgets form
2. User opens dialog to remove response in a line with added date or time
3. User sees doubled option `Remove response` but only the lower is working

| Screenshot_20200403-112159|Screenshot_20200403-111919|Screenshot_20200403-112214|
|---|---|---|

#### Expected behavior
`Remove response` option is not doubled

grzesiek2010:
Closes #3761

I tested the implemented changes manually.

It's just a bug fix. `DateTimeWidget` consists of `DateWidget` and `TimeWidget` so `registerToClearAnswerOnLongPress()` was called multiple times.

We need to test removing answers from: `DateTimeWidget`, `DateWidget` and `TimeWidget`, that would be enough.

All widgets form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java
@@ -42,8 +42,8 @@ public class DateTimeWidget extends QuestionWidget implements BinaryWidget, Widg
     public DateTimeWidget(Context context, QuestionDetails prompt) {
         super(context, prompt);
 
-        dateWidget = new DateWidget(context, prompt);
-        timeWidget = new TimeWidget(context, prompt);
+        dateWidget = new DateWidget(context, prompt, true);
+        timeWidget = new TimeWidget(context, prompt, true);
 
         dateWidget.getAudioVideoImageTextLabel().getLabelTextView().setVisibility(GONE);
         dateWidget.getHelpTextLayout().setVisibility(GONE);
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/DateWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/DateWidget.java
@@ -77,7 +77,11 @@ public class DateWidget extends QuestionWidget implements DatePickerDialog.OnDat
     private DatePickerDetails datePickerDetails;
 
     public DateWidget(Context context, QuestionDetails prompt) {
-        super(context, prompt);
+        this(context, prompt, false);
+    }
+
+    public DateWidget(Context context, QuestionDetails prompt, boolean isPartOfDateTimeWidget) {
+        super(context, prompt, !isPartOfDateTimeWidget);
         createWidget(context);
     }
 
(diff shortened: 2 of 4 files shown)


grzesiek2010:
@seadowg I was thinking about tests here. Ideally not UI. I wanted to verify that `registerToClearAnswerOnLongPress()` is called for every widget just once. Unfortunately it's not possible because it's called when the object is constructed so it would require some changes/workarounds.
A UI test of course would be slower so that's why I wanted to avoid it but on the other hand it also could test removing answers so cover bugs like: https://github.com/opendatakit/collect/pull/3755 as well.
Pros and cons, what do you think?

seadowg:
Maybe instead of calling it in `QuestionWidget`'s constructor we could call it in `ODKView#configureWidgetForQuestion`. That seems be currently where we set the widget up with listeners to its parents. To make it testsable I think we'd need to change the design a little 

* have `WidgetFactory.createWidgetFromPrompt` become an instance method
* pass a `WidgetFactory` instance to a `ODKView` in the constructor

Now we could write a test of ODKView where we pass it a stub `WidgetFactory` that returns a dummy/mock widget that lets us verify `registerToClearAnswerOnLongPress` is called on it.

Does that make sense as a way forward?

grzesiek2010:
> Does that make sense as a way forward?

Seems like to many changes just for one test that wouldn't even allow me to check if the right question is cleared (only whether `registerToClearAnswerOnLongPress()` was called once). I will rather go for espresso I now espresso... but at least I will be able to check both cases not one whether `registerToClearAnswerOnLongPress()` was called.

seadowg:
> Seems like to many changes just for one test that wouldn't even allow me to check if the right question is cleared (only whether registerToClearAnswerOnLongPress() was called once).

Yeah but I guess I'm thinking it might be worth reorganizing the code to make it more testable here. If we do write Espresso tests for concerns that run across the widgets then we need to write an Espresso test for every widget to have coverage. I'd prefer to be able to test the code that carrys out clearing an answer from a widget and also test the code that creates widgets to check it hooks it up with the clearing mechanism.

grzesiek2010:
> Yeah but I guess I'm thinking it might be worth reorganizing the code to make it more testable here.

I agree but since it's not related to just this one issue and would require much more changes than this pr itself I would prefer creating a new issue for that and maybe leaving this one as is (even without tests for now since it's easy to confirm manually). Does that sound good?

seadowg:
Yeah that sounds good. Could you create an issue? We could actually hopefully incorporate it in the Widgets rewrite GSOC project if that goes ahead.

grzesiek2010:
Ok created https://github.com/getodk/collect/issues/3797

grzesiek2010:
Can we add `needs testing` now?

seadowg:
@grzesiek2010 ach sorry. Thought I'd already reviewed.

seadowg (approved):

mmarciniak90:
**Tested with success** 

Verified on Android: 6.0, 7.10, 8.1, 9.0

Verified cases:
- remove response on Date part
- remove response on Time part
- remove response on  first widget part
- widget with no appearance
- widget with no-calendar appearance

Issue #3441, opened 2019-11-08
MemoryLeak in GeoPointActivity


grzesiek2010:
#### Problem description
![Screenshot_1573244675](https://user-images.githubusercontent.com/3276264/68508259-36f67100-026e-11ea-88ab-3f1064b81999.png)

`Instance of android.location.LocationManager$GnssStatusListenerTransport
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$ifTable = java.lang.Object[6]@1880885848 (0x701c0a58)
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$classSize = 260
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$componentType = null
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$numReferenceInstanceFields = 7
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$name = "android.location.LocationManager$GnssStatusListenerTransport"
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$vtable = null
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$referenceInstanceOffsets = 2035
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$numReferenceStaticFields = 0
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$status = -536870912
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$virtualMethodsOffset = 13
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $class$objectSize = 52
2019-11-08 21:23:29.892 8628-8662/org.odk.collect.android:leakcanary D/LeakCanary: |   static $classOverhead = byte[132]@1882941545 (0x703b6869)
[...]

PR #3772, opened 2020-04-09
Use fused location provider everywhere

### LINKED ISSUE #2717: Switch all location capture to FusedLocationProvider
- [x] Implement changes
- [ ] Explore and document how users can force GPS-only results
- [ ] Understand and document implications for external high-precision GPS sensors
- [ ] Document implications for accuracy and location data availability for user-facing docs

`FusedLocationProvider` is the location provider recommended by Google. It's supposed to balance accuracy, resource usage, etc.

Collect used to use it but in Feb 2018, it started capping accuracy at 10m (https://github.com/opendatakit/collect/issues/2008). This turned out to be [an Android bug](https://issuetracker.google.com/u/0/issues/79189573#comment20) which is now fixed. However, accuracy **[is still capped at 3m](https://issuetracker.google.com/issues/118789585)**.

It would probably make sense for the 3m issue to be addressed before switching back to using it.

Regardless, there are some higher-level questions about whether it's the right way to go. It's supposed to provide better battery management but that's only if less accuracy is acceptable. It may make sense to continue only using the GPS since accuracy is almost always important. That said, there are certain environments (urban) where other location sources may be more accurate. Users could force using the GPS by turning other providers off at the Android level, I think, but that may be harder to use.

This may be a conversation more appropriate for the forum. I wanted to at least file the issue so the content and history is not forgotten but feel free to close and start a thread elsewhere or comment here.

### LINKED ISSUE #3313: Verify that correct location provider is visible for Geo widgets
Collect v1.23 with #3309

#### Problem description
During testing #3309 I was wondering if the correct provider is visible for all Geo widgets.
I executed a few testes but precise verification is desirable.

* Internet connection is turned OFF.
* In Android location settings, `Device only` option is set

|Widget|Google Maps|Mapbox|OSM|
|---|---|---|---|
|Geopoint without map apperance|Using GPS|Using GPS|Using GPS|
|Geopoint with map apperance|Using GPS|Using fused|Using GPS|
|GeoTrace|GPS location accuracy:...|GPS location accuracy:...|GPS location accuracy:...|
|GeoShape|GPS location accuracy:...| GPS location accuracy:...|GPS location accuracy:...|

* Internet connection is turned ON.
* In Android location settings, `High accuracy` option is set

Results are the same as in the table above.

lognaturel:
Closes #2717
Closes #3313
Makes progress towards #3441

Some decisions:
- location provider should be communicated explicitly to the user where practical. I didn't add it to geotrace and geoshape because I think those have enough info already and provider wasn't included previously. I could be convinced to remove it from everywhere but I think it could help with troubleshooting.
- accuracy should be displayed before provider information everywhere because it is more important
- keep using the deprecated `FusedLocationProviderApi` because `FusedLocationProviderClient` which is supposed to replace it doesn't work until Google Play Services 11.6.0, released Nov 2017
- `LocationClientListener.onClientStop()` should be called whenever `LocationClient.stop()` is called. I don't think the implementation pre-929dc4f made much sense but I'd like another careful eye on that.

Added tests for the `LocationClientProvider` logic. Verified that `LocationClient` or a Mapbox/Google equivalent is used everywhere that foreground location is requested.

I considered fixing the leaks in another PR but they were getting in my way and I figure they'll be easier to review in this context.

For the `LocationClientProvider` test I considered using Robolectric Shadows Play Services to not have the really strange exception catching pattern but I couldn't immediately get it to work and it didn't feel worth it to add more complexity.

On most Android versions, users won't have any control over what sensors are used to get their current location. From what I have experienced and read, this will be an improvement for almost everyone. However, it could lead to worse performance under some conditions.

The regression risk is focused on geopoint questions without a map but there's also some small amount of risk for all geo questions.

Any form with geo questions.

https://github.com/getodk/docs/issues/1225


collect_app/src/test/java/org/odk/collect/android/location/client/LocationClientProviderTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public Context context;
+
+    @Mock
+    public PlayServicesChecker playServicesChecker;
+
+    @Test
+    public void fusedLocationClient_returnedWhenPlayServicesAvailable() {
+        when(playServicesChecker.isGooglePlayServicesAvailable(any())).thenReturn(true);
+
+        // Could import Robolectric Shadows Play Services to actually build a GoogleFusedLocationClient
```
seadowg: I agree about not pulling in Play Services shadows. One potentially simple way to make this test a little nicer would be to pass a `GoogleAPIClientLocationProvider` interface to `LocationClientProvider` in someway so it could be mocked/faked and not actually call the play services code. That would also let us write tests for `GoogleFusedLocationProvider` without any extra Robolectric stuff. 

So many interfaces though! It's frustrating not having any way to express lambdas in method signatures. #3886 adds a poly fill for Java 8's [Supplier](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html)  (we could be using it as a general "provider" interface) which could make this all a bit simpler. You could think about introducing that here?

collect_app/src/test/java/org/odk/collect/android/location/client/GoogleFusedLocationClientTest.java
```diff
@@ -74,30 +76,29 @@ public void startShouldCallLocationClientOnConnected() {
...
+    public void stopShouldDisconnectFromGoogleApiIfConnected_andAlwaysCallOnClientStopIfListenerSet() {
         TestClientListener testClientListener = new TestClientListener();
         client.setListener(testClientListener);
 
-        // Call through to ApiClient.disconnect(), but don't do anything:
+        // Previously connected, disconnection succeeds
         when(googleApiClient.isConnected()).thenReturn(true);
         client.stop();
+        verify(googleApiClient).disconnect();
+        assertThat(testClientListener.getOnClientStopCount(), is(1));
```
seadowg: Might be easier to use Mockito for this listener verification?

seadowg (changes requested):
Just a couple of changes inline.

It does feel like there might be some missing test coverage around lifecycle hooks performing cleanup etc but I think this is good to get in without looking at that as this code isn't something that there is frequent iteration on. Do you a sense of things that feel uncovered or does it feel good to you @lognaturel? It's hard to get a good sense of these things during a review, so I'd be keen for you to create issues around testing follow-ups if you feel like any are needed.

grzesiek2010:
> keep using the deprecated FusedLocationProviderApi because FusedLocationProviderClient which is supposed to replace it doesn't work until Google Play Services 11.6.0, released Nov 2017

I'm thinking about this. Would that be a problem to check `Google Play Services` version and use `FusedLocationProviderClient` if 11.6.0 or newer? Then in the future we could just remove `FusedLocationProviderApi` once we decide it's safe?

I'm not sure about detailed stats but seems like it's a small group that uses older `Google Play Services` versions so depriving the rest of our users of newer (better) provider seems strange.

lognaturel:
> depriving the rest of our users of newer (better) provider seems strange.

I should have been clearer. As far as I can tell `FusedLocationProviderClient` and `FusedLocationProviderApi` will access whatever the fused location provider implementation is for the device's version of Play Services so the location behavior should be identical. The difference is in [how the connection with Play Services is established](https://stackoverflow.com/a/54781254/137744). `FusedLocationProviderClient` simplifies all of that significantly but since we already have a working implementation that has been previously released, the benefits seem small to me.

> Would that be a problem to check Google Play Services version and use FusedLocationProviderClient if 11.6.0 or newer?

Good thinking. Given that the benefit of switching to `FusedLocationProviderClient` would be simplicity, having two implementations doesn't seem like a win.

> I'm not sure about detailed stats but seems like it's a small group that uses older Google Play Services versions

I've been trying to get info on this. Did you get anything concrete or is it a gut feel based on how Play Services updates? I've seen folks on metered internet connections turn off auto-updates. It used to be that they weren't turned on automatically for some devices but maybe that's only for old Android versions, I'm not sure.

collect_app/src/test/java/org/odk/collect/android/location/client/GoogleFusedLocationClientTest.java
lognaturel: Agreed and done. I made all the tests in that class consistent. I didn't touch `AndroidLocationClientTest` which still uses `TestClientListener`.

collect_app/src/test/java/org/odk/collect/android/location/client/LocationClientProviderTest.java
lognaturel: Hopefully 87439ff3 is roughly what you had in mind. There are so many levels of provider client client providers here it's feeling a little silly but I agree the test looks much better.

grzesiek2010:
> Did you get anything concrete or is it a gut feel based on how Play Services updates?

I didn't, it was just a guess.

Ok thanks for the explanation.

lognaturel:
Thanks for the conversation, @grzesiek2010, it's been helpful to think through what the meaning and value of these different APIs is. @seadowg and I discussed a little further in our weekly sync and are currently thinking that for cases like this where an API is deprecated in favor of one that doesn't provide more functionality but just is easier to use, we can delay upgrading until we run into a problem with the current implementation or new functionality gets added to the modern API.

> missing test coverage around lifecycle hooks performing cleanup etc

I think that `GoogleFusedLocationClientTest` has us pretty well covered. The `GeoPointActivity` should probably have tests around the timer behavior and the satellite count update. In fact, I'm wondering whether the satellite count update should go to `LocationClient`. I'll file a follow-up issue.

lognaturel:
@seadowg Could you please take another look? Let's prioritize this one so we can get the early beta out. The JavaRosa update has already been merged: https://github.com/getodk/collect/pull/3908

seadowg:
@lognaturel ach sorry I missed this today. I've set up a reminder for myself tomorrow and I'll review it first thing in the morning.

seadowg (approved):

mmarciniak90:
@lognaturel @seadowg 

**At the beginning of testing, I started comparing fix to the released version. Unfortunately, the released version was installed on Android 6.0. In the middle of testing, I decided to replace versions installed on devices to improve test value. After that, I noticed that I'm not able to use the app with the fix on Android 6.0. App crashes at the beginning. I was able to reproduce the crash on  emulator with Android 6.0** 
Screenshot from 2020-06-10 14-39-56

---
General:
- Accuracy is visible as a first in the list, before the provider in GeoPoint widget. No change in GeoShape and GeoTrace widgets :heavy_check_mark: 
- MemoryLeaks are still noticeable - after merge we should update #3441
- I was not able to verify all cases on all devices. It was more comparing. My feeling is positive. I'm a bit surprised that I see `fused` when there is no internet connection and `Device only` is set in Android settings but still, the final result is positive. 
---
Testing environment:
- `high accuracy` set on devices
- a lot of WiFi available in the surrounding
- version with fix installed on Android 9.0 and 7.0
- the released version installed on Android 8.1 and 6.0 to compare
---
Test 1 - **Geopoint with NO appearance** - No sim card - only **WiFi connection** - inside a building - hard to find GPS
* Location is set faster with fused even if satellites are not available.
* It's hard to get location >10m

Test 2 - **Geopoint with NO appearance** - **Sim card** - No WiFi connection - inside a building - hard to find GPS
* Location is set faster with fused even if satellites are not available. 
* It's hard to get location >10m. I waited more than 4 minutes and accuracy was ~15m

Test 3 - **Geopoint with NO appearance** - No sim card - only **WiFi connection** - outside a building
* works smoothly, location with good accuracy is set quite fast
* get a location with accuracy >10m takes about 30 sec but it does depend on the device.

Test 4 - **Geopoint with NO appearance** - **Sim card** - No WiFi connection - outside a building
* works smoothly, location with good accuracy is set quite fast
* get a location with accuracy >10m takes about 30 sec but it does depend on the device

Test 5 - **Geopoint with NO appearance** - No internet connection - outside a building
* Location provider: fused
* works smoothly, location with good accuracy is set quite fast

Test 7 - **Geopoint with placement-map appearance** - Internet connection available: WiFi or mobile data - inside a building - hard to find GPS
- not verified because of crash on Android 6.0 

Test 6 - **Geopoint with placement-map appearance** - No internet connection - outside a building
* Accuracy is visible before provider in a line
- not verified because of crash on Android 6.0 

Test 7 - **Geopoint with placement-map appearance** - Internet connection available: WiFi or mobile data - inside a building - hard to find GPS
- not verified because of crash on Android 6.0 
---

Testing environment:
- `device only` set on devices
---

Test  - **Geopoint with NO appearance** - No internet connection - outside a building
* Location provider: fused
* works smoothly, location with good accuracy is set quite fast

mmarciniak90:

lognaturel:
Thanks, @mmarciniak90.

From what I understand of the stack trace, `java.util.Spliterator` is needed by Mapbox via OkHttp3 and it's failing to resolve at runtime even though it was available at compile time. The only change in this PR that I can imagine this would be related to is the addition of `coreLibraryDesugaringEnabled true`. `Spliterator` was added to Android in API 24 so I assume it must be made available through some magic. 

I can't reproduce on Android 5.1 or on my Android 6 emulator so I'm not really sure where to go from here. I'm more suspicious of @mmarciniak90's tooling (Android Studio version?) than an actual issue with the code but I can't explain why there would be an issue only on one Android version. Some things I can think of to try:
- @mmarciniak90 tries https://github.com/getodk/collect/pull/3886 which also introduces the new desugaring. This would give us insights into whether that's the culprit.
- I produce an APK for @mmarciniak90 to try on Android 6. If that works, we don't worry about it.

lognaturel:
I built an APK to try out: [collect-debug-v1.27.1-160-g9f7dd6439.apk.zip](https://github.com/getodk/collect/files/4766703/collect-debug-v1.27.1-160-g9f7dd6439.apk.zip)

That APK works on my Android 5.1 device. It crashes on my Android 6.0 emulator when I tap "Get Blank Form":
```xml
A/art: art/runtime/thread.cc:1321] No pending exception expected: java.lang.NoClassDefFoundError: $r8$wrapper$java$util$Spliterator$-WRP
    art/runtime/thread.cc:1321]   at java.lang.Class dalvik.system.DexFile.defineClassNative(java.lang.String, java.lang.ClassLoader, java.lang.Object) (DexFile.java:-2)
    art/runtime/thread.cc:1321]   at java.lang.Class dalvik.system.DexFile.defineClass(java.lang.String, java.lang.ClassLoader, java.lang.Object, java.util.List) (DexFile.java:226)
    art/runtime/thread.cc:1321]   at java.lang.Class dalvik.system.DexFile.loadClassBinaryName(java.lang.String, java.lang.ClassLoader, java.util.List) (DexFile.java:219)
    art/runtime/thread.cc:1321]   at java.lang.Class dalvik.system.DexPathList.findClass(java.lang.String, java.util.List) (DexPathList.java:338)
    art/runtime/thread.cc:1321]   at java.lang.Class dalvik.system.BaseDexClassLoader.findClass(java.lang.String) (BaseDexClassLoader.java:54)
    art/runtime/thread.cc:1321]   at java.lang.Class java.lang.ClassLoader.loadClass(java.lang.String, boolean) (ClassLoader.java:511)
    art/runtime/thread.cc:1321]   at java.lang.Class java.lang.ClassLoader.loadClass(java.lang.String) (ClassLoader.java:469)
    art/runtime/thread.cc:1321]   at void okhttp3.internal.Util.<clinit>() (Util.kt:61)
    art/runtime/thread.cc:1321]   at java.util.List okhttp3.internal.Util.immutableListOf(java.lang.Object[]) (Util.kt:-1)
    art/runtime/thread.cc:1321]   at void okhttp3.OkHttpClient.<clinit>() (OkHttpClient.kt:1029)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.openrosa.OpenRosaHttpInterface org.odk.collect.android.injection.config.AppDependencyModule.provideHttpInterface(android.webkit.MimeTypeMap, org.odk.collect.utilities.UserAgentProvider) (AppDependencyModule.java:104)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.openrosa.OpenRosaHttpInterface org.odk.collect.android.injection.config.AppDependencyModule_ProvideHttpInterfaceFactory.provideHttpInterface(org.odk.collect.android.injection.config.AppDependencyModule, android.webkit.MimeTypeMap, org.odk.collect.utilities.UserAgentProvider) (AppDependencyModule_ProvideHttpInterfaceFactory.java:41)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.openrosa.OpenRosaHttpInterface org.odk.collect.android.injection.config.AppDependencyModule_ProvideHttpInterfaceFactory.get() (AppDependencyModule_ProvideHttpInterfaceFactory.java:31)
    art/runtime/thread.cc:1321]   at java.lang.Object org.odk.collect.android.injection.config.AppDependencyModule_ProvideHttpInterfaceFactory.get() (AppDependencyModule_ProvideHttpInterfaceFactory.java:11)
    art/runtime/thread.cc:1321]   at java.lang.Object dagger.internal.DoubleCheck.get() (DoubleCheck.java:47)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.openrosa.OpenRosaAPIClient org.odk.collect.android.injection.config.DaggerAppDependencyComponent.getOpenRosaAPIClient() (DaggerAppDependencyComponent.java:164)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.utilities.FormListDownloader org.odk.collect.android.injection.config.DaggerAppDependencyComponent.getFormListDownloader() (DaggerAppDependencyComponent.java:170)
    art/runtime/thread.cc:1321]   at org.odk.collect.android.activities.FormDownloadListActivity org.odk.collect.android.injection.config.DaggerAppDependencyComponent.injectFormDownloadListActivity(org.odk.collect.android.activities.FormDownloadListActivity) (DaggerAppDependencyComponent.java:451)
    art/runtime/thread.cc:1321]   at void org.odk.collect.android.injection.config.DaggerAppDependencyComponent.inject(org.odk.collect.android.activities.FormDownloadListActivity) (DaggerAppDependencyComponent.java:243)
    art/runtime/thread.cc:1321]   at void org.odk.collect.android.activities.FormDownloadListActivity.onCreate(android.os.Bundle) (FormDownloadListActivity.java:148)
```

I swear I tried that yesterday but now I'm not so confident anymore.

I feel like it has to be some kind of tooling problem but I don't have any ideas beyond that. @seadowg, maybe you have some ideas of what to look at?

lognaturel:
Good news (?), it's an Android bug: https://issuetracker.google.com/issues/157681341. I don't understand why I can't reproduce on my Android 5.1 tablet but I'll let it go. 

[collect-release-v1.27.1-160-g9f7dd6439.apk.zip](https://github.com/getodk/collect/files/4767181/collect-release-v1.27.1-160-g9f7dd6439.apk.zip) is a release build that works on my Android 6.0 emulator. Please give that a try when you can, @mmarciniak90.

mmarciniak90:
@lognaturel I was able to run your apk on ALL devices :tada:
I can continue testing this PR tomorrow.

mmarciniak90:
Attached apk installed on Androids: 6.0, 7.0, 8.1, 9.0
I used `GPS reset` app to reset GPS data between tests

---

**Testing environment:**
* `high accuracy` set on devices
* a lot of WiFi available in the surrounding
* 'Don't keep activities' tuned off
* Internet connection available: **wifi or mobile data**
* outside a building

Verified cases:
* Geopoint with NO appearance
   * Location provider: fused
   * accuracy is visible as a first on the list
   * Fused location is visible as a second on the list
   * It's possible to get accuracy >10m. I feel that it takes longer but I'm in a different surrounding than always
   * MemoryLeaks are NOT noticeable
   * light/dark mode

* Geopoint with placement-map / maps appearance 
   * accuracy is visible as a first on the list
   * Location is visible as a second on the list
   * It's possible to get accuracy >10m
   * Google Maps - Provider location: fused; the map is not loaded when I use attached apk
   * Mapbox  - Provider location: fused
   * OSM - Provider location : **GPS**
   * MemoryLeaks are NOT noticeable

---
**Testing environment:**
* `high accuracy` set on devices
* a lot of WiFi available in the surrounding
* 'Don't keep activities' tuned off
* **internet connection disabled**
* outside a building

Verified cases:
* Geopoint with NO appearance
   * Location provider: fused
   * MemoryLeaks are NOT noticeable
   * same results like above

* Geopoint with placement-map / maps appearance 
   * same results like above
   * MemoryLeaks are NOT noticeable
   * Google Maps - Provider location: fused; the map is not loaded when I use attached apk
   * Mapbox  - Provider location: fused
   * OSM - Provider location : **GPS**
---
**Testing environment:**
* `device only` set on devices
* a lot of WiFi available in the surrounding
* 'Don't keep activities' tuned off
* Internet connection available: `wifi or mobile data`
* outside a building

Verified cases:
* Geopoint with NO appearance
   * Location provider: fused
   * MemoryLeaks are NOT noticeable
   * same results like above

* Geopoint with placement-map / maps appearance 
   * same results like above
   * Google Maps - Provider location: fused; the map is not loaded when I use attached apk
   * Mapbox  - Provider location: fused
   * OSM - Provider location : **GPS**
   * MemoryLeaks are NOT noticeable
---
**Testing environment:**
* `device only` set on devices
* a lot of WiFi available in the surrounding
* 'Don't keep activities' tuned off
* **Internet connection disabled**
* outside a building

Verified cases:
* Geopoint with NO appearance
   * Location provider: fused
   * MemoryLeaks are NOT noticeable
   * same results like above

* Geopoint with placement-map / maps appearance 
   * same results like above
   * Google Maps - Provider location : fused ; map is not loaded when I use attached apk
   * Mapbox  - Provider location : fused
   * OSM - Provider location : **GPS**
   * MemoryLeaks are NOT noticeable

lognaturel:
Thanks, @mmarciniak90. Google maps not loading with the APK is expected because I didn't include a key.

The OSM-based widgets still using the old-style providers was a mistake. It turns out that a `LocationClient` was being built but not properly passed on to the map. I didn't notice this in code and apparently didn't check the behavior either. Good catch and sorry about that. I added a commit that only affects the OSM implementation. I verified that this is shared behavior between point with map, shape and trace so verifying a point on a map with a single device should be sufficient. 

This commit does make a behavior change. Previously, if the GPS provider was disabled in settings, a special dialog would be displayed. I removed that check since we're going to use the fused provider in most cases anyway. If the location client fails to initialize, then that dialog will be displayed (this is unchanged).

mmarciniak90:
@lognaturel take a look at disturbing results like:
   * I didn't observe that accuracy value is being improved when map is opened with OSM
   * OSM map is NOT zoomed to current location automatically
---
APK with newest change installed on Androids: 7.0, 9.0

**General testing environment:**
* a lot of WiFi available in the surrounding
* 'Don't keep activities' tuned off
* outside a building
---
Verified cases when `device only` set on devices and internet connection is **available**
* Geopoint with placement-map / maps appearance
   * OSM - Location provider: fused
   * Accuracy is visible as a first on the list
   * Location is visible as a second on the list
   * It's possible to get accuracy >10m. 
   * **I didn't observe that accuracy value is being improved when map is opened with OSM** 
   * OSM map is zoomed to current location automatically
* Geotrace
   * only the accuracy is visible
   * It's possible to get accuracy >10m
   * Google Sheet, Mapbox, OSM
   * **I didn't observe that accuracy value is being improved when map is opened with OSM**
   * **OSM map is NOT zoomed to current location automatically**
* Geoshape
   * only the accuracy is visible
   * It's possible to get accuracy >10m
   * Google Sheet, Mapbox, OSM
   * **I didn't observe that accuracy value is being improved when map is opened with OSM**
   * **OSM map is NOT zoomed to current location automatically**
---
Verified cases when `device only` set on devices and internet connection is **disabled**
* Geopoint with placement-map / maps appearance
   * same results as above
* Geotrace
   * same results as above
* Geoshape
   * same results as above
---
Verified cases when `high accuracy` set on devices and internet connection is **available**
* Geopoint with placement-map / maps appearance
   * same results as above
* Geotrace
  * same results as above
* Geoshape
   * same results as above
---
Verified cases when `high accuracy` set on devices and internet connection is **disabled**
* Geopoint with placement-map / maps appearance
   *  same results as above
* Geotrace
   * same results as above
* Geoshape
   * same results as above

lognaturel:
As long as there’s no obvious regression in accuracy, I’m not too concerned. It’s very highly situational and the most important thing is to get it out to users to try out in their environments. 

I think we should merge this as-is and @mmarciniak90 can file an issue for the zoom behavior. It needs to be fixed for the release but it’s not something we need to get user feedback on. Great if someone reports it because then we know they’re paying attention. This lets us focus on the JR issue and a point release.

seadowg:
@lognaturel agreed. Good to get this out in a beta to let people play with it. If it doesn't work it's pretty easy to revert given we fall back to the good ol' `LocationManager` anyway.

mmarciniak90:
I will create separate issues for the mentioned behaviors tomorrow.

lognaturel:
I thought you meant that the accuracy didn't seem better than with the raw providers. I didn't realize that what you meant was that **location** wasn't updated but the issues helped clarify that. I have fixed it in the latest commit and ended up not merging since it was a severe regression and would not have let us get the beta feedback we need. In verifying this more thoroughly, I saw an accuracy of 2.79 so < 3m accuracy appears possible.

https://github.com/getodk/collect/pull/3927 is the priority and then it would be good to get a quick check here to see if anything else crazy is going on. Again, the change is shared between geopoint with map, geotrace, geoshape.

mmarciniak90:
**Tested with success**

Tested cases described in #3930 and #3929 on Androids: 7.0, 8.1, 9.0

Verified on `high accuracy` and `device only` settings, with and withouth internet connection.

Verified widgets: Geopoint with placement-map appearance, Geopoint with maps appearance, Geotrace, Geoshape

PR #3776, opened 2020-04-10
Fixed changing languages

### LINKED ISSUE #3649: Update `appcompat` dependency
Currently we are stuck on `appcompat:1.1.0-rc01` due to an issue introduce in `appcompat:1.1.0`: https://issuetracker.google.com/issues/141869006.

To reproduce the issue:
1. Change the app language to something other than your system default
1. Change `appcompat` version to 1.1.0 in `collect_app/build.gradle`
1. Redeploy the app

The language will change back to the system default.

It looks like the developers believe the issue is fixed in `1.2.0-alpha02` but trying that with the above steps (instead of `1.1.0`) shows the same problem. It could be that our method of changing the language needs to be updated. The current method (implemented in `CollectAbstractActivity#attachBaseContext`) has never been officially supported but the devs have posted links to examples in that issue that they are using for Night Mode:

> That's not going to work with AppCompat anymore. You can take either of the following approaches at the Activity level:
> 1. Create a configuration context in Activity.attachBaseContext() ([example](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:appcompat/appcompat/src/androidTest/java/androidx/appcompat/app/NightModeCustomAttachBaseContextActivity.java))
> 2. Call applyOverrideConfiguration() from Activity.attachBaseContext() ([example](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:appcompat/appcompat/src/androidTest/java/androidx/appcompat/app/NightModeCustomApplyOverrideConfigurationActivity.java))

It would be good to experiment with these examples to try and see if we can update to the latest `appcompat`. Otherwise we need to make sure the `appcompat` team has an issue to work on.

### LINKED ISSUE #3774: User is not able to change Collect's language on Android 7.x
Collect current master
Android 7.0

#### Problem description
User is not able to change Collect's language

1. User goes to `General Settings` -> `User Interface`
2. User changes language
3. User is moved to the main view but language stays unchanged
4. The selected language is visible in settings

#### Expected behavior
User is able to change Collect's language to different than device language

#### Other information 
Issue is NOT visible on released v1.26.1

grzesiek2010:
Closes #3774
Closes #3649

I tested the fix on many different android versions (mostly emulators).

As I said the problem appeared after updating dagger version:
https://github.com/opendatakit/collect/issues/3774#issuecomment-611993334

I was googling and noticed that other developers had similar problem after migrating to androidX and just used the same fix which turned out to work. Don't ask me for more details I don't know how dagger caused it :smile: 

We need to test changing languages on all android versions we support to be sure everything is fine. Changing the theme (light/dark) is a related functionality so I would also check it to rule out regression.

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -198,14 +198,12 @@ configurations.all {
 dependencies {
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.browser:browser:1.2.0'
-    implementation 'androidx.appcompat:appcompat:1.1.0-rc01' // https://github.com/opendatakit/collect/issues/3649
+    implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'com.google.android.material:material:1.1.0-alpha10'
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.exifinterface:exifinterface:1.1.0'
     implementation 'androidx.multidex:multidex:2.0.1'
-    implementation('androidx.preference:preference:1.1.0') {
-        exclude group: 'androidx.appcompat', module: 'appcompat' // Again because of https://github.com/opendatakit/collect/issues/3649
-    }
+    implementation 'androidx.preference:preference:1.1.0'
 
     implementation "com.google.android.gms:play-services-auth:17.0.0"
     implementation "com.google.android.gms:play-services-maps:17.0.0"
--- a/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
@@ -17,6 +17,7 @@
 package org.odk.collect.android.activities;
 
 import android.content.Context;
+import android.content.res.Configuration;
 import android.os.Bundle;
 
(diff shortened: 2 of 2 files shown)


seadowg (approved):
Absolutely do not understand how Dagger caused this either! It should be unrelated to Android... 🤯

Are you able to see if this change also lets us fix https://github.com/opendatakit/collect/issues/3649? It looks pretty similar to the workarounds posted.

grzesiek2010:
> Are you able to see if this change also lets us fix #3649? It looks pretty similar to the workarounds posted.

Yeah I tested it and seems like the fix works for both issues so I updated `appcompat` in a separate commit.

seadowg:
@grzesiek2010 such good news! That bug was really annoying me.

mmarciniak90:
**Tested with success**

Verified cases:
* change language (especially on 7.0)
* closing app after changing language (especially on 6.0)
* change theme
* scanning QA code with a different language and dark mode

@kkrawczyk123 please confirm on Android 5.1, 7.1, 10.0

kkrawczyk123:

seadowg:
Can we hold off on merging until we have #3656 in?

PR #3777, opened 2020-04-15
Add experimental Material Theme

### LINKED ISSUE #3693: Simplify version parsing code and ensure that different tag structures won't cause a crash
See https://github.com/opendatakit/collect/pull/3685#issuecomment-595494561 for suggested approach. I think that by first handling the beta version suffix and then splitting on the first `-`, there will no longer be a crash risk but this should be added to the tests (e.g. `my-cool-version` or `777` as versions).

seadowg:
Closes #3693 

We've been talking for a while about switching to the newer `MaterialComponents` which has better support for the Material Components library that we are using. There are two core complications to doing this:

1. The `MaterialComponents` theme tries to provide some magic that actually makes our lives harder - an example would be that it automatically converts all the `Button` instances in XML to `MaterialButton`
2. Android moved a lot of it's non material specific UI framework (dialogs, preferences etc) to a new `androidx` package and deprecated the old ones. The `MaterialComponents` theme doesn't behave that well with the older components which we still have a lot of.

To remove problem 1) this PR changes all the `Button` instances in the app to `MaterialButton` and styles them to look (as much as possible) like the buttons did before. 

Solving 2) will take a little longer but so we're not spending the rest of the year changing frameworks without much to say for it so I'm proposing we add a third "experimental" theme to the app (that extends `MaterialComponents`). This will let us offer it to users (maybe just in beta) while we track down and fix problems. It also lets us start a new theme (virtually) from scratch which will be nice for helping to eliminate styles and attributes we don't really need. An extra more fun advantage is that it will let us play with different colors for the app!

Here's a screenshot of the new theme in all it's glory:

Screenshot_1586946446

Yes that's the Central color scheme.

The things we still definitely need to fix to get to a point where this new theme could be our default are. We could either create a few issues or one "super" issues for these:

* `ProgressDialog` buttons are showing as contained buttons. `ProgressDialog` has actually been deprecated and so it used the old `android.app.AlertDialog` which creates styling problems. I've made our `ProgressDialogFragment` use the new `AlertDialog` as part of this PR so we can replace the old ones (or change the design of loading) over time.
* The date picker in `DateWidget` also uses contained buttons. I'm not very familiar with this part of the code but my thinking is we might just need to switch to using the new `MaterialDatePicker` here.
* We use the old `preferences` framework in a bunch of places and this means any dialogs used there will probably not look right.
* Remove `iconColor` theme attribute by using more appropriate colors/multiple icons.

Ran all tests locally and on test lab. Played around with a few different forms.

Shouldn't be any real changes other than the new theme and that across all themes the buttons in widgets are very slightly bigger (as they use one of our standard margins for padding rather than a pixel value in code). There has been some changes to progress dialogs so testing them would be good.


lognaturel:
I like the overall approach. It's nice to see what a material components theme looks like and nice to do low-hanging-fruit changes like making sure resources are used where possible.

Still working through a more detailed code review but in the mean time, I've immediately run into a couple of issues:
- Crash when I try to go to user interface preferences:
```    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.odk.collect.android.version.VersionInformation.isRelease()' on a null object reference
        at org.odk.collect.android.preferences.UserInterfacePreferencesFragment.initThemePrefs(UserInterfacePreferencesFragment.java:90)
        at org.odk.collect.android.preferences.UserInterfacePreferencesFragment.onCreate(UserInterfacePreferencesFragment.java:79)
        at android.app.Fragment.performCreate(Fragment.java:2075)
```
- Blank form and unfinalized filled form icons have black insides instead of white in light theme.

I think it might also be helpful to move to the new test package structure and fix conflicts so that can be reviewed.

seadowg:
@lognaturel I can't reproduce the crash just by going to the User Interface preferences? Are there any other steps you took? If not could you debug and see what's happening? It looks like there is a crash if you move from a beta to a release version with Magenta theme on then try and change themes but want to make sure you haven't found something else!

seadowg:
@lognaturel color problem is that the `form_state_x` icons happen to use `colorSecondary` but they actually have their own color scheme. I've made them use hard coded colors so that they are consistent between themes.

lognaturel:
No other step. Stack trace:
```java
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.odk.collect.android.version.VersionInformation.isRelease()' on a null object reference
        at org.odk.collect.android.preferences.UserInterfacePreferencesFragment.initThemePrefs(UserInterfacePreferencesFragment.java:90)
        at org.odk.collect.android.preferences.UserInterfacePreferencesFragment.onCreate(UserInterfacePreferencesFragment.java:79)
        at android.app.Fragment.performCreate(Fragment.java:2075)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:869)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1083)
        at android.app.BackStackRecord.run(BackStackRecord.java:834)
        at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1470)
        at android.app.FragmentManagerImpl$1.run(FragmentManager.java:452)
        at android.os.Looper.loop(Looper.java:145)
        at android.app.ActivityThread.main(ActivityThread.java:6939)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
```

seadowg:
@lognaturel that would suggest Dagger isn't injecting the object. Are you definitely on the latest commit? Would you be able to debug and investigate? I'm not able to reproduce at all.

Also what device/API version are you using?

collect_app/src/main/java/org/odk/collect/android/preferences/UserInterfacePreferencesFragment.java
```diff
@@ -72,11 +87,19 @@ private void initThemePrefs() {
         final ListPreference pref = (ListPreference) findPreference(KEY_APP_THEME);
 
         if (pref != null) {
+            if (versionInformation.isRelease()) {
```
lognaturel, line 98: How about a comment here or putting it in a method called `hideMaterialThemeForRelease` or something?

lognaturel:
I'm on Android 5.1, Galaxy Tab A. I can't get this to work as written. If I move the preference initialization to `onAttach`, it works. If I move the injection to `onCreate` (not recommended), it works. When I put breakpoints in `onAttach` or methods called from `onAttach`, I never seem to reach them even though I do break in `onCreate`. I can't explain that.

This comes with a pretty big change in behavior -- activities only get the new theme when they are recreated as part of user navigation. I think we should maintain the prior behavior where changing the theme means being sent back to the (recreated and rethemed) main menu.

collect_app/src/main/java/org/odk/collect/android/version/VersionInformation.java
```diff
@@ -0,0 +1,75 @@
+package org.odk.collect.android.version;
+
+import androidx.annotation.Nullable;
+
+public class VersionInformation {
```
lognaturel, line 5: This is much better. It very nearly could close https://github.com/getodk/collect/issues/3693 if you could add the couple of test cases suggested. I still think the regex approach could cut some of this down but it's now much easier to reason about so I'm fine with it as-is.

lognaturel (changes requested):

seadowg:
@lognaturel ah yeah I think this might be a problem in Android where they changed the ordering of some lifecycle. I'll test this against 5.1 on the emulator and see if I get the same thing. I think this is probably only a problem for `PreferenceFragment` objects as it's probably to do with the preference lifecycle methods being called before methods like `onCreateView` would be.

seadowg:
> This comes with a pretty big change in behavior -- activities only get the new theme when they are recreated as part of user navigation. I think we should maintain the prior behavior where changing the theme means being sent back to the (recreated and rethemed) main menu.

This wasn't an intentional change and I also can't reproduce what you're seeing. If I change themes I'm still returned to the main menu which is presented in the theme I've just selected. More details on the device/API level please?

collect_app/src/main/java/org/odk/collect/android/preferences/UserInterfacePreferencesFragment.java
seadowg: Yeah good call. I'll put in a method.

collect_app/src/main/java/org/odk/collect/android/version/VersionInformation.java
seadowg: Ah didn't see that issue. Definitely the right time. Will get to that once other issues are fixed!

lognaturel:
> I also can't reproduce what you're seeing

That's strange. Same Android 5.1 Galaxy Tab A. If there were lifecycle changes (yikes), that could also be related.

seadowg:
@lognaturel wait how did you see that behaviour if you were getting crashes just accessing the user interface screen? Surely the app was crashing before you had the chance to change themes.

lognaturel:
> If I move the preference initialization to onAttach, it works. If I move the injection to onCreate (not recommended), it works.

With either of those changes I can change the theme but then I get the behavior I described -- I am still on the UI preference screen with the old theme, I go back to the main screen with the old theme and then if I access another activity it is themed. But I just realize that maybe `onAttach` is never called and so the change listeners don't exist? It probably makes sense to get to reasonable fix for the crash first and then we can see what's going on with this.

lognaturel:
Oh, right. [`onAttach(context)` was added in API 23](https://developer.android.com/reference/android/app/Fragment#onAttach(android.content.Context)) so we can't use that. I'm surprised there was no warning! So that explains all of the issues I'm getting, I think.

seadowg:
@lognaturel ah right I misunderstood that you were referring to the changes you made causing that. I think the solve is moving the fragment to the new `PreferenceFragmentCompat` which gives us both the `onAttach` we want and the new `onCreatePreferences` that is where preferences should be initialized. It'll just be a similar transition to the one we had to do when adding Install ID.

collect_app/src/main/java/org/odk/collect/android/version/VersionInformation.java
seadowg: Looks like the implementation was there for this but wanted to move the coverage from the ViewModel to the `VersionInformation` so did that.

lognaturel (approved):
I assume that `AndroidXPreferencesActivity` is a temporary solution.

seadowg:
@lognaturel yeah it is and it totally needs a comment! I'll add that quickly.

mmarciniak90:
WOW! What an energizing color :smile: 

1. Sometimes `Cancel` is blue and sometimes is a button in magenta color, eg. when loading forms to download. In the previous version, it was just text, not a button. Additionally, on Android 6.0, there is a button with blue text


2. Text "Get Selected"  on Android 7.0 is in two lines which causes that buttons are not in one line

3. Offline layout dialog has doubled the `None` option. Opening the same dialog from `General Settings` -> `Maps` causes crash

4. Search bar has extra :mag: icon

seadowg:
@mmarciniak90 haha yeah it certainly wakes you up right? These are good finds. As this is only going to be in dev and beta I don't think any of these are show stopping but we'll create issues after you're finished testing. I think short of behaviour problems/crashes we should really only block on problems found with dark/light theme on.

mmarciniak90:
Ok, so I specified which cases are visible on light and dark modes as well

1. Text `Get Selected` on Android 7.0 is in two lines which causes that buttons are not in one line

| master | PR /light mode | PR / dark mode |
|---|---|---|
|Screenshot_20200506-102319|Screenshot_20200506-102044|Screenshot_20200506-102501|

2. Offline layout dialog has doubled the `None` option. 
Opening the same dialog from `General Settings` -> `Maps` causes crash 

|Screenshot_20200506-110125|Screenshot_20200506-110145|
|---|---|

seadowg:
@mmarciniak90 amazing thanks. I'll get to work on those.

mmarciniak90:
@seadowg I executed a sneaky case :sunglasses: Let me know that we accept its result.

1. I added a tag and built apk
2. I didn't see `Magenta theme` on this version
3. I scanned QR code where Magenta theme was set
4. The magenta theme was applied on app with tag

kkrawczyk123:
@seadowg I think there is one more thing that it's worth mentioning about magenta look only except ui issues - I have noticed crashes on all date widgets from Ethiopian to Persian date widgets and when open rank widget.

seadowg:
> I have noticed crashes on all date widgets from Ethiopian to Persian date widgets and when open rank widget.

@kkrawczyk123 hmmm. I think we probably need to fix that before introducing Magenta in beta releases. Could you post full steps for these problems? It might be that for this PR we just have Magenta in dev and then have the crashes as an issue we need to fix.

> I executed a sneaky case 😎 Let me know that we accept its result.

haha nice. Personally I think that's fine. @lognaturel do you have a second opinion on this?

kkrawczyk123:
@seadowg sure. I have used all widgets form, reproduction steps:
1. open Ethiopian/Coptic/Islamic/ Bikram Sambat/Myanmar/Persian date widget
2. click on Select date button
the steps for Rank widgets are actually the same, click on Rank items button in the 2nd step. 

I was able to reproduce it on Androids 5.1, 7.1 and 10. (i don't have any other)

the stacktrace for crash:
`2020-05-06 14:26:44.898 9760-9760/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 9760
    android.view.InflateException: Binary XML file line #54: Binary XML file line #54: Error inflating class TextView
    Caused by: android.view.InflateException: Binary XML file line #54: Error inflating class TextView
    Caused by: java.lang.UnsupportedOperationException: Failed to resolve attribute at index 5: TypedValue{t=0x2/d=0x7f040252 a=-1}
        at android.content.res.TypedArray.getColorStateList(TypedArray.java:528)
        at android.widget.TextView.<init>(TextView.java:1076)
        at android.widget.TextView.<init>(TextView.java:704)
        at androidx.appcompat.widget.AppCompatTextView.<init>(AppCompatTextView.java:99)
        at com.google.android.material.textview.MaterialTextView.<init>(MaterialTextView.java:91)
        at com.google.android.material.textview.MaterialTextView.<init>(MaterialTextView.java:86)
        at com.google.android.material.textview.MaterialTextView.<init>(MaterialTextView.java:81)
        at com.google.android.material.theme.MaterialComponentsViewInflater.createTextView(MaterialComponentsViewInflater.java:109)
        at androidx.appcompat.app.AppCompatViewInflater.createView(AppCompatViewInflater.java:103)
        at androidx.appcompat.app.AppCompatDelegateImpl.createView(AppCompatDelegateImpl.java:1407)
        at androidx.appcompat.app.AppCompatDelegateImpl.onCreateView(AppCompatDelegateImpl.java:1457)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:769)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:727)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:858)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:861)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:518)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:426)
        at androidx.appcompat.app.AlertController.setupCustomContent(AlertController.java:644)
        at androidx.appcompat.app.AlertController.setupView(AlertController.java:475)
        at androidx.appcompat.app.AlertController.installContent(AlertController.java:233)
        at androidx.appcompat.app.AlertDialog.onCreate(AlertDialog.java:279)
        at android.app.Dialog.dispatchOnCreate(Dialog.java:395)
        at android.app.Dialog.show(Dialog.java:294)
        at androidx.fragment.app.DialogFragment.onStart(DialogFragment.java:529)
        at androidx.fragment.app.Fragment.performStart(Fragment.java:2730)
        at androidx.fragment.app.FragmentStateManager.start(FragmentStateManager.java:365)
        at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1194)
        at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1356)
        at androidx.fragment.app.FragmentManager.moveFragmentToExpectedState(FragmentManager.java:1434)
        at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1497)
        at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:447)
        at androidx.fragment.app.FragmentManager.executeOps(FragmentManager.java:2169)
        at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1992)
        at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1947)
        at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1849)
        at androidx.fragment.app.FragmentManager$4.run(FragmentManager.java:413)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6165)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:778)`

mmarciniak90:

seadowg:
@mmarciniak90 the two issues you found are hopefully fixed. I wasn't actually sure how to reproduce the screenshot you posted for the second one (if you could post instructions for me on that that'd be awesome) but I've fixed the crash which I think had the same cause.

@kkrawczyk123 I'll have a go at fixing that but if it seems big I'll probably remove Magenta from beta so we can merge this and fix that after.

seadowg:
@mmarciniak90 oh and I should have added: for the "Get Selected" button I've made it work as much as possible but on a very thin screen it would have to ellipsize the text in the button to avoid creating two lines.

seadowg:
@kkrawczyk123 that should be the crashes fixed!

lognaturel:
Great sneaky case! There is lots of weirdness around scanning settings in so really good that you're thinking about it, @mmarciniak90. I agree that it's ok to leave in.

It sounds like the crashes y'all have found on the magenta theme have been fixed which is fantastic. I think it would be ok to include in beta even with some specific cases/situations/widgets crashing.

kkrawczyk123:
@seadowg I have noticed that Download Selected button in Get Blank From when Google Drive is selected as type of server has changed a little: 
| Android 10 | Android 5.1 | master |
| --------- | --------- | --------- |
| Screenshot from 2020-05-08 10-21-31 | Screenshot from 2020-05-08 10-22-19 | Screenshot from 2020-05-08 10-21-59 |


Is there any way to avoid that text wrap?

seadowg:
@kkrawczyk123 ah. This is related to the first problem @mmarciniak90 found - I've made it so those buttons only take up one line. We could go with making the buttons all two lines if one of them pushes into that instead though which seems better to me. I'll change it!

mmarciniak90:
@seadowg maybe we should think about standardizing this text for Aggregate and Google Drive. We can set it as `Get Selected` for both servers

seadowg:
@mmarciniak90 I like that idea but "Shared with Me" is going to be a problem on screens that size as well. Those buttons probably need to change anyway as they're fairly non standard so I'd rather rethink text then.

mmarciniak90:
@seadowg agree! Also manipulating text size in Android settings will cause similar issues.

seadowg:
@kkrawczyk123 the buttons should be fixed now!

mmarciniak90:

Verified cases on light, dark and magenta themes

Re-tested cases:
* Offline layout dialog from settings, Geo widget and submissions view on map
* Buttons on the `Get Blank Form` view for Aggregate and Google Drive
   * The Spanish version has even 3 lines and buttons stay on the same level
* Date widgets from Ethiopian to Persian date widgets and  rank widget
* English, Spanish language

mmarciniak90:

PR #3778, opened 2020-04-16
Provide submission summary when tapping a submission on the form map

### LINKED ISSUE #3737: Provide submission summary when tapping a submission on the form map
Currently, tapping on a submission from the form map opens that submission for viewing or editing. Instead, it should be possible to see:
- the instance name
- the current instance status
- the last status change time
- the tappable action that can be taken on the form (view, edit, etc)

Future extensions may include:
- displaying the first image from the submission
- displaying a user-specified image from the submission
- displaying user-specified fields from the submission

Todo:
- [x] design basic user interface and share on forum

grzesiek2010:
Closes #3737

This is what we discussed on the forum https://forum.getodk.org/t/provide-submission-summary-when-tapping-a-submission-on-the-form-map/25957

It just changes the way we deal with submissions visible on a map. Only this functionality might be affected. Now we should see a submission every time we click on a marked instead of opening a form immediately.

Any form with geopoints.

No.


grzesiek2010:
@lognaturel this is the pr you wanted for review so here you go.

lognaturel:
I think that this can move forward with a more final implementation and feature tests. I responded to the forum thread about the action button. I’d rather use another visual pattern for that. That said, I’m ok with you and @seadowg discussing and deciding which way to go.

grzesiek2010:
@lognaturel I pushed the latest changes in case you want to play with it on your device. I saw :heart: on the forum so I guess you like it more than the version with FAB but I'll wait for your feedback.

lognaturel:
This looks fantastic to me. The action is explicit and it's where the eye is already to get summary info.

I had completely missed this tag -- as we're getting close to code freeze or things get time-sensitive in some other way, please consider also asking for review in #collect-code so there's less risk of an accidental miss (sorry).

lognaturel:
I'm happy with the functionality and user experience. I expected the bottom sheet to be gone when I came back to the map after viewing/editing a submission. I can see the logic either way and it's a non-disruptive thing to change so I think we can leave it as-is for now.

I'd like to have seen the logic in the ViewModel. Moving forward, we should probably favor the [Saved State ViewModel module](https://developer.android.com/topic/libraries/architecture/viewmodel-savedstate) when UI state needs to be maintained when the activity is stopped. We hadn't discussed this before so maybe it's best to leave this as-is and make the switch moving forward.

Test-wise, I think it would be nice to have a test specifically focused on the appearance of the bottom sheet when a submission is tapped.

@seadowg I didn't have time for a full review today and I'm not very confident in my two comments above. Could you please have a look?

grzesiek2010:
I noticed that if a submission is deleted or encrypted we clear coordinates:
https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java#L312
https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java#L462

but we still have code for displaying info about them in FormMapActivity and test that are failing in this pr.
So I wonder whether it make sense to fix them maybe I can just remove the code and tests if it's not possible to display such instances on a map at all.
@lognaturel

collect_app/src/main/res/drawable/form_state_finalized.xml
```diff
@@ -1,20 +1,17 @@
...
+    android:width="40dp"
+    android:height="40dp"
+    android:viewportWidth="120"
+    android:viewportHeight="120">
 
     <path
-            android:pathData="M60,60m-60,0a60,60 0,1 1,120 0a60,60 0,1 1,-120 0"
-            android:fillColor="@color/deep_purple_500"/>
+        android:pathData="M67.195,26.83h-26.83c-3.688,0 -6.673,2.985 -6.673,6.633L33.66,86.537c0,3.648 2.984,6.634 6.673,6.634L80.61,93.171c3.689,0 6.707,-2.986 6.707,-6.634L87.317,46.732L67.195,26.829zM40.365,86.537L40.365,33.463h23.476L63.841,50.05L80.61,50.05v36.488L40.366,86.538z"
+        android:fillColor="?iconColor"/>
```
seadowg, line 9: I think it's best avoid `iconColor` and go for Material Theme attributes like `colorOnSurface` or `colorOnPrimary`. These let us account for the fact that in some color schemes the icon color would differ depending on where it was rendered.

collect_app/src/main/res/layout/submission_summary_layout.xml
```diff
@@ -0,0 +1,70 @@
...
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical">
+
+        <TextView
+            android:id="@+id/submission_name"
+            style="@style/TextAppearance.AppCompat.Headline"
```
seadowg: This should use of our Material Theme text appearances (`TextAppearance.Collect.X`). I'm guessing `Headline6` is too big but have a play with it!

collect_app/src/main/res/layout/submission_summary_layout.xml
```diff
@@ -0,0 +1,70 @@
...
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textColor="?primaryTextColor" />
```
seadowg: This should be `colorOnSurface`. `primaryTextColor` is actually gone on master.

collect_app/src/main/res/layout/submission_summary_layout.xml
```diff
@@ -0,0 +1,70 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="horizontal">
+
+        <ImageButton
+            android:id="@+id/status_icon"
+            android:layout_width="30dp"
+            android:layout_height="30dp" />
+
+        <TextView
+            android:id="@+id/status_text"
```
seadowg, line 43: Could we use a text appearance for this as well?

collect_app/src/main/res/layout/submission_summary_layout.xml
```diff
@@ -0,0 +1,70 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/margin_small"
+        android:textColor="@android:color/white"
+        app:chipBackgroundColor="?colorSecondary"
+        app:iconStartPadding="@dimen/margin_small"
+        app:chipEndPadding="@dimen/margin_small"/>
+
+    <TextView
+        android:id="@+id/info"
```
seadowg, line 64: Again using a standard text appearance would be nice!

seadowg (changes requested):
This is a cool looking feature!

> Test-wise, I think it would be nice to have a test specifically focused on the appearance of the bottom sheet when a submission is tapped.

Yup I agree with that. Also left some comment inline mostly around theme attributes. Also had some questions around some of the architecture but I've reached out on Slack so we can chat.

collect_app/src/main/res/layout/submission_summary_layout.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/submission_summary_layout.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/submission_summary_layout.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/submission_summary_layout.xml
grzesiek2010: Headline6 looks pretty good. Done.

collect_app/src/main/res/drawable/form_state_finalized.xml
grzesiek2010: That's right but `?iconColor` is used in many places like that so would be better to clear that in a separate pr.

seadowg: Absolutely. It's one of the follow ups to #3777 I'll schedule. It would be nice to not have another one to change in that though!

lognaturel:
> maybe I can just remove the code and tests if it's not possible to display such instances on a map at all

I think that’s ok. They’re certainly not expected or desired states.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -223,47 +216,10 @@ public void onLocationChanged(MapPoint point) {
...
-    private Intent getEditFormInstanceIntentFor(int featureId) {
-        Uri uri = ContentUris.withAppendedId(InstanceColumns.CONTENT_URI, instancesByFeatureId.get(featureId).getDatabaseId());
-        return new Intent(Intent.ACTION_EDIT, uri);
-    }
-
-    private Intent getViewOnlyFormInstanceIntentFor(int featureId) {
-        Intent intent = getEditFormInstanceIntentFor(featureId);
-        intent.putExtra(ApplicationConstants.BundleKeys.FORM_MODE, ApplicationConstants.FormModes.VIEW_SENT);
-        return intent;
+        showSummary(featureId);
```
lognaturel: It doesn't make sense to have both `onFeatureClicked` and `showSummary` since they have the same parameters and `showSummary` is only called from here. I'd keep `onFeatureClicked` and move the code here but you could also rename it to `showSummary` if you think that's clearer.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/InstanceSummaryDialogFragment.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        View view = inflater.inflate(R.layout.submission_summary_layout, container, false);
+        ButterKnife.bind(this, view);
+        return view;
+    }
+
+    @Override
+    public void onSaveInstanceState(Bundle outState) {
+        outState.putLong(INSTANCE_ID, instanceId);
```
lognaturel: This is much better and I guess that if truly only necessary data is being saved continuing to use `onSaveInstanceState` is not so bad.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
```diff
@@ -99,6 +102,15 @@ private void initializeFormInstances() {
         return mappableFormInstances;
     }
 
+    public MappableFormInstance getInstanceById(long instanceId) {
+        for (MappableFormInstance mappableFormInstance : mappableFormInstances) {
```
lognaturel: Doing a sequential search here could be slow if there are lots of instances. I might have considered saving the database id to make this lookup fast. It's probably not that important to do now but wanted to flag it for future consideration.

lognaturel (commented):
Looks good! I don't see anything critical to change in this initial release I've made a couple of comments inline and have a couple more considerations below.

Is the graying out of the map new behavior? I think the position is relevant when the bottom sheet is shown so like in Google Maps, I would expect the map to remain active. 

Relatedly, I would expect that when an instance has been tapped and the bottom sheet is showing, tapping on another instance would immediately show me that second instance's information. One use case for this feature is trying to find a record when you know the general area it was captured in. For example, let's say you visited several households and you want to find Bob's household. You might tap a few different instances in the general neighborhood before finding the right one. Having to tap to close the bottom sheet and then again to select another instance adds a bit of friction. Is there an advantage to doing it as it is now?

grzesiek2010:
> Is the graying out of the map is new behavior?

> Relatedly, I would expect that when an instance has been tapped and the bottom sheet is showing, tapping on another instance would immediately show me that second instance's information.

I used `BottomSheetDialogFragment` which is modal and works like other DialogFragments that's why...

there are some tricks and workarounds for example to get rid of that dimming and maybe even to make the rest of the screen still enabled (touchable) but nothing seems reliable.
I agree with you that it would be better to make the sheet work in the way you described so that i converted `BottomSheetDialogFragment` to `BottomSheetBehavior` https://github.com/getodk/collect/pull/3778/commits/471a4d8351aba5eb99c3d860ce0efb55b12f65f6

**pros:**
-the map is not dimmed and you can scroll it
-it allows avoiding some strange problems like: [Android BottomSheetDialogFragment does not expand completely](https://stackoverflow.com/questions/37555403/android-bottomsheetdialogfragment-does-not-expand-completely)

**cons:**
-I had to remove the class that extended `BottomSheetDialogFragment` and moved the code to the activity
-the sheet disappears on screen rotation by default (but we can handle it)

so to sum up it works better with `BottomSheetBehavior`  but the code might look worse because I had to remove the separate file that extended `BottomSheetDialogFragment`.

I think we should go with `BottomSheetBehavior` do you agree @lognaturel?

grzesiek2010:
There is one problem with `BottomSheetBehavior` `FormMapActivityTest` which is Robolectric is getting stuck on [summarySheet.setState(BottomSheetBehavior.STATE_EXPANDED);](https://github.com/getodk/collect/pull/3778/commits/471a4d8351aba5eb99c3d860ce0efb55b12f65f6#diff-3152f196669edf61566132c2a6d05d1eR242)

@seadowg any idea?

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
grzesiek2010: It is removed after removing BottomSheetDialogFragment

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/InstanceSummaryDialogFragment.java
grzesiek2010: This class no longer exists.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
grzesiek2010: Fixed.

grzesiek2010:
I can think of a few possible improvements but these might be implemented is separate prs:
1. The marker we clicked on is centered but if there are lots of markers close that one it might be unclear which one we selected so we should highlight it somehow - maybe just bigger marker icon?
2. Clicking the back button closes the FormMapActivity even if the bottom sheet is visible I guess it might be a bit confusing and users might expect that dialog to be closed after clicking the back button.
3. The area below map buttons is transparent and we can see the map but it's not touchable and we can't for example click on a marker that is there.

lognaturel (approved):
Thanks for making the change to `BottomSheetBehavior`. 👍 I agree that this is great for a first release and that we can follow up shortly with the other enhancements you list.

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- light mode
- dark mode
- screen rotation
- minimize app
- extinguish screen
- Google maps
- Mapbox
- OSM
- offline layer 
- all statuses: saved, finalized, sent and failed
- delete answer
- edit answer
- change time and status after edition
- zoom buttons
- add submission through displayed map
- changed language 
- encrypted form

mmarciniak90:
Tested with success on Android: 6.0, 7.0, 8.1, 9.0

seadowg (approved):

PR #3779, opened 2020-04-16
Add Material Banner implementation

seadowg:
This adds an implementation of [Material Design's "Banner"](https://material.io/components/banners). @grzesiek2010 had already done most of the work to put this together as part of the Scoped Storage Migration effort and this just makes the banner implementation reusable across the app. 

I've also created a new Gradle submodule for our Material Component implementations. This ensure they stay clean of any Collect specific code and also means the build time while working on them is super fast (as you only have to build the module).

Ran all tests and checked that there weren't any styling regressions.

This code is all pretty well tested so I would usually suggest just going for a merge on this one. However, the storage migration flow isn't something I imagine any of us touch on a day to day basis so it'd be nice to run through that.


material/src/main/java/org/opendatakit/collect/material/MaterialFullScreenDialogFragment.java
```diff
@@ -1,4 +1,4 @@
-package org.odk.collect.android.material;
+package org.opendatakit.collect.material;
```
lognaturel: `org.odk`

material/src/test/java/org/opendatakit/collect/material/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
+package org.opendatakit.collect.material;
```
lognaturel: This example can be removed

material/build.gradle
```diff
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 28
```
lognaturel, line 4: I think this is probably safe to update across the project? Not very important, though.

lognaturel (changes requested):
Package name should match the main project. Other than that, looks like a good example of introducing a submodule.

material/src/main/java/org/opendatakit/collect/material/MaterialFullScreenDialogFragment.java
seadowg: Whoops! Will fix. Interesting that we're using a package name that we don't actually own though (looks like frat beat us to it).

material/build.gradle
seadowg: Yeah thought we migth as well keep it the same right now to avoid any potential problems.

material/src/main/java/org/opendatakit/collect/material/MaterialFullScreenDialogFragment.java
lognaturel: Yes, it's definitely not my favorite thing. The frat has been registered since 1998 so I really don't know why it was done this way.

lognaturel (approved):

lognaturel:
It's too bad that each project needs to have its own build.gradle with redundant directives to keep in sync. I see that there are likely approaches for sharing those configs but couldn't get one to work with a couple minutes of fiddling. Maybe something to explore or to file as help wanted?

@grzesiek2010, this introduces a new pattern so might be worth a quick look!

seadowg:
> with redundant directives to keep in sync

Not sure what this means? Dependencies?

lognaturel:
I mean `minSdkVersion`, `compileSdkVersion`, etc. I think we could use the `subproject` closure: https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:subproject_configuration

seadowg:
@lognaturel hmm but not all our subprojects will be Android projects (hopefully). Maybe that doesn't matter but I'm assuming that would fail for non Android projects. 

We could define version variables at the top level and use them in the Android projects?

lognaturel:
> not all our subprojects will be Android projects

Ah, right you are!

> We could define version variables at the top level and use them in the Android projects?

That seems reasonable. I suppose the errors will be easy to follow and fix when we do have to change one of the versions so let's not worry about it until it becomes a problem.

mmarciniak90:

I didn't notice regression.

Verified cases:
- rotating/locking screen when migration banner is visible
- rotating/locking screen when complete migration banner is visible
- rotating/locking screen on migration view
- open more details page  (also without browser)
- light/dark mode
- English/Spanish language
- Admin password

kkrawczyk123:

PR #3782, opened 2020-04-19
Used Dialog Fragment to retain Exit Form Dialog on Orientation Change

### LINKED ISSUE #3612: Dialogs are disappearing after screen rotation.
Collect v1.25.1, Android v4 -10.x.x, device used...

#### Problem description
Displayed dialogs are disappearing after screen rotation. 

Dialogs:
Exit form dialog
Remove this response
Layer data file dialog on geowidgets with map
Input method dialog on geotrace and geoshape dialog
Change Admin Password Dialog

1. Open one of the dialogs described above 
2. Rotate the screen

#### Expected behavior
Dialogs should not disappear after rotating the screen. 

#### Other information 
I have noticed behavior differences between Android 8.1, 9.0 and other devices on Collect 1.23.3 - dialogs are not disappearing on them after rotating the screen. 
From 1.24.0 it occurs on all devices.

SaumiaSinghal:
Work towards #3612 

I tested it on Android 9.0.

I shifted creating alert dialog inside dialog fragment so that they can be retained by the activity on configuration change. I added tests for the same.

No regression risks.

No 

No


collect_app/src/main/AndroidManifest.xml
```diff
@@ -79,7 +79,8 @@ the specific language governing permissions and limitations under the License.
         android:largeHeap="true"
         android:supportsRtl="true"
         android:usesCleartextTraffic="true"
-        android:allowBackup="true">
+        android:allowBackup="true"
+        android:theme="@style/Theme.AppCompat.Light">
```
seadowg: Why did you have to make this change?

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -80,6 +87,36 @@ public void saveForm(Uri instanceContentURI, boolean shouldFinalize, String upda
         }
     }
 
+    // Cleanup when user exits a form without saving
+    public void removeTempInstance() {
```
seadowg, line 107: Definitely good to move this to a `ViewModel` 👍

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -80,6 +87,36 @@ public void saveForm(Uri instanceContentURI, boolean shouldFinalize, String upda
...
+                int images = MediaUtils.deleteImagesInFolderFromMediaProvider(file);
+                int audio = MediaUtils.deleteAudioInFolderFromMediaProvider(file);
+                int video = MediaUtils.deleteVideoInFolderFromMediaProvider(file);
+
+                Timber.i("Removed from content providers: %d image files, %d audio files and %d audio files.",
+                        images, audio, video);
+                FileUtils.purgeMediaPath(instanceFolder);
+            }
+        } else {
+            Timber.w("null returned by getFormController()");
```
seadowg: We can probably remove this logging now as we should have a clearer picture whether the `FormController` is null or not here.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,54 @@
...
+        FragmentActivity activity = Robolectric.setupActivity(FragmentActivity.class);
+        fragmentManager = activity.getSupportFragmentManager();
+        dialogFragment = new QuitFormDialogFragment();
+    }
+
+    @Test
+    public void dialogIsCancellable() {
+        dialogFragment.show(fragmentManager, "tag");
+
+        fragmentManager.executePendingTransactions();
```
seadowg: I think you can just do `shadowOf(getMainLooper()).idle();` instead here which a more general way of running the looper.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,54 @@
...
+        assertThat(shadowOf(dialogFragment.getDialog()).isCancelable(), equalTo(true));
+    }
+
+    @Test
+    public void dismiss_shouldDismissTheDialog() {
+        dialogFragment.show(fragmentManager, "tag");
+        fragmentManager.executePendingTransactions();
```
seadowg: Same as above!

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void dialogIsCancellable() {
+        dialogFragment.show(fragmentManager, "tag");
+
+        fragmentManager.executePendingTransactions();
+        assertThat(shadowOf(dialogFragment.getDialog()).isCancelable(), equalTo(true));
+    }
+
+    @Test
+    public void dismiss_shouldDismissTheDialog() {
```
seadowg: Do we need this test? I think there are a few things that we could add tests for here but `dismiss` working feels like it's the framework's responsibility. Maybe I'm missing something here?

seadowg (changes requested):

collect_app/src/main/AndroidManifest.xml
SaumiaSinghal: Hi @seadowg! Whenever I run the tests for the newly created fragment, it gives me the error `java.lang.IllegalStateException: You need to use a Theme.AppCompat theme (or descendant) with this activity.`

seadowg: Ah right. Yes this would fix it but it would also change the default theme for the whole app which could cause us some problems. 

What's happening is that a component in the `DialogFragment` relies on the `AppCompat` theme so we can make sure that the Activity being used in tests has that theme. You can call `setTheme(R.style.Theme_AppCompat)` on the Activity or use our `RobolectricHelpers.createThemedActivity` helper.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: I tried using `shadowOf(getMainLooper()).idle()`, but it doen't work. I stil get the error `main looper has queued unexecuted runnables` , unless I use `fragmentManager.executePendingTransactions()`.

collect_app/src/main/AndroidManifest.xml
SaumiaSinghal: Yes @seadowg, I was also concerned for the default theme of the app. so I removed light, but using `RobolectricHelpers.createThemedActivity` helper would be a better option.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Dismissing the dialog on dismissing the fragment is the framework responsibility. I can think of one case if I override the `onDismiss` method and do not call super.onDestroy, it retains the dialog even when I call `dismiss`. 
It would be more interesting to add tests to check whether the dialog is retained during orientation change, when the fragment is recreated.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
seadowg: So you should be able to remove the `@LooperMode(PAUSED)` as well.

You'll see that you're getting that warning from Robolectric yes but the actual error is below that:

```
Caused by: java.lang.UnsupportedOperationException: Failed to resolve attribute at index 5: TypedValue{t=0x2/d=0x7f040251 a=-1}
	at android.content.res.TypedArray.getColorStateList(TypedArray.java:546)
	at android.widget.TextView.readTextAppearance(TextView.java:3542)
	at android.widget.TextView.__constructor__(TextView.java:959)
	at android.widget.TextView.<init>(TextView.java)
	at android.view.LayoutInflater.createView(LayoutInflater.java:647)
	at com.android.internal.policy.PhoneLayoutInflater.onCreateView(PhoneLayoutInflater.java:58)
	at android.view.LayoutInflater.onCreateView(LayoutInflater.java:720)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:788)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730)
	at android.view.LayoutInflater.inflate(LayoutInflater.java:492)
	at android.view.LayoutInflater.inflate(LayoutInflater.java:423)
	at org.odk.collect.android.adapters.IconMenuListAdapter.createView(IconMenuListAdapter.java:69)
	at org.odk.collect.android.adapters.IconMenuListAdapter.getView(IconMenuListAdapter.java:62)
```

This probably means that a layout is trying to grab an attribute (like `?colorSecondary` or `?attr/colorSecondary`) that isn't defined in the theme that's being used. It might be that something is using an attribute from our own theme or the `MaterialComponents` theme.

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private ListView listView;
+    private String title;
+
+    @Override
+    public void onAttach(@NonNull Context context) {
+        super.onAttach(context);
+        viewModel = ViewModelProviders.of(requireActivity(), viewModelFactory).get(FormSaveViewModel.class);
+
+        title = viewModel.getFormName();
```
seadowg: All this view initialization can probably go in `onCreateDialog` instead of here so you don't need the `title` and `listView` fields.

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -0,0 +1,135 @@
...
+                getDialog().dismiss();
+            }
+        });
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        super.onCreateDialog(savedInstanceState);
+        setRetainInstance(true);
```
seadowg: You shouldn't need this. Good to look at the docs and have a think about it but it's best to avoid.

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -0,0 +1,135 @@
...
+                    .setView(listView)
+                    .create();
+        }
+
+        return alertDialog;
+    }
+
+    @Override
+    public void onDestroyView() {
+        AlertDialog dialog = (AlertDialog) getDialog();
```
seadowg: What is this for?

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
seadowg: That is good thinking! I think it's potentially overkill though. I think it's best to think about tests protecting from "lazy" programmers (as we're all pretty lazy really) - protect from things being deleted or changed but not added. The universe of things that someone could add are pretty big and so it's hard to test protectively against it. Does that make sense?

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
SaumiaSinghal: Thanks @seadowg for directing to the docs. I was thinking of using it and then dismissing the dialog in `onDestroy`, so as to confirm that the dialog is destroyed when the fragment is finally detached, and not on just orientation change. But I found that it is deprecated now, so I think it is better not to use it and rely more on ViewModel.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Yes @seadowg, I think it's better if I add the test to check whether the dialog is dismissed when the user clicks the positive button? And to check whether the view is populated correctly and all the options are working or not. I think that will verify the basic functionality of the dialog.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Thanks @seadowg! I removed and got better results. I need not use `fragmentManager.executePendingTransaction` anymore.

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.inject.Inject;
+
+import static android.app.Activity.RESULT_OK;
+
+public class QuitFormDialogFragment extends DialogFragment {
+
+    @Inject
+    Analytics analytics;
+
+    @VisibleForTesting private FormSaveViewModel viewModel;
```
seadowg, line 45: Is this needed?

seadowg: Is this needed?

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -0,0 +1,134 @@
...
+                .get(FormSaveViewModel.class);
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        super.onCreateDialog(savedInstanceState);
+        AlertDialog alertDialog = (AlertDialog) getDialog();
+
+        if (alertDialog == null) {
```
seadowg: I think this will always be true right?

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,96 @@
...
+        dialogFragment.show(fragmentManager, "TAG");
+        AlertDialog dialog = (AlertDialog) dialogFragment.getDialog();
+
+        assertNotNull(dialog);
+        assertThat(dialog.getButton(DialogInterface.BUTTON_NEGATIVE).getVisibility(), equalTo(GONE));
+        assertThat(dialog.getButton(DialogInterface.BUTTON_POSITIVE).getVisibility(), equalTo(VISIBLE));
+        assertThat(dialog.getButton(DialogInterface.BUTTON_POSITIVE).getText(),
+                equalTo(activity.getString(R.string.do_not_exit)));
+
+        String title = dialogFragment.getTitle(viewModel);
```
seadowg: I'm confused by your use of `getTitle` and `setTitle` here. Could you explain why you chose to create a public `getTitle` and `setTitle`?

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,96 @@
...
+        AlertDialog dialog = (AlertDialog) ShadowDialog.getLatestDialog();
+        assertTrue(dialog.isShowing());
+
+        dialog.getButton(DialogInterface.BUTTON_POSITIVE).performClick();
+        assertFalse(dialog.isShowing());
+        assertTrue(shadowOf(dialog).hasBeenDismissed());
+    }
+
+    @Test
+    public void shouldRetainDialogOnOrientationChange() {
```
seadowg: I don't think this test does what you think it does. On a rotation the `Acitivity` will be recreated (which should cause the `DialogFragment` to be recreated). The reason to write a test around recreation would be to check that state of the fragment is persisted (the title for example). The `ProgressDialogFragmentTest` has examples of this kind of test.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
SaumiaSinghal: Oh sorry I forgot that there. Previously I was thinking of some other approach to use `viewModel` in the tests.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Oh alright. I'll make changes

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: I created the public methods so that I can use them in the `QuitFormDialogFragmentTest`.

seadowg: Right. It's generally best for us to avoid creating methods just for tests. Our tests should act like "real" clients of whatever code they are testing and so shouldn't have access to methods or fields we wouldn't want them to have access to. Feeling like you need to make a method public for a test can be a nice hint that something is probably off about the design of the object you're testing or that you're testing the wrong thing!

Here you shouldn't have to do anything but `show` the `DialogFragment` to have the title set right? And the test should be able to ascertain the title without asking the `DialogFragment` as it has control of the mock `FormSaveViewModel` so it can be set up to have a title you choose and then you can test that title is shown on the dialog.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,92 @@
...
+        AlertDialog dialog = (AlertDialog) ShadowDialog.getLatestDialog();
+        assertTrue(dialog.isShowing());
+
+        dialog.getButton(DialogInterface.BUTTON_POSITIVE).performClick();
+        assertFalse(dialog.isShowing());
+        assertTrue(shadowOf(dialog).hasBeenDismissed());
+    }
+
+    @Test
+    public void restoringFragment_retainsTitle() {
```
seadowg: Given that the only "state" in this fragment comes from the ViewModel does it make sense to have this test? There isn't any mutable state that needs to be carried from one instance of the fragment to the other as far as I can see.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void dialogIsCancellable() {
+        dialogFragment.show(fragmentManager, "tag");
+        assertThat(shadowOf(dialogFragment.getDialog()).isCancelable(), equalTo(true));
+    }
+
+    @Test
+    public void clickingCancel_shouldDismissTheDialog() {
```
seadowg, line 87: Do you still feel we need this test? Happy to talk it over some more!

seadowg (changes requested):
Comments inline

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Yes, right @seadowg. Checking it just once would be fine.

SaumiaSinghal: Also I think the part that we are checking that whether the dialog fragment shows the correct title or not can be covered with the test of the ViewModel (whether it returns the correct form name)?

seadowg: Yeah as above we just need that test I think!

SaumiaSinghal: Then, should we even remove the test to check the title of the dialog once, as it can be covered in the tests for the ViewModel?

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: It is something related to the dialog interface. We have, here the positive button set to be the cancel button. So, to check whether the correct button is set as the cancel button and clicking it does as it says (to dismiss the dialog).

seadowg: Ah of course sorry! My mistake. Good test 👍. Let's maybe make it the `NEGATIVE_BUTTON` instead as that's usually the "Cancel" button.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
seadowg: Yeah I think that make sense

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Yes, that seems more accurate.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
seadowg: Ach I'm so sorry I misinterpreted your question! My bad for skimming. 

Try removing line 60 from `QuitFormDialogFragment` and then running the tests to see if you have coverage of the title being displayed in the dialog. I don't think you do now you've removed that test. The ViewModel tests check that the `ViewModel.getFormLabel()` does the right thing but here you're testing the dialog, not the ViewModel.

SaumiaSinghal: Yes sorry @seadowg! I was a little confused too.

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertThat(dialog.getButton(DialogInterface.BUTTON_NEGATIVE).getVisibility(), equalTo(VISIBLE));
+        assertThat(dialog.getButton(DialogInterface.BUTTON_NEGATIVE).getText(),
+                equalTo(activity.getString(R.string.do_not_exit)));
+    }
+
+    @Test
+    public void shouldShowCorrectTitle() {
+        dialogFragment.show(fragmentManager, "TAG");
+        AlertDialog dialog = (AlertDialog) ShadowDialog.getLatestDialog();
+        String title =  formSaveViewModel.getFormName() == null ? activity.getString(R.string.no_form_loaded)
```
seadowg: The problem with this is that it's unclear what the test wants the title to be. For "unit" tests like this you want you test to be as in charge of the dependencies as possible. Given the ViewModel is a mock you can set the value that will be returned from `getFormName` like so:

```
when(formSaveViewModel.getFormName()).thenReturn("blah");
```

This means you could write a test for both the null and non null case. Here I I'd imagine you're only testin the null case as the mock will return null from `getFormName()`.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Oh alright yes. We should test both the cases

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -66,6 +67,21 @@ public void shouldShowCorrectTitle() {
         assertThat(dialogTitle.getText().toString(), equalTo(title));
     }
 
+
+    @Test
+    public void shouldShowCorrectTitle_whenFormIsLoaded() {
```
seadowg, line 70: Yeah I like these tests divided up like this a lot! 👍

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
```diff
@@ -66,6 +67,21 @@ public void shouldShowCorrectTitle() {
...
+        when(formSaveViewModel.getFormName()).thenReturn("blah");
+
+        dialogFragment.show(fragmentManager, "TAG");
+        AlertDialog dialog = (AlertDialog) ShadowDialog.getLatestDialog();
+
+        String title =  formSaveViewModel.getFormName() == null ? activity.getString(R.string.no_form_loaded)
```
seadowg: You don't need this logic in either test right? The test knows what value the form name will be.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/formentry/QuitFormDialogFragmentTest.java
SaumiaSinghal: Ah sorry! This piece of code is not required.

seadowg (approved):

seadowg:
@SaumiaSinghal I edited your PR description just so it didn't close the issue but still kept it linked.

SaumiaSinghal:
Thanks @seadowg! I was wondering from the beginning on this PR if there should be a unit test to check whether the dialog is dismissed or not, when the user clicks on any item in the ListView in the dialog. But, whenever I try `dialog.getListView()` or `shadowAlertDialog.getView(),` it gives NPE.

I was thinking if this test is necessary or not.

seadowg:
@SaumiaSinghal I could go either way on that. That does feel like more of an Espresso level test to me as both options are part of important user flows. I think we could add more but what we have is good for now 👍

mmarciniak90:
@SaumiaSinghal, I still see an issue with disappearing dialogs after screen rotation.
Could you look at this again?

Verified dialogs:
- Exit form dialog :heavy_check_mark: fixed
- Remove this response dialog :no_entry_sign: 
- Layer data file dialog on geowidgets with map dialog  :no_entry_sign: 
- Input method dialog on geotrace and geoshape dialog  :no_entry_sign: 
- Change Admin Password dialog  :no_entry_sign: 

mmarciniak90:
@SaumiaSinghal I see that you split work at #3612 to a few PRs.

Could you define what dialogs from the mentioned list are fixed in this PR?
Is it only the `Exit form` dialog?

mmarciniak90:

SaumiaSinghal:
Hi @mmarciniak90! Yes, it is only for the `Exit form` dialog.

mmarciniak90:
**Tested with success**

Verified on Android: 6.0, 7.0, 8.1, 9.0 and emulator with Android 5.1

I confirm that `Exit form` dialog does not disappear after rotation.

This PR does not close the whole #3612 issue. 

PR #3791, opened 2020-04-20
Replace opendatakit with getodk in README references

seadowg:
This fixes references (including URLs in links) that were using the `opendatakit` instead of `getodk`. I've left the JavaRosa maven references the same as I imagine they won't change until we do another release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # ODK Collect
 ![Platform](https://img.shields.io/badge/platform-Android-blue.svg)
 [![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
-[![Build status](https://circleci.com/gh/opendatakit/collect.svg?style=shield&circle-token=:circle-token)](https://circleci.com/gh/opendatakit/collect)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,11 +72,11 @@ We try to make sure that all issues in the issue tracker are as close to fully s
     **Note:** You can see the emulator setup used on CI in  `.circleci/config.yml`.
 
 ## Testing a form without a server
-When you first run Collect, it is set to download forms from [https://opendatakit.appspot.com/](https://opendatakit.appspot.com/), the demo server. You can sometimes verify your changes with those forms but it can also be helpful to put a specific test form on your device. Here are some options for that:
```
yanokwa: @seadowg I won't be setting up a new appspot instance so this should stay the stay the same for now. 

@ln Maybe we can setup demo.central.getodk.org and get rid of all the appspot references? This should ideally be paired with a change in the defaults of Collect.

seadowg: Whooops. Sorry I should have checked that one.

lognaturel:
Closing in favor of #3794. Looks like you and @MartijnR were looking at this just about at the same time! He made a couple other good changes like using the xlsform URL for the form converter instead of xiframe.

PR #3792, opened 2020-04-20
Retain Input Setting Dialog on GeoPolyActivity on orientation change

SaumiaSinghal:
Work towards #3612 

I tested it.

I used a separate variable `IS_SETTINGS_DIALOG_ACTIVE` to check whether the dialog is active or not, and if the dialog was active last when the activity was destroyed on configuration change, then the dialog is displayed again in `onCreate`.

No regression risks

All Widgets form or any form with Geo widgets 

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -14,28 +14,27 @@
 
 package org.odk.collect.android.activities;
 
-import androidx.appcompat.app.AlertDialog;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.view.View;
 import android.view.Window;
-import android.widget.AdapterView;
-import android.widget.ArrayAdapter;
 import android.widget.Button;
 import android.widget.ImageButton;
 import android.widget.RadioGroup;
-import android.widget.Spinner;
 import android.widget.TextView;
 
 import androidx.annotation.VisibleForTesting;
+import androidx.appcompat.app.AlertDialog;
 
 import org.odk.collect.android.R;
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
```diff
@@ -111,6 +112,8 @@
     private Double restoredMapZoom;
     private List<MapPoint> restoredPoints;
 
+    private static boolean isSettingsDialogActive;
```
seadowg: We shouldn't use static state like this. If one instance of `GeoPolyActivity` set this to `true` and then another was opened later it would behave as if the dialog was open. We could manage all this with extra logic but it'd be a headache. I think you should look at a solution using a `DialogFragment` like you did for the quit form dialog.

I'll convert this PR to a draft so you can make changes in peace.

lognaturel (approved):
It would be nice to use view binding to cut down the noise in the new fragment class and perhaps use a view model to mediate communication between the activity and fragment (e.g. `getCheckedId` feels awkward as does the check on `checkedRadioButtonId == -1`) but I think this is a nice change and we can start here.

SaumiaSinghal:
Thanks for reviewing @lognaturel! I think the same, seeing the communication between fragment and activity, I think it would be better to use `viewModel`. I can try implementing the `viewModel` and maybe write separate tests for the `viewModel` later.

mmarciniak90:
**Tested with success**

Verified cases:
- Geotrace, Geoshape widgets
- Google Maps, Mapbox, OSM
- rotation dialog with selected each of 3 option
- rotation when recording interval options list or accuracy requirement options list is expanded
- minimization
- light/dark mode
- English/Spanish language

kkrawczyk123:

PR #3795, opened 2020-04-21
Disable coverage reporting on CI

seadowg:
And again...

It looks like we're going to need to rework how coverage reports work now that we're sharding our tests. Disabling the upload step makes the most sense for the moment I think!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -243,11 +243,11 @@ workflows:
           filters:
             branches:
               only: master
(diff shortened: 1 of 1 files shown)


seadowg:
Ach something has gone wrong here.

seadowg:
Ok fixed. For some reason Github chose a new branch as the base branch for the PR.

grzesiek2010 (approved):

PR #3803, opened 2020-04-27
Add extra assertion when swiping to end screen

seadowg:
This may make out tests more stable as the assertion hopefully prevents us from trying to click on views to early but if not it will at least give us clearer failures.


lognaturel (approved):

lognaturel:
I think it's a good concept. Looks like another test failed similarly in the latest run. What do you think about identifying every call on `clickSaveAndExit` and making this change?

seadowg:
@lognaturel yeah I think that's the next step!

PR #3804, opened 2020-04-28
Fixed maps offline

### LINKED ISSUE #3660: With Mapbox or Google mapping engines, offline tiles are not rendered without an Internet connection
Collect v1.25.2 and prior

#### Problem description
See https://forum.opendatakit.org/t/mapbox-mbtiles-layer-not-loading-without-network-connection/21774 for user report.

1. Set the basemap source to Mapbox or Google
1. Push an mbtiles file to `/sdcard/odk/layers`
1. Turn off all networking
1. Select that mbtiles file as a reference layer
1. Go to a map-based question

#### Expected behavior
The offline reference layer should always show, even if the device is offline.

grzesiek2010:
Closes #3660

I tested GoogleMaps and MapBox offline and confirmed that if during installation the device is connected to the internet it will work.

Maybe the amount of code I have added is not impressive but it wasn't easy to figure out what's wrong so I also should explain.

The first problem was that we need to initialize maps when we are online to be able to use them when we are offline. It's a know limitation: https://issuetracker.google.com/issues/35823181 and it works in the same way for GoogleMaps and MapBox. The initialization took place when the map is opened so a user had to open a widget with maps to make sure it will work offline in the future.
I improved this behavior adding initialization when the app is installed: 647dc72 7507d08 of course I can't be sure that a device is online during installation but it's much more likely.

The second problem was that MapBox didn't work offline even if we initialized it before (according to what I described above). Here the problem is that MapBox doesn't support using mbtiles offline and in order to do that we implemented a workaround [TileHttpServer](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/geo/TileHttpServer.java) that uses localhost. Unfortunately Mapbox by default detects if a device is offline and doesn't even send requests in such a case. Our implementation is as I noticed based on https://gist.github.com/typebrook/7d25be326f0e9afd58e0bbc333d2a175 and there that limitation is also mentioned. Fortunately I was able to find a fix for it as well 71e8ba5

It should just fix the problems described in the issue and explained more here. It should affect anything else so testing maps would be enough.

Any form with maps, it might be `All widgets` form.

The only thing I can come up with is that we could add somewhere a note that users that use offline maps should make sure they have been initialized. The best way is just open maps when internet connection is available. I try to do that during installation and in 99% cases it should work but some users might install the app not from the store but offline using .apk files. What do you think? is it worth the effort?

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
@@ -29,6 +29,9 @@ import android.view.Window;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 
+import com.mapbox.mapboxsdk.maps.MapView;
+import com.mapbox.mapboxsdk.maps.Style;
+
 import org.odk.collect.android.R;
 import org.odk.collect.android.analytics.Analytics;
 import org.odk.collect.android.application.Collect;
@@ -83,6 +86,7 @@ public class SplashScreenActivity extends Activity {
                 }
 
                 init();
+                initMapBox();
             }
 
             @Override
@@ -224,5 +228,10 @@ public class SplashScreenActivity extends Activity {
             }
         };
         t.start();
+    }
+
+    private void initMapBox() {
+        MapView mapView = findViewById(R.id.mapView);
+        mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
     }
 }
--- a/collect_app/src/main/res/layout/splash_screen.xml
+++ b/collect_app/src/main/res/layout/splash_screen.xml
@@ -7,6 +7,12 @@
     android:orientation="vertical"
     android:padding="10dip">
 
+    <com.mapbox.mapboxsdk.maps.MapView
+        android:id="@+id/mapView"
+        android:layout_width="match_parent"
+        android:layout_height="1dp"
+        android:visibility="invisible"/>
+
     <ImageView
         android:id="@+id/splash"
         android:layout_width="wrap_content"
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -184,6 +184,7 @@ public class Collect extends Application {
         }
 
         setupOSMDroid();
+        initMapProviders();
 
         // Force inclusion of scoped storage strings so they can be translated
         Timber.i("%s %s", getString(R.string.scoped_storage_banner_text),
@@ -192,6 +193,10 @@ public class Collect extends Application {
 
     protected void setupOSMDroid() {
         org.osmdroid.config.Configuration.getInstance().setUserAgentValue(userAgentProvider.getUserAgent());
+    }
+
+    private void initMapProviders() {
+        new com.google.android.gms.maps.MapView(this).onCreate(null);
     }
 
     private void setupDagger() {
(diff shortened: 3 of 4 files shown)


collect_app/src/main/res/layout/splash_screen.xml
```diff
@@ -7,6 +7,12 @@
     android:orientation="vertical"
     android:padding="10dip">
 
+    <com.mapbox.mapboxsdk.maps.MapView
+        android:id="@+id/mapView"
+        android:layout_width="match_parent"
+        android:layout_height="1dp"
+        android:visibility="invisible"/>
```
grzesiek2010, line 14: Can't be gone or with 0dp/0dp I checked it.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -225,4 +229,9 @@ public void run() {
         };
         t.start();
     }
+
+    private void initMapBox() {
```
grzesiek2010, line 233: I know the fix for Mapbox doesn't look good you would probably ask: can't we do the same what works for Google: https://github.com/getodk/collect/pull/3804/commits/647dc72d81aeeefc1a7f483490714950d403b31f?
Unfortunately we can't we need to display that map even if it's not visible like in this case to make sure `getMapAsync` and then `setStyle` is called. Only then MapBox is initialized properly and we can use it offline.

seadowg:
I'm not sure `onCreate` is called after installation - as far as I understand it's [called when the process starts](https://developer.android.com/reference/android/app/Application.html#onCreate()). However I think this is still a good idea as it feels more likely we'll get our "first open" right after installation and therefore while connected to the internet. I"m wondering if to truly close off this issue we're going to need to schedule work to run when we get an internet connection and then have some warning for the user that offline tiles aren't available yet.

In this solution does the user still have to have offline tiles in place **BEFORE** initialization or is initializing the map frameworks while connected once enough that they can copy offline tiles later?

grzesiek2010:
> I"m wondering if to truly close off this issue we're going to need to schedule work to run when we get an internet connection and then have some warning for the user that offline tiles aren't available yet.

It's a good idea but then if we want more improvements maybe it would be good to have a separate pr since the issue this one solves was just about offline layers not about initializing. @lognaturel what do you think?

> In this solution does the user still have to have offline tiles in place BEFORE initialization or is initializing the map frameworks while connected once enough that they can copy offline tiles later?

initializing is enough you don't need to use offline layers when you are online

seadowg:
> It's a good idea but then if we want more improvements maybe it would be good to have a separate pr since the issue this one solves was just about offline layers not about initializing. 

Yeah I agree that should probably be another PR. I'm thinking we shouldn't close the issue yet as there is still a problem if the app is opened while offline.

lognaturel (approved):
Thanks for the extra comments to help make sense of this. I think this is a good improvement and should be easy to expand on. As far as scheduling the initialization, I think an easy thing we could do is triggering it after the first successful form list download.

grzesiek2010:
I added a separate issue for that improvement you can share your thoughts there https://github.com/getodk/collect/issues/3815

kkrawczyk123:
**Tested with success**

Tested scenarios:
- Google maps with offline layer
- Mapbox with offline layer
- OSM: with offline layer, just regression check
- online 
- offline
-geo point, geotrace, geoshape

mmarciniak90:

_Google Maps_
* The set Offline layer is applied even if Geo widget wasn't opened when the internet connection was available
* Geopoint, Geotrace, Geoshape
* online

_Mapbox_
* The set Offline layer is applied 
* Geopoint, Geotrace, Geoshape
* online

_No changes on OSM_

PR #3808, opened 2020-04-29
Disable Jacoco in debug build (by default)

### LINKED ISSUE #3262: Build sometimes hangs on jacocoDebug step
Android Studio 3.4.1, Gradle 3.4.0

I suddenly started experiencing this frequently a couple of weeks ago. @zestyping has mentioned getting it as well. 

When I run from Android Studio, typically the jacocoDebug step takes about 1 second. Occasionally, it will hang for several minutes and my CPU fans will go full blast. I have not investigated this. I'm hoping that when @grzesiek2010 updates all the things after v1.23 goes out, it will go away. 

In the mean time, if anyone else experiences this, it would be helpful to gather a bit more information about the conditions.

seadowg:
Closes #3262.

This disables Jacoco by default in the `debug` build. It can still be enabled by passing `enableJacoco` to a task like so:

```
./gradlew -PenableJacoco=true myTask
```

Ran different build types to check Jacoco is disabled without the flag and enabled with the flag. I also double checked that Android Studio's coverage tool still works (without the flag).

I considered writing docs but I felt that updating CI's config was better as it'd generally be the first place I looked to see how build tasks work for a project.

We could try and work out why the task was occasionally slow but this feels better as unless we want a coverage report it's always a waste of time to have Jacoco's build tasks running.

No change to users. Should be a straight merge once approved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,11 +109,11 @@ jobs:
       - run:
           name: Assemble debug build
           command: |
-            ./gradlew assembleDebug -PdisablePreDex
+            ./gradlew -PenableJacoco=true assembleDebug -PdisablePreDex
       - run:
           name: Assemble test build
           command: |
-            ./gradlew assembleDebugAndroidTest -PdisablePreDex
+            ./gradlew -PenableJacoco=true assembleDebugAndroidTest -PdisablePreDex
       - run:
           name: Run code quality checks
           #           When running the plain lint target, it will execute by default for all flavors.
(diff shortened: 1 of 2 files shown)


lognaturel:
> Android Studio's coverage tool still works (without the flag)

And you checked that for connected and robolectric tests too?

seadowg:
@lognaturel as far as I've seen you can't run connected tests "with coverage" but yeah it works for all the JVM tests. It uses a special runner to do it I think.

lognaturel (approved):

lognaturel:
> you can't run connected tests "with coverage"

🤯 Why was I convinced that was possible? Confirmed it's not a thing on `master`, either.

seadowg:
@lognaturel I absolutely thought that was a thing as well before I was looking into the coverage stuff on CI.

Issue #3811, opened 2020-05-05
Not all settings are imported from older Collect versions


seadowg:
#### Software and hardware versions 
Collect v1.26.2 (and Collect v1.22.0)

#### Problem description

Settings that have been migrated in someway (in their storage in Android's `SharedPreferneces`) aren't imported correctly when going from an old version (before they were migrated) to a new version (after they were migrated).

#### Steps to reproduce the problem

1. Open Collect 1.22.0
2. Change the Map SDK (General Settings > User interface > Mapping SDK) to "OpenStreetMap SDK"
3. Open Collect 1.26.2 
4. Change the Map SDK (General Settings > Maps > Source) to  "Google"
5. Scan the QR code from Collect 1.22.0 in Collect 1.26.2

You'll see that the Maps "Source" is still "Google" rather than "OpenStreetMap".

#### Expected behavior

Settings that had preference migrations applied to them should still import correctly from QR codes or settings files. This means that the import code (potentially `PreferenceSaver`) needs to perform the migrations in `PrefMigrator` in someway.

seadowg:
@grzesiek2010 have you made any progress on this? It looks like we probably need this fixed for #3818 and I think this should really follow on from changes in #3813 so I'm happy to take this on if not!

grzesiek2010:
Not yet do you want to fix it as a part of #3813 if so no problem.

seadowg:
Yeah I can build on top of that! Thanks.

Issue #254, opened 2016-11-18
Remove splash screen from ODK Collect


yanokwa:
The splash screen was added to enable branding. There continue to be forks of Collect because the splash screen with a custom logo is not adequate. I would like to remove the splash screen because it is not sufficient and focus instead on the solutions in #233

### Notes

We should make it so the new [Android 12 Splash Screen](https://developer.android.com/guide/topics/ui/splash-screen) is shown instead of our one (investigation into this in #4592). As part of this work we should make app initialization run while the splash screen is up (using the [new API](https://developer.android.com/guide/topics/ui/splash-screen#suspend-drawing)).

This would be a perfect time to consider updating the Collect icon...

jnordling:
Do this PR relate to his issue? Skip intro screen #224

lognaturel:
Hehe, @jnordling seems they would be related but no. This one refers to the screen that shows on app launch. The other one is the form instruction screen. So many intro screens!

grzesiek2010:
Two months ago we started login events about splash screen https://github.com/grzesiek2010/collect/commit/c882a10ee332ecc70f4083453f03a3fe84d57af8

Now I can see 305 events https://analytics.google.com/analytics/web/#/report/app-content-event-events/a91951913w136214323p140411927/_u.date00=20190118&_u.date01=20190324&explorer-table.plotKeys=%5B%5D&_r.drilldown=analytics.eventCategory:PreferenceChange

It's not a big number and generally, I don't like this option (I think that if someone wants his own name/logo, he should just fork our repo and implement required changes, we don't allow to change the logo or app name so why that image), so I would be in favor of removing it. 

We should at least remove logging those events because in some cases it causes a crash https://console.firebase.google.com/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5c63005ef8b88c296394c625?time=last-seven-days&sessionId=5C9744F4031B0001197497531BFDFD8A_DNE_0_v2

but as I said I would like to remove both.

@yanokwa what do you think?

shobhitagarwal1612:
Are you in favour of removing the ability to change splash screen image or the entire splash screen itself?

Maybe we can do a forum polling as well to get user opinion.

grzesiek2010:
> Are you in favour of removing the ability to change splash screen image or the entire splash screen itself?

Generally both but at least the ability to change splash screen image.

grzesiek2010:
[...]

PR #3813, opened 2020-05-06
Remove KEYS_WE_SHOULD_NOT_RESET

### LINKED ISSUE #3669: Replace KEYS_WE_SHOULD_NOT_RESET with "meta" shared preferences
In https://github.com/opendatakit/collect/pull/3640/files#diff-bf808238c0e5c486c12b05f9477c0fb2, @seadowg introduced a new preference file named `meta` to hold values relevant to Collect's behavior but not surfaced as a user preference. This makes for a nice separation and ensures that we won't do something like accidentally reset those preferences with the reset utility or allow users to change them with preference files. Currently the `KEYS_WE_SHOULD_NOT_RESET` collection is used for this purpose but that requires remembering to manually update it which is error-prone and makes the code more complex.

This will require migrating the prefs in `KEYS_WE_SHOULD_NOT_RESET`.

Marking as `needs discussion` until @seadowg has a chance to review and make sure I captured the intent. @seadowg, it probably makes sense for you to take this on?

seadowg:
Closes #3669

This moves keys to that should not be reset to the new meta shared preferences bucket. Most of  of these keys were removed during this work:

* `KEY_LAST_VERSION` and `KEY_FIRST_RUN` - these keys were only used to show the splash screen on the first run of each new version of Collect. After discussing with @lognaturel, we decided that this wasn't a feature we needed any longer.
* `KEY_METADATA_MIGRATED` - this was used as part of data migration to allow the server username to be copied to the newer metadata username. This was really only needed for a couple of releases after the new fields were added.
* `KEY_AUTOSEND_WIFI` and `KEY_AUTOSEND_NETWORK` - it didn't seem like these should have been avoiding reset. They are the old keys used for autosend prefs and are migrated for old version upgrades. If @grzesiek2010 has a reason then need to stay in then please shout at me!

Added new tests for the new migrations.

There shouldn't be any change in behaviour but checking version upgrades work ok and that resetting the application works as expected would be good. We found #3811 as part of working on this as a heads up.

There is a change in behaviour to the splash screen. Before the splash screen would show on the first run of the app (fresh install) and the first run after an update as well as when it was enabled in User Interface settings. The splash screen will now only show when it is enabled.


collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -63,14 +61,23 @@
     @Inject
     Analytics analytics;
 
+    @Inject
+    ApplicationInitializer applicationInitializer;
```
grzesiek2010: It's unused so we can remove it.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -148,41 +131,14 @@ public void onCreate() {
...
-
-        reloadSharedPreferences();
-
-        AppCompatDelegate.setCompatVectorFromResourcesEnabled(true);
-        JodaTimeAndroid.init(this);
-
-        defaultSysLanguage = Locale.getDefault().getLanguage();
-        new LocaleHelper().updateLocale(this);
-
         initializeJavaRosa();
```
grzesiek2010, line 146: Why this or registering receivers or setupOSMDroid() is not a part of applicationInitializer?

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -0,0 +1,107 @@
...
+        generalSharedPreferences = GeneralSharedPreferences.getInstance();
+        adminSharedPreferences = AdminSharedPreferences.getInstance();
+    }
+
+    public void initializePreferences() {
+        performMigrations();
+        reloadSharedPreferences();
+    }
+
+    public void initializeFrameworks() {
```
grzesiek2010, line 54: I think it would be good to factor out more methods here like `initializeTimber`/`initializeJobManager` etc. It might be strange to create separate methods for those one line initializations so it's ok to keep them in one method.

grzesiek2010 (changes requested):
Generally looks great! Just some small comments.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
seadowg: Oh I'd meant to come back to these! JavaRosa is a little complex as `initializeJavaRosa` is called in lots of places but I think we should be able to include OSMDroid at least. I'll look at this again.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: I think you're right it's pretty hard to read that method!

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
seadowg: OK moved the map stuff in (the OSM problem we were seeing in Robolectric before has apparently gone now which is nice). I decided to keep the receivers and the JavaRosa in `onCreate`. For the former it's because they are directly linked to the application object itself. For the latter it's because that method is called in many places and I think that's a rework for another time 😁

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -88,7 +70,7 @@
     UserAgentProvider userAgentProvider;
```
grzesiek2010: It is unused now.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -0,0 +1,119 @@
...
+        NotificationUtils.createNotificationChannel(context);
+        initializeJobManager();
+        JodaTimeAndroid.init(context);
+        initializeLogging();
+        AppCompatDelegate.setCompatVectorFromResourcesEnabled(true);
+
+        new com.google.android.gms.maps.MapView(context).onCreate(null);
```
grzesiek2010: Is there any reason you unpacked this? I mean having it in a separate method like `initializeMaps` or `initializeMapProviders` seems more elegant.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
seadowg: Ah nice catch. I really dislike that Dagger's generated code sometimes fools Android Studio into thinking a field is used.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: It felt fine to have them in the `initializeFrameworks` to me but I don't have any strong feelings on it. I'll put the method back but call it from the same place.

grzesiek2010 (approved):
Looks good to me there are just  some conflicts to solve.

seadowg:
I've just added a change to account for the `MAPBOX_INITIALIZATION` key that was added last week.

lognaturel:
👍 Can you please describe the intended behavior for the splash screen in the PR template so QA knows what to look for? I think another slight area of risk is all things app initialization since that code moved around. In particular, it seems worth doing a quick check of automatic form downloads.

seadowg:
@lognaturel good call. I've updated the regression section.

kkrawczyk123:
@seadowg Could you please provide more details about the expected behavior change in splash image? I can't see any except one: on Android 5.1 I am not able to select image from device internal storage as splash image - Collect is crashing but unfortunately, I can't see any significant stacktrace from that crash. I can easily add it on master.

seadowg:
@kkrawczyk123 ah right I think it might not be quite right actually. Sorry this PR has been sitting here for a while so I'm bit out of context with my own work. I'll fix conflicts and have a look at this later today. Feel free to move on to testing other PRs in the meantime.

kkrawczyk123:

seadowg:
@kkrawczyk123 Does the 5.1 behaviour exist on master as well? I don't think this PR should change anything with splash screen settings...

I've updated the description of the intended behaviour of the splash screen:

> There is a change in behaviour to the splash screen. Before the splash screen would show on the first run of the app (fresh install) and the first run after an update as well as when it was enabled in User Interface settings. The splash screen will now only show when it is enabled.

We had talked about doing more work to remove the splash screen default image and only show it if the user configures a custom image. That's a separate feature defined in #254 but I got confused as it'd been a while since I'd thought about any of this.

kkrawczyk123:
@seadowg no, I was able to add a splash image from the device's gallery on master on Android 5.1.

seadowg:
@kkrawczyk123 I can take a look

seadowg:
@kkrawczyk123 I can't reproduce. Could you take another look now I've fixed conflicts? If you do see it again it would be interesting to know the exact repro steps and also to post the image you're trying to use.

lognaturel:
@seadowg is it possible you didn't push after fixing conflicts or are there new conflicts? Would be good to be conflict-free before @kkrawczyk123 tries again.

seadowg:
@lognaturel new conflicts! Will fix.

mmarciniak90:
Tested with success
I didn't notice regression in test cases:
- version upgrades
- resetting the application
- scan QR code with splash screen set
- splash screen - appears with first run
- splash screen - reset
- splash screen - default and custom image
- auto send to aggregate
- auto send to 
- auto send - wifi or cellular
- automatic download newer form version
- delete after send

mmarciniak90:
**Tested with success**

Verified on Android: 6.0, 8.1, 9.0

I didn't notice regression in test cases:
-  version upgrade
-  resetting the application
-  scan QR code with splash screen set
-  splash screen - appears with first run
-  splash screen - reset
-  splash screen - default and custom image
-  auto send to aggregate
-  auto send data defined in a form
-  auto send to google sheet
-  auto send - wifi or cellular
-  automatic download newer form version
-  delete after send

kkrawczyk123:

PR #3814, opened 2020-05-07
Add analytics that will help shape changes to the onboarding experience

lognaturel:
Adds analytics for form downloads triggered from Get Blank Form. I also verified that we already track usage of the automatic form update settings.

Wrote and ran tests.

I had trouble deciding how to structure the download analytics because that area of the code doesn't have patterns to follow. There's a viewmodel but it's just holding state that was pulled out to support rotation. I felt there was enough logic involved in the analytics that I wanted the support of tests and I ended up putting that logic in the view model and testing it there. I put the analytics call in the activity and wrote a Robolectric test for that. If any of the tests seem superfluous, you're uncomfortable with the code changes, or you have a better structure to propose, let me know.

A long time ago now, @shobhitagarwal1612 had put in a PR to add some testing here and it turned into a risky refactor with some behavior regressions. I feel badly that no one had the cycles to work with him on that at the time because there are a ton of challenges that would be good to address in this area of the code.

Particularly dangerous is that the naming in both `FormDownloadListActivity` and `FormDownloadListViewModel` is  misleading. I tried to improve it for things I was touching but I think there are still some possibly problematic cases. For example, the viewmodel keeps track of `url` which seems like it could be the URL for the server that is currently configured. It is, but it's only set if the activity is launched from an external intent. In general, the external intent path seems hard to understand and brittle. There are other examples of fields that seem like they should have certain meaning but don't or are often null. Because there are no domain concepts, everything is very low level and some information is duplicated making the data flow hard to follow.

Other things that may need follow-up and/or just bothered me:
- Why is a homegrown insertion sort used to order form definitions in `formListDownloadingComplete` rather than putting them in a list as they come in and then sorting?
- The `item` hashmap built in `formListDownloadingComplete` is hard to understand and should be replaceable by the `Form` domain object, possibly with methods added.
- I really wanted to have a higher level `FormsRepository` rather than using the "dao" abstraction but held back since it was just to get a form count.
- I think `InstanceUploaderListActivity` is missing a call to create the ODK storage directory. Since that class can be launched via intent, that could be a problem if Collect was never launched another way.
- The method naming in `AppDependencyModule` is not consistent. Sometimes the prefix is `provide` and sometimes it's `provides`.

There should be no change to behavior. There is some risk to form download behavior since I did do some refactoring there but I tried to keep it as minimal as possible.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
@@ -205,10 +205,10 @@ public class FormDownloadListActivity extends FormListActivity implements FormLi
             public void onClick(View v) {
                 downloadButton.setEnabled(toggleChecked(listView));
                 toggleButtonLabel(toggleButton, listView);
-                viewModel.clearSelectedForms();
+                viewModel.clearSelectedFormIds();
                 if (listView.getCheckedItemCount() == listView.getCount()) {
                     for (HashMap<String, String> map : viewModel.getFormList()) {
-                        viewModel.addSelectedForm(map.get(FORMDETAIL_KEY));
+                        viewModel.addSelectedFormId(map.get(FORMDETAIL_KEY));
                     }
                 }
             }
@@ -262,7 +262,7 @@ public class FormDownloadListActivity extends FormListActivity implements FormLi
                 }
                 downloadFormsTask = null;
             }
-        } else if (viewModel.getFormNamesAndURLs().isEmpty()
+        } else if (viewModel.getFormDetailsByFormId().isEmpty()
                 && getLastCustomNonConfigurationInstance() == null
                 && !viewModel.wasLoadingCanceled()) {
             // first time, so get the formlist
@@ -288,9 +288,9 @@ public class FormDownloadListActivity extends FormListActivity implements FormLi
         downloadButton.setEnabled(listView.getCheckedItemCount() > 0);
 
         if (listView.isItemChecked(position)) {
-            viewModel.addSelectedForm(((HashMap<String, String>) listView.getAdapter().getItem(position)).get(FORMDETAIL_KEY));
+            viewModel.addSelectedFormId(((HashMap<String, String>) listView.getAdapter().getItem(position)).get(FORMDETAIL_KEY));
         } else {
-            viewModel.removeSelectedForm(((HashMap<String, String>) listView.getAdapter().getItem(position)).get(FORMDETAIL_KEY));
+            viewModel.removeSelectedFormId(((HashMap<String, String>) listView.getAdapter().getItem(position)).get(FORMDETAIL_KEY));
         }
     }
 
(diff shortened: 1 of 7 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
```diff
@@ -191,11 +204,10 @@ private void init(Bundle savedInstanceState) {
...
         downloadButton = findViewById(R.id.add_button);
         downloadButton.setEnabled(listView.getCheckedItemCount() > 0);
-        downloadButton.setOnClickListener(new OnClickListener() {
-            @Override
-            public void onClick(View v) {
-                downloadSelectedFiles();
-            }
+        downloadButton.setOnClickListener(v -> {
+            ArrayList<FormDetails> filesToDownload = getFilesToDownload();
+            logDownloadAnalyticsEvent();
```
seadowg: How would you feel about moving all of the analytics logging code to the ViewModel here? Ideally `startFormDownload` would be a ViewModel method and the analytics would be encapsulated within there but right now we could move logDownloadAnalyticsEvent to the ViewModel and then all the analytics testing/concern could be hidden from the Activity (you might still want a test to check the method is called).

seadowg (changes requested):
That review was meant to be "request changes" but Github borked

seadowg:
> The method naming in AppDependencyModule is not consistent. Sometimes the prefix is provide and sometimes it's provides.

Feel free to go ahead and change them all! They could just be `getX` to avoid that happening in future. The method name doesn't matter to Dagger.

collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
lognaturel: Makes sense. 👍

lognaturel:
> Feel free to go ahead and change them all! 

Something for a less hectic time.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormDownloadListViewModel.java
```diff
@@ -207,4 +226,37 @@ public boolean wasLoadingCanceled() {
...
         this.loadingCanceled = loadingCanceled;
     }
+
+    public void logDownloadAnalyticsEvent(int downloadedFormCount, String serverUrl) {
+        String analyticsEvent = getDownloadAnalyticsEvent(downloadedFormCount);
+        String analyticsDesc = getDownloadAnalyticsDescription(serverUrl);
+        analytics.logEvent(analyticsEvent, analyticsDesc);
+    }
+
+    String getDownloadAnalyticsEvent(int downloadedFormCount) {
```
seadowg: These methods (`getDownloadAnalyticsEvent` and `getDownloadAnalyticsDescription`) feel like they should be private now as they're just used as helpers for `logDownloadAnalyticsEvents` and don't feel like they should be part of the external interface of the object. That would also mean reworking the tests a bit.

seadowg (changes requested):
One comment inline.

seadowg (approved):

seadowg:
I'll merge when we have a green build!

Issue #3812, opened 2020-05-05
Simplify Server settings screen by removing Other type and replacing ODK Aggregate with ODK


lognaturel:
See description at https://forum.getodk.org/t/collect-server-settings-screen-remove-other-server-type-and-replace-odk-aggregate-with-odk/26299

This should not involve any changes to the settings keys because server types ODK Aggregate and Other currently share the URL/username/password parts of their configurations.

While here, it will likely make sense to migrate settings to the androidx libraries.

seadowg:
![](https://media.giphy.com/media/3ornk57KwDXf81rjWM/giphy.gif)

@getodk-bot still working on it my little friend

seadowg:
![](https://media.giphy.com/media/3ornk57KwDXf81rjWM/giphy.gif)

@getodk-bot still working on it my little friend

seadowg:
Finishing the last part of this (converting the prefs screen to AndroidX) keeps getting pushed back as other things come up. I think it'd be best to admit defeat here and make a new issue for the conversion. After looking at it again I think there needs to be more test coverage written up before ripping the screen apart and I think it'd be better to write those tests and do the conversion as part of some user visible work. 

Sound good @lognaturel? If you agree I think we can close this and then just post a PR with my fix for #3901.

EDIT: And we should just add a note to the `STATE.md` about moving to AndroidX.

lognaturel:
Yes, that sounds good to me.

PR #3818, opened 2020-05-08
Server settings

seadowg:
Work towards #3812. ~~Depends on #3777.~~ I'll move straight onto to a refactor of the `ServerPreferencesFragment` (mainly migrating it to AndroidX) which I think we should do before closing the issue. 

This reworks the Server settings screen to match the designs/flow in the issue.

Expanded existing to tests to cover custom server path funcionality as well as the settings.

To echo my comment above I think we really should move this settings screen to AndroidX but I'll follow up with that to actually close out the issue.

Key regression risk would be that server settings aren't applied in someway. Good to check setting and changing server settings (including Google sheets). I'd like to add more test coverage to these screens as part of the upcoming refactor to lower the risk of changes here.

~~It probably does right? I'll add an issue.~~ Filed at https://github.com/getodk/docs/issues/1219


collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/CustomServerPathsTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.FeatureTestRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.StubOpenRosaServer;
+import org.odk.collect.android.support.pages.GeneralSettingsPage;
+import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.utilities.UserAgentProvider;
+
+@RunWith(AndroidJUnit4.class)
+public class CustomServerPathsTest {
```
lognaturel, line 25: Even though this is now a feature test, I think keeping the link to @getodk/testers' internal numbering (NODK-235) would be good so we can identify where those cases are being addressed. @getodk/testers?

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
```diff
@@ -76,23 +78,68 @@
...
 
-    /*
-    private ListPreference transportPreference;
-    private ExtendedPreferenceCategory smsPreferenceCategory;
-    */
+    private ListPopupWindow listPopupWindow;
+    private Preference selectedGoogleAccountPreference;
+    private boolean allowClickSelectedGoogleAccountPreference = true;
+
+    public static ServerPreferencesFragment newInstance(boolean adminMode) {
```
lognaturel, line 93: Might be good to have coverage for this admin mode behavior or at least flag it for QA.

collect_app/src/main/java/org/odk/collect/android/preferences/GeneralKeys.java
```diff
@@ -17,7 +17,7 @@
     public static final String KEY_USERNAME                 = "username";
     public static final String KEY_PASSWORD                 = "password";
 
-    // other_preferences.xml
+    // custom_server_paths_preferences.xmlreferences.xml
```
lognaturel, line 20: Remove `references.xml` suffix.

collect_app/src/main/java/org/odk/collect/android/preferences/GeneralKeys.java
```diff
@@ -132,7 +132,7 @@
         hashMap.put(KEY_GOOGLE_SHEETS_URL,          "");
         // identity_preferences.xml
         hashMap.put(KEY_ANALYTICS,                  true);
-        // other_preferences.xml
+        // custom_server_paths_preferenceshs_preferences.xml
```
lognaturel, line 136: custom_server_paths_preferences.xml

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
```diff
@@ -135,9 +136,6 @@ private void initProtocolPrefs() {
             addAggregatePreferences();
```
lognaturel: Rename?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ServerSettingsPage.java
```diff
@@ -23,12 +21,12 @@ public ServerSettingsPage assertOnPage() {
     }
 
     public ServerSettingsPage clickOnServerType() {
-        onData(PreferenceMatchers.withKey("protocol")).perform(click());
+        onView(withText(getTranslatedString(R.string.type))).perform(click());
         return this;
     }
 
     public ServerSettingsPage clickAggregateUsername() {
```
lognaturel, line 28: Rename?

collect_app/src/main/res/values/strings.xml
```diff
@@ -743,4 +743,5 @@
     <string name="preference_not_available">Not available</string>
     <string name="legacy_custom_server_paths">Legacy custom server paths</string>
     <string name="custom_server_paths">Custom server paths</string>
```
lognaturel, line 732: Maybe "legacy" doesn't add anything and we should use "Custom server paths" everywhere. The detail text about removal is clear enough, I think. I had proposed that wording but now that I've looked at translations for the .settings feature that also uses "legacy", I think it can be hard to translate to other languages.

collect_app/src/main/res/values/strings.xml
```diff
@@ -743,4 +743,5 @@
...
+    <string name="legacy_custom_server_paths_summary">Will be removed in a future version. Please use standard Open Rosa paths on your server.</string>
```
lognaturel: I don't think people know what "OpenRosa" is. Could we say "Please use /formList and /submission on your server"?

collect_app/src/main/res/values/strings.xml
```diff
@@ -174,7 +174,7 @@
     <string name="save_as_error">Saved name must not be blank!</string>
     <string name="edit_prompt">Edit Prompt</string>
     <string name="selected_google_account_text">Google account</string>
-    <string name="server_platform_odk_aggregate">ODK Aggregate</string>
+    <string name="server_platform_odk_aggregate">ODK</string>
```
lognaturel: I think this can now be untranslated? Or at least the resource name should change to `server_platform_odk` or `server_platform_default`.

lognaturel (changes requested):
Another really good example of first introducing an end-to-end test, then changing the behavior. 👍

There are method and resource names that I think should no longer include "Aggregate". This could happen in a follow-on PR but it would be good to do before the issue is closed so it isn't forgotten.

I also have made a couple of wording change suggestions. I feel strongly about not using "OpenRosa". I feel less strongly about using "legacy" or not and I'm happy for you to consider the suggestion and then decide.

Please see what suggestions you may want to do now, see if @grzesiek2010 wants to take a quick look and then mark for QA. @getodk/testers this has text changes that need to be translated and we'd like to get it in v1.27 so it's a good one to prioritize.

seadowg:
> There are method and resource names that I think should no longer include "Aggregate". This could happen in a follow-on PR but it would be good to do before the issue is closed so it isn't forgotten.

Yup agree. Will do those renames as part of the follow up refactor. This PR doesn't close out the issue!

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
seadowg: Great point. @getodk/testers for this PR it would be good to double check the server settings screen in admin mode!

collect_app/src/main/res/values/strings.xml
seadowg: Makes sense. I'll change it!

collect_app/src/main/res/values/strings.xml
seadowg: I thought it looked a little weird but I don't have strong feelings about it. I'll change it.

collect_app/src/main/res/values/strings.xml
lognaturel: This needs to be addressed before this PR is merged so translators get the right thing. Coming back to it, I have a preference for adding an ODK string to `untranslated.xml` for this instead of sending a translatable string.

lognaturel:
The "ODK" string does need to get in the right place before this is merged because merging pushes strings out to translators. This doesn't change behavior, though, so let's get QA started!

collect_app/src/main/res/values/strings.xml
seadowg: Ach my bad for dismissing. WIll change this!

mmarciniak90:
@seadowg take a look at this case

Use case:
1. User scans QR code where `Other` type was selected, e.g. credentials to KOBO server were set
2. Settings are applied and the user is able to use the app but when the user opens `Server` settings user sees view without data

The same, almost empty view is visible when `Other` server is set and apk is upgraded to the version with this fix

seadowg:
@mmarciniak90 ooooft good catch. This seems like a really obvious problem now you've pointed it out but myself and @lognaturel both missed it.  I'll look at this in the afternoon.

mmarciniak90:
Great! 
I noticed one more issue, this time more serious because it causes a crash.

1. Open dialog `Form list path` or `Submission path`
2. Rotate device 

kkrawczyk123:
@seadowg I had some random crashes on Server Settings view no matter what kind of server I have chosen. It happened on Android 10 and Android 5.1. Unfortunately, I am not able to specify reproduction steps but when investigating I have noticed that when:
- open URL dialog and click OK  (not even change anything)
and 
- open Google account dialog and click ok 
There always displays error message in logcat: 
`2020-05-12 14:42:22.101 2409-18089/? E/SQLiteDatabase: Error inserting flex_time=2792000 job_id=-1 period=5585000 source=16 requires_charging=0 preferred_network_type=1 target_class=com.google.android.gms.measurement.PackageMeasurementTaskService user_id=0 target_package=com.google.android.gms tag=Measurement.PackageMeasurementTaskService.UPLOAD_TASK_TAG task_type=0 required_idleness_state=0 service_kind=0 source_version=201516000 persistence_level=1 preferred_charging_state=1 required_network_type=0 runtime=1589287342082 retry_strategy={"maximum_backoff_seconds":{"3600":0},"initial_backoff_seconds":{"30":0},"retry_policy":{"0":0}} last_runtime=0
    android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: pending_ops.tag, pending_ops.target_class, pending_ops.target_package, pending_ops.user_id (code 2067 SQLITE_CONSTRAINT_UNIQUE)
        at android.database.sqlite.SQLiteConnection.nativeExecuteForLastInsertedRowId(Native Method)
        at android.database.sqlite.SQLiteConnection.executeForLastInsertedRowId(SQLiteConnection.java:879)
        at android.database.sqlite.SQLiteSession.executeForLastInsertedRowId(SQLiteSession.java:790)
        at android.database.sqlite.SQLiteStatement.executeInsert(SQLiteStatement.java:88)
        at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:1599)
        at android.database.sqlite.SQLiteDatabase.insert(SQLiteDatabase.java:1468)
        at aplm.a(:com.google.android.gms@201516037@20.15.16 (120400-309763488):76)
        at aplb.a(:com.google.android.gms@201516037@20.15.16 (120400-309763488):173)
        at aplb.a(:com.google.android.gms@201516037@20.15.16 (120400-309763488):21)
        at aplb.a(:com.google.android.gms@201516037@20.15.16 (120400-309763488):167)
        at aphk.run(:com.google.android.gms@201516037@20.15.16 (120400-309763488):8)
        at sob.b(:com.google.android.gms@201516037@20.15.16 (120400-309763488):12)
        at sob.run(:com.google.android.gms@201516037@20.15.16 (120400-309763488):7)
        at sub.run(:com.google.android.gms@201516037@20.15.16 (120400-309763488):0)
        at java.lang.Thread.run(Thread.java:919)`

I am not sure if that error is or might be connected to those random crashes so i think it would be good if you could take a look.

seadowg:
@mmarciniak90 @lognaturel we could put in some hacking fixes for the QR code scanning problem but really I think it needs a new shared prefs migration (which means we should work on top of #3813) and also needs migrations to work when importing configuration (when means fixing #3811). I think it's best to pause on this (pushing it to 1.28) and look at those. Does that sound right?

seadowg:
After talking it over on Slack we're going to push this to 1.28

seadowg:
@mmarciniak90  Ok I've fixed the issues you found. The random crashes @kkrawczyk123 was seeing may well have been related to the rotation problem so good to check on that again. 

When importing settings from a QR code that uses the "Other" server type threre will still be some weirdness in that the "Type" field will look empty (but the ODK type preferences will be displayed). Fixing #3811 will fix this. If we can live with that weirdness we can potentially look at reincluding this in the 1.27 release but I personally think that it'd be better to hold off. Either way I think we could merge this before fixing #3811.

collect_app/src/main/java/org/odk/collect/android/application/initialization/migration/MigrationUtils.java
```diff
@@ -0,0 +1,28 @@
+package org.odk.collect.android.application.initialization.migration;
```
lognaturel, line 1: This and `ValueTranslator` feel somewhat disjoint from the migration tools in `org.odk.collect.android.preferences`. It would be good to consolidate them in your follow-up.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/application/initialization/migration/MigrationUtils.java
seadowg: @lognaturel they fit the structure that #3813 introduces so should fit in when these are merged 😀

seadowg:
Went for a merge rather than a rebase on the conflicts as rebasing with strings generally leads to me having the worst time.

collect_app/src/main/java/org/odk/collect/android/application/initialization/migration/MigrationUtils.java
lognaturel: Oh great, I only went for a skim on that one since it was already reviewed!

kkrawczyk123:
@seadowg, I have confirmed that previously defined issues have been fixed.
I haven't also noticed any random crashes what is great but I have notices some problems:
- I was able to multiclick on custom server paths like on the gif:
multiclick

- When rotate screen after opening custom server paths the text on the header disappears.
That header also disappears in the background when rotating screen when form list path and submission path dialogs are opened.

- Custom server paths header disappears also after minimizing an app when don't keep activities is enabled.

- After input text to both: /formList and /submission (in /formList/bla way) it is reset when going back General Settings (back button twice) - on master it is not 

seadowg:
@kkrawczyk123 the second and third issue where the title disappears after rotating/minimizing actually already exists for the User Interface and Form Metadata screen. I'll post an issue (#3901) and fix as part of the follow up work to this.

I've fixed the multiclick and the server paths getting reset issues!

kkrawczyk123:
**Testes with success!**

Tested scenarios:
- light/dark mode
- other removed from server settings
- use servers: Aggregate 2.0, local Aggregate, Aggregate 1.7, Central, Ona, Kobo, Google Drive
- follow: download, fill form, send submission for all
- import server settings
- remember url when settings imported
- import qr code when other selected on older version
- multiclick
- screen rotation, minimizing
- language change 
- edit and save form list path and submission path 
- reset settings
- server settings in admin settings
- form with submission url 

Issue #3815, opened 2020-05-07
Improve initializing map providers


grzesiek2010:
#### Problem description
`GoogleMaps` and `Mapbox` require initializing online even if a user needs to use them offline. In https://github.com/getodk/collect/pull/3804 I added a small improvement that performs the initialization when the app is installed. There is a good chance a user is online during installing the app but we can't be sure so we should handle that in a better way:
[scheduling work to run when we get an internet connection](https://github.com/getodk/collect/pull/3804#issuecomment-621717349) or [triggering it after the first successful form list download](https://github.com/getodk/collect/pull/3804#pullrequestreview-407055835)
Additionally I think it would be good to provide some explanation for users when initialization hasn't been performed. It could be a banner or a snackbar in map settings and/or in map view.

PR #3819, opened 2020-05-08
Make sure MapBox init fails silently

seadowg:
This stops the app crashing on launch in the emulator (and on other devices where Mapbox native code isn't included).

Ran tests and launched app.

We should double check that this hasn't caused any regressions on the fixes #3804 introduced. I think if the reviewer could double check that we won't need QA for this.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
@@ -231,7 +231,13 @@ public class SplashScreenActivity extends Activity {
     }
 
     private void initMapBox() {
-        MapView mapView = findViewById(R.id.mapView);
-        mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
+        // This "one weird trick" lets us initialize MapBox at app start when the internet is
+        // most likely to be available. This is annoyingly needed for offline tiles to work.
+        try {
+            MapView mapView = getLayoutInflater().inflate(R.layout.mapbox_init, null).findViewById(R.id.mapView);
+            mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
+        } catch (Exception ignored) {
+            // This will crash on devices where the arch for MapBox is not included
+        }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
@@ -29,9 +29,6 @@ import android.view.Window;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 
-import com.mapbox.mapboxsdk.maps.MapView;
-import com.mapbox.mapboxsdk.maps.Style;
-
 import org.odk.collect.android.R;
 import org.odk.collect.android.analytics.Analytics;
 import org.odk.collect.android.application.Collect;
@@ -86,7 +83,6 @@ public class SplashScreenActivity extends Activity {
                 }
 
                 init();
-                initMapBox();
             }
 
             @Override
@@ -228,16 +224,5 @@ public class SplashScreenActivity extends Activity {
             }
         };
         t.start();
-    }
-
-    private void initMapBox() {
-        // This "one weird trick" lets us initialize MapBox at app start when the internet is
-        // most likely to be available. This is annoyingly needed for offline tiles to work.
-        try {
-            MapView mapView = getLayoutInflater().inflate(R.layout.mapbox_init, null).findViewById(R.id.mapView);
-            mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
-        } catch (Exception | Error ignored) {
-            // This will crash on devices where the arch for MapBox is not included
-        }
     }
 }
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -231,7 +231,13 @@ public void run() {
...
 
     private void initMapBox() {
-        MapView mapView = findViewById(R.id.mapView);
-        mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
+        // This "one weird trick" lets us initialize MapBox at app start when the internet is
+        // most likely to be available. This is annoyingly needed for offline tiles to work.
+        try {
+            MapView mapView = getLayoutInflater().inflate(R.layout.mapbox_init, null).findViewById(R.id.mapView);
+            mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
+        } catch (Exception ignored) {
```
grzesiek2010: I would catch Error here as well it might be needed in some cases like https://github.com/getodk/collect/commit/f6bc94ebef3d901734b272d8e5031abd90245bd0

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
seadowg: Good call! We want to account for everything.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -231,7 +231,13 @@ public void run() {
     }
 
     private void initMapBox() {
-        MapView mapView = findViewById(R.id.mapView);
-        mapView.getMapAsync(mapBoxMap -> mapBoxMap.setStyle(Style.MAPBOX_STREETS, style -> { }));
+        // This "one weird trick" lets us initialize MapBox at app start when the internet is
+        // most likely to be available. This is annoyingly needed for offline tiles to work.
+        try {
+            MapView mapView = getLayoutInflater().inflate(R.layout.mapbox_init, null).findViewById(R.id.mapView);
```
grzesiek2010: Are you sure this still does the job? I mean the map is not displayed anywhere but it might be required.

seadowg: Not had a chance to double check yet. I could see that being a problem.

seadowg: @grzesiek2010 if you have basemaps already ready to go could you check? I don't have anything setup for any of this.

grzesiek2010: Ok will do.

seadowg: Thanks!

grzesiek2010: Yeah as I thought it doesn't work. It needs to be displayed.

seadowg: Ach really? That's weird. I wonder what part of the View's lifecycle needs to get hit. Definitely possible to inflate and then add this into the view hierarchy though. I'm not going to be able to look at this again tonight though. @lognaturel if this is getting in your way (i.e. you need to look at the app on an emulator) I'd suggest merging this or reverting #3804. Also feel free to push to this if you have a version that retains the fixes from #3804!

grzesiek2010: I'll take a look.

grzesiek2010: Ok it's done. I moved the code from the SplashScreen to the MainMenuActivity because once we merge https://github.com/getodk/collect/pull/3813 the SplashScreen won't be displayed by default so the fix also won't work.

lognaturel:
I almost always use a physical device myself so wouldn't have caught this for a long time. Good to add a comment, too.

What are the implications of this running every single time the main activity is created? I imagine there's no performance issue since it's async. Does it actually attempt a network request every time or will it only do so if it hasn't "phoned home" yet?

grzesiek2010:
> What are the implications of this running every single time the main activity is created? I imagine there's no performance issue since it's async.

I think it shouldn't affect performance at all.

Does it actually attempt a network request every time or will it only do so if it hasn't "phoned home" yet?

I don't know I would need to check but since it's a temporary fix which will be improved in https://github.com/getodk/collect/issues/3815 I think there is no need to be worried about it.

seadowg:
@lognaturel @grzesiek2010 testing through a proxy I'm not seeing any connections from Mapbox until I actually load a map. I'm guessing these tricks just make Mapbox think it's online but for some reason it doesn't actually need to connect (the `Mapbox.setConnected(true);` makes a lot of sense given this context). I'm going to merge this for the moment to get `master` back in a working state.

Edit: Oh wait of course I can't it's my PR. @grzesiek2010 could you merge?

grzesiek2010 (approved):

grzesiek2010:
Ok I'm merging but also adding `needs-testing` label. It's low-priority so @getodk/testers please do verify it when you are free.

lognaturel:
> since it's a temporary fix which will be improved in #3815 I think there is no need to be worried about it.

I mostly agree but if this were making significant network calls every time the main screen were launched we might not want to leave this as-is for a release. Sounds like that's not the case which is great. I'd also like to document the current behavior since it will be there for one release at least. Is the following accurate?

> To use an offline layer with Google Maps, the device must have been online on first launch after installation.

As far as we know, OSM and Mapbox offline layers will now always work, right?

mmarciniak90:
Tested as a part of regression testing

PR #3831, opened 2020-05-14
Rewriting URL widget

SaumiaSinghal:
As part of the project 'Rewriting the widget framework ' described on ODK forum [here](https://forum.getodk.org/t/gsoc-rewrite-odk-collects-widgets-using-the-new-widget-framework/25328). This PR includes the new framework for the URL widget.

No

All Widgets form, or any form with the URL widget.

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
@@ -24,7 +24,7 @@ import java.util.List;
 public class CustomTabHelper {
     public static final String OPEN_URL = "url";
     private static final String CUSTOM_TAB_PACKAGE_NAME = "com.android.chrome";
-    public CustomTabsClient customTabsClient;
+    private CustomTabsClient customTabsClient;
     private CustomTabsSession customTabsSession;
 
     /*
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
@@ -24,7 +24,7 @@ import java.util.List;
 public class CustomTabHelper {
     public static final String OPEN_URL = "url";
     private static final String CUSTOM_TAB_PACKAGE_NAME = "com.android.chrome";
-    private CustomTabsClient customTabsClient;
+    public CustomTabsClient customTabsClient;
     private CustomTabsSession customTabsSession;
 
     /*
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -23,7 +23,6 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.StringData;
@@ -32,6 +31,7 @@ import org.odk.collect.android.R;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.CustomTabHelper;
 import org.odk.collect.android.utilities.MultiClickGuard;
+import org.odk.collect.android.utilities.ToastUtils;
 
 /**
  * Widget that allows user to open URLs from within the form
@@ -67,25 +67,14 @@ public class UrlWidget extends QuestionWidget {
 
         stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
         String answerText = prompt.getAnswerText();
-        if (answerText != null) {
-            stringAnswer.setText(answerText);
-        }
+        stringAnswer.setText(answerText);
 
         return answerView;
     }
 
-    private boolean isUrlEmpty(TextView stringAnswer) {
-        return stringAnswer == null || stringAnswer.getText() == null
-                || stringAnswer.getText().toString().isEmpty();
-    }
-
-    private Uri getUri() {
-        return Uri.parse(stringAnswer.getText().toString());
-    }
-
     @Override
     public void clearAnswer() {
-        Toast.makeText(getContext(), "URL is readonly", Toast.LENGTH_SHORT).show();
+        ToastUtils.showShortToast("URL is readonly");
     }
 
     @Override
@@ -108,11 +97,21 @@ public class UrlWidget extends QuestionWidget {
         stringAnswer.cancelLongPress();
     }
 
+    @Override
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
         if (customTabHelper.getServiceConnection() != null) {
             getContext().unbindService(customTabHelper.getServiceConnection());
         }
+    }
+
+    private boolean isUrlEmpty(TextView stringAnswer) {
+        return stringAnswer == null || stringAnswer.getText() == null
+                || stringAnswer.getText().toString().isEmpty();
+    }
+
+    private Uri getUri() {
+        return Uri.parse(stringAnswer.getText().toString());
     }
 
     public void onButtonClick() {
@@ -121,7 +120,7 @@ public class UrlWidget extends QuestionWidget {
                 customTabHelper.bindCustomTabsService(getContext(), null);
                 customTabHelper.openUri(getContext(), getUri());
             } else {
-                Toast.makeText(getContext(), "No URL set", Toast.LENGTH_SHORT).show();
+                ToastUtils.showShortToast("No URL set");
             }
         }
     }
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -17,21 +17,23 @@ package org.odk.collect.android.widgets;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.net.Uri;
+import android.util.TypedValue;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
 import android.widget.Button;
-import android.widget.LinearLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.StringData;
+import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
-import org.odk.collect.android.formentry.questions.WidgetViewUtils;
 import org.odk.collect.android.utilities.CustomTabHelper;
+import org.odk.collect.android.utilities.MultiClickGuard;
+import org.odk.collect.android.utilities.ThemeUtils;
 import org.odk.collect.android.widgets.interfaces.ButtonWidget;
-
-import static org.odk.collect.android.formentry.questions.WidgetViewUtils.createSimpleButton;
-import static org.odk.collect.android.formentry.questions.WidgetViewUtils.getCenteredAnswerTextView;
 
 /**
  * Widget that allows user to open URLs from within the form
@@ -42,31 +44,42 @@ import static org.odk.collect.android.formentry.questions.WidgetViewUtils.getCen
 public class UrlWidget extends QuestionWidget implements ButtonWidget {
 
     private Uri uri;
-    final Button openUrlButton;
-    final TextView stringAnswer;
+    private Button openUrlButton;
+    private TextView stringAnswer;
     private final CustomTabHelper customTabHelper;
 
     public UrlWidget(Context context, QuestionDetails questionDetails) {
         super(context, questionDetails);
+        customTabHelper = new CustomTabHelper();
+    }
 
-        openUrlButton = createSimpleButton(getContext(), getFormEntryPrompt().isReadOnly(), context.getString(R.string.open_url), getAnswerFontSize(), this);
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        ViewGroup answerView = (ViewGroup) LayoutInflater.from(context).inflate(R.layout.url_widget_answer, null);
 
-        stringAnswer = getCenteredAnswerTextView(getContext(), getAnswerFontSize());
+        openUrlButton = answerView.findViewById(R.id.url_button);
+        stringAnswer = answerView.findViewById(R.id.url_answer_text);
 
-        String s = questionDetails.getPrompt().getAnswerText();
+        if (prompt.isReadOnly()) {
+            openUrlButton.setVisibility(GONE);
+        } else {
+            openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            openUrlButton.setOnClickListener(v -> {
+                if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
+                    onButtonClick(openUrlButton.getId());
+                }
+            });
+        }
+
+        stringAnswer.setTextColor(new ThemeUtils(context).getColorOnSurface());
+        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        String s = prompt.getAnswerText();
         if (s != null) {
             stringAnswer.setText(s);
             uri = Uri.parse(stringAnswer.getText().toString());
         }
 
-        // finish complex layout
-        LinearLayout answerLayout = new LinearLayout(getContext());
-        answerLayout.setOrientation(LinearLayout.VERTICAL);
-        answerLayout.addView(openUrlButton);
-        answerLayout.addView(stringAnswer);
-        addAnswerView(answerLayout, WidgetViewUtils.getStandardMargin(context));
-
-        customTabHelper = new CustomTabHelper();
+        return answerView;
     }
 
     private boolean isUrlEmpty(TextView stringAnswer) {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -21,7 +21,6 @@ import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.Button;
 import android.widget.TextView;
 
 import org.javarosa.core.model.data.IAnswerData;
@@ -30,8 +29,8 @@ import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.CustomTabHelper;
-import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.utilities.ToastUtils;
+import org.odk.collect.android.views.MultiClickSafeButton;
 
 /**
  * Widget that allows user to open URLs from within the form
@@ -43,7 +42,7 @@ public class UrlWidget extends QuestionWidget {
 
     private final CustomTabHelper customTabHelper;
 
-    protected Button openUrlButton;
+    protected MultiClickSafeButton openUrlButton;
     protected TextView stringAnswer;
 
     public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper) {
@@ -105,13 +104,11 @@ public class UrlWidget extends QuestionWidget {
     }
 
     public void onButtonClick() {
-        if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
-            if (!isUrlEmpty(stringAnswer)) {
-                customTabHelper.bindCustomTabsService(getContext(), null);
-                customTabHelper.openUri(getContext(), getUri());
-            } else {
-                ToastUtils.showShortToast("No URL set");
-            }
+        if (!isUrlEmpty(stringAnswer)) {
+            customTabHelper.bindCustomTabsService(getContext(), null);
+            customTabHelper.openUri(getContext(), getUri());
+        } else {
+            ToastUtils.showShortToast("No URL set");
         }
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -42,10 +42,11 @@ import org.odk.collect.android.widgets.interfaces.ButtonWidget;
 @SuppressLint("ViewConstructor")
 public class UrlWidget extends QuestionWidget implements ButtonWidget {
 
-    private Uri uri;
-    private Button openUrlButton;
-    private TextView stringAnswer;
     private final CustomTabHelper customTabHelper;
+    private Uri uri;
+
+    protected Button openUrlButton;
+    protected TextView stringAnswer;
 
     public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper) {
         super(context, questionDetails);
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -15,22 +15,20 @@
 package org.odk.collect.android.widgets;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
 import android.content.Context;
 import android.net.Uri;
 import android.util.TypedValue;
-import android.view.LayoutInflater;
 import android.view.View;
-import android.view.ViewGroup;
 import android.widget.TextView;
 
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.StringData;
 import org.javarosa.form.api.FormEntryPrompt;
-import org.odk.collect.android.R;
+import org.odk.collect.android.databinding.UrlWidgetAnswerBinding;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.CustomTabHelper;
 import org.odk.collect.android.utilities.ToastUtils;
-import org.odk.collect.android.views.MultiClickSafeButton;
 
 /**
  * Widget that allows user to open URLs from within the form
@@ -41,9 +39,7 @@ import org.odk.collect.android.views.MultiClickSafeButton;
 public class UrlWidget extends QuestionWidget {
 
     private final CustomTabHelper customTabHelper;
-
-    protected MultiClickSafeButton openUrlButton;
-    protected TextView stringAnswer;
+    private UrlWidgetAnswerBinding binding;
 
     public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper) {
         super(context, questionDetails);
@@ -52,20 +48,18 @@ public class UrlWidget extends QuestionWidget {
 
     @Override
     protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
-        ViewGroup answerView = (ViewGroup) LayoutInflater.from(context).inflate(R.layout.url_widget_answer, null);
-
-        openUrlButton = answerView.findViewById(R.id.url_button);
-        stringAnswer = answerView.findViewById(R.id.url_answer_text);
+        binding = UrlWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        View answerView = binding.getRoot();
 
         if (prompt.isReadOnly()) {
-            openUrlButton.setVisibility(GONE);
+            binding.urlButton.setVisibility(GONE);
         } else {
-            openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
-            openUrlButton.setOnClickListener(v -> onButtonClick());
+            binding.urlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            binding.urlButton.setOnClickListener(v -> onButtonClick());
         }
 
-        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
-        stringAnswer.setText(prompt.getAnswerText());
+        binding.urlAnswerText.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        binding.urlAnswerText.setText(prompt.getAnswerText());
 
         return answerView;
     }
@@ -77,7 +71,7 @@ public class UrlWidget extends QuestionWidget {
 
     @Override
     public IAnswerData getAnswer() {
-        String answerText = stringAnswer.getText().toString();
+        String answerText = binding.urlAnswerText.getText().toString();
         return !answerText.isEmpty()
                 ? new StringData(answerText)
                 : null;
@@ -85,14 +79,14 @@ public class UrlWidget extends QuestionWidget {
 
     @Override
     public void setOnLongClickListener(OnLongClickListener l) {
-        openUrlButton.setOnLongClickListener(l);
+        binding.urlButton.setOnLongClickListener(l);
     }
 
     @Override
     public void cancelLongPress() {
         super.cancelLongPress();
-        openUrlButton.cancelLongPress();
-        stringAnswer.cancelLongPress();
+        binding.urlButton.cancelLongPress();
+        binding.urlAnswerText.cancelLongPress();
     }
 
     @Override
@@ -104,12 +98,16 @@ public class UrlWidget extends QuestionWidget {
     }
 
     public void onButtonClick() {
-        if (!isUrlEmpty(stringAnswer)) {
+        if (!isUrlEmpty(binding.urlAnswerText)) {
             customTabHelper.bindCustomTabsService(getContext(), null);
             customTabHelper.openUri(getContext(), getUri());
         } else {
             ToastUtils.showShortToast("No URL set");
         }
+    }
+
+    protected UrlWidgetAnswerBinding getBinding() {
+        return binding;
     }
 
     private boolean isUrlEmpty(TextView stringAnswer) {
@@ -118,6 +116,6 @@ public class UrlWidget extends QuestionWidget {
     }
 
     private Uri getUri() {
-        return Uri.parse(stringAnswer.getText().toString());
+        return Uri.parse(binding.urlAnswerText.getText().toString());
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -48,9 +48,9 @@ public class UrlWidget extends QuestionWidget implements ButtonWidget {
     private TextView stringAnswer;
     private final CustomTabHelper customTabHelper;
 
-    public UrlWidget(Context context, QuestionDetails questionDetails) {
+    public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper) {
         super(context, questionDetails);
-        customTabHelper = new CustomTabHelper();
+        this.customTabHelper = customTabHelper;
     }
 
     @Override
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -32,7 +32,6 @@ import org.odk.collect.android.R;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.CustomTabHelper;
 import org.odk.collect.android.utilities.MultiClickGuard;
-import org.odk.collect.android.widgets.interfaces.ButtonWidget;
 
 /**
  * Widget that allows user to open URLs from within the form
@@ -40,7 +39,7 @@ import org.odk.collect.android.widgets.interfaces.ButtonWidget;
  * @author Yaw Anokwa (yanokwa@gmail.com)
  */
 @SuppressLint("ViewConstructor")
-public class UrlWidget extends QuestionWidget implements ButtonWidget {
+public class UrlWidget extends QuestionWidget {
 
     private final CustomTabHelper customTabHelper;
     private Uri uri;
@@ -65,9 +64,7 @@ public class UrlWidget extends QuestionWidget implements ButtonWidget {
         } else {
             openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
             openUrlButton.setOnClickListener(v -> {
-                if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
-                    onButtonClick(openUrlButton.getId());
-                }
+                onButtonClick();
             });
         }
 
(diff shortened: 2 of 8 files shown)


seadowg:
@SaumiaSinghal just making this a draft as discussed!

SaumiaSinghal:
Oh sorry! somehow created this as a normal PR.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -22,14 +33,74 @@
...
 
+    @Test
+    public void clearAnswer_doesNotClearWidgetAnswer() {
+        UrlWidget widget = createWidget(promptWithAnswer(new StringData("blah")));
+        widget.clearAnswer();
+        assertThat(widget.getAnswer().getDisplayText(), equalTo("blah"));
+    }
+
+    @Test
+    public void whenPromptHasAnswer_setsCorrectAnswer() {
```
seadowg: I would rename this `whenPromptHasAnswer_displaysAnswer` and the below one to `whenPromptAnswerDoesNotHaveAnswer_displayEmptyString`.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -22,14 +33,74 @@
...
+        UrlWidget widget = createWidget(promptWithAnswer(null));
+        Button urlButton = widget.findViewById(R.id.url_button);
+        urlButton.performClick();
+
+        verify(customTabHelper, never()).bindCustomTabsService(null, null);
+        verify(customTabHelper, never()).openUri(null, null);
+    }
+
+    @Test
+    public void clickingButtonWhenUrlIsNotEmpty_callsCorrectMethods() {
```
seadowg: This can just be `clickingButton...` as it's the "normal" case!

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -22,14 +33,74 @@
...
+        UrlWidget widget = createWidget(promptWithAnswer(new StringData("blah")));
+        Button urlButton = widget.findViewById(R.id.url_button);
+        TextView textView = widget.findViewById(R.id.url_answer_text);
+        urlButton.performClick();
+
+        assertThat(textView.getText().toString(), equalTo("blah"));
+        verify(customTabHelper, times(1)).bindCustomTabsService(widget.getContext(), null);
```
seadowg: I'd not bother with `times` here. I'd usually only use it when something happens more than once but I expect a dependency to only get called once in that case (maybe when testing caching or something).

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -48,9 +48,9 @@
     private TextView stringAnswer;
     private final CustomTabHelper customTabHelper;
 
-    public UrlWidget(Context context, QuestionDetails questionDetails) {
+    public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper) {
```
seadowg, line 39: Nice change!

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -22,14 +33,74 @@
...
+
+    @Test
+    public void whenPromptAnswerDoesNotHaveAnswer_setsNullValue() {
+        UrlWidget widget = createWidget(promptWithAnswer(null));
+        assertThat(((TextView) widget.findViewById(R.id.url_answer_text)).getText().toString(), equalTo(""));
+    }
+
+
+    @Test
+    public void clickingButtonWhenUrlIsEmpty_doesNotCallOpenUri() {
```
seadowg, line 86: Good catch!

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -42,31 +44,42 @@
...
+        } else {
+            openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            openUrlButton.setOnClickListener(v -> {
+                if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
+                    onButtonClick(openUrlButton.getId());
+                }
+            });
+        }
+
+        stringAnswer.setTextColor(new ThemeUtils(context).getColorOnSurface());
```
seadowg: This can move into the XML layout.

seadowg:
This is a great start @SaumiaSinghal. I've left some comments inline.

Next I think you should look at adding tests for the `setOnLongClickListener` and `cancelLongPress` and also `onDetachedFromWindow` as this looks like it deals with some important lifecycle stuff.

collect_app/src/main/res/layout/url_widget_answer.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:layout_marginTop="5dp"
+        android:layout_marginBottom="5dp"
+        />
+
+    <com.google.android.material.textview.MaterialTextView
+        android:id="@+id/url_answer_text"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center"
+        android:paddingStart="20dp"
```
seadowg: We should try as much as possible to use the `@dimen` margins so they're consistent in throughout the app. Here you can probably use `@dimen/margin_standard` for the padding as that is the normal left/right spacing (16dp).

collect_app/src/main/res/layout/url_widget_answer.xml
```diff
@@ -0,0 +1,29 @@
...
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/url_button"
+        style="@style/Widget.Collect.Button.Custom"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:padding="@dimen/margin_small"
+        android:text="@string/open_url"
+        android:layout_marginStart="@dimen/margin_standard"
+        android:layout_marginEnd="@dimen/margin_standard"
+        android:layout_marginTop="5dp"
```
seadowg: Lets maybe use `@dimen/margin_extra_small` for this

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
SaumiaSinghal: Oh okay thanks. I will keep it in mind.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: Thank you!

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -1,46 +1,123 @@
...
+        UrlWidget widget = createWidget(promptWithAnswer(new StringData("blah")));
+        Button urlButton = widget.findViewById(R.id.url_button);
+        urlButton.performClick();
+
+        verify(customTabHelper).bindCustomTabsService(widget.getContext(), null);
+        verify(customTabHelper).openUri(widget.getContext(), Uri.parse("blah"));
+    }
+
+    @Test
+    public void onLongPressButton_doesNotCallOpenUri() {
```
seadowg: I don't think this or the test below are correct. I think my question around writing tests for `setOnLongClickListener` and `cancelLongPress` was confusing though now I look at it. Lets breakdown what we need for each method:

* `setOnLongClickListener`: usually this needs to call `setOnLongClickListener` on any child views in the widget so that when I long press on anything in the widget I see the "Edit Prompt" pop up menu. Here you'll see that `setOnLongClickListener` is empty so that long pressing on the URL button won't do anything - it's clickable so it consumes the long press event. We need to write a test that makes sure long pressing on the url button will call any listener passed to the widget's `setOnLongClickListener`.
* `cancelLongPress`: this makes sure that any long press cancellation (usually the user scrolling around) works properly. I don't think there is a way to test that, so I think we should just make sure to review this method and make sure any views in the widget have the call propagated to them.

seadowg: One tip by the way for checking the coverage you have for an object is to use Android Studio's "Run with coverage" feature. If you use the green ▶️ icon to run the test file (in this case `URLWidgetTest`) and then hit "Run X with Coverage" you'll see green and red highligting in the `URLWidget` file for what lines are covered by tests and what ones aren't.

SaumiaSinghal: Thank you so much for the suggestion. I will modify the tests.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -98,26 +106,23 @@ public void clickingButton_callsCorrectMethods() {
...
-        assertThat(textView.hasFocus(), equalTo(false));
+    private TestUrlWidget createWidget(FormEntryPrompt prompt) {
+        return new TestUrlWidget(widgetTestActivity(), new QuestionDetails(prompt, "formAnalyticsID"), customTabHelper, listener);
     }
 
+    private static class TestUrlWidget extends UrlWidget {
 
-    private UrlWidget createWidget(FormEntryPrompt prompt) {
-        return new UrlWidget(widgetTestActivity(), new QuestionDetails(prompt, "formAnalyticsID"), customTabHelper);
+        public TestUrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelper customTabHelper, OnLongClickListener listener) {
```
seadowg: Here's you're extending URLWidget for the test and then using that. That means your test is now of TestURLWidget not URLWidget. We shouldn't use fakes, mocks or stubs of the objects we're testing as this means we're not really testing the object itself!

I'm not sure why you created this as you could just call `setOnLongClickListener` on the widget in the test.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -98,26 +106,23 @@ public void clickingButton_callsCorrectMethods() {
...
     @Test
-    public void onLongPressButton_doesNotCallOpenUri() {
+    public void clickingButtonForLong_callsLongClickListener() {
         UrlWidget widget = createWidget(promptWithAnswer(new StringData("blah")));
         Button urlButton = widget.findViewById(R.id.url_button);
         urlButton.performLongClick();
 
-        verify(customTabHelper, never()).bindCustomTabsService(widget.getContext(), null);
-        verify(customTabHelper, never()).openUri(widget.getContext(), Uri.parse("blah"));
+        assertThat(listener.onLongClick(urlButton), equalTo(true));
```
seadowg: I think a better way to test this is to pass a `mock(Listener.class)` to the Widget's `OnLongClickListener` and then check that it is called (using `when`) after you call `urlButton.performLongClick()`. Does that make sense? That would make sure the listener the URL button has and the listener the widget has are definitely the same

collect_app/src/main/res/layout/url_widget_answer.xml
```diff
@@ -0,0 +1,30 @@
...
+        android:id="@+id/url_button"
+        style="@style/Widget.Collect.Button.Custom"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:padding="@dimen/margin_small"
+        android:text="@string/open_url"
+        android:layout_marginStart="@dimen/margin_standard"
+        android:layout_marginEnd="@dimen/margin_standard"
+        android:layout_marginTop="@dimen/margin_extra_small"
+        android:layout_marginBottom="@dimen/margin_extra_small"
```
seadowg: Just a tiny change while we're here. Often it's nicer to have the margin between two items be on the second item (the `MaterialTextView`) rather than the first (the `MaterialButton`) here. This often helps when the first/last items have different margins and also makes it easier when deleting items. I'd probably move the `marginBottom` here so it's a `marginTop` on the `MaterialTextView`.

SaumiaSinghal: Ohh! Thanks for the advice. I will remember that.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
seadowg: This is much better! But sorry I lead you down a slightly wrong path there. I think instead of using `when` to set up a result and then verifying it later (lines 108 and 111) you could just do `verify(listener).onLongClick(urlButton)`. Does that make more sense?

SaumiaSinghal: I wrote the test like this initially:


```
        UrlWidget widget = createWidget(promptWithAnswer(new StringData("blah")));
        Button urlButton = widget.findViewById(R.id.url_button);
        widget.setOnLongClickListener(listener);
        urlButton.performLongClick();
        verify(listener).onLongClick(urlButton);
```

But this test does not pass.

seadowg: Right but I think that's a correctly failing test: `URLWidget#setOnLongClickListener` doesn't do anything at the moment so I'd expect that test to be red. If you try long clicking on the URL button in the app at the moment right now you'll see it doesn't show the "Edit Prompt" pop up.

SaumiaSinghal: It says `Wanted but not invoked` and `Actually, there were zero interactions with this mock.` I do not understand.

seadowg: Right that means that the `onLongClick` on the listener was never called. It might be worth reading some docs at https://site.mockito.org/ if you're confused by this message. If you look at `URLWidget#setOnLongClickListener` you'll see it's empty so that test should fail in the way you're seeing. You have a red test that you can now make green!

SaumiaSinghal: Hi! I'm sorry. I didn't refresh the page and so kept adding comments without reading your second last comment. I understand it now.

seadowg: @SaumiaSinghal hah I've made exactly the same mistake before!

SaumiaSinghal: :disappointed_relieved:

collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
```diff
@@ -24,7 +24,7 @@
 public class CustomTabHelper {
     public static final String OPEN_URL = "url";
     private static final String CUSTOM_TAB_PACKAGE_NAME = "com.android.chrome";
-    private CustomTabsClient customTabsClient;
+    public CustomTabsClient customTabsClient;
```
seadowg: Is this an accidental change? I can't see a reason for it.

seadowg (dismissed):

collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
SaumiaSinghal: Oh Sorry, I was trying something in `onDetachedFromWindow` test.

seadowg:
This is looking good to me! If you're feeling happy with it then hit "Ready for Review". I think it'd be good for @lognaturel and @grzesiek2010 to take a quick look as well seeing as this is the first of our widget rewrite PRs.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -1,46 +1,146 @@
...
-        when(formEntryPrompt.isReadOnly()).thenReturn(true);
+    public void detachingFromWindow_disconnectsService_whenServiceConnectionIsNotNull() {
+        when(customTabHelper.getServiceConnection()).thenReturn(serviceConnection);
+        UrlWidget widget = createWidget(promptWithAnswer(null));
+        widget.onDetachedFromWindow();
+        verify(serviceConnection).onServiceDisconnected(null);
+    }
+
+    @Test
+    public void detachingFromWindow_doesNotCallOnServiceDisconnected_whenServiceConnectionIstNull() {
```
lognaturel: Extra 't' in `whenServiceConnectionIstNull`

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -1,46 +1,146 @@
...
+        widget.cancelLongPress();
+
+        verify(widget.openUrlButton).cancelLongPress();
+        verify(widget.stringAnswer).cancelLongPress();
     }
 
     @Test
-    public void usingReadOnlyOptionShouldMakeAllClickableElementsDisabled() {
-        when(formEntryPrompt.isReadOnly()).thenReturn(true);
+    public void detachingFromWindow_disconnectsService_whenServiceConnectionIsNotNull() {
```
lognaturel, line 122: This test and the following seem to me like they should be part of `CustomTabHelper`'s tests, not the tests for a specific widget.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -1,46 +1,146 @@
...
+        UrlWidget widget = createWidget(promptWithAnswer(null));
+        Button urlButton = widget.findViewById(R.id.url_button);
+        urlButton.performClick();
+
+        verify(customTabHelper, never()).bindCustomTabsService(null, null);
+        verify(customTabHelper, never()).openUri(null, null);
+    }
+
+    @Test
+    public void clickingButton_callsCorrectMethods() {
```
lognaturel: I think this could be `_opensUri` instead of `_callsCorrectMethods` so that the intended behavior is clear.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -41,32 +42,43 @@
...
+        this.customTabHelper = customTabHelper;
+    }
 
-        openUrlButton = createSimpleButton(getContext(), getFormEntryPrompt().isReadOnly(), context.getString(R.string.open_url), getAnswerFontSize(), this);
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        ViewGroup answerView = (ViewGroup) LayoutInflater.from(context).inflate(R.layout.url_widget_answer, null);
 
-        stringAnswer = getCenteredAnswerTextView(getContext(), getAnswerFontSize());
+        openUrlButton = answerView.findViewById(R.id.url_button);
```
lognaturel: Since we're going to be touching every widget, should we consider using view binding?

collect_app/src/main/res/layout/url_widget_answer.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content">
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/url_button"
+        style="@style/Widget.Collect.Button.Custom"
```
lognaturel, line 9: Could we consider introducing a style that includes all of the options below so that if we want to change the in-form buttons we can do it in one place?

collect_app/src/main/res/layout/url_widget_answer.xml
```diff
@@ -0,0 +1,30 @@
...
+        android:padding="@dimen/margin_small"
+        android:text="@string/open_url"
+        android:layout_marginStart="@dimen/margin_standard"
+        android:layout_marginEnd="@dimen/margin_standard"
+        android:layout_marginTop="@dimen/margin_extra_small"
+        />
+
+    <com.google.android.material.textview.MaterialTextView
+        android:id="@+id/url_answer_text"
+        android:layout_width="match_parent"
```
lognaturel, line 21: Can this also be a style so it's easy to make consistent across widget types?

collect_app/src/main/res/layout/url_widget_answer.xml
seadowg: I'd be interested to see if we can get rid of the button... We're hitting the geowidgets next so might make sense to introduce button styles there if they're hanging around.

collect_app/src/main/res/layout/url_widget_answer.xml
seadowg: Yeah actually this should have a `TextAppearance` style. Probably `TextAppearance.Collect.Body1`.

SaumiaSinghal: Yes, I have seen that style probably in `ProgressDialogFragment`. I will use that.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: Yes, actually. I didn't realize I can directly use them without view binding. And as a lot of tests have to be added, this is a better option.

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
SaumiaSinghal: Ahh.. sorry for the typo.

collect_app/src/main/res/layout/url_widget_answer.xml
lognaturel: That makes sense, @seadowg. Up to y'all to decide whether you want to add do a style with the next set of widgets or not.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
lognaturel: Just a small wording correction, you're actually using the [view binding feature](https://developer.android.com/topic/libraries/view-binding) now rather than explicitly binding your views. I know, it's confusing. Speaking of confusing, where is view binding turned on, @seadowg? I'm confused about not seeing it in the build.gradle file.

collect_app/src/main/res/layout/url_widget_answer.xml
lognaturel: What about also including padding, height, color, etc in a style for answer text? I think we will want to change how answers are displayed and it'd be great to be able to do that in a single place for all widgets.

SaumiaSinghal: Yes, I think we can do that, for TextView and Buttons that we use, as we have the same style for them for many widgets.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
seadowg: I'm also confused. It looks like @SaumiaSinghal is just using `findViewById` with `protected` fields. @lognaturel just to be clear are you saying we should move to using the new [View Binding](https://developer.android.com/topic/libraries/view-binding) feature before we accept this PR? I'd be in favour of that if so.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -41,32 +42,43 @@
 @SuppressLint("ViewConstructor")
 public class UrlWidget extends QuestionWidget implements ButtonWidget {
```
grzesiek2010, line 43: Does that still make sense to implement ButtonWidget? That was created in order to use WidgetViewUtils.createSimpleButton() - one method used in widgets for generating a button but if we are going to use xml files instead of creating those buttons programmatically we probably can get rid of it right?

grzesiek2010: Does that still make sense to implement ButtonWidget? That was created in order to use WidgetViewUtils.createSimpleButton() - one method used in widgets for generating a button but if we are going to use xml files instead of creating those buttons programmatically we probably can get rid of it right?

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -41,32 +42,43 @@
...
+        openUrlButton = answerView.findViewById(R.id.url_button);
+        stringAnswer = answerView.findViewById(R.id.url_answer_text);
 
-        String s = questionDetails.getPrompt().getAnswerText();
+        if (prompt.isReadOnly()) {
+            openUrlButton.setVisibility(GONE);
+        } else {
+            openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            openUrlButton.setOnClickListener(v -> {
+                if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
```
grzesiek2010: I would move this check to onButtonClick() then this expression will look simpler.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: Yes @seadowg :/ I was. I fixed that.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -41,32 +42,43 @@
...
+                    onButtonClick(openUrlButton.getId());
+                }
+            });
+        }
+
+        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        String s = prompt.getAnswerText();
```
grzesiek2010: Please change this name we still have such one-char names in our codebase.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,22 +17,23 @@
...
-import org.odk.collect.android.formentry.questions.WidgetViewUtils;
 import org.odk.collect.android.utilities.CustomTabHelper;
+import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.widgets.interfaces.ButtonWidget;
 
-import static org.odk.collect.android.formentry.questions.WidgetViewUtils.createSimpleButton;
-import static org.odk.collect.android.formentry.questions.WidgetViewUtils.getCenteredAnswerTextView;
-
 /**
  * Widget that allows user to open URLs from within the form
```
grzesiek2010: What do you think about such comments? Even on master (before this big change) there is no single line crated by Yaw (every single line has been edited) of course the whole project exists thanks to Yaw but isn't git enough?

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -41,32 +42,43 @@
...
+            openUrlButton.setOnClickListener(v -> {
+                if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
+                    onButtonClick(openUrlButton.getId());
+                }
+            });
+        }
+
+        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        String s = prompt.getAnswerText();
         if (s != null) {
```
grzesiek2010: The same here when it comes to naming.

grzesiek2010 (changes requested):
Just a couple of small comments.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: I do think having a separate method `onButtonClick()` is redundant. We can have all the functionality in `setOnClickListener()` instead of explicitly calling `onButtonClick()` from there. Should I remove it?

grzesiek2010: I'm not talking about removing that method you can keep since you call it in https://github.com/getodk/collect/pull/3831/files#diff-a5c152dc7425469015c19b7a29b8da07R69
but implementing the interface and using `@Override` above the method doesn't make sense because this method is no longer called like before (using interface from WidgetViewUtils.createSimpleButton()) you do that directly from the same widget (the line I attached).

SaumiaSinghal: Oh okay I understand. Thanks for explaining.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,56 +17,65 @@
...
-        openUrlButton = createSimpleButton(getContext(), getFormEntryPrompt().isReadOnly(), context.getString(R.string.open_url), getAnswerFontSize(), this);
+        openUrlButton = answerView.findViewById(R.id.url_button);
+        stringAnswer = answerView.findViewById(R.id.url_answer_text);
 
-        stringAnswer = getCenteredAnswerTextView(getContext(), getAnswerFontSize());
+        if (prompt.isReadOnly()) {
+            openUrlButton.setVisibility(GONE);
+        } else {
+            openUrlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            openUrlButton.setOnClickListener(v -> {
```
grzesiek2010: You can simplify this: `openUrlButton.setOnClickListener(v ->  onButtonClick());`

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
SaumiaSinghal: Hi @lognaturel! I updated the tests and used  `spy(widgetTestActicity())` instead to verify `unbindService()` is called.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -105,13 +113,14 @@ protected void onDetachedFromWindow() {
...
 
-    @Override
-    public void onButtonClick(int buttonId) {
-        if (!isUrlEmpty(stringAnswer)) {
-            customTabHelper.bindCustomTabsService(getContext(), null);
-            customTabHelper.openUri(getContext(), uri);
-        } else {
-            Toast.makeText(getContext(), "No URL set", Toast.LENGTH_SHORT).show();
+    public void onButtonClick() {
+        if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
```
grzesiek2010: This check will be needed in case of every single widget button. If so it would be better to create our own `WidgetButton` class that will extend `com.google.android.material.button.MaterialButton` (then you can use it in your layout file). In such a class we can then override:
```
    @Override
    public boolean performClick() {
        if (MultiClickGuard.allowClick(getClass().getName())) {
            return super.performClick();
        } else {
            return false;
        }
    }
```

what do you think about this approach?

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,56 +17,63 @@
...
  */
 @SuppressLint("ViewConstructor")
-public class UrlWidget extends QuestionWidget implements ButtonWidget {
+public class UrlWidget extends QuestionWidget {
 
-    private Uri uri;
-    final Button openUrlButton;
-    final TextView stringAnswer;
     private final CustomTabHelper customTabHelper;
+    private Uri uri;
```
grzesiek2010: I would get rid of this field. this is used in just one place `onButtonClick()` so it would be better to add a method like:
```
    private Uri getUri() {
        return Uri.parse(stringAnswer.getText().toString());
    }
```

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: Yes, that would be great. I can create `WidgetButton` class in the `material` package?

SaumiaSinghal: Maybe later, we can shift more common functionality to the `WidgetButton` class.

grzesiek2010: > I can create WidgetButton class in the material package?

I'm not sure that package contains implementations for elements that don't exist the Material Components framework in this case it's about customizing not implementing so maybe our main dir would be a better choice what do you think @seadowg

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,78 +17,88 @@
...
 
-        // finish complex layout
-        LinearLayout answerLayout = new LinearLayout(getContext());
-        answerLayout.setOrientation(LinearLayout.VERTICAL);
-        answerLayout.addView(openUrlButton);
-        answerLayout.addView(stringAnswer);
-        addAnswerView(answerLayout, WidgetViewUtils.getStandardMargin(context));
+        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        String answerText = prompt.getAnswerText();
+        if (answerText != null) {
```
grzesiek2010: Do we need this check? I'm pretty sure we can call setText(null) and it's safe. Could you check and remove it if not needed?

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,78 +17,88 @@
...
+            stringAnswer.setText(answerText);
+        }
 
-        customTabHelper = new CustomTabHelper();
+        return answerView;
     }
 
     private boolean isUrlEmpty(TextView stringAnswer) {
```
grzesiek2010: Please move this method to the bottom of the class since it's private and used only by onButtonClick().

grzesiek2010: The same with getUri()

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -105,13 +115,14 @@ protected void onDetachedFromWindow() {
         }
```
grzesiek2010, line 92: Please add @Override I know it's not required but if we do that for all the other methods let's keep consistency.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -105,13 +115,14 @@ protected void onDetachedFromWindow() {
...
-            customTabHelper.openUri(getContext(), uri);
-        } else {
-            Toast.makeText(getContext(), "No URL set", Toast.LENGTH_SHORT).show();
+    public void onButtonClick() {
+        if (MultiClickGuard.allowClick(QuestionWidget.class.getName())) {
+            if (!isUrlEmpty(stringAnswer)) {
+                customTabHelper.bindCustomTabsService(getContext(), null);
+                customTabHelper.openUri(getContext(), getUri());
+            } else {
+                Toast.makeText(getContext(), "No URL set", Toast.LENGTH_SHORT).show();
```
grzesiek2010: We have `ToastUtils` class which we use for displaying toast so we should use it here as well.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,78 +17,88 @@
...
                 || stringAnswer.getText().toString().isEmpty();
     }
 
+    private Uri getUri() {
+        return Uri.parse(stringAnswer.getText().toString());
+    }
+
     @Override
     public void clearAnswer() {
         Toast.makeText(getContext(), "URL is readonly", Toast.LENGTH_SHORT).show();
```
grzesiek2010: We have `ToastUtils` class which we use for displaying toast so we should use it here as well.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
seadowg: Yeah the `material` module is mostly for implementations of Material Components like @grzesiek2010 suggests. I'd say we could make a `MulticlickSafeButton` that extends `MaterialButton` and just lives in the `org.odk.collect.android.views` seeing as it's just a general view. If it seems like like we want one just for widgets it could live in the `widgets` package.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -108,20 +97,30 @@ public void cancelLongPress() {
...
 
+    @Override
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
         if (customTabHelper.getServiceConnection() != null) {
             getContext().unbindService(customTabHelper.getServiceConnection());
         }
     }
 
+    private boolean isUrlEmpty(TextView stringAnswer) {
```
grzesiek2010, line 108: Even lower please :) both are called only by onButtonClick() so it looks better if we have them below. Sorry for being picky but if you are rebuilding the widget it's good do fix even such small things.

SaumiaSinghal: Oh no that's okay! It's good if we have all these things fixed.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
grzesiek2010: It was a question more to @lognaturel

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -17,78 +17,77 @@
...
         }
 
-        // finish complex layout
-        LinearLayout answerLayout = new LinearLayout(getContext());
-        answerLayout.setOrientation(LinearLayout.VERTICAL);
-        answerLayout.addView(openUrlButton);
-        answerLayout.addView(stringAnswer);
-        addAnswerView(answerLayout, WidgetViewUtils.getStandardMargin(context));
+        stringAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        String answerText = prompt.getAnswerText();
```
grzesiek2010: I think now you can get rid of this variable and just call `stringAnswer.setText(prompt.getAnswerText());` below, it won't be a long or unclear expression.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: @seadowg @grzesiek2010, Should I go ahead and make `MulticlickSafeButton` in the `org.odk.collect.android.views` package?

grzesiek2010: Yes please go ahead you will need it in other widgets either way so I think it's a good idea

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
lognaturel: Sorry, now I'm really confused, too! Yes, @seadowg, that's what I was requesting. I was expecting not to see any more explicit `findViewById` in the final code but they're still there.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
lognaturel: I think in this case it makes sense to remove the whole header comment since it doesn't add much. I also know for sure that @yanokwa doesn't like `@author` tags and is happy for history to be in `git`.

For those who use `@author` tags, the rule is supposed to be that anyone who touches the file adds their name and names are never removed. For that reason I don't typically remove author tags from people I don't know.

grzesiek2010: Great! so @SaumiaSinghal please remove it.

grzesiek2010 (approved):

grzesiek2010:
In 
`How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?`

there is just short `no` but i think if the QA team is available we should test every widget we are reworking, don't you agree?

SaumiaSinghal:
Yes definitely @grzesiek2010! That will ensure there are no regression risks.

seadowg:
Looks like this is ready for QA. As @grzesiek2010 it's best to make sure the PR describes what parts of the app have changed as it may not always be obvious. Here it's good for QA to focus their efforts on the URL Widget.

mmarciniak90:
**Tested with success**

I didn't notice regression

Verified cases:
* no browser installed (Android 8.1, 6.0)
* jump to/from hierarchy view
* light/dark mode
* rotation
* minimization
* lock screen

kkrawczyk123:

PR #3832, opened 2020-05-14
Revise QR code stub setup

seadowg:
This reworks the QR code stub setup so that file saving happens in the test rules. This means that any errors will stop the test rather than us seeing fails later for what looks like unrelated reasons. I've also added cleanup to the stub for good measure (which might help stability if that was a problem) and added and used a `tryAgainOnFail` helper to tackle a couple of flakey actions in the tests.


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.java
```diff
@@ -290,6 +290,15 @@ public T checkIfWebViewActivityIsDisplayed() {
         return (T) this;
     }
 
+    void tryAgainOnFail(Runnable action) {
+        try {
+            action.run();
+        } catch (NoMatchingViewException e) {
+            assertOnPage();
```
seadowg, line 297: We might want to add a sleep in here in future but interested to try this out

grzesiek2010 (approved):

PR #3835, opened 2020-05-14
Fixed scanning inverted QRCodes and refactored the way we scan codes

### LINKED ISSUE #3823: Scanning inverted QR code to apply settings does not work
Collect with #3743 

#### Problem description
Scanning inverted QR code to apply settings does not work

1. User opens `Configure via QR code`
2. User scans inverted QR code with settings
image

#### Expected behavior
QR code is scanned and settings are applied

grzesiek2010:
Closes #3823

I tested the changes manually.

The problem mentioned in the issue was just a small bug which I fixed. Investigating the code I realized we have a lot of duplicated code in `ScannerWithFlashlightActivity` used by `BarcodeWidget` and `QRScannerFragment` used when we scan settings. I refactored the code and now in both cases we use `QRScannerFragment` (`ScannerWithFlashlightActivity` still exists to hold `QRScannerFragment` but the duplicated code is removed).

We need to test scanning settings (no need to check result's just whether the code is scanned or not) and scanning barcodes using the `BarcodeWidget` (with front camera).

Any form with `BarcodeWidget` (+ one using front camera ideally).

No.


collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
```diff
@@ -83,6 +83,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
 
         Intent intent = new IntentIntegrator(getActivity())
                 .setDesiredBarcodeFormats(getSupportedCodeFormats())
+                .setPrompt(getContext().getString(R.string.barcode_scanner_prompt))
```
lognaturel: I think we should make this a separate change for v1.28. Most translators won’t have time to get to this in the next few days so the strings will show up in English where they were translated before.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
```diff
@@ -24,43 +26,70 @@
...
+
+    public enum Action { BARCODE_WIDGET, APPLY_SETTINGS };
+
+    private CaptureManager capture;
+    private DecoratedBarcodeView barcodeScannerView;
     private Button switchFlashlightButton;
     private BeepManager beepManager;
+    private Action action;
+
+    public static QRScannerFragment newInstance(Action action) {
```
lognaturel: Passing in an `Action` to configure the fragment makes it hard to track what the different behavior is between the two types of scanners. Did you consider an abstract `BarcodeScannerFragment` with `SettingsQRCodeScannerFragment` and `BarcodeWidgetScannerFragment` subtypes or something like that? You'd have abstract `handleBarcodeResult` and `getSupportedCodeFormats` methods that are the only things implemented by those concrete classes, making the different behavior really clear.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
```diff
@@ -24,43 +26,70 @@
...
 import org.odk.collect.android.utilities.PermissionUtils;
+import org.odk.collect.android.utilities.WidgetAppearanceUtils;
 
 import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.zip.DataFormatException;
 
 public class QRScannerFragment extends Fragment implements DecoratedBarcodeView.TorchListener {
```
lognaturel: This should be in a different package, now. It's also no longer just a QR code scanner.

lognaturel (changes requested):
Couple of structural suggestions inline.

It'd be great to get tests on the fragment verifying the flashlight and front camera behavior but I think that can follow up.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
grzesiek2010: Right, fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
grzesiek2010: We have over 20 translations for that string the default zxing message also is translatable but I noticed it contains less translations so maybe it's better to have this our own?.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
grzesiek2010: You are right. I Implemented it in that way. Thanks.

collect_app/src/main/java/org/odk/collect/android/preferences/qr/QRScannerFragment.java
lognaturel: Oh, right, this string was already in use for the barcode widget. This change just also uses it in the settings QR code context. That makes a lot of sense, thanks.

lognaturel (approved):
Looks good to me! Great to see duplicated code going away. We can do a separate pass to add automated tests based on this and on the issues found in QA for the QR code scanning improvements.

The refactor looks much better to me. You can see the principle at https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html. I also like the article https://thoughtbot.com/blog/refactoring-replace-conditional-with-polymorphism Clearly there are a lot of violations of this throughout the Collect codebase!

In this case I think it's a simple enough inheritance hierarchy that it feels ok to me. Another option in general to address this kind of issue is to use composition. That's similar to what you were initially going for by passing something into the generic base code scanner. Instead of passing in an enum, though, you'd pass in an object that has some responsibility and can be tested in isolation. The classes would probably look exactly like the subclasses you have now but you'd delegate to the `getSupportedCodeFormats` and `handleScanningResult` methods instead of them being called through polymorphism. We should probably align on which we favor as we get into more of these kinds of refactors.

mmarciniak90:
**Tested with success**

Verified cases:
- scan inverted QR code in barcode widget - back and front camera
- scan standard QR code in barcode widget - back and front camera
- scan inverted QR code with settings
- scan standard QR code with settings
- scan invalid, inverted QR code to apply settings

kkrawczyk123:

PR #3837, opened 2020-05-15
Fix app crashing during form load

### LINKED ISSUE #3836: Collect crashes when open blank form on v1.27.0 beta 0
Collect v1.27.x, Android all from 5.1 to 10 device used...

#### Problem description
Collect crashes when try to open blank form.

1. Download any form
2. Go to Fill Blank Form
3. Click on Downloaded form

#### Expected behavior
The app should not crash, user should be able to fill form

seadowg:
Closes #3836. This stops the app crashing when loading any form.

The core problem ended up being that `DialogFragment#getLayoutInflater` contains a call to `DialogFragment#onCreateDialog` which is where we were calling it from.

Why weren't we seeing this on master? Basically the Fragment dependency (`androix.app.fragment`) was diffferent in our debug and release builds. This was because we we using the [Fragment Testing](https://developer.android.com/training/basics/fragments/testing) dependency which uses `debugImplementation` rather than `testImplementation` (I'd imagine for scary reasons). Our `fragment-testing` dependency was set to version `1.2.3` which meant that in debug versions of the app we'd be using `fragment:1.2.3` where as in release versions of the app we were (implicitly through Jetifier) getting `fragment:1.2.0`. `fragment:1.2.3` does not have the problem (looks like the `getLayoutInflater` code is wrapped in a catch all `try` which is urrrghhh) but `fragment:1.2.0` does.

I first made it so the crash occured on debug as well by adding an explicit `fragment` dependency and setting both it and the `fragment-testing` dep to version `1.2.0`. Tests then failed due to the crash.

I've also made it so the fragment version is set using a variable in gradle which means our debug/release versions should stay the same. This doesn't fix any bugs with Fragments but it makes sure if we run into any they happen in debug and release.

We could just have upgraded to `1.2.3` but I also wanted to protect from this happening again if the problem reappears in the Android lib. Using the Activity's `LayoutInflater` seemed to be a nice change and is also what the docs do in https://developer.android.com/guide/topics/ui/dialogs#CustomLayout.

Should just fix the crash!

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -199,8 +199,6 @@ configurations.all {
 }
 
 dependencies {
-    def fragment_version = "1.2.4"
-
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.browser:browser:1.2.0'
     implementation 'androidx.appcompat:appcompat:1.1.0'
@@ -209,7 +207,7 @@ dependencies {
     implementation 'androidx.exifinterface:exifinterface:1.1.0'
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'androidx.preference:preference:1.1.0'
-    implementation "androidx.fragment:fragment:$fragment_version"
+    implementation "androidx.fragment:fragment:${rootProject.fragmentVersion}"
 
     implementation "com.google.android.gms:play-services-auth:17.0.0"
     implementation "com.google.android.gms:play-services-maps:17.0.0"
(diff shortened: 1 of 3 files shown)


kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.1 and 10

Tested scenarios:
- download form
- fill form
- save form
- edit form
- send form
- server type: Google Drive, Aggregate, Central 

let's wait for @mmarciniak90 to verify on her devices according to high priority.

mmarciniak90:
Tested with success on Android 6.0, 8.1, 9.0

collect_app/build.gradle
```diff
@@ -199,6 +199,8 @@ configurations.all {
 }
 
 dependencies {
+    def fragment_version = "1.2.4"
```
grzesiek2010: Can we fallow the same approach like in case of `daggerVersion` and `okhttp3Version` and keep the version in https://github.com/getodk/collect/blob/master/build.gradle#L64?

grzesiek2010 (changes requested):

collect_app/build.gradle
seadowg: @grzesiek2010 Yes good point! I'm not at my computer right now so happy for you to make and push that change.

grzesiek2010: Ok I'll update it in a minute.

grzesiek2010 (approved):

PR #3838, opened 2020-05-17
Fix code coverage integration

### LINKED ISSUE #3796: Upload coverage reports to codecov.io
Now we are sharding our tests on Firebase Test Lab we get a coverage report for every shard (rather than just one for all the connected tests). In Circle CI we were previously copying the coverage report we downloaded from Firebase with this line (in `.circleci/config.yml`):

```
mkdir -p collect_app/build/outputs/code-coverage/connected && cp firebase/Pixel2-27-en-portrait/artifacts/coverage.ec collect_app/build/outputs/code-coverage/connected/coverage.ec
```

The new reports are downloaded as part of our `test_instrumented` job and are stored for each shard in `firebase/Pixel2-27-en-portrait-shard_x/artifacts/sdcard/coverage.ec` where `x` is the shard number. This means we need to either find some way to include these files as part of our `jacocoTestReport` gradle task or work out a way to merge them before running it.

This should be fixable without modifying CI by downloading coverage files locally (not all 50 just a couple) and then putting together a workflow for merging and uploading to codecov.io. Once we've done that we can make changes that let us automate the upload in Circle CI.

shobhitagarwal1612:
Closes #3796 

Copy all code coverage files from each sharded output dir. 

`jacoco.gradle` already includes `outputs/code-coverage/connected/*coverage.ec` so we don't have to change anything in the plugin config.

I was unable to test the changes locally. For that, I need atleast 2-3 `coverage.ec` from the shards.

This is the only solution I considered.

I noticed that the coverage files are not getting generated in the latest master build. Even if that is the case, then this won't fail the build.

n/a

n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,11 +244,11 @@ workflows:
           filters:
             branches:
               only: master
-#      - report_coverage:
-#          requires:
-#            - test_unit
-#            - test_instrumented
-#          # We'd like to also filter by username and remove the if/fi above, but username filtering is not supported in CircleCI
-#          filters:
-#            branches:
-#              only: master
+      - report_coverage:
+          requires:
+            - test_unit
+            - test_instrumented
+          # We'd like to also filter by username and remove the if/fi above, but username filtering is not supported in CircleCI
+          filters:
+            branches:
+              only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,8 @@ jobs:
           name: Move Firebase coverage report
           command: |
             if [[ "$CIRCLE_PROJECT_USERNAME" == "getodk" ]]; then \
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
The `coverage.ec` files are now visible in the latest master build. I suspect, the last time was a fluke then
https://app.circleci.com/pipelines/github/getodk/collect/277/workflows/f11a487c-0cda-464e-b95b-3a9d68432827/jobs/8882

@seadowg Do you mind sharing a couple of `coverage.ec` files from the GCS or maybe merge this PR to test out the changes (and revert if it doesn't help)?

seadowg (approved):
@shobhitagarwal1612 I'm going to merge and then if we run into problems I'll just revert as you suggest!

seadowg:
This didn't work so I'm reverting. @shobhitagarwal1612 if you're still able to look at this I've realized you're able to try out anything you do in the `circleci/android:api-27` Docker image (that the `report_coverage` step uses). That'll help you make sure commands exist (which was the problem here).

PR #3841, opened 2020-05-18
Upgrade to the new Firebase Crashlytics

### LINKED ISSUE #3807: Upgrade to the new Firebase Crashlytics
#### Problem description
The topic from the forum https://forum.getodk.org/t/fixing-leaks-in-collect-app/26137/3 made me thinking what actually fabric is for. I took a look at the documentation and found out that it's needed fro crashlytics but now it's legacy and we can switch to the new [Firebase Crashlytics](https://firebase.google.com/docs/crashlytics/get-started-new-sdk?platform=android) and get rid of fabric.

@lognaturel @seadowg any thoughts pros/cons?

grzesiek2010:
Closes #3807 

Reviewed implemented changes according to the doc: https://firebase.google.com/docs/crashlytics/upgrade-sdk?platform=android

https://firebase.google.com/docs/crashlytics/upgrade-sdk?platform=android

it does't require testing.

No.

No.


lognaturel (approved):
Whoops, good catch upgrading that. Could you please do a quick sanity check and make sure you actually see a crash in the Crashlytics log for debug builds before merging?

lognaturel:
Also, I didn't realize that we track warnings. I think we should exclude them but we can do that for next release.

grzesiek2010:
> Could you please do a quick sanity check and make sure you actually see a crash in the Crashlytics log for debug builds before merging?

I can see something in logs but unfortunately can't test it according to [Test your Firebase Crashlytics implementation](https://firebase.google.com/docs/crashlytics/force-a-crash?platform=android) because it seems outdated and uses old Fabric/Crashlytics.

> Also, I didn't realize that we track warnings. I think we should exclude them but we can do that for next release.

You mean logging only errors right? If so I did that https://github.com/getodk/collect/pull/3841/commits/fe2bbc47b34100529be1e645e02fd54a969fe992

grzesiek2010:
> You mean logging only errors right? If so I did that fe2bbc4

wait wait wait... I agreed but those logs are then visible if we have a crash and might be helpful sometimes so why do you want to get rid of it?

grzesiek2010:
> wait wait wait... I agreed but those logs are then visible if we have a crash and might be helpful sometimes so why do you want to get rid of it?

I removed the last commit to keep logging warning according to what I said. Do you think we should review warning and remove those we don't need in this pr or add as a separate one?

lognaturel:
For context, @grzesiek2010 and I briefly talked about this in a call. I had never used the log view in Crashlytics so that's nice to know about. But there really is a lot of noise in warnings.

> Do you think we should review warning and remove those we don't need in this pr or add as a separate one?

I think it can go into a separate PR.

PR #3843, opened 2020-05-18
Add Progress Bar in the Blank Forms Fragment in 'Delete Saved Forms'

### LINKED ISSUE #3842: The progress bar is not visible on `Delete Saved Form`
Collect v1.26.3

#### Problem description
The progress bar is not visible on `Delete Saved Form`

1. User download a lot of forms
2. User goes to `Delete Saved Form` view
3. User opens `Blank Forms`

#### Expected behavior
Progress bar is visible

ezgif com-video-to-gif

### LINKED ISSUE #3965: Progress Bar is Not Shown on clicking `Delete Forms` Button in `Blank Forms` Tab in `Delete Saved Forms`
Collect v1.27.x

#### Problem description
The progress bar is not shown when deleting blank forms in Delete Saved Forms Activity. This behaviour differs from the `Saved Form tab`, as there progress bar is shown to account for the progress of deleting forms. 

1. User download a lot of forms
2. User goes to Delete Saved Form view
3. User opens Blank Forms
4. User selects a lot of forms
5. User clicks on `Delete Selected` Button
6. User clicks `Delete Forms`

#### Expected behavior
Progress bar should be displayed

ezgif com-video-to-gif (4)

SaumiaSinghal:
Closes #3965 

I tested it.

I added the progress dialog and published the progress, just as it is there in the Saved Forms fragment.

No

No specific form

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
@@ -22,6 +22,7 @@ import android.widget.ListView;
 
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
+import androidx.fragment.app.DialogFragment;
 import androidx.loader.content.CursorLoader;
 
 import org.odk.collect.android.R;
@@ -154,6 +155,17 @@ public class FormManagerList extends FormListFragment implements DiskSyncListene
         alertDialog.setButton(DialogInterface.BUTTON_NEGATIVE, getString(R.string.delete_no),
                 dialogYesNoListener);
         alertDialog.show();
+    }
+
+    @Override
+    public void progressUpdate(Integer progress, Integer total) {
+        String message = String.format(getResources().getString(R.string.deleting_form_dialog_update_message), progress, total);
+        ProgressDialogFragment existingDialog = (ProgressDialogFragment) requireActivity().getSupportFragmentManager()
+                .findFragmentByTag(ProgressDialogFragment.class.getName());
+
+        if (existingDialog != null) {
+            existingDialog.setMessage(message);
+        }
     }
 
     /**
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
@@ -179,23 +179,19 @@ public class FormManagerList extends FormListFragment implements DiskSyncListene
     private void deleteSelectedForms() {
         // only start if no other task is running
         if (backgroundTasks.deleteFormsTask == null) {
-            runDeleteSelectedFormsTask();
+            Bundle args = new Bundle();
+            args.putSerializable(ProgressDialogFragment.MESSAGE, getResources().getString(R.string.form_delete_message));
+            args.putBoolean(ProgressDialogFragment.CANCELABLE, false);
+            DialogUtils.showIfNotShowing(ProgressDialogFragment.class, args, getActivity().getSupportFragmentManager());
+
+            backgroundTasks.deleteFormsTask = new DeleteFormsTask();
+            backgroundTasks.deleteFormsTask
+                    .setContentResolver(getActivity().getContentResolver());
+            backgroundTasks.deleteFormsTask.setDeleteListener(this);
+            backgroundTasks.deleteFormsTask.execute(getCheckedIdObjects());
         } else {
             ToastUtils.showLongToast(R.string.file_delete_in_progress);
         }
-    }
-
-    private void runDeleteSelectedFormsTask() {
-        Bundle args = new Bundle();
-        args.putSerializable(ProgressDialogFragment.MESSAGE, getResources().getString(R.string.form_delete_message));
-        args.putBoolean(ProgressDialogFragment.CANCELABLE, false);
-        DialogUtils.showIfNotShowing(ProgressDialogFragment.class, args, getActivity().getSupportFragmentManager());
-
-        backgroundTasks.deleteFormsTask = new DeleteFormsTask();
-        backgroundTasks.deleteFormsTask
-                .setContentResolver(getActivity().getContentResolver());
-        backgroundTasks.deleteFormsTask.setDeleteListener(this);
-        backgroundTasks.deleteFormsTask.execute(getCheckedIdObjects());
     }
 
     @Override
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
@@ -88,11 +88,16 @@ public class FormManagerList extends FormListFragment implements DiskSyncListene
                 && backgroundTasks.deleteFormsTask.getStatus() == AsyncTask.Status.FINISHED) {
             deleteComplete(backgroundTasks.deleteFormsTask.getDeleteCount());
         }
+        if (backgroundTasks.deleteFormsTask == null) {
+            DialogUtils.dismissDialog(ProgressDialogFragment.class, getActivity().getSupportFragmentManager());
+        }
     }
 
     @Override
     public void onPause() {
-        backgroundTasks.diskSyncTask.setDiskSyncListener(null);
+        if (backgroundTasks.diskSyncTask != null) {
+            backgroundTasks.diskSyncTask.setDiskSyncListener(null);
+        }
         if (backgroundTasks.deleteFormsTask != null) {
             backgroundTasks.deleteFormsTask.setDeleteListener(null);
         }
@@ -174,20 +179,23 @@ public class FormManagerList extends FormListFragment implements DiskSyncListene
     private void deleteSelectedForms() {
         // only start if no other task is running
         if (backgroundTasks.deleteFormsTask == null) {
-
-            Bundle args = new Bundle();
-            args.putSerializable(ProgressDialogFragment.MESSAGE, getResources().getString(R.string.form_delete_message));
-            args.putBoolean(ProgressDialogFragment.CANCELABLE, false);
-            DialogUtils.showIfNotShowing(ProgressDialogFragment.class, args, getActivity().getSupportFragmentManager());
-
-            backgroundTasks.deleteFormsTask = new DeleteFormsTask();
-            backgroundTasks.deleteFormsTask
-                    .setContentResolver(getActivity().getContentResolver());
-            backgroundTasks.deleteFormsTask.setDeleteListener(this);
-            backgroundTasks.deleteFormsTask.execute(getCheckedIdObjects());
+            runDeleteSelectedFormsTask();
         } else {
             ToastUtils.showLongToast(R.string.file_delete_in_progress);
         }
+    }
+
+    private void runDeleteSelectedFormsTask() {
+        Bundle args = new Bundle();
+        args.putSerializable(ProgressDialogFragment.MESSAGE, getResources().getString(R.string.form_delete_message));
+        args.putBoolean(ProgressDialogFragment.CANCELABLE, false);
+        DialogUtils.showIfNotShowing(ProgressDialogFragment.class, args, getActivity().getSupportFragmentManager());
+
+        backgroundTasks.deleteFormsTask = new DeleteFormsTask();
+        backgroundTasks.deleteFormsTask
+                .setContentResolver(getActivity().getContentResolver());
+        backgroundTasks.deleteFormsTask.setDeleteListener(this);
+        backgroundTasks.deleteFormsTask.execute(getCheckedIdObjects());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
```diff
@@ -153,13 +156,24 @@ public void onClick(DialogInterface dialog, int i) {
...
+        progressDialog.setMessage(message);
+    }
     /**
      * Deletes the selected files.First from the database then from the file
      * system
      */
     private void deleteSelectedForms() {
         // only start if no other task is running
         if (backgroundTasks.deleteFormsTask == null) {
+            progressDialog = new ProgressDialog(getContext());
```
seadowg: We shouldn't be using `ProgressDialog` any longer as [it's deprecated](https://developer.android.com/reference/android/app/ProgressDialog). Annoyingly the warning doesn't show in Android Studio for some reason!

Ideally there would be a different progress indicator that doesn't block the user. Here I don't want to get into that redesign though so for the moment we should just use a `ProgressDialogFragment`.

collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
```diff
@@ -153,13 +156,24 @@ public void onClick(DialogInterface dialog, int i) {
         alertDialog.show();
     }
 
+    @Override
+    public void progressUpdate(int progress, int total) {
```
seadowg: I don't think we need to add the progress update functionality. It is extra complexity that we could add later if we feel we really need it. For the moment the dialog could just have the message `"Deleting..."`.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
SaumiaSinghal: Hi @seadowg! We also use `ProgressDialog` in Delete Saved Forms fragment. Should we remove `ProgressDialog` from there too? Also, I found that the UI would look better if `layout_centerVertical` is set to true in the `progress_dialog.xml` file.

collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
```diff
@@ -160,6 +163,10 @@ public void onClick(DialogInterface dialog, int i) {
     private void deleteSelectedForms() {
         // only start if no other task is running
         if (backgroundTasks.deleteFormsTask == null) {
+            dialogFragment.setMessage(getResources().getString(R.string.form_delete_message));
+            dialogFragment.setCancelable(false);
+            dialogFragment.show(getActivity().getSupportFragmentManager(), ProgressDialogFragment.COLLECT_PROGRESS_DIALOG_TAG);
```
seadowg: Like in the other PR this should use `DialogUtils.showIfNotShowing` and `DialogUtils.dismissDialog`. That way you also don't need to hold a reference to the dialog like you do here. In fact, it's important to realize you shouldn't ever hold a reference to a `Fragment` instead. This is because if the Activity is recreated (like during a rotation) the Fragment will be recreated as well and the instance you have a reference to will not be the instance that is actually displayed. This causes a memory leak (holding a reference to an object that could have been garbage collected) and also might lead to bugs.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
```diff
@@ -160,6 +163,12 @@ public void onClick(DialogInterface dialog, int i) {
     private void deleteSelectedForms() {
         // only start if no other task is running
         if (backgroundTasks.deleteFormsTask == null) {
+
+            Bundle args = new Bundle();
+            args.putSerializable(ProgressDialogFragment.MESSAGE, getResources().getString(R.string.form_delete_message));
+            args.putSerializable(ProgressDialogFragment.CANCELABLE, "true");
```
seadowg: You can just do `putBoolean` here which will be easier on the other side!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/fragments/FormManagerList.java
seadowg: Also could you add a test for this new functionality to `ProgressDialogFragment`? I think before adding/changing functionality like this to a reusable component you should try adding a test. This ensures you actually have coverage and also gives you an opportunity to think about the interface you're adding/changing.

SaumiaSinghal: Yes that would be great! I'll add tests. Also, I wanted to ask, should we use `ProgressDialogFragment` in `DeleteSavedForms Fragment`, as well?

seadowg: @SaumiaSinghal probably a good idea down the line but no need to do it as part of this PR

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragmentTest.java
```diff
@@ -106,6 +106,26 @@ public void setMessage_beforeDialogExists_setsMessageWhenDialogShown() {
...
+
+            Bundle args = new Bundle();
+            args.putBoolean(ProgressDialogFragment.CANCELABLE, false);
+            fragment.setArguments(args);
+            fragment.show(activity.getSupportFragmentManager(), "TAG");
+            shadowOf(getMainLooper()).idle();
+
+            AlertDialog dialog = (AlertDialog) fragment.getDialog();
+
+            activity.finish();
```
seadowg: I don't think this is right. Did you try commenting out your code to check this test passes and fails when you'd expect? 

Closing the activity might not "cancel" the dialog but it definitely wouldn't be on screen anymore so this test doesn't really make sense to me. You should be able to check whether the dialog is cancellable or not by calling `isCancellable` on it.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragment.java
```diff
@@ -104,6 +108,10 @@ private void setupView(@NonNull AlertDialog dialog) {
             ((TextView) dialogView.findViewById(R.id.message)).setText(getArguments().getString(MESSAGE));
         }
 
+        if (getArguments() != null && Objects.equals(getArguments().getBoolean(CANCELABLE), false)) {
```
seadowg: I think this could just be:

```
if (getArguments() != null) {
    dialog.setCancelable(getArguments().getBoolean(CANCELABLE), true));
}
```
Is that right?

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragmentTest.java
```diff
@@ -106,6 +107,23 @@ public void setMessage_beforeDialogExists_setsMessageWhenDialogShown() {
         });
     }
 
+    @Test
+    public void setCancellable_makesTheDialogNotCancellable() {
+        activityScenario.onActivity(activity -> {
```
SaumiaSinghal, line 112: Hi @seadowg! This test is failing. I'm looking into it.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragment.java
```diff
@@ -104,6 +106,11 @@ private void setupView(@NonNull AlertDialog dialog) {
             ((TextView) dialogView.findViewById(R.id.message)).setText(getArguments().getString(MESSAGE));
         }
 
+        if (getArguments() != null) {
+            dialog.setCancelable(getArguments().getBoolean(CANCELABLE));
```
seadowg: Did you have to set this on the dialog and the fragment? I'm wondering if we maybe only need it on one of them.

I find this really confusing with `DialogFragment` 😕. I don't understand why the `Fragment` and the `Dialog` both have ther dialog interface (like `setCancellable`)...

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragment.java
SaumiaSinghal: Thanks @seadowg for pointing it out. We need it only on the `Fragment`. The `Dialog` is shown as part of the `Fragment`, so we need to be concerned about the `fragment` instead of the `dialog`. I found it has been mentioned [here](https://developer.android.com/reference/android/app/DialogFragment.html#setCancelable(boolean)) in the docs as well.

seadowg: Ah cool. Nice find in the docs. That makes sense!

seadowg (approved):

seadowg:

kkrawczyk123:
@SaumiaSinghal I can't see any changes on both Delete Saved Forms tabs - neither the progress bar and the progress dialog that you mentioned in the description above even if I have more than 1000 submissions and ~100 blank forms. I have the same result on all devices even when trying on the slowest device that I have with Android 5.1. 
I was expecting the same results as we have on other menu pages mean:
Screenshot_20200708-140101
When fixing please keep in mind that the Delete Saved Forms view is changing in #3919

kkrawczyk123:

SaumiaSinghal:
Hi @kkrawczyk123! Thanks for testing. I tested on my device with latest master and the progress bar is showing when I click `Delete Form` Button on both tabs.

ezgif com-video-to-gif (2)

kkrawczyk123:
@SaumiaSinghal but the linked issue #3842 doesn't say anything about the actual action of deleting forms, just opening the view. Is there any chance that you linked the wrong issue by mistake?

SaumiaSinghal:
@kkrawczyk123 I noticed that the progress bar was not visible when the user presses `Delete Forms` button in `Blank Forms` tab in Delete Saved Forms activity. However, the progress bar for the same action was visible for the `Saved Forms` tab. This was some inconsistent behavior that I noticed, and the PR fixes the same.  

Does the issue #3842 says that the progress bar should be visible for loading the forms in both the tabs.

mmarciniak90:
@SaumiaSinghal a good practice is to check the previous version (e.g. v1.25 ) than the version added to the issue (v1.26.3). It could be helpful to reproduce. I created a video for you. I hope it will dispel all doubts. If not, let me know.

ezgif com-video-to-gif

SaumiaSinghal:
@mmarciniak90 I think I misunderstood the issue sorry. To fix the issue, a progress bar should be displayed as soon as the user switches on to any tabs in Delete Forms activity,  to load the forms in the tab. Right? 

I think I solved some other issue, as shown in the gif that I attached, that the progress bar is not visible in the blank form tab in Delete forms activity. 

If it is so, shall I solve #3842 in another PR and link this PR to a newly openend issue that concerns this PR?

mmarciniak90:
> To fix the issue, a progress bar should be displayed as soon as the user switches on to any tabs in Delete Forms activity, to load the forms in the tab. Right?

The behavior should be exactly the same as in v1.25

> If it is so, shall I solve #3842 in another PR and link this PR to a newly openend issue that concerns this PR?

Yes, this workflow sounds good

SaumiaSinghal:
Hi @mmarciniak90! I linked to the correct issue. Can you test it now please?

SaumiaSinghal:

kkrawczyk123:
@SaumiaSinghal I have noticed an issue with the scenario:
1. Don't keep activities is turned on
2. Go to Delete Saved Forms/Blank Forms
3. Select All
4. Click on Delete Selected: dialog with progress bar displays
5. Minimize an app
6. Go back to App
The progress number stops to update. When all forms ale removed dialog does not disappear and freezes an app - there is no way to move backward using the back button. The app becomes useable again after killing it in the background.
Gif of reproduction steps:
ezgif com-video-to-gif (12)

SaumiaSinghal:
@kkrawczyk123 Thanks for reporting the issue. I have fixed that. Can you tests again please?

mmarciniak90:

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- progress dialog on Deleting Saved Forms
- progress dialog on Deleting Blank Forms
- light mode
- dark mode
- screen rotation
- minimize app
- extinguish devices screen
- don't keep activities

lognaturel:
I see the `needs testing` label was still on, probably because @mmarciniak90 was still planning to take a look. Sorry about that. @kkrawczyk123's cases and Android versions look sufficient to me but I've certainly been wrong before!

mmarciniak90:
I did a quick check on Android versions: 7.0, 8.1, 9.0 :+1: 

PR #3845, opened 2020-05-19
Prevent crash when adding repeat from hierarchy

### LINKED ISSUE #3844: Crash on creating repeat
Discovered on Firebase: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/dd74083d67f410ebf91ab087006aac32?time=last-seven-days&sessionId=5EC0D6D4012900012EF581F660321501_DNE_0_v2

Collect v1.27.0 Beta 1

#### Problem description

It looks like Collect is crashing when adding repeats from the hierarchy view if "Don't keep activities" is enabled (probably originally happened on a low mem device).

1. Enable "Don't keep Activities" in Developer Settings
2. Open any form with repeats
3. Open Hierarchy view and navigate to the repeat group
4. Click ➕
5. Click "Add"

#### Expected behavior

The Add Repeat dialog should not be shown and the app should not crash when the repeat is added.

seadowg:
Closes #3844 

Tests failed when running with "Don't keep activities" option so was able to verify that way.

I think using the `FormEntryViewModel` in the hierarchy activity seems a little strange but I think that's mainly because of the static state it relies on in the `FormController`. In the future it looks like really we'd want to be able to initialize the `FormEntryViewModel` with a form and a starting point ("index"). That or the hierarchy view would be a Fragment in the `FormEntryActivity`. I think for the moment it's nice that we can pull logic into "view actions" in the view model from the `FormHierarchyActivity`.

Should just fix the crash! Good to check repeats and creating them from the hierarchy view.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -138,13 +138,13 @@ import org.odk.collect.android.utilities.ApplicationConstants;
 import org.odk.collect.android.utilities.DestroyableLifecyleOwner;
 import org.odk.collect.android.utilities.DialogUtils;
 import org.odk.collect.android.utilities.FileUtils;
-import org.odk.collect.android.utilities.QuestionFontSizeUtils;
 import org.odk.collect.android.utilities.FormNameUtils;
 import org.odk.collect.android.utilities.ImageConverter;
 import org.odk.collect.android.utilities.MediaManager;
 import org.odk.collect.android.utilities.MediaUtils;
 import org.odk.collect.android.utilities.PermissionUtils;
 import org.odk.collect.android.utilities.PlayServicesUtil;
+import org.odk.collect.android.utilities.QuestionFontSizeUtils;
 import org.odk.collect.android.utilities.ScreenContext;
 import org.odk.collect.android.utilities.SnackbarUtils;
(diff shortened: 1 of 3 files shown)


grzesiek2010 (approved):

grzesiek2010:
What about enabling `Don't keep activities` option for our espresso tests? Would that be beneficial?

seadowg:
@grzesiek2010 yeah I was just wondering if it would be possible to run both. Maybe we should see how the whole suite runs and then if it's green we could run with "Don't keep activities" as part of regression testing?

grzesiek2010:
> then if it's green we could run with "Don't keep activities" as part of regression testing?

you mean locally?

seadowg:
@grzesiek2010 yeah. Or we could probably set it up on Firebase which would be faster. I wouldn't want to run CI with "Don't keep activities" on as the default as I think we'd run into the opposite set of bugs where things only work when "Don't keep activities" is on.

We could try and set it up so it runs on both on every commit - basically starting a test matrix of different configurations. I'm not sure I'd feel our tests are stable enough right now to add another parallel test configuration though.

What do you think? Should we make a "Don't keep activities" run a standard part of our regression testing?

mmarciniak90:
**Tested with success**

We can confirm that the mentioned crash is no longer visible.

Verified cases:
- we verified a few forms with repeats
- `Don't keep activities` turned on
- `Don't keep activities` turned off

PR #3852, opened 2020-05-21
Fix code coverage integration (II)

### LINKED ISSUE #3796: Upload coverage reports to codecov.io

shobhitagarwal1612:
Closes #3796, follow up of #3838 

Replaced the `cpio` command with `cp --parents` to copy the parent dir structure. 
Tested by creating dummy directories with files (in linux). Then ran this command to confirm if the filtered files are copied in the dest dir.

No

No

n/a

n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,8 @@ jobs:
           name: Move Firebase coverage report
           command: |
             if [[ "$CIRCLE_PROJECT_USERNAME" == "getodk" ]]; then \
-              mkdir -p collect_app/build/outputs/code-coverage/connected && cp firebase/Pixel2-27-en-portrait/artifacts/coverage.ec collect_app/build/outputs/code-coverage/connected/coverage.ec
+            mkdir -p collect_app/build/outputs/code-coverage/connected
+            find firebase -iname 'coverage.ec' -exec cp --parents \{\} collect_app/build/outputs/code-coverage/connected \;
             fi
       - run:
           name: Generate JaCoCo report


seadowg:
@shobhitagarwal1612 I'm a little confused while trying this out as I can't find the coverage files in the bucket. Do you have access? If not I'm wondering if we could find a time to go through it together on a screen share call.

shobhitagarwal1612:
> I can't find the coverage files in the bucket. Do you have access? If not I'm wondering if we could find a time to go through it together on a screen share call.

Yes, that would be fantastic.

seadowg:
Tried this out in the Circle CI docker image and it's looking good. Will give this a merge end of day so reverting isn't disruptive to anyone else (if we have to).

seadowg (approved):

PR #3857, opened 2020-05-21
Update opendatakit links

yanokwa:
I did a search for "open data kit" and "opendatakit" and "opendatakit.org" and made targeted changes. I'd love the reviewer to check to make sure no inadvertent changes were made AND that I didn't miss any necessary changes. I tried to make sure no changes were made to namespaces and the opendatakit.appspot.com links.

Most changes are to non-executable files (so should be safe). 8619d311337097028c2570de8a08a563ee25b21e should be the only dangerous change.


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2837,7 +2837,7 @@ private void updateFieldListQuestions(FormIndex lastChangedIndex) {
         // identification and removal in the same pass but removal has to be done in a loop that
         // starts from the end and itemset-based select choices will only be correctly recomputed
         // if accessed from beginning to end because the call on sameAs is what calls
-        // populateDynamicChoices. See https://github.com/opendatakit/javarosa/issues/436
+        // populateDynamicChoices. See https://github.com/github.com/getodk//javarosa/issues/436
```
seadowg: Little find and replace typo there I think

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -72,7 +72,7 @@
 import static org.odk.collect.android.tasks.sms.SmsSender.SMS_SEND_ACTION;
 
 /**
- * The Open Data Kit Collect application.
+ * The ODK Collect application.
```
seadowg: I think we can get rid of this comment. Doesn't seem like it adds anything.

seadowg (changes requested):
Just a couple of changes needed I think.

@lognaturel  One thing that we should probably change is the Google Drive root folder we're using (https://github.com/getodk/collect/blob/b41f4bc6fcb4290f22cde48990192128ec9589b4/collect_app/src/main/java/org/odk/collect/android/utilities/gdrive/DriveHelper.java#L39). I'm thinking we'd need to create a migration for this as well though?

lognaturel:
> One thing that we should probably change is the Google Drive root folder we're using

I figured it wasn't worth changing but I guess that with a migration it's relatively painless? Can two folders in the same container have the same name? I think it's entirely possible someone would already have an 'ODK' folder. Seems right to treat this as its own issue, anyway.

lognaturel (approved):

lognaturel:
@seadowg I went through the diff and things look good to me. Please merge -- I can't do it unless you update your review.

seadowg (approved):

PR #3859, opened 2020-05-22
Fix crash when user removes all repeat groups and presses back button

### LINKED ISSUE #3799: Collect crashes when user removes all repeat groups and clicks on device back button

#### Problem description
Collect crashes when user removes all repeat groups and clicks on device back button

1. User starts form [nested-repeats.xml.txt](https://github.com/getodk/collect/files/4519317/nested-repeats.xml.txt)
2. User adds one group
3. User goes to hierarchy view
4. User clicks on bin
5. User sees empty hierarchy view

6. User clicks on the device back button
7. Collect crashes

##### Alternative case
1. User starts form [repeatTest.xml.txt](https://github.com/getodk/collect/files/4519385/repeatTest.xml.txt)
2. User adds one group
3. User goes to hierarchy view
4. User clicks on bin
5. User clicks on the device back button
6. User is moved to form, dialog 
7. Error dialog is visible in background

#### Expected behavior
Question dialog allowing to add group should appear without errors and crashes

#### Notes
Noticed during testing #3656

SaumiaSinghal:
Closes #3799 

I tested it for both the cases mentioned in the issue.

When the tree element of the form is null, the user is directed to the beginning of the form to create a new repeat.

No regression risks, as such now.

[nested-repeats.xml.txt](https://github.com/getodk/collect/files/4667376/nested-repeats.xml.txt)
[repeatTest.xml.txt](https://github.com/getodk/collect/files/4667377/repeatTest.xml.txt)

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1200,19 +1200,19 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             case FormEntryController.EVENT_REPEAT:
                 // should only be a group here if the event_group is a field-list
                 try {
-                    AuditUtils.logCurrentScreen(formController, formController.getAuditEventLogger(), System.currentTimeMillis());
-
                     FormEntryCaption[] groups = formController
                             .getGroupsForCurrentIndex();
                     FormEntryPrompt[] prompts = formController.getQuestionPrompts();
 
-                    TreeElement treeElement = formController.getFormDef().getInstance().
-                            resolveReference(prompts[0].getIndex().getReference());
-                    if(treeElement == null) {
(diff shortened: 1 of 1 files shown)


grzesiek2010 (changes requested):
I think this is not a perfect solution. You catch that case and call `createViewForFormBeginning` that means if you have let's say a form with 10 questions and then a repeatable group and you reproduce the steps on that group you will be moved to the begging of the form.
I know that understanding and dealing with indexes might be difficult (it's still not easy for me) so let me know if you need help I can dig deeper into possible solutions.

SaumiaSinghal:
Thanks @grzesiek2010! yes please go ahead. I understand the problem with my solution. I will search more on how this can be fixed without simply going to the beginning of the form, each time this problem is encountered.

grzesiek2010:
@SaumiaSinghal  here is my proposal: https://github.com/getodk/collect/pull/3912 please take a look.

lognaturel:
Good team work here!

PR #3863, opened 2020-05-25
Fix crash in AndroidXPreferencesActivity

### LINKED ISSUE #3862: Crash when rotating Form Metadata screen
Collect v1.26.3, v1.27.0

#### Problem description

The app is crashing when rotating the Form Metadata screen due to some weird lifecycle issues. This was seen spiking on [Firebase](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/4bfc0982e50cafc28984847510f7d859?time=last-seven-days&sessionId=5ECBCD8F00CD000168BC6701CE9E091A_DNE_0_v2).

1. Open the app and navigate to Form Metadata in General Settings
2. Click on the "Username" setting to open the dialog
3. Rotate the screen a few times until there is a crash

#### Expected behavior

The app shouldn't crash!

seadowg:
Closes #3862

Drove change out with a test and also manually verified.

Should just fix the issue! Maybe worth a run through the User Interface and Metadata screens as these are the only ones using this Activity.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/AndroidXPreferencesActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/AndroidXPreferencesActivity.java
@@ -20,9 +20,9 @@ import org.odk.collect.android.activities.CollectAbstractActivity;
 
 public class AndroidXPreferencesActivity extends CollectAbstractActivity {
 
-    private static final String KEY_EXTRA_FRAGMENT = "key_extra_fragment";
-    private static final int EXTRA_FRAGMENT_FORM_METADATA = 1;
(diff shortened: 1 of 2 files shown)


lognaturel (approved):
Normally I wouldn't merge this immediately before a release but in this case I think it's the right thing to do because:
- clearly our user behavior has changed. We can be almost certain that it's because of [this Nigeria-based agriculture study](https://twitter.com/FmardPace/status/1264488477128888320). They explicitly have folks set metadata. Given that hundreds of folks are having Collect just close on them, the benefit of releasing immediately is high.
- It's a null check that's easy to verify as safe and [correct](https://stackoverflow.com/a/26855936/137744). I also manually spent some time on both affected settings screens and the risk seems very low.

PR #3864, opened 2020-05-25
Instance map view improvements

### LINKED ISSUE #3828: Instance map view improvements
#3778 adds a summary view when tapping an instance on a form's instance map. This issue is to track enhancements to that view:
- grow the selected marker icon ~and make it red~, same as in Google Maps
- dismiss bottom sheet on back button press
- dismiss the bottom sheet if there's a tap that's not on a map button or on an instance
- maintain the bottom sheet on device rotation

grzesiek2010:
Closes #3828

This is what we agreed on in the issue.

It requires testing FormMapActivity. Only this functionality might be affected.

Any form with geopoint widget.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -364,7 +364,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
         findViewById(R.id.info).setVisibility(View.GONE);
         Chip openFormButton = findViewById(R.id.openFormChip);
         openFormButton.setVisibility(View.VISIBLE);
-        openFormButton.setText(canEdit ? R.string.review_data : R.string.view_sent_forms);
+        openFormButton.setText(canEdit ? R.string.review_data : R.string.view_data);
         openFormButton.setChipIcon(ContextCompat.getDrawable(this, canEdit ? R.drawable.ic_edit : R.drawable.ic_visibility));
         openFormButton.setOnClickListener(v -> {
             summarySheet.setState(BottomSheetBehavior.STATE_HIDDEN);
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -68,6 +68,7 @@ import timber.log.Timber;
 public class FormMapActivity extends BaseGeoMapActivity {
     public static final String MAP_CENTER_KEY = "map_center";
     public static final String MAP_ZOOM_KEY = "map_zoom";
+    public static final String MAP_SELECTED_SUBMISSION_ID = "map_selected_submission_id";
 
     private FormMapViewModel viewModel;
 
@@ -92,6 +93,8 @@ public class FormMapActivity extends BaseGeoMapActivity {
      * True if the map viewport has been initialized, false otherwise.
      */
     private boolean viewportInitialized;
+
+    private int selectedSubmissionId = -1;
 
     @VisibleForTesting public ViewModelProvider.Factory viewModelFactory;
 
@@ -123,6 +126,19 @@ public class FormMapActivity extends BaseGeoMapActivity {
         setContentView(R.layout.instance_map_layout);
         summarySheet = BottomSheetBehavior.from(findViewById(R.id.submission_summary));
         summarySheet.setState(BottomSheetBehavior.STATE_HIDDEN);
+        summarySheet.addBottomSheetCallback(new BottomSheetBehavior.BottomSheetCallback() {
+            @Override
+            public void onStateChanged(@NonNull View bottomSheet, int newState) {
+                if (newState == BottomSheetBehavior.STATE_HIDDEN) {
+                    selectedSubmissionId = -1;
+                }
+            }
+
+            @Override
+            public void onSlide(@NonNull View bottomSheet, float slideOffset) {
+
+            }
+        });
 
         TextView titleView = findViewById(R.id.form_title);
         titleView.setText(viewModel.getFormTitle());
@@ -154,6 +170,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
         }
         state.putParcelable(MAP_CENTER_KEY, map.getCenter());
         state.putDouble(MAP_ZOOM_KEY, map.getZoom());
+        state.putInt(MAP_SELECTED_SUBMISSION_ID, selectedSubmissionId);
     }
 
     @SuppressLint("MissingPermission") // Permission handled in Constructor
@@ -183,6 +200,10 @@ public class FormMapActivity extends BaseGeoMapActivity {
 
         map.setFeatureClickListener(this::onFeatureClicked);
         updateInstanceGeometry();
+
+        if (selectedSubmissionId != -1) {
+            onFeatureClicked(selectedSubmissionId);
+        }
     }
 
     private void updateInstanceGeometry() {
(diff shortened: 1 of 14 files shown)


lognaturel:
Is this ready for review? If so, could you please update the checklist? Thanks!

grzesiek2010:
I was thinking about new tests to cover some new cases but testing bottomSheet with robolectric is so annoying...
For example i wanted to cover the case when we open the bottomSheet clicking on a marker then we click the back button what should close the bottomSheet, something like:

```
    @Test public void test() {
        MapPoint editableAndFinalized = new MapPoint(10.1, 125.6);
        int featureId = map.getFeatureIdFor(editableAndFinalized);
        activity.onFeatureClicked(featureId);

        assertThat(activity.summarySheet.getState(), is(BottomSheetBehavior.STATE_EXPANDED));
        activity.onBackPressed();
        assertThat(activity.summarySheet.getState(), is(BottomSheetBehavior.STATE_HIDDEN));
        activity.onBackPressed();
        assertThat(activity.isFinishing(), is(true));
    }
```
but `activity.summarySheet.getState()` always returns `BottomSheetBehavior.STATE_HIDDEN` even when I'm sure it is visible.
tricks like `sleep()` don't fix the problem.

then I wanted to confirm that the second click on the device back button finishes the activity so I added:
`assertThat(activity.isFinishing(), is(true));`
it works but the thing is that even if I call it before it also returns true. Robolectric what's wrong with you?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -211,9 +212,11 @@ public void initMap(MapFragment newMapFragment) {
         map.setFeatureClickListener(this::onFeatureClicked);
         updateInstanceGeometry();
 
-        if (selectedSubmissionId != -1) {
-            onFeatureClicked(selectedSubmissionId);
-        }
+        new Handler().postDelayed(() -> {
```
lognaturel: Use `Scheduler`

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
```diff
@@ -26,6 +26,7 @@
      * The count of all filled instances of this form, including unmappable ones.
      */
     private int totalInstanceCount;
+    private int selectedSubmissionId = -1;
```
lognaturel, line 29: Great, I had a note to request this change from previous commits. 😊 A further enhancement would be to use a `LiveData<Instance>` and to drive all of the bottom sheet behavior off of that. But I think this is fine.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -207,20 +248,26 @@ private void updateInstanceGeometry() {
...
 
             instancesByFeatureId.put(featureId, instance);
             points.add(point);
         }
     }
 
+    private void updateSubmissionMarker(int featureId, String status, boolean enlarged) {
+        int drawableId = getDrawableIdForStatus(status, enlarged);
+        map.setMarkerIcon(featureId, drawableId);
+        map.setMarkerAnchorToBottomCenter(featureId);
```
lognaturel: Setting the anchor should only need to happen on marker creation. In general, the API created here feels awkward to me though I know it's tricky because of the different ways the three mapping engines represent anchors. Instead of introducing `setMarkerAnchorToBottomCenter`, how about adding an `addMarker` variant? Mapbox's `Property.ICON_ANCHOR` is a good example of introducing a type for this (see https://guides.codepath.com/android/Replacing-Enums-with-Enumerated-Annotations for more about the pattern used there). I'm thinking `addMarker(MapPoint point, boolean draggable, @ICON_ANCHOR String iconAnchor)`. The `ICON_ANCHOR` annotation can be introduced in the `MapFragment` interface with just a single value for `BOTTOM` for now. Then each `MapFragment` can test for `BOTTOM` and do whatever they need to do to pass that on.

I like that approach better because I don't think we'd ever want the anchor changing after a marker has been added so it better captures the semantics. Most importantly, it avoids the exact duplicate `setMarkerAnchorToBottomCenter` implementations and the `instanceOf` calls. I also like that it's a bit more flexible in case we need different anchors in the future (unlikely).

lognaturel (changes requested):
Ahhh, this really makes it feel great. 😻

Few small things inline.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: Right, @seadowg? We should try not to have raw `Handler`s anymore so we can easily keep all those implementations consistent?

seadowg: It's not merged yet, so I think using `Handler` for the moment is best. I'm not quite sure why this code needs a `postDelayed` though?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
grzesiek2010: I agree your approach looks better. Thanks!
Done.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
grzesiek2010: It was a fix for https://github.com/getodk/collect/issues/3828#issuecomment-628611800 but now I realized  it's still not perfect so I removed this trick at all. I'll try to find a better solution or leave it as-is since it's not a big problem.

grzesiek2010: Ok I tried to figure out what's wrong but it's difficult. Since it's a small issue that affects only Mapbox and OSM let's create a separate issue to avoid blocking this pr:
image
 
@mmarciniak90 @kkrawczyk123 please keep it in mind.

lognaturel (approved):

kkrawczyk123:
I have noticed a small regression issue for Google maps and OSM that does not occur on master and for Mapbox. After clicking on markers one by one displayed details are moved slightly down. I was able to reproduce it on all devices, see attached gif: 
ezgif com-video-to-gif (6)
As it is probably connected to the issue from https://github.com/getodk/collect/issues/3828#issuecomment-628611800 and might have a similar cause so it might be good to investigate both together in another issue. Let me know @grzesiek2010 @lognaturel if you agree with me. I haven't noticed any other issues. Tested on Androids 5.1, 7.1, 8.1, 9.0 and 10.
Tested scenarios:
- Confirmed that issues from the original description have been fixed
- regression check:
 light mode
dark mode
screen rotation
minimize app
extinguish screen
Google maps
Mapbox
OSM
offline layer
all statuses: saved, finalized, sent and failed
delete answer
edit answer
change time and status after edition
zoom buttons
add submission through displayed map
changed language
encrypted form

grzesiek2010:
> I have noticed a small regression issue for Google maps and OSM that does not occur on master and for Mapbox. After clicking on markers one by one displayed details are moved slightly down.

I fixed this most common case when a user clicks on markers. When it comes to https://github.com/getodk/collect/issues/3828#issuecomment-628611800 as I said above it can be a separate issue.

kkrawczyk123:
@grzesiek2010 I have confirmed that it has been fixed for both Google and OSM (also no regression on Mapbox) but it has introduced another regression when clicking on markers one by one on OSM only - informations are not visible when tap on the second marker, see the gif: 
ezgif com-video-to-gif (7)

grzesiek2010:
@kkrawczyk123 ok please try now. Tricky...

kkrawczyk123:
**Tested with success!**

PR #3872, opened 2020-05-27
Exit form when instance file cannot be created/is null

seadowg:
Should fix crash found at https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/30fcda84f13db871dca574a1b2cba23f?time=last-seven-days&sessionId=5ECDE9E80054000132DF2A24E11F24F0_DNE_2_v2.

This also makes sure we never have a `null` `AuditEventLogger`.

Verified we see the error and form entry exits if instance file can't be created. Had to hardcode the failure to reproduce the case.

Discussed before (thanks to @grzesiek2010) for the pointers. I thought adding a specific error message might help us if someone posts this error on the forum.

I'm not sure of a way to reproduce this (which is why I didn't create an issue) but really this should just fix the crash. I think a quick pass on form loading would be good.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -153,19 +153,15 @@ import org.odk.collect.android.utilities.ToastUtils;
 import org.odk.collect.android.widgets.DateTimeWidget;
 import org.odk.collect.android.widgets.QuestionWidget;
 import org.odk.collect.android.widgets.RangeWidget;
-import org.odk.collect.utilities.Clock;
 
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 
 import javax.inject.Inject;
@@ -2394,7 +2390,18 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 }
 
                 if (formController.getInstanceFile() == null) {
-                    createInstanceDirectory(formController);
+                    FormInstanceFileCreator formInstanceFileCreator = new FormInstanceFileCreator(
+                            storagePathProvider,
+                            System::currentTimeMillis
+                    );
+
+                    File instanceFile = formInstanceFileCreator.createInstanceFile(formPath);
+                    if (instanceFile != null) {
+                        formController.setInstanceFile(instanceFile);
+                    } else {
+                        showFormLoadErrorAndExit(getString(R.string.form_instance_file_creation_error));
+                    }
+
                     formControllerAvailable(formController);
 
                     identityPromptViewModel.requiresIdentityToContinue().observe(this, requiresIdentity -> {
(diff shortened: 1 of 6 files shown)


grzesiek2010 (changes requested):
I think this pr doesn't fix the real cause. You handled the exception that occurs but as we discussed on slack  it was caused by null `instanceFile`. That field shouldn't be null when a user click on `menu_goto` icon. It indicates that something went wrong during loading the form. I'm pretty sure the problem stems from [createInstanceDirectory()](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L2483). That method should set `instanceFile` but it's not guaranteed for example
[FileUtils.createFolder(path)](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L2490) might return false somehow or [new File(path + File.separator + file + "_" + time + ".xml");](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L2491) might return null. In such a case we should force closing the form and display a toast for example that something went wrong. Otherwise the app probably will just crash in another place so we will have exactly the same number of crashes.

seadowg:
@grzesiek2010 thanks that makes a lot of sense. I'll stick this back into a draft and build on it 👍

seadowg:
@grzesiek2010 I think this gives us more protection. There is a lot more I'd like to rework but tried to keep this as minimal as possible so we can get it in.

collect_app/src/main/res/values/strings.xml
```diff
@@ -743,4 +743,5 @@
 
     <string name="install_id">Install ID - will replace Device ID in August 2020</string>
     <string name="preference_not_available">Not available</string>
+    <string name="form_instance_file_creation_error">Form instance file could not be created!</string>
```
lognaturel: Would you be open to using `loading_form_failed`: "An error occurred while loading the form. Please try again." That way we don't need to introduce a new string for a case we know is rare. I don't think that knowing the failure was about form instance file creation is actionable for users.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2394,7 +2390,18 @@ public void denied() {
...
+                    FormInstanceFileCreator formInstanceFileCreator = new FormInstanceFileCreator(
+                            storagePathProvider,
+                            System::currentTimeMillis
+                    );
+
+                    File instanceFile = formInstanceFileCreator.createInstanceFile(formPath);
+                    if (instanceFile != null) {
+                        formController.setInstanceFile(instanceFile);
+                    } else {
+                        showFormLoadErrorAndExit(getString(R.string.form_instance_file_creation_error));
```
lognaturel: Could you also do a `Timber.e("Error creating form instance file")` so we can see it in Crashlytics, keep an eye on how common it is, and possibly learn something from the logs?

lognaturel (changes requested):
Changes and tests look good. Couple of suggestions inline.

collect_app/src/main/res/values/strings.xml
seadowg: Do you disagree that having a specific error might help us if someone posts on the forum? Totally hear you on it not being actionable but neither really is a generic "try again" if it's a device/form specific issue.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah good call!

grzesiek2010 (approved):

grzesiek2010:
LGTM.

collect_app/src/main/res/values/strings.xml
lognaturel: I think it’s very unlikely a user would post. If one does, we can usually track down their crashlytics trace from their device info. So I do prefer something generic.

mmarciniak90:
**Tested with success**

**Verified cases:**
- downloading forms from Google Drive
- downloading forms from Aggregate
- filled forms from Basic criteria part of regression testing

lognaturel (approved):

kkrawczyk123:

PR #3875, opened 2020-05-27
Catch exceptions that might take place while customizing a toast

### LINKED ISSUE #3874: ClassCastException in ToastUtils.java:55
Collect v1.27.0 and older versions

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/ea33a1b05089575438f325731c00f23c?time=last-seven-days&sessionId=5ECDEA5A0327000110B59C46659938BA_DNE_0_v2

```
Fatal Exception: java.lang.ClassCastException: android.widget.LinearLayout cannot be cast to android.widget.TextView
       at org.odk.collect.android.utilities.ToastUtils.showToastInMiddle(ToastUtils.java:55)
       at org.odk.collect.android.utilities.ToastUtils.showShortToastInMiddle(ToastUtils.java:45)
       at org.odk.collect.android.activities.FormEntryActivity.createConstraintToast(FormEntryActivity.java:1700)
       at org.odk.collect.android.activities.FormEntryActivity.saveAnswersForCurrentScreen(FormEntryActivity.java:1075)
       at org.odk.collect.android.activities.FormEntryActivity.saveBeforeNextView(FormEntryActivity.java:1494)
       at org.odk.collect.android.activities.FormEntryActivity.showNextView(FormEntryActivity.java:1437)
       at org.odk.collect.android.activities.FormEntryActivity.lambda$onCreate$2(FormEntryActivity.java:371)
       at org.odk.collect.android.activities.-$$Lambda$FormEntryActivity$Y_WgNY5L9k5YQmo4By-nYVVFXUI.onClick(-.java:2)
```

grzesiek2010:
Closes #3874 

Just reviewed implemented changes.

Looks like on some devices the approach we use to customize toasts might cause exceptions  when we try to change font size because `group.getChildAt(0);` doesn't return TextView but LinearLayout. Since it''s rare case we can just catch it and keep the default font size when it happens.

Another possible solution would be to create our custom toasts https://medium.com/@francis4dblues/custom-toast-views-in-android-1e1e7352351c
but I think it doesn't make sense in this case because:
1. The issue is rare
2. Even if the exception takes place and we catch it, it will work but with default font size so not a big problem.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/ToastUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/ToastUtils.java
@@ -51,10 +51,14 @@ public class ToastUtils {
 
     private static void showToastInMiddle(String message, int duration) {
         Toast toast = Toast.makeText(Collect.getInstance(), message, duration);
-        ViewGroup group = (ViewGroup) toast.getView();
-        TextView messageTextView = (TextView) group.getChildAt(0);
-        messageTextView.setTextSize(21);
-        messageTextView.setGravity(Gravity.CENTER);
+        try {
+            ViewGroup group = (ViewGroup) toast.getView();
+            TextView messageTextView = (TextView) group.getChildAt(0);
+            messageTextView.setTextSize(21);
+            messageTextView.setGravity(Gravity.CENTER);
+        } catch (Exception ignored) {
+            // ignored
+        }
 
         toast.setGravity(Gravity.CENTER, 0, 0);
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel:
Seems like the right call, thanks! We definitely don't want folks crashing for font scaling.

PR #3877, opened 2020-05-27
Reworking GeoPoint Widget

SaumiaSinghal:
This PR is the work on GeoPoint Widget.

How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
No

Do we need any specific form for testing your changes? If so, please attach one.
All Widgets form, or any form with the geo widgets.


collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,27 +1,41 @@
...
+    public void usingReadOnlyOption_makesAllClickableElementsDisabled() {
+        GeoPointWidget widget = createWidget(promptWithReadOnly());
+        assertThat(widget.findViewById(R.id.simple_button).getVisibility(), equalTo(View.GONE));
+    }
+
+    private GeoPointWidget createWidget(FormEntryPrompt prompt) {
+        return new GeoPointWidget(widgetTestActivity(), new QuestionDetails(prompt, "formAnalyticsID"));
+    }
+}
+
```
SaumiaSinghal, line 75: Hi @seadowg! I had some doubts. As here, `GeoPointWidgetTest` does not inherit from any other tests. Should I go about it, or maintain the already implemented hierarchy. Also, I wanted to ask, is there any drawback when running tests having some hierarchy (in terms of running time, or some other factor)?

seadowg: I think we should go for the tests not having a hierarchy i.e. `GeoPointWidgetTest` should not extend anything.

As far as implementation goes I think it's a harder decision. I think a good way to start would be to rewrite the tests for these widgets (`GeoPointWidget`, `GeoShapeWidget` and `GeoTraceWidget`) as one PR. Then we can discuss how to rewrite the widgets using the new "framework".

SaumiaSinghal:
Hi @seadowg! I added unit tests for `GeoPointWidget`. But, unit test `buttonsShouldLaunchCorrectIntents()` is failing. I'm looking into it.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,217 @@
...
+        stubAllRuntimePermissionsGranted(widget, true);
+        (widget.findViewById(R.id.geo_button)).performClick();
+
+        Intent startedIntent = shadowOf(widgetTestActivity()).getNextStartedActivity();
+        Intent expectedIntent = new Intent(widgetTestActivity(), GeoPointActivity.class);
+
+        expectedIntent.putExtra(READ_ONLY, false);
+        expectedIntent.putExtra(DRAGGABLE_ONLY, false);
+        expectedIntent.putExtra(ACCURACY_THRESHOLD, DEFAULT_LOCATION_ACCURACY);
+        assertEquals(startedIntent, expectedIntent);
```
seadowg: Instead of building an `expectedIntent` here I would just assert that the `startedIntent` has the correct component and extras. It makes it easier to see what is missing when there is a fail and I'm not sure `equals` on Intent is very good. Are you familiar with how equality works in Java (`equals` vs `==` etc)?

SaumiaSinghal: By building a new Intent, I might get an intent with same component name and extra values, but the `nextStartedActivity()` will not return the same intent. I think that's why the test is failing.

SaumiaSinghal:
Hi @seadowg! I have fixed the red unit tests, should I start writing the unit tests for other GeoWidgets? Also I feel that a lot of code in the unit test of all the geo widgets would be repetitive, so maybe it would be better if there is a class similar to `QuestionWidgetHelper`

seadowg:
@SaumiaSinghal haven't had a chance to go into a full review of the tests but I think the idea of having a helper is good! I think for this PR we should just focus on `GeoPointWidget` and move onto the others after.

One thing that would be interesting to investigate as part of this is if you are able to make `GeoPointWidget` inherit from `QuestionWidget` instead of `BaseGeoWiget`. This might mean moving some code to shared helpers when we rework the other geo widgets so we're not repeating ourselves but I'm interested to see if it's doable. If you think it's a bad idea/too awkward let's discuss though! I think a good place to start would be to change the base class, see what functionality is missing and then write tests for that functionality for the `GeoPointWidget`.

SaumiaSinghal:
Oh okay @seadowg! In future, if we want to remove `BaseGeoWidget`, then, it would be better to check if we can get the `GeoPointWidget` not inherit `BaseGeoWidget`.
I will try that. As far as the helpers are concerned, I think we can transfer the repetitive code late after we are done with all the `GeoWidgets`

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,248 @@
...
+
+    @Test
+    public void whenWidgetHasAppearanceAndNullAsAnswer_buttonShowsCorrectText() {
+        when(mapConfigurator.isAvailable(any())).thenReturn(true);
+        GeoPointWidget widget = createWidget(promptWithAppearanceAndAnswer(PLACEMENT_MAP, null));
+        assertThat(widget.startGeoButton.getText(), equalTo(widget.getContext().getString(R.string.get_point)));
+    }
+
+    @Test
+    public void whenWidgetHasAppearanceAndAnswer_buttonShowsCorrectText() {
```
seadowg: I think it's be good to write tests for the different appearances here. Right now they show the same thing but I think it'd be nice. The behaviour we think about from the outside is "if it has placement-map it's this and if it's maps it's this". 

Try to write the tests you want to describe the behaviour rather than writing tests that match the current implementation! This is definitely a lot easier when writing new code than backfilling tests.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,248 @@
...
+import static org.robolectric.Shadows.shadowOf;
 
 /**
  * @author James Knight
  */
 
-@Config(shadows = {ShadowPlayServicesUtil.class})
-public class GeoPointWidgetTest extends BaseGeoWidgetTest<GeoPointWidget, GeoPointData> {
+@RunWith(RobolectricTestRunner.class)
+public class GeoPointWidgetTest {
```
seadowg, line 48: I think we're missing tests for `setOnLongClickListener` and `cancelLongPress`. Should be very similar to the `URLWidget` ones I'd imagine!

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,248 @@
...
+    @Test
+    public void whenWidgetHasNullAsAnswer_buttonShowsCorrectText() {
+        GeoPointWidget widget = createWidget(promptWithAnswer(null));
+        assertThat(widget.startGeoButton.getText(), equalTo(widget.getContext().getString(R.string.get_point)));
+    }
+
+    @Test
+    public void whenPromptDoesNotHaveAnswer_buttonShouldLaunchCorrectIntent() {
+        GeoPointWidget widget = createWidget(promptWithAnswer(null));
+        stubAllRuntimePermissionsGranted(widget, true);
```
seadowg: This method should really be called `stubLocationPermissions`. It's probably also good to have a test that checks what happens when location permissions are denied.

seadowg (changes requested):
I've left some comments inline. The tests are looking great so far though! 

I noticed when playing with this that the GeoPointWidget does not update when an answer is set. In fact, currently `FormMap#fillingBlankForm_addsInstanceToMap` fails because of this.

I've pushed a refactor of how widgets get data from external Activity's (like the geo widgets do) to make it easier to understand what's going on here. To have `GeoPointWidget` receive the location data from `GeoPointActivity` the following needs to be in place:

* `GeoPointWidget` must implement `BinaryDataReciever`
* `GeoPointWidget` must call `waitForData()` with its form index on the `WaitingForDataRegistry` that is passed into `WidgetFactory.createWidget` (you should be able to pass it into the constructor).

You can see examples of this in `ExStringWidget`. As always I'd suggest writing failing tests for these requirements before changing the `GeoPointWidget`. When you're done `FormMap#fillingBlankForm_addsInstanceToMap` should pass!

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
SaumiaSinghal: Oh okay @seadowg. Yes, I'll change the tests

SaumiaSinghal:
Hi @seadowg! Thanks for pointing out the bug and for guiding me on how to get it fixed. I checked the Espresso test too, that we talked about yesterday. It is green now.

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
```diff
@@ -41,118 +54,172 @@
...
-            } else {
-                startGeoButton.setText(
-                    dataAvailable ? R.string.view_change_location : R.string.get_point);
-            }
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        ViewGroup answerView = (ViewGroup) LayoutInflater.from(context).inflate(R.layout.base_geo_widget_layout, null);
+
+        answerDisplay = answerView.findViewById(R.id.geo_answer_text);
+        answerDisplay.setTextColor(new ThemeUtils(context).getColorOnSurface());
```
seadowg: This can be defined in the layout

seadowg (dismissed):
Just one little comment. I think after feel free to mark it ready for review and someone else can have a go at a final review!

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
SaumiaSinghal: Ahhh...Sorry I forgot

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/FormControllerWaitingForDataRegistry.java
```diff
@@ -0,0 +1,53 @@
+package org.odk.collect.android.widgets.utilities;
+
+import org.javarosa.core.model.FormIndex;
+import org.odk.collect.android.application.Collect;
+import org.odk.collect.android.javarosawrapper.FormController;
+
+public class FormControllerWaitingForDataRegistry implements WaitingForDataRegistry {
```
lognaturel, line 7: I don't think there's any reason for the `FormController` to be involved in this. I think that a single `WaitingForDataRegistry` object that keeps track of the state itself and is tied to the `FormEntryActivity` lifecycle would help reduce what the `FormController` deals with. Doesn't have to be done here but might be a nice next step if I'm not missing anything.

collect_app/src/test/java/org/odk/collect/android/widgets/support/FakeWaitingForDataRegistry.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.javarosa.core.model.FormIndex;
+import org.odk.collect.android.widgets.utilities.WaitingForDataRegistry;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FakeWaitingForDataRegistry implements WaitingForDataRegistry {
+
+    public List<FormIndex> waiting = new ArrayList<>();
```
lognaturel, line 11: This doesn't much matter for tests but note that there's only ever one widget waiting for data.

collect_app/src/main/java/org/odk/collect/android/tasks/MediaLoadingTask.java
```diff
@@ -97,7 +97,7 @@ protected void onPostExecute(File result) {
 
         ODKView odkView = formEntryActivity.get().getCurrentViewIfODKView();
```
lognaturel: This null check seems unrelated now that the data is set by the activity. `FormEntryActivity.setBinaryWidgetData` does that check anyway so I think it's clearer to remove it.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,301 @@
...
+        waitingForDataRegistry = mock(WaitingForDataRegistry.class);
+        when(questionDef.getAdditionalAttribute(anyString(), anyString())).thenReturn(null);
     }
 
-    @NonNull
-    @Override
-    public GeoPointWidget createWidget() {
-        return new GeoPointWidget(activity, new QuestionDetails(formEntryPrompt, "formAnalyticsID"));
+    @Test
+    public void usingReadOnlyOption_makesAllClickableElementsDisabled() {
```
lognaturel: Test and name don't match -- we decided the button should be gone, right? So the test name should reflect that.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,301 @@
...
+        GeoPointWidget widget = createWidget(promptWithAnswer(null));
+        stubLocationPermissions(widget, false);
+        widget.startGeoButton.performClick();
+        Intent startedIntent = shadowOf(widgetTestActivity()).getNextStartedActivity();
+
+        assertNull(startedIntent);
+    }
+
+    @Test
+    public void whenPermissionIsGranted_buttonClickShouldSetTheLocationSavedOnTheMap() {
```
lognaturel: A more accurate name might be `buttonClickLaunchesIntentAndWaitsForLocationData`.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,301 @@
...
+    public void whenPromptHasAnswer_buttonShouldLaunchCorrectIntent() {
+        GeoPointWidget widget = createWidget(promptWithAnswer(new StringData(answer)));
+        stubLocationPermissions(widget, true);
+        widget.startGeoButton.performClick();
+
+        Intent startedIntent = shadowOf(widgetTestActivity()).getNextStartedActivity();
+        Bundle bundle = startedIntent.getExtras();
+
+        assertThat(startedIntent.getComponent(), equalTo(new ComponentName(widgetTestActivity(), GeoPointActivity.class)));
+        assertBundleArgumentEquals(bundle, GeoWidgetUtils.getLocationParamsFromStringAnswer(answer), DEFAULT_LOCATION_ACCURACY, false, true);
```
lognaturel: Would be good to also set a non-default accuracy threshhold in the form entry prompt and verify that goes through to the intent.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,85 +1,301 @@
...
+        stubLocationPermissions(widget, true);
+        widget.startGeoButton.performClick();
+
+        Intent startedIntent = shadowOf(widgetTestActivity()).getNextStartedActivity();
+        Bundle bundle = startedIntent.getExtras();
+
+        assertThat(startedIntent.getComponent(), equalTo(new ComponentName(widgetTestActivity(), GeoPointMapActivity.class)));
+        assertBundleArgumentEquals(bundle, null, DEFAULT_LOCATION_ACCURACY, false, false);
+    }
+
```
lognaturel, line 198: I think there's a test missing here for if there's a map and it's read-only.

lognaturel (changes requested):
Big 'yes' to:
- the new geopoint tests
- having `GeoPointWidget` extend `QuestionWidget` rather than `BaseGeoWidget`
- the change from `BinaryWidget` to `BinaryDataReceiver` and taking that out of the `Widget` hierarchy.
- moving all things about waiting for data out of `ODKView`

I think that sets things up really nicely for looking at the other geo widgets and the binary/button widget hierarchies.

I'm kind of left wondering whether things would be easier to understand if there were two geopoint widgets: one with a map and one without. I particularly found it hard to read all the test cases and keep track of what was covered since there's so much branching. Is that something you considered, @SaumiaSinghal?

I marked a few specific things inline. I'd also love to see all the `findViewById` calls gone. Also please don't forget https://github.com/getodk/collect/pull/3877/files#r438930384!

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
SaumiaSinghal: Ohh totally forgot that!

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
SaumiaSinghal: Right sorry! I will correct it.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
SaumiaSinghal: Sorry @lognaturel I'm not understanding how I would confirm here if the map is read only.

SaumiaSinghal:
hi @lognaturel! As you have pointed out now that it might be better to have two GeoPoint widgets with and without maps appearance. I also feel that there is a lot of branching in some of the methods. It would be good to remove that to make things easier to understand and then, even it would be easier to write up the tests and not many tests would be accommodated for a single class. Also, if I go about it, most of the code for both the classes would remain the same. But as at the end, I need to extract out a helper class, for the three widgets, so I think that part of the redundancy of code would be dealt. :)  

I can write up the separate classes and then if the code is not looking good, I'll remove the commit and force push?

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
```diff
@@ -18,152 +18,183 @@
...
-                    GeoWidgetUtils.convertCoordinatesIntoDegreeFormat(getContext(), Double.parseDouble(parts[1]), "lon"),
-                    GeoWidgetUtils.truncateDouble(parts[2]),
-                    GeoWidgetUtils.truncateDouble(parts[3])
-            );
-        } else {
+    protected GeoWidgetAnswerBinding getBinding() {
+        return binding;
+    }
+
+    private void determineMapProperties(QuestionDef questionDef) {
```
lognaturel: There's no map so the name isn't right. I think that here the code can either go directly in the constructor or be renamed to something like `getAccuracyThreshold`. If it's the later, then I think the method should return a `double`.

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
```diff
@@ -18,152 +18,183 @@
...
-        Intent intent = new Intent(
-            context, useMap ? GeoPointMapActivity.class : GeoPointActivity.class);
+        Intent intent = new Intent(context, GeoPointActivity.class);
 
         if (stringAnswer != null && !stringAnswer.isEmpty()) {
             intent.putExtra(LOCATION, GeoWidgetUtils.getLocationParamsFromStringAnswer(stringAnswer));
         }
         intent.putExtra(READ_ONLY, readOnly);
-        intent.putExtra(DRAGGABLE_ONLY, draggable);
+        intent.putExtra(DRAGGABLE_ONLY, false);
```
lognaturel: I think this can just be removed.

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointMapWidget.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        binding = GeoWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        View answerView = binding.getRoot();
+
+        binding.geoAnswerText.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+
+        readOnly = prompt.isReadOnly();
+        if (readOnly) {
+            binding.simpleButton.setVisibility(GONE);
```
lognaturel: I think this changes the behavior that was there before. Before your changes I see:
```
if (useMap) {
            if (readOnly) {
                startGeoButton.setText(R.string.geopoint_view_read_only);
```

which implies that if the question is read-only, the map should open and show the point but not allow it to be changed. Can you verify what the prior behavior was with a readonly geopoint with map and a default value, make sure it's still the same, and that there's a test for it?

This is the case I was referring to at https://github.com/getodk/collect/pull/3877/files#r449199637.

lognaturel (changes requested):
Looking good! There's redundancy, but as you say, you'll do a later pass to take care of that. Couple small things inline and some failing tests!

lognaturel (changes requested):
Looking good! There's redundancy, but as you say, you'll do a later pass to take care of that. Couple small things inline and some failing tests!

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointMapWidget.java
SaumiaSinghal: I will confirm @lognaturel! I was also confused about this behavior here

SaumiaSinghal: It seems like having the prompt `readOnly` does not set the visibility of the `button` to `View.GONE`. Thanks for pointing this out, I'll revert it.

SaumiaSinghal:
Hi @lognaturel! I verified the behavior on my device for `GeoPointWidget` on `master` branch. When using `readOnly` option, the button visibility is set to `View.GONE`.  However, it does not seem nice, as then this part of code does not make sense.
```
if (useMap) {
            if (readOnly) {
                startGeoButton.setText(R.string.geopoint_view_read_only);
```
Also, I feel that the user should be allowed to view the location on the maps (if maps are available) if the prompt is `readOnly`, and in case when the map view is not available, the button can just be disabled, as it might not look good to show just the answer in the answer textView, without showing any button.

|   |  |
| WhatsApp Image 2020-07-09 at 23 32 24 (1)  | WhatsApp Image 2020-07-09 at 23 32 24  |

I used this form for testing: [TestingGeoPointWidget.xml.txt](https://github.com/getodk/collect/files/4898685/TestingGeoPointWidget.xml.txt)

lognaturel:
> When using readOnly option, the button visibility is set to View.GONE

I see, I wondered whether that might be the case. I think something got broken in the hierarchy along the way. In general, we shouldn't change the behavior of widgets while doing the refactors unless we have to for some reason or it's a bug fix. In this case allowing a view-only map clearly seems like a clear bug fix so let's do it. It's a good opportunity to add a test to make sure it doesn't regress again.

> the button can just be disabled, as it might not look good to show just the answer in the answer textView

We recently agreed on removing the button across the board when it can't be used so that should stay the same. The big advantages that it reduces the vertical space taken by the question and cuts down on visual noise.

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointMapWidget.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private void startGeoActivity() {
+        Context context = getContext();
+        Intent intent = new Intent(context, GeoPointMapActivity.class);
+
+        if (stringAnswer != null && !stringAnswer.isEmpty()) {
+            intent.putExtra(LOCATION, GeoWidgetUtils.getLocationParamsFromStringAnswer(stringAnswer));
+        }
+        intent.putExtra(READ_ONLY, readOnly);
```
lognaturel, line 199: Note here that the `GeoPointMapActivity` accepts an intent extra to represent whether the map should be readonly. So it's very clear that the intended behavior was to show a readonly map.

collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
```diff
@@ -18,152 +18,174 @@
...
+    private void startGeoActivity() {
         Context context = getContext();
-        Intent intent = new Intent(
-            context, useMap ? GeoPointMapActivity.class : GeoPointActivity.class);
+        Intent intent = new Intent(context, GeoPointActivity.class);
 
         if (stringAnswer != null && !stringAnswer.isEmpty()) {
             intent.putExtra(LOCATION, GeoWidgetUtils.getLocationParamsFromStringAnswer(stringAnswer));
         }
         intent.putExtra(READ_ONLY, readOnly);
```
SaumiaSinghal: @lognaturel Should I remove `intent.putExtra(READ_ONLY, readOnly)` from here as well? In case of `readOnly` prompt, the button visibility is set to `View.GONE`. So, the user will not be launching any intent in that case.

lognaturel: Good catch, yes!

lognaturel (approved):

mmarciniak90:
@SaumiaSinghal @lognaturel 
I was a bit worried about showing on the map the readOnly point, that's why I looked at it more closely and I noticed a case where I'm able to edit this point.

1. I used _TestingGeoPointWidget.xml_ form. First of all, I edited second question in XML-> readonly="false()"
2. I filled form and added a point for the second question -> I saved the form
3. Then I again edited second question in XML -> readonly="true()"
4. I opened the saved form and opened the second question with the location.
5. I opened the map view and **I was able to remove the point, add a new point**

Link to gif: https://drive.google.com/file/d/1-bvS-2T0MkrQ8K6IoKOkq1HkPieI3pyM/view?usp=sharing

SaumiaSinghal:
Hi @mmarciniak90! Thanks for reporting the issue, Can you share the form please?

SaumiaSinghal:
Hi @mmarciniak90! I didn't fully understand it. I think the form that is saved will be the copy of the form that has readOnly="false()", so, we can change the value. When I'm trying this case with two different forms `TestingGeoPointWidget.xml`, it doesn't allow to save form with the same name. And if I'm replacing the form, then it gets replaced just for the Blank Forms, and gives me an error when I try to access the same form in 'Saved Forms' saying 
`Unable to edit this saved form because the corresponding blank form was not present or was deleted.`

mmarciniak90:
@SaumiaSinghal sorry, I feel I make this case too complicated. There is a simpler way to check it.
Take a look at this form [TestingGeoPointWidgetWithValues.xml.txt](https://github.com/getodk/collect/files/4954329/TestingGeoPointWidgetWithValues.xml.txt). The second question is readOnly with a point but you are able to edit this point.

SaumiaSinghal:
@mmarciniak90 I tried with the attached form too, but I was not able to edit the `readOnly` point with Blank form or Saved Form. Sorry for not understanding the case. I'll wait for @lognaturel too, to have a look.

https://drive.google.com/file/d/1cp7po9WMfJMpVSIV2Y4MXNBxPoK8aeJV/view?usp=sharing

mmarciniak90:
@SaumiaSinghal you need to click on **trash** icon on a map view. Then the first icon will be available.

SaumiaSinghal:
Hi @mmarciniak90 I solved the issue. It was a simple fix, sorry for taking time. Can you test again?

mmarciniak90:

kkrawczyk123:
@lognaturel I have one more concern about the read-only question for geopoint with map view. Shouldn't the View GeoPoint button be actually hidden when there is no point added? When the point is added it makes perfect sense to view it but when there is no point it might actually give a wrong impression that user can actually interact with the map, screenshots:
| GeoPoint without map read-only | GeoPoint with map read-only NO point | GeoPoint with map read-only point added |
| -------- | ------- | ------- |
| Screenshot_2020-07-22-14-09-34 | Screenshot_2020-07-22-14-10-24 | Screenshot_2020-07-22-14-11-22 | 

I was initially thinking to report it separately but we may also encounter the same thing after reworking GeoTrace and GeoShape widgets.

SaumiaSinghal:
@kkrawczyk123 Thanks for the suggestion. I fixed the visibility of the button in case the prompt is readOnly and answer is available.

lognaturel:
One could argue that including a blank read-only value in a form design is a little bit 🤯🤯🤯🤪🤪🤪 but thanks @SaumiaSinghal for handling that edge case with tests and to @getodk/testers for being so detail-oriented!!

SaumiaSinghal:
HAHA No worries @lognaturel!

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- geowidget without map
- geowidget with map
- scroll on widget
- text on the button is changed when a location is set
- removed response
- Google map, Mapbos, OSM
- offline layer
- light mode, dark mode
- read-only question
- regression check

mmarciniak90:
Tested with success also on Android versions: 7.0, 8.1, 9.0

lognaturel:
High fives all around! ✋

PR #3885, opened 2020-06-01
Fixed google bug 154855417

### LINKED ISSUE #3884: Crash when Google Maps goes down
#### Problem description

It looks like [this crash](https://console.firebase.google.com/u/a0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/71b0f8dfadb8435206e57c3b072dcb92?time=last-ninety-days&sessionId=5ED447A5006F00014CFC40CB8EDAC83C_DNE_0_v2) is occurring when there is an outage on the Google Maps side. 

Google have created a fix and deployed in the new Play Services but have also posted workaround code to prevent the crash on devices with older Play Services: https://issuetracker.google.com/issues/154855417#comment509. We should probably incorporate this in our code.

Not really possible as it's an internal Google Maps SDK problem.

#### Expected behavior

The app does not crash if the Google Maps API is having problems.

grzesiek2010:
Closes #3884

Not much since we can't reproduce it. I just prepared my changes according to https://issuetracker.google.com/issues/154855417#comment509

I implemented changes according to the explanation provided here (as I said above): https://issuetracker.google.com/issues/154855417#comment509
I adjusted the sample fix for android keeping just the fix for maps v2 and using our shared preferences.

We are not able to reproduce the issue and also testing doesn't make sense.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/Collect.java
@@ -47,6 +47,7 @@ import org.odk.collect.android.preferences.AdminSharedPreferences;
 import org.odk.collect.android.preferences.AutoSendPreferenceMigrator;
 import org.odk.collect.android.preferences.FormMetadataMigrator;
 import org.odk.collect.android.preferences.GeneralSharedPreferences;
+import org.odk.collect.android.preferences.MetaSharedPreferencesProvider;
 import org.odk.collect.android.preferences.PrefMigrator;
 import org.odk.collect.android.storage.StoragePathProvider;
 import org.odk.collect.android.tasks.sms.SmsNotificationReceiver;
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -91,6 +92,9 @@
     @Inject
     public CollectJobCreator collectJobCreator;
 
+    @Inject
+    public GeneralSharedPreferences generalSharedPreferences;
```
seadowg: This should be using `MetaSharedPreferencesProvider` instead. That way we don't need to use `KEYS_WE_SHOULD_NOT_RESET`.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -150,6 +154,7 @@ public void onCreate() {
         singleton = this;
 
         setupDagger();
+        fixGoogleBug154855417();
```
seadowg, line 158: Probably nice to add a comment with a link to the issue given this is a weird one!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
grzesiek2010: Done.

seadowg (approved):

seadowg:
@grzesiek2010 seeing some weird issues on Circle CI. Could you try rerunning "from start" to see if it fixes it?

grzesiek2010:
> seeing some weird issues on Circle CI. Could you try rerunning "from start" to see if it fixes it?

it's green.

Issue #3826, opened 2020-05-13
Unable to Save to Drive/Attach to Mail QR Code on Android 5.1 Huawei Y560-L01


kkrawczyk123:
#### Software and hardware versions 
Collect v1.26.x, Android v5.1.x, device used: Huawei Y560-L01

#### Problem description
I have noticed a problem with sharing QR code on my device with Android 5.1.
Google drive seem to crash when try to save file with QR code on Huawei Y560-L01 device. 
stacktrace:
`05-13 12:44:55.935 11606-11606/? E/CAKEMIX_CRASHED: [main] java.lang.SecurityException: Uid 10050 does not have permission to uri 0 @ content://org.odk.collect.android.provider/external_files/Android/data/org.odk.collect.android/files/settings/collect-settings.png
        at android.os.Parcel.readException(Parcel.java:1546)
        at android.os.Parcel.readException(Parcel.java:1499)
        at android.app.ActivityManagerProxy.startActivity(ActivityManagerNative.java:2448)
        at android.app.Instrumentation.execStartActivity(Instrumentation.java:1496)
        at android.app.Activity.startActivityForResult(Activity.java:3794)
        at ci.startActivityForResult(PG:3)
        at ltq.startActivityForResult(PG:2)
        at android.app.Activity.startActivityForResult(Activity.java:3755)
        at ci.startActivityForResult(PG:1)
        at ltq.startActivityForResult(PG:1)
        at android.app.Activity.startActivity(Activity.java:4065)
        at ltq.startActivity(PG:2)
        at android.app.Activity.startActivity(Activity.java:4033)
        at ltq.startActivity(PG:1)
        at hjz.onClick(PG:34)
        at android.support.v7.app.AlertController$b.handleMessage(PG:3)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:135)
        at android.app.ActivityThread.main(ActivityThread.java:5298)
        at java.lang.reflect.Method.invoke(Native Method)
        at java.lang.reflect.Method.invoke(Method.java:372)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:911)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:706)
05-13 12:44:56.124 831-857/? E/JavaBinder: !!! FAILED BINDER TRANSACTION !!!
[...]

PR #3888, opened 2020-06-01
Fixed Failed to find configured root that contains on huawei devices

### LINKED ISSUE #3887: Failed to find configured root that contains /storage/emulated/0/Android/data/org.odk.collect.android/files/settings/collect-settings.png
Collect v1.27.0

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/d9c8a9db66823efa64a701b76295e1e3?time=last-seven-days&sessionId=5ED1DA00035000013E521B1B631AD624_DNE_0_v2

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.odk.collect.android/org.odk.collect.android.preferences.qr.QRCodeTabsActivity}: java.lang.IllegalArgumentException: Failed to find configured root that contains /storage/emulated/0/Android/data/org.odk.collect.android/files/settings/collect-settings.png
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2969)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3047)
       at android.app.ActivityThread.-wrap11(ActivityThread.java)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1777)
       at android.os.Looper.loop(Looper.java:166)
       at android.app.ActivityThread.main(ActivityThread.java:6861)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:450)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

grzesiek2010:
Closes #3887 

Not much because I 'm not able to reproduce the issue. I just confirmed there is no regression on devices I have.

It's a bug that occurs on huawei devices. I implemented a fix according to: https://stackoverflow.com/a/41309223/5479029

it might be difficult to test becuase we might reproduce the bug only on Huawei devices. If we are not able to reproduce we need just confirm that nothing wrong happens when we click `Configure via QE code` and we can scane/share codes.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/ConfigureWithQRCodeTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/preferences/qr/ConfigureWithQRCodeTest.java
@@ -8,7 +8,6 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.net.Uri;
 
-import androidx.core.content.FileProvider;
 import androidx.test.espresso.intent.Intents;
 import androidx.test.espresso.intent.rule.IntentsTestRule;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
@@ -30,6 +29,7 @@ import org.odk.collect.android.injection.config.AppDependencyModule;
 import org.odk.collect.android.support.ResetStateRule;
 import org.odk.collect.android.support.RunnableRule;
 import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.android.utilities.ContentUriProvider;
 
 import java.io.File;
 import java.io.FileOutputStream;
@@ -52,7 +52,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.not;
 import static org.junit.Assert.assertTrue;
-
 
 @RunWith(AndroidJUnit4.class)
 public class ConfigureWithQRCodeTest {
@@ -132,7 +131,7 @@ public class ConfigureWithQRCodeTest {
     @Ignore("Should be replaced at Robolectric level")
     public void onMainMenu_clickConfigureQRCode_andClickingOnShareQRCode_startsExternalShareIntent() {
         String path = new StubQRCodeGenerator().getQrCodeFilepath();
-        Uri expected = FileProvider.getUriForFile(getApplicationContext(),
+        Uri expected = ContentUriProvider.getUriForFile(getApplicationContext(),
                 BuildConfig.APPLICATION_ID + ".provider",
                 new File(path));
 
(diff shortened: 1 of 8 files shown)


grzesiek2010:
For reviewers:
as I mentioned I used the fix provided in https://stackoverflow.com/a/41309223/5479029 but I adjusted it a bit using our cache directory instead of `context.getCacheDir()` like in the original solution.
So looks as if I can skip the second part of the solution (updating our provider_paths.xml file) am I right, or maybe I'm missing something about configuring file provider?

seadowg (approved):
Yuck. Huawei have really not made this pleasant. I'd really like to verify this on a Huawei device before we merge if possible. Do @getodk/testers have any?

kkrawczyk123:
@seadowg jup, I will try it on Huawei Y560-L01.

grzesiek2010:
> I will try it on Huawei Y560-L01.

If you had been able to report this issue https://github.com/getodk/collect/issues/3826 that means the crash doesn't occur because it would happen earlier when you click `Configure via QR code`. 

I'll ask on slack or maybe even on the forum. Maybe someone will help us.

grzesiek2010:
@lognaturel @seadowg 
I'm back to this issue because there is one thing I'm thinking about...
the code that causes crashes on some Huawei devices (`FileProvider.getUriForFile()`) is used in a few different places in our codebase:
- taking photos
- playing videos
- opening files via the file widget

but I can't see reports about crashes in these functionalities. However I would go ahead and use this fix in all of the places just in case. What do you think?

**[Edit] ahh ok I can't see reports because we catch `IllegalArgumentException` in most of those places which extends `RuntimeException` thrown by Huawei devices. Ok so I think we should handle all the other places as well.**

grzesiek2010:
Ok I added changes to handle all cases. If you are strongly against we can get rid of it but I'm pretty sure it's a good solution.

seadowg:
@grzesiek2010 makes sense to me. I think keeping the error handling inside the abstraction is the right thing to do here. In #3886 I converted `FileProvider` to an interface (I can deal with the conflict there) so we could even call out in the implementing class that this deals with Huawei problems (`HuaweaiSafeContentURProvider` example).

mmarciniak90:
**Tested with success**

Verified on devices:
* Sony Xperia Z3 D6603 - Android 6.0.1
* Samsung Galaxy S7 SM-G930F - Android 7.0.0
* LG Nexus 5X - Android 8.1
* Xiaomi Redmi 7 - Android 9.0

Verified cases:
* confirm that nothing wrong happens when I click `Configure via QR code` 
* scan code
* share code
* QR code with admin password
* QR code functionality from Admin Settings

mmarciniak90:
@getodk-bot@getodk-bot

mmarciniak90:

PR #3899, opened 2020-06-03
Save current screen answers when quitting form

seadowg:
This was the reason the Track Changes Reason tests were failing.

Added new tests to make a bug in this area a little more obvious (hopefully) and ran tests to check we're now green.

I think really we need to look at moving the screen answer saving out of the FormEntryActivity and into the FormSaveViewModel. I didn't want to get into that to get green again so just used a listener on QuitFormDialogFragment to call the FormEntryActivity's save logic. 

I think it's a good step forward to have some Espresso tests covering this flow a little better however.

I think it's best to do a straight merge on this as it is mostly a step backwards and gets us green again. 


lognaturel (approved):
Yikes. Well, it's really good additional test coverage to have, certainly.

PR #3910, opened 2020-06-05
Implemented viewmodel for RankingWidgetDialog to solve the problem with serialization

### LINKED ISSUE #3615: Collect crashes after minimizing an app when Rank widget popup displayed.
Collect v1.25.1, Android v4.2-10, device used...

#### Problem description
Collect crashes after minimizing/extinguishing an app when Rank widget popup displayed.

1. Open Rank widget
2. Click Rank items button
3. Minimize/extinguish the app

#### Expected behavior
The app should not crash. User should be able to restore the app from the background. 

#### Other information 
I have used All widgets form with rank widget.

grzesiek2010:
Closes #3615

I tested the fix manually.

It's the solution we discussed in https://github.com/getodk/javarosa/pull/536#issuecomment-633733161

It should just fix the issue and have no other impact. We can test just the rank widget.

Any form with rank widget.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
@@ -22,7 +22,7 @@ import android.os.Bundle;
 import androidx.fragment.app.DialogFragment;
 import androidx.core.widget.NestedScrollView;
 import androidx.appcompat.app.AlertDialog.Builder;
-import androidx.lifecycle.ViewModelProviders;
+import androidx.lifecycle.ViewModelProvider;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.recyclerview.widget.ItemTouchHelper;
@@ -72,7 +72,7 @@ public class RankingWidgetDialog extends DialogFragment {
         if (context instanceof RankingListener) {
             listener = (RankingListener) context;
         }
-        viewModel = ViewModelProviders.of(this, new RankingViewModel.Factory(items, formIndex)).get(RankingViewModel.class);
+        viewModel = new ViewModelProvider(this, new RankingViewModel.Factory(items, formIndex)).get(RankingViewModel.class);
         if (viewModel.getItems() == null) {
             dismiss();
         }
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
@@ -22,6 +22,7 @@ import android.os.Bundle;
 import androidx.fragment.app.DialogFragment;
 import androidx.core.widget.NestedScrollView;
 import androidx.appcompat.app.AlertDialog.Builder;
+import androidx.lifecycle.ViewModelProviders;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.recyclerview.widget.ItemTouchHelper;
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
```diff
@@ -35,57 +36,46 @@
...
-    @Override
-    public void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        items = (List<SelectChoice>) (savedInstanceState == null
-                ? getArguments().getSerializable(ITEMS)
-                : savedInstanceState.getSerializable(ITEMS));
-        formIndex = (FormIndex) (savedInstanceState == null
-                ? getArguments().getSerializable(FORM_INDEX)
-                : savedInstanceState.getSerializable(FORM_INDEX));
+        viewModel = ViewModelProviders.of(this, new RankingViewModel.Factory(items, formIndex)).get(RankingViewModel.class);
```
lognaturel: Since this is deprecated, I think that in new code it makes sense to use the `ViewModelProvider` constructor - `new ViewModelProvider(this, new RankingViewModel.Factory(items, formIndex)).get(RankingViewModel.class)`

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
```diff
@@ -35,57 +36,46 @@
...
-        bundle.putSerializable(ITEMS, (Serializable) items);
-        bundle.putSerializable(FORM_INDEX, formIndex);
-        dialog.setArguments(bundle);
+    public RankingWidgetDialog() {
+    }
 
-        return dialog;
+    public RankingWidgetDialog(List<SelectChoice> items, FormIndex formIndex) {
+        this.items = new ArrayList<>(items);
+        this.formIndex = formIndex;
```
lognaturel, line 66: I don't love that these fields still end up existing since the point of the view model is to host them. What I was imagining originally is the activity creating the view model and the dialog fragment accessing it. I wasn't thinking about how the activity in this case is `FormEntryActivity` but currently it's the widget's responsibility to know about the items. 

Given the existing infrastructure, I find this ok for now. It's something @SaumiaSinghal and @seadowg could give another look at when they get to this widget.

lognaturel (commented):
It would be nice to have a test but I'm not really seeing a reasonable way to create one. My general sense is that it's worth getting this fix in for now and that the code structure will be reconsidered as part of @SaumiaSinghal's GSoC project anyway. Does that seem ok to you @seadowg, @SaumiaSinghal?

collect_app/src/main/java/org/odk/collect/android/widgets/RankingWidget.java
```diff
@@ -98,7 +98,7 @@ public void onButtonClick(int buttonId) {
         if (formController != null) {
             formController.setIndexWaitingForData(getFormEntryPrompt().getIndex());
         }
-        RankingWidgetDialog rankingWidgetDialog = RankingWidgetDialog.newInstance(savedItems == null ? items : savedItems, getFormEntryPrompt().getIndex());
+        RankingWidgetDialog rankingWidgetDialog = new RankingWidgetDialog(savedItems == null ? items : savedItems, getFormEntryPrompt().getIndex());
         rankingWidgetDialog.show(((FormEntryActivity) getContext()).getSupportFragmentManager(), "RankingDialog");
```
lognaturel, line 102: This is not really related to this PR but it just came to mind that I don't really know when `DialotUtils.showIfNotShowing` makes a difference. Should it be used here?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
grzesiek2010: >What I was imagining originally is the activity creating the view model and the dialog fragment accessing it

we can't create viewmodels using activity because then if we have multiple rank widgets they will share the same viewmodel with the same data. We need to create viewmodels using fragment scope.

I didn't think about it much so maybe I'm wrong but generally it would be great if every single widget were a fragment then we could have a viewmodel for every widget easily.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/widgets/RankingWidget.java
grzesiek2010: Can't be sure whether it's needed here I doubt but... generally it's good to use that method it's a defensive approach which could prevent some strange and rare crashes but that method uses bundle to pass data so that's what I need to avoid in this case.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- minimize
- lock screen
- rotation
- reopen from list of last opened apps
- reopen from list of the installed apps
- `Don't keep activities` turned on/off
- rank widgets in field-list

kkrawczyk123:

PR #3912, opened 2020-06-08
Fixed back button behavior in FormHierarchyActivity

### LINKED ISSUE #3799: Collect crashes when user removes all repeat groups and clicks on device back button

grzesiek2010:
Closes #3799

I tested the implemented changes manually and added automated tests.

We didn't handle a case when the index we started from (opening the `FormHierarchyActivity`) is deleted and then clicking device back button we try to get back to that index. My fix checks the index and tries to navigate back to the first relevant question if the starting index has been deleted. It's the best solution. We could also try to navigate to the beginning/end of the form but I find my solution better.

This should just fix the issue and have no side effects. The change in behavior (intended or not) might be only visible if we click device back button being in `FormHierarchyActivity` so just that functionality should be tested.

Any form with repeats.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
@@ -36,6 +36,7 @@ import org.javarosa.core.model.IFormElement;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.form.api.FormEntryCaption;
 import org.javarosa.form.api.FormEntryController;
+import org.javarosa.form.api.FormEntryModel;
 import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.HierarchyListAdapter;
@@ -736,7 +737,7 @@ public class FormHierarchyActivity extends CollectAbstractActivity {
         FormController formController = Collect.getInstance().getFormController();
         if (formController != null) {
             formController.getAuditEventLogger().flush();
-            formController.jumpToIndex(startIndex);
+            navigateToTheLastRelevantIndex(formController);
         }
 
         onBackPressedWithoutLogger();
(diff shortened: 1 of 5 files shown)


SaumiaSinghal (approved):

SaumiaSinghal:
Hi @grzesiek2010, I very much like the changes. My solution just jumps to the beginning of the form. I implemented this as I came across only those crashes in which the `startIndex` was not deleted and forgot to include that case. It is a more promising approach to try and go to the most relevant question if the `startIndex` has been deleted. Thanks for providing a better approach. :)

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -746,6 +747,29 @@ protected void onBackPressedWithoutLogger() {
         super.onBackPressed();
     }
 
+    private void navigateToTheLastRelevantIndex(FormController formController) {
+        FormEntryController fec = new FormEntryController(new FormEntryModel(formController.getFormDef()));
+        formController.jumpToIndex(startIndex);
+
+        // startIndex might no longer exist if it was a part of repeat group that has been removed
+        while (true) {
```
lognaturel, line 755: I don't love while(true), but the other option here would be a do-while and I like that even less so I agree with your choice.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -746,6 +747,29 @@ protected void onBackPressedWithoutLogger() {
         super.onBackPressed();
     }
 
+    private void navigateToTheLastRelevantIndex(FormController formController) {
```
lognaturel, line 750: I think this is a candidate for moving to JavaRosa's `FormEntryController`. It could become something like `jumpToIndexOrFirstPriorAvailable`. If you agree, please file an issue there. The two advantages are that it reduces complexity in Collect and that we would be able to use really fast unit tests for testing it.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
grzesiek2010: That seems fine I'll file an issue.

mmarciniak90:
@grzesiek2010 I can confirm that mentioned cases work correct but when I turned ON `Don't keep activities` option I noticed that dialog allowing to add group is doubled which causes a crash.

grzesiek2010:
> but when I turned ON Don't keep activities option I noticed that dialog allowing to add group is doubled which causes a crash.

it's not directly caused by this pr but I fixed it.

kkrawczyk123:
**Tested with success!**

Verified cases:
- scenarios from issue
- dark mode light mode
- add and delete more groups
- dont keep activities 
- delete group from hierarchy
- delete group from edit prompt
- different forms with groups
- using back button in hierarchy view

PR #3914, opened 2020-06-09
Allow tests to detect failed "fling" gestures

seadowg:
A lot of our tests rely on `FormEntryActivity#onFling` being called to swipe from left to right during form entry. We've had **a lot** of flakiness around this and after chipping away at rewriting tests it seems that there is an underlying problem where `onFling` simply isn't always called when we expect it to be. As far as I can see this is not the fault of any of Collect's code but rather a problem in our test environment.

To account for this I've added a `FlingRegister` component that tests and `FormEntryActivity` can share to let the tests know when a fling gesture hasn't worked so the test can then try again. Interestingly there was still flakes after adding this but upping the amount of retries seems to fix it. I'm guessing that there might be a momentary stall/freeze on the test device that's causing this problem.

This doesn't mean that we can relax and just swipe as much as we want: we still need to write tests that assert what question more so that we don't end up with weird scenarios where we swipe from an earlier or later question than we meant to.

Ran through test lab a lot of times.

It's a little weird but I think it's a good solution to the weirdness of the problem and although it means adding "test" code to the core app it doesn't feel intrusive enough to avoid given the benefit.

No risk to behaviour. Can just be merged after approval.

### DIFF
--- a/collect_app/src/androidTest/assets/forms/encrypted-no-instanceID.xml
+++ b/collect_app/src/androidTest/assets/forms/encrypted-no-instanceID.xml
@@ -14,7 +14,7 @@
     </h:head>
     <h:body>
         <input ref="/encrypted-no-instanceID/question">
-            <label>Question</label>
+            <label>Question 1</label>
         </input>
     </h:body>
 </h:html>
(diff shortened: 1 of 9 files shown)


lognaturel (approved):
Does feel wacky but like you said, not very intrusive, and certainly worth a shot.

PR #3917, opened 2020-06-10
Fixed bug in GeoPolyActivity.clear()

### LINKED ISSUE #3860: Polygon is created as an opened shape on map view when a previously created shape is removed
Collect v1.24.1 (older versions were not verified)

#### Problem description
Polygon is created as an opened shape on map view when a previously created shape is removed 

1. User opens form withGeoshape widget
2. User adds at least 3 points and points create closed shape
3. User clicks on the trash icon and deletes added points
4. User adds again at least 3 points
5. This time, **points do not create closed shape but the break is visible**
6. User saves not closed shape and opens map view again
7. Shape is closed

#### Expected behavior
There is no break in the shape

grzesiek2010:
Closes #3860 

I reproduced the scenario mentioned i the issue and confirmed this fix works.

It is just a bug fix the problem was that in GeoPolyActivity.clear() we always passed false as `closedPolygon` parameter not matter what type of geowidget we used.

It's a safe fix so testing the scenario from the issue would be enough to confirm that everything is fine.

All widgets form would be good.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -445,7 +445,7 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
 
     private void clear() {
         map.clearFeatures();
-        featureId = map.addDraggablePoly(new ArrayList<>(), false);
+        featureId = map.addDraggablePoly(new ArrayList<>(), outputMode == OutputMode.GEOSHAPE);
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- GeoShape widget
- Google Maps, Mapbox, OSM
- 3 input methods
- rotation
- offline layers
- toast when less than 3 points are added
- GeoTrace widgets to look at possible regression

kkrawczyk123:
Verified on Androids: 5.1 and 10

PR #3922, opened 2020-06-12
Reworking GeoShape and GeoTraceWidget

### LINKED ISSUE #3979: `View or Change location` button is visible when point is removed in map view
Collect v1.27.3 (older version were not verified)

#### Problem description
`View or Change location` button is visible when point is removed in map view

1. User starts form with GeoPoint widget with map
2. User adds a point on map and saves answer on map
3. User sees location in GeoPoint widget and clicks on `View or Change location`
4. User clicks on the trash icon on map view -> point is removed
5. User clicks on save icon on map view
6. User is moved to GeoPoint widget
7. **`View or Change location` is still visible on button**
8. User goes to next question and returns -> correct text is visible

ezgif com-video-to-gif

#### Expected behavior
`Start GeoPoint` button is visible even if point was removed in map view

#### Other information 
Discovered while testing #3877

SaumiaSinghal:
Closes #3979 
This PR is the work on GeoShape and GeoTrace Widget.

I pulled the GeoWidgets out of `BaseGeoWidget` class and hence removed the `BaseGeoWidget` and `BaseGeoWidgetTest`. 

I added `ActivityGeoDataRequester` that implements `GeoDataRequester` which requests geo data whenever geo button is clicked.

For testing purposes, I added `GeoWidgetHelpers` that checks whether the data passed to the intent on button click is correct or not. 

It adds an extra functionality of allowing the user to view any of the widgets (with maps/placement maps appearance) data in readOnly mode (when data is available). On master, buttons are hidden for the readOnly mode, whether or not data is available.
 
All Widgets form, or any form with the geo widgets.


collect_app/src/main/java/org/odk/collect/android/widgets/GeoPointWidget.java
```diff
@@ -17,153 +17,111 @@
...
-            if (!readOnly) {
-                startGeoButton.setText(
-                    dataAvailable ? R.string.change_location : R.string.get_point);
-            }
+            binding.simpleButton.setText(getDefaultButtonLabel());
+            binding.simpleButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+
+            binding.simpleButton.setOnClickListener(v -> {
+                Bundle bundle = GeoWidgetUtils.getGeoPointBundle(stringAnswer, accuracyThreshold, null, null);
+                GeoWidgetUtils.onButtonClick(context, prompt, getPermissionUtils(), null,
```
SaumiaSinghal: For all the four GeoWidgets, I call static method `onButtonClick`, when `geoButton` is clicked. In the corresponding test, I was thinking just verifying the method call, and checking different cases later on in unit tests for `GeoWidgetUtils`.

seadowg: I think given you want to test the shared behaviour it's good to try and think hard about what the responsibility of `onButtonClick` really is here. For me what it does is ask for geo data for the widget. It feels to me like there could be an interface called something like `GeoDataRequester` or `GeoDataFetcher` with methods like `requestPoint`, `requestShape` etc:

```
binding.simpleButton.setOnClickListener(v -> {
    geoDataRequester.requestPoint(promp);
});
```

 In each of the different geo widgets you can have a test that the `requestPoint` or whatever is called and then those methods can be tested for some implementation of the interface (`ActivityGeoDataRequester`).

Just to detail my questions here the though process I'm going through is:

1. I have shared code used in multiple places does it make sense to just test the behaviour in each place?
2. If not, what is the shared code responsible for? Is there some kind of interface/object/entity that could be responsible for that?

lognaturel:
I see this, @SaumiaSinghal and @seadowg. I will likely not be able to get to it this week, unfortunately. I think it's not blocking anything, is it?

SaumiaSinghal:
Hi @lognaturel! No worries! I just wanted to confirm something. Currently, for the `GeoTrace` and `GeoShape` widget, the widget button visibility is set to `View.GONE` when the prompt is `readOnly` and data is not available. Do we need to show the data on the maps as in the case of `GeoPointMap` widget when the prompt is `readOnly` and data is available?

lognaturel:
> Do we need to show the data on the maps as in the case of GeoPointMap widget when the prompt is readOnly and data is available?

Yes, that would be good to do. That can be useful if someone already has mapped something like a field or a road and wants to use the form to navigate to it.

SaumiaSinghal:
Oh okay okay. I will make changes in the PR.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
```diff
@@ -230,10 +234,16 @@ public void initMap(MapFragment newMapFragment) {
         }
         featureId = map.addDraggablePoly(points, outputMode == OutputMode.GEOSHAPE);
 
-        if (inputActive) {
+        if (inputActive && !intentReadOnly) {
             startInput();
         }
 
+        if (intentReadOnly) {
```
lognaturel: Won't this already be taken care of by the call on `updateUi`? I think you might be able to remove it.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -1,17 +1,127 @@
...
+
+    public static Bundle getGeoPolyBundle(String stringAnswer, GeoPolyActivity.OutputMode outputMode, boolean readOnly) {
+        final Bundle bundle = new Bundle();
+        bundle.putString(GeoPolyActivity.ANSWER_KEY, stringAnswer);
+        bundle.putSerializable(GeoPolyActivity.OUTPUT_MODE_KEY, outputMode);
+        bundle.putBoolean(READ_ONLY, readOnly);
+        return bundle;
+    }
+
+    private static void onButtonClick(Context context, FormIndex index, PermissionUtils permissionUtils, MapConfigurator mapConfigurator,
```
lognaturel: Why keep this static method? It's not called directly so can be removed (and the contents moved to `onButtonClicked`).

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -1,17 +1,127 @@
...
+        });
+    }
+
+    private static void startGeoActivity(Context context, Class activityClass, Bundle bundle, int requestCode) {
+        Intent intent = new Intent(context, activityClass);
+        intent.putExtras(bundle);
+        ((Activity) context).startActivityForResult(intent, requestCode);
+    }
+
+    private static void updateButtonLabelAndVisibility(GeoWidgetAnswerBinding binding, boolean readOnly, boolean dataAvailable,
```
lognaturel: Move contents to caller.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -1,17 +1,127 @@
...
+                        GeoWidgetUtils.truncateDouble(parts[3])
+                );
+            }
+        } catch (NumberFormatException e) {
+            return "";
+        }
+        return "";
+    }
+
+    public static Bundle getGeoPointBundle(String stringAnswer, double accuracyThreshold, Boolean readOnly, Boolean draggable) {
```
lognaturel: What's the advantage of putting this here rather than directly in `GeoPointWidget`? I don't see how it's implementing shared behavior.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -1,17 +1,127 @@
...
+        }
+        bundle.putDouble(ACCURACY_THRESHOLD, accuracyThreshold);
+        if (readOnly != null && draggable != null) {
+            bundle.putBoolean(READ_ONLY, readOnly);
+            bundle.putBoolean(DRAGGABLE_ONLY, draggable);
+        }
+        return bundle;
+    }
+
+    public static Bundle getGeoPolyBundle(String stringAnswer, GeoPolyActivity.OutputMode outputMode, boolean readOnly) {
```
lognaturel: This is short enough that I don't think there's huge advantage to it being shared. Each poly widget could build its own bundle.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -1,17 +1,127 @@
...
 
 import java.text.DecimalFormat;
 
 import timber.log.Timber;
 
-public class GeoWidgetUtils {
+import static android.view.View.GONE;
 
-    private GeoWidgetUtils() {
+public class GeoWidgetUtils implements GeoWidgetListener {
```
lognaturel: In general, the `Utils` suffix is a red flag because it implies "I didn't really know where to put this stuff so I put it here." It would be good to aim for clean abstractions that have meaningful names and specific responsibilities (e.g. `GeoDataRequester` as @seadowg suggested at https://github.com/getodk/collect/pull/3922#discussion_r459445775).

lognaturel (dismissed):
So sorry for the delay in reviewing this. I started out looking at some of the implementation and that's what my comments start with below. However, I think it would be worthwhile to take a step back and review the design and the structure of the tests. In general, I think you might find it a helpful practice @SaumiaSinghal to really carefully consider the `Why is this the best possible solution? Were any other approaches considered?` question in the PR template. It's particularly important in the case of a PR that mostly refactors. You should be able to describe why the code is in a better place after your changes.

I'd say the biggest change in this PR is the move away from `BaseGeoWidget`. It introduces a new `GeoWidgetListener` which is used by the widgets to communicate with the activity and also to change the label and button of the widget. I think already with this description you can see how `GeoWidgetListener` doesn't have a very clearly-defined responsibility. The concrete implementation of this listener is `GeoWidgetUtils` which mostly has static methods unrelated to the methods in the listener interface.

What @seadowg described at https://github.com/getodk/collect/pull/3922#discussion_r459445775 was an interface with a narrow responsibility. It would have different methods for the different geo activities so things like the activity class and the request code wouldn't be passed in -- they would be the responsibility of that interface's implementation. It's not quite as simple as @seadowg makes it look because of dependencies like `PermissionsUtils` but I think it would simplify your tests and make the code easier to understand. Consider the test [`buttonClick_callsOnButtonClicked`](https://github.com/getodk/collect/pull/3922/files#diff-af3e5bf7ce1a62bf63a5c092e8381606R195). Notice first how the name of the test is not really getting at what we care about. We care about the button click requesting a geopoint. So it could be renamed to `buttonClick_requestsGeopoint` and the assertion could be to make sure a method like `requestGeopoint` is called on a `GeoRequester` interface.

I haven't thought deeply about a structure for `setButtonLabelAndVisibility` yet. My first reaction is that because it's not communicating things between components it might not need to be part of an interface. I know it was previously part of `GeoWidget` but it feels like an implementation detail of the widget classes that tests don't need to know about. I also wonder whether the visibility and label parts should be split because button visibility never changes after a widget is first displayed whereas label does. Maybe an alternative would be something like a `GeoWidgetConfigurator` interface with specific methods like `setReadOnly`, `setViewOnly`, `setEditable`.

I know this is a lot. There's a ton going on here! You're deep in other widgets now so I think it makes sense for you to keep your focus there and then come back here with a fresh perspective once you have a moment.

SaumiaSinghal:
Thanks so much  @lognaturel for your thoughts, and for explaining where I'm going wrong. I should focus more on making the code readable rather than just implementing static methods for shared behavior and interfaces. This seems a lot of change. I will work on it, and let you know when it's ready. :)

lognaturel:
> where I'm going wrong

Please don't use the "w" word! 😄 This is difficult, subjective stuff and it takes a lot of practice to deeply get the implications of the various principles at work. I hope you see my comments as another opportunity for practice and refinement. Also, since @seadowg is your primary mentor for your GSoC project, you should take his word over mine. So feel very free to discuss some of the things I brought up with him and push back as appropriate. You can also take the opportunity to see if I've brought up ideas that might serve you in the other improvements you're making.

SaumiaSinghal:
Thanks for the insights @lognaturel. I removed certain shared methods like `updateButtonLabelAndVisisbility()`, as all the widgets use different labels in different scenarios, and also as you mentioned correctly, I need to change button visibility just once in the widget constructor.

Also, I renamed the interface to `GeoButtonClickListener` as now it just has one method `requestGeoIntent`, that is called every time the user clicks on the geo button. I refactored the unit tests accordingly. Let me know if there is something I can change/add to improve the geo widgets structure. :)

grzesiek2010:
@SaumiaSinghal does this pr fixes https://github.com/getodk/collect/issues/3979 maybe? If not that's fine since it's out of the scope but I'm just asking.

SaumiaSinghal:
Hi @grzesiek2010! I fixed it in the last commit and verified the change with unit tests as well.

seadowg:
@SaumiaSinghal to make this easier to review it would be great if you could answer the template questions. I'm sure there is more detail to be given in "How does this change affect users?" than just "No". 

It'd be good to describe the decisions you've made in the PR description or in comments in the code. Think about explaining the changes so that the reviewer has context on them before looking at the code. What new structures have you added? What structures have you changed? As much detail as possible is helpful when looking at these bigger PRs!

collect_app/src/main/java/org/odk/collect/android/widgets/interfaces/GeoButtonClickListener.java
```diff
@@ -0,0 +1,13 @@
+package org.odk.collect.android.widgets.interfaces;
+
+import android.content.Context;
+import android.os.Bundle;
+
+import org.javarosa.core.model.FormIndex;
+import org.odk.collect.android.utilities.PermissionUtils;
+import org.odk.collect.android.widgets.utilities.WaitingForDataRegistry;
+
+public interface GeoButtonClickListener {
```
seadowg: I think this interface should really be called the `GeoDataRequester` as that's what the widgets use it for. They do that on a button click but that's a detail of the widget.  It should also hide anything to do with the geo activity classes from the widget. The widget doesn't care how the geo data is acquired right? It just wants to get it in `setBinaryData` somehow.

 That being the case it probably makes sense to have different methods like `requestGeoPoint`, `requestGeoShape` etc and then have the implementation control what activity it launches in each case. The implementation should also take responsibility for building the `Bundle` and the `requestCode`. 

Given all that I think the implementation should probably be called something like `ActivityGeoDataRequester` as it's an activity based implementation of the request geo data.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoDataRequester.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String LOCATION = "gp";
+    public static final String ACCURACY_THRESHOLD = "accuracyThreshold";
+    public static final String READ_ONLY = "readOnly";
+    public static final String DRAGGABLE_ONLY = "draggable";
+    public static final double DEFAULT_LOCATION_ACCURACY = 5.0;
+
+    @Override
+    public void requestGeoIntent(Context context, FormIndex index, PermissionUtils permissionUtils,
+                                 WaitingForDataRegistry waitingForDataRegistry, Class activityClass, Bundle bundle, int requestCode) {
+        permissionUtils.requestLocationPermissions((Activity) context, new PermissionListener() {
```
seadowg: `permissionUtils` could probably be a field of this class and come in at the constructor rather than be passed in as a method argument.

seadowg (changes requested):
I think a bit of a restructuring of the `GeoButtonClickListener` interface would really improve this. This will move some testing around as I imagine a lot of tests could move to testing the implementation of that interface.

collect_app/src/main/java/org/odk/collect/android/widgets/interfaces/GeoButtonClickListener.java
SaumiaSinghal: I renamed `GeoDataRequester` class back to `GeoWidgetUtils`, as I thought it also has methods to convert answer string to geo data etc. I renamed the interface as `ActivirtyGeoDatarequester`, which has methods `requesGeoPoint`, `requestGeoShape`, `requestGeoTrace` and `requestGeoIntent`. Also, I am testing the interface separately in the `GeoWidgetUtilsTest`.

seadowg: Might be worth reading my comment again as I don't think that's what I was getting at. I don't think the `GeoWidgetUtils` should implement the `GeoDataRequester` interface (it should just be a non construtable utils class with helper methods) and I don't think the interface should be named `ActivityGeoDataRequester`, that was a suggested name for the implementation. Also, if it wasn't clear I was thinking you would get rid of `requestGeoIntent` and just have the three methods `requesGeoPoint`, `requestGeoShape`, `requestGeoTrace` which would all be `void`.

Very happy to discuss more if it's helpful!

seadowg (changes requested):
As in comment

seadowg:
@SaumiaSinghal could you fix conflicts and re-request review please?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.java
```diff
@@ -6,24 +6,23 @@
...
 
 import java.text.DecimalFormat;
 
 import timber.log.Timber;
 
-import static org.odk.collect.android.utilities.WidgetAppearanceUtils.MAPS;
-import static org.odk.collect.android.utilities.WidgetAppearanceUtils.PLACEMENT_MAP;
-import static org.odk.collect.android.utilities.WidgetAppearanceUtils.hasAppearance;
-
-public class GeoWidgetUtils implements ActivityGeoDataRequester {
```
seadowg: I think the `static` helper methods should stay in the `GeoWidgetUtils` class. There's no reason as far as I can see for them to live in `ActivityGeoDataRequester` and having them there creates a dependency between the widgets and the implementation of `GeoDataRequester` which isn't ideal.

seadowg (changes requested):

seadowg:
@SaumiaSinghal looks like tests and checks are failing

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.java
```diff
@@ -0,0 +1,108 @@
...
+import org.javarosa.form.api.FormEntryPrompt;
+import org.odk.collect.android.activities.GeoPointActivity;
+import org.odk.collect.android.activities.GeoPointMapActivity;
+import org.odk.collect.android.activities.GeoPolyActivity;
+import org.odk.collect.android.listeners.PermissionListener;
+import org.odk.collect.android.utilities.ApplicationConstants;
+import org.odk.collect.android.utilities.PermissionUtils;
+import org.odk.collect.android.utilities.WidgetAppearanceUtils;
+
+public class ActivityGeoDataRequester {
```
seadowg: This should still implement the interface so that the widget depends on an abstraction rather than an implementation.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.java
```diff
@@ -0,0 +1,108 @@
...
+    public static final String LOCATION = "gp";
+    public static final String ACCURACY_THRESHOLD = "accuracyThreshold";
+    public static final String READ_ONLY = "readOnly";
+    public static final String DRAGGABLE_ONLY = "draggable";
+
+    private final PermissionUtils permissionUtils;
+
+    public ActivityGeoDataRequester() {
```
seadowg: I don't think we should have this constructor. The `WidgetFactory` can construct a `PermissionUtils` object to pass in.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
```diff
@@ -1,212 +1,151 @@
...
 import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.promptWithReadOnly;
 import static org.odk.collect.android.widgets.support.QuestionWidgetHelpers.widgetTestActivity;
-import static org.robolectric.Shadows.shadowOf;
-
-/**
- * @author James Knight
- */
 
 @RunWith(RobolectricTestRunner.class)
 public class GeoPointWidgetTest {
```
seadowg, line 29: I think all the widgets are missing tests to check that `setBinaryData` actually updates the widget answer (`getAnswer`). It looks like right now the wigets are broken and don't save answers (just display them).

collect_app/src/test/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequesterTest.java
```diff
@@ -0,0 +1,183 @@
...
+    public void whenPermissionIsNotGranted_requestGeoTrace_doesNotLaunchAnyIntent() {
+        permissionUtils.setPermissionGranted(false);
+        activityGeoDataRequester.requestGeoTrace(testActivity, prompt, waitingForDataRegistry);
+
+        assertNull(shadowActivity.getNextStartedActivity());
+        assertTrue(waitingForDataRegistry.waiting.isEmpty());
+    }
+
+    @Test
+    public void whenPermissionIGranted_requestGeoPoint_launchCorrectIntent_andSetsFormIndexWaitingForData() {
```
seadowg: These three tests should be broken into multiple tests for the different calls to the methods under tests (for readOnly mode and different appearances for example). Otherwise, it's hard to read the test and if one fails it'll be harder to see what exactly has failed.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.java
SaumiaSinghal: I misinterpreted the last comment. The interface would be good here.

collect_app/src/test/java/org/odk/collect/android/widgets/GeoPointWidgetTest.java
SaumiaSinghal: I thought that the answers must be getting saved by calling `widgetValueChanged`

seadowg (approved):
@getodk-bot add label "needs testing"

mmarciniak90:
@SaumiaSinghal, does this PR require your attention or can we test it? It's been a while so I'd rather ask.

SaumiaSinghal:
Hi @mmarciniak90! This PR is ready for testing. You can go ahead. I confirmed with my mentor some time back.

mmarciniak90:
@SaumiaSinghal I noticed unexpected behavior, take a look at this, please.

I reproduce this issue on Android: 7.0, 8.1, 10.0 (other Android versions were not verified)

1. User starts form with GeoPoint widget with map
2. User adds a point on map and saves answer on map
3. User sees location in GeoPoint widget and clicks on View or Change location
4. User clicks on the trash icon on map view -> point is removed
5. User clicks on save icon on map view
6. User is moved to GeoPoint widget -> old point is not visible, `Start GeoPoint` is visible on button
7. User clicks on `Start GeoPoint` button
8. **User sees map with previous point added** -> **User is able to remove this point again** -> it's totally weird because this point was removed

seadowg:
@SaumiaSinghal I think it'd be good to get this one in next (sorry we've been a bit all over the place recently). Could you have a look at conflicts (probably needs a rebase) and the failing tests here?

seadowg:
@SaumiaSinghal ooops so I didn't realize there was also a problem reported by @mmarciniak90. Did you have a chance to look at that?

SaumiaSinghal:
Yes @seadowg that is fixed.

mmarciniak90:
**Tested with success**

Verified cases:
* regression test cases for GeoShape widget
* regression test cases for GeoTrace widget
* exploratory check

kkrawczyk123:

PR #3925, opened 2020-06-17
Implemented flexbox layout

### LINKED ISSUE #3744: Select appearance columns-pack should not take width of labels into account
Collect v1.26.1

#### Problem description
Select appearance columns-pack should not take width of labels into account

Load a form with selects with the `columns-pack` appearance and choices of different lengths. [columns-pack.zip](https://github.com/opendatakit/collect/files/4378966/columns-pack.zip)

device-2020-03-24-211201

#### Expected behavior
The padding between options should be fixed regardless of the label or image sizes. The current behavior is problematic when stacking choice lists where one is a subset of the other such as `Yes No Unknown` and `Yes No` in the form at https://forum.opendatakit.org/t/who-covid-19-minimum-reporting-form/25647

I think we should also add a little bit of extra space to the current padding when all choice labels are the same length. I think the very minimal padding works well with images but is too packed for text labels. Perhaps those could be two different cases?

grzesiek2010:
Closes #3744 

I tested some widgets manually and ran all automated tests.

I was able to get rid of a lot of code removing Grid widgets that didn't use recyclerview. That was something I had always wanted to do. I used google [FlexboxLayout](https://github.com/google/flexbox-layout) which allowed me not only improved the behavior described in the issue but also get rid of those Grid widgets.

| | columns-pack | compact | compact-2 |
| ------------- | ------------- |  ------------- |  ------------- |
| Before | Screenshot_1592390253  | Screenshot_1592390281  | Screenshot_1592390285 |
| After  | Screenshot_1592389989  | Screenshot_1592390006  | Screenshot_1592390015 |

This change should improve the behavior described in the issue and also improve the way we build compact (columns-pack no buttons) widgets (widgets with images instead of radio buttons/checkboxes). It would be good to test as many select widgets as possible since the risk is related to almost all of them.

All widgets form
[SelectsTestColumnsPack.xml.txt](https://github.com/getodk/collect/files/4791928/SelectsTestColumnsPack.xml.txt)
and generally any form with select widgets.
The form I prepared some time ago to document selects might be helpful https://docs.google.com/spreadsheets/d/1uA6NeUbE51WACiKUjErFUxYFeJdGGWglNGKivuzei7I/edit#gid=1181650558

Probably we need to update some screenshots only but it's not very important.


seadowg (approved):
Well this seems a lot nicer. Really happy to be getting rid of the `GridWidget` hierarchy 😄

mmarciniak90:
@grzesiek2010 @lognaturel

mmarciniak90:
@grzesiek2010 @lognaturel 

I do like changes for these widgets

| pr | v1.26.3|
|---|---|
|Screenshot_20200626-111440|Screenshot_20200626-111125|
Screenshot_20200626-111505|Screenshot_20200626-111142|



But I'm not a fan of causing regression on other widgets at the same time.

| pr | v1.26.3|
|---|---|
Screenshot_20200626-111606|Screenshot_20200626-111243|
Screenshot_20200626-112429|Screenshot_20200626-112114|
Screenshot_20200626-112603|Screenshot_20200626-112251

grzesiek2010:
1. The thing with start/end margins is that now we use only RecyclerView for all select widgets what''s awesome but at the same time we don't use margins in order to make the whole screen clickable. In other widgets we just use paddings to fake those start end/margins but if I did the same here every single element would have that padding what would add too much space between elements. I think it's not a big deal We can think of improvements here but it's not important in my opinion.

| Before  | After |
| Webp net-resizeimage  | Webp net-resizeimage (1)  |

-----------------------------------------------------------------------------------------------------------------------------

2. The border width is not even a regression here in fact it's a fix. Apparently GridWidgets used different sizes (look at other attachments where the border size differs) and now it's consistent

| Before  | After |
| 85844128-84decb80-b7a2-11ea-8b71-f60fd3c9bc41  | 85843698-d470c780-b7a1-11ea-9968-b3db271a6404  |

-----------------------------------------------------------------------------------------------------------------------------

3. GridWidgets didn't use paddings but if we used `columns-n` or `columns` appearances with RecyclerView padding was added so it also wasn't consistent and I tried to find balance like that. I think it looks good.

| Before  | After |
| 85844176-97590500-b7a2-11ea-9b79-484cf909bf72  | 85843700-d5095e00-b7a1-11ea-8537-d53881444f0e  |

kkrawczyk123:
@grzesiek2010 I can see an unnecessary search text field on widgets:
- Select one widget: Select one widgets with autocomplete appearance. appearance: compact autocomplete
- Select one widget: Select one widgets with autocomplete appearance. appearance: compact no-buttons autocomplete
- Select multiple widget: Select multiple widgets with autocomplete appearance. appearance: compact autocomplete
- Select multiple widget: Select multiple widgets with autocomplete appearance. appearance: compact no-buttons autocomplete

| master | Android 5.1 | Android 10 |
| ---------------- | ------------------- | ------------- |
| Screenshot_20200626-113817_master | Screenshot_2020-06-26-11-38-02 5 1 | Screenshot_20200626-113741 10 |

There is also a difference in displayed answers in columns.
The form that I have used: 
[selectTest3028.zip](https://github.com/getodk/collect/files/4836359/selectTest3028.zip)

grzesiek2010:
You are using `autocomplete` so you expect that search field. It was a bug on muster that it didn't work like that.
When it comes to:
> a difference in displayed answers in columns

don't you think it's better and expected? You are using compact (which is the same like columns-pack) appearance and it should accommodate as many elements in one row as possible.

kkrawczyk123:
@grzesiek2010  I agree that it looks better on devices with a bigger screen, not on smaller one but they are so rare so I don't think it's worth to do something about that. I've continued my testing and I have noticed that it looks badly in RTL language - answers are missing and scroll is deformed on Androids from 7-10, see the gif:
ezgif com-video-to-gif (5)

It looks in the best way on Android 5.1( list or huge space between questions):
Screenshot_2020-06-26-15-26-31
Screenshot (03_18PM, Jun 26, 2020)

grzesiek2010:
What appearance is that? Does that look worse than on master?

kkrawczyk123:
It is the form from issue, it looks much better on the master, the scroll remains untouched:
Screenshot_20200626-155436

lognaturel:
I am bothered by the left side of the choices not matching up with the question text. I would be ok with treating that as a separate issue, though, as I do think the benefits here far outweigh the inconvenience of that visual change.

Looking at it in more contexts, I also wish the vertical spacing were a little less. @grzesiek2010 I'm guessing you matched what was there for image grids? What about matching the text select vertical spacing? How does that look with images?

grzesiek2010:
> I am bothered by the left side of the choices not matching up with the question text. I would be ok with treating that as a separate issue, though, as I do think the benefits here far outweigh the inconvenience of that visual change.

The easiest solution that we could share across all widgets would be to have a global margin. The thing is that we decided to make the whole screen clickable for selects. I agreed on that but I wasn't totally convinced and now it's a problem here.
Is that really important or maybe we can change?

> Looking at it in more contexts, I also wish the vertical spacing were a little less. @grzesiek2010 I'm guessing you matched what was there for image grids?

Where is it too big? Generally spacing is the same for all elements it differs only between columns-pack (compact) appearance and the others
https://github.com/getodk/collect/pull/3925/files#diff-16abb480481e8a0a0d07cf1a8bee3aa2R209
but then every element might use a different layout with different spacing so it would be good to standardize it.

Another question: can we get rid of that arrow icon that indicates it's the quick appearance? It is used only with normal selects but requires a separate layout. We don't use it with no-buttons appearance and here in case of columns-pack it doesn't make much sense too. I think there is no need to display it. The separate layout which is required just to handle that arrow cause problems with rtl languages so that's why I'm asking about here too.

lognaturel:
> we decided to make the whole screen clickable for selects

Ah, yes, now I remember what this refers to. We've seen people try to tap the very left of their screen while holding their devices and having to tap again. I do think it's important to have the tap region extend all the way to the left of the screen. I don't have enough of the layouts loaded in my brain to know what's possible. The left spacing looks great with `columns-pack` and it's tappable all the way to the left, right? You just can't also achieve that with `columns-pack no-buttons`?

> Where is it too big?

You know what, I was wrong and the vertical spacing now matches all other selects which is what we want. I was referring to the vertical space between elements.

> can we get rid of that arrow icon that indicates it's the quick appearance?

I think I had advocated for keeping it but really, it's not very clear what it means. Given the trouble it's causing I'm up for removing it.

grzesiek2010:
I fixed one issue: the case where in rtl languages items were displayed in separate rows despite the fact column-pack appearance was used. However I'm not able to fix that empty space which is visible also with rtl languages:
Screenshot_1593517141
since:
- it takes place only when we use rtl languages
- it will be visible only when we use multiple questions on one screen
- and only with columns-flex appearance

I think we can ignore it for now, as @lognaturel said:
>the benefits here far outweigh the inconvenience of that visual change

so @kkrawczyk123 please file a separate issue.

> I think I had advocated for keeping it but really, it's not very clear what it means. Given the trouble it's causing I'm up for removing it.

I got rid of it, @kkrawczyk123 keep it in mind.

> The left spacing looks great with columns-pack and it's tappable all the way to the left, right? You just can't also achieve that with columns-pack **no-buttons**

would something like below would be ok?
Screenshot_1593526100

lognaturel:
Quick thing for the RTL `columns-pack` case: the label text should be grouped with the button. There's supposed to be padding after the text label to visually separate choices and it looks like that's always applied to the right. It should be applied to the end which is the left in RTL languages.

The spacing for `columns-pack no-buttons` in the second screenshot looks ok to me. It is a change and certainly if someone got their images just right for a particular screen size this could be problematic. I don't know how likely that is. The padding is not in fact very much, it just looks huge because the images are so small, I think. Let's put it out and see how users react. We can always go back to no padding if needed.

grzesiek2010:
@kkrawczyk123 I think it's ready for testing again.
I noticed that text (in rtl) in radio buttons/checkboxes was displayed like:
Screenshot_1593552720

and I fixed it but generally there are many more bugs in rtl so it's something for separate prs as long as it's not a big regression caused by this pr.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.1, 8.1, 9.0 and 10.

Tested scenarios:
- light mode
- dark mode
- change language
- RTL language
- minimize app
- screen rotate
- forms with columns-flex appearance
- widgets with quick appearance
- regression check on different types of select one question
- regression check on different types of select multiple question

PR #3927, opened 2020-06-17
Fixed selecting elements in no-button widgets

### LINKED ISSUE #3926: Can not unselect a selected item in no-button appearances after navigating back
#### Problem description
Can not unselect a selected item in no-button appearances after navigating back.

ezgif com-video-to-gif (5)

#### Expected behavior
The orange frame should disappear after clicking on a selected item.

grzesiek2010:
Closes #3926 

I tested the fix manually.

The problem was that we have every single item wrapped in a FrameLayout and in our code sometimes we called:
view.getChildAt(0).setBackground (changing a child) but sometimes view.setBackground (changing a parent FrameLayout). Now we always add/remove the frame using the parent FrameLayout.

It's a safe change that should just fix the issue. Testing select no-button widgets would be enough.

All widgets form would be enough.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/adapters/SelectMultipleListAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/adapters/SelectMultipleListAdapter.java
@@ -71,6 +71,7 @@ public class SelectMultipleListAdapter extends AbstractSelectListAdapter {
         void bind(final int index) {
             super.bind(index);
             if (noButtonsMode) {
+                view.setBackground(null);
                 for (Selection selectedItem : selectedItems) {
                     if (filteredItems.get(index).getValue().equals(selectedItem.getValue())) {
                         view.setBackground(ContextCompat.getDrawable(view.getContext(), R.drawable.select_item_border));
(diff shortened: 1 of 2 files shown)


lognaturel (approved):

mmarciniak90:
@grzesiek2010 I noticed that the `remove response` option does not work for this question type.
Maybe it's not totally related to this issue but it's the same result - removing added answer.

Use case:
1. User selects some images
2. User opens option "Remove response"
3. User confirms removing
4. Answers are still visible but when user changes the view, the orange frame is removed. The orange frame is also removed when user opens hierarchy view.

grzesiek2010:
> I noticed that the remove response option does not work for this question type.
Maybe it's not totally related to this issue but it's the same result - removing added answer.

I agree, should be fixed now.

mmarciniak90:
Thanks, @grzesiek2010 

**Tested with success**

Verified also on emulator with Android 5.1 

Verified cases:
* All widgets form
* selectTest3028.xml form
* unclick
* `remove response` option
* light/dark mode
* rotation

PR #3933, opened 2020-06-22
Fix imports from old QR codes

### LINKED ISSUE #3811: Not all settings are imported from older Collect versions

seadowg:
Closes #3811. ~~Depends on #3886.~~

This makes sure our `SharedPreferences` migrations are run after importing settings from a QR code (or from legacy `.settings`/`.settings.json` files). The core of the implementation is the new `SettingsImporter` use case that covers the shared logic of importing and processing settings.

Verified a few old QA myself and added new tests to cover this bug.

There are a lot of different ways to approach this. In the end, to me, it felt like just pulling the logic (or as much as possible) into one place where we can test it extensively was the way to go. The tests and the code both end up being very "and then, and then, and then" (like a bad video game plot) but it feels like it should be easier to understand and change/correct this behaviour in the future now.

As well as making sure the bug is fixed, imports of QR codes and from settings file would be good to cover


collect_app/src/test/java/org/odk/collect/android/configure/StructureAndTypeSettingsValidatorTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public void forNonJSON_returnsFalse() {
+        assertThat(validator.isValid("blah"), is(false));
+    }
+
+    @Test
+    public void withoutGeneralObject_returnsFalse() throws Exception {
+        JSONObject json = emptySettingsObject();
+        json.remove("general"); // Remove mutates and returns removed item :(
+
+        assertThat(validator.isValid(json.toString()), is(false));
```
lognaturel, line 40: I believe that Collect implicitly required both `general` and `admin` previously so good to keep consistent. I think we should follow up and make it only required to have one, not necessarily both to avoid having an empty set of settings. If that sounds reasonable I can file a low-priority issue (and corresponding doc update).

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/PropertyManagerJavaRosaInitializer.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public PropertyManagerJavaRosaInitializer(Application application) {
+        this.application = application;
+    }
+
+    @Override
+    public void initialize() {
+        PropertyManager mgr = new PropertyManager(application);
+
+        // Use the server username by default if the metadata username is not defined
```
lognaturel: Shouldn't this logic be a responsibility of the `PropertyManager`? Then maybe you don't need this class but can implement the interface inline where needed.

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
```diff
@@ -0,0 +1,104 @@
...
+            Map<String, Object> adminEntries = (Map<String, Object>) input.readObject();
+
+            return new Pair<>(generalEntries, adminEntries);
+        } finally {
+            try {
+                if (input != null) {
+                    input.close();
+                }
+            } catch (IOException ex) {
+                Timber.e(ex, "Exception thrown while closing an input stream due to: %s ", ex.getMessage());
```
lognaturel: While here, would you consider changing all this to a `try-with-resources` so it's easier to read and there isn't the `Timber.e` call?

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            return builder.toString();
+        } finally {
+            try {
+                if (br != null) {
+                    br.close();
+                }
+            } catch (IOException ex) {
+                Timber.e(ex, "Exception thrown while closing an input stream due to: %s ", ex.getMessage());
```
lognaturel: While here, would you consider changing all this to a try-with-resources so it's easier to read and there isn't the Timber.e call?

lognaturel (changes requested):
Funny thing: I had completely forgotten about the JSON file configuration path. My hope is that we can replace all the file-based configuration paths with something like a `.odksettings` extension that Collect recognizes. That way I could for example send a `JulyPolioVaccination.odksettings` file via WhatsApp and when a user who has Collect taps on it, Collect would open up and get configured (probably with a confirmation step -- we need to actually design this). In the mean time, if the JSON file path isn't already part of your testing plan @getodk/testers, you should be able to place a file named `collect.settings.json` in the root of the ODK folder.

Types of risk I see with this PR are: something missing after settings change (e.g. JavaRosa initialization), QR codes/settings files that previously worked no longer working (e.g. because of stricter validation), regression in messages shown for error states (e.g. certain errors aren't shown to the user anymore)

I focused my review on tests and high-level decisions made.

* Consolidating all the settings imports and migrations in `SettingsImporter` and `CollectPreferenceMigrator` and having all paths go through JSON makes it all a lot easier to reason about.
* Introducing the `configure` package seems like a nice way to separate out how preferences can be input.
* The validation on settings file is helpful. Eventually we may want to provide more troubleshoothing help to users but I think it's fine for now.
* Introducing an interface for JR initialization is good. Maybe it can be further simplified, see comment inline.
* Not having checkstyle yell at empty lambdas is helpful for testing.

Couple of small suggestions inline.

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    private Pair<Map<String, Object>, Map<String, Object>> readSettingsFile(File src) throws IOException, ClassNotFoundException {
+        // this should probably be in a thread if it ever gets big
+        ObjectInputStream input = null;
+        try {
+            input = new ObjectInputStream(new FileInputStream(src));
+
+            // first object is preferences
```
lognaturel: Remove comment that doesn't add much.

collect_app/src/test/java/org/odk/collect/android/configure/StructureAndTypeSettingsValidatorTest.java
seadowg: Given they can just be empty it feels ok to require both? Requiring both does simplify logic on our side as well.

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
seadowg: With pleasure.

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
seadowg: As above.

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/PropertyManagerJavaRosaInitializer.java
seadowg: Huh yeah I'll play with that.

seadowg: Ended up moving some stuff around so we can just call `reload` on `PropertyManager` and the actual JavaRosa initialization can happen in `ApplicationInitializer`.

collect_app/src/test/java/org/odk/collect/android/configure/StructureAndTypeSettingsValidatorTest.java
lognaturel: Fair enough!

collect_app/src/main/java/org/odk/collect/android/configure/LegacySettingsFileReader.java
lognaturel: 🤝

lognaturel (approved):
I like where `ApplicationInitializer` has ended up! I think there's a new source of risk for @getodk/testers to consider: when metadata settings are changed, the way that this is updated to be available to the form (e.g. with `username` field) has changed.

seadowg:
@getodk/testers if possible it would be great to prioritize this next as I'm going to need to add a migration as part of #3940.

seadowg:
@getodk/testers conflicts fixed!

kkrawczyk123:
@seadowg I have started testing from importing map settings from 1.22 to the build from pr and I have noticed some inconsistencies in behavior:
- I set OSM SDK on 1.22 and select CartoDB Dark Matter, imported qr code: I have Google/Streets in my settings on the build from PR. 
- The same issue is visible when set OSM SDK/ CartoDB Positron after import it is set to Google/Streets.
- I set OSM SDK on 1.22 and select USGS National Map Hybrid, imported qr code: Source has been properly imported as USGS, but map style is Topographic, except for Hybrid
For the same scenarios for OSM Streets, USGS National Map Topo, Stamen both sources and styles are imported correctly.

- I have also verified the same scenarios for Google map and Mapbox: map styles (satellite/terrain/hybrid) after importing qr code are set to default first option so Streets. (map sources are imported correctly)

seadowg:
@kkrawczyk123 I'm not sure what's going on. It could be that there are problems with the migrations themselves. Could you share the QR codes you've been working with?

seadowg:
@kkrawczyk123 I think I've found some bugs in the existing migration code itself. Hopefully they account for all the problems! I'll have a look at fixing.

seadowg:
@seadowg those problems should be fixed now!

seadowg:
@kkrawczyk123 those problems should be fixed now!

mmarciniak90:
@seadowg, ours today's testing focused on `collect.settings`. We know that there is a plan to remove this option in a future version (not sure in which) but for now, this option is available.

1. I generated `collect.settings` file on v1.22 -> [collect.settings.txt](https://github.com/getodk/collect/files/4931033/collect.settings.txt)
2. I tried to load this file on version form PR but I see a toast



3. The same file added to apk built on the master branch is applied correctly but outside the map data.


We also want to ask for instructions because we never checked `collect.settings.json`. How to get this file?

seadowg:
@mmarciniak90 I think the `.json` file uses the format detailed at the end of https://docs.getodk.org/collect-import-export/. 

In terms of the `.settings`: did you try that with the latest release as well? Very little has changed with the code that handles that import.

seadowg:
@mmarciniak90 never mind me I think I found the problem 😩. Will fix!

seadowg:
@mmarciniak90 @kkrawczyk123 legacy file issues should be fixed now!

kkrawczyk123:
@seadowg I have re-tested all map types and styles and started regression checks on other settings import. I have noticed an issue connected to import periodic form update and automatic download, this is the scenario I have followed:
1. Downloaded a Form on app with pr's build
2. On 1.22 I have set server settings to server that has the update of the form. Then I went to Form Management settings and choose Every Fifteen minutes as periodic form updates check and select automatic download checkbox. (I have also unselected hide old form version checkbox but that step probably doesn't even matter)
3. I have scanned QR code from 1.22 to pr's build
4. All settings seemed to look like imported correctly but the automatic download didn't work - notification didn't display. So it looks like a kind of false import.

I have executed the same case on master and settings were also imported but the automatic download worked after 15 minutes.

seadowg:
@kkrawczyk123 ah right yeah I see where that problem is. I'd actually suggest we let that one slide as I'm reworking a bunch of the code that schedules tasks as part of #3940. I'll take a note of that to make sure it'll get fixed but we'll want to check the periodic updates again anyway once Match Exactly is finished!

kkrawczyk123:
@seadowg I have continued testing of regression in importing settings. The admin password is not imported from 1.22. It works correctly on the master.

kkrawczyk123:

seadowg:
@kkrawczyk123 nice find! I'll have a look at this

seadowg:
@kkrawczyk123 ok that's fixed

mmarciniak90:
**Tested with success**

Verified on Android: 7.0, 8.1, 9.0

Verified cases:
* `collect.settings.json` file with admin password (admin),  Carto map with a dark style
* `collect.settings` file generated in v1.22, with admin password (a), Carto map with a dark style
* QR code generated in v1.22, with admin password (a), Carto map with a dark style

Here are files used for testing:
https://drive.google.com/drive/folders/1ozAXsEEmBAhVRM-lguKHGLbQn5qeLOA3?usp=sharing

kkrawczyk123:

PR #3937, opened 2020-06-23
Reworking Rating Widget

SaumiaSinghal:
This PR includes the work on rewriting Rating Widget.

Previously, Rating-Widget displayed stars simply on multiple lines, when the number of stars could not be accommodated in a single line. In this PR, the user just can have maximum of two lines for the rating widget.

Any form with Rating widget like [TestRatingWidget .xml.txt](https://github.com/getodk/collect/files/4817038/TestRatingWidget.xml.txt)

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -70,8 +70,8 @@ def mapboxToken = secrets.getProperty('MAPBOX_ACCESS_TOKEN', '')
 android {
     compileSdkVersion rootProject.androidCompileSdk
 
-    viewBinding {
-        enabled = true
+    buildFeatures {
+        viewBinding true
     }
 
     defaultConfig {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
@@ -20,7 +20,6 @@ import android.content.Context;
 import android.graphics.drawable.Drawable;
 import android.util.DisplayMetrics;
 import android.view.View;
-import android.widget.LinearLayout;
 
 import androidx.core.content.ContextCompat;
 
@@ -38,12 +37,11 @@ public class RatingWidget extends QuestionWidget {
     public static final int ASSUMED_TOTAL_MARGIN_AROUND_WIDGET = 40;
 
     private RatingWidgetAnswerBinding binding;
-    private int numberOfStars;
 
     public RatingWidget(Context context, QuestionDetails questionDetails, RangeQuestion rangeQuestion) {
         super(context, questionDetails);
-
-        numberOfStars = rangeQuestion.getRangeEnd().intValue();
+        binding.ratingBar.setNumStars(rangeQuestion.getRangeEnd().intValue());
+        binding.ratingBar.setStepSize(1.0F);
     }
 
     @Override
@@ -51,11 +49,8 @@ public class RatingWidget extends QuestionWidget {
         binding = RatingWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
         View answerView = binding.getRoot();
 
-        binding.ratingBar.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT));
-        binding.ratingBar.setNumStars(numberOfStars);
         binding.ratingBar.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> widgetValueChanged());
         binding.ratingBar.setEnabled(!prompt.isReadOnly());
-        binding.ratingBar.setStepSize(1.0F);
 
         if (prompt.getAnswerText() != null) {
             binding.ratingBar.setRating(Integer.parseInt(prompt.getAnswerText()));
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
@@ -18,6 +18,7 @@ import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
 import android.os.Build;
+import android.view.MotionEvent;
 import android.view.View;
 import android.widget.AbsSeekBar;
 
@@ -67,32 +68,12 @@ public class RatingWidget extends QuestionWidget {
             binding.ratingBar1.setMax(numberOfStars);
         }
 
-        binding.ratingBar1.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> {
-            binding.ratingBar2.setRating(0.0F);
-            binding.ratingBar1.setRating(rating);
-            widgetValueChanged();
-        });
-
-        binding.ratingBar2.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> {
-            binding.ratingBar1.setRating(maxNumberOfStars);
-            binding.ratingBar2.setRating(rating);
-        });
-
-        // fix for rating bar showing incorrect rating on Android Nougat(7.0/API 24)
-        // See https://stackoverflow.com/questions/44342481
-        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
-            try {
-                Field field = AbsSeekBar.class.getDeclaredField("mTouchProgressOffset");
-                field.setAccessible(true);
-                field.set(binding.ratingBar1, 0.6f);
-                field.set(binding.ratingBar2, 0.6f);
-            } catch (NoSuchFieldException | IllegalAccessException e) {
-                Timber.e(e);
-            }
+        if (prompt.isReadOnly()) {
+            binding.ratingBar1.setEnabled(false);
+            binding.ratingBar2.setEnabled(false);
+        } else {
+            setUpRatingBar(maxNumberOfStars);
         }
-
-        binding.ratingBar1.setEnabled(!prompt.isReadOnly());
-        binding.ratingBar2.setEnabled(!prompt.isReadOnly());
 
         if (prompt.getAnswerText() != null) {
             int rating = Integer.parseInt(prompt.getAnswerText());
@@ -121,6 +102,49 @@ public class RatingWidget extends QuestionWidget {
     @Override
     public void clearAnswer() {
         binding.ratingBar1.setRating(0.0F);
+    }
+
+    @SuppressLint("ClickableViewAccessibility")
+    private void setUpRatingBar(int maxNumberOfStars) {
+        // to quickly change rating on other rating bar in case onRatingChange listener is not called
+        binding.ratingBar1.setOnTouchListener((v, event) -> {
+            if (event.getAction() == MotionEvent.ACTION_DOWN) {
+                binding.ratingBar2.setRating(0);
+            }
+            return false;
+        });
+
+        binding.ratingBar2.setOnTouchListener((v, event) -> {
+            if (event.getAction() == MotionEvent.ACTION_DOWN) {
+                binding.ratingBar1.setRating(maxNumberOfStars);
+            }
+            return false;
+        });
+
+        binding.ratingBar1.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> {
+            binding.ratingBar2.setRating(0);
+            binding.ratingBar1.setRating(rating);
+            widgetValueChanged();
+        });
+
+        binding.ratingBar2.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> {
+            binding.ratingBar1.setRating(maxNumberOfStars);
+            binding.ratingBar2.setRating(rating);
+            widgetValueChanged();
+        });
+
+        // fix for rating bar showing incorrect rating on Android Nougat(7.0/API 24)
+        // See https://stackoverflow.com/questions/44342481
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
+            try {
+                Field field = AbsSeekBar.class.getDeclaredField("mTouchProgressOffset");
+                field.setAccessible(true);
+                field.set(binding.ratingBar1, 0.6f);
+                field.set(binding.ratingBar2, 0.6f);
+            } catch (NoSuchFieldException | IllegalAccessException e) {
+                Timber.e(e);
+            }
+        }
     }
 
     private int calculateMaximumStarsInOneLine() {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
@@ -19,11 +19,8 @@ import android.app.Activity;
 import android.content.Context;
 import android.graphics.drawable.Drawable;
 import android.util.DisplayMetrics;
-import android.view.LayoutInflater;
 import android.view.View;
-import android.view.ViewGroup;
 import android.widget.LinearLayout;
-import android.widget.RatingBar;
 
 import androidx.core.content.ContextCompat;
 
@@ -32,6 +29,7 @@ import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.IntegerData;
 import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
+import org.odk.collect.android.databinding.RatingWidgetAnswerBinding;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 
 @SuppressLint("ViewConstructor")
@@ -39,7 +37,7 @@ public class RatingWidget extends QuestionWidget {
 
     public static final int ASSUMED_TOTAL_MARGIN_AROUND_WIDGET = 40;
 
-    private RatingBar ratingBar;
+    private RatingWidgetAnswerBinding binding;
     private int numberOfStars;
 
     Integer answer;
@@ -52,12 +50,12 @@ public class RatingWidget extends QuestionWidget {
 
     @Override
     protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
-        ViewGroup answerView = (ViewGroup) LayoutInflater.from(context).inflate(R.layout.rating_widget_answer, this);
+        binding = RatingWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        View answerView = binding.getRoot();
 
-        ratingBar = answerView.findViewById(R.id.rating_bar);
-        ratingBar.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT));
-        ratingBar.setNumStars(numberOfStars);
-        ratingBar.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> widgetValueChanged());
+        binding.ratingBar.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT));
+        binding.ratingBar.setNumStars(numberOfStars);
+        binding.ratingBar.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> widgetValueChanged());
 
         if (prompt.getAnswerText() != null) {
             answer = Integer.parseInt(prompt.getAnswerText());
@@ -67,7 +65,7 @@ public class RatingWidget extends QuestionWidget {
 
     @Override
     public void setOnLongClickListener(OnLongClickListener l) {
-        ratingBar.setOnLongClickListener(l);
+        binding.ratingBar.setOnLongClickListener(l);
     }
 
     @Override
@@ -77,8 +75,8 @@ public class RatingWidget extends QuestionWidget {
 
     @Override
     public void clearAnswer() {
-        answer = 0;
-        ratingBar.setRating(0);
+        answer = null;
+        binding.ratingBar.setRating(0);
         widgetValueChanged();
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
@@ -17,7 +17,9 @@ package org.odk.collect.android.widgets;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
+import android.os.Build;
 import android.view.View;
+import android.widget.AbsSeekBar;
 
 import androidx.core.content.ContextCompat;
 
@@ -30,9 +32,12 @@ import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.ScreenUtils;
 import org.odk.collect.android.utilities.UiUtils;
 
+import java.lang.reflect.Field;
+
+import timber.log.Timber;
+
 @SuppressLint("ViewConstructor")
 public class RatingWidget extends QuestionWidget {
-
     private static final int ASSUMED_TOTAL_MARGIN_AROUND_WIDGET = 40;
     private static final int STANDARD_WIDTH_OF_STAR = 48;
 
@@ -73,6 +78,18 @@ public class RatingWidget extends QuestionWidget {
             binding.ratingBar2.setRating(rating);
         });
 
+        // fix for rating bar showing incorrect rating on Android Nougat(7.0/API 24)
+        // See https://stackoverflow.com/questions/44342481
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
+            try {
+                Field field = AbsSeekBar.class.getDeclaredField("mTouchProgressOffset");
+                field.setAccessible(true);
+                field.set(binding.ratingBar1, 0.6f);
+                field.set(binding.ratingBar2, 0.6f);
+            } catch (NoSuchFieldException | IllegalAccessException e) {
+                Timber.e(e);
+            }
+        }
 
         binding.ratingBar1.setEnabled(!prompt.isReadOnly());
         binding.ratingBar2.setEnabled(!prompt.isReadOnly());
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
@@ -17,9 +17,9 @@ package org.odk.collect.android.widgets;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
-import android.graphics.drawable.Drawable;
 import android.util.DisplayMetrics;
 import android.view.View;
+import android.widget.RatingBar;
 
 import androidx.core.content.ContextCompat;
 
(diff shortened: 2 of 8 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,92 +14,68 @@
...
-            for (int i = 0; i < numberOfStars; i++) {
-                ImageButton button = (ImageButton) gridLayout.getChildAt(i);
-                button.setImageResource(i <= position ? R.drawable.ic_star : R.drawable.ic_star_border);
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        binding = RatingWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        View answerView = binding.getRoot();
+
+        binding.ratingBar.setLayoutParams(new LinearLayout.LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT));
+        binding.ratingBar.setNumStars(numberOfStars);
```
SaumiaSinghal, line 56: I wanted to display the number of stars as defined in the `RangeQuestion` but `setNumStars()` here doesn't seem to work. It fails to override the value mentioned in the XML file. I found in the docs that there are some conditions, when this works like the `layout_width` must be set to `WRAP_CONTENT`.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,116 +14,92 @@
...
+        } else {
+            binding.ratingBar.setVisibility(GONE);
+            columns = calculateColumns(36);
+            rows = (int) Math.ceil((double) numberOfStars / columns);
+            if (rows == 1) {
+                binding.ratingBar2.setVisibility(VISIBLE);
+                this.ratingBar = binding.ratingBar2;
+            } else {
+                binding.ratingBar3.setVisibility(VISIBLE);
+                this.ratingBar = binding.ratingBar3;
```
SaumiaSinghal: I have three `Ratingbar` with different styles In the XML file. All of thems visibility is turned off by default. This part of code checks which `Ratingbar` needs to be displayed depending on the number of stars and screen dimensions.

seadowg:
@SaumiaSinghal I'm not sure about the change in behaviour here. Making the stars smaller as the total number increases does let us fit them on one line for longer but it means that each star's touch/tap area is becoming smaller and smaller. I'm worried this will make them very hard to select.

I don't have that much anecdotal evidence around how the rating widget is used (and it doesn't look like we have any analytics around it) but my assumption would be that people would typically use between 1 and 5 stars and that we don't need to optimize the aesthetics for numbers higher than that. Maybe @grzesiek2010 / @lognaturel can confirm or deny here?

collect_app/build.gradle
```diff
@@ -70,6 +70,10 @@ def mapboxToken = secrets.getProperty('MAPBOX_ACCESS_TOKEN', '')
 android {
     compileSdkVersion(28)
 
+    viewBinding {
```
seadowg: It looks like Google have switched up how this should be enabled annoyingly (found [here](https://developer.android.com/topic/libraries/view-binding#setup)). This will throw up a deprecation warning on build. Instead, we need:

```
 android {
    buildFeatures {
        viewBinding true
    }
}
```

config/quality.gradle
```diff
@@ -55,6 +55,7 @@ android {
         abortOnError true
         xmlReport true
         htmlReport true
+        checkGeneratedSources true
```
seadowg: What was this needed for?

grzesiek2010:
Yeah no one complained about having that widget with multiple lines of stars so there is no need to change it. As @seadowg probably in 99% of cases 5 start are used so it wasn't a problem. On the other hand if there is one crazy user which want to have let's say 100 stars probably they wouldn't be able to use the widget at all.

collect_app/src/main/res/layout/rating_widget_answer.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:layout_width="wrap_content"
+    android:layout_height="match_parent"
+    android:paddingStart="@dimen/margin_small"
```
seadowg: All these padding values can just be replaced with one `padding` as they are all the same!

seadowg (changes requested):
A couple of comments inline but I think we should discuss the behaviour change before reviewing most of this.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,116 +14,92 @@
...
+        return ratingBar;
     }
 
-    private int calculateColumns() {
-        Drawable starDrawable = getResources().getDrawable(R.drawable.ic_star);
+    private int calculateColumns(int widthOfStar) {
         DisplayMetrics dm = getScreenDimensions((Activity) getContext());
-        int widthOfStar = starDrawable.getIntrinsicWidth();
-        return (dm.widthPixels - ASSUMED_TOTAL_MARGIN_AROUND_WIDGET) / widthOfStar;
+        return (dm.widthPixels - ASSUMED_TOTAL_MARGIN_AROUND_WIDGET) / ((int) UiUtils.convertDpToPixel(widthOfStar, getContext()));
```
grzesiek2010: I think you can  use `ScreenUtils.getScreenWidth()` here and get rid of `getScreenDimensions()` at all.

grzesiek2010 (changes requested):

config/quality.gradle
SaumiaSinghal: I used this as otherwise `./gradlew lintDebug` fails and it gives an error for the unused resource (`geo_widget_answer.xml`)

collect_app/build.gradle
SaumiaSinghal: I did the same initially, but the Gradle sync fails and it says `No signature of method: build_.android() is applicable for argument types`

config/quality.gradle
seadowg: Ah do you mean `rating_widget_answer.xml`? I was confused initially but guessing this is because of View Binding's magic. Makes sense now!

collect_app/build.gradle
seadowg: Ah it looks like this PR has the older Gradle plugin. If you update your branch (`git rebase master`) so it has the latest changes from `master` you should be able to use the new definition.

SaumiaSinghal: Oh sure I will do that @seadowg

SaumiaSinghal:
I agree with @seadowg and @grzesiek2010. I just changed it because I thought the appearance of `RatingBar.INDICATOR` looks better than rating bar on multiple lines when we have 8 or 9 stars, or maybe for more stars when the user has a smaller screen. However, I feel that in case of 10 stars and with `RatingBar.SMALL` style, there might be problem or the user may not be able to select the desired number of stars with as much ease as before. So, should I revert back to the old behaviour?

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,91 @@
...
-        int numberOfStars = rangeQuestion.getRangeEnd().intValue();
-        int columns = calculateColumns();
-        int rows = (int) Math.ceil((double) numberOfStars / columns);
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        RatingWidgetAnswerBinding binding = RatingWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        View answerView = binding.getRoot();
 
-        gridLayout = new GridLayout(context);
-        gridLayout.setLayoutParams(new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
```
grzesiek2010, line 56: You have two blank lines here please remove one.

grzesiek2010:
> So, should I revert back to the old behaviour?

I would say yes. If during your work you notice something that could be improved but is not directly related to tour goal please file a separate issue where we can discuses it. Here your goal is to refactor the widget not to implement fixes like that one.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -17,69 +17,89 @@
...
 
     @Override
     public void clearAnswer() {
-        binding.ratingBar.setRating(0.0F);
+        ratingBar.setRating(0.0F);
     }
 
-    protected RatingWidgetAnswerBinding getBinding() {
-        return binding;
+    protected RatingBar getRatingBar() {
```
grzesiek2010: This is just for tests right? I would change access modifier (of the field) to package-private and use that field directly without any getter method like this one.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -17,69 +17,89 @@
...
 import org.odk.collect.android.formentry.questions.QuestionDetails;
+import org.odk.collect.android.utilities.UiUtils;
 
 @SuppressLint("ViewConstructor")
 public class RatingWidget extends QuestionWidget {
 
     public static final int ASSUMED_TOTAL_MARGIN_AROUND_WIDGET = 40;
 
-    private RatingWidgetAnswerBinding binding;
+    private RatingBar ratingBar;
```
grzesiek2010: Why do we need this extra field at all? Why can't you use binding.ratingBar everywhere?

grzesiek2010 (changes requested):

collect_app/src/test/java/org/odk/collect/android/widgets/RatingWidgetTest.java
```diff
@@ -2,60 +2,130 @@
...
+    }
+
+    @Test
+    public void ratingBarShowsCorrectNumberOfStars() {
+        RatingWidget widget = createWidget(promptWithReadOnlyAndRangeQuestion(rangeQuestion));
+        assertThat(widget.getRatingBar().getNumStars(), equalTo(5));
+    }
+
+    @Test
+    public void getAnswer_whenPromptAnswerDoesNotHaveAnswer_returnsZero() {
```
grzesiek2010: Should be returns null not zero right?

collect_app/src/test/java/org/odk/collect/android/widgets/support/QuestionWidgetHelpers.java
```diff
@@ -37,4 +38,18 @@ public static FormEntryPrompt promptWithReadOnly() {
                 .withReadOnly(true)
                 .build();
     }
+
+    public static FormEntryPrompt promptWithRangeQuestionAndAnswer(QuestionDef questionDef, IAnswerData answer) {
```
grzesiek2010: You can use it in other widgets as well so you should''t use Range word here.

collect_app/src/test/java/org/odk/collect/android/widgets/support/QuestionWidgetHelpers.java
```diff
@@ -37,4 +38,18 @@ public static FormEntryPrompt promptWithReadOnly() {
...
+        return new MockFormEntryPromptBuilder()
+                .withQuestion(questionDef)
+                .withAnswer(answer)
+                .build();
+    }
+
+    public static FormEntryPrompt promptWithReadOnlyAndRangeQuestion(QuestionDef questionDef) {
```
grzesiek2010: Same as above it's not just for range widgets.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
SaumiaSinghal: Thanks @grzesiek2010! That's a better approach.

SaumiaSinghal:
Thanks @grzesiek2010 for reviewing. I reverted back to the original behavior. I think having the rating bar shown in two lines, not more than that, should suffice.

grzesiek2010:
> I think having the rating bar shown in two lines, not more than that, should suffice.

Please revert this change as well we didn't discuss it so as I said above in this pr where the goal is to improve the code (not behavior) it shouldn't exist.

SaumiaSinghal:
Hi @grzesiek2010! I had a doubt, if I revert this change, then I might need to initialize the `Ratingbar` dynamically, as I would not know how many ratingbars would be required to display the required number of stars. And, as part of the project 'Rewriting the Widget Framework', I want to shift as much code as possible to XML files.

grzesiek2010:
> I had a doubt, if I revert this change, then I might need to initialize the Ratingbar dynamically, as I would not know how many ratingbars would be required to display the required number of stars. And, as part of the project 'Rewriting the Widget Framework', I want to shift as much code as possible to XML files.

I see only creating the layout dynamically we can achieve multi-line rating bar. So I can see two options:
1. We can keep creating this widget dynamically.
2. Or we can display it in one line like all the rating bars I can google. I don't know why we didn't do this in that way since in 99.99% of cases probably our users have 5 stars. @lognaturel what do you think was there anyone who asked for such strange rating bar with many stars?

lognaturel:
I think the benefits of keeping the layout in XML are worth it and as @grzesiek2010 says, it's going to be really rare to have so many stars that they can't fit on one line. As @seadowg pointed out to me, [imdb](https://www.imdb.com/) is an example of a broadly-used system that has a 10-star rating. On smaller screens that would wrap in portrait. So I think @SaumiaSinghal's approach of allowing up to 2 lines is a good middle ground. 

Ideally there'd be a message if not all stars can be displayed but I don't know how easy that is. It's a problem if there are 10 stars, the user only sees 8 and selects 8 thinking that's the max rating, for example. If displaying a message is really impractical, then I think we just need to be really explicit in the documentation.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
 import org.odk.collect.android.formentry.questions.QuestionDetails;
-import org.odk.collect.android.formentry.questions.WidgetViewUtils;
+import org.odk.collect.android.utilities.ScreenUtils;
+import org.odk.collect.android.utilities.UiUtils;
 
+@SuppressLint("ViewConstructor")
 public class RatingWidget extends QuestionWidget {
 
     public static final int ASSUMED_TOTAL_MARGIN_AROUND_WIDGET = 40;
+    public static final int STANDARD_WIDTH_OF_STAR = 48;
```
grzesiek2010: Both can be private.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
-                ImageButton button = (ImageButton) gridLayout.getChildAt(i);
-                button.setImageResource(R.drawable.ic_star);
-            }
-        }
+        RangeQuestion rangeQuestion = (RangeQuestion) prompt.getQuestion();
+        int numberOfStars = rangeQuestion.getRangeEnd().intValue();
 
-        addAnswerView(gridLayout, WidgetViewUtils.getStandardMargin(context));
-    }
+        int maxNumberOfStars = (int) ((ScreenUtils.getScreenWidth() - ASSUMED_TOTAL_MARGIN_AROUND_WIDGET)
```
grzesiek2010: Please factor it out to a separate method, I mean the calculation because you can keep the variable.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
-            gridLayout.getChildAt(i).setOnLongClickListener(l);
-        }
+        binding.ratingBar1.setOnLongClickListener(l);
+        binding.ratingBar2.setOnLongClickListener(l);
     }
 
     @Override
     public IAnswerData getAnswer() {
-        return answer != null ? new IntegerData(answer) : null;
+        return binding.ratingBar1.getRating() == 0.0F ? null :
```
grzesiek2010: I think more readable way of using the ternarry operator in this case would be:
```
return binding.ratingBar1.getRating() == 0.0F 
                        ? null 
                        : new IntegerData((int) (binding.ratingBar1.getRating() + binding.ratingBar2.getRating()));
```
so new line for each option.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
+                binding.ratingBar2.setRating(rating - maxNumberOfStars);
+            } else {
+                binding.ratingBar1.setRating(rating);
+            }
+        }
+        return answerView;
     }
 
     @Override
     public void setOnLongClickListener(OnLongClickListener l) {
```
grzesiek2010: Please fix this one-char parameter name, even simple `listener` would be better.

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
-        String answerText = questionDetails.getPrompt().getAnswerText();
-        if (answerText != null) {
-            answer = Integer.parseInt(answerText);
-            for (int i = 0; i < answer; i++) {
-                ImageButton button = (ImageButton) gridLayout.getChildAt(i);
-                button.setImageResource(R.drawable.ic_star);
-            }
-        }
+        RangeQuestion rangeQuestion = (RangeQuestion) prompt.getQuestion();
+        int numberOfStars = rangeQuestion.getRangeEnd().intValue();
```
grzesiek2010: This also could be factored out to a new method (again the calculation, you can keep the variable - as I said below).

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -14,121 +14,95 @@
...
-        addAnswerView(gridLayout, WidgetViewUtils.getStandardMargin(context));
-    }
+        int maxNumberOfStars = (int) ((ScreenUtils.getScreenWidth() - ASSUMED_TOTAL_MARGIN_AROUND_WIDGET)
+                / UiUtils.convertDpToPixel(STANDARD_WIDTH_OF_STAR, getContext()));
 
-    private void renderGrid(Context context, int numberOfStars, int columns, int rows) {
-        gridLayout.setColumnCount(columns);
-        gridLayout.setRowCount(rows);
+        binding.ratingBar1.setStepSize(1.0F);
+        binding.ratingBar2.setStepSize(1.0F);
```
grzesiek2010: You already have `android:stepSize in the layout file so why do you need this here too?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
SaumiaSinghal: I'll fix this. Thanks for pointing out

grzesiek2010 (approved):

kkrawczyk123:
@SaumiaSinghal I have noticed issues:
- There is a color difference between master and PR. Is that change intentional?
**PR:**

| light mode | dark mode |
| ------ | ------ | 
| Screenshot (Jul 24, 2020 10_54_03 AM) | Screenshot (Jul 24, 2020 10_55_13 AM) |

**Master:**

| light mode | dark mode |
| ------ | ------ | 
| Screenshot_20200724-105833 | Screenshot_20200724-105740 |

- The color difference on Android 5.1: Stars are gray and they have black/white frames.

| light mode | dark mode |
| -------- | -------- | 
| Screenshot (09_45AM, Jul 24, 2020) | Screenshot (09_47AM, Jul 24, 2020) |

- I am able to choose a half or other parts of stars, see the gif:
ezgif com-video-to-gif (14)
The second issue that is visible at the gif is that answers are removed when moving forward-backward and when jumping to hierarchy. The form that I have used: [TestRatingWidget.xml.txt](https://github.com/getodk/collect/files/4970924/TestRatingWidget.xml.txt)



SaumiaSinghal:
@kkrawczyk123 I have fixed all the mentioned issues. Sorry for multiple testing requests. Can you test again, please?

kkrawczyk123:

kkrawczyk123:
@lognaturel I have a question about the behavior on Android 5.1 where not-chosen stars are not fulfilled and have a highlight, it differs from other devices where all not selected stars are fulfilled. Should we fix it now, open separate issue, or maybe changing it is not worth the effort as it's only Android 5 that is used rather rarely? 
See the screenshots:
Android 5.1:
| light mode | dark mode |
| ------ | ------- |
| Screenshot (12_52PM, Jul 27, 2020) | Screenshot (12_54PM, Jul 27, 2020) |

other Androids:
| light mode | dark mode |
| ------ | ------- |
| Screenshot_20200724-105833 | Screenshot_20200724-105740 |

mmarciniak90:
@SaumiaSinghal I noticed unexpected behaviors.

Use case 1:
1. I used this form [TestRatingWidget.xml.txt](https://github.com/getodk/collect/files/4988308/TestRatingWidget.xml.txt)
2. I opened the last question with 10 starts
3. I selected all stars.
4. Then I clicked on the fourth star -> I expected that only 4 stars will be selected and 6 will be unselected
5. Only stars in the second line are unselected



Use case 2:
1. I used this form [TestRatingWidget.xml.txt](https://github.com/getodk/collect/files/4988308/TestRatingWidget.xml.txt)
2. I opened the last question with 10 starts
3. I selected all stars.
4. I opened remove response dialog
5. I accepted removing response
6. Only stars in the second line are unselected

SaumiaSinghal:
Hi @mmarciniak90, Thanks for reporting the issue. I have fixed the problem. Can you test again please?

mmarciniak90:

collect_app/src/main/java/org/odk/collect/android/widgets/RatingWidget.java
```diff
@@ -61,8 +61,8 @@ protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int a
         }
 
         binding.ratingBar1.setOnRatingBarChangeListener((ratingBar, rating, fromUser) -> {
-            binding.ratingBar1.setRating(rating);
             binding.ratingBar2.setRating(0.0F);
+            binding.ratingBar1.setRating(rating);
```
seadowg: Could this change have been driven out with a test? Maybe there was one and I missed it!

SaumiaSinghal: Sorry @seadowg for missing the tests. I added unit tests and confirmed that unit test`changingRating_updatesAnswer_forRatingBarInMultipleLines()` fails for the last commit, and passes on the latest one.

lognaturel:
> Android 5.1 where not-chosen stars are not fulfilled and have a highlight, it differs from other devices where all not selected stars are fulfilled

I think this is fine to leave as-is. It's still very clear what's going on and we have fewer and fewer Android 5.1 users.

mmarciniak90:
@SaumiaSinghal I noticed one more bug. It is visible only on Android 7.0. I checked the real device and emulator.

Sometimes if I select X stars, X+1 stars are selected. I hope you will be able to see this on the attached gif and reproduce it.

ezgif com-video-to-gif

@lognaturel I know that this problem is visible only on one Android (among those that we have) but I feel that we should fix it because it's not possible to easily select the answer you want.

mmarciniak90:
@SaumiaSinghal were you able to reproduce this issue? Do you need some help?

SaumiaSinghal:
Oh sorry @mmarciniak90 Somehow I missed your comment. I will check rn

SaumiaSinghal:
Thanks @mmarciniak90 for reporting the issue. It has something to do with `mTouchProgressOffset` value not set correctly in `AbsSeekBar` class. I found the fix [here](https://stackoverflow.com/questions/44342481/android-nougat-7-0-api-24-rating-count-is-wrong-in-onratingbarchangelistener). Can you test again please?

mmarciniak90:

mmarciniak90:
@SaumiaSinghal there is still unexpected behavior:

1. Open _TestRatingWidget.xml_ form
2. Go to question with 10 stars
3. Select all stars
4. Click on the last star in the first row to unselect stars in the second row -> it's not possible

I checked this case on Android versions: 7.0, 8.1, 9.0



Additionally, I still see that 1 star more is marked than I selected on Android 7.

SaumiaSinghal:
Hi @mmarciniak90! I fixed the problem with selecting rating when clicking on the last star of the `RatingBar`. I should have guessed that, that is the case when `OnRatingChangeListener` is not called as the rating on the top rating bar is still the same. I fixed the problem by setting an `OnTouchListenr` on both the rating bars, which change the rating of the other rating bar as soon as the rating of any one of them is changed (anyone of them is touched).  
However, I didn't see the issue of setting a rating on Android 7. I do not have a real device, but I tested it on an emulator.

seadowg (changes requested):
@SaumiaSinghal Sorry I think this got a little lost. Could you rebase on top of master?

collect_app/build.gradle
```diff
@@ -70,8 +70,8 @@ def mapboxToken = secrets.getProperty('MAPBOX_ACCESS_TOKEN', '')
 android {
     compileSdkVersion(28)
 
-    viewBinding {
-        enabled = true
+    buildFeatures {
+        viewBinding true
```
seadowg, line 74: I don't think we need this anymore right? It's enabled in the `android` block now.

seadowg (approved):

One little tweak but doesn't affect this going to QA!

kkrawczyk123:
@SaumiaSinghal I have noticed another issue. On the rating widget where stars do not fit to one line I have selected a whole 1st line and at least 1 star from the 2nd line. After rotating the screen all-stars are selected, see the gif:
ezgif com-video-to-gif (34)

When I opened hierarchy view it says that I have selected 11 stars from 10:
Screenshot_2020-09-23-10-45-32-219_org odk collect android

seadowg:
@SaumiaSinghal looks like there is a problem that QA found here and some small conflicts. Sorry for this being so slow. Trying to go through these in order!

seadowg:
@SaumiaSinghal could you try rerunning on Circle CI. It won't let me do it 😭

SaumiaSinghal:
Sure @seadowg! Also, I could not fix the problem reported by the QA. I tried debugging the app and commenting out my code, but could not find where my code is setting the rating to max number of stars on screen rotation.

seadowg:
Ach I think this is due to a super annoying, almost completely invisible feature of Android. Basically when the Activity rotates and is recreated it will take the state from views with IDs, save them and then restore them once the view is rendered again. This means that the state from `rating_bar1` is being saved (100%) and then is restored when the view is rotated. This happens after the view has been created so after we've set the rating bar to have the correct number of starts. We ran into this recently with selects in field lists as well. 

The solution is to add `android:saveEnabled="false"` to the `RatingBar` views so this automatic save/restore doesn't happen. Yuck! I'll add a note to the docs about this.

SaumiaSinghal:
Thanks so much @seadowg, it worked. It is ready for testing

mmarciniak90:
@SaumiaSinghal I can still see the issue described here: https://github.com/getodk/collect/pull/3937#issuecomment-665727700

It is a gif from the current version

ezgif com-video-to-gif

@kkrawczyk123 are you able to reproduce it on Android 7.1?

SaumiaSinghal:
Hi @mmarciniak90 I didn't notice this bug on my emulator running Android 7.0 :/

rating_widget

kkrawczyk123:
@mmarciniak90 @SaumiaSinghal I am not able to reproduce it on my Android 7.1.

seadowg:
@mmarciniak90 what version of Android are you seeing that problem on?

mmarciniak90:
Only on a real device with Android 7.0, Samsung Galaxy S7. 
I think creating a separate issue for this is a good solution. We should not block this PR because of this. 
@lognaturel @seadowg are you agree?

EDIT:
New issue created #4192

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.1, 8.1, 9.0 and 10.

Tested scenarios:
- light/dark mode
- remove response
- cancel removing response
- screen rotation 
- minimize app
- jump form hierarchy
- going backward/forward
- RTL language 
- all detected issues have been fixed 

PR #3947, opened 2020-06-25
Call widgetValueChanged on UI thread to avoid race condition

### LINKED ISSUE #3944: ArrayIndexOutOfBoundsException in FormEntryActivity.java line 1102
Collect v1.27.2 and older versions

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8483a6eb0b2743cb499d9a81db50aee3?time=last-seven-days&sessionId=5EF2EC790162000138B7078956868651_DNE_0_v2

```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=2; index=2
       at org.odk.collect.android.activities.FormEntryActivity.saveAnswersForCurrentScreen(FormEntryActivity.java:1102)
       at org.odk.collect.android.activities.FormEntryActivity.updateFieldListQuestions(FormEntryActivity.java:2833)
       at org.odk.collect.android.activities.FormEntryActivity.widgetValueChanged(FormEntryActivity.java:2802)
       at org.odk.collect.android.formentry.ODKView.widgetValueChanged(ODKView.java:663)
       at org.odk.collect.android.widgets.QuestionWidget.widgetValueChanged(QuestionWidget.java:539)
       at org.odk.collect.android.widgets.AbstractSelectOneWidget.onClick(AbstractSelectOneWidget.java:141)
       at org.odk.collect.android.adapters.SelectOneListAdapter.onClick(SelectOneListAdapter.java:69)
       at android.view.View.performClick(View.java:7346)
       at android.widget.TextView.performClick(TextView.java:14167)
       at android.widget.CompoundButton.performClick(CompoundButton.java:143)
       at android.view.View.performClickInternal(View.java:7312)
       at android.view.View.access$3200(View.java:846)
       at android.view.View$PerformClick.run(View.java:27801)
       at android.os.Handler.handleCallback(Handler.java:873)
```

?

grzesiek2010:
Closes #3944 

I reviewed implemented changes and tested forms with field-lists.

I wasn't able to reproduce the issue under normal circumstances but I was able to figure out what causes it.
It's a sheer race condition problem. In some cases `widgetValueChanged()` might be called from a non-main thread (as mentioned in the [comment](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-3944?expand=1#diff-03ff45091550285dd88b56927010c457R2750)) so if that happens and the method is called multiple times in a short period of time strange things might happen like `widgetValueChanged()` might be called again when the previous one is not finished yet because some code we ran on UI thread is still working.
So the solution is to run everything on UI thread not only adding and removing widgets like before. From what i have noticed in most cases everything works on UI thread either way (fro example I had to hardcoded a new thread to reproduce the issue because I wasn't able to do that without such a trick).

It shouldn't be risky since as I said in most cases the whole code responsible for updating the field-list runs on Ui thread either way but testing one form with field-list to make sure there is no regression would be fine.

Any form with field-list + calculations ideally to hide/show some questions depending on results.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2747,15 +2747,21 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         }
 
         if (formController.indexIsInFieldList()) {
-            updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
-
-            odkView.addOnLayoutChangeListener(new View.OnLayoutChangeListener() {
+            // Some widgets may call widgetValueChanged from a non-main thread but odkView can only be modified from the main thread
+            runOnUiThread(new Runnable() {
                 @Override
-                public void onLayoutChange(View v, int left, int top, int right, int bottom, int oldLeft, int oldTop, int oldRight, int oldBottom) {
-                    if (!odkView.isDisplayed(changedWidget)) {
-                        odkView.scrollTo(changedWidget);
-                    }
-                    odkView.removeOnLayoutChangeListener(this);
+                public void run() {
+                    updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
+
+                    odkView.addOnLayoutChangeListener(new View.OnLayoutChangeListener() {
+                        @Override
+                        public void onLayoutChange(View v, int left, int top, int right, int bottom, int oldLeft, int oldTop, int oldRight, int oldBottom) {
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

kkrawczyk123:
**Tested with success!**
I didn't notice any regression.

Tested scenarios:
- form with depending calculation
- a few different forms with field-list
- removing answers from field-list questions
- going forward/backward
- form with rank and field-list
- light mode, dark mode

mmarciniak90:
Tested with success.

Regression was also not observed on Android versions: 7.0, 8.1, 9.0

PR #3954, opened 2020-07-02
Update storage migration banner for v1.28

### LINKED ISSUE #3936: Update scoped storage migration banner for v1.28
The scoped storage migration banner currently reads "You must migrate Collect forms to private storage by August 2020 to comply with Android requirements." We should change this to make sure users know that in the following version the migration will be forced.

I think we need to keep some high-level context: "You must migrate Collect forms to private storage to comply with Android requirements."

Then we need to add some urgency. Possible ideas are:
* "This will be **done automatically** in the next Collect release."
* "In v1.29, **the migration will be forced on install**."

Here is what I think will happen in the next version (v1.29):
* On app launch, if the migration hasn't successfully happened yet, the migration will be triggered and the full screen dialog with status information will be shown.
* The migration results will be shown in the full screen dialog so that the user sees the message.
* If the migration failed, the migration banner will still be displayed and will say something like "The last storage migration attempt failed. Please migrate manually."

Depending on timing, we'd either target SDK 29 with v1.29 or v1.30 (https://github.com/getodk/collect/issues/3360). I think that to play it safe we can set `requestLegacyExternalStorage` to `true` and keep it that way until we need to target SDK 30. Technically this means we could push back the automatic migration but the longer we wait the more new users have to go through the migration so I think we should do it as soon as possible.

CC @seadowg @grzesiek2010 (who I'm sure will be as thrilled as I am when we can fully move on from all this)

lognaturel:
Closes #3936

Ran it and verified that in particular the bold looks as expected.

We discussed the string in the issue. Adding the `setText(CharSequence)` method to `MaterialBanner` seems like the most straightforward way to add styling.

This is only a text change so there should be no risk.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
@@ -654,7 +654,7 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
 
     private void displayStorageMigrationBanner() {
         storageMigrationBanner.setVisibility(View.VISIBLE);
-        storageMigrationBanner.setText(getString(R.string.scoped_storage_banner_text));
+        storageMigrationBanner.setText(getText(R.string.scoped_storage_banner_text));
         storageMigrationBanner.setActionText(getString(R.string.scoped_storage_learn_more));
         storageMigrationBanner.setAction(() -> {
             showStorageMigrationDialog();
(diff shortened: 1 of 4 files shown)


grzesiek2010 (approved):

grzesiek2010:
I think we can merge this without testing since it's a change that will be visible in the Main Menu so that we can just test it on a daily basis. @getodk/testers please just keep in mind that we updated the banner.

PR #3955, opened 2020-07-02
Reworking Range Widget

SaumiaSinghal:
This PR includes the work on rewriting Range Widget.

The only behavioral change that is introduced in this PR is I used Slider instead of Seekbar, and I have to set the colors of Slider manually in the XML file as otherwise it shows `thumbColor` and `trackColor` as primary colors. The reason I used Slider was just that I wanted to use more of material components in my work. Also, I had to upgrade material library to [1.2.0-beta](https://github.com/material-components/material-components-android/releases) in `build.gradle` file to use the `Slider`. 

| Magenta Theme  | Light Theme | Dark Theme |
| magenta-theme-new-slider | light-theme-new-slider | dark-theme-new-slider  |
| magenta-theme-seekbar  | light-theme-seekbar  | dark-theme-seekbar  |

Any form having RangeIntegerWidget like AllWidgets form


collect_app/build.gradle
```diff
@@ -201,6 +201,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.2.0'
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation "androidx.work:work-runtime:2.3.4"
+    implementation 'com.google.android.material:material:1.3.0-alpha01'
```
seadowg: We shouldn't be using an alpha or beta version. Looking at https://maven.google.com/web/index.html it looks like the stable is currently 1.1.0 and they have a beta of 1.20 and an alpha of 1.3.0 out 😒

collect_app/src/main/java/org/odk/collect/android/widgets/RangeDecimalWidget.java
```diff
@@ -40,45 +112,62 @@ public IAnswerData getAnswer() {
     }
 
     @Override
-    protected void setUpActualValueLabel() {
+    public void setOnLongClickListener(OnLongClickListener l) {
+    }
+
+    @Override
+    public boolean suppressFlingGesture(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY) {
+        return suppressFlingGesture;
```
seadowg: There's a lot of code in here concerned with the `Slider`. I think it might make sense to try and extract a custom view to handle all this. Maybe a `RangeSlider`?

SaumiaSinghal:
I'm working on creating separate classes for Number Picker Range Widgets, as `RangePickerIntegerWidget` and `RangePickerDecimalWidget`.

collect_app/src/main/java/org/odk/collect/android/widgets/RangeWidget.java
```diff
@@ -45,10 +47,9 @@
 import static org.odk.collect.android.formentry.questions.WidgetViewUtils.createSimpleButton;
 
 @SuppressWarnings("BigDecimalMethodWithoutRoundingCalled")
-public abstract class RangeWidget extends QuestionWidget implements ButtonWidget, SeekBar.OnSeekBarChangeListener {
+public abstract class RangeWidget extends QuestionWidget implements ButtonWidget, Slider.OnChangeListener, Slider.OnSliderTouchListener {
```
seadowg: I don't think this needs to exist anymore right?

SaumiaSinghal: Yes, right

seadowg (changes requested):
Comments inline. I haven't reviewed the tests yet as I think extracting a custom view around the Slider will probably change up how things are structured.

collect_app/build.gradle
```diff
@@ -201,6 +201,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.2.0'
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation "androidx.work:work-runtime:2.3.4"
+    implementation 'com.google.android.material:material:1.2.0-beta01'
```
seadowg: How about a comment here that we need the beta version for `Slider`?

collect_app/src/main/java/org/odk/collect/android/views/CustomRangeSlider.java
```diff
@@ -0,0 +1,34 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.material.slider.Slider;
+
+import org.jetbrains.annotations.NotNull;
+
+public class CustomRangeSlider extends Slider implements  Slider.OnSliderTouchListener {
+
+    public boolean suppressFlingGesture;
```
seadowg: This should be a private field with a getter right? The problem you have otherwise is that something outside the `CustomRangeSlider` could set `suppressFlingGesture` which wouldn't make any sense - it's a value that only the `CustomRangeSlider` should be able to control (write to). 

Remember to think about encapsulation when adding fields/getters/setters to an object/interface.

seadowg: I think it'd also be good for this have a different name. The `CustomRangeSlider` is just tracking whether it's being swiped or not. It doesn't know what that will be used for. Try and think of a name that describes what this value is tracking. Thinking about where it is used is helpful - "the `RangeWidget` suppresses fling gestures (does not allow scrolling from one question to the other) when the slider X".

collect_app/src/main/java/org/odk/collect/android/widgets/RangeDecimalWidget.java
```diff
@@ -40,45 +76,36 @@ public IAnswerData getAnswer() {
...
     }
 
     @Override
-    protected void setUpDisplayedValuesForNumberPicker() {
-        displayedValuesForNumberPicker = new String[elementCount + 1];
-
-        if (isRangeIncreasing()) {
-            fillDisplayedValuesWithIncreasingValues();
+    public boolean suppressFlingGesture(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY) {
+        return slider.suppressFlingGesture;
```
seadowg: I like what is happening here! Just the naming and how the widget is able to access the field that is bothering me.

collect_app/src/main/java/org/odk/collect/android/widgets/RangeWidget.java
seadowg: Why is this still here?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
```diff
@@ -0,0 +1,118 @@
...
+            minValue = rangeWidgetHorizontalBinding.minValue;
+            maxValue = rangeWidgetHorizontalBinding.maxValue;
+        }
+        setUpWidgetParameters((RangeQuestion) prompt.getQuestion(), minValue, maxValue);
+
+        if (prompt.isReadOnly()) {
+            slider.setEnabled(false);
+        }
+
+        return new Object[] {answerView, slider, currentValue};
```
seadowg: This should probably just return a custom class rather than an array. You could nest it in here and it could be called something like `RangeWidgetLayoutElements`?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
```diff
@@ -0,0 +1,118 @@
...
+    public static Object[] setUpLayoutElements(Context context, FormEntryPrompt prompt) {
+        View answerView;
+        CustomRangeSlider slider;
+        TextView currentValue;
+        TextView minValue;
+        TextView maxValue;
+
+        String appearance = prompt.getQuestion().getAppearanceAttr();
+
+        if (appearance != null && appearance.contains(VERTICAL_APPEARANCE)) {
```
seadowg, line 59: It looks like there aren't any tests to check `VERTICAL_APPEARANCE` works.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
```diff
@@ -0,0 +1,118 @@
...
+        BigDecimal rangeStep = rangeQuestion.getRangeStep().abs();
+
+        slider.setValueFrom(rangeStart.floatValue());
+        slider.setValueTo(rangeEnd.floatValue());
+        slider.setStepSize(rangeStep.floatValue() != 0 ? rangeStep.floatValue() : 0.5F);
+        slider.setValue(actualValue == null ? rangeStart.floatValue() : actualValue.floatValue());
+
+        slider.setOnTouchListener((v, event) -> {
+            int action = event.getAction();
+            switch (action) {
```
seadowg, line 105: What is this code for?

seadowg (changes requested):
Comments inline. One larger point. I'm not sure you need all the tests for `RangeWidgetUtils` and the individual range widgets. Running both sets of tests it looks like there is some doubled coverage. It'd be good for you to have a think about what tests you need to drive out the code that exists. Potentially just having tests for the widgets is enough?

collect_app/src/main/java/org/odk/collect/android/views/CustomRangeSlider.java
SaumiaSinghal: `CusotmRangeSlider` does not reflect what the slider is doing here. Shall I rename it to `SuppressFlingGestureSlider`?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
SaumiaSinghal: @seadowg I think this just adds a touch listener on the `slider`. Without it, I was not able to set the value on the vertical slider. `requestDisallowInterceptTouchEvent()` does not allow the parent of the child to intercept the touch events acoording to the [docs](https://developer.android.com/reference/android/view/ViewParent). It disallows the parent `ViewGroup` to intercept swipe up and swipe down the action and lets the user set the values on the slider.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
SaumiaSinghal: Right @seadowg! I was having some difficulty in writing the tests for the `VERTICAL_APPEARANCE`. I think checking the rotation of the slider would be fine here.

SaumiaSinghal:
@seadowg, Regarding the test coverage I removed `usingReadOnlyOption_disablesTheSlider()`, `whenRangeQuestionHasZeroRangeStep_invalidWidgetToastIsShownAndSliderIsDisabled()` and  `whenPromptHasInvalidWidgetParameters_invalidWidgetToastIsShownAndSliderIsDisabled()` unit tests from `RangeIntegerWidget` and `RangeDecimalWidget`.  Other tests in `RangeWidgetUtils` concerns the `slider` view, I no longer check taht in the respective test classes of the `Rangewidget`

collect_app/src/main/java/org/odk/collect/android/views/SuppressFlingGestureSlider.java
```diff
@@ -0,0 +1,42 @@
...
+import android.util.AttributeSet;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.material.slider.Slider;
+
+import org.jetbrains.annotations.NotNull;
+
+public class SuppressFlingGestureSlider extends Slider implements  Slider.OnSliderTouchListener {
```
seadowg: I like the general pattern here. Do you think it'd make more sense for the value (and the class) to be named around `TrackingTouch...` rather than `SuppressFling...`?  The `Slider` doesn't actually suppress the gesture that's upstream from here.

collect_app/src/main/java/org/odk/collect/android/views/SuppressFlingGestureSlider.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public SuppressFlingGestureSlider(@NonNull Context context) {
+        super(context);
+    }
+
+    public SuppressFlingGestureSlider(@NonNull Context context, @Nullable AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    private void setSuppressFlingGesture(boolean suppressFlingGesture) {
```
seadowg: I'd maybe avoid a private setter and just go for setting the value directly. If you like it it's not a big deal to leave it in though!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2620,15 +2619,6 @@ public void onSaveFormIndexError(String errorMessage) {
 
     @Override
     public void onNumberPickerValueSelected(int widgetId, int value) {
```
seadowg, line 2621: Huh. I don't know if we can just delete the code in here. Is that behaviour not needed any longer?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
seadowg: Yeah that sounds good enough! Maybe just check in both cases all the details are rendered and that the `Slider` is rotated.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetUtils.java
seadowg: Hmmm. Yeah tricky. Could we maybe make this part of the custom `Slider` implementation? It's probably very awkward to test but at least if it's in there (and with a comment) it's more obviously important!

seadowg (changes requested):
Good progress! I've left some more comments inline and replied to your questions. Test coverage is looking pretty good and when we're feeling done with the structure I'll do a final pass of them!

collect_app/src/main/java/org/odk/collect/android/views/SuppressFlingGestureSlider.java
SaumiaSinghal: Oh okay yes right `Slider` is not suppressing the gesture here, it's just passing the value.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
SaumiaSinghal: I deleted it for this PR actually, since this concerns the range picker dialog. I had this back in Range Picker PR. is it fine?

seadowg:
@SaumiaSinghal looks like the quality checks are failing

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Ah. As much as possible we don't want to break things between PRs. Otherwise it means we can't release until the follow up PR comes in I now realize this PR actually removes the picker appearance. Is the "no-ticks" appearance gone as well? I think we should make sure to include all the Range widget work in this PR so we're not breaking anything when we merge.

SaumiaSinghal:
Right @seadowg, that is because there are some unused string resources, as I have removed `Range Picker` widgets in this PR.

Issue #3940, opened 2020-06-23
Update Form Management for Match Exactly


seadowg:
~~Blocked by #3939~~

Form Management settings should be updated to allow easy setup of Match Exactly. This means we can remove whatever mechanism we've been using to enable/disable it up until now.

**Mockup:** https://www.figma.com/proto/j2aYN2j2TKSOqHUHw14K9F/ODK-Collect-Server-Settings?node-id=91%3A69&viewport=683%2C400%2C1&scaling=min-zoom

**Acceptance**

- [x] Given I'm using a fresh install
When I go to the Main Menu
Then I see the "Get Blank Forms" button
And I can fetch forms normally

- [x] Given I'm using a fresh install
When I go to Form Management settings
Then I see "Manually" is selected for "Update forms"

- [x] Given I'm using a fresh install
When I go to Form Management settings
And I tap "Update forms"
And I select "Previously downloaded forms only"
Then "Automatic update frequency" is enabled
And automatic form updates work as before
And I can customize the update frequency

- [x] Given I'm using a fresh install
When I go to Form Management settings
And I tap "Update forms"
And I select "Exactly match server"
Then "Automatic update frequency" is enabled
And Match Exactly mode works as before
And I can customize the update frequency

- [x] Given I'm using a fresh install
When I go to Admin Settings
And then go to User Settings
And I uncheck "Blank form update mode"
And I go to General Settings
And I go to Form Management
Then I don't see the mode option

- [x] Given I've set Google Drive as my server
When I go to Form Management settings
Then I see "Manually" is selected for "Update forms"
When I tap "Update forms"
Then only "Manually" is enabled (I can't select the other options)

- [x] Given I've upgraded from a previous version of Collect
And I was using automatic updates
Then my automatic updates are still enabled

Issue #3939, opened 2020-06-23
Add periodic Match Exactly updates


seadowg:
~~Blocked by #3934~~

Once Match Exactly is enabled the update should occur automatically at a fixed interval (15 mins)

It probably makes sense to implement this using `WorkManager`.

**Acceptance**

- [x] Given I've enabled Match Exactly
And I've downloaded two forms onto my device
When I delete one of the forms on the server
And I update another form
And I add a new form
And I wait 15 minutes
Then the new form has been downloaded on my device
And the deleted form is gone
And the updated form is updated

- [x] Given I've disabled Match Exactly
When I add a new form on server
And I wait 15 minutes
Then the new form does not appear on my device

- [x] Given a sync is running in the background
When I go to Fill Blank Forms
Then the refresh button is disabled
And the progress bar is visible

PR #3956, opened 2020-07-02
Add form sync button to Fill Blank Forms

### LINKED ISSUE #3934: Add "Match Exactly" mode
This adds the enumerator's flow for the "Match Exactly" mode for Collect described [here](https://github.com/seadowg/collect-design/tree/master/prototypes/match-exactly). For this feature, instead of implementing the whole set of Form Management settings changes the mode should be enabled/disabled via a single checkbox that also enables/disables the "Form Update" section to prevent any weirdness.

**Acceptance**

- [x] Given I've enabled "Match Exactly"
When I visit the Main Menu
Then I don't see the "Get Blank Forms" button

- [x] Given I've enabled "Match Exactly"
When I visit Form Management settings
Then the "Form Update" section is disabled
When I disable "Match Exactly"
And I visit Form Management settings
Then the "Form Update" section is enabled

- [x] Given I've enabled "Match Exactly"
When I visit Fill Blank Forms
And I tap the ♻️ (refresh) icon
Then I see a progress indicator
And new forms on the server are downloaded to the device
And deleted forms on the server are removed from the device
And updated forms on the server are updated on the device

- [x] Given I've enabled "Match Exactly"
And I've uploaded new media files for a form that's on my device
When I visit Fill Blank Forms
And I tap the ♻️ (refresh) icon
Then I see a progress indicator
And the form's media files have been updates on the device

seadowg:
Closes #3934 

The match exactly feature can be enabled/disabled in settings by going to General settings > Experimental. This new Experimental section can be made beta/dev only if we end up releasing before we get to implementing the settings screen for Match Exactly (#3940).

This also:
* Adds new `FormRepository` and `MediaFileRepository` interfaces that can wrap the code we have in the `FormsDao` and content providers. Using these makes it simpler to interact with Forms and Media Files in the database/file system. We've also talked about how it would be nice to clean up the relationship between `FormsDao` and the content providers etc and using these interfaces protects them so that they can be refactored without change in too many places.
* Introduces a `FormAPI` interface that essentially represents the Open Rosa form API. This just makes it simpler to write and test code that has to hit the server. The implementation of this also lets us share code that was previously being duplicated. It might be worth pulling parser objects out of this in future.

There's a couple of inline comments about this but I'd thought about completely replacing the `FormListDownloader` with the `ServerFormsDetailsFetcher` (a use case I pulled out of it for the new `ServerFormListSynchronizer`). There isn't a lot of testing for the cases where `FormListDownloader` is used however, so I thought it was best to avoid diving into that work now.

In a similar vein,  there's an opportunity to clean up/replace the `FormDownloader` here but I'd like to on that and save it for later Match Exactly work to keep the size of the PRs down.

Another thing to note is that right now any errors that crop up during "syncing" are swallowed. There is a later issue to add error handling.

Played around with the new feature and double-checked that the standard "Get Blank Forms" flow still works. New feature also driven out with tests.

Major decisions discussed above and inline.

There are potentially risks to fetching forms in "Get Blank Forms" and automatic update but it probably makes sense to merge this and QA when the feature is complete (i.e. the proper settings page is in and automatic updates happen)


collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -135,11 +135,11 @@ WebCredentialsUtils provideWebCredentials() {
     }
 
     @Provides
-    ServerFormsDetailsFetcher provideDownloadFormListDownloader(
+    FormListDownloader formListDownloader(
```
seadowg, line 138: It felt like a good idea to bring this back as the `ServerPollingJob` and `DownloadFormListTask` were both using it and don't have a lot of tests around them. This would make it riskier to iterate on and improve the `ServerFormsDetailsFetcher`.

seadowg: I'll deprecate in a later commit.

collect_app/src/main/java/org/odk/collect/android/formmanagement/BlankFormsListViewModel.java
```diff
@@ -0,0 +1,78 @@
...
+import org.odk.collect.android.forms.FormRepository;
+import org.odk.collect.android.forms.MediaFileRepository;
+import org.odk.collect.android.openrosa.api.FormAPI;
+import org.odk.collect.android.openrosa.api.FormAPIError;
+import org.odk.collect.android.utilities.MultiFormDownloader;
+import org.odk.collect.async.Scheduler;
+
+import javax.inject.Inject;
+
+public class BlankFormsListViewModel extends ViewModel {
```
seadowg, line 18: I think eventually this should also encapsulate `DiskSyncTask`, hold the list of forms that the adapter needs (probably as `LiveData`) and be responsible for sorting.

collect_app/src/main/java/org/odk/collect/android/formmanagement/BlankFormsListViewModel.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static class Factory implements ViewModelProvider.Factory {
+
+        private final Scheduler scheduler;
+        private final FormRepository formRepository;
+        private final MediaFileRepository mediaFileRepository;
+        private final FormAPI formAPI;
+        private final MultiFormDownloader multiFormDownloader;
+
+        @Inject
```
seadowg, line 63: I had completely forgotten we could do this. I'm not usually a huge fan of `@Injects` scattered around the place randomly building the dependency graph for the app - as much as it's terse give me a 500 line module definition any day. However, in the case of `ViewModel` where we have to deal with Factory classes like this I think just being able to `@Inject` the constructor and then injecting it in the Activity cuts down on a lot of explicit dependencies at the `Activity` level that only get used for `ViewModel` construction.

Probably a nice trick to use elsewhere.

collect_app/src/main/java/org/odk/collect/android/dao/FormsDao.java
```diff
@@ -32,7 +32,7 @@
 import java.util.List;
 
 /**
- * This class is used to encapsulate all access to the {@link org.odk.collect.android.database.helpers.FormsDatabaseHelper#DATABASE_NAME}
+ * This class is used to encap`sulate all access to the {@link org.odk.collect.android.database.helpers.FormsDatabaseHelper#DATABASE_NAME}
```
lognaturel: typo

collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormAPI.java
```diff
@@ -1,12 +1,11 @@
 package org.odk.collect.android.openrosa.api;
 
 import org.javarosa.xform.parse.XFormParser;
+import org.jetbrains.annotations.NotNull;
```
lognaturel: Should use the androidx annotations

Edit: hmm, now I see we haven't been very consistent. I think in general we should use the androidx ones because of the android studio nullability analysis.

collect_app/src/main/java/org/odk/collect/android/dao/FormsDaoFormRepository.java
```diff
@@ -0,0 +1,35 @@
+package org.odk.collect.android.dao;
+
+import android.database.Cursor;
+
+import org.odk.collect.android.forms.Form;
+import org.odk.collect.android.forms.FormRepository;
+
+import java.util.List;
+
+public class FormsDaoFormRepository implements FormRepository {
```
lognaturel: I think this needs to be more analogous to `InstancesRepository`. For that interface, the concrete implementation is called `DatabaseInstancesRepository` and it lives in the `instances` package which I like slightly better. I don't mind this if you strongly prefer it, though, but I think the two should be the same. I think `DatabaseXRepository` captures what's really meaningful and contrasts from the test implementation -- that it's fetching X from a database.

collect_app/src/main/java/org/odk/collect/android/openrosa/api/FormAPI.java
```diff
@@ -0,0 +1,10 @@
+package org.odk.collect.android.openrosa.api;
+
+import java.util.List;
+
+public interface FormAPI {
```
lognaturel: API should be Api: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

Would you consider `FormListApi`? That links it more clearly to the corresponding OpenRosa standard https://docs.getodk.org/openrosa-form-list/. It makes sense to me that fetching the manifest is a subset of fetching the form list.

collect_app/src/main/java/org/odk/collect/android/openrosa/api/FormAPIError.java
```diff
@@ -0,0 +1,30 @@
+package org.odk.collect.android.openrosa.api;
+
+public class FormAPIError extends Exception {
```
lognaturel: `FormApiError`

collect_app/src/main/java/org/odk/collect/android/dao/FormsDaoFormRepository.java
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+
+    @Override
+    public void save(Form form) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void delete(Long id) {
```
lognaturel: At https://forum.getodk.org/t/have-collect-exactly-match-the-forms-on-central/23101, we had described doing a soft delete so that saved records could still be edited. Unfortunately that does need adding to the db schema, I think. Maybe something to discuss.

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaXMLFetcher.java
```diff
@@ -21,15 +21,15 @@
 
 import timber.log.Timber;
 
-public class OpenRosaAPIClient {
+public class OpenRosaXMLFetcher {
```
lognaturel: Should be `OpenRosaXmlFetcher`

collect_app/src/main/java/org/odk/collect/android/utilities/MultiFormDownloader.java
```diff
@@ -56,26 +58,28 @@
 import static org.odk.collect.android.utilities.FileUtils.STUB_XML;
 import static org.odk.collect.android.utilities.FileUtils.write;
 
-public class FormDownloader {
+public class MultiFormDownloader implements FormDownloader {
```
lognaturel, line 61: why not in formmanagement?

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MatchExactlyTest.java
```diff
@@ -0,0 +1,93 @@
...
+                .enableMatchExactly()
+                .clickFillBlankForm()
+                .assertText("One Question")
+                .assertText("One Question Repeat")
+                .clickRefresh()
+                .assertText("Two Question") // Check new form downloaded
+                .assertText("One Question Updated") // Check updated form updated
+                .assertTextDoesNotExist("One Question Repeat"); // Check deleted form deleted
+    }
+
```
lognaturel, line 78: Need test and implementation to hide refresh button when setting is off.

collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormAPI.java
```diff
@@ -0,0 +1,325 @@
...
+
+import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.odk.collect.android.openrosa.api.FormAPIError.Type.AUTH_REQUIRED;
+import static org.odk.collect.android.openrosa.api.FormAPIError.Type.FETCH_ERROR;
+import static org.odk.collect.android.openrosa.api.FormAPIError.Type.PARSE_ERROR;
+
+public class OpenRosaFormAPI implements FormAPI {
```
lognaturel: Consider `OpenRosaFormListApi`

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private ManifestFile getManifestFile(FormAPI formAPI, String manifestUrl) {
+        if (manifestUrl == null) {
+            return null;
+        }
+
+        try {
+            return formAPI.fetchManifest(manifestUrl);
+        } catch (FormAPIError formAPIError) {
+            Timber.e(formAPIError);
```
lognaturel: Change to `w` so we don't forget later? I don't think any of these would point to a software error we can fix.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormListSynchronizer.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public void synchronize() throws FormAPIError {
+        ServerFormsDetailsFetcher listDownloader = new ServerFormsDetailsFetcher(formRepository, mediaFileRepository, formAPI);
+        List<FormDetails> formList = listDownloader.fetchFormDetails();
+
+        List<Form> formsOnDevice = formRepository.getAll();
+
+        formsOnDevice.stream().forEach(form -> {
+            if (formList.stream().noneMatch(f -> form.getJrFormId().equals(f.getFormId()))) {
```
lognaturel, line 32: This is n^2, right? Both lists are going to be relatively small in this case so I think it's ok but we should be careful if doing similar things with e.g. instances at some point. I feel like streams can hide complexity in a kind of scary way sometimes.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
```diff
@@ -0,0 +1,144 @@
...
+            return null;
+        }
+    }
+
+    private boolean isNewerFormVersionAvailable(String md5Hash) {
+        if (md5Hash == null) {
+            return false;
+        }
+
+        return formRepository.getAll().stream().noneMatch(f -> f.getMD5Hash().equals(md5Hash));
```
lognaturel: Consider adding a repository method to get a form by md5hash? Seems clearer and more efficient.

collect_app/src/main/java/org/odk/collect/android/logic/FormDetails.java
```diff
@@ -87,4 +90,48 @@ public boolean isNewerFormVersionAvailable() {
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(getErrorStr(), getFormName(), getDownloadUrl(), getManifestUrl(), formID, getFormVersion(), getHash(), getManifestFileHash(), isNewerFormVersionAvailable(), areNewerMediaFilesAvailable);
+    }
+
+    public static FormDetails toFormDetails(FormListItem formListItem) {
+        return toFormDetails(formListItem, null, false, false);
+    }
+
+    public static FormDetails toFormDetails(FormListItem formListItem, String manifestFileHash, boolean isNewerFormVersionAvailable, boolean areNewerMediaFilesAvailable) {
```
lognaturel: I'm not loving all this. Isn't `FormDetails` a wrapper for a `FormListItem` that provides status information? Maybe a `FormDetails` should have a `FormListItem` field? Maybe it should be `FormListItemStatus`?

collect_app/src/main/java/org/odk/collect/android/logic/FormDetails.java
```diff
@@ -87,4 +90,48 @@ public boolean isNewerFormVersionAvailable() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        FormDetails that = (FormDetails) o;
+        return isNewerFormVersionAvailable() == that.isNewerFormVersionAvailable() &&
```
lognaturel: Is all this really necessary? Wouldn't comparing the hash and the error string be sufficient? The download URL would also be needed if somehow you could have downloads from multiple servers going on at the same time but I don't think we have any intention of making that possible. It seems weird to compare these status objects, anyway. Where is this used? Or is it just hashing that ends up getting used?

lognaturel (changes requested):
Things look good at a high level. I'd like to spend a little more time with how `FormListDownloader` has been split up but want to get you preliminary thoughts. The most important things are that the sync button shouldn't exist when match exactly is off and that delete should be soft-delete (we may want to talk through strategy).

collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormAPI.java
seadowg: Hmmm. I think here I actually prefer the JetBrains one as this code has no other Android dependencies. Either work within Android Studio (as it's just IntelliJ under the hood). Does that sound ok?

collect_app/src/main/java/org/odk/collect/android/openrosa/api/FormAPI.java
seadowg: So I've had feedback when pairing with newer or English-as-a-second-language engineers that the uppercasing helps identify acronyms they don't know, so I tend to go with that. However, we do reference that style guide and I don't want to get into debating exceptions! I'll fix.

collect_app/src/main/java/org/odk/collect/android/openrosa/api/FormAPIError.java
seadowg: I'm also going to rename this to `...Exception` while I'm here. I'm way to cavalier around the difference between those.

collect_app/src/main/java/org/odk/collect/android/dao/FormsDaoFormRepository.java
seadowg: For some reason I'd assumed that was what was happening from how we were talking about. Good to add something like this at an acceptance level I think so it's really clear what we mean. I think looking at this as a separate issue would be good!

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MatchExactlyTest.java
seadowg: 🤭 talk about getting caught up in what you're working on! Definitely needs fixed.

collect_app/src/main/java/org/odk/collect/android/utilities/MultiFormDownloader.java
seadowg: I think I didn't feel like it was ready to go there yet as we still haven't taken a stab at cleaning it up? Maybe that's silly. What do you think?

collect_app/src/main/java/org/odk/collect/android/dao/FormsDaoFormRepository.java
seadowg: I'd managed to miss/forget about `InstancesRepository` but totally would have followed it's pattern if I hadn't. Will make those consistent.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
seadowg: Yeah that seems like a nice optimization!

collect_app/src/main/java/org/odk/collect/android/logic/FormDetails.java
seadowg: Yeah actually this is something I meant to come back to. Initially I'd thought I'd be getting rid of `FormDetails` so this was a way of converting between those values while I worked towards that. You're right that it seems really funky now. I'll take a look at cleaning it up.

collect_app/src/main/java/org/odk/collect/android/logic/FormDetails.java
seadowg: This is just using the Android Studio equals/hash generator. Probably good to add a comment when I do that. Generally I'd argue for "data" objects like this it makes sense to just compare the whole thing (like we'd get with Lombok or Kotlin) unless we have a very good reason to do otherwise.

I generated this for testing with mockito but I ended up converting to a fake (which makes life easier) so I can just kill this code.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormListSynchronizer.java
seadowg: Interesting. I tend to find the opposite is true as it feels clearer where the loops are happening. 

This is set difference so without sorting I don't think we can do better than O(n^2) right? I guess maybe using a `HashMap` or `HashSet` could get us O(1) complexity (in theory) on remove operations which gives us linear time. I need more tea ☕️

collect_app/src/main/java/org/odk/collect/android/dao/FormsDaoFormRepository.java
lognaturel: Filed https://github.com/GetODK/collect/issues/3961. You should be able to edit the issue as needed!

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MatchExactlyTest.java
```diff
@@ -0,0 +1,105 @@
...
+                .clickOnMenu()
+                .clickGeneralSettings()
+                .clickFormManagement()
+                .assertDisabled(R.string.periodic_form_updates_check_title)
+                .assertDisabled(R.string.automatic_download)
+                .assertDisabled(R.string.hide_old_form_versions_setting_title);
+    }
+
+    @Test
+    public void whenMatchExactlyEnabled_fillBlankFormRefreshButtonIsGone() {
```
lognaturel: This should be `whenMatchExactlyNotEnabled` or `whenMatchExactlyDisabled` or whatever!

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormRepository.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public boolean contains(String jrFormID) {
+        try (Cursor cursor = new FormsDao().getFormsCursorForFormId(jrFormID)) {
+            return cursor != null && cursor.getCount() > 0;
+        }
+    }
+
+    @Override
+    public List<Form> getAll() {
```
lognaturel, line 21: In our live review we talked about in a follow-up PR considering changing this to a `Set`. There's no reason for the results to be ordered and sets will make some client code cleaner. We also talked about replacing `contains` with a get by jrFormId (which should also return a set with potentially multiple versions). No action needed now but writing it down so we don't forget.

collect_app/src/main/java/org/odk/collect/android/utilities/FormListDownloader.java
```diff
@@ -110,213 +80,38 @@ public FormListDownloader(
...
-                                        R.string.parse_legacy_formlist_failed, error)));
-                        return formList;
-                    }
-                    formList.put(formName,
-                            new FormDetails(formName, downloadUrl, null, formId, null, null, null, false, false));
-
-                    formId = null;
-                }
+        } catch (FormApiException formApiException) {
+            Timber.e(formApiException);
```
lognaturel: Not actionable so let's do `Timber.w` or remove since it goes to the UI.

collect_app/src/main/java/org/odk/collect/android/utilities/FormListDownloader.java
```diff
@@ -110,213 +80,38 @@ public FormListDownloader(
...
+
+            switch (formApiException.getType()) {
```
lognaturel, line 97: I couldn't really come up with a reliable and efficient way to verify that you addressed the exact same exception list when extracting `OpenRosaFormListApi`. That means I'm just trusting you on that and it seems like a potential source of risk.

lognaturel (changes requested):
I've spent a bit more time with this and am generally feeling like I understand and like the new abstractions. There's a test that broke at some point and a couple other tiny comments and then this is ready for a merge!

High-level question, you use the phrase "use case" when talking about `ServerFormsDetailsFetcher`. I see what you mean but I'm not familiar with that usage, does it come from somewhere?

seadowg:
> There's a test that broke at some point and a couple other tiny comments and then this is ready for a merge!

Huh. Looks like this is not breaking locally. That's a little scary. Definitely worth investigating before moving on.

> High-level question, you use the phrase "use case" when talking about ServerFormsDetailsFetcher. I see what you mean but I'm not familiar with that usage, does it come from somewhere?

So I used to just refer to these as "thing doers" or "noun verbers". Basically just an object responsible for some logic. Outside OO it would probably just be a function. It turns out a certified bad dude came up with the name "use case" for these but I'm not going to link to that (because no one needs him in their life). I imagine the "origin" of the term really sits in [UML](https://en.wikipedia.org/wiki/Use_case). Basically the idea is that a "use case" sits between the entities, controller etc in your system architecture, is independent of any frameworks your application may depend on and is responsible for some piece of logic of your application. Does that make sense?

collect_app/src/main/java/org/odk/collect/android/utilities/FormListDownloader.java
seadowg: Yeah I agree this is a source of risk!

lognaturel:
> Looks like this is not breaking locally.

Is that right? It failed for me. The fix makes sense, either way.

> a certified bad dude came up with the name "use case" for these

Hah. That was enough context for me to find what I was looking for, thanks. I found http://www.plainionist.net/Implementing-Clean-Architecture-UseCases/ and https://softwareengineering.stackexchange.com/questions/166462/when-is-a-use-case-layer-needed good reads. I've seen "use case" in clean architecture writings before but I was thinking of it in the requirements/UML sense and didn't realize it was referring to a type of class/layer analogous to domain classes.

I've discussed with @seadowg, @grzesiek2010, @getodk/testers and we've decided to merge this without QA. This is one component of the overall feature that is split out for easier review. It only introduces an experimental setting and there is minimal regression risk. All of the behavior touched in this PR will need to be QAed once the feature is complete, anyway.

Follow up work includes:
- https://github.com/getodk/collect/issues/3959
- https://github.com/getodk/collect/issues/3939
- https://github.com/getodk/collect/issues/3958
- https://github.com/getodk/collect/issues/3842

lognaturel (approved):

PR #3960, opened 2020-07-07
Automatically sync with server when Match Exactly enabled

### LINKED ISSUE #3939: Add periodic Match Exactly updates

seadowg:
Closes #3939 

Lots of new tests and played with automatic sync manually on the emulator.

The biggest decision here was whether to just use `WorkManager` as is or include it in our `async.Scheduler` abstraction. I ended up going with the latter as it means we can write tests for the automatic sync we wouldn't be able to write otherwise  (like those found in `MatchExactlyTest`). This does mean we end up with a weighty abstraction in `TaskSpec`. I'm planning the other background jobs to using this as part of #3940.

As with previous Match Exactly PRs we're going to skip QA on this and wait until the feature is complete.


collect_app/src/main/java/org/odk/collect/android/audio/AudioPlayerViewModel.java
```diff
@@ -19,10 +19,11 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.function.Supplier;
 
 class AudioPlayerViewModel extends ViewModel implements MediaPlayer.OnCompletionListener {
 
-    private final MediaPlayerFactory mediaPlayerFactory;
+    private final Supplier<MediaPlayer> mediaPlayerFactory;
```
seadowg, line 26: I ran into a problem with the `audio` stuff due to it instantiating a `Scheduler` so had to do a slight rework. I like the idea of using `Supplier` instead of an explicit `...Factory` or `...Provider` type now the new desugaring gives us access to it. We might still need a concrete class for Dagger but everything else should be able to use `Supplier`.

collect_app/src/main/java/org/odk/collect/android/utilities/MultiClickGuard.java
```diff
@@ -13,6 +14,10 @@ private MultiClickGuard() {
 
     // Debounce multiple clicks within the same screen
     public static boolean allowClick(String className) {
+        if (test) {
```
seadowg, line 17: I ran into problems with this blocking clicks in tests locally

async/src/main/java/org/odk/collect/async/CoroutineAndWorkManagerScheduler.kt
```diff
@@ -1,8 +1,9 @@
 package org.odk.collect.async
 
-import androidx.work.WorkInfo
-import androidx.work.WorkManager
+import android.content.Context
+import androidx.work.*
```
lognaturel, line 3: No wildcard imports: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

This isn't caught by static analysis because it's in another module, right? Perhaps we should make sure all those checks are consistent?

collect_app/build.gradle
```diff
@@ -256,8 +256,10 @@ dependencies {
     implementation "org.slf4j:slf4j-api:2.0.0-alpha0"
     implementation "com.arcao:slf4j-timber:3.1@aar"
 
-    implementation "com.google.zxing:core:3.3.0" //updating to 3.4.0 requires minSdkVersion >=24 https://github.com/zxing/zxing/releases/tag/zxing-3.4.0
-    implementation "com.journeyapps:zxing-android-embedded:3.6.0" //updating would require minSdkVersion >=24
+    implementation "com.google.zxing:core:3.3.0"
+    //updating to 3.4.0 requires minSdkVersion >=24 https://github.com/zxing/zxing/releases/tag/zxing-3.4.0
```
lognaturel, line 259: Either revert or move above related code.

collect_app/build.gradle
```diff
@@ -270,7 +272,8 @@ dependencies {
     // Dagger:
     implementation "com.google.dagger:dagger:${rootProject.daggerVersion}"
     annotationProcessor "com.google.dagger:dagger-compiler:${rootProject.daggerVersion}"
-    testAnnotationProcessor "com.google.dagger:dagger-compiler:${rootProject.daggerVersion}" // Needed for injection into Robolectric tests
+    testAnnotationProcessor "com.google.dagger:dagger-compiler:${rootProject.daggerVersion}"
+    // Needed for injection into Robolectric tests
```
lognaturel, line 278: Either revert or move above

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDetails.java
```diff
@@ -28,25 +28,25 @@
     private String formVersion;
     private String hash;
     private String manifestFileHash;
-    private boolean isNewerFormVersionAvailable;
-    private boolean areNewerMediaFilesAvailable;
+    private boolean isNew;
```
lognaturel: `isNotOnDevice`?

collect_app/src/main/java/org/odk/collect/android/audio/AudioPlayerViewModel.java
lognaturel: Sounds good. Add a note in the "state of the union"?

collect_app/src/test/java/org/odk/collect/android/formmanagement/BlankFormsListViewModelTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        BlankFormsListViewModel viewModel = new BlankFormsListViewModel(mock(Scheduler.class), syncRepository, mock(ServerFormsSynchronizer.class), mock(PreferencesProvider.class));
+        LiveData<Boolean> syncing = viewModel.isSyncing();
+
+        syncRepository.startSync();
+        assertThat(syncing.getValue(), is(true));
+    }
+
+    @Test
+    public void syncWithServer_starsSyncOnRepository() {
```
lognaturel: stars -> starts?

collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkManager.java
```diff
@@ -0,0 +1,12 @@
+package org.odk.collect.android.backgroundwork;
+
+public interface BackgroundWorkManager {
+
+    boolean isFormUploaderRunning();
```
lognaturel, line 5: Feels weird that this doesn't also schedule and cancel form upload and download but I assume it'll come.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsSynchronizerTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    public void downloadsUpdatedForms() throws Exception {
+        when(serverFormDetailsFetcher.fetchFormDetails()).thenReturn(asList(
+                new ServerFormDetails("form-1", "http://example.com/form-1", null, "form-1", "server", "md5:form-1-hash", null, false, true)
+        ));
+
+        synchronizer.synchronize();
+        assertThat(formDownloader.getDownloadedForms(), containsInAnyOrder("form-1"));
```
lognaturel, line 55: Why not cover a little more ground and have a form-1 with some form version in the repository before the synchronization and then assert that the repository contains the form with the new version? That would better match the level of the delete test below, too.

lognaturel (changes requested):
Couple small things!

async/src/main/java/org/odk/collect/async/CoroutineAndWorkManagerScheduler.kt
seadowg: Ooooft. Didn't realize we were missing those checks. Good catch. I'll fix in the next PR (for #3940)

collect_app/build.gradle
seadowg: I think that's a weird reformat problem. Will just fix.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkManager.java
seadowg: Yeah it's on the way!

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsSynchronizerTest.java
seadowg: Totally agree the test boundaries are inconsistent here. Need to do some refactoring of `MultiFormDownloader` to fix that though. That's something I'd like to fix up in the next PR as I'm going to have to touch the auto download code.

lognaturel (approved):

lognaturel:
@seadowg can you please keep a running list of sources of risk so that QA can do a regression check once the feature is complete? Here I see:
- audio in questions
- pushing forms from adb and having them show up in the right places

seadowg:
@lognaturel that's a good shout. I'll keep a list in my local notes and then post when we get to doing QA on `master`

Issue #3297, opened 2019-08-07
Minimal selects much slower because of HTML styling


lognaturel:
#### Software and hardware versions 
Collect v1.23.0

#### Problem description
Identified by @JohnTheBeloved.

#3204 renders styled options for minimal selects. The additional step of styling each option slows things down significantly because it's done for every option upfront rather than with the RecyclerView-based select implementations when it's done as an option is displayed. 

#### Steps to reproduce the problem
Load a form with hundreds of choices in a select one/multiple question with the minimal appearance.

#### Expected behavior
Performance is the same or better as in v1.22.0 and prior.

lognaturel:
This is potentially a bad regression for users that rely on minimal selects from external files so it could be a candidate for a point release.

This was never very fast to begin with because spinners build every option view at once (as opposed to recycling them as with the `RecyclerView`). I'm not sure how we evaluate how severe this is. @grzesiek2010, what are your measurements of how big the performance hit is?

In the longer term, I think that it would be ideal for minimal selects to use `RecyclerView` as well. Can we use a custom layout to fake a spinner?

grzesiek2010:
> what are your measurements of how big the performance hit is?

it depends, I used Android 5 and a form with 100 choices and it was like 0.5s -> 2.2s
but it will work faster on newer devices but also the more choices the slower performance.

> In the longer term, I think that it would be ideal for minimal selects to use RecyclerView as well. Can we use a custom layout to fake a spinner?

It might be difficult to do that with both the spinner (select_one) and alert dialog with multiple choices (select-multiple). I think our goal should be to create our own dialog which would use RecyclerView and support both select_one and select_multiple. Then we could even use MediaLayouts in such a dialog supporting images/audios/video.

It would be a bigger change so not for now and that's why I can see two options: to leave it as is or to revert. I'm not sure what's better...

grzesiek2010:
Fixed in https://github.com/getodk/collect/pull/3964

lognaturel:
Nice.

PR #3964, opened 2020-07-09
Add support for combination of minimal and autocomplete appearances

### LINKED ISSUE #3736: Add support for combination of minimal and autocomplete appearances
- [x] @seadowg facilitates [design studio](https://miro.com/app/board/o9J_kuoswaE=/) with @grzesiek2010 and @lognaturel to clarify need and design decisions to make
- [x] @seadowg makes higher-fidelity mockup to evaluate usability and feasibility
- [x] share mockup and implementation plan [on the forum](https://forum.opendatakit.org/t/showing-search-option-with-minimal-in-appearance/20758)
- [ ] update issue with mockup, edge cases, etc

grzesiek2010:
Closes #3736 #3297 

I tested select widgets manually.

This is what our users requested and what we decided to implement.
When it comes to the implementation I used RecyclerView like in all other selects (excluding those horizontal: LikertWidget/ListWidget/ListMultiWidget), thanks to that we can mix minimal with the same appearances we use in normal selects: columns-pack/columns-n/no-buttons etc.

We need to test all select widgets that use RecyclerView (basicly all select widgets excluding those horizontal: LikertWidget/ListWidget/ListMultiWidget).

This change also enables mixing more appearances and now we can mix minimal widgets not only with autocomplete (what was the purpose of the issue) but also with other appearances we use in normal selects like: no-buttons/columns-pack/columns-n

|  | | | |
| ------------- | ------------- | ------------- | ------------- |
| Screenshot_2020-07-20-13-43-27-363_org odk collect android  | Screenshot_2020-07-20-13-43-37-387_org odk collect android  | Screenshot_2020-07-20-13-43-47-299_org odk collect android  | Screenshot_2020-07-20-13-43-56-397_org odk collect android  |

that's why I said that we need to test all possible selects with all possible appearances. I know it's not easy since we can mix a lot of appearances.

Any form with any select widgets.
I'm attaching some that I used:
[selectAppearances.zip](https://github.com/getodk/collect/files/4947517/selectAppearances.zip)
[selectAppearancesWithMinimal.zip](https://github.com/getodk/collect/files/4947518/selectAppearancesWithMinimal.zip)

Yes we need to update screenshots.


grzesiek2010:
@lognaturel @seadowg I need to add some final changes and tests probably but could you play with it for a minute and let me know if there is anything I should change? You can use just the All widgets form.

grzesiek2010:
@seadowg @lognaturel as I said above probably I need to think of adding some more tests (the more the better) but it would be good to hear from you at least about the UI.

lognaturel:
Sorry about the delay and hope you had a good week away! This looks great. I hadn't considered that it would allow both `minimal` and `autocomplete` to combine with other appearances. It's always been confusing to know what appearances can combine so that's great. The downside is that now there are a lot more select views than before so careful automated tests that will save ~us~ @getodk/testers from having to manually try the full cross product of appearances is really important!

I didn't immediately know what to do with the input line that looks disabled. In the case of `minimal` and `autocomplete` combined, if that field is the first on a screen, could the field have focus with the keyboard displayed? The full screen dialog would then appear when the user starts typing. That way it would feel like a text field with autocomplete. That's what the original requester described [on the forum](https://forum.getodk.org/t/showing-search-option-with-minimal-in-appearance/20758?u=ln) so it seems there's a precedent for it.

For `minimal` without autocomplete, I think we should keep the same look as on `master`. That's what Material Design's [exposed dropdown menu](https://www.material.io/components/menus#exposed-dropdown-menu) does with a ▼ at the right of the text field. We may even want to keep the text `Select One Answer`. What made you want to remove that?

I think that making a selection should exit the full screen dialog the same way that on `master` making a selection in a `minimal` select collapses the spinner. It's particularly inconvenient on a large screen to have to go all the way up to the top to exit.

On Android 5.1, the background color of the full screen dialog is not the same as the background color of the form entry activity. Seems they should match.

The font size for the select choice should match the font size of text in a text field (and be dynamic in the same way, I didn't check that). It looks smaller on my device.

I get a `SelectOneMinimalWidget.recyclerViewAdapter` leak on Android 5.1.

Other than that, the full screen dialog looks good to me both with and without autocomplete.

grzesiek2010:
> I didn't immediately know what to do with the input line that looks disabled. In the case of minimal and autocomplete combined, if that field is the first on a screen, could the field have focus with the keyboard displayed? The full screen dialog would then appear when the user starts typing

Theoretically it is possible but then if you select for example answer `option1` it should be displayed in that box (after closing the dialog) so if a user opens the dialog again they could expect only that `option1` to be displayed because that's the phrase from the search box? it's weird... maybe we shouldn't display answers in the box but below?

> For minimal without autocomplete, I think we should keep the same look as on master. That's what Material Design's exposed dropdown menu does with a ▼ at the right of the text field. We may even want to keep the text Select One Answer. What made you want to remove that?

I somehow missed it being too focused on the new minimal + autocomplete. I agree that if we have just minimal that search box doesn't make sense. To be clear you just want to have something like a spinner button but not the old spinner widget? I mean it will look like a spinner but it will open the same dialog I use for autocomplete right because it doesn't make sense to keep those old widgets for minimal appearances without autocomplete? What about select multi? it had a normal button.

> I think that making a selection should exit the full screen dialog the same way that on master making a selection in a minimal select collapses the spinner. It's particularly inconvenient on a large screen to have to go all the way up to the top to exit.

I agree. The dialog should be closed in if we use select_one.

> The font size for the select choice should match the font size of text in a text field (and be dynamic in the same way, I didn't check that). It looks smaller on my device.

The font size of every single option is dynamic if that's what you are asking for.

lognaturel:
@grzesiek2010 and I had a brief chat about this and agreed to fake a spinner look for the inline field (basically add ▼ at the right) for all selects with `minimal` appearance regardless of whether they autocomplete or not. We also agreed to put the selected value(s) in the text field, even in the case of multiple selects. That increases consistency across the board.

> if you select for example answer option1 it should be displayed in that box (after closing the dialog) so if a user opens the dialog again they could expect only that option1 to be displayed because that's the phrase from the search box

Agreed this is totally weird and I hadn't thought through the case where something was already selected. Having considered it more, I think that adding the ▼ at the right will likely make it clear enough that the field is tappable. So basically there would be no difference between `minimal` and `minimal autocomplete` within the form. It's only after tapping the text field that there would be a difference. For `minimal autocomplete`, there would be a search field with focus. For `minimal` there would just be the list of items. How does that sound, @grzesiek2010? Might be good for you to take a quick look too, @seadowg, since you did the initial sketches. The two questions we're trying to answer are "how do we make it obvious that for a select with the `minimal` appearance, tapping the field will have some effect" and "should there be any different behavior inline in the form filling view if the `autocomplete` appearance combines with `minimal."

grzesiek2010:
Having a box that would look like spinner for both minimal and minimal autocomplete appearances seems right to me but yeah let's hear from @seadowg

seadowg:
@grzesiek2010 @lognaturel looking at it feels to me like a fake `Spinner` for just `minimal` and a fake `EditText` (with focus where appropriate) for `minimal autocomplete` makes sense. I think in both cases we should use the real Android views but just override the click behaviour. On the `EditText` we could have it open the selection screen when anything is typed so it works properly when focused I think?

Maybe just using a `Spinner` for both is a good first step though and then if we feel like we want to make it slicker later we can?

grzesiek2010:
So basically you don't agree with what we proposed above (having spinner for both cases) @seadowg? My problem with that edittext is like:
- it is not intuitive 
- I don't think that we need to have different widget layouts for minimal vs minimal autocomplete, displaying the search box in the dialog would be enough from a user perspective 
- as I said above displaying answers in such edittext is confusing

So I would be in favour of having spinners in all cases.

seadowg:
> So basically you don't agree with what we proposed above (having spinner for both cases) @seadowg?

No I think it works. Sorry I think I was being confusing! I was trying to explain that my ideal would be having the `minimal autocomplete` look like a standard text widget but when you click on it or type it opens the select screen (like in the [mockup](https://www.figma.com/proto/Dxe1UF9reRdLYy8o5jishW/Minimal-automcomplete-select?node-id=24%3A1774&viewport=826%2C271%2C0.500446617603302&scaling=min-zoom)). I get that this is probably awkward to implement - it would require making sure you'd handled click typing listeners on the `EditText` and could smoothly pass the keys typed from one view to the other. 

Given all that I'm happy to go with a `Spinner` for both.

lognaturel:
@seadowg it’s not about implementation. We just can’t figure out reasonable behavior when something is already selected. The behavior is clear and good when nothing is selected: the text field has focus if it’s the first question on the screen and then if you start typing it opens the dialog and filters the list according to what you’ve typed. But what if you have a fruit list and you’ve already selected “apple”? The edit text will have “apple” in it, right? So then where should the cursor be if the field has focus? What should happen when you start typing?

One new idea I just had is that if there’s already a selection the contents of the edit text could be selected when it gets focus. That way, when you start typing it would start a new search.

grzesiek2010:
Do you like this? I do
Screenshot_1595430687

seadowg:
@grzesiek2010 I like that! Is that just the standard `Spinner`? I think if possible it'd be nice to avoid any custom views.

@lognaturel in my head this all works but maybe I'm crazy. If something is selected it would the cursor would work just like it would with a standard text field and if you start typing (while focused) or click on it the selection screen would pop up and you would continue there. Again though as I've been thinking about it seems really awkward to implement this smoothly! I guess I'd thought about it like a text field that you can only populate with specific answers. Anyway I think the direction we're going in is the pragmatic one so enough about my mad scientist approach.

lognaturel:
> if you start typing (while focused) or click on it the selection screen would pop up

Let’s say the default or previously selected value is “apple” and you want to change it to “banana.” The question is the only one on the screen. If it were a regular edit text, we’d give it focus and put the cursor at the end. But like @grzesiek2010 mentioned if that were the case and you typed “b” then you’d end up with a filter of “appleb” and an empty filtered list in the dialog. Selecting the whole edit text content so it gets replaced when the user starts typing makes some amount of sense for select ones. It doesn’t really work for select multiples where you might want to change a subset of choices (Edit: I don't know, maybe it still makes sense? Then what you're typing is clearly the filter query, not the full value of the field). I think we’d need to have a really precise behavior description for both those cases to consider something other than forcing the user to tap to get to choices in all cases.

grzesiek2010:
> Is that just the standard Spinner?

it is just a text field + image view: https://github.com/getodk/collect/pull/3964/files#diff-731643aa399535acb9f8639380ec98cc

I have also some concerns about possible ways of closing the dialog/accepting selected answers, for now it is possible to close the dialog clicking on the back arrow (from the toolbar) or clicking the device back button. In both cases selected answers are saved (accepted). Do you find it confusing and think that maybe we shouldn't accept new answers when the dialog is closed by the device back button (treat it as cancel button)? On the other hand that back arrow also looks like a way to click cancel maybe it should be :heavy_check_mark: ?

seadowg:
> it is just a text field + image view: https://github.com/getodk/collect/pull/3964/files#diff-731643aa399535acb9f8639380ec98cc

Hmmm I think it would be nice to avoid custom UI if possible (as much as I actually like what you've created more than the Android Spinner style). Was there a specific reason to not use an actual `Spinner`?

grzesiek2010:
> Hmmm I think it would be nice to avoid custom UI if possible (as much as I actually like what you've created more than the Android Spinner style). Was there a specific reason to not use an actual Spinner?

A few reasons:
1. The original spinner looks worse to me.
2. Adding a spinner just to have its layout not functionality (because of course we have our own dialog) is a bit strange
3. It would be more difficult to maintain, it would require an adapter or something that fakes it and changing answers would be more difficult compared to the edit text i use where I just call setText().

lognaturel:
> for now it is possible to close the dialog clicking on the back arrow (from the toolbar) or clicking the device back button. In both cases selected answers are saved (accepted)

This sounds right to me. In the case of a select one, I think it's the only sensible option. Tapping an option selects it and closes the dialog. The back buttons just leaves whatever selection there already might be.

In the case of select multiples, it's a little less clear. But I think that if all ways of exiting the dialog commit the current selections, it's at least consistent and easy to reason about.

seadowg:
> It would be more difficult to maintain, it would require an adapter or something that fakes it and changing answers would be more difficult compared to the edit text i use where I just call setText().

Great point. I think I'm convinced using the custom spinner is less of a maintenance risk than using a real one and faking its setup 👍.

grzesiek2010:
@lognaturel I implemented improvements and fixes I wanted, would be great if you could take another look during my night/your day at the UI. If what I have implemented is ok tomorrow I'll work on tests, more tests...

lognaturel:
I get an NPE when trying to tap a minimal select field with your latest commit. I think context may be accessed prematurely but don't have time for a deep dive right now. The previous commit seems to work as expected.

Although the minimal select looks good, it's not consistent with the rest of the question types we offer. I think this might be some of what @seadowg was getting at too. Does it match some material design guideline I haven't noticed? When I look at [this page](https://www.material.io/components/menus#behavior) I see that selects look like text fields except for the downward triangle.

I see a leak of `SelectMultiMinimalWidget.adapter` (maybe that was fixed by your latest commit).

The background color of the full screen dialog doesn't match the `FormEntryActivity`.

I notice many of the select one appearances don't immediately select an option on tap. I think that should be consistent. Is there a reason it isn't?

It looks like we talked past each other for the select behavior! I was thinking we would treat it like settings full screen dialogs. Look at settings for any of the Google apps and you'll see that they're committed immediately and either the arrow or the hardware back button just returns to wherever the user was before. This is what makes sense to me and I find it really confusing that the hardware back button doesn't commit. If you're not comfortable with that, then I think we should follow the guidelines [here](https://material.io/components/dialogs#full-screen-dialog). That is, we'd have an X at the upper left and SAVE at the upper right. The hardware back button would close without saving, the same behavior as the X. That would make it clear to me that there's only one path to save and everything else cancels. For this to work you'd need to clearly combine the search box clearing icon (currently x) with the search bar. I have a slight preference for the simper model with immediate commit but it's not very strong.

grzesiek2010:
> I get an NPE when trying to tap a minimal select field with your latest commit. I think context may be accessed prematurely but don't have time for a deep dive right now. The previous commit seems to work as expected.

hmmm can't reproduce now... I tried three devices.

> Although the minimal select looks good, it's not consistent with the rest of the question types we offer.

I thought this is what we agreed on... I don't know, this can't be even super consistent because in all the other cases we display choices in a widget and in all the other widgets that display dialogs we use buttons. So how do you even imagine the consistency?

> I notice many of the select one appearances don't immediately select an option on tap. I think that should be consistent. Is there a reason it isn't?

what do you mean? Is there any change in behavior that I implemented? 

> I see a leak of SelectMultiMinimalWidget.adapter (maybe that was fixed by your latest commit).

In canary leak or is it maybe a crash?

> The background color of the full screen dialog doesn't match the FormEntryActivity.

is that bad? all our dialog (at least full screen ones) use that color for example the one that asks for identity which also is a part of filling a form.

> and either the arrow or the hardware back button just returns to wherever the user was before. This is what makes sense to me and I find it really confusing that the hardware back button doesn't commit.

ok sorry I misunderstood our previous conversation. i added the code back and now answers are saved even even if you click device back button.

lognaturel:
Still get the crash in Android 5.1. I thought it was because `Fragment` lifecycle methods changed slightly at some point and there's no `onAttach(context)` in Android 5.1 but I get it on my Android 9 emulator too.

```
    java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.content.Context.getResources()' on a null object reference
        at android.view.ViewConfiguration.get(ViewConfiguration.java:384)
        at android.view.View.<init>(View.java:3834)
        at org.odk.collect.android.adapters.AbstractSelectListAdapter.setUpNoButtonsView(AbstractSelectListAdapter.java:156)
        at org.odk.collect.android.adapters.AbstractSelectListAdapter$ViewHolder.bind(AbstractSelectListAdapter.java:257)
        at org.odk.collect.android.adapters.SelectOneListAdapter$ViewHolder.bind(SelectOneListAdapter.java:110)
        at org.odk.collect.android.adapters.AbstractSelectListAdapter.onBindViewHolder(AbstractSelectListAdapter.java:104)
        at org.odk.collect.android.adapters.AbstractSelectListAdapter.onBindViewHolder(AbstractSelectListAdapter.java:67)
        at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7065)
        at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7107)
        at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6012)
        at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6279)
        at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6118)
        at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6114)
        at com.google.android.flexbox.FlexboxLayoutManager.getFlexItemAt(FlexboxLayoutManager.java:456)
        at com.google.android.flexbox.FlexboxLayoutManager.getReorderedFlexItemAt(FlexboxLayoutManager.java:474)
```

> I thought this is what we agreed on... I don't know, this can't be even super consistent

I should have reacted to that thread previously. I agree that it's not going to be exactly the same as any other question type. Maybe the thing that bothers me most is the contrasting background because it really makes the field stand out in a field list whereas I think it should look more like question types based on edit texts. I find that the padding is also quite tight around the text when compared to Material Design examples. But all that said, I don't think this is too important for now. Like I said, I agree it looks good, and I think that if we make cosmetic changes later on it won't be confusing as long as the behavior is the same. For now, how about you just try it in a field list with a string question and see if there's anything you can do to make things look more balanced. Then we can come back to it later.

> what do you mean? Is there any change in behavior that I implemented?

I am not able to get it running right now but if I remember correctly, select one with `columns` does what I expect: when I tap a choice, the full screen dialog closes and my choice is committed. However, I think it's all the select one `columns-pack no-button` appearances that don't do what I expect: I select a choice and nothing happens. I then have to tap to commit the choice and exit the dialog.

> In canary leak or is it maybe a crash?

Reported by Leak Canary.

> is that bad? all our dialog (at least full screen ones) use that color for example the one that asks for identity which also is a part of filling a form.

Ok, that's reasonable, then!

> i added the code back and now answers are saved even even if you click device back button.

Cool. And does it feel ok to you? Again, I'm trying to be consistent with other apps I've used but if you have examples that point to alternatives, I'm certainly open to those.

grzesiek2010:
> Still get the crash in Android 5.1. I thought it was because Fragment lifecycle methods changed slightly at some point and there's no onAttach(context) in Android 5.1 but I get it on my Android 9 emulator too.

ah ok I got it, fixed.

> I am not able to get it running right now but if I remember correctly, select one with columns does what I expect: when I tap a choice, the full screen dialog closes and my choice is committed. However, I think it's all the select one columns-pack no-button appearances that don't do what I expect: I select a choice and nothing happens. I then have to tap to commit the choice and exit the dialog.

ok I see, fixed.

> Cool. And does it feel ok to you? Again, I'm trying to be consistent with other apps I've used but if you have examples that point to alternatives, I'm certainly open to those.

that seems fine to me and it was the way I implemented it initially so I unnecessarily thought to much about it.

grzesiek2010:
> I should have reacted to that thread previously. I agree that it's not going to be exactly the same as any other question type. Maybe the thing that bothers me most is the contrasting background because it really makes the field stand out in a field list whereas I think it should look more like question types based on edit texts.

How about this then?

image

Do you like it more?

lognaturel:
> Do you like it more?

In isolation I'd probably go with what you had before but in combination with the rest of the question types we have and particularly ones with `EditText`s, this looks a lot more consistent. So yes, in context, I do prefer this. Thanks!!

I also really like that it's consistent with the current `minimal` appearance so it won't surprise anyone now.

grzesiek2010:
I feel like the functionality I prepared is ready for review. **I'm still trying to add more tests** but it shouldn't block you guys from reviewing changes I implemented and decisions I made.
Since it's a pretty complex change more difficult to understand than it seems. I decided to prepare this post for both reviewers and testers to clarity what I've done and what should be tested.

### For all
We already have a lot of options/appearances we can use to customize selects like:
- buttons/no-buttons
- different number of columns: columns-n/columns-pack/columns
- media files image/audio/video
- filtering using autocomplete
- quick in select_one questions

we also have **minimal** appearance but it was built in a completly different way so we couldn't combine it with other appearances. Now it changed and thanks to using the same component (RecyclerView) for both normal and minimal select widgets we can mix all those appearances so we can for example use: **minimal no-buttons columns-3 autocomplete quick** :smiley: 

### For reviewers
I achieved what I described above using [RecyclerView](https://developer.android.com/guide/topics/ui/layout/recyclerview) (my custom [ChoicesRecyclerView](https://github.com/getodk/collect/pull/3964/files#diff-45a1d18f1b79ef88a03daff5352633bb)) so now we use the same RecyclerView in all selects excluding (LikertWidget/ListWidget). The only difference is that for minimal widgets it is displayed in a separate full-screen dialog.
Working on it I refactored a lot of code related to selects. I decided to change the hierarchy of classes a bit:

**Before:**
image

**After:**
image

_Both diagrams don't contain select minimal widgets_

I did that because it simplified the code, I was able to get rid of duplications and now the hierarchy makes more sense.
Reworking the hierarchy I also move the code responsible for creating widgets to layout files so @SaumiaSinghal I also invite you as a reviewer and just keep in mind that you don't need to touch those widgets (excluding LikertWidget/ListWidget because as i said above those are built in a different way without using RecyclerView).

### For testers
Seems like this is a change related to select minimal widgets only but that's not the case. To achieve what I've done I had to implement changes in almost all select widgets (again excluding LikertWidget/ListWidget - list widget is the one we create using `list` or `list-nolabel` appearances). So basically we need to test almost every select widget.

**Initially I was scared of the amount of possible cases we would need to test taking into account that now we can mix minimal appearance with all the others and we have a lot of options to customize the way the list of choices looks like but fortunately the same component is used in all cases so I'm pretty confident that if the way the list works itself is correct in case of normal selects it will work properly in minimal selects (in the dialog) as well (and the other way around). By `the way the list works itself` I mean for example filtering/selecting options/number of columns the code responsible for that is shared so if it works in one place it will work well everywhere.**

collect_app/build.gradle
```diff
@@ -199,7 +199,7 @@ dependencies {
 
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.browser:browser:1.2.0'
-    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.appcompat:appcompat:1.3.0-alpha01'
```
lognaturel: From the commit message: "to fix inflating webView for selectImageMap widgets". Is this fixed in 1.2.0-rc02? In general, I think we should really avoid things off the alpha channel. Ideally we wouldn't have to include any unreleased versions but I understand that's not always possible. Using a release candidate seems less dangerous than a first alpha.

collect_app/src/test/java/org/odk/collect/android/utilities/StringUtilsTest.java
```diff
@@ -1,29 +1,22 @@
...
-     * This is a silly test but it is here to guarantee this behaviour,
-     * since without it the method causes a crash when processing text for
-     * questions with no plain text label. See getodk/getodk#1247.
-     */
     @Test
-    public void textToHtml_BouncesNullInput() {
-        String input = null;
-        CharSequence observed = StringUtils.textToHtml(input);
-        Assert.assertNull(observed);
+    public void textToHtml_BouncesEmptyString() {
```
lognaturel: "bounces" implies the input becomes the output. Maybe rename to `textToHtml_nullBecomesEmptyString`?

lognaturel:
Thanks for that helpful diagram, @grzesiek2010. The high-level changes all seem very positive. I've only started a detailed review with some of the easier parts and only have tiny comments there.

I'm happy with the UX. I'd still prefer a standard back arrow coming out of the minimal dialog instead of the check, though. It's not a super strong preference and I'd be interested to hear what @seadowg thinks.

seadowg:
> I'm happy with the UX. I'd still prefer a standard back arrow coming out of the minimal dialog instead of the check, though. It's not a super strong preference and I'd be interested to hear what @seadowg thinks.

Yeah I agree I'd suggest a back arrow or a cross instead. If there is a worry about it feeling like "cancelling" instead of "selecting" then we could go with a top right "select"/"save" and make "back" a cancel but personally I don't think we need that.

collect_app/src/test/java/org/odk/collect/android/utilities/StringUtilsTest.java
grzesiek2010: Right, fixed.

grzesiek2010:
> I'd still prefer a standard back arrow coming out of the minimal dialog instead of the check

> Yeah I agree I'd suggest a back arrow or a cross instead.

Ok it's added.

collect_app/build.gradle
grzesiek2010: Right, 1.2.0-rc02 also works well so I changed it. Thanks!

collect_app/src/main/java/org/odk/collect/android/adapters/SelectOneListAdapter.java
```diff
@@ -16,86 +16,78 @@
...
             }
             selectedRadioButton = (RadioButton) buttonView;
-            selectedValue = items.get((int) selectedRadioButton.getTag()).getValue();
+            selectedValue = props.getItems().get((int) selectedRadioButton.getTag()).getValue();
         }
     }
 
-    class ViewHolder extends AbstractSelectListAdapter.ViewHolder {
-        ImageView autoAdvanceIcon;
+    public void setSelectItemClickListener(SelectOneItemClickListener listener) {
```
lognaturel: Should always be set from the constructor?

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
```diff
@@ -0,0 +1,5 @@
+package org.odk.collect.android.listeners;
+
+public interface SelectOneItemClickListener extends SelectItemClickListener {
+    void onClearNextLevelsOfCascadingSelect();
```
lognaturel: Why is this a separate interface and method from `SelectItemClickListener`? Can't the `onItemClicked` implementation for `SelectOneWidget` include the call? Only one class that implements this interface has a definition for the method which doesn't feel great.

Note to self: I also reviewed the implementation and I think we should consider revising it. It currently blindly clears any "external itemset" questions after a select one, whether or not they're actually linked or not.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/SelectMultiMinimalDialog.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    protected AbstractSelectListAdapter getAdapter() {
+        return adapter;
+    }
+
+    @Override
+    public void onItemClicked() {
+        // do nothing
```
lognaturel: Why have it implement the interface, then? Can you pass in a null listener when creating the adapter (and make sure that the adapter can handle a null listener)?

collect_app/src/main/java/org/odk/collect/android/widgets/SelectMultiMinimalWidget.java
```diff
@@ -0,0 +1,83 @@
...
+        super(context, prompt);
+        selectedItems = getFormEntryPrompt().getAnswerValue() == null
+                ? new ArrayList<>() :
+                (List<Selection>) getFormEntryPrompt().getAnswerValue().getValue();
+        updateAnswerLabel();
+    }
+
+    @Override
+    protected void showDialog() {
+        SelectMultiMinimalDialog dialog = new SelectMultiMinimalDialog(getSavedSelectedItems(), WidgetAppearanceUtils.isFlexAppearance(getFormEntryPrompt()), WidgetAppearanceUtils.isAutocomplete(getFormEntryPrompt()), getChoicesRecyclerViewProps());
```
lognaturel: How did you decide what goes in the props record vs. not? Why not include flex and autocomplete appearances?

collect_app/src/main/java/org/odk/collect/android/widgets/ItemsetWidget.java
```diff
@@ -75,15 +79,29 @@ public ItemsetWidget(Context context,
         this.fileUtil = fileUtil;
 
         items = getItems();
+    }
+
+    @Override
+    public IAnswerData getAnswer() {
+        Selection selectedItem =  ((SelectOneListAdapter) recyclerViewAdapter).getSelectedItem();
```
lognaturel: Could you use `getSelectedItems().get(0)` instead to avoid having to cast? Generally having to cast is dangerous because it can lead to runtime errors. This same pattern shows up in a couple other places.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/SelectMultiMinimalDialog.java
```diff
@@ -0,0 +1,42 @@
...
+    public SelectMultiMinimalDialog() {
+    }
+
+    public SelectMultiMinimalDialog(List<Selection> selectedItems, boolean isFlex, boolean isAutoComplete, ChoicesRecyclerViewAdapterProps props) {
+        super(isFlex, isAutoComplete);
+        adapter = new SelectMultipleListAdapter(selectedItems, this, props);
+    }
+
+    @Override
+    public void onViewCreated(@NotNull View view, Bundle savedInstanceState) {
```
lognaturel: I don't think this is needed.

lognaturel (changes requested):
I'm still making my way through this but wanted to give you some preliminary feedback.

I think this work is a good opportunity to move towards organizing packages by functional unit rather than by types of files. I'm thinking something like `formentry/selects`. @seadowg has been leading the charge to rework the package structure so would be great to get his thoughts.

`ChoicesRecyclerViewAdapterProps` feels kind of like random stuff that dialog needs. Did you consider any alternative designs? I haven't had a chance to give it much thought yet but wanted to flag it for your consideration. It seems there's some overlap with what it's responsible for and `SelectMinimalViewModel`.

Nit: there are quite a few uses of `protected`. Some are because of overrides that need to maintain the access level. But others (most?) should just be package-private (no modifier).

collect_app/src/main/java/org/odk/collect/android/adapters/SelectOneListAdapter.java
lognaturel: I see, this is for rotation. Note to self: think through alternatives for this a bit.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/SelectMultiMinimalDialog.java
grzesiek2010: Why? Have you seen the comment below?

collect_app/src/main/java/org/odk/collect/android/widgets/SelectMultiMinimalWidget.java
grzesiek2010: Because flex and autocomplete appearances are used in `SelectMinimalDialog` but props are passed to an adapter
https://github.com/getodk/collect/pull/3964/files#diff-c84dd1657c1e1200edf5f3ca7b8f062aR23 that was my reasoning.

collect_app/src/main/java/org/odk/collect/android/widgets/ItemsetWidget.java
grzesiek2010: Maybe not directly getSelectedItems().get(0) because it might be empty but yeah something like:
```
recyclerViewAdapter.getSelectedItems().isEmpty() ? null : recyclerViewAdapter.getSelectedItems().get(0);
```
will be possible.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/SelectMultiMinimalDialog.java
grzesiek2010: Ahh I see because we don't use it either way right right...

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/SelectMultiMinimalDialog.java
grzesiek2010: Right.

collect_app/src/main/java/org/odk/collect/android/widgets/SelectMultiMinimalWidget.java
lognaturel: I see, that makes sense. You know, the more I think about this the more it feels like a pattern we don't want to introduce. It's not very Java/Android idiomatic and I don't think there's much advantage to it. It's ok to have long parameter lists! You could also consider something like a `SelectQuestionPrompt` object that wraps the form entry prompt and booleans for the different appearances. When that object is created with a `FormEntryPrompt`, the appearance is read and sanitized once and all the booleans are set. That also has the advantage of moving away from the static `WidgetAppearanceUtils`.

seadowg:
@lognaturel yeah I like the idea of the widgets starting to go in related packages. Maybe there should be a `widgets.select` package?

grzesiek2010:
> yeah I like the idea of the widgets starting to go in related packages. Maybe there should be a widgets.select package?

I also thought about it but had a lot of other staff to refactor so somehow decided not to add more. But generally I'm not against.

collect_app/src/main/java/org/odk/collect/android/widgets/ItemsetWidget.java
grzesiek2010: Fixed in https://github.com/getodk/collect/pull/3964/commits/e757ea8213db99fe18d88133e6bda0687b232878

lognaturel:
I think the biggest advantage for the package is being able to run a coherent test suite for a single area.

> Maybe there should be a widgets.select package?

So keeping `widgets` at the top level for now and grouping adapters, dialogs, etc related to selects in that `select` package? That sounds reasonable to me.

I think one of the big challenges in this work has been figuring out how the dialog, activity, adapter, etc, all communicate. I mentioned to @seadowg that I was finding it really hard to reason about, especially what the responsibility split and communication should be between the adapters and the viewmodels. He mentioned that he's solved a similar problem before so @seadowg, would be great to get a few thoughts. Some concrete things to look at:
- the dialog is a listener for item clicks from the adapter https://github.com/getodk/collect/pull/3964#discussion_r466137789 -- could this communication happen through the view model?
- the selected items are set on the view through the activity using a dialog listener https://github.com/getodk/collect/pull/3964/files#diff-03ff45091550285dd88b56927010c457R2669. There isn't binary data involved so it doesn't feel quite right.

seadowg:
@lognaturel I'll try and lay this out in a generic example. My immediate though is to handle this by having the Adapter and the Widget (view) share a ViewModel that is attached the Activity lifecycle. The ViewModel could own and expose both the list of selectable items and the currently selected items. In this world the Adapter would take the ViewModel in as a dependency in its constructor and then bind to the items and selected items:

```java
class MyAdapter extends BaseAdapter {

    private List<Item> items;
    private List<Item> selectedItems;

    public MyAdapter(LifecycleOwner lifecycleOwner, MyViewModel viewModel) {
        viewModel.getItems().observer(lifecycleOwner, items -> {
            this.items = items;
            notifyDatasetChanged();
        });

        viewModel.getSelectedItems().observer(lifecycleOwner, items -> {
            this.selectedItems = items;
            notifyDatasetChanged();
        });
    }
}
```

When an item is selected the Adapter could also then call a method like `selectItem` (in `onItemClick`) on the ViewModel to add that item to the ViewModel's `selectedItems`.

This means the widget can then observe or just use the selected items on the ViewModel without the Activity needing to know anything about what's going on.

Let me know if any of that could use clarifying!

**EDIT**: Something I meant to point out is that I think Adapters and the "new" arch components feel a bit awkward together. The adapter solves similar issues (reactive data) and so it can feel a little clunky setting a field in the Adapter and then sort of forwarding the "reactivity" to its machinery. I think this awkwardness comes from the fact that as much as it doesn't feel like it the Adapter is really an optimization detail of the view/UI code - we only have it so we can reduce memory use ansd increase refresh rates. I think it might actually be easier to think about how to design these kinds of UIs if you imagine it with just a `LinearLayout` (that you're swapping the whole list of items in and out of) and a ViewModel.

lognaturel:
> a ViewModel that is attached the Activity lifecycle

Ah, I think this might be where things break down. Won't it be a problem if there are multiple minimal select widgets on the same screen? Or even in the same form, actually?

@grzesiek2010 it would help to get a quick overview of how you ended up with the current communication strategies so we don't have to revisit too many ideas you've already considered. This is tricky.

**EDIT**: maybe there's a not terrible way forward by having the view model keep track of `itemsForCurrentQuestion` or something like that? It'd be `null` if the question isn't a select question?

seadowg:
> Won't it be a problem if there are multiple minimal select widgets on the same screen? Or even in the same form, actually?

Ah maybe. It might be that the view model should account for that and store multiple items/selected items for different form indices? Actually writing this I *think* you can handle this using the `ViewModelProvider`'s key feature: https://developer.android.com/reference/androidx/lifecycle/ViewModelProvider#get(java.lang.String,%20java.lang.Class%3CT%3E)

lognaturel:
> the view model should account for that and store multiple items/selected items for different form indices

The lists can be really big so I don't think we should do that. How about keeping track of just the current form index's items?

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/SelectMultiMinimalWidgetTest.java
```diff
@@ -0,0 +1,224 @@
...
+import org.junit.rules.RuleChain;
+import org.odk.collect.android.activities.FormEntryActivity;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.FormLoadingUtils;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.pages.FormEntryPage;
+
+import static java.util.Arrays.asList;
+
+public class SelectMultiMinimalWidgetTest {
```
lognaturel: These all look to me like they could be robolectric tests entirely within the context of the dialog. Did you try to go that direction and run into issues?

For the last test which verifies the back button and return arrow behavior, I think you can also stay in the dialog and make sure that some method is called to commit the change. How exactly you do that will depend on some of the architecture changes we're discussing. Then there also needs to be a test on the `FormEntryActivity` that verifies that if an answer is set by the mechanism used in the prior test, it's displayed as expected. Depending on what mechanism that ends up being, it may or may not already be addressed.

For the missing audio toast, there's already some coverage in the tests for `AudioPlayerViewModel`. However, there is no test that currently ensures the error is shown as a toast. I think you should be able to do that within the context of the dialog but it looks like setting up the dependencies might be impractical.

collect_app/src/test/java/org/odk/collect/android/widgets/SelectMultiWidgetTest.java
```diff
@@ -71,6 +82,194 @@ public void setup() throws Exception {
         when(audioHelper.setAudio(any(AudioButton.class), any())).thenReturn(new MutableLiveData<>());
     }
 
+    @Test
+    public void byDefault_shouldGridLayoutManagerBeUsed() {
+        populateRecyclerView(getWidget());
```
lognaturel: I don't think there's any reason for populating the recycler view for tests that look at the layout manager.

grzesiek2010:
> @grzesiek2010 it would help to get a quick overview of how you ended up with the current communication strategies so we don't have to revisit too many ideas you've already considered. This is tricky.

when it comes to communication dialog - widget:

What I did is not perfect I know but it's consistent with how other widgets work RankWidget, DateWidget use the same approach with setBinaryData so it's not like it is only for binary widget. I already asked @SaumiaSinghal to rename those methods removing binary words as a part of https://github.com/getodk/collect/pull/3975 because there the same approach is used.

In the future I would like to turn every single widget to a fragment so that we can have a separate viewmodel for every widget. Now it is not possible because we would need to create those widgets using activity context and we would end up with shared viewmodels what would be a problem in case of field-lists.
I said in the future because now @SaumiaSinghal in reworking widgets so it's not a good time to implement such changes.

so that's basically why for now I decided to implement it with that old not perfect but consistent way.

collect_app/src/main/res/layout/select_minimal_widget_answer.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingHorizontal="@dimen/margin_standard"
+    android:paddingVertical="@dimen/margin_small">
+
+    <com.google.android.material.textfield.TextInputEditText
+        android:id="@+id/choices_search_box"
```
lognaturel: Nit: I got really confused about this name because I thought it was related to the `search`/`autocomplete` appearance. This represents the answer, right? Or the selection?

lognaturel:
> In the future I would like to turn every single widget to a fragment so that we can have a separate viewmodel for every widget. [...] now @SaumiaSinghal in reworking widgets so it's not a good time to implement such changes

Thanks for that reminder, we had talked about this concept but I didn't connect the threads. Seems ok to keep this structure for now.

Regarding tests, I'd like to learn a bit about how you think about the Espresso tests and why they're needed. To me it looks like they duplicate coverage that you got at lower levels but I may be missing something! Is there something specific about this context that makes you feel like you need tests on a real device?

Here are examples of duplicate coverage that I see:
- What would it mean for the various `assertAllItemsAreDisplayedInVariousAppearances` tests to fail and the tests in `AudioVideoImageTextLabelVisibilityTest` to pass? I think the label tests are sufficient for ensuring media is properly accounted for.
- Could the various `bothClickingRadioButtonAndImage_shouldSelectItem` Espresso tests be replaced by a single test in `AudioVideoImageTextLabelTest`? With the current structure I think you'd set a mocked listener and make sure the `onItemClicked` method is called but that structure might change depending on what you think of some of the architecture ideas discussed.
- There's a lot of duplication between `SelectOneWidgetTest` and `SelectMultipleWidgetTest` because they have a common base class. Could you introduce `SelectWidgetTest`, have `createWidget` return a `BaseSelectListWidget` and use one of the concrete implementations to verify the behavior that's shared? It's not perfect because since you won't directly test specific methods I don't think you can ensure definitively that what you're testing is indeed shared. Maybe there's a better approach I'm not thinking of. These tests are fast so it doesn't feel like a huge priority either way.

On another note, `ChoicesRecyclerViewTest` and the `no-buttons` and `columns-pack` tests feel really tightly coupled to the implementation details. I don't know how realistic that is but it feels like it'd be nice to verify slightly more generic things like "the view hierarchy contains no radiobuttons or checkboxes" or "the items are displayed next to each other".

collect_app/src/main/res/layout/select_minimal_widget_answer.xml
grzesiek2010: This is that `spinner` button which contains answers inside. I wasn't sure about its name, would spinner_button be better or something else?

collect_app/src/test/java/org/odk/collect/android/widgets/SelectMultiWidgetTest.java
grzesiek2010: Right.

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
grzesiek2010: I think this is broken either way even on master and w didn't have any test for that clearing. Can I file a separate issue for that to fix it and handle in a better way and here just remove?

grzesiek2010:
> yeah I like the idea of the widgets starting to go in related packages. Maybe there should be a widgets.select package?

Do you think all those widgets should be in one package?
image

if so it should be called items not selects because then RankingWIdgets doesn't fit.

collect_app/src/main/java/org/odk/collect/android/widgets/SelectMultiMinimalWidget.java
grzesiek2010: I agree it's not good I didn't like the long list of parameters and wanted to improve it somehow but ended up with another approach that is not perfect. I'll take another look maybe we can improve it or just pass all parameters directly as you said it's not that bad.

lognaturel:
> it should be called items not selects because then RankingWIdgets doesn't fit.

I'm happy with `items`.

collect_app/src/main/res/layout/select_minimal_widget_answer.xml
lognaturel: I would just call it something like `answer` but `spinner_button` feels like a big improvement, too.

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
lognaturel: Unfortunately, I don't think we should just remove it because it does address a real issue and works well in the common case. Is it not possible to move the behavior just into the `onItemClicked` implementation for `ItemsetWidget`?

Definitely good to file an issue to review it and add tests, though. 👍

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
```diff
@@ -227,11 +227,22 @@ public void addMediaFromChoice(AudioVideoImageTextLabel audioVideoImageTextLabel
...
+                    audioVideoImageTextLabel.setBigImage(new File(props.getReferenceManager().deriveReference(bigImageURI).getLocalURI()));
+                }
+                if (videoURI != null) {
+                    audioVideoImageTextLabel.setVideo(new File(props.getReferenceManager().deriveReference(videoURI).getLocalURI()));
+                }
+                if (audioURI != null) {
+                    audioVideoImageTextLabel.setAudio(audioURI, props.getAudioHelper());
+                }
+            } catch (InvalidReferenceException e) {
+                Timber.e(e, "Invalid image reference due to %s ", e.getMessage());
```
lognaturel: While here, change to `Timber.d` since it's not developer-actionable. Message should also probably be "Invalid media ref"

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
```diff
@@ -227,11 +227,22 @@ public void addMediaFromChoice(AudioVideoImageTextLabel audioVideoImageTextLabel
             String imageURI = getImageURI(index, items);
             String videoURI = props.getPrompt().getSpecialFormSelectChoiceText(item, "video");
             String bigImageURI = props.getPrompt().getSpecialFormSelectChoiceText(item, "big-image");
-            audioVideoImageTextLabel.setImageVideo(imageURI, videoURI, bigImageURI, props.getReferenceManager());
-
             String audioURI = getPlayableAudioURI(props.getPrompt(), item, props.getReferenceManager());
-            if (audioURI != null) {
-                audioVideoImageTextLabel.setAudio(audioURI, props.getAudioHelper());
+            try {
+                if (imageURI != null) {
```
lognaturel, line 281: My preference would be for all the setters to handle possible nulls so these null checks don't have to be here. Alternately the setters could have `@NonNull` parameter annotations and these null checks could stay but come out of the setters.

lognaturel (commented):
These last commits are poetry! It's really great to see the tests get more targeted and so much faster to run. 🎉🎉🎉 Having them all in the same package is also really nice for using them as documentation.

The move away from the `props` object is going to shift tests around so might be good to do sooner than later.

collect_app/src/main/res/layout/select_minimal_widget_answer.xml
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
grzesiek2010: I need to have those checks here not in setters because I want to pass files to those setters not URIs so that I can mock such files for tests (if a file exist or not).

grzesiek2010:
> yeah I like the idea of the widgets starting to go in related packages. Maybe there should be a widgets.select package?

Done, but the package is called items as we agreed.

> The move away from the props object is going to shift tests around so might be good to do sooner than later.

This is already done.

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
grzesiek2010: Ok I made a mistake (clearing works as before) I used the form from the doc https://docs.google.com/spreadsheets/d/1CCjRRHCyJXaSEBHPjMWrGotnORR4BI49PoON6qK01BE/edit#gid=1 which seems broken because it doesn't display the last level. It's called new cascading select so it indicates that something has changed do you know anything about it? I know that the old version  has two columns (for filters) and I was able to transform it and now it works but I'm curious what's going on.
Regarding the listener you are right I don't need that second listener. I already removed it.

collect_app/src/main/java/org/odk/collect/android/adapters/SelectOneListAdapter.java
grzesiek2010: Can't see any simpler solution.

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
lognaturel: That form is just crazy. I have no idea why it was like that and I just merged https://github.com/XLSForm/xlsform.github.io/pull/196 to improve it.

I don't think `onClearNextLevelsOfCascadingSelect` would be relevant to that form, though. It's a special case only for "external itemsets" that cascade (http://xlsform.org/en/#external-selects).

grzesiek2010: Yea I agree with internal choices it works without that fix but now I can confirm that it also works with external choices. At least as it worked before.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneWidget.java
```diff
@@ -100,15 +99,21 @@ public void onItemClicked() {
...
+        super.clearAnswer();
+    }
+
     /**
      * It's needed only for external choices. Everything works well and
      * out of the box when we use internal choices instead
      */
-    @Override
-    public void onClearNextLevelsOfCascadingSelect() {
+    private void onClearNextLevelsOfCascadingSelect() {
```
lognaturel: You can go further and move this to `ItemsetWidget`. Then you can name it something like `clearNextLevelsOfCascade` and remove the comment which is a little misleading because there are three things that are "external choices."

collect_app/src/main/java/org/odk/collect/android/listeners/SelectOneItemClickListener.java
lognaturel: Great. Just be careful when saying "external choices" because that can mean three things: `search()`, "external itemsets" (`ItemsetWidget`) or external secondary instances.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneWidget.java
grzesiek2010: I'm afraid it must be in this widget because the first level of cascading select is always `SelectOneWidget` and then next levels are represented by `ItemsetWidgets`. So if I change it like that it won't work.

lognaturel: I see. I guess that's what the provided examples do show though it's certainly not a requirement. This is all kind of crazy so let's at least improve the comment and naming. How about `clearFollowingItemsetWidgets` and something like "If there are "fast external itemset" selects right after this select, assume that they are linked to the current question and clear them."

grzesiek2010: Sounds good, will do.

grzesiek2010:
@lognaturel 
I think this is ready for your review.

collect_app/src/androidTest/java/org/odk/collect/android/support/actions/RecyclerViewAction.java
```diff
@@ -0,0 +1,33 @@
+package org.odk.collect.android.support.actions;
+
+import android.view.View;
+
+import androidx.test.espresso.UiController;
+import androidx.test.espresso.ViewAction;
+
+import org.hamcrest.Matcher;
+
+public class RecyclerViewAction {
```
lognaturel: I think this may no longer be needed? It's not currently used so I think it makes sense to remove it for now and re-introduced if it's useful in the future.

collect_app/src/androidTest/java/org/odk/collect/android/support/assertions/RecyclerViewAssertions.java
```diff
@@ -0,0 +1,24 @@
...
+import android.widget.FrameLayout;
+
+import androidx.test.espresso.ViewAssertion;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+
+public class RecyclerViewAssertions {
```
lognaturel: No longer needed.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormEntryPage.java
```diff
@@ -291,4 +297,99 @@ private void flingLeft() {
...
+        onView(allOf(instanceOf(AppCompatImageButton.class), withParent(withId(R.id.toolbar)))).perform(click());
+        return this;
+    }
+
+    public FormEntryPage assertSelectMinimalDialogAnswer(String answer) {
+        onView(withId(R.id.answer)).check(matches(withText(answer)));
+        return this;
+    }
+
+    public FormEntryPage assertSearchBoxIsHidden(boolean minimalAppearance) {
```
lognaturel: A few methods starting here are no longer used.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2662,6 +2665,16 @@ public void onRankingChanged(List<SelectChoice> items) {
         }
     }
 
+    @Override
+    public void updateSelectedItems(List<Selection> items) {
```
lognaturel, line 2673: How about adding a TODO with some of the discussion we had about this? Just a quick note to help the next person puzzling over this code realize that it's not in an ideal state. Something like "replace `SelectMinimalDialogListener` and use of the binary data flow, likely with a ViewModel"

collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
```diff
@@ -136,22 +141,22 @@ public void setAudio(String audioURI, AudioHelper audioHelper) {
...
-    public void setImageVideo(String imageURI, String videoURI,
-                              String bigImageURI, ReferenceManager referenceManager) {
-        this.bigImageURI = bigImageURI;
-        this.videoURI = videoURI;
-        this.referenceManager = referenceManager;
-
-        if (videoURI != null) {
-            setupVideoButton();
+    public void setImage(File imageFile) {
+        if (imageFile != null) {
```
lognaturel: If you're going to have the callers check for `null`, can you mark the params as `@NonNull` and not have additional null checks here?

lognaturel (approved):
Thanks, @grzesiek2010! I'm happy with where this puts us and I'm particularly happy with how the tests are looking on the whole. I think it's important to move to QA to do a further sanity check on how complete the coverage is and any other strange edge cases that might have come up. Couple of small things to clean up inline.

collect_app/src/androidTest/java/org/odk/collect/android/support/actions/RecyclerViewAction.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/assertions/RecyclerViewAssertions.java
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormEntryPage.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Fixed.

kkrawczyk123:
**Tested with success!**
Verified on Android 5.1, 7.1 and 10.

Tested scenarios:
- light/dark mode
- moving forward/ backward
- jump from hierarchy
- minimize app
- screen rotation
- remove response
- read only questions
- RTL language
- different types of select one widget
- different types of select multi widget
- select one/multi with quick appearance
- select one/multi with autocomplete
- selects with icons
- external itemsets
- cascasing external
- fieldlist
- select mininal
- search and select
- gridwidget
- gridwidget with autoplay: issue visible on master too and reported in #4043
- select with autoplay
- selects with columns-pack appearance
- don't keep activities
- regression check on all widgets

great work @grzesiek2010!

mmarciniak90:
I also confirm successful tests on Android versions: 7.0, 8.1, 9.0
@grzesiek2010 :+1: no bugs in such a complex PR!!

grzesiek2010:
> no bugs in such a complex PR!!

wow I can finish my career satisfied then :smile:

PR #3967, opened 2020-07-10
Update All Widgets Form Test

### LINKED ISSUE #3963: Update AllWidgetsFormTest
Currently, `AllWidgetsFormTest` checks the behaviour of almost all widgets. There is a lot of scope to simplify it. As the behavior of all the widgets ranging from displaying intents to changing answers can be verified in corresponding unit tests, `AllWidgetsFormTest` can be just used to verify whether all the widgets are rendered on screen or not.

Performing a simple swipe to check whether the next widget appears on the screen would be enough rather than navigating to the widgets using the Form Heirarchy menu.

SaumiaSinghal:
Closes #3963 

I verified the new `AllWidgetsFormTest` on my emulator.

I replaced the `all-widgets.xml` with the latest version of the form and removed all the extra checks in `AllWidgetsFormTest`.

No

No

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/AllWidgetsFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/AllWidgetsFormTest.java
@@ -45,7 +45,7 @@ public class AllWidgetsFormTest {
     @ClassRule
     public static final LocaleTestRule LOCALE_TEST_RULE = new LocaleTestRule();
 
-    public AllWidgetsFormTestRule activityTestRule = new AllWidgetsFormTestRule();
+    public FormActivityTestRule activityTestRule = new FormActivityTestRule(ALL_WIDGETS_FORM);
 
     @Rule
     public RuleChain copyFormChain = RuleChain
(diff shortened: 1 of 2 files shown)


seadowg:
@SaumiaSinghal did you not tick the last checkbox because it doesn't really apply? Even if that's the case I think it's best to go ahead and check it. Then we know the PR is ready for review!

SaumiaSinghal:
Oh sorry @seadowg! I forgot about that. Yes it is ready for review.

collect_app/src/androidTest/java/org/odk/collect/android/AllWidgetsFormTest.java
```diff
@@ -846,189 +555,72 @@ public void testImageSelectOne() {
...
-
-    private String randomIntegerSeparator() {
-        int number = 123456;
-        NumberFormat numberFormat = NumberFormat.getNumberInstance(Locale.US);
-        return numberFormat.format(number);
-     }
-
-    //endregion
-
     private static class AllWidgetsFormTestRule extends FormActivityTestRule {
```
seadowg: We probably don't need this class any more. The test should just use the `FormActivityTestRule` directly.

seadowg (changes requested):

seadowg (approved):

PR #3969, opened 2020-07-14
Reworking Range Widget

SaumiaSinghal:
This PR includes rewriting the test coverage and class of Range Picker Widget 

I verified the behavior on the `master` branch and on this branch for the attached form. I didn't find any behavioral changes in the widget. I updated to the latest google material library version 1.2.1 to display Range Widgets using Material Slider.

There is a slight clipping of the ripple effect for the vertical slider, but I found the same on the master branch as well. @seadowg and I decided to leave that for now. I will keep looking for solutions.

Form having both Range Integer and Decimal widgets like 
[Test Range Picker Widgets.xml.txt](https://github.com/getodk/collect/files/4917565/Test.Range.Picker.Widgets.xml.txt)
[Test Range Widget.xml.txt](https://github.com/getodk/collect/files/4931460/Test.Range.Widget.xml.txt)


collect_app/src/test/java/org/odk/collect/android/widgets/RangePickerDecimalWidgetTest.java
```diff
@@ -0,0 +1,147 @@
...
+        RangePickerDecimalWidget widget = createWidget(promptWithQuestionDefAndAnswer(rangeQuestion, null));
+        widget.binding.widgetButton.performClick();
+
+        NumberPickerDialog numberPickerDialog = (NumberPickerDialog) widgetActivity.getActivity().getSupportFragmentManager()
+                .findFragmentByTag(NumberPickerDialog.NUMBER_PICKER_DIALOG_TAG);
+        shadowOf(getMainLooper()).idle();
+
+        assertNotNull(numberPickerDialog);
+    }
+
```
SaumiaSinghal: @seadowg `clickingPickerButton_showsNumberPickerDialog()` unit test is failing. I'm looking onto it. If you have a hint, please let me know.

seadowg: Yeah if you debug you'll see that in `RangeWidgetUtils.showNumberPickerDialog` the `catch` branch will run as there is a `ClassCastException`. The code is trying to cast the `context` to a `FormEntryActivity` but it doesn't need to be one of those to add a `Fragment` right? I think it just needs to cast it to a `FragmentActivity` (`Activity` doesn't have `getSupportFragmentManager` on it).

I think you'll also need to have the activity in fully created state to add a fragment (rather than one that's just been "built").  Maybe just for those tests you can create a different test activity using `createThemedActivity`. The extra work is pretty slow so good to only use it in tests that need it!

SaumiaSinghal: Right I think that's why the tests simply failed instead of giving an error

seadowg (changes requested):
The Espresso tests are not compiling. I know the tests are slow and hard to run for every change but it's a good idea to at least be running `AllWidgetsFormTest` (as it's pretty relevant to the widgets) before marking these PRs as ready for review.

SaumiaSinghal:
@seadowg! Everything seems fine now. I'll check the espresso tests as well from now on.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2633,8 +2634,12 @@ public void onSaveFormIndexError(String errorMessage) {
     public void onNumberPickerValueSelected(int widgetId, int value) {
         if (currentView != null) {
             for (QuestionWidget qw : ((ODKView) currentView).getWidgets()) {
-                if (qw instanceof RangeWidget && widgetId == qw.getId()) {
-                    ((RangeWidget) qw).setNumberPickerValue(value);
+                if (qw instanceof RangePickerIntegerWidget && widgetId == qw.getId()) {
```
seadowg, line 2544: I'm not a big fan of the `FormEntryActivity` having to know about widget implementations here but this structure existed a long time before this PR and I think it's be awkward to unravel.

collect_app/src/test/java/org/odk/collect/android/views/TrackingTouchSliderTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    public void onStartTrackingTouch_suppressesFlingGesture() {
+        slider.onTouchEvent(MotionEventBuilder.newBuilder().setAction(ACTION_DOWN).build());
+        assertThat(slider.isTrackingTouch(), equalTo(true));
+    }
+
+    @Test
+    public void onStopTrackingTouch_doesNotSuppressFlingGesture() {
```
seadowg: Probably worth combining this with the test above so it tests that the `ACTION_UP` **after** the `ACTION_DOWN` works. With these tests the implementation could just set `isTrackingTouch` to `true` on an `ACTION_DOWN` and then never change it back 😉

collect_app/src/test/java/org/odk/collect/android/views/TrackingTouchSliderTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+@RunWith(RobolectricTestRunner.class)
+public class TrackingTouchSliderTest {
+
+    private TrackingTouchSlider slider;
+
+    @Before
+    public void setUp() {
+        ApplicationProvider.getApplicationContext().setTheme(R.style.Theme_Collect_Light);
+
```
seadowg: Oh this is very cool! It never occurred to me to just set the application context theme. This could probably help us speed up tests where we're building an `Activity` just to set the context theme. Maybe you could add this to `RobolectricHelpers` as `createThemedContext(int theme)` so we  can use it other places?

collect_app/src/main/java/org/odk/collect/android/widgets/RangeDecimalWidget.java
```diff
@@ -18,20 +18,56 @@
...
+    TextView currentValue;
 
     public RangeDecimalWidget(Context context, QuestionDetails prompt) {
         super(context, prompt);
     }
 
+    @Override
+    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+        rangeQuestion = (RangeQuestion) getFormEntryPrompt().getQuestion();
+        RangeWidgetUtils.RangeWidgetLayoutElements layoutElements = RangeWidgetUtils.setUpLayoutElements(context, prompt);
```
seadowg, line 51: I think in future this could probably be extracted to its own custom view but for the moment it's great progress!

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -43,18 +44,20 @@ public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelp
 
     @Override
     protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
-        binding = UrlWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+        binding = WidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
```
seadowg: I think this layout should be called `button_and_answer` or something. I don't like suggesting that it's the "default" answer view.

collect_app/src/main/res/layout/range_widget_horizontal.xml
```diff
@@ -31,16 +34,15 @@ limitations under the License.
...
-        <SeekBar
-            android:id="@+id/seek_bar"
+        <org.odk.collect.android.views.TrackingTouchSlider
+            android:id="@+id/slider"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:theme="@style/Base.Widget.AppCompat.SeekBar.Discrete" />
+            app:thumbColor="?colorSecondary"
+            app:tickColor="?colorOnBackground"
+            app:haloColor="?dividerColor"
```
seadowg: I don't think we should use `dividerColor` here. Could you find a color to use in the standard [material color set](https://material.io/develop/android/theming/color) and use that instead? If it isn't in our themes already you might have to define it. 

If it seems doable you could then replace `dividerColor` everywhere and delete it? No worries if that seems like a distraction though.

collect_app/src/main/res/layout/range_widget_vertical.xml
```diff
@@ -33,22 +34,18 @@ limitations under the License.
...
         android:layout_height="330dp">
 
-        <SeekBar
-            android:id="@+id/seek_bar_no_ticks"
+        <org.odk.collect.android.views.TrackingTouchSlider
+            android:id="@+id/slider"
             android:layout_width="330dp"
             android:layout_height="wrap_content"
+            app:thumbColor="?colorSecondary"
+            app:tickColor="#696969"
```
seadowg: This should use theme colors rather than hardcoded colors.

seadowg (changes requested):
Just a few changes but looking great!

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
SaumiaSinghal: @seadowg I think the other option that we can use is having a separate `viewModel` for `RangeWidget`. I also don't like having a callback implemented in `FormEntryActivity`.

collect_app/src/main/java/org/odk/collect/android/widgets/RangeDecimalWidget.java
SaumiaSinghal: Yes correct @seadowg Thank you!

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
SaumiaSinghal: I renamed it to `widget_button_and_answer_layout`

collect_app/src/main/res/layout/range_widget_horizontal.xml
SaumiaSinghal: Yes sure

collect_app/src/test/java/org/odk/collect/android/views/TrackingTouchSliderTest.java
SaumiaSinghal: Yes sure @seadowg :)

collect_app/src/main/res/values/attrs.xml
```diff
@@ -2,6 +2,7 @@
 <resources>
     <declare-styleable name="Collect">
         <attr name="iconColor" format="color" />
+        <attr name="sliderTrackColor" format="color" />
```
seadowg: I think you misunderstood what I was trying to communicate before. I just meant if we were missing one of the standard material theme colors we might have to add it to our themes. Sorry if that was confusing! 

The theme shouldn't have to know that there is a slider that needs a color right? The slider should use a material theme color we have a value for not a new custom one.

seadowg (changes requested):

collect_app/src/main/res/values/attrs.xml
SaumiaSinghal: Oh now I get it. We have values for almost all the theme colors except for `colorBackground`. However, I feel that I shouldn't be using it as `sliderTrackColor`, as it doesn't sound correct. I feel like I should use `colorOnBackground` as slider track color. Am I correct? I found that we are not using the value of  `colorOnBackground` anywhere in the app. Should I update it?

seadowg: I'd say `colorOnSurface` probably makes sense?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/NumberPickerDialog.java
```diff
@@ -55,12 +53,7 @@ public static NumberPickerDialog newInstance(int widgetId, String[] displayedVal
     @Override
     public void onAttach(Context context) {
         super.onAttach(context);
-
-        try {
-            listener = (NumberPickerListener) context;
-        } catch (ClassCastException e) {
-            throw new ClassCastException(context.toString() + " must implement OnHeadlineSelectedListener");
-        }
+        rangePickerViewModel = new ViewModelProvider(requireActivity(), new ViewModelProvider.NewInstanceFactory()).get(RangePickerViewModel.class);
```
seadowg: I don't think you need to pass a `NewInstanceFactory` here. Could you not just use the 1 arg constructor for `ViewModelProvider`?

collect_app/src/main/java/org/odk/collect/android/widgets/RangePickerDecimalWidget.java
```diff
@@ -0,0 +1,100 @@
...
+    private int progress;
+
+    public RangePickerDecimalWidget(Context context, QuestionDetails questionDetails) {
+        super(context, questionDetails);
+        RangePickerViewModel rangePickerViewModel = new ViewModelProvider(((ScreenContext) getContext()).getActivity(),
+                new ViewModelProvider.NewInstanceFactory()).get(RangePickerViewModel.class);
+
+        rangePickerViewModel.getNumberPickerValue().observe(((ScreenContext) getContext()).getViewLifecycle(), answer -> {
+            if (answer != null) {
+                progress = rangePickerViewModel.getNumberPickerProgress(rangeStart, rangeStep, rangeEnd, answer);
```
seadowg: I feel like `getNumberPickerProgress` should probably remain a util method as it doesn't rely on any state in the view model.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/widgets/RangePickerDecimalWidgetTest.java
```diff
@@ -0,0 +1,128 @@
...
+        when(rangeQuestion.getRangeStep()).thenReturn(new BigDecimal("0.5"));
+    }
+
+    @Test
+    public void whenRangeEndIsGreaterThanRangeStart_widgetsSetsCorrectValuesForNumberPicker() {
+        RangePickerDecimalWidget widget = createWidget(promptWithQuestionDefAndAnswer(rangeQuestion, null));
+        int index = 0;
+
+        for (double i = 5.5F; i >= 1.5F; i -= 0.5F, index++) {
+            assertThat(widget.displayedValuesForNumberPicker[index], equalTo(String.valueOf(i)));
```
seadowg, line 48: I'm not a fan of testing this field here as there is no reason for it to be public. Really the test should click on the button and test that the dialog receives the right values. I know you're reworking how all that functions though so I think we go with this for the moment!

As a general tip for tests like this I would avoid the `for` loop and just write out what I expect the values to be. That way it's clearer to the reader of the test. Tests serve as documentation as much as they serve as checks so it's good for them to be very clear what the behaviour they expect is.

seadowg (changes requested):
Swiping the slider all the way to left or right cuts of the grey highlight circle. I think we should work out if we can fix that before movign this to QA.

seadowg (approved):

mmarciniak90:
@SaumiaSinghal take a look at mentioned issues

|Description |released v1.27.3|PR|
|---|---|---|
|When you open Range widgets it looks like an answer is added because a blue dot is visible|Screenshot_20200807-164529|Screenshot_2020-08-07-16-48-34-217_org odk collect android|
|Read Only question is not grey out. There is no difference between editable and not editable question. |Screenshot_20200807-164753|Screenshot_2020-08-07-16-51-10-164_org odk collect android|
|Read only question with filled answer is not grey out |Screenshot_20200807-170356|Screenshot_2020-08-07-17-02-48-815_org odk collect android|
|I'm not sure about showing the inactive button for the read-only picker. Sorry If I missed this requirement|Screenshot_20200807-164921|Screenshot_2020-08-07-16-52-26-927_org odk collect android|

mmarciniak90:

SaumiaSinghal:
Hi @mmarciniak90 I fixed the mentioned issues with the Ui of the disabled slider and the picker button. Can you please test again?

kkrawczyk123:
@SaumiaSinghal I can see another issue connected to dot displayed on the axis without choosing any answer when I follow scenario:
1. Open any range widget with an axis
2. Do not choose any number
3. Rotate screen - unexpected dot appears on the axis and the lowest value is selected automatically
Dot is still visible after rotating back to portrait.
Screenshot (Aug 18, 2020 12_02_11 PM)

SaumiaSinghal:
Hi @kkrawczyk123! Thanks for reporting the issue. I fixed it. Can you test again please?

grzesiek2010:
@SaumiaSinghal does this pr fix https://github.com/getodk/collect/issues/4030 too?

SaumiaSinghal:
Hi @grzesiek2010! This PR https://github.com/getodk/collect/pull/3995 fix #4030. It involves work on top of this PR.

kkrawczyk123:
@SaumiaSinghal I have noticed an issue when select RTL language and then rotate the screen, axis in LTR is visible after screen rotation, see the gif: 
ezgif com-video-to-gif (35)

SaumiaSinghal:
Thanks @kkrawczyk123 for reporting the issue. Sorry for the late reply, I somehow missed the message. It seems there was an issue that the RTL language was getting reset on the branch on screen orientation. I rebased with the master branch and everything works fine now. Also, I updated google material library to a more stable and latest version 1.2.1.

kkrawczyk123:

kkrawczyk123:
**Tested with success!**
Verified on Androis: 5.1, 7.1 and 9.0.

Tested scenarios:
- light/dark mode
- RTL lamguage
- rotate
- minimize 
- going backward/forward
- readonly
- jump from hierarchy 
- remove response 
- dont keep activities turned on/off

mmarciniak90:
Verified also on Android versions: 7.0, 8.1, 10.0

grzesiek2010:
@SaumiaSinghal please fix the conflicts

lognaturel:
Thanks so much for getting this over the finish line, @SaumiaSinghal! 🎉 

There's a little bit of risk here since we can't easily track the conflict resolution through the rebase but any issues would be caught on usage or at least on regression testing.

I went back and forth on whether to squash. I think there are several commits that do provide hints as to decisions made so I decided to keep them all.

PR #3923, opened 2020-06-14
Migrate to `androidx` Library for Implementing Preferences in Collect

SaumiaSinghal:
Work towards #3849

#### What has been done to verify that this works as intended?
I tested all the preference screen and preference dialogs on every theme. It fixes the material contained button on dialogs. I checked it for all the dialogs shown on the settings preferences. There is a minor UI issue of indentation on `MapsPreference` Screen.

#### Why is this the best possible solution? Were any other approaches considered?
I replaced the `android` Preference library with `androidx` one, as the old framework is deprecated now. It solves a lot of visual issues related to the dialogs shown on Settings Preference in new Experimental theme.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
I found that in the new `androidx` preference library, the dialogs are separated from their `DialogPreference` class. Most of the changes are for shifting the code that concerns the dialog from the `DialogPreference` class to a newly created `PreferenceDialogFragment` class. I was careful not to add any new behavior while working on this. However, I feel the part of the settings preference that is most prone to regression risk is the `ReferenceLayerDialog` and `ResetSettingsDialog`.   

#### Do we need any specific form for testing your changes? If so, please attach one.
No

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
No

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/opendatakit/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #3977, opened 2020-07-21
Revise Form Management for Match Exactly

### LINKED ISSUE #3940: Update Form Management for Match Exactly

seadowg:
Closes #3940

In addition to implementing the feature as descirbed this:

* Removes Evernote Job framework
* Backfills tests for manual and "previously downloaded only" form updates
* Moves testing around to create a `smoke` package that covers core parts of the app
* Adds ` TestRuleChain` that makes it easier to write Espresso tests with standard deps and rules

A lot of new tests and manually verified acceptance criteria.

Not a lot of big changes here that don't just build on the previous Match Exactly PRs so I think we're good on discussing structural decisions here.

There are risks to any of the auto update/auto upload features. I'm keeping track of risks and will address with QA when we do checks on `master`.


collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrator.java
```diff
@@ -104,11 +107,11 @@ public StorageMigrationResult migrate() {
     }
 
     private boolean isFormUploaderRunning() {
-        return backgroundWorkManager.isFormUploaderRunning();
+        return formSubmitManager.isSubmitRunning();
```
seadowg, line 110: I think really we just need a shared "lock" in the app that stops things that make changes running at the same time as here we have to care about multiple things, not everything is covered and there is no real thread safety. I've noted this down and I want to work on it later once we've added error handling for match exactly.

collect_app/src/main/res/values/strings.xml
```diff
@@ -707,4 +707,7 @@
     <string name="legacy_custom_server_paths_summary">Will be removed in a future version. Please use /formList and /submission on your server.</string>
     <string name="experimental">Experimental</string>
     <string name="match_exactly">Exactly match server</string>
+    <string name="update_blank_forms">Update blank forms</string>
+    <string name="manually">Manually</string>
+    <string name="form_update_mode_title">Mode</string>
```
lognaturel: Consider "Blank form update mode"? With "Mode" you can blink and kind of forget what you're setting the mode of. I think it can use the same string as for the Admin setting (`form_update_mode`).

async/src/main/java/org/odk/collect/async/Scheduler.kt
```diff
@@ -4,18 +4,27 @@ import java.util.function.Consumer
 import java.util.function.Supplier
 
 /**
- * Runs tasks in the foreground and background
+ * Run and schedule tasks in the foreground (UI thread) and background. Based on terminology
+ * used in Android's Background Processing documentation: https://developer.android.com/guide/background
```
lognaturel, line 8: I think this is a good call and turned out nicely. 👍

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.java
```diff
@@ -51,6 +51,29 @@ public Scheduler providesScheduler(WorkManager workManager) {
...
+
+    @Test
+    public void whenMatchExactlyEnabled_disablesPrefs() {
+        rule.mainMenu()
+                .clickOnMenu()
+                .clickGeneralSettings()
+                .clickFormManagement()
+                .clickUpdateForms()
+                .clickOption(R.string.match_exactly)
+                .assertDisabled(R.string.automatic_download);
```
lognaturel, line 78: Additionally, I think the checkbox should be selected to make it clear that updates always happen automatically in match exactly mode. If the mode changes from match exactly to one of the other modes, the checkbox should be unchecked to indicate that automatic downloads aren't applicable to manual mode or are not on by default for updates to existing forms.

I think it's ok for the value for the update frequency to stay whatever it was last set to in the case of manual update but I find the disabled automatic download option to be confusing.

async/src/main/java/org/odk/collect/async/Scheduler.kt
```diff
@@ -46,4 +37,13 @@ interface Scheduler {
...
     fun isRunning(tag: String): Boolean
+
+    /**
+     * Run a task and then repeat in the foreground
+     *
+     * @param foreground the task to be run
+     * @param repeatPeriod the period between each run of the task
+     * @return object that allows task to be cancelled
+     */
+    fun repeat(foreground: Runnable, repeatPeriod: Long): Cancellable
```
lognaturel, line 48: Yeah, ok, `repeat` is a better name.

collect_app/src/main/java/org/odk/collect/android/application/initialization/MetaPreferenceMigrator.java
```diff
@@ -32,26 +31,22 @@
 /**
  * Migrates old preference keys and values to new ones.
  */
-public class CollectPreferenceMigrator implements PreferenceMigrator {
+public class MetaPreferenceMigrator implements PreferenceMigrator {
```
lognaturel: I don't understand the changes here. Why not pass in all the `SharedPreferences` on initialization? It seems they're always available on initialization or it would be more consistent to pass everything in on migration. I thought it was for testing but the one test that uses it does construction and migration on the same line.

collect_app/src/test/java/org/odk/collect/android/application/initialization/migration/KeyExtractorTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    public void whenOldKeyMissing_doesNothing() {
+        initPrefs(prefs);
+
+        extractNewKey("newKey").fromKey("oldKey")
+                .fromValue("oldBlah").toValue("newBlah")
+                .apply(prefs);
+
+        assertPrefsEmpty(prefs);
+    }
```
lognaturel, line 70: And what about the actually does something case?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2484,16 +2480,7 @@ public void next() {
...
-                .setRequiredNetworkType(NetworkType.CONNECTED)
-                .build();
-        OneTimeWorkRequest autoSendWork =
-                new OneTimeWorkRequest.Builder(AutoSendWorker.class)
-                        .addTag(AutoSendWorker.TAG)
-                        .setConstraints(constraints)
-                        .build();
-        workManager.beginUniqueWork(AutoSendWorker.TAG,
-                ExistingWorkPolicy.KEEP, autoSendWork).enqueue();
+        formSubmitManager.scheduleSubmit();
```
lognaturel: Since you're replacing with a single line that now has very clear behavior, maybe get rid of this method?

I think the TODO should be moved to an issue. Yes, I would expect that if auto send is turned on, that immediately attempts to send all pending submissions or enqueues the work if no connection is available. If you're ok with all that I can file the (very much not urgent) issue.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/SchedulerFormUpdateAndSubmitManager.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private void scheduleMatchExactlySync(long repeatPeriod) {
+        scheduler.networkDeferred(MATCH_EXACTLY_SYNC_TAG, new SyncFormsTaskSpec(), repeatPeriod);
+    }
+
+    private void scheduleAutoUpdate(long repeatPeriod) {
+        scheduler.networkDeferred(AUTO_UPDATE_TAG, new AutoUpdateTaskSpec(), repeatPeriod);
+    }
+
+    private void cancelWork() {
```
lognaturel: Seems either this should cancel submission as well or should be renamed.

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
```diff
@@ -145,4 +168,22 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
         });
     }
 
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+        if (key.equals(KEY_FORM_UPDATE_MODE) || key.equals(KEY_PERIODIC_FORM_UPDATES_CHECK)) {
+            formUpdateManager.scheduleUpdates();
```
lognaturel, line 174: I think `whenPreviouslyDownloadedOnlyDisabled_stopsCheckingForUpdates` verifies that if the mode is changed to manual after being one of the automatic ones, automatic work is cancelled. However, I can't find code for that. Are you sure this line doesn't need to be outside the `if` to guarantee that a change to `manual` cancels work?

collect_app/src/main/java/org/odk/collect/android/preferences/Protocol.java
```diff
@@ -0,0 +1,29 @@
+package org.odk.collect.android.preferences;
+
+import android.content.Context;
+
+import org.odk.collect.android.R;
+
+public enum Protocol {
```
lognaturel, line 7: As more of these simple enums get introduced, maybe we should consider constants with typedef annotations? https://developer.android.com/studio/write/annotations.html#enum-annotations It's not a huge amount of memory but it does add up and I think constants/typedef is just as nice.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormRepository.java
```diff
@@ -48,4 +49,32 @@ public void save(Form form) {
...
+    public void setLastDetectedUpdated(String jrFormId, String formHash, String manifestHash) {
+        String formVersionHash = MultiFormDownloader.getMd5Hash(formHash) + manifestHash;
+
+        ContentValues values = new ContentValues();
+        values.put(LAST_DETECTED_FORM_VERSION_HASH, formVersionHash);
+        new FormsDao().updateForm(values, JR_FORM_ID + "=?", new String[]{jrFormId});
+    }
+
+    @Override
+    public Form getByLastDetectedUpdate(String formHash, String manifestHash) {
```
lognaturel, line 63: Nit: `lastDetectedHash`?

collect_app/src/main/java/org/odk/collect/android/formmanagement/previouslydownloaded/ServerFormsUpdateChecker.java
```diff
@@ -0,0 +1,79 @@
...
+    private List<ServerFormDetails> fetchUpdatedForms() throws FormApiException {
+        List<ServerFormDetails> formList = null;
+
+        try {
+            formList = serverFormsDetailsFetcher.fetchFormDetails();
+        } catch (FormApiException e) {
+            switch (e.getType()) {
+                case AUTH_REQUIRED:
+                    try {
+                        serverFormsDetailsFetcher.fetchFormDetails();
```
lognaturel, line 56: What's the point of the single retry?

lognaturel (changes requested):
Phew! This represents a lot of work. Great to see all the new tests, and especially the smoke tests.

collect_app/src/main/java/org/odk/collect/android/formmanagement/previouslydownloaded/ServerFormsUpdateChecker.java
seadowg: This is the existing behaviour (see [here](https://github.com/getodk/collect/blob/8e6591552913f6a69894c6a8f2ccdefa2246fa0a/collect_app/src/main/java/org/odk/collect/android/tasks/ServerPollingJob.java#L93)). I'd actually meant to ask about this as it felt really weird. I'm thinking we can just get rid of it? Otherwise, we should write a test to keep it in (and hopefully describe what it's about).

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormRepository.java
seadowg: I was actually trying to abstract away from the "hash" here. In the cases we're using it we want to know if this "update" or "version" of the form exists on the device. We don't need to hash anything to do that (it could have just stored the manifest hash rather than hashing it with the form hash) so to me this felt closer to what the intention was while hiding the implementation detail. Does that feel ok?

collect_app/src/main/java/org/odk/collect/android/preferences/Protocol.java
seadowg: I had it in my mind that the enum problem wasn't a big deal any more. I managed to find this [post from Chet Haase](https://medium.com/androiddevelopers/when-using-enums-and-r8-3f8f314c0a13) that goes through optimizations they've made in R8 for enums but also says:

> Yes, we used to talk about avoiding enums… but that was many years and an entire runtime ago — enums are fine

I think maybe to side step the typedef stuff we could consider switching enum definitons and their calling code candidates for switching to Kotlin where we could be using [`sealed` classes](https://kotlinlang.org/docs/reference/sealed-classes.html).

collect_app/src/main/res/values/strings.xml
seadowg: Sounds good!

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.java
seadowg: Oh that's a really nice touch. Good thinking.

async/src/main/java/org/odk/collect/async/Scheduler.kt
seadowg: Haha I still don't know. I'm trying to stop myself from changing it again...

collect_app/src/main/java/org/odk/collect/android/application/initialization/MetaPreferenceMigrator.java
seadowg: Not for testing it's just me playing with the interface. I've flipped-flopped on this a few times. I guess my thinking here is that in the `SettingsImporter` we're actively dealing with general and admin preferences so it's weird that we then just call `migrate`. There w know what we're trying to migrate we just don't care about the detail (and we also don't know that the meta prefs exist). 

I think it's also me not being able to escape from the discussions about having profiles in the future. In that world you'd have different general and shared prefs but one shared meta prefs. 

Any of that make sense or you still hate it?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah that sounds like a good way forward!

collect_app/src/test/java/org/odk/collect/android/application/initialization/migration/KeyExtractorTest.java
seadowg: Ha. This is my very hippie TDD style: I never needed to write that test to get the behaviour in there as it was driven out by the `MetaPreferencesMigratorTest` (which I wrote first). 

I think you're right that we should have the test for the actual behaviour though as these migrations are being treated like a framework - we'd want to have confidence that each one is tested outside in isolation.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/SchedulerFormUpdateAndSubmitManager.java
seadowg: Yeah good point. I think I'll rename for the moment.

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
seadowg: `SchedulerFormUpdateAndSubmitManager#scheduleUpdates` cancels any existing deferred work and then sets it back up again. The `if` here is just making sure we only do that when the relevant settings change (as this will fire for any changes).

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Created an issue: https://github.com/getodk/collect/issues/3980

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.java
seadowg: I think we should create an issue for this. Testing this is going to be really hard without converting the fragment to `androidx` and I think we're just going to end up with conflicts with #3923. What do you think? 

I'll put a note for QA to target that PR next as it'd be really helpful to get that in.

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
lognaturel: Ohhh, hah. Of course, that makes a lot more sense. I was going too quickly and interpreted it as if we go into match exactly or update downloaded modes.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormRepository.java
lognaturel: Fair enough, thanks for the explanation.

collect_app/src/main/java/org/odk/collect/android/preferences/Protocol.java
lognaturel: Ah, great! I'm pretty sure I saw the "avoid enum" thing in Android docs recently but maybe I'm remembering wrong. Let's leave it for now. I'm not super excited about peppering Kotlin around but we should discuss.

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.java
lognaturel: Ok, all sounds good!

collect_app/src/main/java/org/odk/collect/android/application/initialization/MetaPreferenceMigrator.java
lognaturel: The rationale for passing in general and admin preferences to migrate makes sense to me now, thanks. The name still really bothers me. `CollectPreferenceMigrator` or `SharedPrefsMigrator` or something that implies it can migrate all the prefs would make a lot more sense to me. Even if it doesn't keep track of general and admin preferences, it still knows how to migrate all the relevant preferences.

collect_app/src/main/java/org/odk/collect/android/formmanagement/previouslydownloaded/ServerFormsUpdateChecker.java
lognaturel: Wondered whether that might be old stuff. Hmm. In the foreground usage could it be giving an opportunity to show the dialog or something?

collect_app/src/main/java/org/odk/collect/android/application/initialization/MetaPreferenceMigrator.java
seadowg: Yeah that's very fair. I'll rework the name.

collect_app/src/main/java/org/odk/collect/android/formmanagement/previouslydownloaded/ServerFormsUpdateChecker.java
seadowg: Yeah that'd make more sense. Probably want to view that as another issue/feature though. I think for the moment I'm happy for this to merge in like this if we add an issue around making the auto update prompt the user for creds if there is an auth error (like Manual and Match Exactly both will). How does that sound?

lognaturel:
I think that with the `MetaPreferenceMigrator` rename and a couple issues filed, this can be merged.

collect_app/src/main/java/org/odk/collect/android/formmanagement/previouslydownloaded/ServerFormsUpdateChecker.java
seadowg: Issue filed: https://github.com/getodk/collect/issues/3983

lognaturel (approved):

PR #3986, opened 2020-07-24
Match exactly errors

seadowg:
Closes #3959

Although the intention of the issue is just to add errors for Match Exactly mode we talked back and forth about how possible it would be to also improve other error messages in the app. In the end only really Get Blank Form's initial form list fetch had improvements along with Match Exactly. I think any further changes would require us to spec out what we want to see for each flow. Plus, this PR was big enough already.

#### What has been done to verify that this works as intended?

Played around with the app and also added a lot of new tests (including back fills for existing flows)

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Again, keeping track of this for QA. I'd be tempted to say we should start playing with some of the features on master after we merge this as the core UI flows are now all in.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

Issue #3959, opened 2020-07-06
Add error handling to Match Exactly sync


seadowg:
**Acceptance**

- [x] Given I've set up Match Exactly and have navigated to Fill Blank Forms
And my server is down/not available (`UnknownHostException`)
When I tap ♻️ (refresh)
Then I see an error notification
When I click on the notification
Then it takes me to Fill Blank Form
And I see an error badge on the refresh button

- [x] Given I've set up Match Exactly and have navigated to Fill Blank Forms
And my credentials have been revoked/changed (a `401`)
When I tap ♻️ (refresh)
Then I see an error notification
When I click on the notification
Then it takes me to Fill Blank Form
And I see a prompt to fix credentials (like Get Blank Forms)
And I see an error badge on the refresh button

- [x] Given I've set up Match Exactly and have navigated to Fill Blank Forms
When I tap ♻️ (refresh)
And then **ANY** (includes form list, manifest or download) of the network requests fail
Then I see an error notification
When I click on the notification
Then it takes me to Fill Blank Form
And I see an error badge on the refresh button

- [x] Given I've set up Match Exactly and have navigated to Fill Blank Forms
When I tap ♻️ (refresh)
And then there is file system error
Then I see an error notification
When I click on the notification
Then it takes me to Fill Blank Form
And I see an error badge on the refresh button

- [x] Given I've had an error
When I navigate to Fill Blank Forms
When I tap ♻️ (refresh)
And the sync succeeds
Then the error state disappears

- [x] Given I've set up Match Exactly and have navigated to Fill Blank Forms
And the server returns bad form list and or manifest responses
When I tap ♻️ (refresh)
Then I see an error notification
When I click on the notification
Then it takes me to Fill Blank Form
And I see an error badge on the refresh button

- [x] All errors while syncing in the background should result in same notification and error badge

seadowg:
@lognaturel good to get your thoughts here. The copy is straight off the top of my head so could definitely use some work but I wanted to focus on capturing the different error states and how we present them. My logic here was to continue to use a dialog for most errors as they involve trying to push the enumerator to check details/contact someone. For the "sync error" a banner felt more appropriate as we want to make sure we keep alerting them to the problem but without blocking them.

lognaturel:
[...]

PR #3992, opened 2020-07-29
Allow forms to be edited after blank form deletion

### LINKED ISSUE #3961: Soft delete form definitions if device has corresponding form instances
Currently, manually deleting a form definition removes both the database entry and the files for that form definition. With the addition of a form download mode that exactly matches the form list on the server (#3956), this is problematic. A user might still have submissions for a form that will no longer be in their formlist because it is marked as closing. They may need to edit those submissions and doing so requires the form definition. It's possible to get in this bad situation with manual delete of form definitions as well -- users may not expect that a side effect of deleting a form definition is that they can no longer open filled forms.

Instead of always hard deleting form definitions, hard delete should only occur for a formID/version combination if there are no corresponding submissions. If there are corresponding submissions, the formID/version combination's database entry should be updated to indicate soft delete and the files should remain. That formID/version combination should not be visible anywhere in the user interface. As soon as the last submission for that formID/version combination is deleted, the files and database entry for that formID/version combination should be fully removed.

- [x] Given I've enabled "Match Exactly"
 And there's a form definition formID/version combination currently on the device that is not in the server-provided form list
 And that form definition formID/version combination **does not** have corresponding submissions
 When I visit Fill Blank Forms
 And I tap the ♻️ (refresh) icon
 That form definition formID/version combination's files and database entry should be fully deleted

- [x] Given there's a form definition formID/version combination currently on the device that **does not** have corresponding submissions
 When I visit Delete Saved Form
 And I select that Blank Form and delete it
 That form definition formID/version combination's files and database entry should be fully deleted

- [x] Given I've enabled "Match Exactly"
 And there's a form definition formID/version combination currently on the device that is not in the server-provided form list
 And that form definition formID/version combination **does** have corresponding submissions
 When I visit Fill Blank Forms
 And I tap the ♻️ (refresh) icon
 That form definition formID/version combination's database entry should be modified to indicate that it has been soft deleted
 And that form definition formID/version combination should not appear anywhere in the user interface

- [x] Given there's a form definition formID/version combination currently on the device that **does** have corresponding submissions
 When I visit Delete Saved Form
 And I select that Blank Form and delete it
 That form definition formID/version combination's database entry should be modified to indicate that it has been soft deleted
 And that form definition formID/version combination should not appear anywhere in the user interface

- [x] Given there's a form definition formID/version combination that has been soft deleted
 When I visit Delete Saved Form
 And I delete the last Saved Form corresponding to that formID/version combination
 That form definition formID/version combination's files and database entry should be fully deleted

- [x] Given there's a form definition formID/version combination that has been soft deleted
 When the last Saved Form corresponding to that formID/version combination is auto deleted
 That form definition formID/version combination's files and database entry should be fully deleted

Note: it's also possible for a user to delete instances through `adb` or an external file manager. I don't believe `InstanceSyncTask` currently identifies this kind of change and I think it would be fine not to address that case.

I believe all of this could be handled by modifying the implementations for `FormsProvider.delete` (to look for corresponding instances to decide whether to soft or hard delete) and `InstanceProvider.delete` (to look for a corresponding soft deleted form definition that may need to be hard deleted).

### LINKED ISSUE #3991: Experimental settings section should be hidden or removed
`master`

#### Problem description
There's an Experimental settings section that was added temporarily while working towards match exactly. It should be hidden or removed now that nothing is in it. @seadowg maybe you had a plan for it already?

Go to general settings

#### Expected behavior
The Experimental section doesn't show up

seadowg:
Closes #3961. Closes #3991. ~~Blocked by #3986.~~

This also introduces and uses a `ChangeLock` object that lets make sure that work that updates the forms or instances on device doesn't interfere with each other. There are probably more places this could be used but I focused on making sure places that already made an effort in this direction were covered.

Tried out flows manually and wrote a bunch of new tests.

To me it made the most sense to implement this using method objects rather than trying to change the delete logic at the bottom level. This means the logic is separate from any of the implementation of the repositories.

With this I think we want to merge to master and then QA there for the whole match exactly feature set.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
@@ -33,4 +33,28 @@ public class DeleteBlankFormTest {
                 .clickFillBlankForm()
                 .assertTextDoesNotExist("One Question");
     }
+
+    @Test
+    public void deletingAForm_whenThereFillForms_allowsEditing() {
+        rule.mainMenu()
+                .startBlankForm("One Question")
+                .answerQuestion("what is your age", "22")
+                .swipeToEndScreen()
+                .clickSaveAndExit()
+
+                .clickDeleteSavedForm()
+                .clickBlankForms()
+                .clickForm("One Question")
+                .clickDeleteSelected(1)
+                .clickDeleteForms()
+                .assertTextDoesNotExist("One Question")
+                .pressBack(new MainMenuPage(rule))
+
+                .clickEditSavedForm()
+                .clickOnForm("One Question")
+                .clickOnQuestion("what is your age")
+                .answerQuestion("what is your age", "30")
+                .swipeToEndScreen()
+                .clickSaveAndExit();
+    }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
@@ -35,7 +35,7 @@ public class DeleteBlankFormTest {
     }
 
     @Test
-    public void deletingAForm_whenThereFillForms_allowsEditing() {
+    public void deletingAForm_whenThereFilledForms_allowsEditing() {
         rule.mainMenu()
                 .startBlankForm("One Question")
                 .answerQuestion("what is your age", "22")
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
@@ -1,0 +1,36 @@
+package org.odk.collect.android.feature.formmanagement;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.TestRuleChain;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+@RunWith(AndroidJUnit4.class)
+public class DeleteBlankFormTest {
+
+    public final CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public final RuleChain chain = TestRuleChain.chain()
+            .around(new CopyFormRule("one-question.xml"))
+            .around(rule);
+
+    @Test
+    public void deletingAForm_removesFormFromBlankFormList() {
+        rule.mainMenu()
+                .clickDeleteSavedForm()
+                .clickBlankForms()
+                .clickForm("One Question")
+                .clickDeleteSelected(1)
+                .clickDeleteForms()
+                .pressBack(new MainMenuPage(rule))
+                .clickFillBlankForm()
+                .assertTextDoesNotExist("One Question");
+    }
+}
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
@@ -22,7 +22,7 @@ public class DeleteFilledFormTest {
             .around(rule);
 
     @Test
-    public void deletingAForm_removesFormFromBlankFormList() {
+    public void deletingAForm_removesFormFromFinalizedForms() {
         rule.mainMenu()
                 .startBlankForm("One Question")
                 .answerQuestion("what is your age", "30")
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
@@ -1,0 +1,39 @@
+package org.odk.collect.android.feature.instancemanagement;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.TestRuleChain;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+@RunWith(AndroidJUnit4.class)
+public class DeleteFilledFormTest {
+
+    public final CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public final RuleChain chain = TestRuleChain.chain()
+            .around(new CopyFormRule("one-question.xml"))
+            .around(rule);
+
+    @Test
+    public void deletingAForm_removesFormFromBlankFormList() {
+        rule.mainMenu()
+                .startBlankForm("One Question")
+                .answerQuestion("what is your age", "30")
+                .swipeToEndScreen()
+                .clickSaveAndExit()
+
+                .clickDeleteSavedForm()
+                .clickForm("One Question")
+                .clickDeleteSelected(1)
+                .clickDeleteForms()
+                .pressBack(new MainMenuPage(rule))
+                .assertNumberOfFinalizedForms(0);
+    }
+}
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -116,7 +116,7 @@ the specific language governing permissions and limitations under the License.
         <activity android:name=".activities.FillBlankFormActivity" />
         <activity android:name=".activities.FormDownloadListActivity" />
         <activity
-            android:name=".activities.FileManagerTabs"
+            android:name=".activities.DeleteSavedFormActivity"
             android:configChanges="orientation|screenSize" />
         <activity
             android:name=".activities.GoogleSheetsUploaderActivity"
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
@@ -12,7 +12,11 @@ public class ReentrantLockChangeLock implements ChangeLock {
         try {
             return function.apply(lock.tryLock());
         } finally {
-            lock.unlock();
+            try {
+                lock.unlock();
+            } catch (IllegalMonitorStateException ignored) {
+                // Ignored
+            }
         }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
@@ -1,0 +1,18 @@
+package org.odk.collect.android.backgroundwork;
+
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
+
+public class ReentrantLockChangeLock implements ChangeLock {
+
+    private final ReentrantLock lock = new ReentrantLock();
+
+    @Override
+    public <T> T withLock(Function<Boolean, T> function) {
+        try {
+            return function.apply(lock.tryLock());
+        } finally {
+            lock.unlock();
+        }
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
@@ -34,6 +34,7 @@ import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.AUT
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.AUTO_SEND;
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.BASE64_RSA_PUBLIC_KEY;
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.DATE;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.DELETED;
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.DESCRIPTION;
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.DISPLAY_NAME;
 import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.DISPLAY_SUBTEXT;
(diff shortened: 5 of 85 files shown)


collect_app/src/main/res/xml/form_management_preferences.xml
```diff
@@ -14,7 +14,8 @@
             android:entries="@array/form_update_mode_entries"
             android:entryValues="@array/form_update_mode_values"
             android:key="form_update_mode"
-            android:title="@string/form_update_mode_title" />
+            android:title="@string/form_update_mode_title"
+            app:iconSpaceReserved="false" />
```
lognaturel: Oops, I already merged Saumia's change for this. Probably should have assigned it to you in the first place.

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationResult.java
```diff
@@ -6,19 +6,16 @@
...
-    MOVING_FILES_FAILED;
+    MOVING_FILES_FAILED,
+    CHANGES_IN_PROGRESS;
 
     public String getErrorResultMessage(Context context) {
         String errorMessage = context.getString(R.string.error) + " ";
         switch (this) {
             case NOT_ENOUGH_SPACE:
                 return errorMessage + context.getString(R.string.storage_migration_not_enough_space);
-            case FORM_UPLOADER_IS_RUNNING:
```
lognaturel, line 19: I don't understand this.

collect_app/src/main/java/org/odk/collect/android/notifications/NotificationManagerNotifier.java
```diff
@@ -84,11 +84,10 @@ public void onUpdatesDownloaded(HashMap<ServerFormDetails, String> result) {
     @Override
     public void onSyncFailure(FormApiException exception) {
         Intent intent = new Intent(application, FillBlankFormActivity.class);
-        intent.putExtra(FillBlankFormActivity.EXTRA_AUTH_REQUIRED, true);
 
-//        if (exception.getType() == FormApiException.Type.AUTH_REQUIRED) {
-//
-//        }
+        if (exception.getType() == FormApiException.Type.AUTH_REQUIRED) {
```
lognaturel: Ah, I see. Given this, I'll merge #3986 so you can clear out those commits here and not deal with merge conflicts.

collect_app/src/main/res/xml/form_management_preferences.xml
seadowg: No worries!

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationResult.java
seadowg: With the change lock we don't know which background service is running, but we know we shouldn't start the migration. I figured given these cases were rare it'd be ok to combine them into one error message?

The message should definitely change though. How about "Automatic jobs are running. Please try again later."?

collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
```diff
@@ -0,0 +1,60 @@
...
+                .clickForm("One Question")
+                .clickDeleteSelected(1)
+                .clickDeleteForms()
+                .pressBack(new MainMenuPage(rule))
+                .clickFillBlankForm()
+                .assertTextDoesNotExist("One Question");
+    }
+
+    @Test
+    public void deletingAForm_whenThereFillForms_allowsEditing() {
```
lognaturel: `whenThereFillForms`? When there are filled forms?

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public final CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public final RuleChain chain = TestRuleChain.chain()
+            .around(new CopyFormRule("one-question.xml"))
+            .around(rule);
+
+    @Test
+    public void deletingAForm_removesFormFromBlankFormList() {
```
lognaturel: this should be filled form list?

collect_app/src/main/AndroidManifest.xml
```diff
@@ -116,7 +116,7 @@ the specific language governing permissions and limitations under the License.
         <activity android:name=".activities.FillBlankFormActivity" />
         <activity android:name=".activities.FormDownloadListActivity" />
         <activity
-            android:name=".activities.FileManagerTabs"
+            android:name=".activities.DeleteSavedFormActivity"
```
lognaturel, line 119: We should introduce an alias for this. I guess I only put in aliases for activities with intent-filters but it's entirely possible applications would use explicit intents.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormDeleter.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public FormDeleter(FormsRepository formsRepository, InstancesRepository instancesRepository) {
+        this.formsRepository = formsRepository;
+        this.instancesRepository = instancesRepository;
+    }
+
+    public void delete(Long id) {
+        Form form = formsRepository.get(id);
+        List<Instance> instances = instancesRepository.getAllByJrFormId(form.getJrFormId());
+        long instancesForVersion = instances.stream().filter(instance -> Objects.equals(instance.getJrVersion(), form.getJrVersion())).count();
```
lognaturel: Just use `getAllByJrFormIdAndJrVersion`?

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormsRepository.java
```diff
@@ -34,6 +37,22 @@ public boolean contains(String jrFormId) {
...
+    public Form get(Long id) {
+        try (Cursor cursor = new FormsDao().getFormsCursor(_ID + "=?", new String[]{id.toString()})) {
+            return getFormOrNull(cursor);
+        }
+    }
+
+    @Nullable
+    @Override
+    public Form get(String jrFormId, String jrVersion) {
+        try (Cursor cursor = new FormsDao().getFormsCursor(JR_FORM_ID + "=? AND " + JR_VERSION + "=?", new String[]{jrFormId, jrVersion})) {
```
lognaturel, line 51: Why not use `getFormsCursor(String formId, String formVersion)`? I think you may need some `@NonNull` annotations if you just do this.

collect_app/src/main/java/org/odk/collect/android/fragments/DataManagerList.java
```diff
@@ -56,8 +62,16 @@
     private InstanceSyncTask instanceSyncTask;
     private ProgressDialog progressDialog;
 
-    public static DataManagerList newInstance() {
```
lognaturel, line 59: These names are really reprehensible. Maybe another thing to note for v1.29 cleanup is going further down the line of the `DeleteSavedFormActivity` rename.

collect_app/src/main/java/org/odk/collect/android/tasks/DeleteInstancesTask.java
```diff
@@ -100,8 +97,9 @@ public void setDeleteListener(DeleteInstancesListener listener) {
         deleteInstancesListener = listener;
     }
 
-    public void setContentResolver(ContentResolver resolver) {
-        contentResolver = resolver;
+    public void setRepositories(InstancesRepository instancesRepository, FormsRepository formsRepository) {
```
lognaturel: Why not use constructor params? Seems catastrophic to not call this so why even make it possible?

collect_app/src/main/res/layout/main_menu.xml
```diff
@@ -102,7 +102,7 @@
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:layout_marginTop="@dimen/margin_extra_small"
-                        tools:text="@string/manage_files" />
```
lognaturel: Typo?

collect_app/src/main/java/org/odk/collect/android/storage/migration/StorageMigrationResult.java
lognaturel: Whoops, this was a note to self I meant to come back to! But luckily you knew why I was confused.

Ok, I agree a single message is sufficient. I don't think that users know what "jobs" means in that context. How about "Automatic send or blank form download is running. Please try again later." You'll also need to change the resource name.

collect_app/src/main/java/org/odk/collect/android/tasks/DeleteFormsTask.java
```diff
@@ -96,15 +95,16 @@ public void setDeleteListener(DeleteFormsListener listener) {
...
-
     public int getDeleteCount() {
         return successCount;
     }
 
     public int getToDeleteCount() {
         return toDeleteCount;
     }
+
+    public void setRepositories(FormsRepository formsRepository, InstancesRepository instancesRepository) {
```
lognaturel: Why not use parameter args to guarantee these are set? This seems dangerous.

collect_app/src/main/java/org/odk/collect/android/database/helpers/FormsDatabaseHelper.java
```diff
@@ -125,7 +128,7 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
     public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         try {
             SQLiteUtils.dropTable(db, FORMS_TABLE_NAME);
-            createFormsTableV8(db);
+            createFormsTableV9(db);
```
lognaturel, line 131: Previously, all of the data in the forms table could be recreated from a form definition except for the last detected hash. Now delete adds more state that can't be recreated. Still, the consequences of "undeleting" forms on downgrade are not very dire and we don't expect downgrade to be common so let's stick with this and not worry about it. (This is just a note to make the decision explicit to our future selves)

collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
```diff
@@ -0,0 +1,18 @@
...
+public class ReentrantLockChangeLock implements ChangeLock {
+
+    private final ReentrantLock lock = new ReentrantLock();
+
+    @Override
+    public <T> T withLock(Function<Boolean, T> function) {
+        try {
+            return function.apply(lock.tryLock());
+        } finally {
+            lock.unlock();
```
lognaturel: I can't imagine how it would happen but technically this could throw a `IllegalMonitorStateException`. Maybe catch and ignore out of an abundance of caution?

lognaturel (changes requested):
I got pretty confused looking at `BooleanChangeLock` and I think it would be helpful to move that to a test package because it can't be used in real code.

lognaturel:
I'm marking this as `needs testing` because while there are a couple of things that really should be updated before it hits master (the activity alias, the error string text), they're not big behavioral changes. It'd be great to get a first pass of exploratory testing from @getodk/testers so they get some familiarity with the "match exactly" feature and check some of the areas at highest risk for regression (e.g. auto send, auto form update). The goal would be to catch anything really major before we put out a beta. I suggest starting by reading through and/or watching the Espresso tests in `feature/formmanagement`.

[This forum post](https://forum.getodk.org/t/have-collect-exactly-match-the-forms-on-central/23101/9?u=ln) describes the intended workflow. The thread also explains the reasoning behind the feature. The following issues provide acceptance criteria: #3934, #3939, #3940, #3959, #3992. I recommend looking at them in that order.

@seadowg please share your notes on regression risks.

seadowg:
@lognaturel @mmarciniak90 @kkrawczyk123 so for me the areas/features of the app that are at greatest risk are:

* Audio buttons on questions
* Audio widget
* Pushing forms to device from `adb`
* Form auto update (now called "Previously Downloaded Only")
* Auto submit
* Correct auto/background jobs (match exactly, previously downloaded etc) being scheduled (or not scheduled) when settings are updated or imported

There is test coverage for most of these areas but it would be great to put them through their paces as well.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
seadowg: Good catch

seadowg:
@lognaturel `BooleanChangeLock` is in `org.odk.collect.android.support` in the `test` source root. That's where all our other fakes and test support code live. I feel like that's the best place for it to live.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/DeleteFilledFormTest.java
seadowg: Again nice. Catch I think I was too focused on getting `...FormFrom...` right in these cases haha

collect_app/src/main/res/layout/main_menu.xml
seadowg: Oh whoops. Yup.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/ReentrantLockChangeLock.java
seadowg: Yeah I missed that in the docs. Thanks for pointing it out. Will add that in.

collect_app/src/main/java/org/odk/collect/android/tasks/DeleteFormsTask.java
seadowg: I was following the original structure which used setters to get the dependencies in. You're right though that we should improve on that while we're here!

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormsRepository.java
seadowg: Totally missed this was a thing. On one hand it feels like it might make sense to start reducing the number of methods on `FormsDao` and just write queries as much as possible in our new repos. I'll see how used the `getFormsCursor(String formId, String formVersion)` is and think about inlining it.

seadowg: Actually going to avoid that method as it adds sorting in which we don't need here.

collect_app/src/main/java/org/odk/collect/android/fragments/DataManagerList.java
seadowg: Agreed!

collect_app/src/main/AndroidManifest.xml
seadowg: As far as I can see where ok as the Activity isn't `exported`. I checked the docs to remind myself but the default is `false` ([docs here](https://developer.android.com/guide/topics/manifest/activity-element#exported)) so other apps shouldn't be able to launch this.

seadowg:
Think I've found a problem so removing from testing

seadowg:
Fixed

collect_app/src/main/AndroidManifest.xml
lognaturel: Ohhhh, yes. I convinced myself it was true by default but should just have looked it up.

lognaturel:
> BooleanChangeLock is in org.odk.collect.android.support in the test source root. 

That’s absolutely right. Looks like I didn’t notice that the path I was seeing in Github was truncated from the left. I’ve made that mistake before so need to watch it.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified Android versions: 5.1, 7.0, 7.1, 8.1, 9.0, 10.0

We spent a lot of time verifying different test cases with
* `Exactly match server` option
* `Previously downloaded forms only` option
* Aggregate and Central servers and Google Drive 
* Experimental settings section removed when the tag is added
* Soft delete form
* notifications

We checked areas suggested by Callum. We reported all noticed issues. 
This change affects many activities and we did our best to find issues and regressions but it's possible that we didn't discover all incorrect behavior.

PR #3994, opened 2020-07-29
Stop null pointer on progress update

### LINKED ISSUE #3989: Crash when cancelling a form download
`master`, this is a regression caused by the recent match exactly work

#### Problem description
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void org.odk.collect.android.formentry.RefreshFormListDialogFragment.setMessage(java.lang.String)' on a null object reference
        at org.odk.collect.android.activities.FormDownloadListActivity.progressUpdate(FormDownloadListActivity.java:689)
        at org.odk.collect.android.tasks.DownloadFormsTask.onProgressUpdate(DownloadFormsTask.java:79)
        at org.odk.collect.android.tasks.DownloadFormsTask.onProgressUpdate(DownloadFormsTask.java:35)
        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:656)
        at android.os.Looper.loop(Looper.java:145)
        at android.app.ActivityThread.main(ActivityThread.java:6939)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
```

1. Connect to the default server
1. Start downloading All Widgets
1. Cancel the download before all media files are downloaded

#### Expected behavior
Canceling should roll back the download and have no other effect. I do think it's valuable to show progress as the media file download is happening because it gives a sense of whether I'm likely to be able to complete the download. For example, if I'm stuck at attachment 2/10 for a really long time, I might want to cancel whereas if I'm at 9/10 maybe I'm willing to wait longer or put some effort to finding a stronger signal.

seadowg:
Closes #3989 

I wasn't able to reproduce but this change should protect us from the possibility of the null pointer.

I initially spiked out reworking this code the status/progress of the download could use `LiveData` and have a reworked dialog observe the status (rather than the activity pass the progress message from a listener to the dialog). This would hopefully give us safer code but it seemed like it was going to be a lot of work (mainly due to progress updates being embedded in the `MultiFormDownloader`). I'm keen to look at pushing this kind of rework into the 1.29 work as it'd be great to clean all this up and also get rid of `AsyncTask` as much as possible. I'll be back.

Should just fix the regression!

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
@@ -684,10 +684,12 @@ public class FormDownloadListActivity extends FormListActivity implements FormLi
 
     @Override
     public void progressUpdate(String currentFile, int progress, int total) {
-        RefreshFormListDialogFragment fragment = (RefreshFormListDialogFragment) getSupportFragmentManager()
-                .findFragmentByTag(RefreshFormListDialogFragment.class.getName());
-        fragment.setMessage(getString(R.string.fetching_file, currentFile,
-                String.valueOf(progress), String.valueOf(total)));
+        RefreshFormListDialogFragment fragment = (RefreshFormListDialogFragment) getSupportFragmentManager().findFragmentByTag(RefreshFormListDialogFragment.class.getName());
+
+        if (fragment != null) {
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Thanks, this seems like a pragmatic approach. Merging without QA since it's part of the match exactly work.

PR #3995, opened 2020-07-29
Use view model for Number Picker Dialog

### LINKED ISSUE #4030: Chosen answer is reset after screen rotation on Range picker widget
Collect v1.27.3, Android v all, device used...

#### Problem description
After screen rotation on Range picker widget the answer is reset also when is selected one more time.

Use case 1:
1. Open range picker
2. Select an answer
3. Rotate screen: Initial value is selected
4. Click OK: No value is selected

Use case 2:
1. Open range picker
2. Rotate screen
4. Select an answer
5. Click OK button: No value is selected

Use case 3:
1. Open range picker
2. Rotate screen
3. Rotate back to portrait
4. Select an answer
5. Click OK button: No value is selected

#### Expected behavior
Screen rotation should not reset the chosen answers. User should be able to save his answers. 

#### Other information 
- I used All widgets form. 
- I was able to reproduce it on Collect 1.25 and 1.22 so it is old issue not regression issue.

SaumiaSinghal:
Closes #4030. In continuation of #3969. Blocked by #3969.

I used `ViewModel` to observe any changes in the number picker value and update the answer shown in the answer text view in `RangePickerIntegerWidget` and `RangePickerDecimalWidget`.

This refactors the RangeWidgets. I renamed `RangeWidgetUtils` to `RangeWidgetDataRequester`. I used viewmodel to update range picker widget values in the FormEntryActivity.  This PR refactors RangePickerWidget and RangeWidgets.

Any form having `RangePickerIntegerWidget` and `RangePickerDecimalWidget`. 

[Range Picker Test .xml.txt](https://github.com/getodk/collect/files/4998012/Range.Picker.Test.xml.txt)

No


seadowg:
@SaumiaSinghal this is unblocked now! If you're able to fix the conflicts with a rebase and then mark as ready tor review when you're happy I can take a look at it.

seadowg:
@SaumiaSinghal looks like checks are failing

SaumiaSinghal:
Yeah @seadowg! I am working on the same. Will text you once everything's fixed.

SaumiaSinghal:
Hi @seadowg! It is ready for review. Sorry for replying late, was caught up in some refactors that I should have done in #3969

collect_app/src/test/java/org/odk/collect/android/widgets/RangePickerDecimalWidgetTest.java
```diff
@@ -57,71 +67,76 @@ public void whenRangeStartIsGreaterThanRangeEnd_widgetsSetsCorrectValuesForNumbe
...
         RangePickerDecimalWidget widget = createWidget(promptWithQuestionDefAndAnswer(rangeQuestion, null));
-        widget.setOnLongClickListener(listener);
-        widget.binding.widgetButton.performLongClick();
-        widget.binding.widgetAnswerText.performLongClick();
+        WidgetValueChangedListener valueChangedListener = mockValueChangedListener(widget);
+        rangePickerViewModel.setNumberPickerValue(5);
 
-        verify(listener).onLongClick(widget.binding.widgetButton);
-        verify(listener).onLongClick(widget.binding.widgetAnswerText);
+        rangePickerViewModel.getNumberPickerValue().observe(fakeLifecycleOwner, integer -> {
```
seadowg: There's no reason for this to do an `observe`. You should just be able to call `verify(valueChangedListener).widgetValueChanged(widget);` after setting the value.

collect_app/src/test/java/org/odk/collect/android/widgets/RangePickerIntegerWidgetTest.java
```diff
@@ -61,72 +68,80 @@ public void whenRangeStartIsGreaterThanRangeEnd_widgetsSetsCorrectValuesForNumbe
...
-        when(rangeQuestion.getRangeEnd()).thenReturn(BigDecimal.ONE);
-
+    public void updatingValueInRangePickerViewModel_updatesAnswer() {
         RangePickerIntegerWidget widget = createWidget(promptWithQuestionDefAndAnswer(rangeQuestion, null));
-        widget.setNumberPickerValue(4);
+        rangePickerViewModel.setNumberPickerValue(5);
 
-        assertThat(widget.getAnswer().getDisplayText(), equalTo("5"));
+        rangePickerViewModel.getNumberPickerValue().observe(fakeLifecycleOwner, integer -> {
+            assertEquals(widget.binding.widgetAnswerText.getText(), "5");
```
seadowg: Same here. No reason for this to happen in `observe`.

collect_app/src/test/java/org/odk/collect/android/widgets/RangePickerIntegerWidgetTest.java
```diff
@@ -61,72 +68,80 @@ public void whenRangeStartIsGreaterThanRangeEnd_widgetsSetsCorrectValuesForNumbe
...
         RangePickerIntegerWidget widget = createWidget(promptWithQuestionDefAndAnswer(rangeQuestion, null));
-        widget.setOnLongClickListener(listener);
-        widget.binding.widgetButton.performLongClick();
-        widget.binding.widgetAnswerText.performLongClick();
+        WidgetValueChangedListener valueChangedListener = mockValueChangedListener(widget);
+        rangePickerViewModel.setNumberPickerValue(5);
 
-        verify(listener).onLongClick(widget.binding.widgetButton);
-        verify(listener).onLongClick(widget.binding.widgetAnswerText);
+        rangePickerViewModel.getNumberPickerValue().observe(fakeLifecycleOwner, integer -> {
```
seadowg: And here.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/NumberPickerDialog.java
```diff
@@ -19,76 +19,68 @@
...
         super.onAttach(context);
-
         try {
             listener = (NumberPickerListener) context;
         } catch (ClassCastException e) {
             throw new ClassCastException(context.toString() + " must implement OnHeadlineSelectedListener");
         }
+
+        rangePickerViewModel = new ViewModelProvider(this).get(RangePickerViewModel.class);
+        rangePickerViewModel.setDisplayedValuesForNumberPicker((String[]) getArguments().getSerializable(DISPLAYED_VALUES));
```
seadowg, line 51: Instead of passing this as an argument to the dialog and then to the view model, couldn't the data requester set the value on the view model?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/NumberPickerDialog.java
```diff
@@ -19,76 +19,68 @@
...
+        rangePickerViewModel.setProgress(getArguments().getInt(PROGRESS));
+
+        rangePickerViewModel.getNumberPickerValue().observe(this, answer -> {
```
seadowg, line 54: Why does this value get observed to then just get passed to a listener? Couldn't wherever the listener is trying to get the value (the widget I think) listen? Maybe the data requester should have a method that lets the widget listen for changes and then the data requester can handle the live data subscription (look at `AudioPlayer#onPositionChanged` and its implementation as an example).

seadowg: As another thought that would potentially mean you wouldn't need the `WaitingForDataRegistry` for any of this (as the FormEntryActivity would not have to be involved)

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetDataRequester.java
```diff
@@ -0,0 +1,186 @@
...
+        public TextView getMinValue() {
+            return minValue;
+        }
+
+        public TextView getMaxValue() {
+            return maxValue;
+        }
+    }
+
+    public static RangeWidgetLayoutElements getLayoutElements(Context context, String appearance) {
```
seadowg: All these static helpers feel like they don't fit here? I think they should just move back to a `...Utils` class.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetDataRequester.java
```diff
@@ -0,0 +1,186 @@
...
+            for (BigDecimal i = rangeEnd; i.compareTo(rangeStart) < 1; i = i.add(rangeStep.abs())) {
+                displayedValuesForNumberPicker[index] = isIntegerDataType ? String.valueOf(i.intValue())
+                        : String.valueOf(i.doubleValue());
+                index++;
+            }
+        }
+        return displayedValuesForNumberPicker;
+    }
+
+    public static void requestRangePickerValue(Context context, WaitingForDataRegistry waitingForDataRegistry, FormIndex formIndex,
```
seadowg: I think this has all been a little mixed up. To me, it feels like the `RangeWidgetDataRequester` should be an interface with the `requestRangePickerValue` method on it and there should be an implementation that uses the ViewModel/DialogFragment. In my head this should work like the `GeoDataRequester`. Does that make sense?

Happy to chat about this more on Slack or somewhere if that's helpful.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/NumberPickerDialog.java
SaumiaSinghal: Yes correct. I was just confused that if I do so I will be initializing the viewmodel in `RangeWidgetDataRequester` class and then I will have to use the activity's (`FormEntryActivity`) as the owner of the viewmodel. Currently, the viewmodel lives until the dialog is shown and if the user opens up the `NumberPicker` again, a new viewmodel is instantiated and we don't have to worry about the viewmodel which was atttached to the `FormEntryActivity` previously.

I think if I use the activity as the owner, then I will be having the viewmodel as long as I am using the widgets of a particular form?

seadowg: Right I see. I think it'd be nice to get to a point where the ViewModel is the sole method of communication between the widget and the dialog but I it would also involve making the ViewModel work for multiple range widgets (like in a field list). For the moment then lets just go with the args being passed to the dialog and then then the dialog calling the FormEntryActivity via a listener.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RangeWidgetDataRequester.java
seadowg: Based on our discussion in the other comment we  don't want the data requester to use the viewmodel but it should still be reworked as an interface (and have the static methods split out). Also the `WaitingForDataRegistry` and `Content` could be passed in at the constructor level rather than in this method.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/NumberPickerDialog.java
SaumiaSinghal: Oh okay yes I like he idea of having the viewmodel just the sole method of communication

SaumiaSinghal: While I was working refactoring the Date and Time widgets, I came across the same problem when using the viewmodel. @grzesiek2010 guided and helped me in setting them up as I have done in this PR. However, he also liked the idea of excluding the activity from the conversation between viewmodel and widget. He told that later we will be using fragments to implement widgets, then it will be easy to use viewmodel without involving the activity.

seadowg:
@SaumiaSinghal it looks like something might be funky with your git setup. You're potentially using two different emails/names for committer and author? You'll see Github is crediting your account twice. Maybe a good idea to try and fix and then rebase/reset author (https://stackoverflow.com/questions/750172/how-to-change-the-author-and-committer-name-and-e-mail-of-multiple-commits-in-gi).

seadowg (changes requested):
As above

SaumiaSinghal:
Thanks for the link @seadowg. It is fixed now.

seadowg (commented):
There are a few things wrong here that I'm confused about:

* The PR doesn't solve the problem as far as I can see. Opening the dialog, moving to a value and then rotating still doesn't work. I might be confused about the original issue though. As far as I can see `RangePickerViewModel` doesn't do anything to address that. In fact...
* I'm not sure what `RangePickerViewModel` is for. All the state that it holds is either passed to it in `onAttach` from args or it just uses `LiveData` to a fire a listener. While it is nice to use a ViewModel I think here as only one object deals with that state or the listener it feels like overkill and makes it harder to follow.

@kkrawczyk123 could you possibly update #4030 with a GIF so it's clear exactly what the problem is?

I'm going to revert this PR back to a draft as I think it might not be solving the issue and I'm also unclear on whether we want to fix that issue.

SaumiaSinghal:
This PR does not solve the issue you mentioned, I did not touch that code with the viewmodel. But I think previously, when I tested this, the problem was that on rotating the screen, the answer was getting reset for the Range picker widget. I think this was solved in the previous PR #3969 itself

seadowg:
Ah ok! So I'll close this for the moment and we can wait for an update/discussion in the issue.

PR #3997, opened 2020-07-30
Make sure Automatic Download setting makes sense for Form Update mode

### LINKED ISSUE #3982: Automatic download checkbox should be disabled for Never periodic form update check
Collect v1.28.- current master, Android all device used...

#### Problem description
Automatic download checkbox is now enabled for default choice: Never in Periodic form updates check, see screenshots:
| master v1.28.0 | store version 1.27.3 |
| --------- | ---------|
| Screenshot (Jul 23, 2020 9_02_13 AM) | Screenshot_20200723-090551 |

1. Go to General Settings/ Form Management
2. Click on Automatic download checkbox without changing Periodic form updates check

#### Expected behavior
User should not be able to select automatic download checkbox when periodic form updates check is set to Never.

#### Other information 
It works well on the store version: 1.27.3, so it is a recently introduced regression issue.

seadowg:
Closes #3982

Played around with it manually and added new tests

I also considered trying to make the setting just display correctly rather than changing it when the form update changes. This seemed like it was going to end up being pretty complex though, so I just went with the path of least resistance.

Should just fix the bug! Good to double check how the settings change when the form update mode changes makes sense to everyone else.

This shouldn't skip QA like the other Match Exactly PRs.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ServerSettingsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ServerSettingsTest.java
@@ -1,61 +1,46 @@
 package org.odk.collect.android.feature.settings;
 
-import android.Manifest;
-import android.webkit.MimeTypeMap;
-
+import androidx.test.espresso.intent.rule.IntentsTestRule;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
-import androidx.test.rule.GrantPermissionRule;
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.junit.runner.RunWith;
 import org.odk.collect.android.R;
-import org.odk.collect.android.injection.config.AppDependencyModule;
-import org.odk.collect.android.openrosa.OpenRosaHttpInterface;
-import org.odk.collect.android.support.CollectTestRule;
-import org.odk.collect.android.support.ResetStateRule;
-import org.odk.collect.android.support.StubOpenRosaServer;
+import org.odk.collect.android.activities.MainMenuActivity;
+import org.odk.collect.android.support.TestDependencies;
+import org.odk.collect.android.support.TestRuleChain;
 import org.odk.collect.android.support.pages.GeneralSettingsPage;
 import org.odk.collect.android.support.pages.MainMenuPage;
-import org.odk.collect.utilities.UserAgentProvider;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 
 @RunWith(AndroidJUnit4.class)
 public class ServerSettingsTest {
 
-    public final StubOpenRosaServer server = new StubOpenRosaServer();
+    private final TestDependencies testDependencies = new TestDependencies();
 
-    public CollectTestRule rule = new CollectTestRule();
+    public IntentsTestRule<MainMenuActivity> rule = new IntentsTestRule<>(MainMenuActivity.class);
 
     @Rule
-    public RuleChain copyFormChain = RuleChain
-            .outerRule(GrantPermissionRule.grant(
-                    Manifest.permission.READ_EXTERNAL_STORAGE,
-                    Manifest.permission.WRITE_EXTERNAL_STORAGE,
-                    Manifest.permission.READ_PHONE_STATE,
-                    Manifest.permission.GET_ACCOUNTS
-            ))
-            .around(new ResetStateRule(new AppDependencyModule() {
-                @Override
-                public OpenRosaHttpInterface provideHttpInterface(MimeTypeMap mimeTypeMap, UserAgentProvider userAgentProvider) {
-                    return server;
-                }
-            }))
+    public RuleChain copyFormChain = TestRuleChain.chain(testDependencies)
             .around(rule);
 
     @Test
     public void whenUsingODKServer_canAddCredentialsForServer() {
-        server.setCredentials("Joe", "netsky");
-        server.addForm("One Question", "one-question", "one-question.xml");
+        testDependencies.server.setCredentials("Joe", "netsky");
+        testDependencies.server.addForm("One Question", "one-question", "one-question.xml");
 
-        rule.mainMenu()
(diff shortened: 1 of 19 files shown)


lognaturel (approved):
Maybe add a test for scheduled updates being canceled if the protocol changes to Google Drive?

seadowg:
@lognaturel good call

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ServerSettingsTest.java
```diff
@@ -90,4 +75,14 @@ public void selectingGoogleAccount_showsGoogleAccountSettings() {
...
                 .assertText(R.string.google_sheets_url);
     }
+
+    @Test
+    public void selectingGoogleAccount_disablesAutomaticUpdates() {
+        MainMenuPage mainMenu = new MainMenuPage(rule).assertOnPage()
+                .enablePreviouslyDownloadedOnlyUpdates();
+        assertThat(testDependencies.scheduler.getDeferredTasks().size(), is(1));
+
+        mainMenu.setGoogleDriveAccount("steph@curry.basket");
```
lognaturel, line 85: How very American.

lognaturel (approved):
I like the changes that came out of adding that test. 👍

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ServerSettingsTest.java
seadowg: Heh. I'm enjoying the NBA restart but missing my team.

seadowg:
I realized I've introduced a weird problem here where updates will be scheduled when the user changes any setting. Will revise and then request a re-review.

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.java
```diff
@@ -62,7 +62,13 @@ public boolean fromJSON(@NonNull String json) {
         loadDefaults(generalSharedPrefs, generalDefaults);
         loadDefaults(adminSharedPrefs, adminDefaults);
 
-        settingsChangedHandler.onSettingsChanged();
+        for (String key: generalSharedPrefs.getAll().keySet()) {
```
seadowg, line 65: Iterating through *ALL* the keys is future proofing really but it felt weird for `SettingsImporter` to just pass one through knowing it would do the right thing.

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
```diff
@@ -77,40 +84,59 @@ public void onCreate(Bundle savedInstanceState) {
         setupFormUpdateMode();
     }
 
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+        super.onSharedPreferenceChanged(sharedPreferences, key);
+
+        if (key.equals(KEY_FORM_UPDATE_MODE) || key.equals(KEY_PERIODIC_FORM_UPDATES_CHECK)) {
```
lognaturel, line 91: This is where I was getting confused when we were talking about the most recent changes. I think the split you're making is things that need to happen any time that a preference is updated even if the user is not anywhere near preferences in the app go to the custom handler which is always called by the base preference fragment. On the other hand, changes that only need to happen if the user is sitting in preferences at the moment are locally handled. Typically it would just be inline like with guidance hints but in this case since there are two keys that do the same thing, that ends up in `onSharedPreferenceChanged`. Does that sound right? I probably would have introduced a method to wrap the work that happens when either `KEY_FORM_UPDATE_MODE` or `KEY_PERIODIC_FORM_UPDATES_CHECK` changes and also defined handlers for those inline to avoid overriding `onSharedPreferenceChanged` but this is ok too.

Now that this structure is in place, I think there are things we should move from inline handlers to `CollectSettingsChangeHandler` like sending an analytics event for a server URL change, for example. Something to file an issue for v1.29 about?

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.java
lognaturel: One weirdness about this is that the behavior will be different depending on how the form update settings are ordered relative to the server URL and credential settings in the key set. Perhaps this can be considered at the same time as #4011.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
seadowg: Yeah I think we could use this for a lot more. Analytics is a great shout as it would let us capture more from the imports. I'd like to take a proper pass over some of these preference screens again as I was avoiding too many changes due to the AndroidX stuff being in flight. I'll add a note to my project backlog.

As to the split feeling a little weird: I agree. It might be that local changes like that are just better handles using the individual preference listeners.

mmarciniak90:
I tried to look at this PR more comprehensively because the reported issue is no longer valid.

I noticed that:
:heavy_check_mark: Automatic download is unselected as default 
:heavy_check_mark: Automatic download is unselected as default and inactive for `Manually` mode
:heavy_check_mark: Automatic download is selected as default and inactive for `Exactly match server` mode
:heavy_check_mark: Automatic download is unselected as default and active for `Previously downloaded forms only` mode
:heavy_check_mark: First run -> user changed server to Google Drive -> `Manually` mode is set and Automatic download is unselected; all options are inactive

:heavy_check_mark: Another mode than `Manually` was selected and `Automatic download` was selected -> then user set `Manually` mode -> Automatic download is **unselected**
:warning: Aggregate server was set -> Another mode than `Manually` was selected and `Automatic download` was selected ->  then user changed server to Google Drive -> **This time, Automatic download is selected but not editable**

@seadowg, do you think that you can improve with this one marked case?

seadowg:
@mmarciniak90 yeah that can be improved. Could you add it to #4007? I was looking at some related stuff there so would be easier to fix as part of that rather than coming back for it here if that's ok?

kkrawczyk123:
**Tested with success!**

PR #4000, opened 2020-07-31
Revise notification behaviour for Match Exactly

### LINKED ISSUE #3999: Fill Blank Form can appear multiple times on back stack and auth dialog should be displayed immediately
`master`

#### Problem description
These are two related issues connected to @seadowg's match exactly work.

First, if the user receives a notification related to "match exactly" work happening and taps it, that always adds a new instance of Fill Blank Form on the back stack. However, Fill Blank Form may already be on the back stack. This is particularly noticeable when the user manually triggers a form update because s/he is already in Fill Blank Form and tapping a notification always creates a new instance of it.

Second, if the server requires auth and the user manually triggers a form update from Fill Blank Form, the user will receive a notification and have to tap that notification to get the auth prompt. Since this case is user-recoverable, it would be better if the dialog could immediately be shown to the user rather than s/he having to tap through a notification.

These are likely related because it may be easier to solve the second issue first and not have to deal with the case where Fill Blank Form is displayed without the auth dialog, a update attempt fails with a `401`, and the user needs to tap the notification to get that dialog.

1. Turn on match exactly
1. Configure a server that requires auth and don't provide credentials
1. Manually trigger a server update from Fill Blank Form
1. Tap on the notification to get the auth dialog (and note that it would be better if the dialog just appeared)
1. Dismiss the dialog
1. Tap back and notice that you're still in Fill Blank Form

#### Expected behavior
Auth dialog appears in the foreground if match exactly fails with 401 and Fill Blank Form is displayed. Tapping on a match exactly notification uses the existing instance of Fill Blank Form if it exists. There is never more than one instance of Fill Blank Form.

seadowg:
Closes #3999. ~~Blocked by #3992.~~

In addition to the problems described in the issue I also made it so error notifications are cleared/cancelled when a sync succeeds.

New/revised tests and was able to verify manually

I've left some comments inline.

Good to just check the issue is fixed and that Match Exactly notifications behave as expected.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/MatchExactlyTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/MatchExactlyTest.java
@@ -52,13 +52,33 @@ public class MatchExactlyTest {
     }
 
     @Test
+    public void whenMatchExactlyEnabled_clickingFillBlankForm_andClickingRefresh_whenThereIsAnError_showsNotification_andClickingNotification_returnsToFillBlankForms() throws Exception {
+        testDependencies.server.alwaysReturnError();
+
+        rule.mainMenu()
+                .setServer(testDependencies.server.getURL())
+                .enableMatchExactly()
+                .clickFillBlankForm()
+                .clickRefresh();
+
+        notificationDrawerRule
+                .open()
+                .clickNotification(
+                        "ODK Collect",
+                        "Form update failed",
+                        "If you keep having this problem, report it to the person who asked you to collect data.",
+                        "Fill Blank Form",
+                        new FillBlankFormPage(rule)
+                ).pressBack(new MainMenuPage(rule)); // Check we return to Fill Blank Form, not open a new one
+    }
+
+    @Test
     public void whenMatchExactlyEnabled_clickingFillBlankForm_andClickingRefresh_whenThereIsAnAuthenticationError_promptsForCredentials() throws Exception {
-        String url = testDependencies.server.getURL();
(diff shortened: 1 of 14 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/MatchExactlyTest.java
```diff
@@ -53,40 +52,14 @@ public void whenMatchExactlyEnabled_clickingFillBlankForm_andClickingRefresh_get
     }
 
     @Test
-    public void whenMatchExactlyEnabled_getsLatestFormsFromServer_automaticallyAndRepeatedly() throws Exception {
```
seadowg, line 56: I decided to revise how the tests are written to drive out the new behaviour rather than just adding more. Along with the changed/new unit tests everything was TDD'd out and I'm pretty sure we didn't lose anything.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -113,7 +113,7 @@ the specific language governing permissions and limitations under the License.
             android:configChanges="orientation|screenSize"
             android:windowSoftInputMode="stateHidden" />
         <activity android:name=".activities.InstanceChooserList" />
-        <activity android:name=".activities.FillBlankFormActivity" />
+        <activity android:name=".activities.FillBlankFormActivity" android:launchMode="singleTop"/>
```
seadowg, line 116: Not a fan of changes here but at least it's driven out by a test in Espresso

lognaturel (approved):

PR #4002, opened 2020-07-31
Match exactly analytics

### LINKED ISSUE #3958: Add analytics for Match Exactly
~~Blocked by #3959 / #3940.~~

* Send event with server hash when clicking refresh button
* Set user property for selected "form update" mode ("manually", "match_exactly" and "previously_downloaded_only")
* Send event when a sync completes with failure or success (like storage migration). Include error type with failures.

seadowg:
Closes #3958

New tests.

Comments inline

Just adds analytics so should be good to merge as long as we try it out properly with the Play Store creds in the app.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
@@ -16,11 +16,14 @@ import org.javarosa.core.util.JavaRosaCoreModule;
 import org.javarosa.model.xform.XFormsModule;
 import org.javarosa.xform.parse.XFormParser;
 import org.odk.collect.android.BuildConfig;
+import org.odk.collect.android.analytics.Analytics;
 import org.odk.collect.android.application.Collect;
+import org.odk.collect.android.formmanagement.FormUpdateMode;
 import org.odk.collect.android.geo.MapboxUtils;
 import org.odk.collect.android.logic.PropertyManager;
 import org.odk.collect.android.logic.actions.setgeopoint.CollectSetGeopointActionHandler;
 import org.odk.collect.android.preferences.AdminSharedPreferences;
+import org.odk.collect.android.preferences.GeneralKeys;
 import org.odk.collect.android.preferences.GeneralSharedPreferences;
 import org.odk.collect.android.utilities.LocaleHelper;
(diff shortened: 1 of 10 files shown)


collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -55,21 +60,30 @@ public void initialize() {
...
+        initializeAnalytics();
     }
 
-    public void initializeLocale() {
+    private void initializeAnalytics() {
+        String formUpdateMode = generalSharedPreferences
+                .getSharedPreferences()
+                .getString(GeneralKeys.KEY_FORM_UPDATE_MODE, FormUpdateMode.MANUAL.getValue(context));
+
+        analytics.setUserProperty("FormUpdateMode", formUpdateMode);
```
seadowg, line 83: I wanted to do this here rather than when the settings are changed/imported as it avoids handling all the different possible cases and also makes sure we always have a value for this property set. It means the values we have are probably going to be slightly behind (as we'll be waiting for the next time the app boots for the property to be set) but I think that'll be ok.

collect_app/src/main/java/org/odk/collect/android/formmanagement/BlankFormsListViewModel.java
```diff
@@ -87,6 +99,12 @@ public void syncWithServer() {
         });
     }
 
+    private void logManualSync() {
+        String url = preferencesProvider.getGeneralSharedPreferences().getString(GeneralKeys.KEY_SERVER_URL, "");
+        String urlHash = FileUtils.getMd5Hash(new ByteArrayInputStream(url.getBytes()));
```
lognaturel: What do you think of using the hostname instead of the raw setting value? The raw server setting value would mean we differentiate between different Central App Users. Hostname would give us a sense of organizations that use match exactly and specifically the manual sync button.

seadowg: Yeah that makes more sense

lognaturel (approved):

lognaturel:
@getodk/testers are currently doing their work off master so I think it's best to get the green light from them before merging. This doesn't affect behavior but it does move code and tests around.

lognaturel:
Merging because it will be good to have for the beta.

PR #4006, opened 2020-08-04
Reworking Audio widget

### LINKED ISSUE #4005: Audio is hearable after removing response on Audio widget

1. User adds audio to Audio widget
2. User plays it
3. User removes response when audio is played
4. **The response is removed but the sound is still hearable**

#### Expected behavior
* Audio is not hearable after removing response

#### Other information 
Noticed while testing #3992

### LINKED ISSUE #4008: The blue line does not show audio progress on Audio widget
Collect master

#### Problem description
The blue line does not show audio progress

1. User adds audio to Audio widget
2. User plays it -> the blue line shows audio progress
3. User goes to the next question and returns to the audio widget
4. User plays audio again -> the blue line does not show audio progress

#### Expected behavior
The blue line always shows audio progress

#### Other information 
Noticed while testing #3992

SaumiaSinghal:
Closes #4005 and Closes #4008 
This PR includes rewriting the test coverage and class of Audio Widget

This PR removes `MediaManager` enum class and the functionality is shifted to `FormSaveViewModel`. Hence, it changes the structure of `FileWidgets` and scenarios involving rotation of activities and killing of activities in the middle of saving, replacing or deleting answers in file widgets would be good to test.

Any form having Audio Widget like AllWidgetsForm

No


seadowg:
@SaumiaSinghal do you think it'd be plausible to look at fixing #4005 as part of this? If it seems like a lot of extra work then no worries but thought it'd be good to look at while you're in the "neighbourhood".

SaumiaSinghal:
yes sure @seadowg I will look at it

seadowg:
Ah I'm so sorry @SaumiaSinghal I meant #4008. There were two issues but only one was a regression. I see you've fixed #4005 now which is good but if you can look at #4008 it is higher priority.

collect_app/src/test/java/org/odk/collect/android/widgets/AudioWidgetTest.java
```diff
@@ -1,117 +1,188 @@
...
+        assertThat(widget.getAnswer(), nullValue());
     }
 
-    @Override
-    public void setUp() throws Exception {
-        super.setUp();
-        destinationName = RandomString.make();
+    @Test
+    public void getAnswer_whenPromptHasAnswer_returnsAnswer() {
+        AudioWidget widget = createWidget(promptWithAnswer(new StringData(destinationPath)));
```
seadowg: I think better to just have the values inline (`destinationPath`) rather than defined anywhere. Easier to see in the test that the value is just a string.

collect_app/src/main/java/org/odk/collect/android/widgets/interfaces/MediaManagerListener.java
```diff
@@ -0,0 +1,7 @@
+package org.odk.collect.android.widgets.interfaces;
+
+public interface MediaManagerListener {
```
seadowg: Not sure the naming is right here. It's not really a "listener". Maybe the interface is `QuestionMediaManager` and the implementation is something else?

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
```diff
@@ -101,24 +98,33 @@ public AudioWidget(Context context, QuestionDetails prompt, WaitingForDataRegist
...
+        binding = AudioWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
+
+        if (prompt.isReadOnly()) {
+            binding.captureButton.widgetButton.setVisibility(View.GONE);
+            binding.chooseButton.widgetButton.setVisibility(View.GONE);
+        } else {
+            binding.captureButton.widgetButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+            binding.chooseButton.widgetButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+
+            binding.captureButton.widgetButton.setText(getContext().getString(R.string.capture_audio));
```
seadowg: Feels like the button text should just live in a layout?

collect_app/src/main/java/org/odk/collect/android/audio/AudioPlayerViewModel.java
```diff
@@ -178,16 +178,14 @@ public void errorDisplayed() {
     }
 
     private void schedulePositionUpdates() {
-        if (positionUpdatesCancellable == null) {
-            positionUpdatesCancellable = scheduler.repeat(() -> {
-                CurrentlyPlaying currentlyPlaying = this.currentlyPlaying.getValue();
+        positionUpdatesCancellable = scheduler.repeat(() -> {
```
seadowg, line 181: We want a test for this change.

seadowg (changes requested):
Just done a first pass but this looks ace 🙌. Liking the use of fakes and pulling out interfaces to deal with the file stuff.

Just a few comments/tweaks inline.

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
SaumiaSinghal: I use 
`
    <include
        android:id="@+id/capture_button"
        layout="@layout/widget_button"/>

    <include
        android:id="@+id/choose_button"
        layout="@layout/widget_button"/>`

in audio_widget.xml file. So I have added the button text here.

collect_app/src/main/java/org/odk/collect/android/widgets/interfaces/MediaManagerListener.java
SaumiaSinghal: Yes right the implementation mostly concerns with the deletion or replacing the file at the current index. But I think I might need to extend this implementation for working on other widgets

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
seadowg: I think I'd be happier duplicating the layout here than having hardcoded text set programmatically. My thinking is that it's good to be able to see the layout as it will appear (with the text) in the design preview for it.

SaumiaSinghal: For the other file widgets, I was using the same `widget_button.xml` layout. So, I should have a separate layout for all of them too?

seadowg: I think instead of sharing an `include` maybe we could create a custom button class with the shared values. Or, create a style that can be shared. These would let us avoid duplication but avoid having to set the text programmatically (which I'd like to avoid as much as possible).

collect_app/src/main/java/org/odk/collect/android/audio/AudioPlayerViewModel.java
SaumiaSinghal: I verified `playMultipleClips_updatesProgress_forAllClips()` and `playInOrder_playsClipsOneAfterTheOther_andUpdatesProgress` fails on master.

collect_app/src/test/java/org/odk/collect/android/support/FakeScheduler.java
```diff
@@ -13,6 +13,7 @@
     private Runnable foregroundTask;
     private Runnable backgroundTask;
     private Boolean cancelled = false;
+    private Boolean isProgressUpdating = false;
```
seadowg: I like what's happening here but I think the variable/method should be named something more related to the scheduler than the test here. Maybe `isRepeatRunning`?

collect_app/src/main/res/values/buttons.xml
```diff
@@ -6,6 +6,15 @@
         <item name="android:textStyle">normal</item>
     </style>
 
+    <style name="Widget.Collect.Button.Layout" parent="Widget.Collect.Button.Custom">
```
seadowg: The naming scheme here is pretty confusing as by convention we use `Widget` at the beginning but that's referring to the general definition of the word rather than a "question widget". Maybe the name here could be `Widget.Collect.Button.WidgetAnswer`?

seadowg (changes requested):

collect_app/src/main/res/values/buttons.xml
SaumiaSinghal: Oh I thought the same initially.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaManager.java
```diff
@@ -24,7 +24,7 @@
  * remember an original media file answer (no matter how many times this answer is replaced), in order
  * to be able to restore the original answer in case of ignoring changes.
  */
-public enum MediaManager {
+public enum MediaManager implements QuestionMediaManager {
```
seadowg: I don't think this is quite right after a second look. As much as we can we want to avoid static singletons. If there is some state that we want to keep shared at an application level (i.e. between multiple activities or services) then we should use Dagger to setup and inject a singleton (using `@Singleton`).  

In this case I don't believe the state needs to persist outside of form entry so it really only needs to last as long as the Activity.  This means we could confuse ourselves by ending up with state hanging around between form entry "sessions". A good way to share state like this would be to use a view model where the `ViewModelProvider` uses the activity as its owner. Does that make sense as a way forward?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/utilities/MediaManager.java
seadowg: If so, it might be good to discuss the plans/approach here so we can come up with something that makes sense as it's a little awkward I think!

SaumiaSinghal: @seadowg If I use the `ViewModel` then I would be creating it using the widget context, and it will be there for all the widgets on the screen?
I feel like avoiding using `QuestionMediaManager` too and using the previous implementation. That would mean there would be additional unit tests for the widgets. Will that be fine?

seadowg: I think if that's simpler to do right now that's a good plan. Maybe we can move the logic to a viewmodel in the future.

seadowg (changes requested):
@SaumiaSinghal and I talked about this and agreed it made sense to experiment with using a view model to handle files.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -370,7 +377,7 @@ public Factory(Analytics analytics) {
         @NonNull
         @Override
         public <T extends ViewModel> T create(@NonNull Class<T> modelClass) {
-            return (T) new FormSaveViewModel(System::currentTimeMillis, new DiskFormSaver(), analytics);
+            return (T) new FormSaveViewModel(System::currentTimeMillis, new DiskFormSaver(), analytics, MediaManager.INSTANCE);
```
seadowg: Maybe this hasn't come through in previous comments and conversations but I'm very against the `MediaManager` being a singleton. Just to make it very clear I think the `MediaManager` should be a `class` (not an `enum`) and that this line should be:

```java
return (T) new FormSaveViewModel(System::currentTimeMillis, new DiskFormSaver(), analytics, new MediaManager());
```

As we talked about before we could get in trouble with the `enum` singleton implementation as its state will be carried between different instances of the `FormEntryActivity`. Also in tests the singleton **would not** be recreated for each test (as the static state is retained throughout the process) so any state in the `MediaManager` set up in one test might inadvertently affect another.

collect_app/src/test/java/org/odk/collect/android/widgets/AudioWidgetTest.java
```diff
@@ -98,6 +102,12 @@ public void setUp() {
         when(activityAvailability.isActivityAvailable(any())).thenReturn(true);
     }
 
+    @After
+    public void tearDown() {
+        widgetActivity.originalFiles.clear();
```
seadowg: You shouldn't need to do this as each test will get a new instance of `TestWidgetActivity`.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2674,6 +2675,16 @@ public void onCancelFormLoading() {
         finish();
     }
 
+    @Override
+    public void deleteOriginalFile(String questionIndex, String fileName) {
```
seadowg: I think we're getting close with the structure but I don't like that the `FormEntryActivity` implements the `QuestionMediaManager`. I think instead:

* the `AudioWidget` should take a `QuestionMediaManager` in as a dependency at it's constructor rather than casting the `context` to one.
* the `WidgetFactory` should also take a `QuestionMediaManager` in as one of the args to `createWidgetFromPrompt` so it can pass it to any widget that needs it. 
* `FormSaveViewModel#getMediaManager` should return a `QuestionMediaManager` not a `MediaManager`
* `FormSaveViewModel#getMediaManager` should be passed as the `QuestionMediaManager` argument to `WidgetFactory#createWidgetFromPrompt`
* `QuitFormDialogFragment`  should call `viewModel.getMediaManager().revertChanges` rather than relying on the static singleton instance

I'm not sure if I like that currently the `QuestionMediaManager` is simply owned by the `FromSaveViewModel` and then exposes it. I'm thinking that potentially the `FormSaveViewModel` could implement the `QuestionMediaManager` interface and then there would be no need for the `MediaManager` class. In that world the "revert" call (in `QuiteFormDialogFragment` could just be part of the logic `FormSaveViewModel#removeTempInstance`.

seadowg (changes requested):
Getting closer and closer. This is a really hard one to strucutre...

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
SaumiaSinghal: Oh right! I didn't think about it. It will cause problems when writing tests for other file widgets.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
SaumiaSinghal: Nice idea @seadowg, and Thanks for explaining so well. I implemented all the functionality of `MediaManager` in `FormSaveViewModel`, as you said if we have the `FormSaveViewModel` implements `QuestionMediaManager`, then `MediaManager` class is no longer required.

seadowg (changes requested):
@SaumiaSinghal looks like unit tests are failing

SaumiaSinghal:
Sorry for that @seadowg

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -126,6 +127,7 @@ public void removeTempInstance() {
                 FileUtils.purgeMediaPath(instanceFolder);
             }
         }
+        releaseMediaManager();
```
seadowg: Do we need to do this in `onCleared()` as well? Have you played around with the behaviour when "Don't save activities" is enabled? It might be good to give that a go and see if we need to handle the case when we leave the app, Android kills the process, and then we return to the app. Would files get left undeleted but on in the form in this case? Would things break?

Let me know if you have questions about trying that out!

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -171,7 +173,7 @@ public String getReason() {
     }
 
     private void saveToDisk(SaveRequest saveRequest) {
-        saveTask = new SaveTask(saveRequest, formSaver, formController, mediaManager, new SaveTask.Listener() {
+        saveTask = new SaveTask(saveRequest, formSaver, formController, this, new SaveTask.Listener() {
```
seadowg: Hmmm. Little worried about the viewmodel being passed into an async task here... It's another awkward change but I think really what should happen here is that a list of files to save should get passed into the save task. Does that make sense? In that world we wouldn't need `saveChanges` (as it would be handled in `DiskFormSaver`.

seadowg (changes requested):
Like where this is going! It feels like this is working towards a better pattern for how we deal with files, external apps etc with widgets (basically listener interfaces with android arch components handling the actual implementations).

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
SaumiaSinghal: This is an interesting case. I will try that out with all the media widgets.

SaumiaSinghal: I tried clearing the `originalFiles` and `recentFiles` HashMap in `onCleared()`, when `Don't keep activities` is turned on. I didn't find any anomaly in saving answers to the widgets, as I think, the answer is already saved with the widget, so clearing the files won't cause any break. Am I missing some test case here?

seadowg: Here's the problem I found:

1. Enabled "Don't keep activities"
1. Open the "All widgets form"
1. Go to the file widget
1. Select a file
1. Select a different file
1. Go to the hierarchy view with the arrow button
1. Press "Go To End"
1. Save and exit the form

If you now go to the "Device File Explorer" and check the instance dir (`sdcard/odk/instances` and then the newest dir) you'll see both the files you've selected are still in the directory. I'm guessing this is because when the activity is killed the view model state is wiped with the activity and the `originalFiles` list will be empty when we return to the `FormEntryActivity` from the hierarchy screen. Before (on `master`) this worked because the `MediaManager` was a process level singleton and so state would be retained. 

It feels to me like it would be a lot nicer if when the form was saved or (or ignore changes is hit) it just deleted any files not being used in the form but that could require a lot of rework. I'm thinking we should use the new [Saved State module](https://developer.android.com/topic/libraries/architecture/viewmodel-savedstate) to persist `originalFiles` between activity instances (as opposed to going back to a process singleton) and then file an issue around carrying out the rework.

I'd like to pull in @lognaturel and @grzesiek2010 for sanity checks here though!

seadowg (changes requested):
Needs discussion as mentioned in comment

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
lognaturel: This case does seem possible if someone needs to get clarification on what file to attach, for example. Certainly it'd be best not to have a known regression in behavior.

We haven't used the saved state module yet. It might be simpler to use `onSavedInstanceState`. But yes, saving `originalFiles` between activity instances for now seems good.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -398,20 +420,21 @@ protected void onPostExecute(SaveToDiskResult saveToDiskResult) {
...
-        private final Analytics analytics;
+    public static class Factory extends AbstractSavedStateViewModelFactory {
+        private Analytics analytics;
 
+        public Factory(@NonNull SavedStateRegistryOwner owner, @Nullable Bundle defaultArgs) {
+            super(owner, defaultArgs);
+        }
 
-        public Factory(Analytics analytics) {
+        public void setAnalytics(Analytics analytics) {
```
seadowg: This could still be set in the constructor right? Could the constructor not just call `super` then assign the field?

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -85,7 +84,9 @@ public FormSaveViewModel(SavedStateHandle stateHandle, Clock clock, FormSaver fo
 
     @Override
     protected void onCleared() {
-        releaseMediaManager();
+        originalFiles.clear();
+        recentFiles.clear();
+
         stateHandle.set(ORIGINAL_FILES, originalFiles);
```
seadowg: Do we need to do this if we're also handling it in `markOriginalFileOrDelete` (same for the `RECENT_FILES`)?

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -368,6 +395,47 @@ public void resumeFormEntry_clearsSaveResult() {
         assertThat(saveResult.getValue(), equalTo(null));
     }
 
+    @Test
+    public void markOriginalFileOrDelete_whenFileNameIsEmpty_doesNotSaveFileForDelete() {
```
seadowg: It looks like all these new tests are mainly there to drive out the use of `SavedStateHandle`. I think a better way to test it would be more functionally by checking that if I call the methods that set state then use that saved state to create another view model then I get the behaviour I expect (probably that save or ignore do the right thing). This checks that the values are loaded from save state as well as written to it and also means we probably just need one test.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
SaumiaSinghal: I forgot to change this. I changed the structure when I was trying different things.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
SaumiaSinghal: Hi @seadowg! Do you mean if I need the `onCleared` method or not?

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -353,6 +353,24 @@ public void whenFormSaverFinishes_mediaManagerIsCleared() {
...
+        FormSaveViewModel restoredViewModel = new FormSaveViewModel(savedStateHandle, () -> CURRENT_TIME, formSaver, mediaUtils, null);
+        restoredViewModel.formLoaded(formController);
+
+        restoredViewModel.saveForm(Uri.parse("file://form"), true, "", false);
+        whenFormSaverFinishes(SaveFormToDisk.SAVED);
+
+        Map<String, String> originalFiles = savedStateHandle.get(FormSaveViewModel.ORIGINAL_FILES);
+        Map<String, String> recentFiles = savedStateHandle.get(FormSaveViewModel.RECENT_FILES);
+
+        assertThat(originalFiles.isEmpty(), equalTo(true));
```
seadowg: I think this should check that the files are deleted etc (so check the correct calls are made to `mediaUtils`) rather than checking the `savedStateHandle`).

seadowg: Actually so maybe we only need `markOriginalFileOrDelete_whenQuestionIndexHasAnswer_onRecreatingViewModel_deletesFile`? Unless there are reasons to check `savedStateHandle` directly.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -26,37 +29,66 @@
...
+            originalFiles = stateHandle.get(ORIGINAL_FILES);
+        }
+        if (stateHandle.get(RECENT_FILES) != null) {
+            recentFiles = stateHandle.get(RECENT_FILES);
+        }
+    }
+
+    @Override
+    protected void onCleared() {
+        originalFiles.clear();
```
seadowg: I don't think we need to clear these variables out here. They are part of the ViewModel which will be removed from memory.

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
seadowg: I don't **think** we need anything in here.

seadowg (changes requested):
Still think the tests could be reworked a little further. Again, I think the tests for driving out the use of SavedStateHandle shouldn't have to check what's actually in the instance being used by the view models. The test should just check that the viewmodel behaves as expected (i.e. clearing up files on save) when it is recreated.

seadowg:

I think we're just in a back and forth here around some tweaks to the code but would be good to get started testing this as it'd be nice to include in 1.28.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -325,6 +338,39 @@ public void whenFormSaverFinishes_isSaving_returnsFalse() {
...
+        assertThat(recentFiles.isEmpty(), equalTo(true));
+    }
+
+    @Test
+    public void whenFormSaverFinishes_onRecreatingFormSaveViewModel_mediaManagerIsCleared() {
```
seadowg: I'm not sure you need this test and `whenFormSaverFinishes_mediaManagerIsCleared`. Could you explain it or do you think we should just delete this one?

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -325,6 +338,39 @@ public void whenFormSaverFinishes_isSaving_returnsFalse() {
...
+        viewModel.markOriginalFileOrDelete("index", "blah");
+        viewModel.replaceRecentFileForQuestion("index", "blah");
+
+        viewModel.saveForm(Uri.parse("file://form"), true, "", true);
+        whenFormSaverFinishes(SaveFormToDisk.SAVED);
+
+        Map<String, String> originalFiles = savedStateHandle.get(FormSaveViewModel.ORIGINAL_FILES);
+        Map<String, String> recentFiles = savedStateHandle.get(FormSaveViewModel.RECENT_FILES);
+
+        assertThat(originalFiles.isEmpty(), equalTo(true));
```
seadowg: Instead of verifying the files here I think this test should run `saveForm` again and check that `deleteImageFileFromMediaProvider` is only called the first time. Does that make sense? Again, this is to test the intended behaviour rather than an implementation detail of that behaviour.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -59,12 +65,19 @@ public void setup() {
...
         logger = mock(AuditEventLogger.class);
         formSaver = mock(FormSaver.class);
+        mediaUtils = mock(MediaUtils.class);
         Analytics analytics = mock(Analytics.class);
 
         when(formController.getAuditEventLogger()).thenReturn(logger);
         when(logger.isChangeReasonRequired()).thenReturn(false);
 
-        viewModel = new FormSaveViewModel(() -> CURRENT_TIME, formSaver, analytics);
+        Map<String, String> originalFiles = new HashMap<>();
```
seadowg: It looks like you don't need these 4 lines for anything?

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
SaumiaSinghal: I think we can delete `whenFormSaverFinishes_mediaManagerIsCleared`. We do not have any other test to check whether mediaManager is released when form save finishes. I'm not sure is it is really important to check whether all files get deleted when form save task is finished.

seadowg: Yup I agree. I think my comment below lays out one test that would capture what `whenFormSaverFinishes_mediaManagerIsCleared` is trying to test.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
SaumiaSinghal: To verify calls to `deleteImageFileFromMediaProvider()`, I should use actual `new DisFormSaver()` instead of `FormSaver`, right?

seadowg: Oh I just meant like you did in the other test and call `saveForm` on the view model. Does that make sense. The test is to verify that if we call save once and then call save again then we don't try and delete the files from the first save. That's my understanding anyway. Does that make sense?

seadowg: So pseudo code would be:

* set up files to be deleted/saved
* call save on view model
* check files are saved/deleted correctly (using mock of `MediaUtils`)
* call save again
* check no files are saved/deleted (you can pass `once()` to `verify` to check it method has only been called once)

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -253,8 +279,35 @@ public AuditEventLogger getAuditEventLogger() {
...
+        if (questionIndex != null && fileName != null) {
+            if (recentFiles.containsKey(questionIndex)) {
+                mediaUtils.deleteImageFileFromMediaProvider(recentFiles.get(questionIndex));
+            }
+            recentFiles.put(questionIndex, fileName);
+            stateHandle.set(RECENT_FILES, recentFiles);
+        }
+    }
 
+    private void clearMediaFiles() {
```
SaumiaSinghal, line 305: When `saveToDisk()` task is complete, I just call `clearMediaFiles()` . So when I try to verify call to `mediaUtils.deleteImageFileFromMediaProvider()`, the test fails.

seadowg: Ah sorry because the `formSaver` is mocked! My bad. The way to verify things are set up correctly then is to assert on the files passed to the `fromSaver` (as part of `verify`) instead of the call the `mediaUtils`.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
SaumiaSinghal: Hi! the unit test `saveForm_runsWith_correctFiles()` is failing. Sorry I meant to name it `saveForm_savesCorrectFiles()`. I confirmed the failure is bacause of calling `clearMediaFiles()` in OnComplete, but that should not effect the FakeFormSaver, right?

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android: 7.0, 8.1, 9.0

Verified cases:
* Audio is not hearable after removing response on Audio widget
* The blue line always shows audio progress
* Choose Sound is not available for appearance="new"
* Adding audio file from Google Drive is possible
* Record sound
* Device rotation
* Light/dark mode
* Don't keep activities

lognaturel:
@mmarciniak90's pass looks quite thorough. I'm going to merge this because it's going to make tracking what's in the final beta a little easier. If any other problem is found, please file an issue. 🙏 Feel free to shake your fist in my general direction if I've caused trouble here.

mmarciniak90:

PR #4010, opened 2020-08-04
Fix deleting Instances of Forms without version

### LINKED ISSUE #4004: Unable to delete submission of softly deleted form without version
Collect v1.28.x, Android v5-7-10.x, device used...

#### Problem description
Testing result from #3992 scenario from #3961  is failing:
> Given there's a form definition formID/version combination that has been soft deleted
When I visit Delete Saved Form
And I delete the last Saved Form corresponding to that formID/version combination
That form definition formID/version combination's files and database entry should be fully deleted

1. Download form ex. from Drive or copy to form folder manually
2. Fill submission of the form
3. Choose match exactly server in Form management (with default server or any other that doesn't have that specific form.
4. Go to fill blank form - see that form is softly deleted
5. Go to form folder - form is visible
6. Go to Delete Saved Form and try to remove submission - submission fails to delete - toast is visible. The form is still not fully deleted. The other thing is that Delete Selected button becomes disabled when saved form to delete is selected. 
7. The form is still not fully deleted when I choose to delete saved forms in reset application. The form is fully deleted only after selecting blank forms in reset application. 

gif of example attached:
deletesub

#### Expected behavior
- User should be able to delete submission using Delete saved form tab
- Form should be fully deleted after deleting submission manually
- Form should be fully deleted after deleting submission using reset application/reset saved form

#### Other information 

seadowg:
Closes #4004

Comments inline.

Shouldn't be any changes other than the issue being fixed.

Any form without a version.


collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.hamcrest.Matchers.notNullValue;
+
+public abstract class FormsRepositoryTest {
+
+    public abstract FormsRepository buildSubject();
+
+    public abstract String getFormFilesPath();
+
+    @Test
+    public void get_whenFormHasNullVersion_returnsForm() {
```
seadowg, line 19: Given that the error here was the SQL query being constructed incorrectly it felt like the right time to start adding tests. Using a contract testing pattern like we're doing here lets us verify that our "real" (`DatabaseFormsRepository`) and "fake" (`InMemFormsRepository`) both pass the same set of tests. This lets us be more confident that we can switch the "real" object for the "fake" object in tests.

collect_app/src/main/java/org/odk/collect/android/forms/Form.java
```diff
@@ -136,7 +137,7 @@ public Builder jrVersion(String jrVersion) {
         }
 
         public Builder formFilePath(String formFilePath) {
-            this.formFilePath = new StoragePathProvider().getFormDbPath(formFilePath);
```
seadowg, line 139: It felt wrong that the `Form` object should have to care about storage details, so I removed this dependency. This also made it simpler to write tests for the `InMemFormsRepository` (which was my clue that something could be improved).

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormsRepository.java
```diff
@@ -40,16 +40,16 @@ public boolean contains(String jrFormId) {
...
+        return queryForForm(_ID + "=?", new String[]{id.toString()});
     }
 
     @Nullable
     @Override
-    public Form get(String jrFormId, String jrVersion) {
-        try (Cursor cursor = new FormsDao().getFormsCursor(JR_FORM_ID + "=? AND " + JR_VERSION + "=?", new String[]{jrFormId, jrVersion})) {
-            return getFormOrNull(cursor);
+    public Form get(String jrFormId, @Nullable String jrVersion) {
+        if (jrVersion != null) {
```
lognaturel, line 49: Feels too bad to have a mix of raw queries passed in to the dao layer and higher-level calls on the dao. Wouldn't `getFormsCursorSortedByDateDesc` do what you need here?

collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
lognaturel: I think another option would be to let the dao layer deal with queries and do testing at that level.

lognaturel (approved):
This is ok but see what you think about my questions/comments.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormsRepository.java
seadowg: You're right we should be clearer around which way we're going here: Are we wrapping the Dao or trying to remove it and wrapping the content provider? Personally I'm leaning towards the latter.

To your comment below (as it's related and I can't reply for some reason): I think I'd like to have testing around the repositories as I would envision us being able to create a new one (using Room for example) and we'd be able to just add another `FormRepositoryTest` to drive that out. Right now the actual test is completely divorced from the querying/storage. 

Maybe it makes sense to follow up with a PR trying to inline as much of the Dao as possible into the repo? That's conditional on us agreeing that's the direction we want to go of course!

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- forms with version 
- forms without assigned version
- Aggregate, Google Drive and Cenrtal 
- form copied to stogare manually
- form copied to storage when execute espresso tests
- automatic delete opotion
- auto delete + auto send
- auto delete defined in form 
- delete submissions of previously soft deleted form 
- different statuses of submissions 
- reset application: blank forms

mmarciniak90:
Tested with success also on Android versions: 7.0, 8.1, 9.0

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseFormsRepository.java
lognaturel: Ah yes, right, I forgot that we have too many layers going on. I agree that a goal would be to remove the existing DAO layer. I think that for now making the repositories use the content providers makes sense. If we were to switch to something like using Room, I think we'd flip that around. But doesn't matter for now.

PR #4013, opened 2020-08-05
Make sure Automatic Download downloads updates that have already been notifiied

### LINKED ISSUE #4012: Automatic download doesn't work for 'previously downloaded form only' after refusing to update manually

#### Problem description
When testing #3992 I have noticed a regression issue for previously downloaded forms only option. When you select that option you will get notification about available form update and if you do not download it and change to automatic download - automatic download doesn't work even if you see that there is form uptade info in Get Blank Forms. Collect stops to recognize there is form update after refusing to download the first try. 

1. Download form from server versions of the form (i have tried for Aggregate)
2. Log to another server where you have version2/ upload version2 to the same server (it doesn't matter which way you choose it's reproducible for both)
3. This is an update to a form you have info that should be visible in get blank form.
4. Go to General Settings/Form Management
5. Choose Previously downloaded forms only  
6. You will get a notification immediately
7. Ignore available update 
8. Select Automatic download checkbox
9. Wait for notification and form update

You will never get notification and form won't be updated.

#### Expected behavior
After 15 minutes, notification about automatic download should display, form should be updated automatically.

#### Other information 
I have tried the same case for 1.27.3 store version and it works properly.

seadowg:
Closes #4012

Added tests to capture the edge case

Comments inline

Good to test Previously Downloaded Only with and without Automatic Download enabled to double-check everything is the way it should be.


collect_app/src/main/java/org/odk/collect/android/database/FormDatabaseMigrator.java
```diff
@@ -45,97 +24,51 @@
...
 
-/**
- * This class helps open, create, and upgrade the database file.
- */
-public class FormsDatabaseHelper extends SQLiteOpenHelper {
-    private static final String DATABASE_NAME = "forms.db";
-    public static final String FORMS_TABLE_NAME = "forms";
-
-    public static final int DATABASE_VERSION = 9;
+public class FormDatabaseMigrator {
```
seadowg, line 28: I wanted to pull this out so we could have an object that was easier to setup and work with than the `SQLiteOpenHelper` implementations that need `DatabaseContext`.

collect_app/src/main/java/org/odk/collect/android/notifications/NotificationManagerNotifier.java
```diff
@@ -35,18 +42,26 @@
...
+    public NotificationManagerNotifier(Application application, PreferencesProvider preferencesProvider) {
         this.application = application;
         notificationManager = (NotificationManager) application.getSystemService(NOTIFICATION_SERVICE);
+        this.preferencesProvider = preferencesProvider;
     }
 
     @Override
-    public void onUpdatesAvailable() {
+    public void onUpdatesAvailable(List<ServerFormDetails> updates) {
+        SharedPreferences metaPrefs = preferencesProvider.getMetaSharedPreferences();
```
seadowg, line 59: This is the key change here: switch away from using the database to record the versions the user is notified about and also making it more of a UI's (the `Notifier`) concern.

collect_app/src/test/java/org/odk/collect/android/database/FormDatabaseMigratorTest.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.JRCACHE_FILE_PATH;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.JR_FORM_ID;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.JR_VERSION;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.LANGUAGE;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.MD5_HASH;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns.SUBMISSION_URI;
+import static org.odk.collect.android.provider.FormsProviderAPI.FormsColumns._ID;
+
+@RunWith(AndroidJUnit4.class)
+public class FormDatabaseMigratorTest {
```
seadowg, line 40: There were tests for the database helper checking migrations, but they were out of date and `Ignore`d. Interested to know if people think we need more coverage here.

collect_app/src/main/java/org/odk/collect/android/database/FormDatabaseMigrator.java
```diff
@@ -45,97 +24,51 @@
...
+                upgradeToVersion4(db, oldVersion);
+            case 4:
+                upgradeToVersion5(db);
+            case 5:
+                upgradeToVersion6(db);
+            case 6:
+                upgradeToVersion7(db);
+            case 7:
+                upgradeToVersion8(db);
+                break;
```
lognaturel: Move `break` down to allow upgrades from versions other than 8.

collect_app/src/test/java/org/odk/collect/android/database/FormDatabaseMigratorTest.java
lognaturel: I know you've been trying to kill those tests for a long time. 😂 They were added after we had several significant bugs related to `onDowngrade`. I think the structure and test are good but that it would be good to also downgrade from a fictitious high version with an extra and missing column to avoid regressions there. We can assume we wouldn't touch `onDowngrade` anymore but it's hard to reason about and tempting to try to improve.

lognaturel (changes requested):

collect_app/src/test/java/org/odk/collect/android/database/FormDatabaseMigratorTest.java
seadowg: haha yeah totally happy to add more. The old tests were failing and needed rewritten anyway. I'll add a couple of downgrade tests (one with extra and one with missing columns).

collect_app/src/main/java/org/odk/collect/android/database/FormDatabaseMigrator.java
seadowg: Oh damn. Yes. I might try and add some tests to drive this out.

seadowg:
@lognaturel test issues fixed. Wasn't going to be working today as usual but couldn't get this out of my head!

lognaturel (approved):
Yay for cleaning up unfinished business!

kkrawczyk123:
@seadowg I have a question. Scenario from the issue seems to be fixed but I am wondering if you ignore notification about the update and do not change anything in settings - shouldn't you see the notification about an update again after 15 minutes? I think that it should appear.

seadowg:
@kkrawczyk123 I believe the behaviour in this PR will match master though right? I do see where you're coming from: right now if you dismiss the notification it will never be seen again. I'm not sure if this is bad or not so it might be something to discuss in an issue or forum post. Apologies if I'm totally wrong and there is a regression here though!

kkrawczyk123:
ok, @seadowg I checked on older versions and you are right. It's always good to double-check.

**Tested with success!**

Tested scenarios:
- automatic download
- match exactly
- previously downloaded forms
- scenario from issue
- used servers: Drive, Aggregate, Central
- auto send
- auto delete
- check on fill submission, match exactly and soft delete

mmarciniak90:
It has also been successfully tested on Android versions: 7.0, 8.1, 9.0
Verified cases work as expected.

PR #4016, opened 2020-08-07
Adjusted gesture navigation

### LINKED ISSUE #3966: Improve gesture navigation while filling forms
Collect - all versions, Android - all versions

#### Problem description
According to our discussion on slack:
https://getodk.slack.com/archives/C35ENHA64/p1594315293484400
it would be good to adjust gesture navigation in `FormEntryActivity` a bit in favor of scrolling up/down because new not trained users might swipe to next/previous questions accidentally while scrolling.
 
#### Expected behavior
Navigating backward/forward should require a bit more horizontal gesture.

grzesiek2010:
Closes #3966 

I tested navigating in a form.

It's the easiest solution. We also talked about viewpager but it would require more changes so despite the fact it's something we should do in the future, I think for now this quick adjustment is fine.

It just requires swiping right/left that's all. It doesn't even require testing navigation using buttons only gesture navigation.

Any form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
+++ b/collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
@@ -1,0 +1,145 @@
+package org.odk.collect.android.listeners;
+
+import android.content.Context;
+import android.view.GestureDetector;
+import android.view.MotionEvent;
+
+import org.odk.collect.android.formentry.ODKView;
+import org.odk.collect.android.preferences.GeneralKeys;
+import org.odk.collect.android.preferences.GeneralSharedPreferences;
+import org.odk.collect.android.utilities.FlingRegister;
+import org.odk.collect.android.utilities.ScreenUtils;
+
+import timber.log.Timber;
+
+public class SwipeHandler {
+    private final GestureDetector gestureDetector;
+    private final OnSwipeListener onSwipe;
+    private ODKView odkView;
+    private boolean allowSwiping = true;
+    private boolean beenSwiped;
+
+    public interface OnSwipeListener {
+        void showPreviousView();
+        void showNextView();
+    }
+
+    public SwipeHandler(Context context) {
+        gestureDetector = new GestureDetector(context, new GestureListener());
+        this.onSwipe = (OnSwipeListener) context;
+    }
+
+    public void setOdkView(ODKView odkView) {
+        this.odkView = odkView;
+    }
+
+    public void setAllowSwiping(boolean allowSwiping) {
+        this.allowSwiping = allowSwiping;
+    }
+
+    public void setBeenSwiped(boolean beenSwiped) {
+        this.beenSwiped = beenSwiped;
+    }
+
+    public boolean beenSwiped() {
+        return beenSwiped;
+    }
+
+    public GestureDetector getGestureDetector() {
+        return gestureDetector;
+    }
+
+    public class GestureListener implements GestureDetector.OnGestureListener {
+
+        @Override
+        public boolean onDown(MotionEvent event) {
+            return false;
+        }
+
+        @Override
+        public void onShowPress(MotionEvent e) {
+
+        }
+
+        @Override
+        public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
+            // The onFling() captures the 'up' event so our view thinks it gets long pressed. We don't want that, so cancel it.
+            if (odkView != null) {
+                odkView.cancelLongPress();
+            }
+            return false;
+        }
+
+        @Override
+        public void onLongPress(MotionEvent e) {
+
+        }
+
+        @Override
+        public boolean onSingleTapUp(MotionEvent e) {
+            return false;
+        }
+
+        @Override
+        public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY) {
+            FlingRegister.flingDetected();
+
+            // only check the swipe if it's enabled in preferences
+            String navigation = (String) GeneralSharedPreferences.getInstance()
+                    .get(GeneralKeys.KEY_NAVIGATION);
+
+            if (e1 != null && e2 != null
+                    && navigation.contains(GeneralKeys.NAVIGATION_SWIPE) && allowSwiping) {
+                // Looks for user swipes. If the user has swiped, move to the
+                // appropriate screen.
+
+                // for all screens a swipe is left/right of at least
+                // .25" and up/down of less than .25"
+                // OR left/right of > .5"
+                int xpixellimit = (int) (ScreenUtils.getScreenWidth() * .25);
+                int ypixellimit = (int) (ScreenUtils.getScreenHeight() * .25);
+
+                if (odkView != null && odkView.suppressFlingGesture(e1, e2, velocityX, velocityY)) {
+                    return false;
+                }
+
+                if (beenSwiped) {
+                    return false;
+                }
+
+                float diffX = e1.getX() - e2.getX();
+                float diffY = e1.getY() - e2.getY();
+
+                if (odkView != null && odkView.canScrollVertically() && getGestureAngle(diffX, diffY) > 30) {
+                    return false;
+                }
+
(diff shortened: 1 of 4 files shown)


lognaturel (changes requested):
I believe that what this is doing is requiring that the finger travel in the horizontal dimension be at least 300 pixels greater than the finger travel in the vertical dimension. This seems like a dangerous approach to me because it doesn't take pixel density into account and because it makes short horizontal swipes much harder. I have to significantly change my "swiping technique" to navigate a form on my Galaxy Tab A. I think we need this change to be incremental and to have as little effect as possible on folks who don't intend to do any vertical scrolling at all.

There are a lot of different levers to play with here so it's especially critical to get a description of how you came about your proposed solution and why you think it's the best. It would be good to know what devices (physical or emulator) you were able to try, too.

My guess is that we're usually hitting the `Math.abs(diffX) > xpixellimit * 2` part of the condition which means any swipe with a horizontal travel of more than a half inch is interpreted as a horizontal swipe. Maybe changing that threshhold to `xpixellimit * 2` (.75in) would help? Did you also take a look at what other applications or components that detect both horizontal and vertical scrolls do?

Perhaps you could also consider a variation on your strategy that instead of using a fixed allowed difference between the horizontal and vertical dimension allows some percentage deviation between the two.

grzesiek2010:
> This seems like a dangerous approach to me because it doesn't take pixel density into account and because it makes short horizontal swipes much harder.

Shame on me, sorry :disappointed: 

Of course a better approach is to just calculate an angel in degrees (0-90 degrees for x-axis). I improved my solution. Now it blocks swiping if the angel is > 30 where originally it was 45. I don't know if it's a perfect solution because it's hard to say but through trail and error I could say it works pretty well.

What do you think.

grzesiek2010:
@lognaturel 
I added that check we were talking about to adjust swiping only if a view is scrollable vertically: https://github.com/getodk/collect/pull/4016/commits/7958bce0916eddfde7cff6f236082b8bea328010
I also move the code responsible for gesture navigation to a separate class to make ForEntryActivity a bit smaller: https://github.com/getodk/collect/pull/4016/commits/70a3b6a707f9353e8de207eff964fdc96c84bb0e

seadowg:
Just passing through but I'm liking the swipe logic being pulled out here. I'm curious whether the `FormIndexAnimationHandler` could help simplify any of the logic here or if this new handler could replace that in some way. Just throwing that out there as it's easy to miss things in `FormEntyActivity`! It might be something to look at in the future rather than now.

lognaturel:
The approach of not capturing swipes with an angle greater than some degree if the view is vertically scrollable seems great to me. I've applied just that change to `master` and I feel pretty good about your choice of 30 degrees. I'd still really like to read a little bit about how you selected it to make sure our trials and thought process were complementary.

I feel like I'm going a bit crazy, though, because when I use this branch, I really feel like my normal short swipes don't work. I can't find anything in the code that would explain that, though! Can you see if you reproduce what I'm talking about? I'm guessing my normal swipes are just about .25" because I've probably trained myself to just reach the threshold.

I hadn't thought about `FormIndexAnimationHandler` but agree with @seadowg that the functionality feels related and combining them would be good.

Minor: it'd be better for `diffX` and `diffY` to include the absolute value call since the sign isn't ever needed.

grzesiek2010:
> I feel like I'm going a bit crazy, though, because when I use this branch, I really feel like my normal short swipes don't work. I can't find anything in the code that would explain that, though! Can you see if you reproduce what I'm talking about? I'm guessing my normal swipes are just about .25" because I've probably trained myself to just reach the threshold.

You are right something changed what I missed. Before we calculated:
[int xpixellimit = (int) (dm.xdpi * .25);](https://github.com/getodk/collect/pull/4016/files#diff-03ff45091550285dd88b56927010c457L2548)
but now it is:
[int xpixellimit = (int) (ScreenUtils.getScreenWidth() * .25);](https://github.com/getodk/collect/pull/4016/files#diff-9363bce37fd275eaebf6e245e7a1d283R102)
where
dm.xdpi - the exact physical pixels per inch of the screen in the X dimension
ScreenUtils.getScreenWidth() uses widthPixels - the absolute width of the available display size in pixels
so for example on my device returned values are:
dm.xdpi - 440
ScreenUtils.getScreenWidth() - 1080

so yeah swipes should be based on xdpi  not on widthPixels because then the bigger your screen is the longer swipe you need.
**It's fixed!**

> Minor: it'd be better for diffX and diffY to include the absolute value call since the sign isn't ever needed.

I agree. Done.

> I'm curious whether the FormIndexAnimationHandler could help simplify any of the logic here or if this new handler could replace that in some way. Just throwing that out there as it's easy to miss things in FormEntyActivity! It might be something to look at in the future rather than now.

`FormIndexAnimationHandler` and `SwipeHandler` have different responsibilities and I don't think we should merge them or something.

collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
```diff
@@ -0,0 +1,152 @@
...
+            // only check the swipe if it's enabled in preferences
+            String navigation = (String) GeneralSharedPreferences.getInstance()
+                    .get(GeneralKeys.KEY_NAVIGATION);
+
+            if (e1 != null && e2 != null
+                    && navigation.contains(GeneralKeys.NAVIGATION_SWIPE) && allowSwiping) {
+                // Looks for user swipes. If the user has swiped, move to the appropriate screen.
+
+                // For all screens a swipe is left/right of at least .25" and up/down of less than .25" OR left/right of > .5"
+                int xpixellimit = (int) (ScreenUtils.getXPixelsNumberPerInch() * .25);
```
lognaturel: I think this is clearer with `xdpi` and `ydpi` inline. The comment is very clear and `dpi` is common for "dots per inch".

lognaturel:
> the bigger your screen is the longer swipe you need

Ahhh, don't mess with my head like that!! 😬😄

I'm happy with this. I think the approach is easy to explain and feels very low risk. To summarize, if there's a `R.id.odk_view_container`, then it's assumed the view can scroll vertically, and any fling motion that is more steeply angled than 30 degrees from positive x are discarded from consideration as swipes. I feel a pretty significant improvement.

As we discussed, @seadowg should also agree this is low-risk for us to consider it for v1.28. I don't mind revisiting handlers at a future time.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
grzesiek2010: Right but still I think it's good to keep the code in a separate class (ScreenUtils) so I just change naming.

lognaturel: I guess that by avoiding the direct dependency on the Collect instance it gets things a little closer to being testable. Fine by me.

lognaturel:
Test failures are because of CircleCI's switch to JDK 11. See https://github.com/getodk/collect/pull/4029.

seadowg (approved):
Yeah this seems to make sense to me! If anything the core advantage here for me is that the code is isolated and is easier to understand so if we do notice problems during beta/regression I think it'll be easier to fix than it was before. I really like the `canScrollVertically` idea as well!

Still failing but just needs a rebase before adding `needs testing`.

grzesiek2010:
> Still failing but just needs a rebase before adding needs testing.

Fixed so I'm adding the label.

mmarciniak90:
**Tested with success**

Verified on Android 7.0, 8.1, 9.0

Verified cases:
* form with a lot of questions on one page - I can swipe even diagonally and I'm not moved to the next page (not intentionally). I can easily go to the next page when I swipe horizontally.
* form with one question on one page - there is no difference
* `All widgets` form, especially the first page with long description and `list group`
* horizontal and vertical device orientation 
* RTL language

grzesiek2010:
We agreed that it's enough if the pr has been tested by @mmarciniak90, @kkrawczyk123 you will just test it on a daily basis during regression testing.

Issue #3988, opened 2020-07-27
Notifications not being translated


seadowg:
~~Blocked by #3986.~~

#### Software and hardware versions 

Collect v1.27.3 (older versions as well)

#### Problem description

It seems notifications for auto send and auto update are not being translated.

#### Steps to reproduce the problem

1. Set Collect's language to something other than English
1. Enabled auto send
1. Fill out and finalize a form

You'll see the notification is in English.

#### Expected behavior

Notifications use the same language as the rest of the app.

#### Other information 

This will be problem wherever we use the `Application` to grab resource strings rather than an `Activity`. This is because the way we change the language only does so for `Activity` contexts, but we're unable to change the resource locale at an `Application` level. We're seeing this in Notifications as they are mostly created from background jobs that would use the application context as they don't have access to an `Activity`. 

#3986 introduces code that lets us create resource objects with the correct locale from the application context (in `NotificationManagerNotifier`). Once that is merged we can use that for all the notifications.

grzesiek2010:
It's a common problem in our app and there are other reported issues like: https://github.com/getodk/collect/issues/3029
In some cases we just can't use activity context.
How about creating a helper class for dealing with strings?
```java
public class TranslationHandler {
    public static String getString(int stringId) {
        return getLocalizedResources(Collect.getInstance()).getString(stringId);
    }

    // The application context will give us the system's locale
    private static Resources getLocalizedResources(Context context) {
        Configuration conf = context.getResources().getConfiguration();
        conf = new Configuration(conf);
        conf.setLocale(new Locale(LocaleHelper.getLocaleCode(context)));
        Context localizedContext = context.createConfigurationContext(conf);
        return localizedContext.getResources();
    }
}
```

and calling:
```java
TranslationHandler.getString(R.string.sting_name);
```
across the app?

seadowg:
Yeah I like that! I think it'd be nicer for it to take in `Context` as well as a `int` string id, so we're not relying on `Collect.getInstance()`.

grzesiek2010:
> I think it'd be nicer for it to take in Context as well as a int string id, so we're not relying on Collect.getInstance().

why do you find it important? Then every call will look like:
```java
TranslationHandler.getString(getActivity(), R.string.sting_name);
TranslationHandler.getString(getContext(), R.string.sting_name);
TranslationHandler.getString(Collect.getInstance(), R.string.sting_name);
```
[...]

PR #4017, opened 2020-08-07
Fixed auto-delete and auto-sent options

### LINKED ISSUE #3768: Form-level auto-delete option is not read from correct form version
Collect v1.26.1 and prior

#### Problem description
The `InstanceUploader.formShouldBeAutoDeleted` method does not take form version into account. This means that if a form update changes the auto-delete option, that may or may not be reflected in Collect's behavior.

1. Transfer a form with auto-delete set to false such as [this one](https://docs.google.com/spreadsheets/d/1rvg8VDX9Z_F_1KGvzPGNyxzw_bl3hImFgpOMk4Xn5UE/edit?usp=sharing)
1. Modify the form to set auto-delete to true and change the form version
1. See that the old auto-delete setting is used because the old form def is returned first in `InstanceUploader.formShouldBeAutoDeleted`

#### Expected behavior
For each instance sent, the attribute value for the version that was used to fill the instance should be used.

#### Other information 
This is a good opportunity to clean up the code here and use higher-level constructs like `Form` domain objects. Ideally each instance wouldn't have to hit the database to get this value.

It would also be good to check `auto-send` and the other columns that may be read in this way.

grzesiek2010:
Closes #3768 

I reproduced the issue and confirmed that this pr fixes it.

It is just a bug fix. Before we didn't take form versions into account while reading auto-sent and auto-delete props from database.

It's a safe change that should just fix the issue. Nothing else should be affacted.

No.


lognaturel:
Did you consider tests for this? I'm wondering whether you could use the `FormsRepository` to make that possible. It might then be better to wait for #4013 to be merged.

seadowg:
@lognaturel @grzesiek2010 it looks like `FormsRespository#get(String jrFormId, String jrVersion)` is on `master` so this could just use that. It'd make sense for the `FormsRepository` to be a constructor level dependency of the `InstanceSubmitter` I think. That way it should be feasible to write tests for it with the `InMemFormsRepository` (similar to those in `FormDeleterTest`).

seadowg (changes requested):
As above

grzesiek2010:
Ok I used `FormsRespository` and added tests using `InMemFormsRepository`.
I decided to place methods that check autoSend/autoDelete in InstanceUploaderUtils because they are called from various places so I think that utils class is a good choice.
Please review.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1887,7 +1887,7 @@ private void handleSaveResult(FormSaveViewModel.SaveResult result) {
                         // finalized specifies that it should always be auto-sent.
                         String formId = getFormController().getFormDef().getMainInstance().getRoot().getAttributeValue("", "id");
                         String formVersion = getFormController().getFormDef().getMainInstance().getRoot().getAttributeValue("", "version");
-                        if (InstanceUploaderUtils.formShouldBeAutoSent(formsRepository, formId, formVersion, GeneralSharedPreferences.isAutoSendEnabled())) {
+                        if (InstanceUploaderUtils.shouldFormBeSent(formsRepository, formId, formVersion, GeneralSharedPreferences.isAutoSendEnabled())) {
```
seadowg: Could this not just always call `formSubmitMananer.scheduleSubmit()` at this point? I might have missed something but I think the `InstanceSubmitter` (that the `AutoSendTaskSpec` will use will do the right thing and only send if it needs to.

If that's the case we could just test `InstanceSubmitter` rather than the static method (as it will be the only place we make that decision).

collect_app/src/main/java/org/odk/collect/android/utilities/InstanceUploaderUtils.java
```diff
@@ -154,7 +154,7 @@ public static boolean formShouldBeAutoSent(FormsRepository formsRepository, Stri
      *
      * If the form explicitly sets the auto-delete property, then it overrides the preference.
      */
-    public static boolean formShouldBeAutoDeleted(FormsRepository formsRepository, String jrFormId, String jrFormVersion, boolean isAutoDeleteAppSettingEnabled) {
+    public static boolean shouldFormBeDeleted(FormsRepository formsRepository, String jrFormId, String jrFormVersion, boolean isAutoDeleteAppSettingEnabled) {
```
seadowg, line 100: Might be nice to have these static methods in a new `InstanceUtils` class in either `instancemanagement` or `instances`. I can see us pulling the other methods out at some point (maybe as part of #3988).

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: I was thinking about it as well. I didn't change it because I think it would change the default behavior a bit. Now autosend is triggered if we save a form that requires it, that's why we have that check (`InstanceUploaderUtils.shouldFormBeSent()`). If we get rid of it it will be triggered even if just one of already saved forms requires autosend.

grzesiek2010: I know it's misleading because if you save a form that requires autosend the app will also send other already saved forms that require it too but it doesn't work the other way around - even if all already saved forms require autosend but the one you save in this step doesn't it won't be triggered. Probably we should rethink it @lognaturel

lognaturel: I think (hope?) in most cases the distinction won't matter because the instance currently being saved will be the only one available for auto submission.

One extreme case is if there are multiple submissions that are hanging around and systematically fail AND there's a form in play that doesn't auto send. In that case there's some advantage to the existing check. That doesn't seem very likely to me so I'd be ok with the simpler implementation.

grzesiek2010: Ok I removed that check and moved `shouldFormBeSent()` to `InstanceSubmitter`. However `shouldFormBeDeleted()` is still placed in `InstanceUploaderUtils` because it's used from different classes (InstanceUploaderTask). I noticed that the code between that `InstanceUploaderTask` and `InstanceSubmitter` is duplicated so probably it would be good to refactor it using `InstanceSubmitter` in `InstanceUploaderTask` maybe? Then both `shouldFormBeSent()` and `shouldFormBeDeleted()` would be placed in just one class (`InstanceSubmitter`). What do you think? Maybe as a separate pr if you agree?

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.java
```diff
@@ -35,20 +35,26 @@
...
+        this.formsRepository = formsRepository;
+        this.instancesRepository = instancesRepository;
     }
 
     public Pair<Boolean, String> submitUnsubmittedInstances() throws SubmitException {
         List<Instance> toUpload = getInstancesToAutoSend(GeneralSharedPreferences.isAutoSendEnabled());
+        return submitSelectedInstances(toUpload);
+    }
 
+    public Pair<Boolean, String> submitSelectedInstances(List<Instance> toUpload) throws SubmitException {
```
seadowg, line 60: I think this is meant to be `private`

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.java
```diff
@@ -147,19 +149,11 @@ public InstanceSubmitter(Analytics analytics) {
...
-        Cursor cursor = new FormsDao().getFormsCursorForFormId(jrFormId);
-        String formLevelAutoSend = null;
-        if (cursor != null && cursor.moveToFirst()) {
-            try {
-                int autoSendColumnIndex = cursor.getColumnIndex(AUTO_SEND);
-                formLevelAutoSend = cursor.getString(autoSendColumnIndex);
-            } finally {
-                cursor.close();
-            }
+    public static boolean shouldFormBeSent(FormsRepository formsRepository, String jrFormId, String jrFormVersion, boolean isAutoSendAppSettingEnabled) {
```
seadowg, line 164: In my head this would ideally be a `private` method that is just an implementation detail of `InstanceSubmitter`. The `InstanceSubmitter` would then have tests around it that setup instances and check that they are sent in the right contexts. Looking at though there is a lot of dependencies on static calls we'd need to pull out to write that test (`Collect.getInsance().getComponent()` and `GeneralSharedPreferences.isAutoSendEnabled()` for example). This PR moves the needle forward so I don't think it makes sense to block it on that rework.

Could you maybe just add a `@Deprecated` to this method and a `@deprecated` Javadoc explaining that really it should be a private method and that that requires refactoring `InstanceSubmitter` to make it testable/modular?

seadowg (changes requested):
Just a couple of tweaks!

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.java
grzesiek2010: I thought about future and refactoring the code where we could use this method from InstanceUploaderTask because as I sad there are duplicated code here and in that class.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.java
grzesiek2010: Ok, done.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.java
seadowg: Ah I see. I would tend to leave this private until then. It might be that when we make that change it would make sense for this logic to move to a different object that's responsible for submitting a single instance and then have this object use that which nicely separates collecting the instances to auto submit from the logic to actually submit one. Not a big deal for the moment though!

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified cases:
* uploaded to Aggregate 4 different versions
   * orx:auto-delete="false" orx:auto-send="true"
   * orx:auto-delete="true" orx:auto-send="true"
   * orx:auto-delete="true" orx:auto-send="false"
   * orx:auto-delete="false" orx:auto-send="false"
* forms downloaded manually
* `Exactly match server` set
* unselect `Hide old form versions` and fill a version with a lower number

PR #4020, opened 2020-08-10
Match exactly improvements

### LINKED ISSUE #4007: Improvements to match exactly.
Collect v1.28.x, Android vx.x.x, device used...

There are few small things that could be improved, testing result from #3992:

- When Match Exactly is selected, Get Blank Form is not visible in menu but it is still visible as enabled in Admin Settings/Main Menu Settings: 
Screenshot_20200804-094952
It should be disabled like Edit Saved Form option when Moving backward is selected.

- When Match Exactly is selected there is only one default option: Automatic download that is disabled and the checkbox is empty what might be little confusing:
Screenshot (Aug 4, 2020 9_55_34 AM)
It should still be disabled be the checkbox should be selected by default then. 

- When Use refresh arrow on Fill Blank Form list when there is no internet connection or airplane mode is turned on, the auto-download refresh keeps failing and failing. It might not be sure for user that he lost internet connection as it might happen suddenly and be totally unexpected so I feel like toast informing about the connection lost would be much friendlier for users, like when try to get blank form without connection:
Screenshot_20200804-100014

- In Admin Settings/User settings Blank form update mode has a weird indentation what differs it from other option:
Screenshot_20200804-101735
This issue is not visible on Android 5.1 but it is on all other devices.

seadowg:
Closes #4007

New and reworked tests. Verified the different issues manually as well.

Comments inline.

Should just fix issues. Good to double-check any features around them but everything has pretty good test coverage at this point.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/configure/SettingsUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/configure/SettingsUtils.java
@@ -1,0 +1,29 @@
+package org.odk.collect.android.configure;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import androidx.annotation.NonNull;
+
+import org.odk.collect.android.formmanagement.FormUpdateMode;
+import org.odk.collect.android.preferences.GeneralKeys;
+import org.odk.collect.android.preferences.Protocol;
+
+public class SettingsUtils {
+
+    private SettingsUtils() {
+
+    }
+
+    @NonNull
+    public static FormUpdateMode getFormUpdateMode(Context context, SharedPreferences generalSharedPreferences) {
+        String protocol = generalSharedPreferences.getString(GeneralKeys.KEY_PROTOCOL, null);
+
+        if (Protocol.parse(context, protocol) == Protocol.GOOGLE) {
+            return FormUpdateMode.MANUAL;
+        } else {
+            String mode = generalSharedPreferences.getString(GeneralKeys.KEY_FORM_UPDATE_MODE, null);
+            return FormUpdateMode.parse(context, mode);
+        }
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
@@ -41,6 +41,7 @@ import static org.odk.collect.android.preferences.GeneralKeys.KEY_GUIDANCE_HINT;
 import static org.odk.collect.android.preferences.GeneralKeys.KEY_IMAGE_SIZE;
 import static org.odk.collect.android.preferences.GeneralKeys.KEY_PERIODIC_FORM_UPDATES_CHECK;
 import static org.odk.collect.android.preferences.GeneralKeys.KEY_PROTOCOL;
+import static org.odk.collect.android.preferences.utilities.PreferencesUtils.displayDisabled;
 
 public class FormManagementPreferences extends BasePreferenceFragment {
 
@@ -105,9 +106,7 @@ public class FormManagementPreferences extends BasePreferenceFragment {
         } else {
             switch (FormUpdateMode.parse(getActivity(), formUpdateMode)) {
                 case MANUAL:
-                    automaticDownload.setEnabled(false);
-                    automaticDownload.setChecked(false);
-
+                    displayDisabled(automaticDownload, false);
                     updateFrequency.setEnabled(false);
                     break;
                 case PREVIOUSLY_DOWNLOADED_ONLY:
@@ -115,9 +114,7 @@ public class FormManagementPreferences extends BasePreferenceFragment {
                     updateFrequency.setEnabled(true);
                     break;
                 case MATCH_EXACTLY:
-                    automaticDownload.setEnabled(false);
-                    automaticDownload.setChecked(true);
-
+                    displayDisabled(automaticDownload, true);
                     updateFrequency.setEnabled(true);
                     break;
             }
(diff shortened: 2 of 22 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.java
```diff
@@ -116,22 +116,6 @@ public void whenPreviouslyDownloadedOnlyEnabled_checkingAutoDownload_downloadsUp
                 .assertAndDismissNotification("ODK Collect", "ODK auto-download results", "Success");
     }
 
-    @Test
-    public void whenGoogleDriveUsingAsServer_disablesPrefsAndOnlyAllowsManualUpdates() {
```
seadowg: It ended up being a lot simpler to move a bunch of the testing of these preferences into the fragment tests.

collect_app/src/main/java/org/odk/collect/android/preferences/FormManagementPreferences.java
```diff
@@ -91,44 +92,36 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
     }
 
     private void updateDisabledPrefs(String formUpdateMode, String protocol) {
         Preference updateFrequency = findPreference(KEY_PERIODIC_FORM_UPDATES_CHECK);
         CheckBoxPreference automaticDownload = findPreference(KEY_AUTOMATIC_UPDATE);
 
         if (Protocol.parse(getActivity(), protocol) == Protocol.GOOGLE) {
-            findPreference(KEY_FORM_UPDATE_MODE).setEnabled(false);
-            automaticDownload.setEnabled(false);
+            displayDisabled(findPreference(KEY_FORM_UPDATE_MODE), getString(R.string.manually));
```
seadowg, line 93: To me it feels a lot simpler for us to simply display the disabled prefs with the appropriate values rather than change the underlying value. This has two advantages:

* The code is more declarative - we don't have to actually change values of other prefs when a pref changes (like changing the value of form update mode when the protocol changes).
* If the user switches between different options the disabled prefs will revert to the values they had them at before.

collect_app/src/main/java/org/odk/collect/android/preferences/utilities/PreferencesUtils.java
```diff
@@ -0,0 +1,23 @@
...
+import androidx.preference.Preference;
+
+public class PreferencesUtils {
+
+    private PreferencesUtils() {
+
+    }
+
+    public static void displayDisabled(CheckBoxPreference preference, boolean displayValue) {
+        preference.setPersistent(false);
```
seadowg, line 13: This is clunky but I couldn't find any other way to just display a value in a `CheckBoxPreference`. You'll see the `String` version is much simpler.

lognaturel (approved):

kkrawczyk123:
@seadowg, all of the improvements look just great but very unfortunate regression has been introduced when match exactly is enabled and change server from Aggregate to Google drive - in form management manually blank form update mode is selected but:
- Get Blank Form is not visible in Main Menu
- Delete Blank Form is not visible in Delete Saved Form
- Match exactly arrow is visible in Fill Blank Form 

reproduction steps:
1. Select match exactly for default server
2. Change server to Google Drive
gif attached: 
ezgif com-video-to-gif (18)
The issue does not occur for master, match exactly option is turned off after selecting Google Drive as a server.

seadowg:
@kkrawczyk123 ach I think I know where the problem is. Will fix!

collect_app/src/main/java/org/odk/collect/android/configure/SettingsUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import org.odk.collect.android.preferences.Protocol;
+
+public class SettingsUtils {
+
+    private SettingsUtils() {
+
+    }
+
+    @NonNull
+    public static FormUpdateMode getFormUpdateMode(Context context, SharedPreferences generalSharedPreferences) {
```
seadowg, line 19: I was incredibly tempted to add a Settings interface for abstracting away this kind of logic (and shared prefs themselves). This would be in a weird place however as in most places it's pretty simple to just grab the actual pref value to work out what "state" we're in. I think if these utils start to grow we can revisit the idea.

kkrawczyk123:
@seadowg I can see that issue has been fixed but when following exactly the same scenario, get blank form is still visible as disabled in Admin Settings/ Main Menu Settings:
ezgif com-video-to-gif (19)

seadowg:
@kkrawczyk123 ah that's because I forgot about that case. Sorry this stuff is a bit all over the place as having settings disable others is a bit of a logic bomb 💣. Will fix that.

seadowg:
@kkrawczyk123 fixed!

mmarciniak90:
Verified cases:
* Get blank form is disabled and not editable when Match Exactly is set
* Automatic download that is enabled and not editable Match Exactly is set
* Toast informing about the connection lost is visible when user clicks on refresh arrow

mmarciniak90:
**Tested with success**

Verified cases:
* `Get blank form` is disabled and not editable when Match Exactly is set
* `Automatic download` is enabled and not editable when Match Exactly is set
* Toast informing about the connection lost is visible when user clicks on refresh arrow
* `Admin Settings` -> `User settings` -> `Blank form update mode` is in the same line as all options
* `Automatic download` is unselected and not editable when server is changed to Google Drive 
* `Get blank form` is enabled when Google Drive is set as a server
* Scan QR code
* light / dark mode

lognaturel:
@kkrawczyk123 has already had a first pass and @mmarciniak90's looks quite thorough. I'm going to merge this because it's going to make tracking what's in the final beta a little easier. If any other issue is found, please file an issue. 🙏

mmarciniak90:

PR #4025, opened 2020-08-11
Remove Answer Text View in Url Widget

SaumiaSinghal:
This PR updates the UI of the URL widget. As discussed in ODK forum [here](https://forum.getodk.org/t/use-materialcardview-to-update-the-ui-of-url-widget/27500/6), the answer text view is removed and user simply sees the button to open the url in the widget. 

As discussed with @seadowg, this PR does not change the button text for the moment, so `Open Url` button is shown. As the user is not allowed to change the url, the widget is `readonly` all times, so I am removing any specific behavior concerning the `readonly` widget in this PR.

WhatsApp Image 2020-08-12 at 03 13 53

I changed the value fo the string resource `open_url` to `Visit Website`, and that change is not migrated to all the other languages of Collect, so this change won't appear good in when the user views the form in a language other than English. 

Any form having URL Widget like AllWidgetsForm.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -73,6 +73,12 @@ public class AnalyticsEvents {
     public static final String AUDIO_QUESTION = "Prompt";
 
     /**
+     * Track form definitions with url widget. The action should be the widget name and
+     * the label should be a hash of the form definition.
+     */
+    public static final String URL_WIDGET = "Prompt";
+
+    /**
      * Track scoped storage migration attempts. The action should be the result of the attempt.
      */
     public static final String SCOPED_STORAGE_MIGRATION = "ScopedStorageMigration";
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -67,16 +67,10 @@ public class AnalyticsEvents {
     public static final String SAVE_INCOMPLETE = "WidgetAttribute";
 
     /**
-     * Track displays of audio question types. The action should be the type of audio question and
+     * Track displays of widget/question types. The action should be the type of widget and
      * the label should be a hash of the form definition.
      */
-    public static final String AUDIO_QUESTION = "Prompt";
-
-    /**
-     * Track form definitions with url widget. The action should be the widget name and
-     * the label should be a hash of the form definition.
-     */
-    public static final String URL_WIDGET = "Prompt";
+    public static final String PROMPT = "Prompt";
 
     /**
      * Track scoped storage migration attempts. The action should be the result of the attempt.
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -43,13 +43,8 @@ public class UrlWidget extends QuestionWidget {
     @Override
     protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
         binding = UrlWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
-
-        if (prompt.isReadOnly()) {
-            binding.urlButton.setVisibility(GONE);
-        } else {
-            binding.urlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
-            binding.urlButton.setOnClickListener(v -> onButtonClick());
-        }
+        binding.urlButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, answerFontSize);
+        binding.urlButton.setOnClickListener(v -> onButtonClick());
 
         return binding.getRoot();
     }
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
@@ -20,7 +20,6 @@ import android.content.Context;
 import android.net.Uri;
 import android.util.TypedValue;
 import android.view.View;
-import android.widget.TextView;
 
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.StringData;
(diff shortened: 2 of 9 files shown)


seadowg:
> I still need to decide upon some issues regarding the behavior of URL widget. According to the current implementation, url button is not shown when the widget is readonly, but as the user is not allowed to change the answer anytime, so the widget is readonly at all times. I feel like not showing/disabling the url button when the url is empty.

My thinking here is that in the `readonly` case the widget should be no different as the "answer" is never changeable. Would be good to get second opinions from @grzesiek2010 and @lognaturel on that though.

seadowg:
Another point to make here is that it's been really hard to make calls about the URL widget as we don't have a clear picture of how it is used or even if it is used at all. I think as part of this change it would be nice to add an analytics call that gets made when a URL widget is included in a form. 

We could probably have the `WidgetFactory` make a call whenever it creates a `URLWidget`. I'd imagine this could use the "Prompt" category (that audio labels are using) and pass the form analytics ID (`QuestionDetails#getFormAnalyticsID()`) to make sure we can answer the question "how many forms contain URL widgets?" rather than just "how many times is a URL widget rendered?". @SaumiaSinghal could you look at adding that in as part of these changes?

collect_app/src/test/java/org/odk/collect/android/widgets/UrlWidgetTest.java
```diff
@@ -58,9 +59,9 @@ public void getAnswer_whenPromptHasAnswer_returnsAnswer() {
     }
 
     @Test
-    public void usingReadOnlyOption_makeAllClickableElementsDisabled() {
+    public void usingReadOnlyOption_doesNotShowUrlButton() {
         UrlWidget widget = createWidget(promptWithReadOnly());
-        assertThat(widget.getBinding().urlButton.getVisibility(), equalTo(View.GONE));
+        assertThat(widget.binding.urlButton.getVisibility(), equalTo(View.GONE));
```
seadowg: As discussed the URL widget should behave exactly the same in read only mode (as it is always "read only"). That being the case you shouldn't need any tests for read only for the widget.

collect_app/src/main/java/org/odk/collect/android/widgets/UrlWidget.java
```diff
@@ -44,7 +43,6 @@ public UrlWidget(Context context, QuestionDetails questionDetails, CustomTabHelp
     @Override
     protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
         binding = UrlWidgetAnswerBinding.inflate(((Activity) context).getLayoutInflater());
-        View answerView = binding.getRoot();
 
         if (prompt.isReadOnly()) {
```
seadowg: As in other comments: lets remove any change in behaviour based on `isReadOnly`

collect_app/src/main/res/values/strings.xml
```diff
@@ -217,7 +217,7 @@
     <string name="constraint_behavior_on_finalize">Defer validation until finalized</string>
     <string name="view_change_location">View or Change Location</string>
     <string name="change_location">Change Location</string>
-    <string name="open_url">Open Url</string>
+    <string name="open_url">Visit Website</string>
```
seadowg: Lets not change this text at the moment.

seadowg (changes requested):

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
```diff
@@ -72,6 +72,12 @@ private AnalyticsEvents() {
      */
     public static final String AUDIO_QUESTION = "Prompt";
 
+    /**
+     * Track form definitions with url widget. The action should be the widget name and
+     * the label should be a hash of the form definition.
+     */
+    public static final String URL_WIDGET = "Prompt";
```
seadowg: This doesn't seem right as it's the same as the `AUDIO_QUESTION`. I think the `AUDIO_QUESTION` naming and description is just wrong here and probably should be `PROMPT` and really be used to just track any kind of question/widget.

collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
```diff
@@ -122,6 +124,8 @@ public static QuestionWidget createWidgetFromPrompt(FormEntryPrompt prompt, Cont
                             questionWidget = new StringNumberWidget(context, questionDetails, readOnlyOverride);
                         } else if (appearance.equals(WidgetAppearanceUtils.URL)) {
                             questionWidget = new UrlWidget(context, questionDetails, new CustomTabHelper());
+
+                            analytics.logEvent(URL_WIDGET, "UrlWidget", questionDetails.getFormAnalyticsID());
```
seadowg: I think the action here should just be `"Url"`. What we care about is that the user saw a url question.

seadowg (changes requested):
Just a couple of small changes to go in!

seadowg (approved):

seadowg:
@lognaturel I think this would be nice to include in the release as it adds analytics. What do you think?

mmarciniak90:
@SaumiaSinghal please take a look at my test results. 



>I changed the value fo the string resource open_url to Visit Website

:red_circle: **Visit Website** text on the button is not visible

> I feel like not showing/disabling the url button when the url is empty.

:red_circle: I added a screenshot where url is empty and as you can see, button is visible and it's clickable. Toast with message appears

:large_blue_circle:  website address is not visible under the button -> the result is that the user has no idea whether the URL is set or not until the user clicks on button or open hierarchy view. It seems ok but it is worth paying attention to.
:green_circle: remove reponse dialog is not available for readonly url widget
:green_circle: readonly url widget has button and visit website is possible

SaumiaSinghal:
Hi @mmarciniak90! Thanks for testing the PR. I am sorry, I forgot to change the description of this PR. We decided not to change the button text for the moment, as we have to look into the string translations then. Also, we will not be showing the answer textview of the widget as mentioned in the header of the PR. Regarding the case when there is no answer set, I'll confirm about that case with my mentor and update the PR. Thanks!

seadowg:
I think for the no answer case we can just go with the current behaviour (that @mmarciniak90 is seeing). I think we're very unlikely to see that case in real life as it would be a mistake in the form and the current behaviour is actually pretty good for bubbling that mistake up to the user without interrupting the form.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.0, 7.1, 8.1, 9.0 and 10

Tested cases:
- light/dark mode
- read only
- open url
- minimize app
- screen rotation
- RTL language 
- try to remove answer

kkrawczyk123:

seadowg:
@SaumiaSinghal looks like we've got conflicts and failing checks.

SaumiaSinghal:
Done!

PR #4035, opened 2020-08-24
Handle NPE in DatabaseMediaFileRepository#getAll()

### LINKED ISSUE #4034: NPE in DatabaseMediaFileRepository.java line 15
Collect v1.28.0-beta

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/776021de55f48298e0fedd54d67f76c5?time=last-seven-days&sessionId=5F4244EC034300015F9A81F660321501_DNE_0_v2

```
Caused by java.lang.NullPointerException
       at java.io.File.<init>(File.java:283)
       at org.odk.collect.android.forms.DatabaseMediaFileRepository.getAll(DatabaseMediaFileRepository.java:15)
       at org.odk.collect.android.formmanagement.ServerFormsDetailsFetcher.areNewerMediaFilesAvailable(ServerFormsDetailsFetcher.java:122)
       at org.odk.collect.android.formmanagement.ServerFormsDetailsFetcher.fetchFormDetails(ServerFormsDetailsFetcher.java:72)
       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
```

grzesiek2010:
Closes #4034

The issue was caused by null `formMediaPath`. if that value is null and we call `newFile(formMediaPath)` NPE is thrown. Another solution would be to catch that exception but since we now what is the cause its' a better approach to just avoid it.

I think this doesn't require testing. A review from another contributor should be enough.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
+++ b/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
@@ -13,7 +13,10 @@ public class DatabaseMediaFileRepository implements MediaFileRepository {
     @Override
     public List<File> getAll(String jrFormID, String formVersion) {
         String formMediaPath = new FormsDao().getFormMediaPath(jrFormID, formVersion);
-        File[] files = new File(formMediaPath).listFiles();
+        File[] files = new File[0];
+        if (formMediaPath != null) {
+            files = new File(formMediaPath).listFiles();
+        }
         return asList(files);
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
+++ b/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
@@ -10,10 +10,15 @@ import java.util.List;
 import static java.util.Arrays.asList;
 
 public class DatabaseMediaFileRepository implements MediaFileRepository {
+    private final FormsDao formsDao;
+
+    public DatabaseMediaFileRepository(FormsDao formsDao) {
+        this.formsDao = formsDao;
+    }
 
     @Override
     public List<File> getAll(String jrFormID, String formVersion) {
-        String formMediaPath = new FormsDao().getFormMediaPath(jrFormID, formVersion);
+        String formMediaPath = formsDao.getFormMediaPath(jrFormID, formVersion);
         return formMediaPath == null
                 ? new ArrayList<>()
                 : asList(new File(formMediaPath).listFiles());
--- a/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
+++ b/collect_app/src/main/java/org/odk/collect/android/database/DatabaseMediaFileRepository.java
@@ -4,6 +4,7 @@ import org.odk.collect.android.dao.FormsDao;
 import org.odk.collect.android.forms.MediaFileRepository;
 
 import java.io.File;
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


grzesiek2010:
@seadowg 
I'm thinking about that class `DatabaseMediaFileRepository` you have created. Shouldn't that be one class with `DatabaseFormsRepository`? It uses the same database and at this moment contains just one method so what was the rationale for creating a separate repository class?

seadowg:
@grzesiek2010 I guess my thinking here was that the two different repositories are for fetching two distinct "entities": "forms" and "media files". I could see it being one repository though if you just think of the media files as a part of the form rather than a related to the form. My assumption is that some parts of the code care about the forms and not the media files and vice versa but that could totally be wrong! 

I guess if nothing else it feels like a nice split to me to avoid the interfaces getting too large.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseMediaFileRepository.java
```diff
@@ -12,7 +12,10 @@
     @Override
     public List<File> getAll(String jrFormID, String formVersion) {
         String formMediaPath = new FormsDao().getFormMediaPath(jrFormID, formVersion);
-        File[] files = new File(formMediaPath).listFiles();
+        File[] files = new File[0];
```
seadowg: I think we could avoid the array initialization here with an `if-else` on the `formMediaPath != null` and just returning `emptyList` in the `else` case. The `new File[0]` going into the `asList` confused me.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseMediaFileRepository.java
```diff
@@ -12,7 +12,10 @@
...
+        if (formMediaPath != null) {
```
seadowg: I think we need a test to drive this out so we avoid regressing at any point. As far as I can see the situation that gives us the NPE (for the previous code) was if `getAll` was queried for a `jrFormID` and `formVersion` not in the database. If that's right we could have a `DatabaseMediaFileRepositoryTest` similar to the `DatabaseFormRepositoryTest` and a test that just checks we get back an empty list when there is nothing in the database.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseMediaFileRepository.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/forms/DatabaseMediaFileRepository.java
grzesiek2010: Done.

grzesiek2010:
> I guess my thinking here was that the two different repositories are for fetching two distinct "entities": "forms" and "media files".

I'm not sure about it but we can leave it as is for now and we will see how it evolves.

seadowg (approved):

seadowg:
@grzesiek2010 looks like the last PR merged added a conflict but this looks good to go!

seadowg:
@grzesiek2010 oh sorry you'd pointed out we could probably skip QA on this one and I think I agree. If you fix the conflicts I'll merge!

grzesiek2010:
> If you fix the conflicts I'll merge!

fixed.

PR #4028, opened 2020-08-15
Make sure soft-deleted form can be re-downloaded

seadowg:
Closes #4026

#### What has been done to verify that this works as intended?

Manually verified and wrote new test.

#### Why is this the best possible solution? Were any other approaches considered?

I'm not a fan of testing this at an Espresso level. I started at writing tests at a lower level but it would mean a bit of a complicated dance around the `MultiFormDownloader` and `ServerFormDownloader`. I think this would get everything to a better place and potentially mean killing the `MultiFormDownloader` which would be nice. I'm going to continue that work immediately but didn't want to hold the regression fix up or add more risk to it.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Good to just check the issue is fixed and that form download still seems to be working!

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4040, opened 2020-08-24
Isolate MultiFormDownloader

### LINKED ISSUE #1698: Reuse manifest download/parse code between DownloadFormListTask and DownloadFormsTask
#1653 introduces checks to see whether form media has changed when forms are listed. This duplicates code from the `downloadManifestAndMediaFiles` in the `DownloadFormsTask` class.

### LINKED ISSUE #3538: Rework `FormDownloader`

seadowg:
Closes #1698
Closes #3538

~~Blocked by #4028.~~

The goal of this work was to replace any calls to `MultiFormDownloader` with calls to the new `FormDownloader` interface. This lets other places in the code mock or fake the interface, not having to worry about it's details. As part of this rework we wanted to have more testing around the implementation of `FormDownloader` (`ServerFormDownloader`) so that we can have more confidence mocking/fake it elsewhere, have a better understanding of what it actually does and move test coverage of some of the more detailed business logic to a unit test.

I've also moved code to three new packages created outside the `android` package: `forms`, `openrosa` and `server`. The code here is non-Android specific and should not have any dependencies on our "app" code. I'd like to move these out to submodules but will avoid that for the moment to prevent even more noise in the PR. The change was motivated by "auditing" what parts of the code I was playing with had Android dependencies (and didn't need Robolectric) and what parts didn't.

New tests and played around with the app.

Comments inline.

One change here is that we now always fetch the manifest for forms when checking for updates instead of sometimes avoiding it. This isn't actually a big problem as far as I can see: In the old code we'd avoid a call to fetch the manifest only in the cases where a form wasn't on the device or when the form XML had been updated. Given that these will be very rare compared to the case where nothing has been updated there isn't a big change here.


collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -339,13 +354,13 @@ FileResult downloadXform(String formName, String url, FormDownloaderListener sta
...
          * is okay, so that garbage is not left over on cancel.
          */
-        private void writeFile(File file, FormDownloaderListener stateListener, InputStream inputStream)
+        private void writeFile(File file, FormDownloaderListener stateListener, InputStream inputStream, File tempDir)
                 throws IOException, InterruptedException {
 
             File tempFile = File.createTempFile(
                     file.getName(),
                     TEMP_DOWNLOAD_EXTENSION,
-                    new File(storagePathProvider.getDirPath(StorageSubdirectory.CACHE))
```
lognaturel: @seadowg and I briefly talked about whether the use of the ODK cache directory could have any significance. All files are eventually cleaned up and I don't see any reason why any particular directory has to be used. Passing in the base cache directory, whatever it is, is certainly an improvement.

Am I understanding correctly that you've kept using `storagePathProvider.getDirPath(StorageSubdirectory.CACHE)` but create an enclosing directory for each form's files so that can be deleted on cancel rather than having to touch individual files? That seems fine, as would using the Android scoped cache directory.

seadowg: Yeah that's right! I think going one step further and use the Android cache directory (directly) in the dependency declaration (`AppDependencyModule`) would be nice as well.

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           path: collect_app/build/reports
           destination: reports
 
-  test_unit:
+  test_modules:
```
seadowg, line 40: This new job allows us to run all the submodule tests in parallel to the app tests and the quality checks which as we move more things out to submodules will hopefully keep our test CI time under control

collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
```diff
@@ -63,36 +63,6 @@ public void deletingAForm_whenThereFilledForms_removesFormFromBlankFormList_butA
                 .clickSaveAndExit();
     }
 
-    @Test
-    public void afterFillingAForm_andDeletingIt_allowsFormToBeReDownloaded() {
```
seadowg: This is now covered in the `ServerFormDownloaderTest`

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMetadataParser.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.File;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.odk.collect.android.utilities.FileUtils.LAST_SAVED_FILENAME;
+import static org.odk.collect.android.utilities.FileUtils.STUB_XML;
+import static org.odk.collect.android.utilities.FileUtils.write;
+
+public class FormMetadataParser {
```
seadowg, line 17: This was begging to be pulled out

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDetails.java
```diff
@@ -28,22 +28,22 @@
...
-    private final String manifestFileHash;
     private final boolean isNotOnDevice;
     private final boolean isUpdated;
+    private final ManifestFile manifest;
 
     public ServerFormDetails(String formName, String downloadUrl, String manifestUrl, String formID,
-                             String formVersion, String hash, String manifestFileHash,
-                             boolean isNotOnDevice, boolean isUpdated) {
+                             String formVersion, String hash,
+                             boolean isNotOnDevice, boolean isUpdated, ManifestFile manifest) {
```
seadowg, line 37: We now just always include the manifest as part of the form details so logic that deals with it can be simpler (and doesn't end up fetching the manifest when it doesn't need to).

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -1,30 +1,538 @@
...
+import static org.odk.collect.utilities.PathUtils.getAbsoluteFilePath;
 
 /**
- * Provides a sarcophagus for {@link org.odk.collect.android.utilities.MultiFormDownloader} so it
+ * Provides a sarcophagus for {@link MultiFormDownloader} so it
  * can eventually be disposed of.
  */
 public class ServerFormDownloader implements FormDownloader {
 
     private final MultiFormDownloader multiFormDownloader;
```
seadowg, line 37: The `MultiFormDownloader` lives on in here as pulling it apart entirely is going to be pretty time-consuming and we've already hit the core goal of getting rid of any references to it in the rest of the code.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
```diff
@@ -96,10 +99,6 @@ public String getFormDbPath(String filePath) {
         return getDbPath(getDirPath(StorageSubdirectory.FORMS), filePath);
     }
 
-    public String getAbsoluteFormFilePath(String filePath) {
```
seadowg, line 99: It was clear when working with `ServerFormDownloader` that this part of the interface feels like too much as the provider becomes responsible for calculations that are based on it's state but have logic (calculating a relative or absolute path) that can live separately. All the extra helpers increase the coupling on whatever object is using it.

config/quality.gradle
```diff
@@ -46,18 +46,4 @@ tasks.register("pmd", Pmd) {
             setDestination new File("$reportsDir/pmd/pmd.html")
         }
     }
-}
-
-//------------------------Lint------------------------//
-
-android {
```
seadowg, line 53: This is moved into the core `collect_app` module

collect_app/src/main/java/org/odk/collect/android/preferences/FormUpdateMode.java
```diff
@@ -1,4 +1,4 @@
-package org.odk.collect.android.formmanagement;
+package org.odk.collect.android.preferences;
```
lognaturel, line 1: What made you want this package change? It's kind of a weird one because it's for the preferences part of the formmanagement feature. Curious to know how you make that decision as we move more towards feature-based packages.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
```diff
@@ -71,13 +70,13 @@ public void updateFormListApi(String url, WebCredentialsUtils webCredentialsUtil
...
             if (thisFormAlreadyDownloaded) {
-                isNewerFormVersionAvailable = isNewerFormVersionAvailable(getMd5HashWithoutPrefix(listItem.getHashWithPrefix()));
+                isNewerFormVersionAvailable = isNewerFormVersionAvailable(listItem);
 
                 if (manifestFile != null) {
                     List<MediaFile> newMediaFiles = manifestFile.getMediaFiles();
 
                     if (newMediaFiles != null && !newMediaFiles.isEmpty()) {
-                        areNewerMediaFilesAvailable = areNewerMediaFilesAvailable(listItem.getFormID(), listItem.getVersion(), newMediaFiles);
+                        isNewerFormVersionAvailable = areNewerMediaFilesAvailable(listItem.getFormID(), listItem.getVersion(), newMediaFiles);
```
lognaturel, line 80: I think this changes behavior. Previously, the presence of media updates was only checked if we didn't already know new XML was available. Now, if media files are available, we ignore whether or not new XML is available and only do an update if new media is available. I think the enclosing condition can be changed to `!isNewerFormVersionAvailable && manifestFile != null` -- we only care to check media for updates if we already know there isn't new XML.

```java
    @Test
    public void whenAFormExists_andIsUpdatedOnServer_andDoesNotHaveNewMedia_isUpdated() throws Exception {
        formsRepository.save(new Form.Builder()
                .id(2L)
                .jrFormId("form-2")
                .md5Hash("form-2-hash-old")
                .build());

        File localMediaFile = File.createTempFile("blah", ".csv");
        writeToFile(localMediaFile, "blah");
        when(mediaFileRepository.getAll("form-2", "server")).thenReturn(asList(localMediaFile));

        List<ServerFormDetails> serverFormDetails = fetcher.fetchFormDetails();
        assertThat(serverFormDetails.get(1).isUpdated(), is(true));
    }
```

collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
```diff
@@ -67,7 +67,7 @@ public DownloadFormsTask(FormDownloader formDownloader) {
 
                 results.put(serverFormDetails, Collect.getInstance().getString(R.string.success));
             } catch (FormDownloadException e) {
-                results.put(serverFormDetails, e.getMessage());
+                results.put(serverFormDetails, Collect.getInstance().getString(R.string.failure));
```
lognaturel, line 73: Just showing "Failure" rather than any more specific message was already part of the match exactly changes, right? So this doesn't change behavior?

config/quality.gradle
lognaturel: Doesn't this mean lint won't be run on the submodules? Shouldn't it?

collect_app/src/main/java/org/odk/collect/utilities/PathUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.File;
+
+public class PathUtils {
+
+    private PathUtils() {
+
+    }
+
+    public static String getRelativeFilePath(String dirPath, String filePath) {
```
lognaturel, line 11: Really, I think we should be using `Path`s and [`relativize`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#relativize(java.nio.file.Path)). For some other time.

collect_app/src/main/java/org/odk/collect/utilities/PathUtils.java
```diff
@@ -0,0 +1,25 @@
...
+        return filePath.startsWith(dirPath)
+                ? filePath.substring(dirPath.length() + 1)
+                : filePath;
+    }
+
+    public static String getAbsoluteFilePath(String dirPath, String filePath) {
```
lognaturel, line 17: [`Path.resolve`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.nio.file.Path))

collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
```diff
@@ -33,32 +38,51 @@
  * @author carlhartung
  */
 public class DownloadFormsTask extends
-        AsyncTask<ArrayList<ServerFormDetails>, String, HashMap<ServerFormDetails, String>> implements FormDownloaderListener {
+        AsyncTask<ArrayList<ServerFormDetails>, String, Map<ServerFormDetails, String>> {
```
lognaturel, line 41: Somehow I haven't noticed this signature. Why is the first param `ArrayList<ServerFormDetails>` rather than just `ServerFormDetails`? Feels unusual that `doInBackground` needs to access the 0th element. Did this change at some point or has it always been like this?

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
```diff
@@ -77,8 +80,18 @@
...
-                                notifier.onUpdatesDownloaded(result);
+                                HashMap<ServerFormDetails, String> results = new HashMap<>();
+                                for (ServerFormDetails serverFormDetails : updatedForms) {
+                                    try {
+                                        formDownloader.downloadForm(serverFormDetails, null, null);
+                                        results.put(serverFormDetails, Collect.getInstance().getString(R.string.success));
+                                    } catch (FormDownloadException e) {
+                                        results.put(serverFormDetails, e.getMessage());
+                                    } catch (InterruptedException e) {
+                                        return null;
```
lognaturel: This is not actually possible through the UI, is it? If it is, shouldn't the status for forms that did download be updated?

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
```diff
@@ -77,8 +80,18 @@
...
                             if (acquiredLock) {
-                                final HashMap<ServerFormDetails, String> result = multiFormDownloader.downloadForms(updatedForms, null);
-                                notifier.onUpdatesDownloaded(result);
+                                HashMap<ServerFormDetails, String> results = new HashMap<>();
+                                for (ServerFormDetails serverFormDetails : updatedForms) {
+                                    try {
+                                        formDownloader.downloadForm(serverFormDetails, null, null);
+                                        results.put(serverFormDetails, Collect.getInstance().getString(R.string.success));
+                                    } catch (FormDownloadException e) {
+                                        results.put(serverFormDetails, e.getMessage());
```
lognaturel: There will never be a message for this. I believe the only place that exception is thrown is at https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-8ae847a7cbae59b9a61bd5868b5c6514R65. Is the client robust to nulls? Should this maybe be the failure message?

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -1,31 +1,532 @@
...
+
+        @Override
+        public void progressUpdate(String currentFile, String progress, String total) {
+            if (progressReporter != null) {
+                progressReporter.onDownloadingMediaFile(Integer.parseInt(progress));
+            }
+        }
+
+        @Override
+        public boolean isTaskCanceled() {
```
lognaturel: `Canceled` -> `Cancelled`

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
```diff
@@ -0,0 +1,349 @@
...
+                true,
+                false,
+                null);
+
+        CancelAfterFormDownloadFormListApi formListApi = new CancelAfterFormDownloadFormListApi(xform);
+        ServerFormDownloader downloader = new ServerFormDownloader(formListApi, formsRepository, cacheDir, formsDir.getAbsolutePath());
+
+        try {
+            downloader.downloadForm(serverFormDetails, null, formListApi);
+            fail("Excepted exception");
```
lognaturel: Expected?

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
```diff
@@ -0,0 +1,349 @@
...
+            assertThat(formsRepository.getAll(), is(empty()));
+            assertThat(asList(new File(getCacheFilesPath()).listFiles()), is(empty()));
+
+            // The media directory is created early for some reason
+            assertThat(asList(new File(getFormFilesPath()).listFiles()), contains(new File(getFormFilesPath() + "/Form-media")));
+        }
+    }
+
+    @Test
+    public void beforeDownloadingMediaFile_reportsProgress() throws Exception {
```
lognaturel: Add test for media file progress?

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
```diff
@@ -1,414 +0,0 @@
...
-    }
-
-    /**
-     * Forms with references to external secondary instance need to have the secondary instance
-     * available at time of form parse.
-     *
-     * See https://github.com/getodk/collect/issues/3635
-     */
-    @Test
-    public void downloadingFormWithXmlExternalSecondaryInstance_Succeeds() throws Exception {
```
lognaturel, line 159: I think this case should get coverage somewhere. I think the CSV case can be skipped but the last-saved one should also be captured.

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
```diff
@@ -1,414 +0,0 @@
...
-     *
-     * This is an unfortunate side effect of using the form media folder to store the contents that
-     * jr://instance/last-saved resolves to.
-     *
-     * Additionally, immediately replacing a secondary instance with name last-saved.xml avoid users
-     * exploiting this current implementation quirk as a feature to preload defaults for the first
-     * instance.
-     * */
-    @Test
-    public void downloadingFormWithExternalSecondaryInstanceNamedLastSavedXml_Succeeds() throws Exception {
```
lognaturel, line 324: I think this edge case is also important to capture somewhere.

lognaturel (changes requested):
> This isn't actually a big problem as far as I can see: In the old code we'd avoid a call to fetch the manifest only in the cases where a form wasn't on the device or when the form XML had been updated.

Let's say I have access to 52 forms on my server, all with media, and I have just one downloaded to my device. Previously, a background form update would have involved up to 2 requests (form list and manifest for the 1 form on device) and now it would would involve up to 53 (form list and manifest for each form), right? We do see from analytics that folks can have access to a lot of forms. It's unlikely all have media but this could increase the number of requests made significantly if I'm understanding correctly.

I looked primarily at:
- [AutoUpdateTaskSpec](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-29f9504e77d644c683aeaedf712d3654)
- [FormMetadataParser](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-7ce5b83a70d6de41807e2c70a98a3276)
- [ServerFormDetails](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-33d5c16848c88eaefc41e7ccd05ce2ce)
- [ServerFormDownloader](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-8ae847a7cbae59b9a61bd5868b5c6514)
- [ServerFormDetailsFetcher](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-74ea94be69d987bd937bf4f13879aa83)
- [DownloadFormTask](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-a0a2d90fbe75ec05be821f16c24e74c3)
- [ServerFormDownloaderTest](https://github.com/getodk/collect/pull/4040/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-4f10ae6e96b8c8ffe460db9c26cbe8d7)

Can you please say a little bit about the motivation for putting things that were in `openrosa/api` into a separate `server` package? Everything in there seems to be very much tied to the openrosa API.

Should we perhaps capture the use of packages outside of `android` somewhere? Maybe the state doc? This starts the new pattern but now there are inconsistencies like many of the `android.utilities` classes could really be outside of `android`. Side note, `org.odk.collect.android` is the app ID. I know the app ID and the package names are independent but wanted to point out that this is presumably where the `android` package came from, not from the classes it contains having Android dependencies.

seadowg:
Another amazing and detailed PR review from @lognaturel. I felt really bad about how big this got and I'm sorry I couldn't break it into more digestible chunks. Thanks for managing to get through it and provide a useful critique!

> Let's say I have access to 52 forms on my server, all with media, and I have just one downloaded to my device. Previously, a background form update would have involved up to 2 requests (form list and manifest for the 1 form on device) and now it would would involve up to 53 (form list and manifest for each form), right? We do see from analytics that folks can have access to a lot of forms. It's unlikely all have media but this could increase the number of requests made significantly if I'm understanding correctly.

Urgh yeah. As I was writing this up I had a feeling you'd have a good use case to counter that change! I'm tempted to say we could merge this in and then fix that as a follow up? That would cut down on the review/QA window for this PR. I had originally gone down a path of keeping that optimization in so have some ideas of a nice(ish) way forward.

> Can you please say a little bit about the motivation for putting things that were in openrosa/api into a separate server package? Everything in there seems to be very much tied to the openrosa API.

My thinking here is that as much as the `FormListApi` looks a lot like Open Rosa we could definitely get to a point (and I think we should) where that interface has disk based and Google Drive implementations so our method object/use case objects could be shared between the different form synchronization, update etc flows. I think you're right that this thinking is completely unclear though. Maybe some more comments on the interfaces/data classes? Also, a `STATE.md` update would be good.

> Should we perhaps capture the use of packages outside of android somewhere? Maybe the state doc? This starts the new pattern but now there are inconsistencies like many of the android.utilities classes could really be outside of android. Side note, org.odk.collect.android is the app ID. I know the app ID and the package names are independent but wanted to point out that this is presumably where the android package came from, not from the classes it contains having Android dependencies.

I'm thinking I shouldn't have skipped on just pulling them out as submodules. That hardwires the intended separation and direction of dependencies in. What do you think?

collect_app/src/main/java/org/odk/collect/android/preferences/FormUpdateMode.java
seadowg: I think I bounced around on this a few times. For me it was that the actual setting of the mode is a settings/preference concern and then other parts of the app ask about it. I'm starting to think about settings as its own section of the app that provides config for everything else. In that way (in my head) it makes sense that it owns the domain objects/data models associated with it. Does that make sense?

collect_app/src/main/java/org/odk/collect/utilities/PathUtils.java
seadowg: huh. Yeah probably!

collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
seadowg: Ah it's weirdness with `AsyncTask` where `execute` takes a varargs of `Params`. Generally to avoid how consfusing this ends up looking I'd leave `Params` as `Void` and pass anything I need into the constructor (as you can't reuse an `AsyncTask` anyway). I think we can leave this for the moment as it's existing and ideally we want to just get rid of the `AsyncTask` rather than improve it.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
seadowg: I'm pretty sure I ended up not needing another test to drive that out (I basically used commenting/deleting to work out adding tests). You're right that from a documentation and regression standpoint though it would be nice to have the test.

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
seadowg: Same as the media file progress test I think. Didn't need it but I think you're right we should add it in any way.

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
seadowg: Same as the media file progress test I think. Didn't need it but I think you're right we should add it in any way.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
seadowg: Ah nice. So many points for communicating the problem with a failing test.

collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
seadowg: Yeah I think this might be a mistake I made during a rebase... Will fix.

seadowg: No wait sorry this looks right I read the diff in reverse.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
seadowg: Yeah. This should really just be a `break` I think?

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
seadowg: Yeah I think this should just be the failure message.

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
lognaturel: It's not the same. Only external secondary instances will actually use the `ReferenceManager` configured by `FormMetadataParser`. It's really easy for code there to get removed or modified (as it had been before).

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
lognaturel: The comment is the most important part, I think. But I think it's also important to have a test around which `last-saved` external secondary instance gets used in that specific case.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
lognaturel: Yes, `break` sounds right.

collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java
seadowg: Sorry I mean "same story", not "same code".

lognaturel:
> Thanks for managing to get through it and provide a useful critique!

Of course! I went through the commits really quickly and then the "mark file as viewed" Github feature was particularly helpful for this one. I was able to get to just that handful of files with minimal effort.

> could merge this in and then fix that as a follow up?

Definitely. The case I described is definitely not very common but I still think it's worth addressing.

> we could definitely get to a point (and I think we should) where that interface has disk based and Google Drive implementations

That makes sense but the interfaces would be different. That is, there's no manifest or web credentials in the Google Drive world, e.g. So the `ManifestFile` class is very much in the OpenRosa world, and all the other classes in the `server` package refer to manifests in some way. It feels premature to me and hard to explain but I don't feel strongly about it.

> just pulling them out as submodules

Having an `openrosa` submodule feels like it makes sense. I'm less convinced about `forms` and especially `utilities` because those feel kind of like implementation details. Are there any downsides to having a bunch of submodules?

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
seadowg: Wait looking back at this I'm not sure what test you think is missing here? I thought I was missing media file progress but I think this test is just badly named as it covers that?

seadowg:
@lognaturel oh and on linting: you'll see if you run `./ gradlew lintDebug` on this branch that all the android submodules have `lintDebug` run on them. As we don't specific a submodule it just runs the task for everything that has it I believe.

I moved the lint config around so we wound't have `collect_app` specific setup in the other submodules.

seadowg:
Issue around optimization created: https://github.com/getodk/collect/issues/4161

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+        FormMetadataParser formMetadataParser = new FormMetadataParser(referenceManager);
+        formMetadataParser.parse(formXml, mediaDir);
+
+        assertThat(mediaDir.listFiles().length, is(0));
+    }
+
+    @Test
+    @SuppressWarnings("PMD.JUnitUseExpected")
+    public void cleansUpReferenceManager() throws Exception {
```
lognaturel, line 107: YES. Hadn't thought of this one. Great case to have.

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -64,6 +65,32 @@ public void canParseFormWithLastSaved() throws Exception {
...
+     * is ok if they only try launching the form once.
+     *
+     * This is an unfortunate side effect of using the form media folder to store the contents that
+     * jr://instance/last-saved resolves to.
+     *
+     * Additionally, immediately replacing a secondary instance with name last-saved.xml avoid users
+     * exploiting this current implementation quirk as a feature to preload defaults for the first
+     * instance.
+     * */
+    @Test(expected = XFormParseException.class)
```
lognaturel, line 82: That's not the behavior I'd expect! The media file should be ignored and the last-saved instance should be used.

lognaturel (changes requested):
Nice. There's just one test I'm not sure about.

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
seadowg: If you look at the [previous test](https://github.com/getodk/collect/blob/c4bddb406a33159fcdfde78fac75ebf0001941e5/collect_app/src/test/java/org/odk/collect/android/utilities/MultiFormDownloaderTest.java#L324) does this not match that behaviour? The test name agrees with you but the test itself doesn't as far as I can see (it ends with `assertThat(messages.get(serverFormDetails), containsString("Failure"))`)

lognaturel: Oh, that's confusing, I'm sorry! Because the external file named `last-saved.xml` and the primary instance have different shapes AND because the references for a dynamic select get validated at parse time, the parse crashes. So if the form definition had a different kind of query on the instance, there might not be a crash. Phew, that's clever/diabolical. You could add another comment above the `parse` call about how it's going to crash because the `last-saved` virtual instance doesn't have `/root/item/name` and `/root/item/label` paths to avoid going through this again. But maybe it's just confusing no matter what.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -101,7 +101,6 @@
 import org.odk.collect.android.formentry.repeats.DeleteRepeatDialogFragment;
 import org.odk.collect.android.formentry.saving.FormSaveViewModel;
 import org.odk.collect.android.formentry.saving.SaveFormProgressDialogFragment;
-import org.odk.collect.android.forms.FormsRepository;
```
lognaturel: Looks like your IntelliJ settings aren't set to alphabetize imports. Would be great to revert changes to files where this is the only change so they don't get touched by this PR.

lognaturel:
I caught some regressions in review around form updates and forms with external secondary instances. For form updates, the issue had to do with if the form definition was updated but not the media so doing some checks around the different combinations there would be wise to make sure we have coverage for all of it.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Looks like that's default it's just that Studio doesn't do that when you move code between packages. Optimizing imports fixes it. Annoying!

mmarciniak90:
@seadowg I see a crash when I download some form on Android 7. @kkrawczyk123 please try to reproduce it on Android 7.1

```java
2020-10-13 16:20:42.933 26685-26981/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #3
    Process: org.odk.collect.android, PID: 26685
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:318)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
        at java.util.concurrent.FutureTask.run(FutureTask.java:242)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:762)
     Caused by: java.lang.NoSuchMethodError: No virtual method toPath()Ljava/nio/file/Path; in class Ljava/io/File; or its super classes (declaration of 'java.io.File' appears in /system/framework/core-oj.jar)
        at org.apache.commons.io.FileUtils.isSymlink(FileUtils.java:1967)
        at org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1205)
        at org.odk.collect.android.formmanagement.ServerFormDownloader.downloadForm(ServerFormDownloader.java:70)
        at org.odk.collect.android.tasks.DownloadFormsTask.doInBackground(DownloadFormsTask.java:61)
        at org.odk.collect.android.tasks.DownloadFormsTask.doInBackground(DownloadFormsTask.java:40)
        at android.os.AsyncTask$2.call(AsyncTask.java:304)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243) 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607) 
        at java.lang.Thread.run(Thread.java:762) 
2020-10-13 16:20:42.997 26685-26695/org.odk.collect.android E/StrictMode: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
    java.lang.Throwable: Explicit termination method 'close' not called
        at dalvik.system.CloseGuard.open(CloseGuard.java:180)
        at android.content.ContentResolver$CursorWrapperInner.<init>(ContentResolver.java:2713)
        at android.content.ContentResolver.query(ContentResolver.java:562)
        at android.content.ContentResolver.query(ContentResolver.java:478)
        at org.odk.collect.android.dao.FormsDao.getFormsCursor(FormsDao.java:53)
        at org.odk.collect.android.dao.FormsDao.getFormsCursorForFormFilePath(FormsDao.java:181)
        at org.odk.collect.android.database.DatabaseFormsRepository.getByPath(DatabaseFormsRepository.java:80)
        at org.odk.collect.android.formmanagement.ServerFormDownloader$MultiFormDownloader.findExistingOrCreateNewUri(ServerFormDownloader.java:257)
        at org.odk.collect.android.formmanagement.ServerFormDownloader$MultiFormDownloader.installEverything(ServerFormDownloader.java:201)
        at org.odk.collect.android.formmanagement.ServerFormDownloader$MultiFormDownloader.processOneForm(ServerFormDownloader.java:180)
        at org.odk.collect.android.formmanagement.ServerFormDownloader.downloadForm(ServerFormDownloader.java:63)
        at org.odk.collect.android.tasks.DownloadFormsTask.doInBackground(DownloadFormsTask.java:61)
        at org.odk.collect.android.tasks.DownloadFormsTask.doInBackground(DownloadFormsTask.java:40)
        at android.os.AsyncTask$2.call(AsyncTask.java:304)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:762)

```

I will continue testing on higher Android versions

EDIT: add stacktrace to message

kkrawczyk123:
@seadowg @mmarciniak90 I am also able to reproduce the crash on Android 7.1 and 5.1.

seadowg:
@mmarciniak90 @kkrawczyk123 I'll have a look into the crash

seadowg:
@mmarciniak90 @kkrawczyk123 ok that's fixed. Looks like some updates to our dependencies brought in some code that wasn't compatible with Android which is a little scary.

One positive is that I was able to reproduce by running `GetBlankFormTest` against API 24. For big changes like this I'll try and remember to run the `smoke` tests on a few API levels.

seadowg:
@lognaturel @grzesiek2010 I actually had to downgrade commons to 2.5 for this PR (rather than 2.6 which was on `master`)

mmarciniak90:
@lognaturel @seadowg I spent time on this issue and I described all verified cases. Let me know if you have more ideas.

Compared with the current master. 

I didn't notice different behavior in cases:
- downloading forms from Aggregate
- downloading forms from Central
- downloading forms from Google Drive
- downloading forms from KOBO
- downloading forms from ONA
- downloading forms with media
- form designation with the new version to download
- `Exactly match server` 
   * verified on Central and Aggregate
   * refresh button
   * auto-refresh after 15 minutes
   * failed connection when required credentials are not set
   * successful connection when required credentials are set
   * successful connection as an anonymous user
- `Previously downloaded forms only`
   * manual download the latest form version
   * auto-download 
   * the latest version is auto-downloaded even if the first notification is ignored
- delete downloaded forms from GD and download them again when submission is still on the device
- Central - removed access to download form
- Aggregate - removed access to download form and send submission
- downloading forms via an external app
- auto-send to Aggregate
- auto-send to Central
- auto-send to Google Sheet

seadowg:
@mmarciniak90 looks good to me. Anything you think is blocking this from a `behaviour verified`?

mmarciniak90:
> Anything you think is blocking this from a behaviour verified?

No, I just wanted your opinion about verified test cases.

PR #4042, opened 2020-08-25
Rework Bearing Widget

SaumiaSinghal:
This PR includes the work on rewriting the framework of Bearing widget.

I don't think there are any regression risks as of now.

Any form having Bearing widget like AllWidgetsForm.

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
@@ -32,7 +32,6 @@ import org.odk.collect.android.activities.BearingActivity;
 import org.odk.collect.android.databinding.BearingWidgetAnswerBinding;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.ToastUtils;
-import org.odk.collect.android.utilities.UiUtils;
 import org.odk.collect.android.widgets.interfaces.BinaryDataReceiver;
 import org.odk.collect.android.widgets.utilities.WaitingForDataRegistry;
 
@@ -117,18 +116,13 @@ public class BearingWidget extends QuestionWidget implements BinaryDataReceiver 
     private void onButtonClick() {
         if (areSensorsAvailable()) {
             Intent intent = new Intent(getContext(), BearingActivity.class);
+            ((Activity) getContext()).startActivityForResult(intent, RequestCodes.BEARING_CAPTURE);
 
             waitingForDataRegistry.waitForData(getFormEntryPrompt().getIndex());
-            ((Activity) getContext()).startActivityForResult(intent, RequestCodes.BEARING_CAPTURE);
         } else {
             ToastUtils.showLongToast(R.string.bearing_lack_of_sensors);
 
             binding.bearingButton.setEnabled(false);
-
-            MarginLayoutParams marginLayoutParams = (MarginLayoutParams) binding.answerText.getLayoutParams();
-            marginLayoutParams.setMarginStart(UiUtils.getDimen(R.dimen.margin_standard));
-            marginLayoutParams.setMarginEnd(UiUtils.getDimen(R.dimen.margin_standard));
-            binding.answerText.setLayoutParams(marginLayoutParams);
 
             binding.answerText.setBackground(new EditText(getContext()).getBackground());
             binding.answerText.setFocusable(true);
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
@@ -21,9 +21,7 @@ import android.content.Intent;
 import android.graphics.drawable.Drawable;
 import android.hardware.Sensor;
 import android.hardware.SensorManager;
-import android.text.Editable;
 import android.text.InputType;
-import android.text.TextWatcher;
 import android.text.method.DigitsKeyListener;
 import android.util.TypedValue;
 import android.widget.Button;
@@ -57,10 +55,12 @@ public class BearingWidget extends QuestionWidget implements BinaryDataReceiver,
     final EditText answer;
     private final Drawable textBackground;
     private final WaitingForDataRegistry waitingForDataRegistry;
+    private final SensorManager sensorManager;
 
-    public BearingWidget(Context context, QuestionDetails questionDetails, WaitingForDataRegistry waitingForDataRegistry) {
+    public BearingWidget(Context context, QuestionDetails questionDetails, WaitingForDataRegistry waitingForDataRegistry, SensorManager sensorManager) {
         super(context, questionDetails);
         this.waitingForDataRegistry = waitingForDataRegistry;
+        this.sensorManager = sensorManager;
 
         isSensorAvailable = checkForRequiredSensors();
 
@@ -79,9 +79,6 @@ public class BearingWidget extends QuestionWidget implements BinaryDataReceiver,
         String s = questionDetails.getPrompt().getAnswerText();
         if (s != null && !s.equals("")) {
             getBearingButton.setText(getContext().getString(R.string.replace_bearing));
-            if (!isSensorAvailable) {
-                answer.setText(s);
-            }
             setBinaryData(s);
         }
         addAnswerView(answerLayout, WidgetViewUtils.getStandardMargin(context));
@@ -130,18 +127,15 @@ public class BearingWidget extends QuestionWidget implements BinaryDataReceiver,
     }
 
     private boolean checkForRequiredSensors() {
-
         boolean isAccelerometerSensorAvailable = false;
         boolean isMagneticFieldSensorAvailable = false;
 
-        SensorManager sensorManager = (SensorManager) getContext().getSystemService(Context.SENSOR_SERVICE);
         if (sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER) != null) {
             isAccelerometerSensorAvailable = true;
         }
         if (sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD) != null) {
             isMagneticFieldSensorAvailable = true;
         }
-
         return isAccelerometerSensorAvailable && isMagneticFieldSensorAvailable;
     }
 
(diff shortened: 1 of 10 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
 
-        LinearLayout answerLayout = new LinearLayout(getContext());
-        answerLayout.setOrientation(LinearLayout.VERTICAL);
-
-        answer = getEditText();
-        textBackground = answer.getBackground();
-        answer.setBackground(null);
-
-        getBearingButton = createSimpleButton(getContext(), getFormEntryPrompt().isReadOnly(), getContext().getString(R.string.get_bearing), getAnswerFontSize(), this);
+        String answerText = questionDetails.getPrompt().getAnswerText();
```
grzesiek2010: As I remember in other widgets you have refactored you set answer in `onCreateAnswerView()`, am I right? If so please be consequent.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
+        return binding.getRoot();
     }
 
     @Override
     public void clearAnswer() {
-        answer.setText(null);
-        getBearingButton.setText(getContext().getString(R.string.get_bearing));
-
+        binding.answerText.setText(null);
+        binding.bearingButton.setText(getContext().getString(R.string.get_bearing));
```
grzesiek2010, line 81: There is something wrong with the button text. You set it to default here but it's not set to `R.string.replace_bearing` when we record bearing (only if we create the widget from scratch and there is an answer).

collect_app/src/main/res/values/buttons.xml
```diff
@@ -6,6 +6,15 @@
         <item name="android:textStyle">normal</item>
     </style>
 
+    <style name="Widget.Collect.Button.WidgetAnswer" parent="Widget.Collect.Button.Custom">
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">wrap_content</item>
+        <item name="android:padding">@dimen/margin_small</item>
+        <item name="android:layout_marginTop">@dimen/margin_extra_small</item>
+        <item name="android:layout_marginStart">@dimen/margin_standard</item>
```
grzesiek2010: I think those start/end margins should be used in the parent layout then you don't need to duplicate it for each child. For example it looks bad for the edittext:
image

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
     private final WaitingForDataRegistry waitingForDataRegistry;
+    private final SensorManager sensorManager;
 
-    public BearingWidget(Context context, QuestionDetails questionDetails, WaitingForDataRegistry waitingForDataRegistry) {
+    public BearingWidget(Context context, QuestionDetails questionDetails, WaitingForDataRegistry waitingForDataRegistry, SensorManager sensorManager) {
         super(context, questionDetails);
         this.waitingForDataRegistry = waitingForDataRegistry;
+        this.sensorManager = sensorManager;
 
         isSensorAvailable = checkForRequiredSensors();
```
grzesiek2010: I would call the filed and the method areSensorsAvailable.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
         super.cancelLongPress();
         if (isSensorAvailable) {
-            getBearingButton.cancelLongPress();
+            binding.bearingButton.cancelLongPress();
         }
-        answer.cancelLongPress();
+        binding.answerText.cancelLongPress();
     }
 
     private boolean checkForRequiredSensors() {
```
grzesiek2010: I think this method is unnecessarily big can we just have:
```java
        return sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER) != null 
                && sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD) != null;
```

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
         if (isSensorAvailable) {
-            Intent i = new Intent(getContext(), BearingActivity.class);
+            Intent intent = new Intent(getContext(), BearingActivity.class);
             waitingForDataRegistry.waitForData(getFormEntryPrompt().getIndex());
-            ((Activity) getContext()).startActivityForResult(i,
-                    RequestCodes.BEARING_CAPTURE);
+            ((Activity) getContext()).startActivityForResult(intent, RequestCodes.BEARING_CAPTURE);
         } else {
-            getBearingButton.setEnabled(false);
+            binding.bearingButton.setEnabled(false);
```
grzesiek2010: I don't like mixing things here I mean showing the toast i the middle of other stuff. It should be at the begging or at the end.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
-        return manualData;
-    }
-
-    @Override
-    public void onButtonClick(int buttonId) {
+    private void onButtonClick() {
         if (isSensorAvailable) {
-            Intent i = new Intent(getContext(), BearingActivity.class);
+            Intent intent = new Intent(getContext(), BearingActivity.class);
             waitingForDataRegistry.waitForData(getFormEntryPrompt().getIndex());
```
grzesiek2010: Could you move it one line down? The reason is like below (mixing stuff).

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
+            binding.bearingButton.setOnLongClickListener(l);
         }
-        answer.setOnLongClickListener(l);
+        binding.answerText.setOnLongClickListener(l);
     }
 
     @Override
     public void cancelLongPress() {
         super.cancelLongPress();
         if (isSensorAvailable) {
```
grzesiek2010: Do we really need this check? I think it's redundant.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
     @Override
     public void setBinaryData(Object answer) {
-        this.answer.setText((String) answer);
+        binding.answerText.setText((String) answer);
         widgetValueChanged();
     }
 
     @Override
     public void setOnLongClickListener(OnLongClickListener l) {
         if (isSensorAvailable) {
```
grzesiek2010: Do we really need this check? I think it's redundant.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
-        LinearLayout answerLayout = new LinearLayout(getContext());
-        answerLayout.setOrientation(LinearLayout.VERTICAL);
-
-        answer = getEditText();
-        textBackground = answer.getBackground();
-        answer.setBackground(null);
-
-        getBearingButton = createSimpleButton(getContext(), getFormEntryPrompt().isReadOnly(), getContext().getString(R.string.get_bearing), getAnswerFontSize(), this);
+        String answerText = questionDetails.getPrompt().getAnswerText();
+        if (answerText != null && !answerText.equals("")) {
```
grzesiek2010: `isEmpty()` would be better here instead of `equals("")`

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -21,176 +21,126 @@
...
+            ((Activity) getContext()).startActivityForResult(intent, RequestCodes.BEARING_CAPTURE);
         } else {
-            getBearingButton.setEnabled(false);
+            binding.bearingButton.setEnabled(false);
             ToastUtils.showLongToast(R.string.bearing_lack_of_sensors);
-            answer.setBackground(textBackground);
-            answer.setFocusable(true);
-            answer.setFocusableInTouchMode(true);
-            answer.requestFocus();
+            binding.answerText.setBackground(textBackground);
```
grzesiek2010: Can we get rid of `textBackground` field and for example call here:
```java
binding.answerText.setBackground(new EditText(getContext()).getBackground());
```

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
SaumiaSinghal: Oh right! I will change that. I might have set the answer in the constructor in some others too. I'll change them

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
SaumiaSinghal: I need to update the button label in `setBinaryData()`. I will add a unit test too.

collect_app/src/main/res/values/buttons.xml
SaumiaSinghal: I tried settings the margins in parent view but that didn't fix the problem. The widget looks fine but as soon as the `EditText` starts accepting answers, it requires an extra horizontal margin, so, I added the margin dynamically, using 
```
marginLayoutParams.setMarginStart(UiUtils.getDimen(R.dimen.margin_standard));
marginLayoutParams.setMarginEnd(UiUtils.getDimen(R.dimen.margin_standard));
```

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
SaumiaSinghal: In case the button is disabled, the listener will not listen to any click events, so yes, it is redundant.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
grzesiek2010: Something went wrong here because you just added and empty line instead of moving the line I mentioned down.

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
```diff
@@ -18,179 +18,125 @@
...
 @SuppressLint("ViewConstructor")
-public class BearingWidget extends QuestionWidget implements BinaryDataReceiver, ButtonClickListener {
-    final Button getBearingButton;
-    private final boolean isSensorAvailable;
-    final EditText answer;
-    private final Drawable textBackground;
+public class BearingWidget extends QuestionWidget implements BinaryDataReceiver {
+    BearingWidgetAnswerBinding binding;
+
+    private final boolean areSensorsAvailable;
```
grzesiek2010: Now we use this only in one place so it doesn't make sense to keep the field we can just use the method which initializes it (`areSensorsAvailable()`).

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/BearingWidget.java
SaumiaSinghal: Oh sorry right

collect_app/src/main/res/values/buttons.xml
grzesiek2010: You did something wrong I think because this seems to work:
https://github.com/getodk/collect/commit/99ee788add2fdc20c799bf9e6311f6835a7ff94f

grzesiek2010 (approved):

grzesiek2010:
:+1:

mmarciniak90:
**Tested with success**

All my devices have a compass so I'm not able to manually enter the value. 

Verified cases:
* adding value
* removing response
* read-only
* rotation
* light/dark mode
* RTL language

kkrawczyk123:
Verified on Andoirds: 5.1, 7.1 and 9.
I was able to verify on a device without a compass. 

seadowg:
@SaumiaSinghal are you able to rebase to fix conflicts?

Issue #3029, opened 2019-04-19
Some toasts are visible in the language selected on the device, not on Collect


mmarciniak90:
#### Software and hardware versions 
Verified on Collect v1.19
Android v8.1, v7.0
device used: Nexus 5X, Samsung S7

#### Problem description
Some toasts are visible in the language selected on the device, not on Collect on Android 8.1

#### Steps to reproduce the problem
1. **Spanish** language is set on device
2. **English** language is set on Collect
3. User fills form and clicks on Save icon -> `Formulario guardado correctamente` toast is visible
4. User goes to Form Metadata and adds incorrect email -> `Email inválido`  toast is visible
5. User imports settings -> `Ajustes importados exitosamente` toast is visible

|![Screenshot_20190419-142003](https://user-images.githubusercontent.com/3907554/56424552-07bbea80-62b1-11e9-849a-f12eb161e6b7.png)|![Screenshot_20190419-141700](https://user-images.githubusercontent.com/3907554/56424553-07bbea80-62b1-11e9-8d18-984d0ce49778.png)|![Screenshot_20190419-141819](https://user-images.githubusercontent.com/3907554/56424554-07bbea80-62b1-11e9-88df-c5447a63260b.png)|
|---|---|---|

#### Expected behavior
All toasts are visible in the language selected on Collect

grzesiek2010:
Sometimes we get strings using the application context like `Collect.getInstance().getString(R.string.data_saved_ok);` (becuase an activiti context is not available)) and in those cases (depending on android version) the problem occurrs.

According to https://proandroiddev.com/change-language-programmatically-at-runtime-on-android-5e6bc15c758
`Another pitfall of using createConfigurationContext is that you can’t actually update the resources for Application after you change the language at runtime since attachBaseContext is never called again. Therefore, you have to restart the application to update the resources.`

resources are not updated after changing the language so getting a string using the application context as I described above cause that we receive wrong translations.

So the solution would be to restart the app after changing the language not only closing all activities. What do you think about this solution @lognaturel

mmarciniak90:
I'm closing this issue because the mentioned problem no longer exists.
Verified on released version **1.30.1**, on Android **7.0, 8.1, 10.0** 

Toasts are consistent and are presented in the language selected in Collect's settings, even if a completely different language is set on the device.

PR #4045, opened 2020-08-26
Fixed handling translations

### LINKED ISSUE #3988: Notifications not being translated

grzesiek2010:
Closes #3988 #3029

I tested the scenario from #3988 

It's the only solution. Sometimes we just can access application context only and we need to get translations using it however the application context is not updated when we change language so we need the helper class I created.

This pr touched many classes but I don't think we should test the whole app. I think we can confirm that it fixes #3988 and merge then test it on a daily basis.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
@@ -118,10 +118,10 @@ public class AutoSendTaskSpec implements TaskSpec {
                     } catch (SubmitException e) {
                         switch (e.getType()) {
                             case GOOGLE_ACCOUNT_NOT_SET:
-                                notifier.onSubmission(true, Collect.getInstance().getString(R.string.google_set_account));
+                                notifier.onSubmission(true, context.getString(R.string.google_set_account));
                                 break;
                             case GOOGLE_ACCOUNT_NOT_PERMITTED:
-                                notifier.onSubmission(true, Collect.getInstance().getString(R.string.odk_permissions_fail));
+                                notifier.onSubmission(true, context.getString(R.string.odk_permissions_fail));
                                 break;
                             case NOTHING_TO_SUBMIT:
                                 break;
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkUtils.java
@@ -2,6 +2,7 @@ package org.odk.collect.android.backgroundwork;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.application.Collect;
+import org.odk.collect.android.utilities.TranslationHandler;
 
 public class BackgroundWorkUtils {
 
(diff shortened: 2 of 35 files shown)


collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
```diff
@@ -102,10 +103,10 @@
                     } catch (SubmitException e) {
                         switch (e.getType()) {
                             case GOOGLE_ACCOUNT_NOT_SET:
-                                notifier.onSubmission(true, Collect.getInstance().getString(R.string.google_set_account));
+                                notifier.onSubmission(true, TranslationHandler.getString(Collect.getInstance(), R.string.google_set_account));
```
seadowg: We can actually use the `context` param here instead of `Collect.getInstance()`

collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkUtils.java
```diff
@@ -15,13 +16,13 @@ private BackgroundWorkUtils() {
     }
 
     public static long getPeriodInMilliseconds(String period) {
-        if (period.equals(Collect.getInstance().getString(R.string.every_one_hour_value))) {
+        if (period.equals(TranslationHandler.getString(Collect.getInstance(), R.string.every_one_hour_value))) {
```
seadowg: This method could take a `Context` param in our just be inlined it looks like as it's only called from one place that has access to `Application`.

collect_app/src/main/java/org/odk/collect/android/preferences/GeneralKeys.java
```diff
@@ -95,7 +96,7 @@
     private static HashMap<String, Object> getHashMap() {
         HashMap<String, Object> hashMap = new HashMap<>();
         // odk_server_preferences.xmll
-        hashMap.put(KEY_SERVER_URL,                 Collect.getInstance().getString(R.string.default_server_url));
+        hashMap.put(KEY_SERVER_URL,                 TranslationHandler.getString(Collect.getInstance(), R.string.default_server_url));
```
seadowg: All the values in here should be in `unstranslated.xml` right? I think we should probably just use `getString` directly here.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormApiExceptionMapper.java
```diff
@@ -1,23 +1,17 @@
...
 import org.odk.collect.android.R;
+import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.openrosa.api.FormApiException;
+import org.odk.collect.android.utilities.TranslationHandler;
 
 public class FormApiExceptionMapper {
 
-    private final Resources resources;
-
-    public FormApiExceptionMapper(Resources resources) {
```
seadowg, line 12: I think it'd still be nice for this object to take in `Context` as a constructor dependency so it doesn't have to rely on `Collect.getInstance()`

seadowg (changes requested):
Looks great! Just noted a few places where it looks like we could use a local context object.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormApiExceptionMapper.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/BackgroundWorkUtils.java
grzesiek2010: Done.

seadowg (approved):

seadowg:
@grzesiek2010 looks like this has collected conflicts ☹️

grzesiek2010:
> looks like this has collected conflicts

fixed.

mmarciniak90:
**Tested with success**

Verified cases:
* Spanish language set on device, German language set on Collect
* English language set on device, Spanish language set on Collect
* English language set on device, French language set on Collect
* toast when user clicks on save icon while filling form
* toast when user clicks on button `Save Form and Leave`
* toast when user puts incorrect email in Form Metadata
* toast when user imports settings
* autosend notification
* autodownload the latest form version notification
* light/dark mode

kkrawczyk123:

PR #4047, opened 2020-08-27
Remove analytics that are no longer actionable

### LINKED ISSUE #3952: Review analytics events and drop ones that are no longer actively used
For example, we now know `likert` is broadly used and can stop tracking those events. It's hard to know what to pay attention to when we have a lot of events and Firebase caps the number of actions that we can see in its interface so it's worth it to clean up periodically.

lognaturel:
Closes #3952

I decided to only add submission analytics for custom endpoints because I don't think tracking custom formList endpoints adds anything. Current analytics show they're almost always set together.

Otherwise I followed the issue. Let's close it for now and I'll keep track of the other suggestions elsewhere.

Get, filled, submitted form. I only visually verified the analytics calls but I think that's sufficient.

I didn't add tests because this seems short-lived and testing would generally require refactoring that doesn't seem like a good idea to take on now.

There's no intended change. The one risk I can think of is that I do introduce some conditionals and there could be things like nullpointerexception opportunities that snuck in.

No.

No.


collect_app/src/test/java/org/odk/collect/android/configure/qr/QRCodeActivityResultDelegateTest.java
```diff
@@ -53,62 +57,72 @@ public QRCodeDecoder providesQRCodeDecoder() {
     }
 
     @Test
-    public void forSelectPhoto_importsSettingsFromQRCode_showsSuccessToast() {
-        QRCodeActivityResultDelegate delegate = new QRCodeActivityResultDelegate(context, settingsImporter, fakeQRDecoder);
+    public void forSelectPhoto_importsSettingsFromQRCode_showsSuccessToast_andLogsAnalytics() {
```
seadowg: Yay for delegate tests! I think here I'd personally split the tests so that each "then" of a "given"/"when" gets its own test. Can just make it easier to navigate the assertions and change individual features. The trade off is run time of course but these tests are pretty fast once Robolectric has booted up (which is a small cost when you're running everything).

seadowg (changes requested):
Just one tweak but looks great!

seadowg (approved):

seadowg (approved):

PR #4049, opened 2020-08-27
Fixed automatic location recording

### LINKED ISSUE #3984: The automatic location recording method does not work as expected
Collect master

#### Problem description
The automatic location recording method does not work as expected

1. User opens GeoTrace or GeoShape widgets
2. User tries to add points via Automatic location recording method

#### Expected behavior
* The first point is collected immediately for automatic mode
* It's possible to add points with interval 1 second with Accuracy 20m or other
* It's possible to add points with Accuracy None
* It's possible to clear previously added points and add new points with other parameters. New parameters appear on black footer without delay.

grzesiek2010:
Closes #3984

This is a bug fix and the only solution. The problem was that we tried to start recording first after clicking Start button and then in `onDestroy` method we set properties we need to start the process. It should work the other way around.

It's just a bug fix so I find it pretty safe. I think that testing recording locations would be enough.

Any form with GeoTrace/GeoShape widgets like the all widgets form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/SettingsDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/SettingsDialogFragment.java
@@ -14,6 +14,7 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.fragment.app.DialogFragment;
 
+import org.jetbrains.annotations.NotNull;
 import org.odk.collect.android.R;
 
 public class SettingsDialogFragment extends DialogFragment {
@@ -35,7 +36,7 @@ public class SettingsDialogFragment extends DialogFragment {
     private int accuracyThresholdIndex = -1;
 
     @Override
-    public void onAttach(Context context) {
+    public void onAttach(@NotNull Context context) {
         super.onAttach(context);
 
         if (context instanceof SettingsDialogCallback) {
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/geo/SettingsDialogFragment.java
```diff
@@ -35,7 +36,7 @@
     private int accuracyThresholdIndex = -1;
 
     @Override
-    public void onAttach(Context context) {
+    public void onAttach(@NotNull Context context) {
```
lognaturel: How I wish these annotations were more standardized! Since this is an android class, we should use `NonNull` which is already imported here. I think it makes sense to use this one for classes that are not related to android at all.

lognaturel (changes requested):
Good catch and fix.

collect_app/src/main/java/org/odk/collect/android/geo/SettingsDialogFragment.java
grzesiek2010: Oh right, fixed.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified on Android 7.0, 8.1, 9.0

Verified cases:
* GeoTrace and GeoShape
   * The first point is collected immediately for automatic mode
   * It's possible to add points with interval 1 second with Accuracy 20m or other
   * It's possible to add points with Accuracy None
   * It's possible to clear previously added points and add new points with other parameters. New parameters appear on the black footer without delay
* quick check `Manual location recording` method 
* removing response
* light / dark mode
* RTL language
* change device orientation

kkrawczyk123:

PR #4053, opened 2020-08-27
Update JavaRosa to v3.0.0

lognaturel:
There's no difference in behavior between 2.18-snapshot and this one. This just gets us on a release build rather than a snapshot.


yanokwa (approved):

lognaturel:
@getodk/testers Heads up that I merged this. There should be no change in behavior, though, so I haven't tagged as needing testing.

PR #4055, opened 2020-08-28
Do not save answers again after closing select minimal dialog to improve performance

### LINKED ISSUE #4052: Exiting minimal select dialog triggers recomputation even if selected value hasn't changed
`master`

#### Problem description
Exiting minimal select dialog triggers recomputation even if selected value hasn't changed. This can be time consuming in the case of a cascade of selects with lots of choices.

Load a form with selects with long lists and choice filters: [nigeria_wards_internal_select_minimal.xlsx](https://github.com/getodk/collect/files/5138667/nigeria_wards_internal_select_minimal.xlsx)

#### Expected behavior
If the selected value hasn't changed, don't trigger recomputation.

grzesiek2010:
Closes #4052 

I tested the changes manually and added automated tests.

There is just no need to update answers if they changed. There is no better solution.

I'm not sure about testing because it's an improvement that might be not even visible so it will be difficult to verify. Maybe trying older devices would help?

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/adapters/AbstractSelectListAdapter.java
@@ -323,4 +323,6 @@ public abstract class AbstractSelectListAdapter extends RecyclerView.Adapter<Abs
     }
 
     public abstract void clearAnswer();
+
+    public abstract boolean hasAnswerChanged();
 }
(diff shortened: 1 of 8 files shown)


lognaturel (approved):

lognaturel:
> an improvement that might be not even visible so it will be difficult to verify

I've run it on my Galaxy Tab A with the form I attached to the issue and it's a huge difference. 

I agree it's not risky and I'd be in favor of merging as-is since there will be a regression pass shortly. The only real risk I see is if there's a mistake in the algorithm for identifying identical selections then forms with multiple selects in field lists might not correctly recompute. I tried to be as careful as I can reviewing it. It's a straightforward sequential search and using the `xmlValue` for comparison should be the right thing to do. There's a targeted test that verifies comparison against an empty list and against a list of a different order.

@getodk/testers tagging as `needs testing` for now but if you're satisfied with my analysis, please give the 🟢 to merge.

mmarciniak90:
**Tested with success**

I observed a difference in the Android versions 7.0, 8.1, 9.0

Verified cases:
* fill form as the first time
* edit saved form
* click on the same option
* click on back arrow
* clicks on device back button
* light/dark mode
* RTL language

kkrawczyk123:
Verified with success on Androids 5.1, 7.1 and 10.

PR #4061, opened 2020-09-02
Fixed invisible radio buttons / checkboxes on android 5.0

### LINKED ISSUE #4060: Radio buttons/checkboxes are not visible in select on Android 5.0
####  Software and hardware versions 
Collect v1.28, Android v5.0

#### Problem description
Radio buttons/checkboxes are not visible in select on Android 5

The bug has been originally reported on the forum: https://forum.getodk.org/t/boxes-for-select-multiple-in-android-5/29621/4

Try any form with selects on android 5.0

grzesiek2010:
Closes #4060 

I tested the fix on various devices.

The issue stems from the fact that Android 5.0 doesn't work well with insets https://github.com/getodk/collect/commit/967537f10eed60491adfebdf5ec79964926fd72b#diff-e4fc71042ed4849e61170b860e16592cR4

when we used only one inset left or right we had that problem with ugly buttons:
image
using both they are not visible at all.

Since the issue is visible only on Android 5.0 I created separate drawables for that Android version without insets. It would look worse than on all the other versions but at least everything will be displayed:
| Android 5.0  | Android 5.1 and newer |
| Screenshot_1599047434  | Screenshot_1599047443  |

We need to test selects on various android versions. It's not a risky change so there is no need to test different appearances, just simple select_one and select_multiple questions. 

Any form with selects like the AllWidgets form.

No.

### DIFF
--- a/collect_app/src/main/res/drawable-v22/inset_check_box.xml
+++ b/collect_app/src/main/res/drawable-v22/inset_check_box.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<inset xmlns:android="http://schemas.android.com/apk/res/android"
+    android:drawable="?android:attr/listChoiceIndicatorMultiple"
+    android:insetLeft="@dimen/margin_standard"
+    android:insetRight="@dimen/margin_standard"/>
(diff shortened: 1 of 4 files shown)


seadowg (approved):

grzesiek2010:
The QA team doesn't even have a device with Android 5.0 so we agreed (@lognaturel @kkrawczyk123 @grzesiek2010) that we can merge since I tested it. If anything was wrong it will be visible during regression testing for sure.

PR #4063, opened 2020-09-03
Fixed strange bug with mixing answers in field-list

### LINKED ISSUE #4062: Incorrect answers are displayed in form with selects with long lists when user changes device orientation
master

#### Problem description
Incorrect answers are displayed in form with selects with long lists when user changes device orientation

1. User starts form [nigeria_wards_internal_select_minimal.xml.txt](https://github.com/getodk/collect/files/5163424/nigeria_wards_internal_select_minimal.xml.txt) , 
[search_and_select.zip](https://github.com/getodk/collect/files/5168109/search_and_select.zip)

2. User adds answers
Screenshot_20200902-180431
3. User changes device orientation -> the last answer shows everywhere
Screenshot_20200902-180442

#### Expected behavior
Correct answers are visible

#### Other information 
Noticed during testing #4055

grzesiek2010:
Closes #4062

I tested the solution manually.

I haven't seen anything so confusing for a very long time. At first when I saw the issue I said to myself ok I must have made some stupid bugs implementing changes in selects, it happens I will fix it in 30 minutes but then I started investigating the issue and I wasn't able to find any bug.
After debugging, playing with other widgets that use view binding my gut said try to just use a different view element to display answers and bingo it turned out that the issue happens when we use `EditTexts`.
I have no idea why. What's interesting the problem occurs only if we rotate a device, if we navigate to another question and then back everything is fine despite the fact that exactly the same methods are called building widgets in both cases.
@seadowg @SaumiaSinghal can you maybe explain that magic?

this fix shouldn't cause any change or risk so testing only the issue would be enough.

The forms attached to the issue.

No.

### DIFF
--- a/collect_app/src/main/res/layout/select_minimal_widget_answer.xml
+++ b/collect_app/src/main/res/layout/select_minimal_widget_answer.xml
@@ -15,6 +15,7 @@
         android:focusable="false"
         android:drawableEnd="@drawable/ic_arrow_drop_down"
         android:cursorVisible="false"
-        android:inputType="textMultiLine|textNoSuggestions">
+        android:inputType="textMultiLine|textNoSuggestions"
+        android:saveEnabled="false">
     </com.google.android.material.textfield.TextInputEditText>
 </FrameLayout>
--- a/collect_app/src/test/java/org/odk/collect/android/widgets/items/SelectOneMinimalWidgetTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/widgets/items/SelectOneMinimalWidgetTest.java
@@ -33,6 +33,11 @@ public class SelectOneMinimalWidgetTest extends GeneralSelectOneWidgetTest<Selec
     }
 
     @Test
+    public void answerView_shouldHaveSavingStateDisabled() {
+        assertThat(getSpyWidget().binding.answer.isSaveEnabled(), is(false));
+    }
+
+    @Test
     public void usingReadOnlyOptionShouldMakeAllClickableElementsDisabled() {
         when(formEntryPrompt.isReadOnly()).thenReturn(true);
         assertThat(getSpyWidget().binding.answer.getVisibility(), is(View.VISIBLE));


seadowg:
@grzesiek2010 I think this could be because the Activity will automatically save the text of `EditText` views in the `onSaveInstanceState` bundle and then restore it after rotation using the view ID. Here the view ID for all the `EditText`s would be the same, so I guess it would restore every one of them with the same value. This is just a theory though and finding docs on `EditText` behaving like this is really hard (there is an SO [here](https://stackoverflow.com/questions/12420396/how-to-retain-edittext-data-on-orientation-change) though). I guess we can verify by checking the bundle in `onSaveInstanceState` and `onCreate` in the debugger.

If that is the case then an alternative fix might be to use a `View.generateViewId()` in the widget to set a different ID to on the `EditText`. One advantage of that is that it'd be testable but I'm worried it could cause problems for view binding?

seadowg:
Ok partly confirmed. I used the `FieldListUpdateTest` form to quickly get a Select Minimal widget and if I rotate after selecting "Mango" I do end up with an item in the bundle with `"Mango"` which I think is what ends up getting loads into the view:

Screenshot 2020-09-03 at 17 10 22

grzesiek2010:
Oh right I knew but somehow forgot about that tricky behavior. In that case I think we can just disable it using [saveEnabled="false"](https://developer.android.com/reference/android/view/View#attr_android:saveEnabled)
We also need to remember about it because once we refactor text widgets to use view binding we will have the same issue.

seadowg:
@grzesiek2010 hah that's a nicer solution. Yeah it's scarily subtle and not something you expect.

seadowg (changes requested):
As @grzesiek2010 suggested.

seadowg (approved):
Nice!

mmarciniak90:
**Tested with success**

Verified cases:
* both forms
* rotate device orientation several times (on a new and edited form)
* light/dark mode
* RTL language
* minimize
* don't keep activities ON/OFF

kkrawczyk123:
Tested on Androids: 5.1, 7.1 and 10.

PR #4064, opened 2020-09-03
Show friendlier errors on form download failure

seadowg:
While working on something else I noticed that for v1.28 we're throwing the exception messages up to the user when form download fails for things like server errors (a 500 or a 404 for example). I think for 1.29 we want to look at improving these errors but for the moment I think making them simpler is the best/easiest thing. For now any form download error will just show as "Failure" (if it succeeds we simply put "Success") as the result of the individual download.

Wrote a new test as it's hard to verify manually.

I think there are a lot of better things we could do. My ideal would be to make the details of the error available to the user but maybe hide them behind an expanding text view or something. Then we could potentially have a simpler "everything succeeded" or "everything failed" message.

I think creating an issue and fixing as part of #4040 might be a nice plan? If people agree I can start an issue for discussion/mockups.

Should just make errors friendlier. Good to check other form download failures than just server errors to check nothing weird has been introduced.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/GetBlankFormsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/GetBlankFormsTest.java
@@ -48,4 +48,17 @@ public class GetBlankFormsTest {
                 .assertText(R.string.report_to_project_lead)
                 .clickOK(new GetBlankFormPage(rule));
     }
+
+    @Test
+    public void whenThereIsAnErrorFetchingForms_showsError() {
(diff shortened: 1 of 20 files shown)


grzesiek2010 (approved):

kkrawczyk123:
My results:
- invalid url: "Collect can't reach the server at ... Did you enter the URL correctly? If you keep having this problem, report it to the person who asked you to collect data"
- not existing server: "If you keep having this problem, report it to the person who asked you to collect data".
- server with required credetials: invalid credentials - Serve Required Autenthication dialog 
- offline: No network available toast
- invalid central url: "If you keep having this problem, report it to the person who asked you to collect data".
- download form: Success message
- turned off connection while downloading forms: Failure message

PR #4067, opened 2020-09-07
Fixed getFileExtensionFromUri() method

### LINKED ISSUE #4056: The dot between name and file extension is not visible for choosed file on hierarchy view
Android: 7.0, 8.1

#### Problem description
The dot between name and file extension is not visible for choosed file on hierarchy view

1. User adds audio/video file from existing files on device
2. User opens hierarchy view
3. The dot between name and file extension is not visible
Screenshot_20200828-172556

#### Expected behavior
The dot between name and file extension is visible

#### Other information 
Problem is not visible for recorded files

grzesiek2010:
Closes #4056 

The problem was that in [one case](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4056?expand=1#diff-e6929594f15505b42123fc05d98b17dcR85) getFileExtensionFromUri() returned values with dots but in [other cases](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4056?expand=1#diff-e6929594f15505b42123fc05d98b17dcR87) without dots and that's why sometimes we had doubled dots but sometimes no dots at all.

It's juts a bug fix so there is no other solution.

This should be a safe change that will just fix the bug, nothing else should be affected so testing attaching and capturing media files would be enough.

Any form with media widgets like the AllWidgets form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
@@ -89,7 +89,7 @@ public final class ContentResolverHelper {
                     filename = returnCursor.getString(nameIndex);
                 }
                 if (filename != null && filename.lastIndexOf('.') != -1) {
-                    return filename.substring(filename.lastIndexOf('.'));
+                    return filename.substring(filename.lastIndexOf('.') + 1);
                 } else {
                     return fileUri.getScheme() != null && fileUri.getScheme().equals(ContentResolver.SCHEME_CONTENT)
                             ? MimeTypeMap.getSingleton().getExtensionFromMimeType(context.getContentResolver().getType(fileUri))
--- a/collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
@@ -77,7 +77,7 @@ public final class ContentResolverHelper {
      * Using contentResolver to get a file's extension by the uri
      *
      * @param fileUri Whose name we want to get
-     * @return The file's extension
+     * @return The file's extension without a dot eg. "mp3" not ".mp3"
      */
     public static String getFileExtensionFromUri(Context context, Uri fileUri) {
         try (Cursor returnCursor = getContentResolver().query(fileUri, null, null, null, null)) {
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
```diff
@@ -18,14 +18,14 @@
...
 import android.webkit.MimeTypeMap;
 
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.logic.FormInfo;
 import org.odk.collect.android.provider.FormsProviderAPI.FormsColumns;
 import org.odk.collect.android.provider.InstanceProviderAPI.InstanceColumns;
 import org.odk.collect.android.storage.StoragePathProvider;
+import org.odk.collect.android.utilities.MediaUtils;
 
 public final class ContentResolverHelper {
```
seadowg, line 30: I think it'd be nice to change the name here to `...Utils` as that seems to have become idiomatic for static method buckets.

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
grzesiek2010: You mean ContentResolverUtils? What about the package then? It's in helpers package along with other classes that use helper suffixes.

seadowg: Hmmm good point. I'd be up for moving those out into `utils` (or more appropriate packages) but now is maybe not the time.

mmarciniak90:
**Tested with success**

Verified on Android: 7.0, 8.1, 10.0

| 1.27.3 | PR|
|---|---|
Screenshot_20200930-121543|Screenshot_20200930-121255|

Verified cases:
* record sound -> open hierarchy view -> one dot is visible before file extension
* add a record from existing files (select `Open from Audio` option) -> open hierarchy view -> one dot is visible before file extension
* record video -> open hierarchy view -> one dot is visible before file extension
* add a video from existing files (select `Open from Video` option) -> open hierarchy view -> one dot is visible before file extension
* remove reponse
* read only
* minimize
* light/dark mode
* RTL language

kkrawczyk123:

PR #4079, opened 2020-09-11
Lint improvements

grzesiek2010:
We are in the middle of regression testing so it's always a good time to add some small less important improvements.

I tested parts of the app that might by affected somehow but the changes I implemented.

Those are just code improvements suggested by Lint.

It doesn't require testing just reviewing.

No.

No.


config/lint.xml
```diff
@@ -17,6 +17,7 @@
     <issue id="UselessParent" severity="error" />
     <issue id="BinaryOperationInTimber" severity="error" />
     <issue id="RelativeOverlap" severity="error" />
+    <issue id="Typos" severity="error" />
```
seadowg, line 20: Is this potentially risky? I'm worried there might be colloquialisms etc that show up as typos from the translators.

seadowg (changes requested):
One worry but overall looks like some great spring (late summer) cleaning!

config/lint.xml
grzesiek2010: You might be right, if that happen then we can turn this rule off but for now I find it helpful because it will help us fix bugs in new translations.

seadowg: Good point. It's really easy to disable if we run into problems.

seadowg (approved):

PR #4081, opened 2020-09-14
Fixed changing languages

### LINKED ISSUE #4077: Change language issue
Collect v1.28.x, Android v7, 9, 10 probably all device used...

#### Problem description
I have noticed that changing the language does not work in all tried. On Android 7 it doesn't work at all. On other ones it works sometimes - not sure exactly why. In all tries after closing the app after opening it again default language is restored. 

ezgif com-video-to-gif (22)

1. Change language to Spanish
2. Change language to RTL
3. Close an app 
4. Go back to Collect

#### Expected behavior
- the language should be updated immediately 
- default language should not be restored after closing app and going back

#### Other information 
- compared with 1.27: the issue is not visible

grzesiek2010:
Closes #4077

I tested the fix on various different APIs 21...29

The issue was caused by changes in androidx.appcompat. I implemented my fix according to what they said in release notes:
https://developer.android.com/jetpack/androidx/releases/appcompat#1.2.0
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:appcompat/appcompat/src/androidTest/java/androidx/appcompat/app/NightModeCustomApplyOverrideConfigurationActivity.java

We should test changing language on as many different android versions as possible. Would be also good to verify if everything is fine after reseting settings.

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -199,7 +199,7 @@ dependencies {
 
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.browser:browser:1.2.0'
-    implementation 'androidx.appcompat:appcompat:1.2.0-rc02'
+    implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation "androidx.work:work-runtime:2.3.4"
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.exifinterface:exifinterface:1.2.0'
--- a/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/dao/FormsDaoTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/dao/FormsDaoTest.java
@@ -34,7 +34,6 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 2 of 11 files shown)


lognaturel (approved):
Great to get off a release candidate and even better that there's now an explicitly documented way to do all this.

kkrawczyk123:
@grzesiek2010 I have noticed a small issue with the text cursor in RTL. It is also visible on 1.27.3 so it was not introduced in that PR. I was wondering if you could take a look if it's quick to fix. If it's not, don't bother and we will report it separately as it's just not aesthetic: 
Screenshot_2020-09-15-12-31-13

I have verified cases:
- jumping between different languages
- RTL
- filling form
- reset settings
- use different devices language 
on Androids: 5.1, 7.1, 8.1, 9.0 and 10.

grzesiek2010:
Since the issue with doubled cursor is not visible on the newest versions 9+ and I can't see any easy fix let's merge without it. To be honest I'm not sure if I would even report it in a separate issue since it's not a big problem. I would rather wait once we refactor those widgets and move the code responsible for creating those text boxes to layout files. Maybe this will fix the issue.

kkrawczyk123:
Sure I agree.

PR #4082, opened 2020-09-14
Fixed checking for spaces in underlying values

### LINKED ISSUE #4078: Warnings are not visible for select multi questions

#### Problem description
Warning messages are not visible for select multi widget on Collect 1.28 see gif:
ezgif com-video-to-gif (23)

Gif of the same form on 1.27/1.27.3:
ezgif com-video-to-gif (24)

1. Fill form 
[multi-warning.xml.txt](https://github.com/getodk/collect/files/5208067/multi-warning.xml.txt)

#### Expected behavior
Warnings should be visible for all questions.

#### Other information 
the issue does not occur on 1.27

grzesiek2010:
Closes #4078

I moved checking for spaces in underlying values to the main ItemsWidget class what is the best solution because all widgets that need the warning extend that class, so we have everything in one place and we can test it easily.

We need to test the form attached to he issue and I think it would be enough since the change is pretty safe and we have automated tests.

The form attached to the issue is ok.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/items/ListMultiWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/items/ListMultiWidget.java
@@ -42,6 +42,7 @@ import org.odk.collect.android.external.ExternalSelectChoice;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
 import org.odk.collect.android.utilities.FileUtils;
 import org.odk.collect.android.widgets.interfaces.MultiChoiceWidget;
+import org.odk.collect.android.widgets.warnings.SpacesInUnderlyingValuesWarning;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -233,6 +234,10 @@ public class ListMultiWidget extends ItemsWidget implements MultiChoiceWidget {
 
                 buttonLayout.addView(answer, answerParams);
             }
+
+            SpacesInUnderlyingValuesWarning
+                    .forQuestionWidget(this)
+                    .renderWarningIfNecessary(items);
         }
     }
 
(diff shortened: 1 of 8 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/items/RankingWidget.java
```diff
@@ -50,8 +49,8 @@
 
     public RankingWidget(Context context, QuestionDetails prompt) {
         super(context, prompt);
-
-        setUpLayout(getOrderedItems());
+        getOrderedItems();
```
lognaturel: This is not a new change but a getter that mutates is dangerous, I think. A rename to `buildOrderedItems` would help.

lognaturel (changes requested):
It's ok to have spaces in single-select question types so I think this is too aggressive.

This regression shows the tests were clearly not covering enough -- it's important to have a test that verifies the check for spaces is actually done. I wonder whether passing in the `SpacesInUnderlyingValuesWarning` to each relevant widget type is the best way, though. Did you consider a Robolectric test that checks for the warning text in the view? I think it may then be ok for the `SpacesInUnderlyingValuesWarning` dependency to be an implementation detail (managed internally). That does mean you'd have to have an explicit, separate test for each question type that allows multiple selections. Might be worth a quick sanity check with @seadowg on the approach.

grzesiek2010:
@lognaturel you are right I made a mistake implementing the warning for all selects. I prepared a new solution, please review.

lognaturel (approved):
Thanks, looks great! For QA, the risk here is forgetting one of the multi select types. I double checked and I think all are accounted for.

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- form from issue
- select one questions
- select multiple questions
- column pack form
- light/dark mode 
- screen rotation/minimize

mmarciniak90:
Tested with success on Androids 7.0, 8.1, 10.0

PR #4097, opened 2020-09-16
Add self signed release build

seadowg:
This adds a build variant that has same configuration as `odkCollectRelease` but uses the default Android tooling debug keystore. This means you can build and test it from a machine without having a signing key (and update/install consistently from that machine).

Installed on emulator using new variant.

We could include our own generated debug keystore but I think this is neater and also prevents us running into problems where the debug keystore is mistaken for the real signing key (as the default debug keystore is different on different machines).

No changes to behaviour. Should be good to merge once reviewed!

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -96,6 +96,7 @@ android {
             }
         }
 
+        // This uses the default debug keystore that the Android tooling creates https://stackoverflow.com/questions/16622528/android-studio-debug-keystore
         selfSignedRelease {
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel:
@getodk/testers @grzesiek2010 This means you can do `./gradlew selfSignedRelease` to get a signed release build. We've occasionally had issues that only show up in signed release builds so this could be helpful.

PR #4098, opened 2020-09-17
Sanitize form version

### LINKED ISSUE #4095: Unnecessary update information for a specific form with external CSV files
Collect master

#### Problem description
Unnecessary update information for a specific form

1. Scan QR code
75986504-79fbe480-5eee-11ea-8f72-b0c18142bacd
or 
image

2. Download form `Form with external CSV file`
3. Back to the main menu
4. Open `Get Blank Form` view
5. Information about the form update is visible

#### Expected behavior
Information about the form update is NOT visible - it works properly on v1.27.3

grzesiek2010:
Closes #4095

I tested the fix manually and added tests.

The first thing I should mention is that this issue is not a regression towards the previous version. The behavior changed a bit because we refactored the code and now is more visible but this has never worked as expected.

The problem was that the form contained and empty version (it's a specific form that's also true):
```xml
<ecsv id="ecsv" version="">
```

Reading the list of forms we can download we care about it and convert such empty values to null:
https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormListApi.java#L187

```java
if (version != null && version.length() == 0) {
    version = null;
}
```

but we have never done the same saving files to database.
The result was that having such a form with an empty version and trying to compare it to the one we already have we weren't able to read proper values from database.

I think this solution is the best because the method I fixed is used every time we try to parse and save a form so it fixes cases when we download a form from a server or when we scan it from a disk.

The issue should be fixed and nothing else should be affected. We can test just the scenario described in the issue.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormListApi.java
+++ b/collect_app/src/main/java/org/odk/collect/android/openrosa/api/OpenRosaFormListApi.java
@@ -184,7 +184,7 @@ public class OpenRosaFormListApi implements FormListApi {
                         break;
                     case "version":
                         version = XFormParser.getXMLText(child, true);
-                        if (version != null && version.length() == 0) {
+                        if (version != null && version.trim().isEmpty()) {
                             version = null;
                         }
                         break;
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
> the method I fixed is used every time we try to parse and save a form so it fixes cases when we download a form from a server or when we scan it from a disk.

Thank you for answering that question up front!

lognaturel:
> The behavior changed a bit because we refactored the code and now is more visible

Let's add it to v1.28, then. I think some form builders may automatically add an empty version attribute.

mmarciniak90:
**Tested with success**

As a consequence of these changes, the empty `Version:` is not visible :+1: 

|1.27.3 | PR|
|---|---|
|Screenshot_20200918-132613|Screenshot_20200918-132942|

Verified cases:
- both servers from issue
- new server
- real update with added form version
- send submission

PR #4099, opened 2020-09-17
Stop server list being cleared on import

### LINKED ISSUE #4093: Server list is reset after scanning Central QR code
Collect v1.28.x, Android v all  device used...

#### Problem description
Previously used server URL should be stored in dropdown list and gives user chance to go back to already used ones - especially Centrals as they are hard to input manually. 
ezgif com-video-to-gif (31)

1. Use aggregate server other than default one
2. Scan any Central QR code: only default one is visible on dropdown list now
3. Go back to default server: Central URL is also gone from dropdown list

#### Expected behavior
URLs should be stored on the dropdown list up to 5 URLs: original PR: https://github.com/getodk/collect/pull/3352
screenshot from 1.27:
Screenshot (15 Sep 2020 18_16_33)

#### Other information 

seadowg:
Closes #4093 

Verified manually and added some new tests (the actual import flow is still not tested).

Definitely want to follow-up by removing `ChangingServerUrlUtils` and I think we should redesign the UI for setting up servers so it's more accessible and testable (probably something to think about in Onboarding).

Good to check that both importing and adding URLs still works and that they are remembered for both flows.


collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -355,8 +358,23 @@ public PropertyManager providesPropertyManager(Application application, RxEventB
     }
 
     @Provides
-    public SettingsChangeHandler providesSettingsChangeHandler(PropertyManager propertyManager, FormUpdateManager formUpdateManager) {
-        return new CollectSettingsChangeHandler(propertyManager, formUpdateManager);
+    public ServerRepository providesServerRepository() {
+        return new ServerRepository() {
```
seadowg: We can follow up with an actual tested implementation (replacing `ChangingServerUrlUtils`)

collect_app/src/main/java/org/odk/collect/android/preferences/ServerPreferencesFragment.java
```diff
@@ -280,7 +283,6 @@ public boolean onTouch(View v, MotionEvent event) {
                         sendAnalyticsEvent(url);
 
                         preference.setSummary(newValue.toString());
-                        ChangingServerUrlUtils.addUrlToList(url);
```
seadowg, line 283: This is now taken care of by our shared preference change listeners.

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ConfigureWithQRCodeTest.java
```diff
@@ -113,23 +113,6 @@ public void clickConfigureQRCode_opensScanner_andThenScanning_importsSettings()
                 .assertText("http://gallops.example");
     }
 
-    @Test
-    public void clickConfigureQRCode_opensScanner_andThenScanning_whenUsingOldQR_importsSettingsAndMigrates() {
```
seadowg, line 117: This is covered by the regression tests put in place when we found a bunch of broken edge cases. It's really hard to test here and get much value (other than we get from the normal import test).

lognaturel (dismissed):

kkrawczyk123:
@seadowg Server list is not reset when after scanning Central QR code but it becomes reset-proof at all. see the gif:
ezgif com-video-to-gif (33)

lognaturel:
We talked about this briefly in Slack. My initial reaction was that it’s not obviously a setting so it did feel a little strange to be reset with settings. 

However, in the case of Central servers the URLs do contain credentials so I can imagine someone wanting to clear out the list. With the behavior in this PR, they’d have to set 5 other URLs which is not at all intuitive. I have to think about this some more but wanted to put initial reactions out in case others have ideas on the best way to address this.

seadowg:
@kkrawczyk123 sorry we'd already discussed that and I should have mentioned it here.

@lognaturel I think we have two options:

1. Intentionally reset the server list when the user resets All Settings
2. Point users towards Android settings in docs for clearing the app "securely". 

I think if we go down the road of adding server credentials (as well as URLs) as saved items I think we might want to have "Saved Servers" explicitly in the list of reset options, so I'd opt for **1** here. I'll look at implementing that today! 

Edit: Just to clarify I'm not sure I think the credentials hanging around is a big deal but I'm worried people who might hand over devices will have become used to using "Reset application" rather than using Android settings/wiping the device (which I think should be used when wiping everything) which could cause them to leak server access.

seadowg:
@kkrawczyk123 @lognaturel I've changed the behaviour so it will reset the server list when settings are reset in the app which also afforded me an opportunity to take care of some tech debt that was left hanging around from the initial implementation. I'm marking this as ready for review and testing so if everyone's happy it has an opportunity to be ready to go by the end of the day.

kkrawczyk123:
**Tested with success!**
Tested on Androids: 5.1, 7.1 and 9.
Verified scenarios:
- scan Central QR
- reset settings
- scan QR code between devices
- import collect-settings.png
- collect.settings file 

lognaturel (approved):

Issue #4102, opened 2020-09-21
java.lang.IllegalArgumentException: radius is negative


grzesiek2010:
#### Software and hardware versions 
Collect v1.28.0-beta

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/47ec09d2126181fa55a724cddb3d6fc0?time=last-seven-days&sessionId=5F64F0FC000C0001441C131C2E306E6E_DNE_0_v2

```
Fatal Exception: java.lang.IllegalArgumentException: radius is negative
       at com.google.maps.api.android.lib6.common.m.d(m.java)
       at com.google.maps.api.android.lib6.impl.al.<init>(al.java:3)
       at com.google.maps.api.android.lib6.impl.bi.a(bi.java:84)
       at com.google.android.gms.maps.internal.i.a(i.java:102)
       at dd.onTransact(dd.java:4)
       at android.os.Binder.transact(Binder.java:675)
       at com.google.android.gms.internal.maps.zza.zza(zza.java:10)
       at com.google.android.gms.maps.internal.zzg.addCircle(zzg.java:174)
       at com.google.android.gms.maps.GoogleMap.addCircle(GoogleMap.java:45)
       at org.odk.collect.android.geo.GoogleMapFragment.updateLocationIndicator(GoogleMapFragment.java:550)
       at org.odk.collect.android.geo.GoogleMapFragment.onLocationChanged(GoogleMapFragment.java:356)
       at org.odk.collect.android.location.client.GoogleFusedLocationClient.onLocationChanged(GoogleFusedLocationClient.java:217)
       at com.google.android.gms.internal.location.zzay.notifyListener(zzay.java:4)
       at com.google.android.gms.common.api.internal.ListenerHolder.notifyListenerInternal(ListenerHolder.java:17)
       at com.google.android.gms.common.api.internal.ListenerHolder$zaa.handleMessage(ListenerHolder.java:6)
       at android.os.Handler.dispatchMessage(Handler.java:112)
       at com.google.android.gms.internal.base.zar.dispatchMessage(zar.java:8)
       at android.os.Looper.loop(Looper.java:216)
       at android.app.ActivityThread.main(ActivityThread.java:7625)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:524)
[...]

PR #4107, opened 2020-09-22
Fixed crash caused by deleted media folder

### LINKED ISSUE #4101: Collect crashes when user opens Get blank form view if a folder with media is removed
Collect master

#### Problem description
Collect crashes when user opens Get blank form view if a folder with media is removed

1. User downloads form with media
2. Media folder is removed
3. User opens `Get Blank Form`
4. **Collect crashes**

#### Expected behavior
User opens `Get Blank Form` and sees information about available update

grzesiek2010:
Closes #4101

It''s just a bug fix that didn't take into account that the file we call `listFiles()` on might not exist. I can't imagine any better solution.

This fix should just solve the problem and shouldn't affect anything else so testing the scenario from the issue should be enough.

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
@@ -2,6 +2,9 @@ package org.odk.collect.android.database;
 
 import org.junit.Test;
 import org.odk.collect.android.dao.FormsDao;
+import org.odk.collect.android.utilities.FileUtil;
+
+import java.io.File;
 
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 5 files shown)


collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
```diff
@@ -15,6 +18,32 @@ public void whenThereIsNoFormForgivenFormIdAndVersion_shouldGetAllMethodReturnEm
...
         when(formsDao.getFormMediaPath("1", "1")).thenReturn(null);
 
-        assertThat(new DatabaseMediaFileRepository(formsDao).getAll("1", "1").size(), is(0));
+        assertThat(new DatabaseMediaFileRepository(formsDao, new FileUtil()).getAll("1", "1").size(), is(0));
+    }
+
+    @Test
+    public void whenMediaFolderFileDoesNotExist_shouldGetAllMethodReturnEmptyArray() {
+        FormsDao formsDao = mock(FormsDao.class);
+        FileUtil fileUtil = mock(FileUtil.class);
```
seadowg: I think we could cut out a lot of the mocking here by having `formsDao.getFormMediaPath` just return a file path to something that doesn't exist.  You could create a temp dir (using `File.createTempDir`) and then return a path of a directory in there (that you don't create). Then we wouldn't need to mock a `File` or `FileUtils` and `listFiles` might even work as just a static helper rather than needing to pass in the util as a dependency. What do you think?

collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
```diff
@@ -15,6 +18,32 @@ public void whenThereIsNoFormForgivenFormIdAndVersion_shouldGetAllMethodReturnEm
...
+    }
+
+    @Test
+    public void whenMediaFolderFileIsNull_shouldGetAllMethodReturnEmptyArray() {
+        FormsDao formsDao = mock(FormsDao.class);
+        FileUtil fileUtil = mock(FileUtil.class);
+        String formMediaPath = "/samplePath/";
+
+        when(formsDao.getFormMediaPath("1", "1")).thenReturn(formMediaPath);
+        when(fileUtil.getFileAtPath(formMediaPath)).thenReturn(null);
```
seadowg: I could be missing something but I don't understand how this case can happen? `FileUtil#getFileAtPath` looks like it couldn't return `null`.

seadowg (changes requested):
I think the tests need a bit of rework/discussion here. I'll mark this as ready for testing as I think the core logic change makes sense.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
grzesiek2010: Theoretically you are right but after fixing bugs like https://github.com/getodk/collect/issues/4102 I'm trying to check more than usually. Is that bad? Those are just tests so I think it's ok to tests **more** sometimes.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
grzesiek2010: But then we would need to move this test to android tests you you want to check if a file really exists or not. Would that improve anything?

seadowg: It should work in normal `test` as we're just using `File` rather than any Android specific code right? I think if you just have `FormsDao#getFormMediaPath` return `"/samplePath/"` (like you're doing) then the as that path doesn't exist you should end up creating the environment you need to check that `getAll` returns an empty list.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseMediaFileRepositoryTest.java
seadowg: Yeah true. I guess I'm thinking about if we've not mocked `FileUtil` as we have we can't guarantee that the method won't return `null` from this test.

kkrawczyk123:
**Tested with success!**
Verified on Androids:5.1, 7.1 and 9
Tested cases:
- scenario from issue before migartion
- scenario from issue after migartion
- download before migaration, delete after
- download and delete before migration
- the same scenarios with filled submission

seadowg (approved):

PR #4109, opened 2020-09-22
Restore soft deleted forms that have been re added to the forms folder

### LINKED ISSUE #4103: Deleting filled forms and then redownloading from Google Drive or adding via adb doesn't work
Collect master

#### Problem description
The re-downloaded form from GD is not visible on the Fill Blank form page if this form was filled and deleted

1. User uses Google Drive as a server
2. User downloads form
3. User fills downloaded form
4. User goes to `Delete Saved Form` and removes the form 
5. User goes to `Get Blank Form` and download the same form again
6. Form is not visible on the `Fill Blank Form` list

or

1. Copy form to device using `adb`
1. User fills form
1. User goes to `Delete Saved Form` and removes the form 
1. Copy form to device again using `adb`
1. Form is not visible on the `Fill Blank Form` list

#### Expected behavior
The form is visible, user is able to fill it

seadowg:
Closes #4103

The main change here is that the form now has a `deleted_date` column rather than a `deleted` one. This allows us to check whether the form file on disk is newer than the deleted date or not so we can "restore" soft deleted forms.

Checked manually and wrote new Espresso test.

Comments inline.

Should just fix the problem but good to check form downloading (with servers and google drive) and form deletion with and without filled forms.


collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
```diff
@@ -89,4 +92,25 @@ public void afterFillingAForm_andDeletingIt_allowsFormToBeReDownloaded() {
                 .clickOK(new MainMenuPage(rule))
                 .startBlankForm("One Question");
     }
+
+    @Test
+    public void afterFillingAForm_andDeletingIt_allowsFormToBeReloadedDirectly() {
```
seadowg, line 97: It would be nice to have this tested at a lower level but the disk sync code is definitely not ready for that. For another day.

Honestly it feels like getting a disk (and Google Drive) based  implementation of `FormApi` together would solve a lot of problems. Then code to add and sync forms could be shared between all these sources. I've added a note to `STATE.md` about this.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConstants.java
```diff
@@ -3,7 +3,7 @@
 public class DatabaseConstants {
 
     public static final String FORMS_TABLE_NAME = "forms";
-    public static final int FORMS_DATABASE_VERSION = 9;
+    public static final int FORMS_DATABASE_VERSION = 10;
```
seadowg, line 6: My thinking is we need a migration here as we already have version 9 of the database out in the wild.

collect_app/src/main/java/org/odk/collect/android/database/FormDatabaseMigrator.java
```diff
@@ -223,6 +225,17 @@ private void upgradeToVersion9(SQLiteDatabase db) {
         SQLiteUtils.dropTable(db, temporaryTable);
     }
 
+    private void upgradeToVersion10(SQLiteDatabase db) {
+        String temporaryTable = FORMS_TABLE_NAME + "_tmp";
```
seadowg, line 229: This migration won't actually carry over the state from the old `deleted` column. My assumption is that it's ok for beta users to have their form "restored" if they'd already soft deleted it when they install the update.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConstants.java
lognaturel: Yes, absolutely.

collect_app/src/main/java/org/odk/collect/android/database/FormDatabaseMigrator.java
lognaturel: Yes, fine by me. @opendatakit/testers note that this is expected.

lognaturel (approved):

grzesiek2010:
@seadowg what changed compared to v1.27 that this is regression and we need an additional column in db table to handle this?

lognaturel:
@grzesiek2010 it's not so much an additional column as a modified one. You may have seen that the `deleted` column was added to make soft delete possible when a user wants to not see a form definition but removing it from the system would make certain operations impossible (editing saved instances). We thought we could get away with just a boolean to keep track of whether something was deleted but then we had no way of knowing whether files on the filesystem were ~newer than when the delete happened~ intended to be brought back by the user.

~@seadowg wasn't sure whether to bring this up but did you consider using the form hash instead? Then couldn't the delete state be boolean and if the hash has changed, it's restored?~

seadowg:
> wasn't sure whether to bring this up but did you consider using the form hash instead? Then couldn't the delete state be boolean and if the hash has changed, it's restored?

@lognaturel if you copied over the same form file as before the hash wouldn't change though right?

lognaturel:
> if you copied over the same form file as before the hash wouldn't change though right?

😳 Of course! Sorry, carry on. That's right, I went through that line of thinking yesterday.

seadowg:
@kkrawczyk123 @mmarciniak90 a case that would be interesting to check as part of this would be to check what happens if you delete a form (that you've filled in) and then perform the storage migration. We expect that the form will (magically) reappear, which is probably ok but worth double-checking.

mmarciniak90:
@seadowg 
> We expect that the form will (magically) reappear, which is probably ok but worth double-checking.

There's no magic here :crying_cat_face: Form is not available and I think it's still OK. I see it in migrated files.

kkrawczyk123:
@seadowg @lognaturel there is one weird case: when copy form manually to forms folder, then fill it and delete blank form - when you go back to form folder you can see form and after pasting form once again to that folder and overwrite the existing file it doesn't show up on fill blank form list. I doubt that anyone then will try that but wanted to let you know about that. What do you think?

lognaturel:
> There's no magic here 😿 Form is not available and I think it's still OK. I see it in migrated files.

That's good, we don't want that type of magic! I have a vague memory of @grzesiek2010 (or maybe you?) discovering that file modification timestamp after the migration were different depending on Android version. It sounds like in your case the timestamp didn't change after the migration.

lognaturel:
> the existing file it doesn't show up on fill blank form list

Hmm, I think that's the case I had run into. I guess the filesystem notices they're the same and doesn't update the timestamp? But this doesn't happen if you download the same exact form from Google Drive?

kkrawczyk123:
@lognaturel I have tried:
- Google Drive
- Aggregate
- Central
- form with defined version
- form without defined version
- form with media/without 
- send submission and the same scenario
- send submission and delete both submission and form

and it looked good in all those cases.

seadowg:
> when you go back to form folder you can see form and after pasting form once again to that folder and overwrite the existing file it doesn't show up on fill blank form list.

@kkrawczyk123 what are you using to copy the file in? The test does this with a copy and it's works as expected. I wonder if it's a file browser is doing something "smart" here.

mmarciniak90:
@lognaturel 
> Hmm, I think that's the case I had run into. I guess the filesystem notices they're the same and doesn't update the timestamp? But this doesn't happen if you download the same exact form from Google Drive?

Exactly! When you copy manually it is still the same file with exactly the same timestamp, that's why it is not visible. When you download the same forms from Drive after all steps, it appears with a new timestamp.

lognaturel:
Confirmed that even with `adb` the subsequent push doesn't have any effect. I think if it fixes it for GDrive, it's good enough. As a weird workaround, I can `touch` the file. Or it's also recoverable by deleting blank forms from settings.

mmarciniak90:
> what are you using to copy the file in?

We are using this file manager https://play.google.com/store/apps/details?id=com.ghisler.android.TotalCommander&hl=pl

mmarciniak90:
Tested with success on Android 8.1 ad 10.0

PR #4114, opened 2020-09-23
Update translations for v1.28

lognaturel:
Routine translation update before release. This adds Danish thanks to @Damorck.

Quick note to @Damorck about process -- we pull all strings from Transifex at every release including betas. Our releases generally come about every 2 months. Strings that haven't yet been translated in a language fall back to English.

Ran it and verified that Danish is listed and selectable. Verified that new strings are visible in French language.

No alternative available.

I verified that this only changes text files so there's no risk of behavior change. The biggest risk is that some strings were changed and don't appear in certain languages anymore so users may be slightly confused. However, these are for install id, subscriber id and simserial which are not frequently used.

No.

No.


yanokwa (approved):
LGTM

PR #4115, opened 2020-09-23
Remove code to hide experimental themes in settings

lognaturel:
There's now an Experimental section. The result of this is that the `Dark theme` was being removed. This only was observable if on a tagged release build.

Created a fake release tag, verified that the dark theme is visible. Ran `userInterfaceSettings_shouldBeReset` and verified that it passes. Verified that the Experimental menu is still hidden.

This was a mistake and there aren't any alternatives to fixing it.

This code is straightforward to understand and verify so it should be low risk. We don't want any special behavior based on the build type in the main settings.

No.

No.


seadowg (approved):
Looks good to merge once checks pass!

PR #4117, opened 2020-09-24
Update dependencies

grzesiek2010:
I ran test on firebase and confirmed seems to be fine.

This is what we do after every release: https://github.com/getodk/collect#release-cycle

It doesn't require testing since almost everything might be affected we always test such changes on a daily basis.

No.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,6 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:4.0.1'
         classpath 'com.google.gms:google-services:4.3.4'
-        classpath 'org.ow2.asm:asm:9.0' // https://github.com/jacoco/jacoco/issues/639#issuecomment-355424756
         classpath 'org.jacoco:org.jacoco.core:0.8.6'
         classpath 'com.google.firebase:firebase-crashlytics-gradle:2.3.0'
         classpath 'com.google.firebase:perf-plugin:1.3.1'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -5,14 +5,14 @@ buildscript {
...
+    ext.kotlin_version = '1.4.10'
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.0'
+        classpath 'com.android.tools.build:gradle:4.0.1'
         classpath 'com.google.gms:google-services:4.3.3'
-        classpath 'org.ow2.asm:asm:7.3.1' // https://github.com/jacoco/jacoco/issues/639#issuecomment-355424756
-        classpath 'org.jacoco:org.jacoco.core:0.8.5'
-        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.1.0'
+        classpath 'org.ow2.asm:asm:9.0' // https://github.com/jacoco/jacoco/issues/639#issuecomment-355424756
```
seadowg: Did you investigate if we still need this comment? I'm wondering if we don't as Jacoco is mostly disabled at the moment.

seadowg (changes requested):
Just one comment but looks good!

build.gradle
grzesiek2010: Right, it seems like we don't need it anymore. I removed it.

seadowg (approved):

PR #4118, opened 2020-09-24
Prevent NPE in QuitFormDialogFragment

### LINKED ISSUE #4036: NPE in FormSaveViewModel
From https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8ca03d40871475f87917f721719cd4aa?time=last-seven-days&sessionId=5F429670034E0001384141414334DC52_DNE_0_v2

Collect v1.28.0-beta.1

#### Problem description

Stacktrace:

```java
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'org.odk.collect.android.formentry.audit.AuditEventLogger org.odk.collect.android.javarosawrapper.FormController.getAuditEventLogger()' on a null object reference
       at org.odk.collect.android.formentry.saving.FormSaveViewModel.getAuditEventLogger(FormSaveViewModel.java:253)
       at org.odk.collect.android.formentry.QuitFormDialogFragment.lambda$onCreateDialog$0(QuitFormDialogFragment.java:93)
       at org.odk.collect.android.formentry.-$$Lambda$QuitFormDialogFragment$kkJKBUgJhZUYv_cQAWuh7PWDoCM.onItemClick(-.java:8)
       at android.widget.AdapterView.performItemClick(AdapterView.java:321)
       at android.widget.AbsListView.performItemClick(AbsListView.java:1217)
       at android.widget.AbsListView$PerformClick.run(AbsListView.java:3203)
       at android.widget.AbsListView$3.run(AbsListView.java:4151)
       at android.os.Handler.handleCallback(Handler.java:808)
       at android.os.Handler.dispatchMessage(Handler.java:101)
       at android.os.Looper.loop(Looper.java:166)
       at android.app.ActivityThread.main(ActivityThread.java:7529)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:245)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:921)
```

My guess it that somehow the form is being quit before the `formController` has been set on the `FormSaveViewModel`.

#### Expected behavior

A `null` `formController` shouldn't cause a crash (as much as it wants to).

#### Other information 

I think a good fix for this kind of thing is to move the logging logic to the view model so that it can just no-op on a `null` `formController`. A good idea for this specific one would be to create a `quit` or `ignore` action on the `FormSaveViewModel` that could take care of logging and removing temp instances etc.

seadowg:
Closes #4036

This hopefully fixes the crash but it's hard to validate as we don't have reproduction steps.

Isolated code with bug and wrote new test to cover it.

Comments inline.

Should just prevent the one case where a `null` could be set. I think it's a simple enough change that we can merge without QA but maybe still good for them to check loading forms and then ignoring changes here so we're confident when we do a patch release.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -45,6 +45,7 @@ import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.widget.Toolbar;
@@ -459,7 +460,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         formSaveViewModel.getSaveResult().observe(this, this::handleSaveResult);
     }
 
-    private void formControllerAvailable(FormController formController) {
+    private void formControllerAvailable(@NonNull FormController formController) {
         menuDelegate.formLoaded(formController);
         identityPromptViewModel.formLoaded(formController);
         formEntryViewModel.formLoaded(formController);
@@ -519,8 +520,9 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             formLoaderTask = (FormLoaderTask) data;
         } else if (data == null) {
             if (!newForm) {
-                if (getFormController() != null) {
-                    FormController formController = getFormController();
+                FormController formController = getFormController();
+
+                if (formController != null) {
                     formControllerAvailable(formController);
                     refreshCurrentView();
                 } else {
@@ -529,6 +531,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     showIfNotShowing(FormLoadingDialogFragment.class, getSupportFragmentManager());
                     formLoaderTask.execute(formPath);
                 }
+
                 return;
             }
 
@@ -2291,6 +2294,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         DialogUtils.dismissDialog(FormLoadingDialogFragment.class, getSupportFragmentManager());
 
         final FormController formController = task.getFormController();
+
         if (formController != null) {
             if (readPhoneStatePermissionRequestNeeded) {
                 new PermissionUtils().requestReadPhoneStatePermission(this, true, new PermissionListener() {
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
@@ -19,10 +19,7 @@ import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.IconMenuListAdapter;
 import org.odk.collect.android.adapters.model.IconMenuItem;
 import org.odk.collect.android.analytics.Analytics;
-import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.dao.helpers.InstancesDaoHelper;
-import org.odk.collect.android.external.ExternalDataManager;
-import org.odk.collect.android.formentry.audit.AuditEvent;
 import org.odk.collect.android.formentry.saving.FormSaveViewModel;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.preferences.AdminKeys;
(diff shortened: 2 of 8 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/QuitFormDialogFragment.java
```diff
@@ -84,16 +81,7 @@ public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
                     listener.onSaveChangesClicked();
                 }
             } else {
-                ExternalDataManager manager = Collect.getInstance().getExternalDataManager();
-                if (manager != null) {
-                    manager.close();
-                }
-
-                if (viewModel.getAuditEventLogger() != null) {
```
seadowg, line 92: The weird thing here is that there is `null` protection in here.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -519,8 +520,9 @@ private void loadForm() {
             formLoaderTask = (FormLoaderTask) data;
         } else if (data == null) {
             if (!newForm) {
-                if (getFormController() != null) {
```
seadowg, line 522: This is the one place where it seemed like we could pass a `null` to `formLoaded` as we are checking the singleton field isn't `null` and then calling the method again. Assigning it here means that we only deal with the one value (`null` or not).

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -83,7 +84,11 @@ public FormSaveViewModel(SavedStateHandle stateHandle, Clock clock, FormSaver fo
     }
 
     @Override
-    public void formLoaded(FormController formController) {
+    public void formLoaded(@NotNull FormController formController) {
+        if (formController == null) {
+            Timber.e(new IllegalArgumentException("null FormController passed to formLoaded"));
```
seadowg, line 89: My thinking here is that we can track this in non fatals in Crashlytics and hopefully gain some insight into whats happening (if this continues).

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -127,8 +128,15 @@ public void saveForm(Uri instanceContentURI, boolean shouldFinalize, String upda
...
     // Cleanup when user exits a form without saving
-    public void removeTempInstance() {
+    public void ignoreChanges() {
+        ExternalDataManager manager = Collect.getInstance().getExternalDataManager();
+        if (manager != null) {
+            manager.close();
+        }
+
         if (formController != null && formController.getInstanceFile() != null) {
+            formController.getAuditEventLogger().logEvent(AuditEvent.AuditEventType.FORM_EXIT, true, System.currentTimeMillis());
```
seadowg, line 138: Much easier to protect form `null` in the view model

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -411,6 +411,12 @@ public void replaceRecentFileForQuestion_whenQuestionIndexHasAnswer_onRecreating
         verify(mediaUtils).deleteImageFileFromMediaProvider("blah");
     }
 
+    @Test
+    public void ignoreChanges_whenFormControllerNotSet_doesNothing() {
+        FormSaveViewModel viewModel = new FormSaveViewModel(savedStateHandle, () -> CURRENT_TIME, formSaver, mediaUtils, null);
+        viewModel.ignoreChanges(); // Checks nothing explodes
```
seadowg, line 417: Definitely could be expanded on but this is good enough to give us a harness to prevent a regression on this particular bug.

lognaturel: Haven't you shaken your fist in my general direction about tests without assertions before?! 😅 

More seriously, I'm not really thinking of anything meaningful to assert and I do think it's ok.

lognaturel (approved):
It's not urgent so let's have QA look at it.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
seadowg: Haha probably. Do as I say not as I do right...

But yeah I think this test would ideally check that calls aren't made but here it felt the real focus was just to make sure nothing exploded. There is a comment 😉

lognaturel:
Since we weren't able to get QA on this today and it is increasing in frequency, we've decided to merge and release. I think this may be a symptom of a deeper issue (why is the `formController` `null` so often?) but I can't think of a way to identify it other than putting this fix out there and seeing if other crashes or user-reported problems show up.

I went through carefully one more time. There is a subtle change in that if `formController.getInstanceFile()` is `null`, a form exit audit event won't be written.

lognaturel (approved):

mmarciniak90:
On Friday, I spent some time trying to reproduce this crash but I failed. I haven't established repro steps. That's why I can only confirm that the actions listed below work correctly.

Verified cases on released v1.28.1:
* change theme in the middle of filling form with audit and load form from savepoint
* change language in the middle of filling form with audit and load form from savepoint
* close app in the middle of filling form with audit and load form from savepoint
* close app on hierarchy view and load form from savepoint
* restart device in the middle of filling form with audit and load form from savepoint
* close app in the middle of repeat group with audit and load form from savepoint
* close app in the middle of filling form with audit  -> clear cashe -> no crash
* close app in the middle of filling form with audit and tracking location in background and load form from savepoint
* close app in the middle of filling form with audit and nested groups in background and load form from savepoint
* ignored changes which were loaded from savepoint
* saved changes which were loaded from savepoint

kkrawczyk123:

PR #4123, opened 2020-09-25
Handle widgets waiting for data using WaitingForDataRegistry

### LINKED ISSUE #4122: NPE in FormEntryActivity.java line 2677
Collect v1.28.0

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/d6df6553f58a3d3f828bdfe036cd855d?time=last-seven-days&versions=v1.28.0%20(3921)&sessionEventKey=5F6C5B9200330001597993BC55946D5D_1454417872317206815

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'org.javarosa.form.api.FormEntryPrompt org.odk.collect.android.widgets.QuestionWidget.getFormEntryPrompt()' on a null object reference
       at org.odk.collect.android.activities.FormEntryActivity$10.run(FormEntryActivity.java:2677)
       at android.app.Activity.runOnUiThread(Activity.java:5643)
       at org.odk.collect.android.activities.FormEntryActivity.widgetValueChanged(FormEntryActivity.java:2674)
       at org.odk.collect.android.activities.FormEntryActivity.updateSelectedItems(FormEntryActivity.java:2573)
       at org.odk.collect.android.fragments.dialogs.SelectMinimalDialog.closeDialogAndSaveAnswers(SelectMinimalDialog.java:100)
       at org.odk.collect.android.fragments.dialogs.SelectOneMinimalDialog.onItemClicked(SelectOneMinimalDialog.java:38)
       at org.odk.collect.android.formentry.questions.AudioVideoImageTextLabel.lambda$setTextView$0(AudioVideoImageTextLabel.java:118)
       at org.odk.collect.android.formentry.questions.-$$Lambda$AudioVideoImageTextLabel$MqCYJ2yJoQ6DnhH04sxqp-kc3wc.onClick(-.java)
       at android.view.View.performClick(View.java:5265)
       at android.widget.CompoundButton.performClick(CompoundButton.java:122)
       at android.view.View$PerformClick.run(View.java:21534)
       at android.os.Handler.handleCallback(Handler.java:815)
       at android.os.Handler.dispatchMessage(Handler.java:104)
       at android.os.Looper.loop(Looper.java:207)
       at android.app.ActivityThread.main(ActivityThread.java:5737)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:789)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:679)
```

grzesiek2010:
Closes #4122

I tested the app manually.

WaitingForDataRegistry is what we use in other places to handle widgets waiting for data.

Since we are not able to reproduce the issue we can merge without testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
@@ -160,7 +160,7 @@ public class WidgetFactory {
                 // considered in each widget by calls to ExternalDataUtil.getSearchXPathExpression.
                 // This means normal appearances should be put before search().
                 if (appearance.contains(WidgetAppearanceUtils.MINIMAL)) {
-                    questionWidget = new SelectOneMinimalWidget(context, questionDetails, isQuick);
+                    questionWidget = new SelectOneMinimalWidget(context, questionDetails, isQuick, waitingForDataRegistry);
                 } else if (appearance.contains(WidgetAppearanceUtils.LIKERT)) {
                     questionWidget = new LikertWidget(context, questionDetails);
                 } else if (appearance.contains(WidgetAppearanceUtils.LIST_NO_LABEL)) {
@@ -180,7 +180,7 @@ public class WidgetFactory {
                 // considered in each widget by calls to ExternalDataUtil.getSearchXPathExpression.
                 // This means normal appearances should be put before search().
                 if (appearance.contains(WidgetAppearanceUtils.MINIMAL)) {
-                    questionWidget = new SelectMultiMinimalWidget(context, questionDetails);
+                    questionWidget = new SelectMultiMinimalWidget(context, questionDetails, waitingForDataRegistry);
                 } else if (appearance.startsWith(WidgetAppearanceUtils.LIST_NO_LABEL)) {
                     questionWidget = new ListMultiWidget(context, questionDetails, false);
                 } else if (appearance.startsWith(WidgetAppearanceUtils.LIST)) {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectMinimalWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectMinimalWidget.java
@@ -6,19 +6,20 @@ import android.view.View;
 
 import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
-import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.databinding.SelectMinimalWidgetAnswerBinding;
 import org.odk.collect.android.formentry.questions.QuestionDetails;
-import org.odk.collect.android.javarosawrapper.FormController;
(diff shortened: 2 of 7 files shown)


lognaturel:
> The problem is related to null form controller

I'm not seeing that relationship. I see "Attempting to return data to a widget or set of widgets not looking for data" in logs before the crash so that means the `formController` was non-null right before. It's all from `SelectOneMinimalDialog.onItemClicked` so I think there might be a bug there, maybe concurrency related. I see examples where it happens right after opening a form and others where it's deep in a form. I think we should hold off on making this change.

grzesiek2010:
Yes just before the controller object is not null but I thought that maybe the scenario was:
1. Form is loaded
2. A user opens the dialog with options
3. FormController is cleared somehow (that's what is a mystery for us)
4. A user closes the dialog/goes back to the form
5. FormController is reloaded so it's not null but the info about the widget waiting for data is lost

lognaturel:
Very interesting, that does seem plausible. I don't think we've seen this with other widget types that get binary data, right? It feels like there's something specific to selects. I continue to think it would be good to collect more traces and do more experimentation before we silence the issue. I'm sure QA tried things with rotation, Do Not Save Activities, etc, but is there some other way that the selection dialog could be leading to the `FormController` going out of memory?

> FormController is reloaded so it's not null

Where would that be happening?

grzesiek2010:
@lognaturel 
I updated the pr removing any new null checks. There is just refactoring to use WaitingForDataRegistry in SelectminimalWidget and [a new timber log](https://github.com/getodk/collect/pull/4123/files#diff-17c1dca497f11810d344a9f3f190c524R20) that might help us to track down the root cause. It would be good to add this pr to the point release.

The new log will give us an answer whether it's caused by null form controller. If not i think that this issue might be related: 
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/85177cf7b388af8b6ec086cd77c7cc05?time=last-ninety-days&versions=v1.28.1%20(3923)&sessionEventKey=5F7DE6010053000120243B374972D6DE_1459356560332557638
the issue is caused by a different amount of questions on the screen so maybe some questions are lost somehow, maybe the one we marked as waiting for data might be lost.

lognaturel (approved):

lognaturel:
Thanks, @grzesiek2010, that's great. I've been looking at that index out of bounds issue as well and your comments give me an idea. I wonder whether maybe there's a relevance condition based on the selected answer. We may not correctly account for question indexes changing in that case.

I have some other work I need to prioritize but will experiment with that before doing the point release.

grzesiek2010:
I tested field lists with relevance condition when I looked at that issue for the first time to no avail but you can try of course maybe you will have better luck.

lognaturel:
Great that you thought of that, @grzesiek2010. No luck reproducing on my end either. Let's see what this new logging tells us.

PR #4140, opened 2020-10-05
Fixed loading downloaded media files

### LINKED ISSUE #4139: The app crashes when a user tries to attach downloaded media files on Android 10
Collect v1.x.x, Android vx.x.x, device used...

#### Problem description
The app crashes when a user tries to attach downloaded media files on Android 10
it's what I discovered thanks to firebase reports:
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5321cc71b505e4165f16c7ea6223e44b?time=last-seven-days&versions=v1.28.1%20(3923)&sessionEventKey=5F79A2DE0062000179EACD3BF8DFBB82_1458154635563220359

1. Open the All widgets form
2. Navigate to the ImageWidget
3. Try to attach an image that you previously downloaded

Screenshot_1601887944

grzesiek2010:
Closes #4139

I was able to reproduce the issue and confirmed the fix does the job.

This is the only solution. We just need to handle URIs that contain `msf` prefixes.

It's a safe fix so we can just test the described scenario, nothing else should be affected.

Any form with media files like the All widgets form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
@@ -449,9 +449,12 @@ public class MediaUtils {
             } else if (isDownloadsDocument(uri)) {
                 // DownloadsProvider
 
-                final String id = DocumentsContract.getDocumentId(uri);
+                String id = DocumentsContract.getDocumentId(uri);
                 if (id.startsWith("raw:")) {
                     return id.replaceFirst("raw:", "");
+                }
+                if (id.startsWith("msf:")) {
+                    id = id.replaceFirst("msf:", "");
                 }
 
                 String[] contentUriPrefixesToTry = {


lognaturel (approved):
I know this code comes from elsewhere but it seems like a very strange way to do things! I think we should move forward with this since it's affecting users and not worry too much about it for now but at some point it seems like we should be able to simplify this and not have hardcoded download folder prefixes. Can we really not use the URI directly to resolve to a real path?

grzesiek2010:
> I know this code comes from elsewhere but it seems like a very strange way to do things! I think we should move forward with this since it's affecting users and not worry too much about it for now but at some point it seems like we should be able to simplify this and not have hardcoded download folder prefixes. Can we really not use the URI directly to resolve to a real path?

I wish there was one simple way to resolve the issue and maybe there is something but for now I would be afraid to touch it since it handles so many cases. We can do that after having a point release or generally later.

mmarciniak90:
**Tested with success**

Verified on Android 10.0, checked also on Android 8.1 and 7.0

Verified cases:
* Image widget
* Annotate widget
* File widget
* Video widget
* rotation
* minimize
* RTL language
* light/dark mode
* new/saved form

kkrawczyk123:
Verified on Androids:
- device with Android 10
- emulator with Android 10
- quick check on Androids: 5.1, 7.1 and 9.0

PR #4141, opened 2020-10-05
Alternative fix for redownloading delete forms from Google Drive

### LINKED ISSUE #4124: Look into flakey Espresso tests
https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/7206684696906624618/executions/bs.10b8f2cb9c5496c6/testcases/3/test-cases

https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/7206684696906624618/executions/bs.10b8f2cb9c5496c6/testcases/3/test-cases

seadowg:
Closes #4124.

Alternative fix for #4103. Instead of the (we suspect) flakey solution that allows forms to be restored if they are modified on disk this just explicitly restores forms downloaded from Google Drive.

Not much at all. I figured it is best to let QA check it out rather than spending time setting up Google Drive for myself locally (I believe they have it set up already).

If there are any problems I can come back to this and take the time to set up a proper environment.

On testing, it feels to me like we need to rip apart a bunch of the Google Drive code and have it use the use case/method objects we're building like `ServerFormDownloader` so we can write tests for it without spending a lot more time implementing stub Google Drive code for a feature we don't consider "production ready". Going down that road right now doesn't seem right so I've left this change untested so we can just move on.

Discussed before but explicitly restoring forms like this avoids us relying on last modified data on files which can cause timing issues in tests (and potentially in real life).

Should be no effect but I think we should check this as if we were fixing the issue again.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formmanagement/DeleteBlankFormTest.java
@@ -2,7 +2,6 @@ package org.odk.collect.android.feature.formmanagement;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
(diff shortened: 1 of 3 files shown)


lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified on Android: 7.0, 8.1, 9.0, 10.0

Verified cases:
* case from the main issue on Google Drive
* form with defined version
* form without defined version
* form with media/without
* send a submission and the same scenario
* case from the main issue on Aggregate
* case from the main issue on Central

Noticed one change, it's positive because the result is the same as in 1.27.3 and the same as in Aggregate
1. User downloads a form and fills it
2. User removes form
3. User downloads again the same form and opens Fill Blank Form view
Result on PR: the time of downloading the first form is visible
Result on MASTER: current time is visible

PR #4146, opened 2020-10-06
Use latest Circle CI Android image

seadowg:
This makes the min SDK used across the app consistent and bumps us up to using the latest Circle CI image so we get more cache hits when pulling the image in.

### DIFF
--- a/async/build.gradle
+++ b/async/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 30
 
     defaultConfig {
         minSdkVersion 21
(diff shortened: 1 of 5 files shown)


async/build.gradle
```diff
@@ -3,11 +3,11 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 28
```
lognaturel: Why not make this 30?

collect_app/build.gradle
```diff
@@ -68,16 +68,16 @@ def googleMapsApiKey = secrets.getProperty('GOOGLE_MAPS_API_KEY', '')
 def mapboxToken = secrets.getProperty('MAPBOX_ACCESS_TOKEN', '')
 
 android {
-    compileSdkVersion(28)
+    compileSdkVersion 28
```
lognaturel: 30?

lognaturel (commented):
This looks ready to me except for the compile version!

lognaturel (approved):

Issue #4128, opened 2020-09-29
Crashlytics stacktrace when repeat is in field list


lognaturel:
#### Software and hardware versions 
Collect v1.28

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/960eea5fd9eed2e302699052950fdaee?time=last-seven-days&versions=v1.28.1%20(3923)&sessionEventKey=5F72DDCB02D10001502FDDFB220A9324_1456249272105355095

```
Fatal Exception: java.lang.RuntimeException
Only questions and regular groups are allowed in 'field-list'. Bad node is: /data/abc/xyz
org.odk.collect.android.javarosawrapper.FormController.getQuestionPrompts (FormController.java:922)
org.odk.collect.android.activities.FormEntryActivity.updateFieldListQuestions (FormEntryActivity.java:2708)
org.odk.collect.android.activities.FormEntryActivity.access$1900 (FormEntryActivity.java:200)
org.odk.collect.android.activities.FormEntryActivity$10.run (FormEntryActivity.java:2681)
android.app.Activity.runOnUiThread (Activity.java:6045)
org.odk.collect.android.activities.FormEntryActivity.widgetValueChanged (FormEntryActivity.java:2678)
org.odk.collect.android.formentry.ODKView.widgetValueChanged (ODKView.java:621)
org.odk.collect.android.widgets.QuestionWidget.widgetValueChanged (QuestionWidget.java:499)
org.odk.collect.android.widgets.items.SelectMultiWidget.onItemClicked (SelectMultiWidget.java:85)
org.odk.collect.android.adapters.SelectMultipleListAdapter.lambda$createButton$0 (SelectMultipleListAdapter.java:106)
org.odk.collect.android.adapters.-$$Lambda$SelectMultipleListAdapter$apSaKhIavlUfvDCnua9cTtxGgo4.onCheckedChanged (-.java)
```

#### Steps to reproduce the problem
I can reproduce a dialog that shows this exception's message by opening a form like [this one](https://docs.google.com/spreadsheets/d/159jGIYoZ5-f1JQTAWyfUFqk9-3k4mmbaM14KgbKbCbA/edit?usp=sharing) where there's a repeat in a field-list. I believe that there's no hard crash but that the `Timber.e` call on line 1253 in `FormEntryActivity` is what's making us see the log in Crashlytics. I think we just need to remove that call (and others like it).

#### Expected behavior
[...]

PR #4147, opened 2020-10-06
Catch updateFieldListQuestions when updating field list

### LINKED ISSUE #4145: RuntimeException in FormController.java:922
#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/960eea5fd9eed2e302699052950fdaee?time=last-seven-days&versions=v1.28.1%20(3923)&sessionEventKey=5F7B186800E50001483327869100C764_1458566851285633918

```
Fatal Exception: java.lang.RuntimeException: Only questions and regular groups are allowed in 'field-list'.  Bad node is: /aGgNSwSLxKcrJWY8NtVfjt/Section_B/B_chitralada/asset
       at org.odk.collect.android.javarosawrapper.FormController.getQuestionPrompts(FormController.java:922)
       at org.odk.collect.android.activities.FormEntryActivity.updateFieldListQuestions(FormEntryActivity.java:2708)
       at org.odk.collect.android.activities.FormEntryActivity.access$1900(FormEntryActivity.java:200)
       at org.odk.collect.android.activities.FormEntryActivity$10.run(FormEntryActivity.java:2681)
       at android.app.Activity.runOnUiThread(Activity.java:5289)
       at org.odk.collect.android.activities.FormEntryActivity.widgetValueChanged(FormEntryActivity.java:2678)
       at org.odk.collect.android.formentry.ODKView.widgetValueChanged(ODKView.java:621)
       at org.odk.collect.android.widgets.QuestionWidget.widgetValueChanged(QuestionWidget.java:499)
       at org.odk.collect.android.widgets.items.SelectMultiWidget.onItemClicked(SelectMultiWidget.java:85)
       at org.odk.collect.android.adapters.SelectMultipleListAdapter.lambda$createButton$0(SelectMultipleListAdapter.java:106)
       at org.odk.collect.android.adapters.-$$Lambda$SelectMultipleListAdapter$apSaKhIavlUfvDCnua9cTtxGgo4.onCheckedChanged(-.java)
```

grzesiek2010:
Closes #4145

Nothing since I wan't able to reproduce the issue.

I wasn't able to reproduce the issue so I decided to just catch it. Generally it's because of having repeats inside field-list group what we don't support. We catch this scenario in other parts of our code https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L1250 and display a dialog. I don't know why updateFieldListQuestions() is called before what cause the crash I guess it might be race condition problem. Catching the exception and doing noting in that case is totally fine.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2678,7 +2678,11 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
+                    try {
+                        updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
+                    } catch (RuntimeException e) {
+                        Timber.w(e);
+                    }
 
                     odkView.addOnLayoutChangeListener(new View.OnLayoutChangeListener() {
                         @Override


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2678,7 +2678,11 @@ public void widgetValueChanged(QuestionWidget changedWidget) {
             runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
+                    try {
+                        updateFieldListQuestions(changedWidget.getFormEntryPrompt().getIndex());
+                    } catch (RuntimeException e) {
+                        Timber.w(e);
```
lognaturel, line 2684: Let's do a `Timber.i` so it doesn't go to Crashlytics. I don't think we can do anything with the information.

lognaturel:
Did you see my note in #4128? 

> I believe that there's no hard crash but that the Timber.e call on line 1253 in FormEntryActivity is what's making us see the log in Crashlytics. I think we just need to remove that call (and others like it).

What do you think about that quick analysis? Or do you think the dialog is shown as expected but there's a special race condition case where there's an actual crash?

I'm not totally clear on the effect of silently catching this kind of exception. It seems like it leaves the form in a weird state.

grzesiek2010:
I'm pretty sure it can't be caused by the way we use Timber. Even if we call Tmber.e such reports don't end'up in `crashes` category but in `non-fatals`. Here you have that particular report: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/05043cc4ecb58cc8956f42be36541558?time=last-seven-days&versions=v1.28.1%20(3923)&sessionEventKey=5F7D682101B100012595A2208C47F942_1459215957434772665

> Or do you think the dialog is shown as expected but there's a special race condition case where there's an actual crash?

Yes I think that updateFieldListQuestions() might be called after we catch the exception and display a dialog but before we move to next questions in:
https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L1258
or maybe somehow before... everything is possible I can't be sure so I'm ok if you prefer to leave it as is for now.

grzesiek2010:
Let's close this for now.

PR #4148, opened 2020-10-06
Catch and display error when new repeat created

### LINKED ISSUE #4119: Crash on RuntimeException when adding repeat
Collect v1.27.4; [Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5b4d80d96007d59fcdcad981?time=last-seven-days&sessionEventKey=5F6CCB66029800011C980B8DF8617D36_1454540976116208809) shows it occurred back to at least v1.25.2

#### Problem description
If a form has a reference to a value inside of a repeat that is not qualified (doesn't specify a specific repeat instance) and is used in a context where a single node is expected, there's a hard crash.

1. Load a form such as [bad-instance-name.xml.zip](https://github.com/getodk/collect/files/5280566/bad-instance-name.xml.zip)
1. Add a first repeat instance
1. Attempt to add a second repeat instance and experience a crash

#### Expected behavior
The user should see the text from the `RuntimeException` ~and the form should close~.

#### Additional information
Validate should catch these kinds of issues so it's likely these forms aren't being built with XLSForm or converted with pyxform-based tools.

I paid close attention to the crashes that happened in v1.28 ([Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5b4d80d96007d59fcdcad981?time=1599609600000:1600991999000&sessionEventKey=5F6C63F400160001361633C70654677F_1454426636061422683)) to see if there was anything about them that could point to a change in behavior. It looks like a single form that has a structure like [relevant-repeat.xml.zip](https://github.com/getodk/collect/files/5280608/relevant-repeat.xml.zip).

seadowg:
Closes #4119

Add a new test and tried out example form.

We already had a lot of the plumbing to make this work in so it ended up being a very simple fix.

Should just fix the issue but good to check there aren't weird side effects in cases where the error shows now we aren't just crashing out.

The form in the issue is the best starting point.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
@@ -80,13 +80,17 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
             analytics.logEvent(ADD_REPEAT, "Hierarchy", formController.getCurrentFormIdentifierHash());
         }
 
-        formController.newRepeat();
+        try {
+            formController.newRepeat();
+        } catch (RuntimeException e) {
+            error.setValue(e.getCause().getMessage());
+        }
         
         if (!formController.indexIsInFieldList()) {
             try {
                 formController.stepToNextScreenEvent();
-            } catch (JavaRosaException exception) {
-                error.setValue(exception.getCause().getMessage());
+            } catch (JavaRosaException e) {
+                error.setValue(e.getCause().getMessage());
             }
         }
     }
--- a/collect_app/src/test/java/org/odk/collect/android/formentry/FormEntryViewModelTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formentry/FormEntryViewModelTest.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.mockito.Mockito.atMostOnce;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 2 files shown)


lognaturel (approved):

mmarciniak90:
@seadowg take a look at my test results, please. Let me know if all described behaviors are expected.

- user loads the form, swipes to the next question and adds a new group -> error is visible -> user is able to continue the form and finalize it
- user loads the form, user clicks on :heavy_plus_sign: icon, then clicks on `Add` new group -> error is visible -> user is able to continue the form and finalize it
- user loads form, swipes to the next question, adds a new group -> user opens hierarchy view and clicks on :heavy_plus_sign: icon on hierarchy view -> error is visible -> but then **user is moved to main menu**
- user loads form, opens hierarchy view and clicks on :heavy_plus_sign: icon on hierarchy view to add a new group -> error is **NOT** visible -> adding groups only by :heavy_plus_sign: icon at hierarchy view cause that error is not visible at all

lognaturel:
> but then user is moved to main menu

I'm ok with this. The form design is faulty so I don't think we need to have any particular guaranteed behavior. We just want to make sure the form designer knows what's going on.

> adding groups only by ➕ icon at hierarchy view cause that error is not visible at all

I certainly didn't expect this. But I think it's ok -- not allowing repeats in a field list is because we don't have any reasonable way to display them at the moment. As @grzesiek2010 mentioned it's something we could add with relative ease. Using only the hierarchy view bypasses the display issue.

I'm personally satisfied with these results.

mmarciniak90:
@lognaturel thanks for the clarification.

PR was tested with success.

Verified on Android versions: 7.0, 7.1, 8.1, 9.0, 10.0

PR #4156, opened 2020-10-07
Don't attempt to stop location updates if GoogleAPIClient is not already connected

### LINKED ISSUE #4155: Crash on attempt to stop location updates when GoogleApiClient is not yet connected
Collect v1.28

#### Problem description
```
Caused by java.lang.IllegalStateException
GoogleApiClient is not connected yet.
com.google.android.gms.common.api.internal.zaat.execute (zaat.java:12)
com.google.android.gms.internal.location.zzq.removeLocationUpdates (zzq.java:5)
org.odk.collect.android.location.client.GoogleFusedLocationClient.stopLocationUpdates (GoogleFusedLocationClient.java:135)
org.odk.collect.android.location.client.GoogleFusedLocationClient.stop (GoogleFusedLocationClient.java:107)
org.odk.collect.android.activities.GeoPointActivity.onStop (GeoPointActivity.java:163)
android.app.Instrumentation.callActivityOnStop (Instrumentation.java:1473)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:980)
```

#### Other information 

grzesiek2010:
Closes #4155 

I tested the app with hardcoded disconnecting GoogleApiClient and added tests. 

We just need to check if the client is connected before performing some operations.

It doesn't require testing since it's difficult to reproduce.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
+++ b/collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
@@ -120,7 +120,7 @@ public class GoogleFusedLocationClient
 
     @SuppressLint("MissingPermission") // Permission checks for location services handled in widgets
     public void requestLocationUpdates(@NonNull LocationListener locationListener) {
-        if (!isMonitoringLocation()) {
+        if (!isMonitoringLocation() && googleApiClient.isConnected()) {
             fusedLocationProviderApi.requestLocationUpdates(googleApiClient, createLocationRequest(), this);
         }
 
@@ -128,7 +128,7 @@ public class GoogleFusedLocationClient
     }
 
     public void stopLocationUpdates() {
-        if (!isMonitoringLocation()) {
+        if (!isMonitoringLocation() || !googleApiClient.isConnected()) {
             return;
         }
 
@@ -141,7 +141,7 @@ public class GoogleFusedLocationClient
     public Location getLastLocation() {
         // We need to block if the Client isn't already connected:
         if (!googleApiClient.isConnected()) {
-            googleApiClient.blockingConnect();
+            new Thread(googleApiClient::blockingConnect).start();
         }
 
         return fusedLocationProviderApi.getLastLocation(googleApiClient);
--- a/collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
+++ b/collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
@@ -128,11 +128,13 @@ public class GoogleFusedLocationClient
     }
 
     public void stopLocationUpdates() {
-        if (!isMonitoringLocation() || !googleApiClient.isConnected()) {
+        if (!isMonitoringLocation()) {
             return;
         }
 
-        fusedLocationProviderApi.removeLocationUpdates(googleApiClient, locationListener);
+        if (googleApiClient.isConnected()) {
+            fusedLocationProviderApi.removeLocationUpdates(googleApiClient, locationListener);
+        }
         locationListener = null;
     }
 
@@ -141,7 +143,7 @@ public class GoogleFusedLocationClient
     public Location getLastLocation() {
         // We need to block if the Client isn't already connected:
         if (!googleApiClient.isConnected()) {
-            new Thread(googleApiClient::blockingConnect).start();
+            googleApiClient.blockingConnect();
         }
 
         return fusedLocationProviderApi.getLastLocation(googleApiClient);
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
```diff
@@ -141,7 +141,7 @@ public void stopLocationUpdates() {
     public Location getLastLocation() {
         // We need to block if the Client isn't already connected:
         if (!googleApiClient.isConnected()) {
-            googleApiClient.blockingConnect();
+            new Thread(googleApiClient::blockingConnect).start();
```
grzesiek2010: This is a fix for another issue https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8bf1ac2f771699358e0237615316760d?time=last-seven-days&versions=v1.28.1%20(3923)&sessionEventKey=5F7C02C1035300012FE4B899DA981953_1458846445087682870

seadowg: I'm a little concerned about this. If the client isn't connected, and we spin up a new thread to connect could that cause problems for the call to `getLastLocation` (as it might happen before the client is connected)?

Also, it's likely the testing of this in `GoogleFusedLocationClientTest` (using a `timeout`) will be flakey as the call to `blockingConnect` is now async (and make the behaviour of the method indeterminate). 

Could we maybe pull this fix out of the PR and look at it as a part of a different PR?

seadowg (changes requested):
Just concerned about the other fix bundled into this PR but happy with the fix for the issue!

collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
grzesiek2010: I tested it hardcoding `googleApiClient.disconnect();` just before `if (!googleApiClient.isConnected()) {` and it didn't cause any crashes. @lognaturel what do you think should we exclude this fix?

collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
```diff
@@ -120,15 +120,15 @@ public void stop() {
...
+        if (!isMonitoringLocation() && googleApiClient.isConnected()) {
             fusedLocationProviderApi.requestLocationUpdates(googleApiClient, createLocationRequest(), this);
         }
 
         this.locationListener = locationListener;
     }
 
     public void stopLocationUpdates() {
-        if (!isMonitoringLocation()) {
+        if (!isMonitoringLocation() || !googleApiClient.isConnected()) {
```
lognaturel: I think it's possible for a `locationListener` to be set and `googleApiClient` not to be connected. In that case, `locationListener` would not be set to `null`. I think that's not a good state to be in. Just to be safe, I'd wrap only the `removeLocationUpdates` call.

grzesiek2010: Ah you mean that locationListener won't be cleared in such a case?

collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
lognaturel: If we can't block the UI thread, then this code structure doesn't makes sense. That is, like @seadowg says, if you're doing it in a different thread, that thread is blocked, but you still attempt to `getLastLocation` on the UI thread which could fail if the connection isn't instant. I think you could use `CoroutineAndWorkManagerScheduler.immediate(background, foreground)`. Splitting it off to another PR sounds best to me.

collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
lognaturel: Whoops, missed this. Yes, exactly. That means `isMonitoringLocation()` would evaluate to `true` even after `stopLocationUpdates` has been called. I don't know if that really matters currently because I think we only `stopLocationUpdates` when we're closing out activities but I'd rather not risk it.

grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/location/client/GoogleFusedLocationClient.java
grzesiek2010: Ok I revoked that change.

lognaturel (approved):

lognaturel:
Ooh, I can't merge! @seadowg you'll need to accept and merge.

seadowg (approved):

PR #4166, opened 2020-10-09
Define new style for Image Button in Magenta Theme

### LINKED ISSUE #4163: All icons on map view look like active when Magenta theme is set
Collect master
Magenta theme

#### Problem description
All icons on map view look like active when Magenta theme is set

1. User fills `All widgets` form
2. User goes to some Geo widget with map view
3. User opens map and sees that all buttons look like active

|Screenshot_20201009-122145| Screenshot_20201009-122159| Screenshot_20201009-122212|
|---|---|---|
|Screenshot_20201009-122306| Screenshot_20201009-122322| Screenshot_20201009-122335|

#### Expected behavior
User sees active and inactive buttons on map view for Geo widget

SaumiaSinghal:
Closes #4163 

I tested it.
<table>
  <tr>
    <td> </td>
    <td></td>
    <td></td>
  </tr>
</table>

I defined a new tint list `map_button_tint_list.xml` a new style `Widget.Collect.ImageButton.MapButton` that uses this tint list, to set the color of the map buttons to `colorOnSurface` for both enabled and disabled state (with reduced opacity).

This change introduces a new style that is used by the map buttons on `geo_point_layout`, `geo_poly_layout` and `instance_map_layout`.

Any form having GeoWidgets or All Widgets Form.

No

### DIFF
--- a/collect_app/src/main/res/color/magenta_theme_tint_list.xml
+++ b/collect_app/src/main/res/color/magenta_theme_tint_list.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="#81ffffff" android:state_enabled="false" />
+    <item android:color="#de000000" android:state_enabled="true" />
+</selector>
--- a/collect_app/src/main/res/color/magenta_theme_tint_list.xml
+++ b/collect_app/src/main/res/color/magenta_theme_tint_list.xml
@@ -1,5 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<selector xmlns:android="http://schemas.android.com/apk/res/android">
-    <item android:color="#81ffffff" android:state_enabled="false" />
-    <item android:color="#de000000" android:state_enabled="true" />
-</selector>
--- a/collect_app/src/main/res/values/themes.xml
+++ b/collect_app/src/main/res/values/themes.xml
@@ -11,4 +11,10 @@
         <item name="android:buttonStyle">@style/Widget.Collect.Button.Custom</item>
         <item name="imageButtonStyle">@style/Widget.Collect.ImageButton.Custom.Light</item>
     </style>
+
+    <style name="Theme.Collect.Magenta" parent="Theme.Collect.BaseMagenta">
+        <item name="android:buttonStyle">@style/Widget.Collect.Button.Custom</item>
+        <item name="imageButtonStyle">@style/Widget.Collect.ImageButton.Custom.Magenta</item>
+    </style>
+
 </resources>
--- a/collect_app/src/main/res/values/themes.xml
+++ b/collect_app/src/main/res/values/themes.xml
@@ -12,9 +12,4 @@
         <item name="imageButtonStyle">@style/Widget.Collect.ImageButton.Custom.Light</item>
     </style>
 
-    <style name="Theme.Collect.Magenta" parent="Theme.Collect.BaseMagenta">
-        <item name="android:buttonStyle">@style/Widget.Collect.Button.Custom</item>
-        <item name="imageButtonStyle">@style/Widget.Collect.ImageButton.Custom.Magenta</item>
-    </style>
-
 </resources>
--- a/collect_app/src/main/res/color/map_button_tint_list.xml
+++ b/collect_app/src/main/res/color/map_button_tint_list.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="?attr/colorOnSurface" android:state_enabled="true"/>
+    <item android:alpha="0.12" android:color="?attr/colorOnSurface" android:state_enabled="false"/>
+</selector>
--- a/collect_app/src/main/res/layout/geopoint_layout.xml
+++ b/collect_app/src/main/res/layout/geopoint_layout.xml
@@ -50,6 +50,7 @@
 
             <androidx.appcompat.widget.AppCompatImageButton
                 android:id="@+id/place_marker"
+                style="@style/Widget.Collect.ImageButton.MapButton"
                 android:layout_width="wrap_content"
                 android:layout_height="match_parent"
                 android:layout_margin="5dp"
@@ -59,6 +60,7 @@
 
             <androidx.appcompat.widget.AppCompatImageButton
                 android:id="@+id/zoom"
+                style="@style/Widget.Collect.ImageButton.MapButton"
                 android:layout_width="wrap_content"
                 android:layout_height="match_parent"
                 android:layout_margin="5dp"
@@ -68,6 +70,7 @@
 
             <androidx.appcompat.widget.AppCompatImageButton
                 android:id="@+id/layer_menu"
+                style="@style/Widget.Collect.ImageButton.MapButton"
                 android:layout_width="wrap_content"
                 android:layout_height="match_parent"
                 android:layout_margin="5dp"
@@ -77,6 +80,7 @@
 
             <androidx.appcompat.widget.AppCompatImageButton
                 android:id="@+id/clear"
+                style="@style/Widget.Collect.ImageButton.MapButton"
                 android:layout_width="wrap_content"
                 android:layout_height="match_parent"
                 android:layout_margin="5dp"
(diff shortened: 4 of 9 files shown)


collect_app/src/main/res/values/themes.xml
```diff
@@ -11,4 +11,10 @@
         <item name="android:buttonStyle">@style/Widget.Collect.Button.Custom</item>
         <item name="imageButtonStyle">@style/Widget.Collect.ImageButton.Custom.Light</item>
     </style>
+
+    <style name="Theme.Collect.Magenta" parent="Theme.Collect.BaseMagenta">
```
seadowg: I'd like to get away from having to use a base theme and a theme. Do we really need to have all image buttons and buttons take on custom styles? I'd rather just have the buttons in the map view explicitly add this style. 

It might seem nice to have things like this happen automatically but the trade-off is that it can be hard to work out why a view looks a certain way if you don't know a global customization is in place. Especially in a code base of this size!

collect_app/src/main/res/color/magenta_theme_tint_list.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
seadowg: I think this should be more explicitly named so it's obvious it's for the image buttons.

seadowg (changes requested):

seadowg:
@SaumiaSinghal For "How does this change affect users?" you've answered "No" but this changes every `ImageButton` in the app which could of course have some unintended consequences. 

Just a reminder that the question isn't trying to catch you out or be antagonistic: it's just asking you to highlight what has changed so QA and reviewers have a heads up. Even if I can't think of a problem I might have introduced I would always try to think about what areas of the app might be affected by my changes so that the reviewer and QA can also think about that.

collect_app/src/main/res/values/themes.xml
SaumiaSinghal: Do you mean for all the other themes, light, dark, and magenta, color for enabled buttons should be defined as a separate value in the theme.xml file, and then I can move away from `Theme.Collect.BaseLight` and `Theme.Collect.BaseDark` and have all the image buttons in map view use a style that defines custom colors for the enabled and disabled state like `@color/state_enabled` and like `@color/state_disabled`? 

This would define the style specifically for the image buttons in the map view. I am just skeptical about adding a new color resource in the themes file. Is this the best way to go about it?

SaumiaSinghal:
Yes sorry, I will keep that in mind. Once this is finalized, I'll update the answers in the PR description.

seadowg:
@SaumiaSinghal I think you might have forgotten to push your changes before hitting "request review" (I've done that so many times). The PR description also doesn't look like it's changed.

SaumiaSinghal:
Oh sorry @seadowg! Actually, I just wanted to confirm if I am thinking in the right direction: https://github.com/getodk/collect/pull/4166#discussion_r504411367

seadowg:
@SaumiaSinghal oh weird. I only got one notification so assumed it was the PR review. That's a bit of a confusing feature from Github there! Will look at your comment now.

collect_app/src/main/res/values/themes.xml
seadowg: No I don't mean changing the light and dark theme setup. Magenta is an experiment to see how we can simplify theming and then we'll probably move to setting Dark/Light up in the same way after. 

These buttons are a a bit of weird setup as they should really be [Floating Action Buttons](https://material.io/components/buttons-floating-action-button) but I don't think we want to make that change now. For the moment I think we should give them a custom style (could be `MapButton`) and then define a tint list specific to that. We shouldn't need to use custom colors here either: I think we should follow what [Material Components does for the FAB](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/floatingactionbutton/res/color/mtrl_fab_bg_color_selector.xml) and use `colorSecondary` for enabled and `colorOnSurface` with a low alpha (so basically transparent) for disabled.

seadowg (changes requested):
Comments inline

collect_app/src/main/res/values/themes.xml
SaumiaSinghal: Oh okay! That seems like a better solution. Thanks!

SaumiaSinghal:
I defined a new style for MapButton, and it uses `colorOnSurface` for both enabled and disabled states (with a low alpha for the disabled state). I tried using `colorSecondary` as well for enabled. It looks a bit more radiant, should we go with `colorOnSurface` for now?
| Magenta Theme         |  Light Theme | DarkTheme |
:-------------------------:|:-------------------------:|:-------------------------:
Screenshot_1603111647  |  Screenshot_1603111679 | Screenshot_1603111707
Screenshot_1603112032  |  Screenshot_1603112004 | Screenshot_1603111972

seadowg:
I'm actually really confused as to how the background for these buttons ends up being grey in Magenta (and black in Dark for that matter). Ultimately, these buttons look weird because they should really just be FABs and use the standard FAB styling (but that's a bigger change). I think for the moment I'm happy with where it is as I think it's at least a lot clearer which buttons are disabled and which aren't. I think this can go to QA to see what they think!

seadowg (approved):

SaumiaSinghal:
Yes, these buttons confuses me too! :-/

lognaturel:
Sorry to come in over the top here but can we please keep the primary color / black icons for now? I agree we'll want to modernize these buttons eventually so I don't think there's much benefit to changing just the color now. We have the pattern of black buttons a few other places (exit form confirmation dialog, instance map icon, etc).

SaumiaSinghal:
Hi @lognaturel! Sorry! my last comment makes the work on this PR a bit confusing. Following on what the material components does for FAB, I just wanted to display how map buttons would appear when the color is set to secondary color.

I have used the primary black color in this PR and just dropped its alpha value to 0.12 for the case when the buttons are disabled. 

| Magenta Theme         |  Light Theme | DarkTheme |
:-------------------------:|:-------------------------:|:-------------------------:
Screenshot_1603111647  |  Screenshot_1603111679 | Screenshot_1603111707

seadowg:
@lognaturel don't worry it is using `colorOnSurface` (black/white) for the moment. @SaumiaSinghal's screenshots are just an example of what `colorSecondary` would look like.

lognaturel:
Ohhhh! Sorry about that, @SaumiaSinghal. I should have read the full thread more carefully. 😊 Thanks so much for the clarification.

SaumiaSinghal:
No worries @lognaturel!

mmarciniak90:
**Tested with success**

Verified cases:
- all Geo widgets in `All widgets` form
- Magenta theme
- light/dark theme
- icons color
- icons activity
- different map types
- rotation
- minimization
- lock the screen

kkrawczyk123:
@SaumiaSinghal There is one weird visual result that I can see on Android 5.1 for all themes, all icons
 are red right now, see the screenshot:
Screenshot (02_32PM, Nov 04, 2020)
Could you please take a look?

seadowg:
@SaumiaSinghal @kkrawczyk123  I had a vague idea of what the problem was here, so I had a look into it. The problem is that using theme attributes like (`?colorOnSurface`) in `selector` (like in you're doing in `map_button_tint_list.xml`) doesn't work until API 23 (our minimum is 21, Android 5.1 is 22). 

Solving this is pretty awkward as we'd probably want to define different tint list xml files for each theme, add an attribute (like `map_image_button_tint_list` for every theme, and then use that in the map button style (attributes work fine in styles below API 23). My instinct is we should just go with the weird red icons for the moment as it doesn't change any behaviour and is just a problem for older versions of Android that we won't support for that much longer. I want to pull in either @grzesiek2010 or @lognaturel for a second opinion here.

lognaturel:
That sounds fine to me, @seadowg. Once we are ready to make the "magenta theme" the default one, we can decide what to do. We may still have to address it but we don't have to make the decision now.

kkrawczyk123:
Ok so, I don't see any other reasons to keep this open then.
Tested on Androids: 5.1, 7.1 and 9.0.

lognaturel:
> Once we are ready to make the "magenta theme" the default one, we can decide what to do.

I missed the "all themes" part of this until I just experienced it on Android 5.1 and the default theme. It's not just the geo buttons, the divider lines in the list activities are also red. It's a small percentage of our user base but it is ~25k individuals so we may get questions or push back but let's wait and see.

seadowg:
> It's a small percentage of our user base but it is ~25k individuals so we may get questions or push back but let's wait and see.

Fair. I don't think it really disrupts them too much though? My thinking with older versions is that it's OK for cosmetics to degrade.

Issue #4154, opened 2020-10-06
Enumerator is asked to stop recording before changing screen


seadowg:
**Acceptance**

- [x] Given I'm recording using the in-app recorder
When I swipe to another screen
Then I see a warning dialog about recording
Then I remain on the screen I was on

- [x] Given I'm recording using the in-app recorder
When I click the jump menu
Then I see a warning dialog about recording
Then I remain on the screen I was on

- [x] Given I'm recording using the in-app recorder
When I quit the form
Then I see a warning dialog about recording
Then I remain on the screen I was on

- [x] Given I'm recording using the in-app recorder
When I open general settings
Then I see a warning dialog about recording
Then I remain on the screen I was on

PR #4168, opened 2020-10-13
Rework `AudioWidget` view/flow and extract module for playing audio

seadowg:
~~Blocked by #4171~~

Screenshot_20201013-143302

This changes the Audio widget to match the flow mocked up at https://github.com/seadowg/collect-design/tree/master/prototypes/in-app-audio-recording. There are two key behaviour changes:

1. The "Choose Sound" and "Record Sound" button will be hidden once the question has been answered
1. The audio controller [Card](https://material.io/components/cards/) now has a "Remove response" action that lets the enumerator clear the answer and start again

This PR also extracts some key audio code to its own Gradle module (and converts it to Kotlin 🎉) and introduces a new approach for the widget to interact with the "audio player" code that's closer to the way other integrations with widgets have been looking recently.

A bunch of new tests and played around with the audio widget manually.

Comments inline.

Intended behaviour changes are up above but it'd be good to put the audio widget through its paces.

Any form with `audio` questions.

We should potentially update screenshots but I'm not sure we need docs updates yet.


collect_app/build.gradle
```diff
@@ -310,6 +311,7 @@ dependencies {
     testImplementation "junit:junit:4.13"
     testImplementation "org.mockito:mockito-core:3.5.11"
 
+    testImplementation project(path: ':test-shared')
```
seadowg: We ended up having test helpers that weren't specific to our "app code" so moved some of that here, so we're not repeating anything.

collect_app/src/main/java/org/odk/collect/android/audio/AudioControllerView.java
```diff
@@ -24,64 +26,65 @@
...
-import butterknife.OnClick;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.odk.collect.android.databinding.AudioControllerLayoutBinding;
 
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 
 public class AudioControllerView extends FrameLayout {
 
-    @BindView(R.id.currentDuration)
```
seadowg, line 38: Just did a straight replacement of Butterknife with view binding here.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/AudioPlayer.java
```diff
@@ -0,0 +1,20 @@
...
+
+public interface AudioPlayer {
+
+    void play(Clip clip);
+
+    void pause();
+
+    void setPosition(String clipId, Integer position);
+
+    void onPlayingChanged(String clipID, Consumer<Boolean> playingConsumer);
```
seadowg, line 15: I wanted to avoid using `LiveData` inside `View` components as it means we need to pass them some kind of lifecycle to use for observations. This approach lets us use a more standard listener hook in the `View` but lets us use `LiveData` as the implementation so we have lifecycle awareness.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppAudioDataRequester.java
```diff
@@ -0,0 +1,68 @@
...
+import android.widget.Toast;
+
+import org.javarosa.form.api.FormEntryPrompt;
+import org.odk.collect.android.R;
+import org.odk.collect.android.listeners.PermissionListener;
+import org.odk.collect.android.utilities.ActivityAvailability;
+import org.odk.collect.android.utilities.ApplicationConstants;
+import org.odk.collect.android.utilities.PermissionUtils;
+
+public class ExternalAppAudioDataRequester implements AudioDataRequester {
```
seadowg, line 15: We should be able to swap this implementation with another one for internal recording

seadowg:
Interested to see opinions from @grzesiek2010 @lognaturel, @mmarciniak90, @kkrawczyk123 and @SaumiaSinghal on this one

collect_app/src/main/java/org/odk/collect/android/audio/AudioHelper.java
```diff
@@ -31,21 +33,26 @@
  * construct multiple instances (within a {@link android.view.View} or
  * {@link androidx.fragment.app.Fragment} for instance) if needed within one
  * {@link android.app.Activity}.
+ *
+ * @deprecated wrapping the ViewModel like this doesn't really fit with other ways we've integrated
+ * widgets with "external" services. Instead of this widgets should talk to {@link AudioPlayer}
+ * and the Activity/Fragment components should talk to the ViewModel itself.
  */
 
+@Deprecated
```
seadowg, line 42: I definitely could have removed this everywhere but it felt like it would be pretty time intensive and actually pretty isolated from what I was working on. We can come back for this next time we touch anything in the neighbourhood.

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
```diff
@@ -157,83 +112,116 @@ public IAnswerData getAnswer() {
...
+        if (getFormEntryPrompt().getAppearanceHint() != null && getFormEntryPrompt().getAppearanceHint().toLowerCase(Locale.ENGLISH).contains(WidgetAppearanceUtils.NEW)) {
             binding.chooseButton.setVisibility(View.GONE);
         }
     }
 
     private void updatePlayerMedia() {
         if (binaryName != null) {
-            audioHelper.setAudio(audioController, new Clip(String.valueOf(ViewCompat.generateViewId()), getAudioFile().getAbsolutePath()));
-            audioController.showPlayer();
+            Clip clip = new Clip("audio:" + getFormEntryPrompt().getIndex().toString(), getAudioFile().getAbsolutePath());
```
seadowg, line 185: This now uses a predictable ID for the clip. I think before I wasn't as clued up on the structure of the form. Not quite sure why we went with random IDs

test-shared/src/main/java/org/odk/collect/testshared/FakeLifecycleOwner.kt
```diff
@@ -0,0 +1,20 @@
+package org.odk.collect.testshared
+
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
+
+class FakeLifecycleOwner : LifecycleOwner {
+
+    private val lifecycle = LifecycleRegistry(this).apply {
```
seadowg: I have really missed `apply`

test-shared/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+
+android {
+    compileSdkVersion 30
+    buildToolsVersion "30.0.0"
+
+    defaultConfig {
+        minSdkVersion 21
+        targetSdkVersion 30
```
lognaturel: Will need to pull in references for these numbers once the string module PR is merged.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/AudioDataRequester.java
```diff
@@ -4,4 +4,6 @@
 
 public interface AudioDataRequester {
     void requestRecording(FormEntryPrompt prompt);
+
+    void requestFile(FormEntryPrompt prompt);
```
lognaturel, line 8: Feels like this should be separate? But I suppose that can be discovered later once there are multiple implementations for recording.

audioclips/build.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+
+android {
+    compileSdkVersion 29
+    buildToolsVersion "30.0.0"
+
+    defaultConfig {
+        minSdkVersion 21
```
lognaturel: Note to come back and set these to shared vars once the strings module pr is merged.

audioclips/src/main/java/org/odk/collect/audioclips/AudioClipViewModel.kt
```diff
@@ -0,0 +1,203 @@
+package org.odk.collect.audioclips
+
+import android.media.MediaPlayer
+import androidx.lifecycle.*
```
lognaturel: We don't allow wildcard imports typically. Is this going through because it's Kotlin or because the style checks aren't applied? I do think explicit imports can avoid errors and are worth it.

test-shared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt
```diff
@@ -0,0 +1,65 @@
+package org.odk.collect.testshared
```
lognaturel: It feels weird for the module and package names to be mismatched. Is that intentional? So far all modules have matching package names. Except for `collect_app`, I suppose.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
```diff
@@ -390,25 +394,50 @@ public void resumeFormEntry_clearsSaveResult() {
     }
 
     @Test
-    public void markOriginalFileOrDelete_whenQuestionIndexHasAnswer_onRecreatingViewModel_deletesFile() {
-        viewModel.markOriginalFileOrDelete("index", "blah");
+    public void deleteAnswerFile_whenFileHasAlreadyBeenDeleted_actuallyDeletesNewFile() {
```
lognaturel: These tests about deleting/replacing media files are good to have. This behavior is hard to reason about and I don't see any context for what it's for anywhere. Maybe you could put these tests in a region and just say something like "we want to be able to go back to the exact state we were in at the last save if a user exits an instance without saving. This means keeping track of any media files that were previously attached so they can be restored." Alternately something like that could go above the two maps in `FormSaveViewModel` used for that purpose.

Also this save and discard changes behavior would be good for QA to check once the audio rework is complete. Want to keep a list like you did for the match exactly feature?

lognaturel (commented):
I realize this is a stepping stone so some of my comments may not be so relevant depending on what's in your head for the future! I figure I'll still write out what questions come to mind in case it gives you new ideas.

I'm a little surprised by the `audioclips` module name. Are you thinking that's the "final" name and scope of the module? It's not immediately obvious to me why `AudioPlayer` and `ViewModelAudioPlayer` aren't in there too. I thought it might have to do with Android dependencies but `AudioClipViewModel` has them.

Am I understanding correctly that `AudioHelper` and `AudioPlayer` currently play a similar role wrapping `AudioClipViewModel` so views don't need to know about lifecycle concerns but the former is for form-defined audio (question prompt audio, choices audio) and the latter is for user-defined audio? Is the fact that `AudioHelper` is not involved in user-defined audio what you mean by it being isolated from what you're working on?

I think we do want to have a confirmation dialog for the response removal. I imagine that ux will get refined anyway so could be added later but let's not lose track of it.

I happened to notice that audio playback resets on screen rotation. I didn't check whether that's the case on `master` or not.

How should we do QA for this feature? I think we want to avoid @getodk/testers doing the same checks over and over again. Test coverage is high and I'd advocate for merging this PR as-is and doing occasional sanity checks on some of the follow up PRs. That process seemed to work well for match exactly. Let me know what y'all think.

test-shared/build.gradle
seadowg: Great shout. Will rebase this on top of that (making it dependent)

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/AudioDataRequester.java
seadowg: Yeah this ends up changing in the next PR so your feeling is correct haha

audioclips/src/main/java/org/odk/collect/audioclips/AudioClipViewModel.kt
seadowg: Right we're not getting style checks here but I'll rebase on top of `master` which will give us that setup.

test-shared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt
seadowg: Er annoyingly no dashes in package names. Maybe we just go with one word for module names then?

seadowg:
> I'm a little surprised by the audioclips module name. Are you thinking that's the "final" name and scope of the module? It's not immediately obvious to me why AudioPlayer and ViewModelAudioPlayer aren't in there too. I thought it might have to do with Android dependencies but AudioClipViewModel has them.

Yeah I think it might be the scope of the module. So my thinking is that `AudioPlayer` is owned by the widgets (why I have it in `widgets.utilities`). It's the interface they speak to play audio. `ViewModelAudioPlayer` is the implementation the "form entry" world passes to the widgets (I think I should move this to the `formentry` package to make that clearer). I was trying to go for a similar relationship to what we have with `QuestionMediaManager` and `FormSaveViewModel`. The difference is just that we need a wrapper implementation rather than just having the ViewModel implement the interface. 

> Am I understanding correctly that AudioHelper and AudioPlayer currently play a similar role wrapping AudioClipViewModel so views don't need to know about lifecycle concerns but the former is for form-defined audio (question prompt audio, choices audio) and the latter is for user-defined audio? Is the fact that AudioHelper is not involved in user-defined audio what you mean by it being isolated from what you're working on?

It probably makes sense to sweep back and clean this up once we've got a beta of the audio out. When I started the refactor I thought I'd have to change everything (completely replacing the `AudioHelper` with the `AudioPlayer`) but as I got going the two different kinds of audio (form-defined and user-defined as you say) ended up being pretty separate from each other so I opted for deprecating so I could carry on with the recoridng work.

> I think we do want to have a confirmation dialog for the response removal. I imagine that ux will get refined anyway so could be added later but let's not lose track of it.

Totally agree. I think we talked about this before but we ended up with an issue: https://github.com/getodk/collect/issues/4154.

> I happened to notice that audio playback resets on screen rotation. I didn't check whether that's the case on master or not.

Huh. I'll check on that.

audioclips/src/main/java/org/odk/collect/audioclips/AudioClipViewModel.kt
seadowg: Nope you're right it also won't work because Kotlin. It looks like PMD support is coming but not quire there yet. I'll have a look at Kotlin specific projects for the moment to see if we can get a few of our rules applied to Kotlin so we're not too far behind when PMD supports both.

collect_app/src/test/java/org/odk/collect/android/formentry/audit/FormSaveViewModelTest.java
seadowg: I'll add regions/comments to try and make this clearer. Yeah I think it's a good idea to have a list of things to check!

seadowg:
@lognaturel looks like audio playback also resets on master

test-shared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt
lognaturel: I see. How about a `reused` module with `test` and `utility` packages or something like that? It would all be pure Java so should be no big deal to compile and I can't imagine there would be a ton shared anyway.

lognaturel:
> I think I should move this to the formentry package to make that clearer

Yes. Do you want to do this now or in a follow up?

> I think we talked about this before but we ended up with an issue: #4154.

Indeed 🤦🏻‍♀️

> looks like audio playback also resets on master

Good to know. Not a big deal either way.

So revise the testshared module/package names, maybe move `ViewModelAudioPlayer` to `formentry` and then we can merge without QA for now?

seadowg:
@lognaturel sorry I moved `ViewModelAudioPlayer` to the `widgets.utilities` package as that's where all the other implementations for the widgets' integration interface live right now. Forgot to leave a comment about that.

I'll revise the module name now as I think it should just change to one word for simplicity.

lognaturel:
Sounds good, @seadowg! See what you think about https://github.com/getodk/collect/pull/4168#discussion_r507887013.

test-shared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt
seadowg: Feels like a good idea to keep test and non-test code separate, so we're not bringing both in when we add a dependency to a module. Maybe for the moment we keep it as `testshared` and work out what we want to do when we have "shared" non-test code that doesn't feel like it needs its own module (like `async`). What do you think?

lognaturel: Fair enough.

seadowg:
@lognaturel ok modules names are fixed!

lognaturel (approved):

lognaturel:
FYI, @getodk/testers. We're going to do a similar process with the audio features in v1.29 as we did for the "match exactly" feature in v1.28 where some of the intermediate PRs go in without QA. This means you won't have to review the same flows over and over again but we can still review isolated chunks of work. We're going to make sure we have good automated coverage and bring you in every couple of PRs to make sure that no major behavior has been forgotten about.

PR #4169, opened 2020-10-15
Fix sort by displaying unexpectedly issue after using map button or search icon 

### LINKED ISSUE #4132: Sort by displays unexpectedly after using map button and search icon
Collect v1.28.x, Android v all .x.x, device used...

#### Problem description
Sort by menu displays unexpectedly after using map view button on Fill Blank form view. 

scenario 1:
1. Open fill blank form 
2. Click on Sort by icon
3. Click on view map button
4. Leave the map using back button

scenario 2:
1. Open fill blank form 
2. Click on search icon 
3. Click on view map button
4. Leave the map using back button

gif of both scenarios:
ezgif com-video-to-gif (37)

#### Expected behavior
Fill blank form view should display, sort by menu should not be visible, it should be hidden in both cases. 

#### Other information 

SaumiaSinghal:
Closes #4132 

I tested it.

I removed all the checks responsible for saving the state of `bottomSheetDialog` in `AppListActivity`, and so when the use moves to another activity, the already shown dialog is simply dismissed.

Regression issue is the bottom sheet dialog will disappear on rotating the device, but as @grzesiek2010 and everyone discussed that this is not that important dialog, and probably the user won't mind too much opening it again on rotating the device.

No specific form required

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/AppListActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/AppListActivity.java
@@ -61,7 +61,6 @@ abstract class AppListActivity extends CollectAbstractActivity {
     protected static final int LOADER_ID = 0x01;
     private static final String SELECTED_INSTANCES = "selectedInstances";
     private static final String IS_SEARCH_BOX_SHOWN = "isSearchBoxShown";
-    private static final String IS_BOTTOM_DIALOG_SHOWN = "isBottomDialogShown";
     private static final String SEARCH_TEXT = "searchText";
 
     protected CursorAdapter listAdapter;
@@ -72,7 +71,6 @@ abstract class AppListActivity extends CollectAbstractActivity {
     protected LinearLayout llParent;
     protected ProgressBar progressBar;
     private BottomSheetDialog bottomSheetDialog;
-    private boolean isBottomDialogShown;
 
     private String filterText;
     private String savedFilterText;
@@ -143,24 +141,18 @@ abstract class AppListActivity extends CollectAbstractActivity {
     protected void onResume() {
         super.onResume();
         restoreSelectedSortingOrder();
-        setupBottomSheet();
     }
 
     @Override
     protected void onSaveInstanceState(Bundle outState) {
         super.onSaveInstanceState(outState);
         outState.putSerializable(SELECTED_INSTANCES, selectedInstances);
-        outState.putBoolean(IS_BOTTOM_DIALOG_SHOWN, bottomSheetDialog.isShowing());
 
         if (searchView != null) {
             outState.putBoolean(IS_SEARCH_BOX_SHOWN, !searchView.isIconified());
             outState.putString(SEARCH_TEXT, String.valueOf(searchView.getQuery()));
         } else {
             Timber.e("Unexpected null search view (issue #1412)");
-        }
-
-        if (bottomSheetDialog.isShowing()) {
-            bottomSheetDialog.dismiss();
         }
     }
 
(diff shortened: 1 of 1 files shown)


grzesiek2010:
I think this is a perfect example of what we were talking about with @seadowg some time ago. General conclusion was that we pay to much attention to what happens after an activity is recreated.
We agreed that we should do that if we display something really important (an important dialog or something) but in other cases we could just let it go.

In this case we have a special code that makes sure we display the menu after for example rotating a device. For what? We only ended up with more code that in this case even introduced a bug.

would you agree with me that we should change our approach a bit and in cases like this just stop trying to restore everything? It applies to all of us @lognaturel @seadowg @mmarciniak90 @kkrawczyk123 @SaumiaSinghal

seadowg:
> In this case we have a special code that makes sure we display the menu after for example rotating a device. For what? We only ended up with more code that in this case even introduced a bug.

Yeah I think I agree with you here @grzesiek2010. It's easy for us to get over obsessed by rotation as it's such a headache to deal with but I think it's actually unlikely that people are rotating their device while carrying out other actions. There are times where we need to care it as we might end up losing some data the user entered or let them skip a step (like with the identity/change reason stuff) but it doesn't feel like losing the sorting bottom sheet is one of those cases.

seadowg:
@grzesiek2010 so we thinking it makes sense for @SaumiaSinghal to just remove the code that deals with rotation here?

grzesiek2010:
Yes, I would be in favor of such a solution.

seadowg:
Cool. I think it'd be good to get everyone else's inputs as you first suggested but for the moment I'll make this a draft. @SaumiaSinghal please reach out if it's not clear what changes we're suggesting!

SaumiaSinghal:
Yep Following on @grzesiek2010 suggestion, I tried implementing the `BottomSheetDialog` in dialog fragment, and used a listener for communicating with the activity. Should I move on with this implementation.

grzesiek2010:
> Yep Following on @grzesiek2010 suggestion, I tried implementing the BottomSheetDialog in dialog fragment, and used a listener for communicating with the activity. Should I move on with this implementation.

No, it doesn't make sense to complicate the code here. Just remove those checks responsible for restoring the menu (as we agreed, unless @lognaturel is against) and that's all.

SaumiaSinghal:
Oh alright I understand @grzesiek2010, we should be concerned about the orientation changes just for important dialogs. I will remove the checks.

SaumiaSinghal:
@grzesiek2010 you can take a look whenever you have time

grzesiek2010:
Yes I think this is what I expected.

grzesiek2010 (approved):

mmarciniak90:
> we should be concerned about the orientation changes just for important dialogs. I will remove the checks.

@SaumiaSinghal I noticed that almost on each view `Sort by` list disappears after rotation. `Delete Saved Form` is only one view where it's still visible. Will you remove it also from this place?

SaumiaSinghal:
Sorry @mmarciniak90. When I said that, I was not considering Sort by bottom sheet dialog as an important dialog. So, I expect that it won't be retained on most of the activities. In the case of `Delete Form`, it is retained as we show that on a fragment.  I think @grzesiek2010 can confirm it for one more time here, if I am understanding things correctly.

grzesiek2010:
Yes the difference is that in case of that view (responsible for deleting forms) we use fragments, in all the other cases we use activities. In the future we are going to use fragments everywhere so it will be consistent. For now we don't need to care if it's the default behavior that's ok.

mmarciniak90:
@SaumiaSinghal and @grzesiek2010 thanks for the clarification.

**Tested with success**

Verified on Android versions; 7.0, 8.1, 10.0

Verified cases:
- sort by functionality
- search functionality
- scenario 1 from issue
- scenario 2 from issue
- disappearing `Sort by` list after rotation (except Delete Saved Form and Send Finalized views)
- RTL language
- light/dark theme

kkrawczyk123:

PR #4171, opened 2020-10-15
Move strings to their own module

seadowg:
This moves all our `strings.xml` files to their own Android Library module. This allows us to extract or write new code in other modules that use translated strings without having to have multiple resource files that translators need to deal with.

It felt like a good idea to give this change its own PR rather than lumping it into something else.

Ran checks and tests to check everything is working.

An alternative would be to let modules look after their own strings which would give us faster build times (on a per module basis) but this would result in a more awkward process for our translators so this seems like a good compromise. 

Shouldn't be any change in behaviour. As far as I can tell this will either compile or it won't so I think we can skip QA and merge when a reviewer is happy.


strings/build.gradle
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id 'com.android.library'
+}
+
+android {
+    compileSdkVersion 30
```
lognaturel: Is there really no way to set all these values once across modules? It's going to be really easy to make a mistake and not update all.

lognaturel (approved):
I think @seadowg captured all the tradeoffs in the PR description. The compile/assembly/whatever magic time of other modules probably isn't a big deal since this one would only have to be compiled when there are string changes.

Since this is a significant organizational change, I think @grzesiek2010 should have another opportunity to consider it before we merge.

lognaturel (changes requested):
I think it's worth looking into something like https://stackoverflow.com/questions/20404476/how-to-define-common-android-properties-for-all-modules-using-gradle as this change is introduced.

strings/build.gradle
seadowg: Yeah we can probably use the same mechanism we use for the Dagger version. I think you're right that this would be the time to do it.

lognaturel (approved):
Great! Just some conflicts. 😞

seadowg:
@grzesiek2010 you feeling ok about this?

grzesiek2010:
What about untranslated strings shouldn't we move that file as well?

seadowg:
> What about untranslated strings shouldn't we move that file as well?

@grzesiek2010 good question. I'm thinking we don't need to as I'd imagine we won't have to share values in `unstranslated.xml` and translators won't have to touch that file.

grzesiek2010:
Ok.

grzesiek2010 (approved):

PR #4173, opened 2020-10-16
Add in-app recording

### LINKED ISSUE #4149: An enumerator can record audio without using an external app
Prototype for in-app audio recording is [here](https://github.com/seadowg/collect-design/tree/master/prototypes/in-app-audio-recording)

For the moment we'll record audio to `.m4a` files using AAC encoding at 32kHz and 128kbps.

**Acceptance**

These assume you've enabled "In-app audio recording" in the "Experimental" preferences

- [x] Given I'm on an `audio` question
When I click "Record sound"
And I wait for a few seconds
And I then I click the ☑️ button
And I click the ▶️ button
Then the question shows the audio player
And I can listen to the audio I just recorded

- [x] Given I'm on an `audio` question
When I click "Record sound"
And I tap the back button
Then I return to the question
And it does not have an answer (no audio player)

- [x] Given I'm on an `audio` question
And have not granted Collect permission to record audio
When I click "Record sound"
Then I see a prompt to grant audio recording permission
When I click "Allow"
Then I move on to recording

- [x] Given I'm on an `audio` question
And have not granted Collect permission to record audio
When I click "Record sound"
Then I see a prompt to grant audio recording permission
When I click "Deny"
Then I return to the question
And it does not have an answer (no audio player)

seadowg:
Closes #4149 
~~Blocked by #4171~~
~~Blocked by #4168 (branched off that PR)~~

This adds recording to the app when enabled in "Experimental" settings.

Driven out by tests and tried out on emulator and device.

Comments inline.

Should just add new feature. Good to check that using an external app still works as expected.

Any form with audio.

We should write docs before release but right now I think it'd be good to let the feature evolve without writing anything up.


audioclips/build.gradle
```diff
@@ -27,7 +27,6 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```
seadowg, line 30: We don't need these for library modules

audiorecorder/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="org.odk.collect.audiorecorder">
+
+    <application>
+        <activity
+            android:name=".recording.AudioRecorderActivity"
+            android:label="Audio Recorder"
+            android:screenOrientation="portrait"
```
seadowg, line 9: Let's deal with rotation later as, once we have a Foreground Service responsible for recording, it won't be as big a deal.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/Config.kt
```diff
@@ -0,0 +1,67 @@
...
+import dagger.BindsInstance
+import dagger.Component
+import dagger.Module
+import dagger.Provides
+import org.odk.collect.audiorecorder.recorder.MediaRecorderRecorder
+import org.odk.collect.audiorecorder.recorder.RealMediaRecorderWrapper
+import org.odk.collect.audiorecorder.recorder.Recorder
+import org.odk.collect.audiorecorder.recording.AudioRecorderActivity
+
+private var _component: AudioRecorderDependencyComponent? = null
```
seadowg, line 15: This file provides a pretty minimal (for Dagger) DI setup. We could come up with something more complex to attach the component to the `Application` object which would mean we wouldn't run into static state problems (see the code comment below) but I don't think we need to go there yet.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recorder/MediaRecorderWrapper.kt
```diff
@@ -0,0 +1,63 @@
+package org.odk.collect.audiorecorder.recorder
+
+import android.media.MediaRecorder
+
+/**
+ * Allows faking/stubbing/mocking with our interactions with Android's MediaRecorder
+ */
+
+interface MediaRecorderWrapper {
```
seadowg, line 9: This interface brought to you by a slightly funky looking `ShadowMediaRecorder` and the "don't mock/fake/stub things you don't own" rule.

audiorecorder/src/test/java/org/odk/collect/audiorecorder/recording/MediaRecorderRecorderTest.kt
```diff
@@ -0,0 +1,224 @@
...
+        assertThat(mediaRecorder.getOutputFile()!!.exists(), equalTo(false))
+    }
+
+    @Test
+    fun cancel_beforeStart_works() {
+        recorder.cancel()
+    }
+}
+
+private class FakeMediaRecorderWrapper : MediaRecorderWrapper {
```
seadowg, line 92: I went with a fake rather than a mock here as it lets me capture the importance of the method call order (`prepare` before `start` for example) without having to specify exactly what order things are called in.

collect_app/src/main/java/org/odk/collect/android/utilities/PermissionUtils.java
```diff
@@ -41,6 +33,12 @@
 
 public class PermissionUtils {
```
seadowg, line 34: I didn't end up having to pull this out to its own module but I like the changes I made to it in preparation so kept them in.

audiorecorder/build.gradle
```diff
@@ -0,0 +1,54 @@
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-kapt'
+
+android {
+    compileSdkVersion 30
```
lognaturel: Use var

audiorecorder/src/main/java/org/odk/collect/audiorecorder/Config.kt
```diff
@@ -14,6 +14,10 @@ import org.odk.collect.audiorecorder.recording.AudioRecorderActivity
 
 private var _component: AudioRecorderDependencyComponent? = null
 
+/**
```
lognaturel, line 17: I don’t get this. Might be good to talk through it and see if we can clarify.

strings/src/main/res/values/strings.xml
```diff
@@ -705,4 +705,7 @@
     <string name="security_error">Collect can\'t connect securely to the server at %s. Did you enter the URL correctly?</string>
     <string name="failure">Failure</string>
     <string name="form_update_succeeded">Form update succeeded</string>
+    <string name="in_app_recording">Use in-app audio recording</string>
```
lognaturel: What do you think about flipping this and going with "Use external app for audio recording" (true by default)? That way it most likely won't need to be translated again. We can absolutely change the wording if we find we need to but at least we send something to translators that's likely to stick around.

lognaturel (commented):
Just a couple things that seem worth talking through.

The build took 8mins trying this out and I don't think it's ever been that long. Is that what others are seeing too on a fresh build? It's not a huge deal but felt surprising.

audiorecorder/build.gradle
seadowg: Oops!

strings/src/main/res/values/strings.xml
seadowg: Good idea!

seadowg:
> The build took 8mins trying this out and I don't think it's ever been that long. Is that what others are seeing too on a fresh build? It's not a huge deal but felt surprising.

On CI or locally? `./gradlew clean assembleDebug` is ~40 secs locally for me.

lognaturel (approved):
👍 @getodk/testers this is entirely behind an experimental flag for now. We'll bring you in once a few more stories are complete.

lognaturel:
> On CI or locally? ./gradlew clean assembleDebug is ~40 secs locally for me.

@seadowg and I talked briefly about this. I was doing run from Android Studio. I'm going to update Android Studio, explore the build analyzer, and we can talk in Slack if it persists.

PR #4178, opened 2020-10-21
Add analytics for audio prompt

seadowg:
This adds analytics to the Audio widget to track:

* How many forms include it
* How many enumerators/forms use Record vs Choose to answer the question

Ran tests and manually tested audio widget.

Not a lot of choices made here! Just made changes to get the tracking in.

Should be no change.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/Analytics.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/Analytics.java
@@ -1,7 +1,7 @@
 package org.odk.collect.android.analytics;
 
 public interface Analytics {
-
+    
     void logEvent(String category, String action);
 
     void logEvent(String category, String action, String label);
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/Analytics.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/Analytics.java
@@ -6,6 +6,8 @@ public interface Analytics {
 
     void logEvent(String category, String action, String label);
 
+    void logFormEvent(String event, String formId);
+
     void setAnalyticsCollectionEnabled(boolean isAnalyticsEnabled);
 
     void setUserProperty(String name, String value);
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -143,4 +143,6 @@ public class AnalyticsEvents {
      * should be a hash of the endpoint setting.
      */
     public static final String CUSTOM_ENDPOINT_SUB = "CustomEndpointSub";
+
+    public static final String AUDIO = "audio";
 }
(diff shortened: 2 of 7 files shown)


lognaturel (changes requested):
Could we do separate events for record and choose with the form ID as the action? If we do it as in the PR, we won't be able to tell workflows associated with unique forms. I think it would be nice to say the Choose vs. Record decision is consistent for a given form definition vs. data collectors are making on the fly decisions about which to use.

If you look at the [AddRepeat](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/events/~2Freport~2FAddRepeat%3Ft%3D1603327436777&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_r..eventId%253DAddRepeat&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1) event as an example, `action[AddRepeat]` gives a sense of how folks are adding repeats. And `label[AddRepeat]` gives a sense of unique forms using repeats. But we don't get a sense of whether a single project is relying on `Inline` or lots of folks are trying it out or what.

To be explicit, I'm proposing `AudioChoose` and `AudioRecord` events. I don't see how the count of the widget being shown is all that helpful so I'd remove it! But if you want to be fancy, I think you could log an event when audio is in a field list. I feel like that might be actionable or at least help us validate one of our hypotheses in the longer term that people may want to record an answer and also write notes/answer questions.

seadowg:
> Could we do separate events for record and choose with the form ID as the action? If we do it as in the PR, we won't be able to tell workflows associated with unique forms. I think it would be nice to say the Choose vs. Record decision is consistent for a given form definition vs. data collectors are making on the fly decisions about which to use.

Right that makes sense. I was going with the style we've been using but as you point out that's not working really. I'm not into sending the form id  as an "action" though. Should we potentially use a different parameter than "action"? 

What we're usually doing is we're sending an event name (`AudioRecord` or `SettingsImportQR`) and then some kind of "scope" or "differentiator" (a hashed analytics ID or a server URL). In other cases we send have multiple "types" of events (`Submission` or `MatchExactlySyncCompleted`). I'm wondering if we could just introduce the following dimensions to make our events easier to understand:

* `form`
* `server`
* `type`

For these new events they would send the events (`AudioRecord` and `AudioChoose`) and then send the `form` dimension. Something like `MatchExactlySyncCompleted` where we really just care about what type of thing happened could just send the `type` dimension. I guess with `AddRepeat` we'd want different events rather than using different `actions` and send the `form` dimension. How does that sound? I just mean introducing this for the new events in this PR btw!

Just to clarify the dimensions I'm talking about here are in documentation here: https://support.google.com/firebase/answer/10075209.

lognaturel:
Great, the changes that Firebase Analytics just went through finally make a little bit of sense. AudioRecord and AudioChoose with the form dimension sounds great.

seadowg:
@lognaturel I've removed the prompt event and added the new ones. I think we'll need to create custom dimensions on the FIrebase side for it to work.

lognaturel (approved):

PR #4180, opened 2020-10-22
Use Foreground Service to manage audio recording

### LINKED ISSUE #4150: An enumerator can leave the app while recording audio
**Acceptance**

When I lock my device
Then I see a notification that audio is recording

When I lock my device
And I unlock it
Then the recording is still in progress

- [x] Given I'm recording
And "Don't keep activities" is enabled
When I switch to another app
And then return to Collect
Then the recording is still in progress (doesn't restart)

- [x] Given I'm recording
When I force close the app (using the task manager)
Then the notification disappears

seadowg:
Closes #4150 

This moves the recording code into a Foreground Service which allows it to run while the phone is locked or the app is in the background.

New tests and verified manually on emulator and on device.

Comments inline.

Should just add new feature. There has been a change to the translation code. I had a quick play around with changing languages in the app and everything seemed ok however.


audiorecorder/src/main/AndroidManifest.xml
```diff
@@ -2,12 +2,17 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.odk.collect.audiorecorder">
 
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
```
seadowg, line 5: This is only required on Android 28+ as far as I know.

audiorecorder/src/test/resources/robolectric.properties
```diff
@@ -0,0 +1 @@
+application=org.odk.collect.audiorecorder.TestApplication
```
seadowg, line 1: This was the magic I needed to get Dagger setup properly for tests. I'd completely forgotten robolectric could be configured in a properties file.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderViewModel.kt
```diff
@@ -1,29 +1,46 @@
...
 import javax.inject.Inject
 
-internal class AudioRecorderViewModel(private val recorder: Recorder) : ViewModel() {
+internal class AudioRecorderViewModel(private val application: Application, private val recordingRepository: RecordingRepository) : ViewModel() {
+
+    val recording: LiveData<File?> = recordingRepository.getRecording()
 
     fun start() {
-        recorder.start()
+        application.startService(
```
seadowg, line 19: I'm not 100% sure about the ViewModel being responsible for the service like this but it feels good to hide that there is a Service involved from the UI.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/RecordingRepository.kt
```diff
@@ -0,0 +1,22 @@
+package org.odk.collect.audiorecorder.recording
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import java.io.File
+
+class RecordingRepository {
```
seadowg, line 7: "Repository" might be the right phrase but this pattern of a singleton that communicates between Services, ViewModels and UI seems pretty standard/nice.

audiorecorder/src/test/java/org/odk/collect/audiorecorder/recording/AudioRecorderActivityTest.kt
```diff
@@ -15,81 +14,104 @@ import org.junit.Before
...
+                }
             }
         )
     }
 
     @Test
-    fun launching_startsRecording() {
-        launchActivity<AudioRecorderActivity>()
-        assertThat(fakeRecorder.isRecording(), equalTo(true))
+    fun launching_startsRecordingService_withStart() {
```
seadowg, line 45: I thought about introducing Espresso tests which mean we could just test the whole flow without having to separate the bulk of testing out between the Activity and the Service. That would become redundant if then got rid of the Activity and exposed the ViewModel as the main interface of the module though so I think for the moment I'll stick with this structure and see where we end up

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -33,15 +34,18 @@
...
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.util.Locale;
 
 import javax.inject.Inject;
 
 import static org.odk.collect.android.preferences.MetaKeys.KEY_GOOGLE_BUG_154855417_FIXED;
 
-public class Collect extends Application {
+public class Collect extends Application implements LocalizedApplication {
```
seadowg, line 48: I looked at moving the `TranslationHandler` to the `strings` module but it would have meant moving all our code that deals with the custom locale. I ended up favouring this solution where we keep the locale stuff in `collect_app` and just have the `strings` code expect to be given a locale.

audiorecorder/src/main/AndroidManifest.xml
lognaturel: Looks like it's based on the targetSdk version.

strings/build.gradle
```diff
@@ -11,6 +14,15 @@ android {
...
         versionCode 1
         versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles 'consumer-rules.pro'
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
```
lognaturel, line 24: I just noticed that modules other than the main app don't get minified. What's the reason? I'm thinking about app size.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderService.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        return START_STICKY
+    }
+
+    private fun setupNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val notificationChannel = NotificationChannel(
+                NOTIFICATION_CHANNEL,
+                getLocalizedString(R.string.recording_channel),
```
lognaturel, line 62: Had to think about this one. I briefly wondered whether there might be a more general channel name that would make sense but I think this is the way to go.

strings/src/main/java/org/odk/collect/strings/LocalizedApplication.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Locale
+
+interface LocalizedApplication {
+
+    val locale: Locale
+}
+
+fun Context.getLocalizedString(stringId: Int, vararg formatArgs: Any): String {
+    return when (applicationContext) {
+        is LocalizedApplication -> {
```
lognaturel, line 14: You won't be surprised to hear me say that as cool as extensions are, they make me nervous. It just feels unexpected to have modifications to `Collect` happening way over here. In this specific case, can the actual `getLocalizedString` definition go in `Collect` and the extension only handle the case where it doesn't exist?

strings/build.gradle
seadowg: Yeah I made these consistent in this commit: https://github.com/getodk/collect/pull/4173/commits/879b9c6c34cb43a153b93ddb5ba8450f5e2544fa. Basically minification doesn't need to happen until you build an APK/App Bundle - that's where it decides what code it can remove from and then shrinks it. So, it doesn't make sense to have minification enabled or configured for library modules. Docs on this are pretty weak but here's an SO: https://stackoverflow.com/questions/50737180/proguard-on-multimodule-app

lognaturel: Oh!

strings/src/main/java/org/odk/collect/strings/LocalizedApplication.kt
seadowg: This is extending `Context`, not `Collect` (although I just typed that wrong way round twice 😂). `strings` is a dependency of `collect_app` so this is just adding a helper that any `Context` object can use rather than modifying a concrete type we own.

 I don't know if you spotted it, but in Java you don't end up with `Content#getLocalizedString(Int, Any)` you get  `getLocalizedString(Context, Int, Any)` (so a static with the first argument as the "receiver"). Extensions in Kotlin are just sugar and don't modify the interface of their receivers (so they can't access private state for example). Does that all still make you nervous? My Kotlin is all coming back to me pretty fast so definitely happy to tone it down! We could just make this a static in both Kotlin and Java.

audiorecorder/src/main/AndroidManifest.xml
seadowg: Ah right I always get confused about minimum API levels and when target changes the behaviour.

strings/src/main/java/org/odk/collect/strings/LocalizedApplication.kt
lognaturel: > This is extending Context, not Collect

`Collect` is the only implementation that would go through the first branch because it's the only `Context` that implements `LocalizedApplication`, right?

> you get getLocalizedString(Context, Int, Any)

And you can pass any `Context` in and it will "just work" but actually you'll only get the correct strings if you pass in a `LocalizedApplication`?

lognaturel: Now I'm super confused. Won't the usage made in `TranslationHandler` not work because generally the context passed in will be a fragment or activity which will go through to the else branch?

seadowg: > Collect is the only implementation that would go through the first branch because it's the only Context that implements LocalizedApplication, right?

Right. I originally looked at moving the `TranslationHandler` into `strings` but it ends up calling `LocaleHelper` which is currently pretty wrapped up in our preferences. So my thought was to separate out the logic of loading localized resources from however the preference of the locale was stored/defined.

> And you can pass any Context in and it will "just work" but actually you'll only get the correct strings if you pass in a LocalizedApplication?

Yeah. In my head the story is that `strings` supports lots of languages but it doesn't know if its dependents actually provide resource localization at an application level (rather than a system level). Also, it means we don't have to have our test Application objects care about provide a locale or implementing `LocalizedApplication` I can totally accept that I'm trying to over architect myself out of the hax that makes all this work (the messing with context to fool Android into loading different resources).

seadowg: > Now I'm super confused. Won't the usage made in TranslationHandler not work because generally the context passed in will be a fragment or activity which will go through to the else branch?

Sorry just saw this. It's branching on `applicationContext` (from `Context`) so it should get the application object in those cases.

lognaturel: I had totally missed that. I'm not very familiar with Kotlin so may not be best positioned for some of these reviews. I thought `applicationContext` was just the name you were giving to the context being extended.

lognaturel (approved):

strings/src/main/java/org/odk/collect/strings/LocalizedApplication.kt
seadowg: Ah it also looks like we're not being helped by Github's syntax highlighting here. Extension functions have `this` as an implicit "context" and then Kotlin is using its property access sugar to "sweeten" up the call to `getApplicationContext()`. In IntelliJ/Android Studio it highlights as a field but Github isn't giving us that ☹️.

PR #4182, opened 2020-10-23
Set main instance name to null when editing saved form

### LINKED ISSUE #4181: Constraints not immediately evaluated correctly when editing saved form
Collect v1.28

#### Problem description
Constraints not immediately evaluated correctly when editing saved form.

1. Load any form with a constraint
1. Fill out some data, save it
1. Edit the saved record
1. Put in an invalid value and notice you can swipe forward
1. Swipe back twice and notice you see the constraint message
1. Put in a valid value and notice you can't swipe forward
1. Swipe back twice and notice you can swipe forward (the valid value is accepted)

#### Expected behavior
Constraints should work correctly on the first try even when editing an existing instance.

#### Other information 
This is related to https://github.com/getodk/javarosa/pull/578. The main instance name is being set on instance deserialization which makes certain `TreeReference` comparisons fail including the one to check whether the user input should be used instead of the model value.

lognaturel:
Closes #4181

I started out writing a test in JavaRosa: https://github.com/getodk/javarosa/compare/master...lognaturel:constraints. However, the code that needs changing is copied in Collect so I'm not really sure where a test should ultimately live. I also tried the form [singleQ.xml.zip](https://github.com/getodk/collect/files/5427434/singleQ.xml.zip) with a single constraint.

This is a non-invasive change appropriate for bug fixing. The duplication between Collect and JavaRosa here is not ideal so I considered some rework but I don't think it's something we should do as part of a point release.

In https://github.com/getodk/javarosa/pull/578, we had identified that we had to reset the main instance name on form definition deserialization. We didn't consider that this was also necessary on form instance deserialization.

I can't think of any regression risk. This ensures that the main instance name is consistent before and after instance serialization.

Any form with constraints. [singleQ.xml.zip](https://github.com/getodk/collect/files/5427434/singleQ.xml.zip) is a simple one. The user at https://forum.getodk.org/t/collect-v1-28-2-error-about-two-different-forms-with-same-formid-and-version-and-constraints-not-accepting-valid-values/30645 reports that they've seen the problem the first time they were filling a blank form. It would be worth trying for a bit to reproduce that though I don't think it would be related.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java
@@ -396,6 +396,7 @@ public class FormLoaderTask extends AsyncTask<String, String, FormLoaderTask.FEC
         }
     }
 
+    // Copied from XFormParser.loadXmlInstance in order to set ExternalAnswerResolver for search()
     public static void importData(File instanceFile, FormEntryController fec) throws IOException, RuntimeException {
         // convert files into a byte array
         byte[] fileBytes = org.apache.commons.io.FileUtils.readFileToByteArray(instanceFile);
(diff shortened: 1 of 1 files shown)


lognaturel:
Marking this as `needs testing` right away because this may not address everything the user has experienced. Would be good to get @getodk/testers looking at it and see if they can find additional cases where constraints might not work as expected.

seadowg (approved):
I'm happy with this going in now but I think we should continue to the work on the JavaRosa side and see if this is something we can expose a tested API on that side or if we need to add coverage on the Collect side. It's a subtle bug so it feels like there is a rather high chance of regression.

mmarciniak90:
**Tested with success**

Verified on Android versions: 5.1, 7.0, 8.1, 9.0, 10.0

Cooperating with @kkrawczyk123, we verified the following cases:
* form attached in PR
* form attached on the forum
* regression forms with constraints
* newest created form with constraints on text and number question
* jump in/from hierarchy view
* jump to start/end
* rotation / minimization / lock screen
* verified finalize / unfinalized submissions

PR #4183, opened 2020-10-23
Fix displaying recordings over an hour

### LINKED ISSUE #4143: Audio recordings longer than an hour display incorrect length
Collect v1.28.1 (and earlier)

#### Problem description

Audio recordings longer than an hour only display the number of minutes past the hour. So, if you record for 1 hour and 14 seconds the length of the recording would be "00:14":

69143b791eef27e2a3d60a7db8fbcfbe2c74acda

1. Open a form with an audio question
2. Choose an audio file longer than an hour (easier than recording)

#### Expected behavior

If the time is greater than an hour we should probably display the time as "hh:mm:ss:". We should make sure this works for 100+ hours (although I doubt we'll see anyone record for that long).

seadowg:
Closes #4143

Added tests.

I was surprised I couldn't find a solution for this in JodaTime! Not a big deal to have a custom implementation of this as it doesn't need a lot of test coverage.

Only part of the app that's changed is the audio widget and how it displays the position and duration.

Any form with an audio question will do!

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/audio/AudioControllerView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/AudioControllerView.java
@@ -23,15 +23,20 @@ import android.widget.ImageButton;
 import android.widget.SeekBar;
 import android.widget.TextView;
 
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 import org.odk.collect.android.R;
 import org.odk.collect.android.databinding.AudioControllerLayoutBinding;
 
+import java.util.Locale;
+
 import static java.lang.Math.max;
 import static java.lang.Math.min;
+import static java.lang.String.format;
 
 public class AudioControllerView extends FrameLayout {
+
+    public static final int ONE_HOUR = 3600000;
+    public static final int ONE_MINUTE = 60000;
+    public static final int ONE_SECOND = 1000;
 
     public final AudioControllerLayoutBinding binding;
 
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/audio/AudioControllerView.java
```diff
@@ -101,10 +103,6 @@ private void onPositionChanged(Integer newPosition) {
         }
     }
 
-    private static String getTime(long seconds) {
-        return new DateTime(seconds, DateTimeZone.UTC).toString("mm:ss");
```
lognaturel, line 105: What happens if you just do HH:mm:ss here? Does it fail beyond 24 hours? Does that matter?

seadowg: Yeah beyond 24 hours we roll back over. It's probably not a huge problem as I don't imagine people are recording/choosing audio that long but using `DateTime` is a little weird.

lognaturel: Oh, I see. It also always puts the hours which is not what we want for clips under an hour.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified cases:
* added ~1h audio
* added ~2h audio
* added audio from GD
* added audio from device
* recorded ~1h audio
* removed response from a new option in player
* removed response from dialog
* light / dark / magenta mode
 
Additionally, I took a look at the new audio player. Audio player remembers where audio left off before rotation -> after the second rotation, it looks like audio goes back to the beginning but when a user plays it, it is continued -> next rotations cause that audio goes back to the beginning. This behavior seems to be better than before where audio goes back to the beginning after each rotation

lognaturel:
Thanks, @mmarciniak90! Since we're going to be doing a lot more QA on audio recording soon, I don't think we need to verify other Android versions for this one. I think we're ready to merge.

mmarciniak90:
> I don't think we need to verify other Android versions for this one.

OK!

PR #4185, opened 2020-10-23
Add additional audio analytics

seadowg:
Just a quick follow up to #4178


collect_app/src/main/java/org/odk/collect/android/audio/AudioControllerView.java
```diff
@@ -193,10 +210,20 @@ public void onStopTrackingTouch(SeekBar seekBar) {
...
+                analytics.onSeek();
+            }
         }
 
         Boolean isSwiping() {
             return swiping;
         }
     }
+
+    public interface AnalyticsListener {
```
seadowg, line 224: Felt less intrusive just to create a separate listener here rather than adding (and kind of duplicating) methods to the main listener.

lognaturel (approved):

PR #4187, opened 2020-10-26
Use JavaRosa v3.0.3 to fix bug in complex calculates

lognaturel:
See https://github.com/getodk/javarosa/issues/611 for description of bug and https://forum.getodk.org/t/repeats-and-calculations-inside-nested-repeats-behaving-differently-in-odk-1-27-3-vs-1-28-2/30694 for a user form.

Wrote and ran JR test at https://github.com/getodk/javarosa/pull/612, interactively reviewed with @seadowg, ran with [complex_calc.xml.zip](https://github.com/getodk/collect/files/5436315/complex_calc.xml.zip).

It's the simplest.

Should be a safe bug fix.

[complex_calc.xml.zip](https://github.com/getodk/collect/files/5436317/complex_calc.xml.zip). This is a form with repeats. Each repeat has a question with text that represents 2*repeat position-4. Prior to the fix, it was showing 2*repeat position-.

There's also a user form at https://forum.getodk.org/t/repeats-and-calculations-inside-nested-repeats-behaving-differently-in-odk-1-27-3-vs-1-28-2/30694 but it's pretty complex. I ended up jumping to the section highlighted in the original writeup and then going backwards to fill needed answers.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -262,7 +262,7 @@ dependencies {
     implementation "com.rarepebble:colorpicker:3.0.1"
     implementation "commons-io:commons-io:2.5" // Commons 2.6+ introduce java.nio usage that we can't access until our minSdkVersion >= 26 (https://developer.android.com/reference/java/io/File#toPath())
     implementation "net.sf.opencsv:opencsv:2.4"
-    implementation("org.getodk:javarosa:3.0.2") {
+    implementation("org.getodk:javarosa:3.0.3") {
         exclude group: 'joda-time'
         exclude group: 'org.slf4j'
     }


lognaturel:
@getodk/testers I'm not really sure this warrants QA but since you have a full day ahead before I can do the release, I think it makes sense for you to at least verify the small test form.

I do suspect this is related to https://github.com/getodk/collect/issues/4122 but I haven't had a chance to try it. I think it'd be something like a relevance condition in a repeat that depends on a value that might not yet be computed when a select minimal question is first open. The relevance then changes when the select value is saved.

seadowg (approved):

mmarciniak90:
**Tested with success**

* _complex_calc_ form
* _Digital_IFP_Filing_202010_ form
* add repeat group by :heavy_plus_sign: icon on hierarchy view
* add repeat group by dialog
* removed group and add it again
* removed eg. the first group - values are recalculated 
* load data from savepoint when the app was closed
* load data from savepoint when theme was changed
* added additional text to a calculated value
* rotation, minimization, lock screen
* light/dark mode

I compared each result with behavior on v1.27.3.

mmarciniak90:

PR #4188, opened 2020-10-26
Extend existing external application functionality to receive binary files in field-list groups

### LINKED ISSUE #3653: Undesirable button is visible on readOnly video widgets.
Collect v maser, Android v probably all, checked on 7-10, device used...

#### Problem description
The undesirable button is visible on readOnly version of video widget and selfie video widget:
Screenshot (Feb 26, 2020 10_46_04 AM)

Just open readOnly version of video/selfie video widget. 

#### Expected behavior
The behavior from Collect 1.25.2:
Screenshot_20200226-104517

#### Other information 
The form that I have used: 
[all-widgetsReadOnly.xml.txt](https://github.com/opendatakit/collect/files/4254861/all-widgetsReadOnly.xml.txt)

### LINKED ISSUE #4071: Extend existing external application functionality to receive binary files
[Draft specification](https://docs.google.com/document/d/1JbKiwgBCuI2t3Cd8n5OnyrNce3woq20ncZbN3oGcPhU/edit)

grzesiek2010:
Closes #4071
Closes #3653 

This is the first part which contains support for field-list groups what is more important at this moment than supporting single widgets. I'll add support for single widgets once we merge this pr.

Testing this app requires an external app and I prepared one which you can find here https://github.com/grzesiek2010/CollectAnswersProvider

I tested the app manually and added some automated tests.

This extends the current functionality described in https://docs.getodk.org/launch-apps-from-collect/#launching-external-apps-to-populate-multiple-fields so it's the best possible solution.

This change shouldn't affect normal users (users that doesn't use external apps to receive values). It just extendshttps://docs.getodk.org/launch-apps-from-collect/#launching-external-apps-to-populate-multiple-fields and we can focus on that functionality during testing. 

[external_widgets.xlsx](https://github.com/getodk/collect/files/5439234/external_widgets.xlsx)

Yes.


collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -80,17 +95,109 @@ public void appMissingErrorText_ShouldComeFromSpecialFormText() {
...
+    public void externalApp_ShouldPopulateStringFields() {
         Intent resultIntent = new Intent();
-        resultIntent.putExtra("sometext", "Here is a text value");
 
-        int randomInteger = (new Random()).nextInt(255);
-        resultIntent.putExtra("someinteger", randomInteger);
+        int randomInteger = new Random().nextInt(255);
+        double scale = Math.pow(10, 2);
+        double randomDecimal = Math.round(new Random().nextDouble() * scale) / scale;
+        String randomText = RandomString.make();
```
seadowg: Do these values need to be random or are you just using random values to generate test data?

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -80,17 +95,109 @@ public void appMissingErrorText_ShouldComeFromSpecialFormText() {
...
+
+    @Test
+    public void externalApp_ShouldPopulateImageField() throws IOException {
+        onView(withTagValue(is("ImageView"))).check(doesNotExist());
+        onView(withId(R.id.capture_image)).check(doesNotExist());
+        onView(withId(R.id.choose_image)).check(doesNotExist());
+
+        Intent resultIntent = new Intent();
+
+        File tmpJpg = new File(new StoragePathProvider().getTmpFilePath());
```
seadowg: You can just use `Files.createTempFile` here I think. Then there's no need to use the `StoragePathProvider` in here.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -80,17 +95,109 @@ public void appMissingErrorText_ShouldComeFromSpecialFormText() {
...
         onView(withText("This is buttonText")).perform(click());
 
-        onView(withText("Here is a text value")).check(matches(isDisplayed()));
         onView(withText(Integer.toString(randomInteger))).check(matches(isDisplayed()));
+        onView(withText(String.valueOf(randomDecimal))).check(matches(isDisplayed()));
+        onView(withText(randomText)).check(matches(isDisplayed()));
+    }
+
+    @Test
+    public void externalApp_ShouldPopulateImageField() throws IOException {
```
seadowg: Could this just be one test where the group contains every kind of data? That might be harder to read though I guess... Interested in what you think!

collect_app/src/main/java/org/odk/collect/android/formentry/questions/QuestionDetails.java
```diff
@@ -24,4 +30,8 @@ public FormEntryPrompt getPrompt() {
     public String getFormAnalyticsID() {
         return formAnalyticsID;
     }
+
+    public boolean isReadOnly() {
```
seadowg, line 34: Nice! Really like this being here.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -100,6 +103,39 @@
     private FileUtils() {
     }
 
+    public static void saveMediaFileFromUri(Uri uri, File destFile, Context context) {
```
seadowg: Could this maybe be `answerFile` or `instanceFile`? The `QuestionMediaManager` refers to these as "answer files". I'm frequently getting confused between files in the instance folder and files that come with the form.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -475,40 +483,70 @@ public void scrollTo(@Nullable QuestionWidget qw) {
...
+                for (QuestionWidget questionWidget : widgets) {
+                    FormEntryPrompt prompt = questionWidget.getFormEntryPrompt();
+
+                    if (FormEntryPromptUtils.getQuestionName(prompt).equals(key.toString())
+                            && prompt.getDataType() == Constants.DATATYPE_BINARY) {
+                        Uri uri = clipData.getItemAt(i).getUri();
+                        if (uri != null) {
+                            try {
+                                File destFile = FileUtils.createDestinationMediaFile(formController.getInstanceFile().getParent(), ContentResolverHelper.getFileExtensionFromUri(getContext(), uri));
+                                FileUtils.saveMediaFileFromUri(uri, destFile, getContext());
```
seadowg: This could maybe use the `QuestionMediaManager` in some way. I've ended up playing with that in #4190 so might be better to come back fro that later!

seadowg (changes requested):
Just a few changes. Probably good for @lognaturel to have a look as well as she' much more familiar with the feature than me!

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
grzesiek2010: Right

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
grzesiek2010: It was just for test data it's not required and I got rid of it to simplify the long `externalApp_ShouldPopulateFields` test.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
grzesiek2010: The test will be long but yeah in this case where it's about populating multiple fields at the same time it's maybe better to have one test.
It's Improved.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
grzesiek2010: Ok, done.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
grzesiek2010: That might be a good idea, I've added a TODO comment to improve it later.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -80,17 +92,67 @@ public void appMissingErrorText_ShouldComeFromSpecialFormText() {
...
+        onView(withId(R.id.play_video)).perform(nestedScrollTo()).check(matches(isDisplayed()));
+        onView(withId(R.id.play_video)).check(matches(isEnabled()));
+
+        // Check ArbitraryFileWidget with answer
+        onView(withTagValue(is("ArbitraryFileWidgetAnswer"))).perform(nestedScrollTo()).check(matches(isDisplayed()));
+    }
+
+    @SuppressWarnings("PMD.DoNotHardCodeSDCard")
+    private Uri createTempFile(String name, String extension) throws IOException {
+        File file = File.createTempFile(name, extension, new File("/sdcard"));
```
seadowg, line 253: Out of interest did you need the parent (`new File("/sdcard")`) here? I was under the impression we could just create a file outright and Android handles it.

seadowg (approved):
Looks good!

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -479,40 +487,71 @@ public void scrollTo(@Nullable QuestionWidget qw) {
...
+                        ((StringWidget) questionWidget).setDisplayValueFromModel();
+                        break;
                     }
+                }
+            }
+        }
+
+        if (clipData != null) {
+            for (int i = 0; i < clipData.getItemCount(); i++) {
+                CharSequence key = clipData.getItemAt(i).getText();
```
lognaturel: See my comments on the test -- I don't think this matches the spec we had written up. It looks like it has the external app set both the text and a URI for each clip data item. The way we had written it up was to for ClipData, only set the URI and use an intent extra to do the match between Collect field name and clipdata URI. You should be able to read the extra to match with the Collect field (same as non-binary types) and then use the extra's value to get the file. I don't think you need to walk the `ClipData` at all, it's just there to provide permissions for when you resolve the content.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -80,17 +92,67 @@ public void appMissingErrorText_ShouldComeFromSpecialFormText() {
...
-        resultIntent.putExtra("sometext", "Here is a text value");
 
-        int randomInteger = (new Random()).nextInt(255);
-        resultIntent.putExtra("someinteger", randomInteger);
+        resultIntent.putExtra("questionInteger", "25");
+        resultIntent.putExtra("questionDecimal", "46.74");
+        resultIntent.putExtra("questionText", "sampleAnswer");
 
+        ClipData clipData = ClipData.newRawUri(null, null);
+        clipData.addItem(new ClipData.Item("questionImage", null, createTempFile("famous", "jpg")));
```
lognaturel: According to our spec, an external app is supposed to:
> Add content URIs for the files to attach to the form to the return intent’s ClipData

> Add an intent extra with desired form field name and set its value to the content URI for the corresponding file.

lognaturel (changes requested):
I don't think this matches the spec. Let's touch base if my comments don't seem right.

I also didn't check the behavior in case of any of these failures:
> There is an intent extra with the name of the field but its value doesn’t match any of the ClipData items’ content URIs
> The question type and URI MIME type don’t match (i.e. anything is allowed if the question type is file, only image types are allowed if the question type is image, etc)
> Collect doesn’t have permissions to read the file at a provided URI
> Collect can’t copy the file into its instance directory (e.g. device out of space)

Any existing value should be cleared. Maybe add a test for one of the cases?

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
lognaturel: The spec document says:
> Look for an intent extra with the name of the field. If one is found, attempt to exactly match the intent extra value to one of the ClipData items’ content URI
> If there’s a match, copy the file to the form instance folder and write its local filename in the corresponding form field

We don't really need to do that, I don't think. We can just try to resolve the URI from the extra as I described above. If that works, it means the ClipData was correctly populated.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -480,39 +486,58 @@ public void scrollTo(@Nullable QuestionWidget qw) {
...
+                            case Constants.DATATYPE_BINARY:
+                                try {
+                                    Uri uri = (Uri) bundle.get(key);
+                                    try {
+                                        File destFile = FileUtils.createDestinationMediaFile(formController.getInstanceFile().getParent(), ContentResolverHelper.getFileExtensionFromUri(getContext(), uri));
+                                        //TODO might be better to use QuestionMediaManager in the future
+                                        FileUtils.saveAnswerFileFromUri(uri, destFile, getContext());
+                                        ((WidgetDataReceiver) questionWidget).setData(destFile);
+                                    } catch (SecurityException e) {
+                                        Timber.w(e);
```
lognaturel: Will this implicitly clear any existing value?

I think there's one more error case missing -- if the type of the file passed from the external app doesn't match the question type. For example, a .mp3 file shouldn't be allowed for an image widget.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
lognaturel: This test needs to be updated!

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
grzesiek2010: Without it I get `java.io.IOException: Read-only file system`

grzesiek2010:
> There is an intent extra with the name of the field but its value doesn’t match any of the ClipData items’ content URIs

I iterate extras if i find URI I just use it I don't check clipdata too as we discussed yesterday. So if URIs are not equal that means permission for the one we are going to use (from extras) is not granted and it will throw `SecurityException` what I catch in https://github.com/getodk/collect/pull/4188/files#diff-3c11347037b799c07e149defff1aea5cfb36d6e611117bcff73c1096a8058d4dR528

and I added a test that throws such SecurityException.

so generally this case and the one you described below: `Collect doesn’t have permissions to read the file at a provided URI` is the same.

> The question type and URI MIME type don’t match (i.e. anything is allowed if the question type is file, only image types are allowed if the question type is image, etc)

We don't check file types we receive starting those widgets normally so why would we do that here. Of course we can improve that but then it should be a separate pr which would fix all the cases not only intents group. Every widget that can receive such files has its own `setData()` method and there probably we should check file types if we need.

> Collect can’t copy the file into its instance directory (e.g. device out of space)

[I catch any Exception and Error](https://github.com/getodk/collect/pull/4188/files#diff-3c11347037b799c07e149defff1aea5cfb36d6e611117bcff73c1096a8058d4dR528) and I added tests that confirm that the app doesn't crash.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
grzesiek2010: When it comes to file formats please take a look below I described that it should be a separate pr if needed because we don't check file types using binary widgets normally so why here in intents group?

> Will this implicitly clear any existing value?

we call `setData()` in the last step so if any exception occurs it's just not called and doesn't remove an existing answer.

lognaturel: > doesn't remove an existing answer

This is a weird case -- a user would have to get data from an external app, have it work fine, then get data again and have it fail. It's not very likely but then if it does happen I think it's nice for the answer to be cleared so it's easier for a user to see that something unexpected happened. It should be straightforward to clear in each `catch` block, right?

grzesiek2010: Again like in case of checking file types I think it wouldn't align with what normal widgets outside intent groups do. If something goes wrong we don't remove an existed answer. We remove the old answer only in setData() so just before adding a new one where we are almost sure that everything is fine.

lognaturel: Next time let's make sure we have these conversations when we're considering the spec and make sure the implementation is based on the spec! I'll file issues once this is merged.

lognaturel (approved):

lognaturel:
> We don't check file types we receive starting those widgets normally so why would we do that here.

Those widgets filter what can be selected so it shouldn't be possible through the UI to select a file of the wrong type. An app returning a binary file could be too lenient and cause problems downstream. I'm ok doing the type check as another PR but I think it's special behavior for external applications.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
seadowg: Huh strange. The same thing just worked™️ for me here: https://github.com/getodk/collect/pull/4190/files#diff-83ee5909b031fabf1a76172164745099e708b293a264d0ae3f2cf9dce844ae21R154.

grzesiek2010: Ah ok this way it could work but then I need to use that file via fileprovider that means I would need to edit our `provider_paths.xml` file and add `<cache-path path="/" name="files" />` there. Since we would need that only for tests I would need to create such a separate file for tests only. Is it worth the effort?

lognaturel:
Quick thing about the layout I noticed playing with a "real" form. Binary widgets now have no vertical space below the question label:

 

I think that looks strange in a field-list. I assume that with the EditText-based question types the EditText is there but not displayed. Maybe doing something similar with the top button for binary question types would be possible?

grzesiek2010:
@mmarciniak90 @kkrawczyk123 
this pr contains #4199, both should be tested at once.

grzesiek2010:
> I think that looks strange in a field-list. I assume that with the EditText-based question types the EditText is there but not displayed. Maybe doing something similar with the top button for binary question types would be possible?

we agreed to treat those widgets that are in such an intents group as read-only so that the layout is the same as in case of single read-only widgets and I think that was a good decision because it's easy and we can avoid complexity.
What we can do here is for example adding some empty same for read-only widgets that don't have any answer but again to make it easy we should do that for all read-only widgets (without answers) not matter if they are placed in an intents group or in a normal field-list group or just used as single questions, otherwise we will end up with plenty if check or even completly new widgets.

lognaturel:
I think the question is whether we should rethink the display for read-only binary widgets and I think the answer is yes. I'll file an issue once this is merged.

mmarciniak90:
**Tested with success**

Verified cases:
* generated answer to single question on page
* generated answers to question in field-list
* button on readOnly video widget (3653)
* style of button for launching an external app (4199)
* send submission to Central and Aggregate
* increased font size
* RTL language
* used forward/backward navigation buttons
* removed response
* toast is visible when external app is not available
* launch counter app
* used external Tester app
* Light/dark/magenta themes
* edited form
* don't keep activities
* rotation
* minimization

The only thing that we noticed is that page is not displaying from the beginning after the answers generation.

kkrawczyk123:
Tested with success on Androids: 5.1, 7.1 and 9 as well.

grzesiek2010:
> The only thing that we noticed is that page is not displaying from the beginning after the answers generation.

I'm not sure but I guess it might be caused by this part of code:
https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L2729

@lognaturel if you feel like we need to fix it please file an issue.

Issue #4151, opened 2020-10-06
Clicking notification while recording returns to app


seadowg:
**Acceptance**

- [x] Given I'm recording using the in-app recorder
When I switch to another app
And then open the notification draw
And I click on the "Recording..." notification
Then I return to Collect
And the recording is still in progress

The notification should also be low priority (like other "ongoing" notifications) and not make a sound/vibrate.

PR #4190, opened 2020-10-28
Make audio question record within form entry

### LINKED ISSUE #4186: Can answer other questions while recording
- [x] Given I'm on a screen with multiple questions
And one of the questions is an audio question
And I am using the in-app recorder
When I click "Record Sound"
Then the "Stop Recording" button is visible on screen
And the audio question buttons are gone
And the audio question showing "Recording..." in the answer
And I can answer other questions

When I rotate the screen
Then recording continues
When I swipe to another screen
Then the recording stops
When I swipe back
Then the question that was recording is unanswered

- [x] Given I'm on a screen with multiple questions
And more than one of the questions is an audio question
And I am using the in-app recorder
When I click "Record Sound"
Then all the audio questions have disabled "Record Sound" buttons

- [x] Given I'm on a screen with multiple questions
And one of the questions is an external string question
And I am using the in-app recorder
When I answer the external string question
And stop the recording
Then nothing goes wrong

seadowg:
Closes #4186
~~Blocked by #4180~~

Comments inline.

Should just add the new feature! I think it would be good to have QA play with both internal and external recording (you can switch between them in "Experimental" settings). There is a lot that could go wrong with have the recording happen while you answer questions, so I'd love to see if they can break it!

I'm going to mark this as "needs testing" even though it's still a draft so we can find any problems early while reviewing.

Any form with audio questions!


audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderViewModel.kt
```diff
@@ -1,46 +1,14 @@
...
-        recordingRepository.clear()
-    }
-
-    class Factory @Inject constructor(private val application: Application, private val recordingRepository: RecordingRepository) : ViewModelProvider.Factory {
-
-        override fun <T : ViewModel?> create(modelClass: Class<T>): T {
-            return AudioRecorderViewModel(application, recordingRepository) as T
-        }
-    }
+abstract class AudioRecorderViewModel : ViewModel() {
```
seadowg, line 12: The thinking here is that `collect_app` doesn't need to know how to construct `AudioRecorderViewModel`. It also means we can mock/fake safely in `collect_app` (which we want to do) without needing to deal with the real constructor or method implementations.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/RealAudioRecorderViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Transformations.map
+import org.odk.collect.audiorecorder.recording.AudioRecorderViewModel
+import org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.Companion.ACTION_CANCEL
+import org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.Companion.ACTION_START
+import org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.Companion.ACTION_STOP
+import org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.Companion.EXTRA_SESSION_ID
+import java.io.File
+
+internal class RealAudioRecorderViewModel internal constructor(private val application: Application, private val recordingSession: RecordingSession) : AudioRecorderViewModel() {
```
seadowg: Very lazy naming. Couldn't think of anything better.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/RecordingSession.kt
```diff
@@ -0,0 +1,31 @@
+package org.odk.collect.audiorecorder.recording.internal
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import java.io.File
+
+class RecordingSession {
```
seadowg: This felt like a more apt name than `Repository`

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/AudioRecordingTest.java
```diff
@@ -0,0 +1,175 @@
...
+import java.io.IOException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.odk.collect.android.support.FileUtils.copyFileFromAssets;
+
+@RunWith(AndroidJUnit4.class)
+public class AudioRecordingTest {
+
+    private final FakeAudioRecorderViewModel fakeAudioRecorderViewModel = new FakeAudioRecorderViewModel();
```
seadowg: Faking here is nice as we have a clearly defined boundary between the `collect_app` and `audiorecorder` and it means the test isn't exposed to any issues that could occur given we're trying to access the mic and using a service in the real implementation.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/AudioRecordingTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.odk.collect.android.support.FileUtils.copyFileFromAssets;
+
+@RunWith(AndroidJUnit4.class)
+public class AudioRecordingTest {
```
seadowg, line 37: I'd been trying to resist writing an Espresso test due to the feature being isolated within the widget but making recording work inline requires collaboration between a few different things (the widget and the activity for instance) so it started to feel too disjointed just writing tests at the component level. 

I do like where these tests have ended up however as they capture the high level navigation features without having to deal with any of the detail of audio recording or playback.

collect_app/src/main/java/org/odk/collect/android/formentry/FormIndexAnimationHandler.java
```diff
@@ -10,6 +10,11 @@
  */
 public class FormIndexAnimationHandler {
```
seadowg, line 11: I think I managed to take this another step in the right direction. A lot of work to do still though!

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppRecordingRequester.java
```diff
@@ -50,4 +52,14 @@ public void denied() {
...
     }
+
+    @Override
+    public void onIsRecordingChanged(Consumer<Boolean> isRecordingListener) {
+        isRecordingListener.accept(false);
+    }
+
+    @Override
+    public void onRecordingAvailable(FormEntryPrompt prompt, Consumer<String> recordingAvailableListener) {
+        // This could be implemented using the new Activity Result API  once it's stable
```
seadowg, line 63: This basically. We could massively simplify `AudioWidget`'s implementation by just speaking to this interface.

seadowg:
I'm removing "needs testing" as I'd like to change some things.

seadowg:
Ready to go again.

mmarciniak90:
@seadowg I downloaded the latest changes and I'm not able to save recorded audio.



The second case is that changes in audio recorders in experimental settings are not applied immediately when we set it in the middle of the form.



seadowg:
@mmarciniak90 hhmmm. For the second one could you make an issue? I think we'll add that separately. I'm just going to check on the first as I thought I had it all working on Friday. What device/OS are you using?

EDIT: No that's definitely broken. I must have forgotten to double-check when I pushed Friday night

seadowg:
@mmarciniak90 sorry about that! I had written a test incorrectly so that I thought everything was working but it clearly wasn't. I've fixed the test and the code so you should be able to play with recording now.

mmarciniak90:
@seadowg I noticed a lot of crashes. I think all of them are related to one problem. When I swipe to another screen then the recording DOES NOT stop.

seadowg:
@mmarciniak90 ah I think I see the problem. I think I need to revise some of the way this stuff is tested but I'll push a fix ASAP for you so it's not blocking you.

seadowg:
@mmarciniak90 ok fixed!

mmarciniak90:

kkrawczyk123:
@seadowg I have noticed some not necessarily desired behavior:
- When you minimize the app in the middle of recording with don't keep activities enabled - recording is still continued.
- There is also inconsistency in behavior: when use the hierarchy button recording is interrupted but when go to general settings it is continued, you can change some settings like mode or language in the middle of recording and go back to the form and see that recording is still ongoing. I believe it should be unified.

seadowg:
> When you minimize the app in the middle of recording with don't keep activities enabled - recording is still continued.

Yeah this is intended. The phone could be locked while recording is in progress and the activity killed to free up memory. In that case (which "Don't keep activities" is simulating) we want the recording to continue.

> There is also inconsistency in behavior: when use the hierarchy button recording is interrupted but when go to general settings it is continued, you can change some settings like mode or language in the middle of recording and go back to the form and see that recording is still ongoing. I believe it should be unified.

Great point. I've addressed that in #4154 so I think we're fine to let that be inconsistent here.

mmarciniak90:
@seadowg thanks! So we are ready to close this PR

**Tested with success**

Verified cases:
- lock screen
- rotation
- minimization
- phone conversation
- open hierarchy view -> recording is stopped and the record is lost
- swipe to next/prev question -> recording is stopped and the record is lost
- General settings -> possible unexpected behavior
- one audio question on page
- multi audio questions on page -> audio questions have disabled "Record Sound" buttons 
- multi audio questions on page -> using other widgets like location, image, video, barcode
- launching exteral app when audio is recorded
- readOnly audio widget
- denied permissions
- form `All widgets field-list-NO-OSM`
- remove response
- RTL language
- edit saved form

kkrawczyk123:

collect_app/src/main/java/org/odk/collect/android/formentry/saving/FormSaveViewModel.java
```diff
@@ -321,11 +322,19 @@ public void replaceAnswerFile(String questionIndex, String fileName) {
 
     @Override
     public String createAnswerFile(File file) {
+        String newFileHash = FileUtils.getMd5Hash(file);
+        String instanceDir = formController.getInstanceFile().getParent();
+
+        File[] answerFiles = new File(instanceDir).listFiles();
+        for (File answerFile : answerFiles) {
+            if (FileUtils.getMd5Hash(answerFile).equals(newFileHash)) {
+                return answerFile.getName();
```
lognaturel: Huh. I think this might actually address https://forum.getodk.org/t/testing-form-on-central-upload-error-a-resource-already-exists/30947? That would be good. @getodk/testers could you please quickly give this a try? On master, submitting All Widgets to Central with the same image file attached twice should fail as in the forum post. I think it will succeed after this change.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/DaggerSetup.kt
```diff
@@ -66,7 +67,7 @@ internal open class AudioRecorderDependencyModule {
 
     @Provides
     @Singleton
-    open fun providesRecordingRepository(): RecordingRepository {
+    open fun providesRecordingSession(): RecordingRepository {
```
lognaturel: Should be `Repository` again.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/RecordingRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+            sessions.getOrPut(it) { MutableLiveData(null) }.value = recording
+        }
+
+        _currentSession.value = null
+    }
+
+    fun clear() {
+        _currentSession.value = null
+
+        sessions.values.forEach {
```
lognaturel, line 33: The UI doesn't actually make it possible for multiple sessions to have files associated with them, right? That would mean that more than one recording were started, then stopped, and no cleanup happened?

And in fact, doesn't `AudioRecorderService` already ensure the file is deleted with `cancel`?

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/AudioRecordingTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public final TestDependencies testDependencies = new TestDependencies() {
+        @Override
+        public AudioRecorderViewModelFactory providesAudioRecorderViewModelFactory(Application application) {
+            return new AudioRecorderViewModelFactory(application) {
+                @Override
+                public <T extends ViewModel> T create(@NonNull Class<T> modelClass) {
+                    if (stubAudioRecorderViewModel == null) {
+                        try {
+                            File stubRecording = File.createTempFile("test", ".4a");
```
lognaturel: .4a -> m4a (doesn't really matter but might as well)

lognaturel (commented):
The notification makes a sound on my Android 10 device. It's really important it doesn't make one or it will show up on every recording. Even if part of the answer is to set the notification priority to low, I think it would be good to manually set the sound off. Seems sound behavior for notifications is pretty inconsistent - https://stackoverflow.com/questions/45919392/disable-sound-from-notificationchannel It could also go in another PR. @getodk/testers neither of you got notification sounds?

Other than that, I don't really have much to say. It was definitely helpful to walk through chunks of it together.

One thing that I think should Just Work but that I didn't verify is what happens when you save a value for a recording (rec1), re-open the instance, record again (rec2), exit without saving, reopen the instance. The recording should be rec1. @getodk/testers did you try that as part of "edit saved form"?

lognaturel:
For @getodk/testers:
- [x] save a value for a recording (rec1), re-open the instance, record again (rec2), exit without saving, reopen the instance. The recording should be rec1. @getodk/testers did you try that as part of "edit saved form"?
- [ ] https://github.com/getodk/collect/pull/4190#discussion_r520796476

lognaturel (approved):
Since there are no significant changes to make, approving so you can merge at your convenience, @seadowg.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/RecordingRepository.kt
seadowg: > The UI doesn't actually make it possible for multiple sessions to have files associated with them, right?

Yeah that's right. `InternalRecordingRequester` calls `cleanUp` after each recording. This implementation means that we don't **need** to do that and could call `cleanUp` when exiting instead. Assuming that `cleanUp` was called after each recording actually made it awkward (for me) so I ended up making it a bit more general purpose but giving myself the tools for the "optimal" implementation on the UI/app side.

> And in fact, doesn't AudioRecorderService already ensure the file is deleted with cancel?

This only happens if `cleanUp` is called.

seadowg:
> The notification makes a sound on my Android 10 device. It's really important it doesn't make one or it will show up on every recording.

Ah yeah I think I missed this as I have the sound off on my test emulator but I do think we need to change the notification priority and turn sound off. I can do that as part of #4151.

seadowg:
> save a value for a recording (rec1), re-open the instance, record again (rec2), exit without saving, reopen the instance. The recording should be rec1. @getodk/testers did you try that as part of "edit saved form"?

I couldn't reproduce this. How are you doing the "exit without saving": ignore changes, quitting the app?

seadowg:
I'll merge for the moment. We can add a bug for @lognaturel's case once we end up reproducing.

lognaturel:
Sorry, I should have been more specific! The first item in my checklist describes the intended behavior. There's somewhat unexpected code that saves media files in a map if an instance is edited so that if changes are discarded the original media is re-established. I don't think there are automated tests for this. I also don't think this PR touched that behavior but I think a sanity check would be good.

My second point relates to an error case in Central when two images with different filenames but the same contents are uploaded. I think that this PR makes that case impossible because the same filename would be used. It would be good to double check so we can confirm whether we need to do anything on Central.

PR #4191, opened 2020-10-28
Changes in metadata required to update the app to API 29

grzesiek2010:
This pr addresses two first steps from https://github.com/getodk/collect/issues/3360#issuecomment-691734743

I tested the changes manually.

This is what we have to do because of changes in privacy policy.

It removes two of visible in settings properties SIM serial and subscriber ID then it makes the deviceId not dangerous so if a user opens a form with that metadata they shouldn't be asked for `ReadPhoneState` permission (unless phone number is also present in metadata which is the last property which requires the permission).

[meta.xlsx](https://github.com/getodk/collect/files/5452519/meta.xlsx)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
+++ b/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
@@ -57,7 +57,6 @@ public class PropertyManager implements IPropertyManager {
     public static final String SCHEME_USERNAME     = "username";
     private static final String SCHEME_TEL          = "tel";
     private static final String SCHEME_MAILTO       = "mailto";
-    private static final String SCHEME_IMEI         = "imei";
 
     private final Map<String, String> properties = new HashMap<>();
 
@@ -92,7 +91,7 @@ public class PropertyManager implements IPropertyManager {
 
     public PropertyManager reload() {
         try {
-            putProperty(PROPMGR_DEVICE_ID,     SCHEME_IMEI,         deviceDetailsProvider.getDeviceId());
+            putProperty(PROPMGR_DEVICE_ID,     "",         deviceDetailsProvider.getDeviceId());
             putProperty(PROPMGR_PHONE_NUMBER,  SCHEME_TEL,          deviceDetailsProvider.getLine1Number());
         } catch (SecurityException e) {
             Timber.e(e);
--- a/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
+++ b/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
@@ -153,9 +153,7 @@ public class PropertyManager implements IPropertyManager {
      * @return True if the given property is dangerous, false otherwise.
      */
     private boolean isPropertyDangerous(String propertyName) {
-        return propertyName != null
-                && (propertyName.equalsIgnoreCase(PROPMGR_DEVICE_ID)
-                || propertyName.equalsIgnoreCase(PROPMGR_PHONE_NUMBER));
+        return propertyName != null && propertyName.equalsIgnoreCase(PROPMGR_PHONE_NUMBER);
     }
 
     @Override
--- a/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
+++ b/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
@@ -53,7 +53,6 @@ import static org.odk.collect.android.preferences.GeneralKeys.KEY_USERNAME;
 public class PropertyManager implements IPropertyManager {
 
     public static final String PROPMGR_DEVICE_ID        = "deviceid";
-    public static final String PROPMGR_SUBSCRIBER_ID    = "subscriberid";
     public static final String PROPMGR_SIM_SERIAL       = "simserial";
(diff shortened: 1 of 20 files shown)


grzesiek2010:
@lognaturel 
in the https://github.com/getodk/collect/issues/3360#issuecomment-691734743 you said that we should remove READ_PHONE_STATE permission but it's still required in case of phone number which we still keeps or maybe I'm missing something?

If we open a form with phone number metadata we are asked for permission, if we deny the form is closed. Maybe now since the permission is related only to that one property it's time to change that behavior and open the form but without that value?

lognaturel:
> it's still required in case of phone number

Ah, yes. https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers describes the change we should make.

> since the permission is related only to that one property it's time to change that behavior and open the form but without that value

Closing the form seems right to me -- if the form designer is requesting the phone number, they intend it to be populated. Essentially the data collector is refusing to fill in part of the form.

grzesiek2010:
>Ah, yes. https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers describes the change we should make.

that doc applies to Android 11 but our target is 10 so it's something we will need to do later not now.

> Closing the form seems right to me -- if the form designer is requesting the phone number, they intend it to be populated. Essentially the data collector is refusing to fill in part of the form.

ok so let's leave it as is.

collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
```diff
@@ -112,12 +92,8 @@ public PropertyManager(Application application, RxEventBus rxEventBus, Permissio
 
     public PropertyManager reload() {
         try {
-            // Device-defined properties
-            IdAndPrefix idp = findDeviceId(context, deviceDetailsProvider);
-            putProperty(PROPMGR_DEVICE_ID,     idp.prefix,          idp.id);
+            putProperty(PROPMGR_DEVICE_ID,     SCHEME_IMEI,         deviceDetailsProvider.getDeviceId());
```
lognaturel: What effect does having `SCHEME_IMEI` here have? Seems it should be `collect`, maybe?

grzesiek2010: That's a good question, I wanted to ask about it because I'm not sure... I kept SCHEME_IMEI for now but it's not IMEI anymore so it doesn't make much sense. I could do what you say and add collect: but our new device id is already like `collect:5sf45sf5dsf5d4f` (it starts with collect: always) so a result the value with scheme would be `collect:collect:5sf45sf5dsf5d4f`

Of course depends where we use it because in settings and in a form it would be still just `collect:5sf45sf5dsf5d4f` (because there we get the value without scheme) but during sending a form we use:
```
deviceId = new PropertyManager(Collect.getInstance().getApplicationContext())
                    .getSingularProperty(PropertyManager.withUri(PropertyManager.PROPMGR_DEVICE_ID));
```

what will return doubled collect: 
`collect:collect:5sf45sf5dsf5d4f
`

maybe we should get rid of that prefix collect: which now is a part of device id and add it instead of SCHEME_IMEI so as a result in settings and in a form we would have `5sf45sf5dsf5d4f` but sending forms it would be `collect:5sf45sf5dsf5d4f`?

@lognaturel @seadowg

lognaturel: Interesting, I didn't realize there was that difference between in-form and at submission time. I think they should be consistent and both include the scheme. How about just using the literal String `""` as the scheme here?

grzesiek2010: That would be a solution, what do you think @seadowg?

seadowg: Ah yeah I hadn't realized `PropertyManager` expects a scheme here. I think passing `""` as the scheme makes the most sense to me as because then (as @lognaturel points out) it'll be displayed consistently.

We could then just add a `putProperty(String propName, String value)` that doesn't append a scheme (or just calls the other `putProperty` with `""`).

grzesiek2010: Ok, I removed that scheme but I needed to add some extra changes in `InstanceServerUploaderTask` and `InstanceSubmitter` where we access that property to avoid using PropertyManager.withUri because then even if the scheme is empty we would have an extra semicolon left, like: `:collect:5sf45sf5dsf5d4f `.
All the changes are in my last commit:
https://github.com/getodk/collect/pull/4191/commits/94e954c9456be9b5654fc759d80c31bac9da5c28

lognaturel (approved):

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- relaunch app
- light/dark mode
- reset settings
- import settings qr code
- display deviceID in form
- save/finalize form
- rotate screen
- minimize app
- RTL language 
- fill few submissions
- install app again: deviceID updated
- regression check on username, email, phone number
- update
- kill app
- form metadata through admin settings
- form that displays old data: empty values visible

mmarciniak90:
Verified also on Android versions: 7.0, 8.1, 10.0

PR #4196, opened 2020-11-02
Force storage migration

### LINKED ISSUE #3360: Update targetSdkversion to 29

grzesiek2010:
Closes #3360

I tested the changes manually and added automated tests.

This is what we need to do because of google's policy.

- fresh installs should start with scoped storage by default
- for other users migrations should happen automatically once they install and open the app for the first time
- if the automated migration fails the banner as in v1.28.x should be visible and allow users to perform migration manually
- if migration fails and users don't trigger it manually next attempt should take place automatically when such users open the app again after 24h or if the app was killed and reopened

We reuse the same code which performs migration so migration itself shouldn't be affected. During testing we don't need to test the process a lot. 

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -200,6 +200,9 @@ android {
         unitTests {
             includeAndroidResources = true
             returnDefaultValues = true
+            all {
+                maxHeapSize = "1024m"
+            }
         }
     }
 
--- a/strings/src/main/res/values/strings.xml
+++ b/strings/src/main/res/values/strings.xml
@@ -670,7 +670,7 @@
     <string name="save_legacy_settings_warning">Will be removed in a future version. Please use the QR code above.</string>
     <string name="reason_for_changes_description">You need to explain the reason for changes to this form</string>
 
-    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. In Collect v1.29, <b>the migration will be forced on install</b>.</string>
+    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. The migration should have happened automatically. If you still see this message it indicates that something went wrong. Please try to perform migration manually.</string>
     <string name="scoped_storage_learn_more">Learn more and migrate</string>
     <string name="storage_migration_completed">Storage migration completed!</string>
     <string name="scoped_storage_dismiss">Dismiss</string>
--- a/strings/src/main/res/values/strings.xml
+++ b/strings/src/main/res/values/strings.xml
@@ -670,7 +670,7 @@
     <string name="save_legacy_settings_warning">Will be removed in a future version. Please use the QR code above.</string>
     <string name="reason_for_changes_description">You need to explain the reason for changes to this form</string>
 
-    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. The migration should have happened automatically. If you still see this message it indicates that something went wrong. Please try to perform migration manually.</string>
+    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. An automatic migration attempt failed. Please migrate manually.</string>
     <string name="scoped_storage_learn_more">Learn more and migrate</string>
     <string name="storage_migration_completed">Storage migration completed!</string>
     <string name="scoped_storage_dismiss">Dismiss</string>
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ task checkInstrumented(type: GradleBuild) {
 ext {
     androidCompileSdk = 30
     androidMinSdk = 21
-    androidTargetSdk = 28
+    androidTargetSdk = 29
 
     fragmentVersion = '1.2.5'
     daggerVersion = '2.29.1'
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalDataFileNotFoundTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalDataFileNotFoundTest.java
@@ -10,6 +10,8 @@ import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.odk.collect.android.R;
 import org.odk.collect.android.activities.FormEntryActivity;
+import org.odk.collect.android.storage.StoragePathProvider;
+import org.odk.collect.android.storage.StorageSubdirectory;
 import org.odk.collect.android.support.pages.FormEntryPage;
 import org.odk.collect.android.support.CopyFormRule;
 import org.odk.collect.android.support.ResetStateRule;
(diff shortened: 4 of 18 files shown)


grzesiek2010:
@lognaturel @seadowg I would like to hear from you what you think about forcing storage migration.
I start it from the Main menu because it's easy I can reuse the code we already have. In order to avoid multiple attempts if something goes wrong I added a check to confirm it happens at most once a day (of course if app is killed it might happen more often but thats ok).
Additionally for now I didn't remove the banner because maybe it's better to have it and still give a user a possibility to start migrating at their convenience when the automated one attempt failed.

any thoughts?

seadowg:
I had been thinking we would just run the migration rather than popping up the UI. I think if that was the case we could just start the service in `ApplicationInitializer`.

That said I can see why you would want to make the user acknowledge that it's happening, and also give them a way change to perform a backup. In this case I think starting from `MainMenuActivity` makes sense! I'm not sure about limiting the number of times we show it though... Maybe we could fire this up on `SplashScreenActivity` and then we'd show it every time the app is "launched"?

grzesiek2010:
@lognaturel @seadowg 
please take a look at this pr. I got stuck with one test `when_migrationFinishedWithSuccess_should_formsWorkAsBefore` which keeps failing on firebase test lab but locally everything is fine and the behavior which that test checks also seems fine (I tested it manually). Do you have any idea how my changes could affect that particular test?

seadowg:
@grzesiek2010 is the failure you're seeing this: https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/5654843488442284014/executions/bs.266c24ea218112f0/testcases/5/errors

seadowg:
@grzesiek2010 playing around with it a bit it doesn't feel like a timing issue as I can put a bunch of waits in, add in pressing back and then returning to Edit Saved Forms and it never loads the form. 

I'm now looking at running it as part of a suite (instead of just running the test itself or the class) and it looks like it always fails (for the same reason) for me if I run all tests in `org.odk.collect.android.feature.storage` which would suggest it's a pollution issue. Potentially the way the dependencies are set up or some static state? I'm going to play around with a bisect as I'm in this now 🤣.

seadowg:
@grzesiek2010 ok `StorageMigrationTest` and `AutomaticStorageMigrationTest` appear to be interfering with each other. If `AutomaticStorageMigrationTest#when_storageMigrationNotPerformed_shouldBePerformedAutomatically` runs before `StorageMigrationTest#when_migrationFinishedWithSuccess_should_formsWorkAsBefore` the latter fails and if I reverse the order the opposite is true. It does look automatic migration involves some static state so it could be something to do with that. 

I'll pause my investigation now in case you're already going down the same path but happy to keep looking if it's proving problematic!

grzesiek2010:
@seadowg 
Thanks for looking at the issue. When it comes to `AutomaticStorageMigrationTest` even if I remove this tests the result is the same so it's not the reason but probably your thinking that other tests somehow impact the one which is failing is right. I'll try to figure it out.

grzesiek2010:
oh I got it https://github.com/getodk/collect/pull/4196/commits/084a5786a0d6eb474d70c4f3a3f680bcad1a3ca8 obviously my brain needed those two days off which I had :) @seadowg your insights were also helpful thanks!

collect_app/build.gradle
```diff
@@ -200,6 +200,9 @@ android {
         unitTests {
             includeAndroidResources = true
             returnDefaultValues = true
+            all {
+                maxHeapSize = "1024m"
```
grzesiek2010, line 204: Without this my build kept failing it's probably connected somehow to bumping targetSdkVersion https://circleci.com/docs/2.0/language-android/#handling-out-of-memory-errors

strings/src/main/res/values/strings.xml
```diff
@@ -670,7 +670,7 @@
     <string name="save_legacy_settings_warning">Will be removed in a future version. Please use the QR code above.</string>
     <string name="reason_for_changes_description">You need to explain the reason for changes to this form</string>
 
-    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. In Collect v1.29, <b>the migration will be forced on install</b>.</string>
+    <string name="scoped_storage_banner_text">You must migrate Collect forms to private storage to comply with Android requirements. The migration should have happened automatically. If you still see this message it indicates that something went wrong. Please try to perform migration manually.</string>
```
lognaturel: "The migration should have happened automatically. If you still see this message it indicates that something went wrong. Please try to perform migration manually." -> "An automatic migration attempt failed. Please migrate manually."

lognaturel (changes requested):
Small suggested wording change to keep things short but looks great to me!

strings/src/main/res/values/strings.xml
grzesiek2010: Done. Do you also want to add something to the message that is displayed when migration is performed without errors? Now it's just `Storage migration completed!` but I remember that you were thinking about it, that it might be unclear for users that don't perform it manually.

lognaturel: I think it's ok. Maybe a toast would be less disruptive since it's not really all that important for users to see. Let's get to beta and ask for feedback.

lognaturel (approved):

kkrawczyk123:
I have verified on Androids 5.1, 7.1 and 9.0 cases, I only have some doubts about the scenario with admin password:
- app update with:
- a lot of forms
- 1000 submissions
- offline layers
- splash image
- media of form with external itemsets
- form with different kind of media files
- form with last-save
- admin password: Here, I can see the difference here with manual migration. Automatic migration ignores admin password and manual migration won't run without input admin password. @grzesiek2010  @lognaturel  please confirm that it's expected behavior.
- offline install
- edited db files to crash automatic migration
- manual migration after failed automatic

grzesiek2010:
> Automatic migration ignores admin password and manual migration won't run without input admin password. @grzesiek2010 @lognaturel please confirm that it's expected behavior.

we didn't discuss it but I think it's ok there was time when we asked politely but now we have less and less time. I think it would be risky to block storage migration on some devices used by enumerators that do not own them and have the password set. Maybe I would even remove that check from manual migration just in case to make sure as many users as possible will do that.

mmarciniak90:
PR verified also on Android versions: 7.0, 8.1, 10.0

I haven't noticed any issues other than the one mentioned by @kkrawczyk123.
Grzegorz's argument is accurate to me, that's why I accept this PR. 
@lognaturel, if you decides otherwise, I'll create a separate problem.

lognaturel:
> we asked politely but

I agree, we need to get folks migrated at this point. Projects should have had the time to adjust if needed.

> remove that check from manual migration just in case

I think it's ok to keep as-is. The automatic migration will be attempted again and if it fails it may be a good opportunity for a data collector who doesn't control their device to check in with their supervisor.

PR #4197, opened 2020-11-02
Show warning when importing legacy settings

### LINKED ISSUE #4111: Show warning when legacy settings import used
Collect currently supports importing files from a `.settings` or `.json` file when the app loads. It'd be good for this to eventually go away, so we don't have to keep maintaining the code paths.

**Acceptance**

- [x] Given I've set up a settings `.settings` file
When I open Collect
I see a dialog warning me that `.settings ` support will be removed

- [x] Given I've set up a settings `.json` file
When I open Collect
I see a dialog warning me that `.json` support will be removed

Text for the dialog will be: 

> Settings were successfully imported from a <file type> file. This functionality will be removed in a future version. Please use QR code configuration instead.

seadowg:
Closes  #4111

New test to check flow.

This could probably be improved a bit but I didn't want to spend much time on this given we want to remove it soon!

Shouldn't change anything. Good to check the `.settings` and `.settings.json` flows.


collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -574,31 +582,4 @@ private void displayBannerWithSuccessStorageMigrationResult() {
             storageMigrationRepository.clearResult();
         });
     }
-
-    private void importSettingsFromLegacyFiles() {
-        try {
-            String settings = new LegacySettingsFileReader(storagePathProvider).toJSON();
-
-            if (settings != null) {
-                String type = new File(storagePathProvider.getStorageRootDirPath() + "/collect.settings.json").exists()
```
seadowg, line 583: This actually wouldn't ever have worked as `LegacySettingsFileReader` deleted the files. I've moved that out to a separate action so this should now fire correctly.

lognaturel (approved):

mmarciniak90:
> This could probably be improved a bit but I didn't want to spend much time on this given we want to remove it soon!

@seadowg I've noticed that the light/dark mode is not applied immediately after reopening the app but I'm not sure that I can argue when you have such a good argument as above :rofl: 

For both versions, I copied this file [collect.settings.json.txt](https://github.com/getodk/collect/files/5502494/collect.settings.json.txt)

|released v1.28.3 | PR|
|---|---|
|v283|pr|

seadowg:
@mmarciniak90 huh. That's weird. I don't think it's a big deal but I want to look into it to check nothing bad is happening.

seadowg:
@mmarciniak90 realized the reason is that the reload that changes the theme doesn't happen until you click "OK" on the dialog. We could rework it so that the reload happens and then the dialog shows but it'd be more work and (as you pointed out) probably not worth it!

If that's the only problem are we done with testing for this?

mmarciniak90:
I noticed that @lognaturel suggested a different title for the dialog  `Title - Successful settings import` but I think that `Settings imported successfully` is also acceptable.

**Tested with success**

Verified cases:
- json file
- .settings file with admin password
- .settings file without admin password
- .settings file which changes theme
- import before and after migration
- imported QR code from file
- scanned QR code

kkrawczyk123:

PR #4199, opened 2020-11-03
Fixed launchIntentButton's style

### LINKED ISSUE #4193: Button for launching an external app to get multiple values is unstyled
Collect v1.18 and earlier

1. Launch a form with an intent group (https://docs.google.com/spreadsheets/d/1bWBACjdC-ZdQgFYCHCrQmUoORIPjU64dfcqiLm_c1fg/edit#gid=0)
1. See that the Launch button is hard to see because it doesn't look like a button

#### Expected behavior
For now we should probably use the same button style as we use everywhere else.

grzesiek2010:
Closes #4193

I tested the fix on real devices.

The previous styling was bad so we decided to fix it. In #4193 I mentioned two possible solutions and the second one seems like a better option.

It's safe issue that shouldn't affect anything. Maybe we can merge this pr even without testing first and then test at the same time with https://github.com/getodk/collect/pull/4188?

[external_widgets.xlsx](https://github.com/getodk/collect/files/5481297/external_widgets.xlsx)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
@@ -29,15 +29,15 @@ import android.util.TypedValue;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.View.OnLongClickListener;
-import android.widget.Button;
 import android.widget.FrameLayout;
 import android.widget.LinearLayout;
-import android.widget.TableLayout;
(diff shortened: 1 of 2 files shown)


lognaturel (approved):
The fontsize + 2 is a little strange and maybe something we should reconsider but let's treat that separately. At least this makes the button look clickable! I agree with @grzesiek2010's recommendation to merge and QA with #4188.

PR #4200, opened 2020-11-03
Make clicking recording notification return to app

### LINKED ISSUE #4151: Clicking notification while recording returns to app

seadowg:
Closes #4151 
~~Blocked by #4190~~

Drove out with tests after spiking out a working solution.

We could use Activity flags to deal with this but that seems less versatile as it means we need to know which Activity the recording is taking place in (from the user's perspective). Given that's already change once in the last couple of weeks I think going with this is the best idea.

I actually wasn't sure this would work, so I'm pretty happy it ended up being this simple.

Good to try and work out a way to break this. I couldn't!

### DIFF
--- a/audiorecorder/src/main/AndroidManifest.xml
+++ b/audiorecorder/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
 
     <application>
         <service android:name=".recording.internal.AudioRecorderService" />
+        <activity android:name=".recording.internal.ReturnToAppActivity" />
     </application>
 
 </manifest>
--- a/audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/AudioRecorderService.kt
+++ b/audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/internal/AudioRecorderService.kt
@@ -8,6 +8,7 @@ import android.content.Intent
 import android.os.Build
 import android.os.IBinder
 import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationCompat.PRIORITY_LOW
 import org.odk.collect.audiorecorder.R
 import org.odk.collect.audiorecorder.getComponent
 import org.odk.collect.audiorecorder.recorder.Recorder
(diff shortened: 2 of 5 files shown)


lognaturel (approved):
Note to @getodk/testers: this should also remove all sounds from notifications on all Android levels.

kkrawczyk123:
@seadowg I have noticed a difference in behavior when using an external app for audio recording. On Android 9.0 I am able to record in the background and return to app using notification but on Androids 5.1 and 7.1 when I minimize app recording is interrupted and notification disappears. Don't keep activities is turned off. This difference is not visible when using an internal recorder. What is the expected behavior here?

seadowg:
@kkrawczyk123 the behaviour while recording will be completely dictated by the external app here so it's not something we control. What app are you using out of interest? It'd be good to know if one of the popular apps is unable to record in the background.

kkrawczyk123:
**Tested with success!**
@seadowg you are right I installed another recording app and I was able to get back to the app using notification. 

Tested scenarios:
- clearing notifications
- displaying auto-send and recording notifications
- sound is turned off
- ignored changes in audio widget in saved form
- switching between apps
- rotation
- minimization
- RTL language
- light / dark mode
- use external app
- click on auto-send notification when recording is in progress

mmarciniak90:
Tested also on Android versions 7.0, 8.1, 10.0

seadowg:
@kkrawczyk123 ah cool. It'd still be interesting to know which app you were seeing the problem with? We advise people around which ones to use in our docs so good to know if one of them has a problem!

kkrawczyk123:
@seadowg It was Sony Audio Recorder (installed a long time ago on my devices) but when I uninstall it was not able to download it again from the play store and it was only on devices when there is no default recording app. Probably, there is a low chance that some of the users actually use it.

seadowg:
A lot of people are as it was probably has the nicest UI out of all the options! Makes sense it has some problems though as it hasn't been updated for a long time.

PR #4206, opened 2020-11-05
Fixed tests

grzesiek2010:
I ran tests locally.

To be honest I'm not sure what exactly was wrong. There was something wrong with doThrow depending on the order of running tests so I fixed using it thenThrow().

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
@@ -58,8 +58,8 @@ import static androidx.test.espresso.matcher.ViewMatchers.withText;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.core.StringEndsWith.endsWith;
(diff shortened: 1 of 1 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -181,7 +181,7 @@ public void collect_shouldNotCrashWhenAnyExceptionIsThrownWhileReceivingAnswer()
         Intent resultIntent = new Intent();
 
         Uri uri = mock(Uri.class);
-        doThrow(new SecurityException()).when(uri);
```
seadowg, line 184: Yeah it looks like this setup is wrong. `when` is just being passed the `Uri` object. I think it should have been:

```
doThrow(new SecurityException()).when(uri).getScheme();
```

Does that seem right? If so the problem makes sense.

lognaturel: Ohh, I see. That makes sense. Seems we should generally go for `when` first anyway, so the fix is good.

seadowg (approved):

PR #4208, opened 2020-11-05
Try using the package manager  to get a launch intent for external apps, support string uris for binary external widgets

### LINKED ISSUE #4194: External app launch features only works for apps that specify the DEFAULT category in their intent filter
Collect v1.28 and previous

#### Problem description
External app launch features only works for apps that specify the DEFAULT category in their LAUNCHER intent filter.

Try to launch one of the apps mentioned in https://forum.getodk.org/t/launch-a-scanner-app-and-attach-a-pdf-from-collect/19061 from a form. For example, for https://printerplus.rcti.es/:

type | name | label | appearance
-- | -- | -- | --
integer | test | Test | ex:es.rcti.printerplus()

#### Expected behavior
The application should launch.

#### Other information 
https://stackoverflow.com/questions/54492708/why-is-no-activity-found-to-handle-intent/54552362#54552362

https://stackoverflow.com/questions/62535856/intent-resolveactivity-returns-null-in-api-30/62856745#62856745

lognaturel:
Closes #4194

Tried launching external applications that were previously failing with both a single question and in an intent field list.

I don't fully understand the intent rules. From my reading, it seems that trying both `new Intent(intentName)` and `getLaunchIntentForPackage(intentName)` gives us the biggest chance of success. I couldn't find any other good alternatives.

I also broadened the allowable types for binary widgets. I realized that in our initial spec we weren't explicit about types and so users might be providing the URIs as strings. The change here is the simplest way I can find to support both.

Since these are additive, the regression risk is low. Only binary widgets and binary questions in intent field-lists are at any risk.

See forum post linked from issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
@@ -509,7 +509,17 @@ public class ODKView extends FrameLayout implements OnLongClickListener, WidgetV
                                 break;
                             case Constants.DATATYPE_BINARY:
                                 try {
-                                    Uri uri = (Uri) bundle.get(key);
+                                    Object uriValue = bundle.get(key);
+
+                                    Uri uri;
+                                    if (uriValue instanceof Uri) {
+                                        uri = (Uri) uriValue;
+                                    } else if (uriValue instanceof String) {
+                                        uri = Uri.parse(bundle.getString(key));
+                                    } else {
+                                        throw new RuntimeException("The value for " + key + " must be a URI but it is " + uriValue);
+                                    }
+                                    
                                     if (uri != null) {
                                         File destFile = FileUtils.createDestinationMediaFile(formController.getInstanceFile().getParent(), ContentResolverHelper.getFileExtensionFromUri(getContext(), uri));
                                         //TODO might be better to use QuestionMediaManager in the future
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
@@ -405,48 +405,49 @@ public class ODKView extends FrameLayout implements OnLongClickListener, WidgetV
         launchIntentButton.setText(buttonText);
         launchIntentButton.setTextSize(TypedValue.COMPLEX_UNIT_DIP, QuestionFontSizeUtils.getQuestionFontSize() + 2);
         launchIntentButton.setVisibility(VISIBLE);
-        launchIntentButton.setOnClickListener(new OnClickListener() {
-            @Override
-            public void onClick(View v) {
-                String intentName = ExternalAppsUtils.extractIntentName(intentString);
-                Map<String, String> parameters = ExternalAppsUtils.extractParameters(
-                        intentString);
+        launchIntentButton.setOnClickListener(view -> {
+            String intentName = ExternalAppsUtils.extractIntentName(intentString);
+            Map<String, String> parameters = ExternalAppsUtils.extractParameters(intentString);
 
-                Intent i = new Intent(intentName);
-                try {
-                    ExternalAppsUtils.populateParameters(i, parameters,
-                            c.getIndex().getReference());
+            Intent i = new Intent(intentName);
+            if (i.resolveActivity(Collect.getInstance().getPackageManager()) == null) {
+                i = Collect.getInstance().getPackageManager().getLaunchIntentForPackage(intentName);
+                i.setFlags(0);
+            }
 
-                    for (FormEntryPrompt p : questionPrompts) {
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -405,48 +405,49 @@ private void addIntentLaunchButton(Context context, FormEntryPrompt[] questionPr
...
-                                    break;
-                            }
+        launchIntentButton.setOnClickListener(view -> {
+            String intentName = ExternalAppsUtils.extractIntentName(intentString);
+            Map<String, String> parameters = ExternalAppsUtils.extractParameters(intentString);
+
+            Intent i = new Intent(intentName);
+            if (i.resolveActivity(Collect.getInstance().getPackageManager()) == null) {
+                i = Collect.getInstance().getPackageManager().getLaunchIntentForPackage(intentName);
+                i.setFlags(0);
```
lognaturel: I can't remember the details of why this is needed but can try to pull it up again if needed. There are flags set by default when doing `getLaunchIntentForPackage` and they suppress something, maybe returning a result.

grzesiek2010: I tested the fix without the flag (not here but using external widgets) and it seems to work so would be good if you could double check it.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -405,48 +405,49 @@ private void addIntentLaunchButton(Context context, FormEntryPrompt[] questionPr
...
-                                            (Serializable) value);
-                                    break;
-                            }
+        launchIntentButton.setOnClickListener(view -> {
+            String intentName = ExternalAppsUtils.extractIntentName(intentString);
+            Map<String, String> parameters = ExternalAppsUtils.extractParameters(intentString);
+
+            Intent i = new Intent(intentName);
+            if (i.resolveActivity(Collect.getInstance().getPackageManager()) == null) {
+                i = Collect.getInstance().getPackageManager().getLaunchIntentForPackage(intentName);
```
grzesiek2010: getLaunchIntentForPackage() might return null so wee need some null check now I think.

lognaturel: YES, thank you.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
lognaturel: Now I remember that at some point I got into a state where the OpenRDT app was immediately returning its result rather than first going through its UI flow and then returning. That led me to [this Stackoverflow post](https://stackoverflow.com/a/24312398/137744). The OpenRDT examples I was sent (not public, unfortunately) do use `setFlags(0)`.

>  getLaunchIntentForPackage("com.your.package.here") would create an intent with Intent.FLAG_ACTIVITY_NEW_TASK added by default, so call: in.setFlags(0);

Always creating the target activity and avoiding the `FLAG_ACTIVITY_NEW_TASK` flag seems like the way to go to me. There are differences in Android versions with how the other launch modes work and their behavior will also depend on whether or not an activity is already in memory.

lognaturel: Is it just me or are the docs around this particularly bad?

lognaturel:
Thanks for the review, @grzesiek2010, and especially for catching the null issue. I verified that trying to launch an app not on the device now gives the expected toast instead of crashing.

I tried to explain the flags both in https://github.com/getodk/collect/pull/4208#discussion_r520016609 and in a code comment. Let me know if you still have questions.

lognaturel:
Another thing. It would be good to have tests for things like if an implicit intent can't be resolved, the explicit intent path is attempted or if nothing can be resolved, the toast is shown. I'm hoping that when you come back to this to add single-question external binary functionality you'll be able to consolidate some of the code and add some testing, @grzesiek2010. It feels like things are still going to move with that change so I didn't want to spend too much time on it.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
grzesiek2010: Ok this is confusing but makes more sense now thanks!

grzesiek2010 (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- launch HP Smart app
- launch Printer+ app
- launch Tacho Center app
- jump from hierarchy
- open saved form
- rotation
- minimization
- RTL language
- light / dark mode

kkrawczyk123:
I was not able to install HP Smart app as it supports Androids 7+.

PR #4211, opened 2020-11-06
Revert "Revert "Reworking Rating Widget""

SaumiaSinghal:
Reverts getodk/collect#4210

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FieldListUpdateTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FieldListUpdateTest.java
@@ -360,7 +360,7 @@ public class FieldListUpdateTest {
         onView(withText(startsWith("Source13"))).perform(click());
 
         onView(withText("Target13")).check(doesNotExist());
-        onView(allOf(withId(R.id.rating_bar1), isDisplayed())).perform(setRating((3.0f)));
+        onView(allOf(withId(R.id.rating_bar1), isDisplayed())).perform(setRating(3.0f));
         onView(withText("Target13")).check(matches(isDisplayed()));
 
         onView(withText("Source13")).perform(longClick());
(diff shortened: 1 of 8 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FieldListUpdateTest.java
```diff
@@ -354,9 +359,13 @@ public void selectingARating_ShouldChangeRelevanceOfRelatedField() throws Except
         jumpToGroupWithText("Rating");
         onView(withText(startsWith("Source13"))).perform(click());
 
-        for (int i = 0; i < 10; i++) {
+        for (int i = 1; i <= 10; i++) {
```
seadowg: Definitely from the previous version but I'm not sure why this test loops. The goal of the test seems to be to check that setting a rating updates the relevance of another question. In my head the test could just do this once: check the "Target13" question doesn't exist, set a (any) rating, check it does, remove the rating, check "Target13". @lognaturel can you think of any reason we might be checking every value here?

SaumiaSinghal: Yes, I too think you are right here @seadowg. Checking just once whether changing the rating makes changes to another question should suffice.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FieldListUpdateTest.java
```diff
@@ -359,19 +359,13 @@ public void selectingARating_ShouldChangeRelevanceOfRelatedField() throws Except
...
+        onView(allOf(withId(R.id.rating_bar1), isDisplayed())).perform(setRating((3.0f)));
+        onView(withText("Target13")).check(matches(isDisplayed()));
 
-            onView(withText("Source13")).perform(longClick());
-            onView(withText(R.string.clear_answer)).perform(click());
-            onView(withText(R.string.discard_answer)).perform(click());
-        }
+        onView(withText("Source13")).perform(longClick());
+        onView(withText(R.string.clear_answer)).perform(click());
+        onView(withText(R.string.discard_answer)).perform(click());
```
seadowg, line 368: Probably good to check the view disappears once the answer is removed as well.

seadowg (changes requested):

seadowg (approved):

Issue #4072, opened 2020-09-07
Add context to existing strings, write up requirements for new strings


lognaturel:
It can be hard for translators to make sense of strings that aren't complete sentences. Sometimes the part of speech is unclear (noun vs verb) and other times the translation can depend a lot on context (dialog vs button, etc).

Some suggestions for strings to add details to: https://forum.getodk.org/t/adding-context-to-particular-strings/13540

We should always add context for new short strings and document how that should be done in the contribution guide.

PR #4215, opened 2020-11-10
Stop user leaving recording

### LINKED ISSUE #4154: Enumerator is asked to stop recording before changing screen

seadowg:
Closes #4154 
~~Blocked by #4190~~

The user will now be instructed to stop recording before they can leave form entry. I also reworked some of the logic that deals with switching between screens in `FormEntryActivity` to make it easier to deal with checks that happen when we swipe left or right.

New tests and manually verified.

Comment inline.

Definitely good to check if there are other ways to "leave" form entry that haven't been covered here! Also good to check that form entry generally (particularly swiping through forms) is all still ok.


.circleci/config.yml
```diff
@@ -63,6 +63,9 @@ jobs:
       - run:
           name: Run audioclips unit tests
           command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 audioclips:testDebug
+      - run:
```
seadowg, line 66: Forgot this earlier

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
```diff
@@ -82,40 +94,43 @@ public void onPrepareOptionsMenu(Menu menu) {
...
     public boolean onOptionsItemSelected(MenuItem item) {
-        switch (item.getItemId()) {
-            case R.id.menu_add_repeat:
-                getFormSaveViewModel().saveAnswersForScreen(answersProvider.getAnswers());
-                getFormEntryViewModel().promptForNewRepeat();
-                formIndexAnimationHandler.handle(getFormEntryViewModel().getCurrentIndex());
-                return true;
-
-            case R.id.menu_preferences:
+        if (item.getItemId() == R.id.menu_add_repeat) {
```
seadowg, line 102: Changing from a `switch` to an `if` is now suggested by Android Studio.

strings/src/main/res/values/strings.xml
```diff
@@ -702,10 +702,22 @@
...
     <string name="failure">Failure</string>
     <string name="form_update_succeeded">Form update succeeded</string>
+
+    <!-- Title for setting to let the user switch between using Collect or another app (on the device) for recording audio -->
     <string name="external_app_recording">Use external app for audio recording</string>
+
+    <!-- Text for button that stops audio recording  -->
     <string name="stop_recording">Stop</string>
+
+    <!-- Text to let the user now a recording is in progress  -->
```
lognaturel: now -> know

strings/src/main/res/values/strings.xml
```diff
@@ -702,10 +702,22 @@
...
     <string name="recording">Recording…</string>
+
+    <!-- Title of Android Notification Channel for audio recording notifications  -->
     <string name="recording_channel">Audio recording</string>
-    <string name="play_audio">Play audio</string>
+
+    <!-- Text for button that plays audio clip  -->
+    <string name="play_audio">Play chosen audio</string>
```
lognaturel: Isn't this sometimes chosen and sometimes recorded?

strings/src/main/res/values/strings.xml
```diff
@@ -720,4 +720,7 @@
 
     <!-- Text to let the user know a file is being saved  -->
     <string name="saving_file">Saving file</string>
+
+    <!-- Text in dialog that instructs the user to stop recording audio before leaving the form -->
+    <string name="recording_warning">Must stop recording before moving on!</string>
```
lognaturel: How about "You must stop recording before leaving this screen."

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1476,99 +1467,77 @@ public boolean dispatchTouchEvent(MotionEvent mv) {
...
+            swipeHandler.setBeenSwiped(false);
+            return;
+        }
 
-                            if (event != EVENT_PROMPT_NEW_REPEAT) {
-                                // Returning here prevents the same view sliding when user is on the first screen
-                                return;
-                            }
-                        }
+        state = null; // This is needed for something but is definitely suspect
```
lognaturel: Looks like it's used to keep track of which questions' guidance hints are expanded and the search text for selects with the search appearance. Would be great to move all that to a view model, I think.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1476,99 +1467,77 @@ public boolean dispatchTouchEvent(MotionEvent mv) {
...
+            DialogUtils.showIfNotShowing(RecordingWarningDialogFragment.class, getSupportFragmentManager());
+            swipeHandler.setBeenSwiped(false);
+            return;
+        }
 
-                        // If we are the beginning of the form, there is no previous view to show
-                        if (event == FormEntryController.EVENT_BEGINNING_OF_FORM) {
-                            event = formController.stepToNextScreenEvent();
-                            swipeHandler.setBeenSwiped(false);
+        if (direction == FORWARDS && formController.getEvent() == FormEntryController.EVENT_END_OF_FORM) {
```
lognaturel: Why pull this direction-specific stuff out? Seems it should either be in the switch below or in the direction specific methods. Or if it's pulled out and their bodies are the same maybe it should be a single conditional.

If pulling it out, why not also pull the analogous beginning of form test out here? (Not that I think it's actually possible to be at the beginning of form event)

lognaturel (changes requested):
I'm a bit puzzled by some of the restructuring in the swiping code and want to make sure we limit risk there. Maybe a quick talk through would help!

The biggest thing is I've proposed alternate text for the dialog.

strings/src/main/res/values/strings.xml
seadowg: Yup good catch.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: I initially pulled all these out so the method had more of a "guard clauses followed by the happy logic" structure as it was really cluttered. That helped me work out what I could move out to the view models. I'll try putting them back into the `switch` now and see how it feels though.

seadowg: Ah it's good to keep them where they are so `state = null` happens in the same place. Initially that had to be in both parts of the `switch`. It would be cool to pair on this now you've got some context in your head. I'll ping you on Slack.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1506,7 +1508,11 @@ private void moveScreen(Direction direction) {
 
                 formEntryViewModel.moveBackward();
                 formIndexAnimationHandler.handle(formController.getFormIndex());
-                break;
+                return true;
+
+            default:
+                // Java switch statements aren't smart enough to know there are only two possibilities
```
lognaturel: Maybe a reason to go with if/else?

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -128,6 +134,44 @@ public boolean canAddRepeat() {
...
+        } catch (JavaRosaException e) {
+            error.setValue(e.getCause().getMessage());
+            return;
+        }
+
+        formController.getAuditEventLogger().flush(); // Close events waiting for an end time
+    }
+
+    public void moveBackward() {
+        if (formController.getEvent() != FormEntryController.EVENT_BEGINNING_OF_FORM) {
```
lognaturel: Why doesn't this also move to `moveScreen`?

strings/src/main/res/values/strings.xml
```diff
@@ -525,9 +525,9 @@
     <string name="invalid_email_address">Invalid email address!</string>
     <string name="sort_by">Sort by</string>
     <string name="server">Server</string>
-<!--
-    <string name="submission_transport_admin_setting">Submission transport</string>
--->
+    <!--
```
lognaturel, line 528: This was an oversight when removing the SMS feature. Would be great to remove it entirely at some point.

lognaturel (approved):
Great to get rid of that confusing `state` field. The view model approach looks good.

I feel like the check for the beginning of the form can be moved to be analogous to the check for the end of the form but I might be wrong on that. Either way, it's not a big deal and would be good for this to go to QA ASAP.

strings/src/main/res/values/strings.xml
lognaturel: I'll do it in https://github.com/getodk/collect/issues/4072

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Yeah, you're right. I just miss `when` so much. It feels like all recent signs (Android tools changes, me making mistakes) are pointing to avoid `switch` altogether. I'll change this.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Ah yeah I got distracted by the search/rotate problem and didn't clean this up. Will move it around today.

seadowg: Realized we don't actually need this check at all so avoiding have it anywhere.

lognaturel (approved):

lognaturel:
Some specific areas of risk for QA to consider:
- Swiping backwards from the first screen on forms that start with repeats or nested repeats
- Guidance hints. We expect that the collapsed state will be lost on rotation and have decided that's ok to reduce code complexity.
- Selects with search. The search query should be maintained on rotation, even if there are multiple such questions in a field list. There's a slight change in behavior: the search queries should be maintained when swiping to another question screen and then swiping back.

mmarciniak90:
@seadowg I've noticed one case which is not covered. Dialog is not visible when you click :heavy_plus_sign: to add a repeat group

[AllQ4190.xml.txt](https://github.com/getodk/collect/files/5566496/AllQ4190.xml.txt)

ezgif com-video-to-gif

seadowg:
@mmarciniak90 ah nice find! I'll fix this.

seadowg:
@mmarciniak90 fixed

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- light mode/dark mode
- jump to hierarchy
- add group
- go to General Settings
- swipe backward
- swipe forward
- back button
- use navigation buttons 
- use swipes and buttons
- try all those actions in portrait and landscape 
- tried swipes when moving backward disabled
- minimize app

mmarciniak90:
> Swiping backwards from the first screen on forms that start with repeats or nested repeats

I've used this form [AllQ4190_v2.xml.txt](https://github.com/getodk/collect/files/5568514/AllQ4190_v2.xml.txt) and I've noticed that **top of page is visible** when I swipe backward. Navigation buttons look correct.

> Guidance hints. We expect that the collapsed state will be lost on rotation and have decided that's ok to reduce code complexity.

I can confirm, this regression is visible. There is no difference in the others options

>Selects with search. The search query should be maintained on rotation, even if there are multiple such questions in a field list. There's a slight change in behavior: the search queries should be maintained when swiping to another question screen and then swiping back.

I like these changes

@seadowg I've noticed one more, crazy and probably unlikely case but it **blocks swiping**.
1. `Use swipes and buttons` is set on Naviation option in User Interface settings
2. Start form [AllQ4190_v2.xml.txt](https://github.com/getodk/collect/files/5568735/AllQ4190_v2.xml.txt)
3. Start recording 
4. Swipe -> dialog is visible, close it
5. Click NEXT -> dialog is visible, close it
6. Swipe -> nothing happens, looks like swiping has stopped working. Even if recording is stopped, swipe still doesn't work. User is forced to use NEXT button. After that, everything backs to normal.

ezgif com-video-to-gif(1)

seadowg:
@mmarciniak90 hmmm there is code that stops swipes from happening if it thinks one is in progress so there's probably something amiss there. I'll check that out and hopefully have a fix for you in the morning!

lognaturel:
> I've noticed that top of page is visible when I swipe backward

Ahhh, thank you, I didn't think to scroll down on the first question screen and then try to swipe back but from the code it seemed something could go wrong without a pre-emptive check for the start of form. @seadowg you can see this just with All Widgets on a device where the introductory text doesn't fit on the screen. Scroll down a bit, swipe back, and it will go to the top because the screen is re-rendered.

> crazy and probably unlikely

Seems worth digging into.

seadowg:
@mmarciniak90 those cases are hopefully fixed now!

mmarciniak90:
**Tested with success!**

Verified on Androids: 7.0, 8.1, 10.0

Verified cases
- cases mentioned by Kasia
- swipe and button bug -> user is not forced to use button
- swipe backward -> user is not moved to the top of the page
- guidance hints -> accepted regression is visible
- selects with search -> swipe, rotate

seadowg:
@mmarciniak90 ace! Is testing finished now?

kkrawczyk123:

PR #4220, opened 2020-11-13
Allow external recording using `quality` parameter

### LINKED ISSUE #4216: Can use external recorder app
Based on https://forum.getodk.org/t/form-spec-proposal-configure-audio-quality-for-internal-recording/31094

**Acceptance**

- [x] Given I'm using a form with an audio question with `quality` parameter as `foo`
When I click "Choose Sound"
Then it uses the in-app recording

- [x] Given I'm using a form with an audio question with `quality` parameter as `external`
When I click "Choose Sound"
Then it uses the default recorder app on my device

- [x] Given I've enabled "Use external recording app" in settings
When I record with an audio question
Then the recording is made using the default recorder app on my device

- [x] Given I've enabled "Use external recording app" in settings
When I record with an audio question that has `normal` or `voice-only` as `quality`
Then the recording is made using the in-app audio recorder

seadowg:
Closes #4216

New tests and manually verified.

The implementation is actually pretty straight forward for this. I've left a couple of comments inline.

I think we can skip QA for this until we have support for the other audio qualities in there.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
@@ -62,6 +62,7 @@ import org.odk.collect.android.formentry.media.PromptAutoplayer;
 import org.odk.collect.android.formentry.questions.QuestionTextSizeHelper;
 import org.odk.collect.android.javarosawrapper.FormController;
 import org.odk.collect.android.listeners.WidgetValueChangedListener;
+import org.odk.collect.android.preferences.GeneralKeys;
 import org.odk.collect.android.preferences.PreferencesProvider;
 import org.odk.collect.android.utilities.ActivityAvailability;
 import org.odk.collect.android.utilities.FileUtils;
@@ -284,7 +285,9 @@ public class ODKView extends FrameLayout implements OnLongClickListener, WidgetV
      */
     private QuestionWidget configureWidgetForQuestion(FormEntryPrompt question, boolean readOnlyOverride) {
         QuestionWidget qw = WidgetFactory.createWidgetFromPrompt(question, getContext(), readOnlyOverride,
-                waitingForDataRegistry, questionMediaManager, analytics, audioPlayer, activityAvailability, audioRecorderViewModelFactory);
+                waitingForDataRegistry, questionMediaManager, analytics, audioPlayer,
+                activityAvailability, audioRecorderViewModelFactory,
+                preferencesProvider.getGeneralSharedPreferences().getBoolean(GeneralKeys.KEY_EXTERNAL_APP_RECORDING, false));
         qw.setOnLongClickListener(this);
         qw.setValueChangedListener(this);
 
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
@@ -35,8 +35,10 @@ import android.widget.LinearLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import androidx.activity.ComponentActivity;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.lifecycle.ViewModelProvider;
 
 import com.google.android.material.button.MaterialButton;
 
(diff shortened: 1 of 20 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -124,35 +127,27 @@
     PreferencesProvider preferencesProvider;
 
     @Inject
-    AudioRecorderViewModelFactory audioRecorderViewModelFactory;
+    ActivityAvailability activityAvailability;
+
+    private final WidgetFactory widgetFactory;
```
seadowg, line 132: It felt like it made sense to rearrange this relationship so the `ODKView` owns a `WidgetFactory` object. In my mind this makes the dependency clearer.

collect_app/src/main/java/org/odk/collect/android/utilities/FormEntryPromptUtils.java
```diff
@@ -124,4 +126,13 @@ public static String markQuestionIfIsRequired(String questionText, boolean isReq
 
         return questionText;
     }
+
+    public static Optional<String> getAttributeValue(FormEntryPrompt prompt, String attributeName) {
```
lognaturel: We talked briefly about using optionals vs sticking to nulls. Are we diving into optionals?

collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
```diff
@@ -194,11 +193,8 @@ public static QuestionWidget createWidgetFromPrompt(FormEntryPrompt prompt,
             case Constants.CONTROL_AUDIO_CAPTURE:
                 RecordingRequester recordingRequester;
 
-                Optional<TreeElement> audioQuality = prompt.getBindAttributes().stream().filter(attr -> {
-                    return attr.getName().equals("audio-quality");
-                }).findAny();
-
-                if (audioQuality.isPresent() && audioQuality.get().getAttributeValue().equals("external")) {
+                Optional<String> audioQuality = FormEntryPromptUtils.getAttributeValue(prompt, "audio-quality");
+                if (audioQuality.isPresent() && audioQuality.get().equals("external")) {
```
lognaturel: This feels like a very `null` structure if you see what I mean. Not really much benefit to an `Optional` if you're not chaining, for example.

lognaturel (commented):
One nitpicky question/conversation point about `Optional` since this would be the first one in the Java codebase.

I also realize we didn't talk about the upgrade experience and whether users who are already happily using external applications should get thrown into the internal recorder. I think it would be better for folks who are updating to continue with the existing behavior and new users to use internal recording by default. That seems like a separate story.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RecordingRequesterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+        this.audioRecorderViewModel = audioRecorderViewModel;
+        this.permissionUtils = permissionUtils;
+        this.activity = activity;
+        this.lifecycle = lifecycle;
+    }
+
+    public RecordingRequester create(FormEntryPrompt prompt, boolean externalRecorderPreferred) {
+        Optional<String> audioQuality = FormEntryPromptUtils.getAttributeValue(prompt, "quality");
+
+        if (audioQuality.isPresent() && (audioQuality.get().equals("normal") || audioQuality.get().equals("voice-only"))) {
```
lognaturel: This feels like a very null structure if you see what I mean. Not really much benefit to an Optional if you're not chaining, for example.

collect_app/src/main/java/org/odk/collect/android/utilities/FormEntryPromptUtils.java
seadowg: Ah yeah good point. I'm in two minds. I think your point below is a good one. Java Optionals really don't add much as calling `get` is super easy. I think I'm happy to convert the `Optional` from the stream into a normal type and using `Nullable`.

seadowg:
> I think it would be better for folks who are updating to continue with the existing behavior and new users to use internal recording by default.

@lognaturel yeah I think I agree.  I'll create a story!

lognaturel (approved):

lognaturel:
New upgrade story at https://github.com/getodk/collect/issues/4223!

PR #4221, opened 2020-11-13
Don't always save media captured through Collect to the Media Store

### LINKED ISSUE #4070: Don't always save media captured through Collect to the Media Store
Placeholder issue for https://forum.getodk.org/t/remove-previously-taken-images-from-gallery-via-setting/26593 so it can be put in milestone.

grzesiek2010:
Closes #4070

I tested the app manually and ran tests.

We agreed that we don't need to add media files to media store. That required a bit more code to remove than I expected and probably I could introduce more refactoring but I had to draw the line somewhere.

Media files should not be added to media store what means they should not be visible in galleries. When it comes to testing it would be good to confirm that taking media files in media widgets (ImageWidget/VideoWidget/AudioWidget/ArbitraryFileWidget) work as before:
- we can add media files
- old media files are removed
- we can open media files clicking for example on images/videos/arbitrary files
- media files recorded using ODK Collect are not visible in gallery apps

Any form with media widgets.

No.


collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
```diff
@@ -91,49 +93,8 @@ public static final Uri getImageUriFromMediaProvider(String imageFile) {
         }
     }
 
-    public int deleteImageFileFromMediaProvider(String imageFile) {
```
lognaturel, line 94: Funny, I've looked at this code and all the related methods in `MediaUtils` a few times wondering why we were going through a content resolver instead of just accessing the files directly and it never clicked that it was part of all this business with the Media Store. I really wonder why it was done that way.

lognaturel (approved):
This is SO satisfying, thank you. Really easy commits to follow, too.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
grzesiek2010: Yeah it's difficult to tell what the reason was.

grzesiek2010:
The pr is rebased and scoped storage will be used by default.

mmarciniak90:
**Tested with success**

Verified cases:
- recent images
- .thumbnails directory
- all widgets with an image on the `All widgets` form
- filled form with created photos
- filled form with attached images

kkrawczyk123:
@grzesiek2010 @lognaturel The fix works well on my devices with Androids 5.1 and 9.0 but on my Motorola Moto G4 Play with Android 7.1 images created using camera are still stored by the device in DCIM/Camera folder and I am also able to attach it to other submissions uploading from the recent folder. The only change visible is that images don't show up in .thumbnails folder. Before this fix, it was stored in both Camera and .thumbnails on that device.

grzesiek2010:
If you want we can file a separate issue to investigate it but I doubt we can force external app not to make copies.

kkrawczyk123:

PR #4226, opened 2020-11-17
Trust root certificate that will root all certs issued by Let's Encrypt after Sept 2021

### LINKED ISSUE #4219: Trust Let's Encrypt's ISRG Root X1 root for Android <7.1.1
Let's Encrypt's root CA will be used to sign all intermediaries starting Jan 2021. There will be no alternative after Sept 2021.

https://forum.getodk.org/t/effect-of-lets-encrypt-root-certificate-authority-changes-on-odk-ecosystem/31099

As far as I know, we can easily add trust for the root for Android 6+: https://developer.android.com/training/articles/security-config#TrustingAdditionalCas

As discussed in the forum post above, we still need to figure out what to do for Android 5.0 and 5.1 but we can do that in a separate issue. We might be able to do something like https://community.letsencrypt.org/t/mobile-client-workarounds-for-isrg-issue/137807/3?u=lognaturel

https://valid-isrgrootx1.letsencrypt.org/ is a test URL

lognaturel:
Closes #4219

This PR affects Android 5.1-7.1. There are separate fixes for Android 5.1-6.0 and Android 7.0-7.1. I verified by configuring the server to https://valid-isrgrootx1.letsencrypt.org. Prior to the fix, when I did "Get Blank Form", I would get no results because there was a security error. After the fix, I get an error about not being able to reach the server (because it's not an ODK server).

Discussed with Let's Encrypt community and an OkHTTP maintainer at https://community.letsencrypt.org/t/mobile-client-workarounds-for-isrg-issue/137807. The OKHttp fix works for all Android versions but it is potentially inefficient if there are many system-level certs specified. It also doesn't work for webviews. That makes it still worthwhile to use the `network-security-config` for API 24/25.

The fix applied for API <= 23 is tested at https://github.com/square/okhttp/pull/6423/files. I don't see any reasonable test for us to write.

It should just allow users to seamlessly keep connecting to servers that use Let's Encrypt certs. This does touch how OkHTTP is initialized so there could be a risk to all network connections. However, it's mitigated by the fact that it only applies for API <= 23 and that issues should be really easy to spot (nothing would work).

No.

No.


audiorecorder/src/main/res/raw/isrgrootx1.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
seadowg: This and the `network_security_config` should be in `collect_app` right?

seadowg (changes requested):
This all seems right but there were some files that I think need moved.

audiorecorder/src/main/res/raw/isrgrootx1.pem
lognaturel: So many questions. How did I manage that. How did I not notice. Why did it work. Whut?? 😩

seadowg (approved):

kkrawczyk123:
@lognaturel I have verified url that you have attached to the issue and I have the same good result for all devices that I have tried. I have also done a quick regression check on other servers (Aggregate, Central, Kobo, Ona, Google Drive) and I have noticed regression when tried to get blank form from local aggregate but only on Androids 9.0 and 10. The issue seems not to be visible on lower Android versions, and also not visible form Android 9 and 10 on master. 

lognaturel:
> local aggregate

Is it HTTP?

lognaturel:
I think it's probably an HTTP issue. I would expect it to happen on API 24+, though. Could you please try the latest commit? I don't have an HTTP server easily accessible!

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- URL form issue
- accept error and refresh the list
- Aggregate
- local Aggregate
- anonymous access allowed/denied
- Central
- Google Drive
- Kobo
- Ona
- display form list
- download submissions
- send submissions

mmarciniak90:
Verified also on Android 7.0, 8.1, 10.0
The fix is visible on Android 7.0 - `Form list download failed` dialog is visible
I didn't notice regression on Android 8.1 and 10.0

lognaturel:
I'm so glad the HTTP issue was so simple! It's definitely not what I expected from the docs but 🤷🏻‍♀️. Thanks for catching that.

PR #4228, opened 2020-11-17
Fixed: Read-only binary questions have no spacing below the question text

### LINKED ISSUE #4209: Read-only binary questions have no spacing below the question text
Collect v1.28 and prior

#### Problem description
Binary widgets hide their buttons when they're read-only. Currently they leave no spacing below the question text which looks strange in field lists, especially when they're combined with text fields:
 

Load a form with a field list including one or more binary questions such as https://github.com/getodk/collect/pull/4188#event-3962296705.

#### Expected behavior
There should be spacing below question text for read only binary questions.

grzesiek2010:
Closes #4209

I tested the fix manually and added some tests.

We just agreed that if a question is read-only and has no answer we need some space to separate such widgets especially in field-list groups.

Since now if a question is read-only and has no answer an empty container should be displayed instead of its question container. In order to test this fix it would be good to just go through a form with all widgets like the one I attached below.

I used:
[all-widgets-read-only-field-list.xlsx](https://github.com/getodk/collect/files/5603942/all-widgets-read-only-field-list.xlsx)
[all-widgets-read-only.xlsx](https://github.com/getodk/collect/files/5603943/all-widgets-read-only.xlsx)

No.


grzesiek2010:
@lognaturel it turned out that we don't need all those checks if the answer container is gone or not. I don't know what I did wrong that I needed it in my first solutions and I don't want to waste time now to figure it out. It's even better, the solution is cleaner, sorry for the confusion.

lognaturel (approved):
I'd tend to put a comment next to where `space_box` is defined in the XML layouts but don't feel super strongly about it. Looks good!

mmarciniak90:
**Tested with success**

Verified cases:
- read only widgets - the difference is visible especially in widgets:
   - Range
   - Video
   - Data/time
   - Select one/multi. _Keyboard on autocomplete appearance reported as a separate issue._
   - Trigger
- read only widgets on field-list - _Additional issue noticed also on master branch - no space below ex printer widget._
- read only widgets on field-list with answers
- light/dark mode
- RTL language
- go forward/backward
- different font size settings
- rotate screen
- jump from hierarchy
- minimize app

kkrawczyk123:

PR #4230, opened 2020-11-18
Expose more information in HEAD request error

lognaturel:
During some troubleshooting with a user, I discovered that there's a confusing message about a web proxy when there's an attempt to make a submission to a URL that exists but is not an OpenRosa submission endpoint. This is really only likely for people who specify a submission URL in their form because they need to remember to add the `/submission` to the address.

1. Put a form on my device with `<submission action="https://test.central.getodk.org" method="post"/>`: 
[external_widgets.xml.zip](https://github.com/getodk/collect/files/5557879/external_widgets.xml.zip)
1. Finalized an instance, attempted to send it.
1. Verified that the error message is more useful.

This is a simple way to help direct users who make a form design mistake. I considered pulling the message out to be translated but I don't think it's common enough to be necessary. I think it also mostly targets project managers and form builders because a form with this error is very unlikely to make it out to production. 

I decided to move the actual status code to the user-facing message instead of leaving it buried in logs. There's nothing we as developers can do with this information so it should not go to Crashlytics.

This doesn't change any code structure and only modifies a message. It should be extremely safe and could probably even skip QA.

[external_widgets.xml.zip](https://github.com/getodk/collect/files/5557879/external_widgets.xml.zip)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/upload/InstanceServerUploader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/upload/InstanceServerUploader.java
@@ -148,12 +148,12 @@ public class InstanceServerUploader extends InstanceUploader {
                     }
                 }
             } else {
-                Timber.w("Status code on Head request: %d", headResult.getStatusCode());
                 if (headResult.getStatusCode() >= HttpsURLConnection.HTTP_OK
                         && headResult.getStatusCode() < HttpsURLConnection.HTTP_MULT_CHOICE) {
                     saveFailedStatusToDatabase(instance);
-                    throw new UploadException(FAIL + "Invalid status code on Head request. If "
-                            + "you have a web proxy, you may need to log in to your network.");
+                    throw new UploadException("Failed to send to " + uri + ". Is this an OpenRosa " +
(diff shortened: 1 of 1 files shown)


seadowg (approved):

seadowg:
The `InstanceServerUploader` is in a pretty bad place in terms of testing but we know it needs a rework and this isn't the change to push that in. Happy for this to go through without a test as it really just a cosmetic change but lets still stick this through QA given it's a riskier area of the code.

mmarciniak90:
**Tested with success**

Verified on Android: 7.0, 8.1, 10.0

|old version | PR|
|---|---|
|Screenshot_20201124-145455|Screenshot_20201124-145525|


Verified cases:
- tried to send submission manually -> new error message is visible
- auto-send -> new error message is visible
- failed send to Central
- failed send to Google spreadsheet
- status code is visible
- scroll on error dialog
- send correct submission to Central
- send correct submission to Aggregate (also local)
- rotation on dialog
- lock screen
- minimize
- light/dark mode

kkrawczyk123:

Issue #1115, opened 2017-06-08
Include some kind of versioning with settings files


lognaturel:
Currently, there are two ways to import settings through files: putting a `collect.settings` file in the ODK root (usually /sdcard/odk) and reading from a QR code which includes settings JSON. As the pace of development on Collect increases, settings will be more likely to change and we should have a way to ensure that old config files or QR codes continue to be usable.

One example of that can be seen in #1101 -- two settings having to do with autosend were combined into one. In that case, it is possible to check which keys already exist and use that information to do the migration but ideally we would have a standardized way to also deal with brand new keys or changed defaults.

On the json side, we can easily use a `version` key. On the serialized file side, I think we can add another `HashMap` including a `version` key.

How should version be represented? Options I can think of:
1. An integer that increases by 1 each time there's a breaking change. This is similar to database versions used for migrations.
2. A string representing the Collect tag (e.g. `v1.7.0`). We'd have to do some manipulations to it to turn it into a number that can be compared.

I have a slight preference for 1 because it's simple and matches the database approach.

nribeka:
Personally, I prefer the second one since it's going to much easier if there's a need for developers to validate the version number (and I'm pretty sure we will need to manually validate the settings sometimes). The first approach will need a mapping between settings version and the collect version.

It should be a lot of effort to convert the string representation to integer, right?

seadowg:
This would be a good discussion to move to the [forum](https://forum.getodk.org) if it needs to be continued.

PR #4231, opened 2020-11-19
Show navigation buttons for new installs

seadowg:
We've been talking about this for a while. This makes the swipe and button navigation the default option rather than the buttons not being there by default. The hope is that this helps new users to learn how to use Collect.

There are tests (which I updated) and checked manually.

Just went with the simplest change here.

I think we can skip QA on this.

Yes it probably does. I can follow up on that.


grzesiek2010 (approved):

grzesiek2010:
> I think we can skip QA on this.

let's have a quick verification, there is no need to use different devices/android versions just one device and it should be enough.

lognaturel (approved):
I did not internalize that a side effect of always running `ApplicationInitializer.reloadSharedPreferences` on app start is that every preference has a key no matter whether it was set by the user or not. I'm not sure when that behavior started but I think it's within the last year. If someone updates from a Collect version before that, they will get buttons and swipes, I believe. I don't think this is super important but I wanted to make sure it's explicit. I find it a bit strange that settings migration and reloading is done every time Collect is launched but I suppose it's very fast and poses no real risk. Maybe it's something we've discussed before and I'm just forgetting.

Having thought about this more, I think the ideal would be to offer this default to everyone who currently uses `on_swipe`. I'll write a story for it and we'll see how to prioritize it.

seadowg:
> I'm not sure when that behavior started but I think it's within the last year.

I'm pretty sure it's much older than that. Here's `Collect` from 1.20 doing the same thing: https://github.com/getodk/collect/blob/5017a8310019d8d88ab22419d35d50cbcf4e6977/collect_app/src/main/java/org/odk/collect/android/application/Collect.java#L248.

Not sure I want to keep diving back through but I imagine you have to go back pretty far before we hit a version that this would cause problems for right?

lognaturel:
> it's much older than that

Well, then! No need to identify when either way: it's not very common for folks to jump several versions as far as I know.

kkrawczyk123:
@seadowg I have noticed weird behavior when scanning qr code. I have used horizontal swipes on the current store version and scan qr code with PR-Collect version with swipes and buttons selected. The result after scanning is a default selection for the app so swipes and buttons. Don't you think that the user's personal configuration should be scanned instead of the default value? 
When select buttons only on store version and scan with PR app - the setting is scanned correctly. Also when set any option on PR app and scan with PR app - settings are correctly displayed.

seadowg:
> I have noticed weird behavior when scanning qr code. I have used horizontal swipes on the current store version and scan qr code with PR-Collect version with swipes and buttons selected. The result after scanning is a default selection for the app so swipes and buttons. Don't you think that the user's personal configuration should be scanned instead of the default value?

Not sure I understand. Is there a mistake in your description here? It sounds like you're scanning one QR where the settings are "Swipes and buttons" and then getting "Swipes and buttons" after scan which would be expected.

kkrawczyk123:
> Not sure I understand. Is there a mistake in your description here? It sounds like you're scanning one QR where the settings are "Swipes and buttons" and then getting "Swipes and buttons" after scan which would be expected.

No.  I have used **horizontal swipes** on the current store version and scan qr code with PR-Collect version with **swipes and buttons** selected.  The result after scanning is **swipes and buttons**. I expected that horizontal swipes will be the result.

seadowg:
> PR-Collect version with swipes and buttons selected

Ah right I had this backwards. So you're saying you'd expect it to be horizontal swipes as you wouldn't expect swipes and buttons to be included in the QR code (as it's the default)? Did you actually select swipes and buttons in settings or are you just referencing that it will be the setting? Maybe just a full set of reproduction steps would help here.

kkrawczyk123:
Ok, exact steps that I have done:
1. On the Collect v1.28.4 I have selected **horizontal swipes** (default selection)
2. On PR-app I have selected **swipes and buttons** (also default selection)
3. I generated QR code on 1.28.4 
4. I scanned QR code using PR-app
5. On PR-app I have **swipes and buttons** selected (default selection for that app) not horizontal swipes that I wanted to import. 

Does it make sense right now?

seadowg:
@kkrawczyk123 right got it. Sounds like for some reason the default is being included in the PR. I'll check it out!

kkrawczyk123:

lognaturel:
> not horizontal swipes that I wanted to import.

I don't think there's really a whole lot we can do about this if we indeed do want to change the default (and I think we do). In order to address this, we'd need to version the settings file as described in https://github.com/getodk/collect/issues/1115. For QR code size purposes, we don't include settings keys with default values in the code. So we'd need to know that a QR code was generated before the default change in order to maintain that behavior across imports.

In this case, I think the worst side-effect of ending up with navigation buttons when you didn't want them are that some question screens could require scrolling. This seems annoying but not fatal so I think it's acceptable.

seadowg:
> I don't think there's really a whole lot we can do about this if we indeed do want to change the default (and I think we do). In order to address this, we'd need to version the settings file as described in #1115.

@lognaturel you're totally right I still had this backwards in my head. As soon as I re-read the repro steps this morning I was confused at what yesterday Callum was talking about. I'll mark this as "needs testing" again as I agree it's not something we're going to change right now.

kkrawczyk123:
**Tested with success, then**
Verified on Androids: 5.1, 7.0, 7.1, 8.1, 9.0 and 10.

Tested cases:
- light/dark mode
- update app
- scan qr code
- use swipes only
- use buttons only
- RTL language 
- disable moving backwards 

PR #4232, opened 2020-11-19
Use glide to display images in selects

### LINKED ISSUE #4205: Performance improvements in loading images
Collect v1.28.4 and older versions

#### Problem description
This is a continuation of what I discovered and described in https://github.com/getodk/collect/pull/4201#issuecomment-721690499

To sum up the problem is that having choices with images and columns-n appearance is particular leads to heavy operations that on many devices especially those poor might cause ANRs and OutOfMemoryErrors.

I prepared a sample form that show everything clearly:
[selectMinimalTest.zip](https://github.com/getodk/collect/files/5493111/selectMinimalTest.zip)

The form contains a few selects with different configurations (different number of choices, columns appearance or not, styling or not, images or not).
It clearly shows that having images is the most problematic case. It takes a lot of time and in many cases might cause OutOfMemoryErrors.

We can't leave it as is especially now when our audience is growing and more people are experiencing such problems.
I made a quick experiment and used [Glide](https://github.com/bumptech/glide) to load images in choices. The result was more than satisfying (on master opening a dialog took over 20s but using Glide just like 1s):

| Master branch  | using Glide |
| ezgif com-video-to-gif (8)  | ezgif com-video-to-gif (9)  |

Glide is recommended by Google but there are also other popular libraries like [Picasso](https://square.github.io/picasso/) or [Coil](https://github.com/coil-kt/coil) so it might be worth investigating which one is best for us.

Maybe this is not even the only possible improvement we can implement (but I think the most important one). We should review our RecyclerView to check for other possible bottlenecks.

If we add Glide (or any other library) to load images in choices we should do the same in ImageWidegets.

grzesiek2010:
Closes #4205

This pr fixes displaying images in selects I'll create a separate pr for displaying images in image widgets since:
- it will require more refactoring so I didn't want to mix everything in one pr
- it's less important since the behavior there is not that bad

I tested the fix using the forms I attached.

I decided to use Glide since it's the most popular library recommended by google and the performance is also good. I also investigated other libraries:
- **Picasso** I didn't use it because it works much worse in recyclerview
- **Coil** it's an interesting tool but it's relatively new and rather designed for Kotlin so using it in Java is not that convenient

This change might affect displaying images in selects (buttons and no-buttons modes) so that's what we need to test. I atatched forms that should be helpful. Nothing else should be affected just selects.

The form attached to the issue - to confirm that it works faster and maybe this form: 
[selectTest.zip](https://github.com/getodk/collect/files/5569394/selectTest.zip)

No.


collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
```diff
@@ -143,7 +142,20 @@ public void setAudio(String audioURI, AudioHelper audioHelper) {
...
+            Glide.with(this)
+                    .load(imageFile)
+                    .centerInside()
+                    .into(imageView);
+
+            imageView.setVisibility(VISIBLE);
+            imageView.setOnClickListener(this);
+        } else {
+            missingImage.setVisibility(VISIBLE);
+            missingImage.setText(getContext().getString(R.string.file_missing, imageFile));
```
grzesiek2010, line 157: @lognaturel maybe it's time to get rid of those error messages? Glide can display a special image that will be used if something is wrong so we could have such an error/missing file image?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -291,40 +285,6 @@ public QuestionDetails getQuestionDetails() {
         return questionDetails;
     }
 
-    // http://code.google.com/p/android/issues/detail?id=8488
-    private void recycleDrawablesRecursive(ViewGroup viewGroup, List<ImageView> images) {
```
grzesiek2010, line 283: All of these is removed because it caused crashes since Glide use caching. I could disable caching in Glide and then keep this code but it's not recommended and I believe it's a very old code that we no longer need because we have more powerful devices and also glide should deal with it better.

collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
lognaturel: I think using a missing image file is a good idea and will make it more obvious that something is wrong.

The advantage of the message is that it lets the user know where the image is expected to be and what its file name is. How about leaving this message and adding the missing image? Or is there a strong reason to get rid of it?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
lognaturel: This seems reasonable to me. Am I understanding correctly that the Glide cache will be some kind of singleton so on activity rotation all images should be able to come from some cache layer?

grzesiek2010: Yes I'm not sure about rotating and would need to double check but it uses caching for example when we navigate between questions so probably during rotating as well.

collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
grzesiek2010: Ok I'll file a separate issue for that.

lognaturel (approved):

lognaturel:
For QA, the most insightful thing to check would probably be a form with a lot of images on the slowest device you have. Loading, rotation, should hopefully be smoother than before.

mmarciniak90:
**Tested with success**

> form with a lot of images on the slowest device you have

I can see performance improvements even on new and quite fast devices

Verified cases:
- selectMinimalTest form
- selectTest form
- rotation
- minimization
- lock screen

kkrawczyk123:

PR #4233, opened 2020-11-20
Rework ResetStateRule 

seadowg:
This fixes the build. The problem seemed to be that `ResetStateRule` was trying to clean up using `ApplicationResetter`. This meant that it needed the databases to exist for it to be able to delete records in it. Really we don't care as we're just trying to simulate an uninstall so instead it now just deletes the scoped and unscoped root dirs.

I had also had to make changes to a couple of tests that had problems revealed by the changes to to `ResetStateRule`.

Ran locally and on test lab.

Comments inline.

No just changes tests.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/CustomSQLiteQueryExecutionTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/CustomSQLiteQueryExecutionTest.java
@@ -20,18 +20,18 @@ import android.Manifest;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 
-import androidx.test.rule.GrantPermissionRule;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.rule.GrantPermissionRule;
 
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.junit.runner.RunWith;
-import org.odk.collect.android.storage.StoragePathProvider;
-import org.odk.collect.android.storage.StorageSubdirectory;
-import org.odk.collect.android.support.ResetStateRule;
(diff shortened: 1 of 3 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/CustomSQLiteQueryExecutionTest.java
```diff
@@ -20,30 +20,29 @@
...
+import java.io.File;
+import java.io.IOException;
+
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertFalse;
 import static junit.framework.Assert.assertTrue;
 
 @RunWith(AndroidJUnit4.class)
 public class CustomSQLiteQueryExecutionTest {
-    private static final String DATABASE_PATH = new StoragePathProvider().getDirPath(StorageSubdirectory.METADATA) + "/test.db";
```
seadowg, line 42: This was causing problems as the path it would get would be determined when the class was loaded.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/CustomSQLiteQueryExecutionTest.java
```diff
@@ -53,14 +52,15 @@
...
-
-    @Before
-    public void setUp() {
-        sqLiteDatabase = SQLiteDatabase.openOrCreateDatabase(DATABASE_PATH, null, null);
-        dropTable(TEST_TABLE_NAME);
-        dropTable(TEST_TABLE_NAME_2);
-    }
+            .around(new RunnableRule(() -> {
+                try {
+                    File dbPath = File.createTempFile("test", ".db");
```
seadowg, line 57: This is just testing our db helpers so it doesn't need to talk to our "real" db.

grzesiek2010 (approved):

PR #4222, opened 2020-11-16
Add AMR recordings

seadowg:
Closes #4217 
~~Blocked by #4220~~

#### What has been done to verify that this works as intended?

New tests and verified manually.

#### Why is this the best possible solution? Were any other approaches considered?

Didn't need to add much here. I'm not 100% confident we're leaving the door open for someone to add a codec not supported by `MediaRecorder` but my theory is that this could be done with an alternative implementation of `MediaRecorderWrapper`. I think what we have is good for now.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Good just to check we get the right files out based on the right forms.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4241, opened 2020-11-24
Show recording duration and waveform

### LINKED ISSUE #4152: Can see how long a recording has been going on
**Acceptance**

Then I see a timestamp of how long my recording has been going on
And it updates every second

When I pull down the notification drawer on my device
Then I see a timestamp of how long my recording has been going on in the "Recording in progress" notification
And it updates every second

### LINKED ISSUE #4153: Can see a waveform while recording
**Acceptance**

Then I see a waveform that reflects the noise around me

seadowg:
Closes #4152 
Closes #4153
~~Blocked by #4222~~

Comments inline.

I think we want to skip QA on this as it really only touches the new features.


.circleci/config.yml
```diff
@@ -57,6 +57,9 @@ jobs:
       - run:
           name: Run async unit tests
           command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 async:testDebug
+      - run:
```
seadowg, line 60: Now that we have a test file in `strings` we should really be running it on CI.

async/src/main/java/org/odk/collect/async/CoroutineScheduler.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import java.util.function.Consumer
+import java.util.function.Supplier
+import kotlin.coroutines.CoroutineContext
+
+open class CoroutineScheduler(private val foregroundContext: CoroutineContext, private val backgroundContext: CoroutineContext) : Scheduler {
```
seadowg, line 12: We won't always need deferred tasks so thought I'd split the implementations. I think long term it might make sense to split the interfaces but lets see how it evolves.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderViewModel.kt
```diff
@@ -7,12 +7,12 @@ import java.io.File
 
 /**
  * Interface for a ViewModel that records audio. Can only record once session
```
seadowg, line 9: I think we're now in a better place with this abstraction. Adding time/amplitude helped it make more sense as it forced the key data exposed to be the current "session" details.

collect_app/build.gradle
```diff
@@ -320,6 +320,8 @@ dependencies {
     // Better "Subjects" for Rx:
     implementation "com.jakewharton.rxrelay2:rxrelay:2.1.1"
 
+    implementation 'com.github.Armen101:AudioRecordView:1.0.2'
```
seadowg, line 323: This was the simplest waveform library I could find that still looked supported. I think we may want to improve this in the future or add our own flair.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/IntentGroupTest.java
```diff
@@ -128,7 +128,7 @@ public void externalApp_ShouldPopulateFields() throws IOException {
         resultIntent.setClipData(clipData);
         resultIntent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
         intending(not(isInternal())).respondWith(new Instrumentation.ActivityResult(Activity.RESULT_OK, resultIntent));
-        onView(withText("This is buttonText")).perform(click());
+        onView(withText("This is buttonText")).perform(nestedScrollTo(), click());
```
seadowg, line 131: These tests were failing due to the audio widget getting slightly taller and it forcing other items in the test to be scrolled off screen.

strings/src/main/java/org/odk/collect/strings/format/LengthFormatter.kt
```diff
@@ -0,0 +1,18 @@
+package org.odk.collect.strings.format
+
+import java.util.Locale
+
+private const val ONE_HOUR = 3600000
+private const val ONE_MINUTE = 60000
+private const val ONE_SECOND = 1000
+
+fun formatLength(milliseconds: Long): String {
```
seadowg, line 9: Could totally see others disagreeing with moving this here. For me, it makes sense to have formatters etc live in `strings` but could see an argument that they should live in a separate utils module.

testshared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt
```diff
@@ -3,58 +3,72 @@ package org.odk.collect.testshared
 import org.odk.collect.async.Cancellable
 import org.odk.collect.async.Scheduler
 import org.odk.collect.async.TaskSpec
+import java.util.LinkedList
 import java.util.function.Consumer
 import java.util.function.Supplier
 
 class FakeScheduler : Scheduler {
```
seadowg, line 10: For some of the tests this needed to be a bit more advanced so it now actually keeps scheduled tasks in queues/lists so it runs them in the correct order and can run repeated tasks.

collect_app/build.gradle
lognaturel: I see a size increase of .2mb from 1.28.4 to this branch. Not sure how much is from this lib but either way it seems ok.

strings/src/main/java/org/odk/collect/strings/format/LengthFormatter.kt
lognaturel: Seems fine to me.

lognaturel (changes requested):
Looks good to me!

There's a rotation issue that's deeper than I thought -- after rotating during recording, it's not possible to see the amplitude before exiting the form. Even starting a new recording doesn't reset things. I think that should be addressed before merge.

Regarding QA, I think we should do a pass over the amplitude/time functionality before freeze next Weds even if small visual tweaks will be done after that. There might be other weird state things I didn't notice and it'd be better to get those addressed before regression testing begins. Are you thinking you'd like to do that after your design pairing on Friday?

seadowg:
@lognaturel fixed! It looks like it's a bug/quirk in the library.

> I see a size increase of .2mb from 1.28.4 to this branch. Not sure how much is from this lib but either way it seems ok.

Yeah would be really nice if we could replace with something simple we write ourselves.

lognaturel (approved):

lognaturel:
@getodk/testers Heads up that we'll get these recording visualization features to you after a few more visual tweaks!

PR #4242, opened 2020-11-24
Fixed: Minimal appearance doesn't work with "fast external itemsets"

### LINKED ISSUE #4051: Minimal appearance doesn't work with "fast external itemsets"
`master`

#### Problem description
Minimal appearance doesn't work with ["fast external itemsets"](http://xlsform.org/en/#external-selects).

Load a form with "fast external itemsets" and the minimal appearance such as [select_one_external.xlsx](https://github.com/getodk/collect/files/5138302/select_one_external.xlsx).

#### Expected behavior
The `minimal` and `minimal autocomplete` appearances should have the same behavior as with other selects.

I believe this was intended to be addressed by #3964 but you'll have to confirm, @grzesiek2010.

grzesiek2010:
Closes #4051

I decided to get rid of the `ItemsetWidget` and just read all items in our `ItemsWidget` thanks to that I can use the same appearances we use in other selects no mater if items come from a form file on an external csv  file.

Prior to this change we used a separate widget called ItemsetWIdget to handle fast external itemsets now I go rrid of it and we sue the same widgets everywhere selects no mater if items come from a form file on an external csv  file. That means the risk is in handling those fast external itemsets and that's what we need to test additionally it would be good to check normal selects for regression.

I used:
[selectOneExternal.zip](https://github.com/getodk/collect/files/5590442/selectOneExternal.zip)

No.


grzesiek2010:
@lognaturel @seadowg 
I can imagine your disappointment when you notice that I added an espresso test instead of a bunch of unit tests so I would like to explain my rationale. I think we need refactoring that would allow us to easily test select widgets without mocking javarosa objects because it's a nightmare even to me a person who knows the code, if we had a volunteer such a person would probably say bye seeing that part of code.
We need something like ItemsRepository to be able to prepare choices for our tests as I said above without mocking what's happening on javarosa's side. We need such a refactoring especially now when we are going to merge another pr that will allow us to come up with even more possible variations of appearances.
So I didn't want to start dinging into it now because it would be a problem to me to find a balance (a good place where to stop my refactoring) and it would take me a lot of time.

Such ItemsRepository could provide choices along with media files so that we won't need to read everything using FormEntryPrompts everywhere in our codebase and in tests.

Have you maybe thought about it as well?

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidget.java
```diff
@@ -40,7 +44,15 @@
 
     public ItemsWidget(Context context, QuestionDetails prompt) {
         super(context, prompt);
-        readItems();
+        if (isFastExternalItemsetWidget()) {
```
lognaturel: This should be another branch along with the `search()` identification. Currently that is in `readItems`. I vaguely remember we had some back and forth about whether `readItems` should be separate and maybe it was related to "fast external itemsets." My preference would be to have the three branches in the constructur and not have `readItems` at all.

collect_app/src/main/java/org/odk/collect/android/utilities/SelectOneWidgetUtils.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public static Selection getSelectedItem(FormEntryPrompt prompt, List<SelectChoice> items) {
+        IAnswerData answer = prompt.getAnswerValue();
+        if (answer == null) {
+            return null;
+        } else if (answer instanceof SelectOneData) {
+            return (Selection) answer.getValue();
+        } else if (answer instanceof StringData) { // Fast external itemset
+            for (SelectChoice item : items) {
```
lognaturel, line 25: How was this done before? It'd be ideal not to loop through all items.

collect_app/src/main/java/org/odk/collect/android/fastexternalitemset/ItemsetDbAdapter.java
```diff
@@ -1,11 +1,12 @@
-package org.odk.collect.android.database;
```
lognaturel, line 1: I think introducing a new package to group what you have is a great idea. I wonder whether it should be top-level, though. It could be in `formentry` or `widgets`. While here, I wonder what the `itemset` package and `Itemset` class were supposed to be for. Seems they can be removed.

lognaturel (changes requested):
Unifying all the ways to define items is great.

I agree with doing integration tests for now. Perhaps `ExternalDataFileNotFoundTest` could be combined with the one you added. I was initially a little concerned to see `ItemsetWidgetTest` go but it's really not doing much.

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidget.java
grzesiek2010: You mean three methods for different sources what would you call them especially I have a problem with those two external sources? fastExternalItemsets and just externalItemsets or what?

collect_app/src/main/java/org/odk/collect/android/utilities/SelectOneWidgetUtils.java
grzesiek2010: We have a method in [SelectOneWidget](https://github.com/getodk/collect/pull/4242/files#diff-6ccb223484e6d1101c41b833d6cd419cab66f2b6d973907112dea076cb6b7275R82) and also it was overrode in [ItemsetWidget](https://github.com/getodk/collect/pull/4242/files#diff-03d8e0549dc87e59bb6e27487024b6b81daa3dc0152a334628da85b33b4723b3L84)

The problem is that we use DATATYPE_TEXT (so different data types to read answers) in case of external itemsets probably in the future we should change it adding maybe even support for select_multiple then we won't need such checks but that's on javarosa's side.

collect_app/src/main/java/org/odk/collect/android/fastexternalitemset/ItemsetDbAdapter.java
grzesiek2010: I added it there because we have also external package it's top-level too which contains classes responsible for reading external csv files we use to handle search/pulldata.

> While here, I wonder what the itemset package and Itemset class were supposed to be for. Seems they can be removed.

Yes it's unused. I added it implementing storage migration to handle migrating itemsets databases but then we decided not to do that because we can clear them just and such databases will be recreated and probably I forgot to remove it.

grzesiek2010:
> Perhaps ExternalDataFileNotFoundTest could be combined with the one you added.

Yes and no, my test check fast external itemsets only but the one you mentioned checks also those which use search/pulldata again do we have good names for those which I could use to make it clear because in both cases we use external csv files.

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidget.java
lognaturel: Yes, three branches. I think `searchItems` or something like that would work. I'm not too picky but it should include `search`. It shouldn't be `externalItemsets` because that really sounds like external secondary instances. That one doesn't show up here because JavaRosa handles it entirely.

The weird thing about `search()` and fast external itemsets is that they live entirely outside of the XForms spec and JavaRosa. That's why they're not supported by Enketo and lead to problems like the one you saw recently on the forum where competing instances were being generated by `pyxform`. Internal secondary instances, external secondary instances and static choice lists in the form body are all the same as far as Collect is concerned.

collect_app/src/main/java/org/odk/collect/android/utilities/SelectOneWidgetUtils.java
lognaturel: Both of the methods you linked to just return the string value. Is `getSelectedItem` only needed for certain appearances? The reason folks use "fast external itemsets" is because they're database-driven and therefore fast. I wouldn't want to add overhead for the common case which is a simple select one with no appearance.

grzesiek2010: Oh sorry my mistake it's not how I wanted to explain it. The problem is that if we use fast external itemsets DATATYPE_TEXT is used which uses StringData but for normal selects we use CONTROL_SELECT_ONE which uses SelectOneData. So when we save a form with answers and open it it might receive StringData or SelectOneData and now we need to handle it since we use the same wights no matter what the source of items is.
You can notice that difference in methods I linked as well. That's why I said it would be good to use the same datatypes implementing changes in javarosa isn't that possible?

lognaturel: Ah, I think I see what you mean. 

> That's why I said it would be good to use the same datatypes implementing changes in javarosa isn't that possible?

JavaRosa doesn't know anything at all about either "fast external itemsets" or `search()`. They're entirely client-side features. So no, not really. Maybe you could look at how the `search()` code ends up with `CONTROL_SELECT_ONE` types?

grzesiek2010: The difference is that when we convert xls to xml using `search()` it's still `select1` element in xml but `select_one_external` used in xls ends up as `<input query="...` and then in Collect [if such `<input` has query](https://github.com/getodk/collect/pull/4242/files#diff-d65ca3cdb4834d0b00ca74ecde18b9930e40762a6c48e2fbb991d8e6cfd4c28dR153) we treat it as select_one question. Probably we could add such a check earlier - on javarosa's side and treat it as `CONTROL_SELECT_ONE` not `DATATYPE_TEXT`

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidget.java
grzesiek2010: > I'm not too picky but it should include search

Now as I'm thinking about it I would rather use `readFastExternalItems()` and `readDynamicItems()` + that last one option that comes from a form file maybe `readStaticItems()`?
I would rather avoid that search word because it might use search but also might use populate right it's also like search but dynamic somehow sounds better to me?

collect_app/src/main/java/org/odk/collect/android/utilities/SelectOneWidgetUtils.java
lognaturel: Hmm. I see, the body element types are different, that's right. I really don't want us to start leaking these weird features into JavaRosa where there's already enough going on. Could you please do a quick performance check before and after your change for 5k items? I think it should be really quick to generate in Excel but let me know if not. There already was a list of `SelectChoice` items generated for "fast external itemsets" prior to your change, right?

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidget.java
lognaturel: I've seen the terms "static" vs "dynamic" to contrast choice lists that are embedded in the form body of the primary instance vs. in a secondary instance. So I think that's confusing.

> it might use search but also might use populate

Could you please describe "populate"? The `search()` appearance/function is what identifies that external data feature in form design.

Other ideas:
- `itemsetsCsvDatabaseItemsets` vs `arbitraryCsvDatabaseItemsets`
- `databaseQueryItemsets` vs `databaseSearchItemsets`
- `readFastExternalItems` vs `readSearchExternalItems`
- `readFastExternalItems` vs `readSearchPulldataItems`
- `readDatabaseItemsetsItems` vs `readDatabaseSearchItems`

grzesiek2010: > Could you please describe "populate"?

sorry I meant pulldata ok `readFastExternalItems vs readSearchPulldataItems` seams fine.

collect_app/src/main/java/org/odk/collect/android/utilities/SelectOneWidgetUtils.java
grzesiek2010: I checked it using android 6 with a device which is not very fast and didn't notice any change in performance comparing this branch with the master branch. I used this form if you want to check it on your own: 
https://drive.google.com/file/d/1lXZtqhGyMJZW5BFrRyPdTcSwRgQ95wQZ/view?usp=sharing
I selected one of the last choices filling the form on purpose to make the searching longer.

grzesiek2010:
I updated the form with all possible combinations of appearances it's here https://docs.google.com/spreadsheets/d/1uA6NeUbE51WACiKUjErFUxYFeJdGGWglNGKivuzei7I/edit?usp=sharing
cases that will be fixed by this pr are marked in yellow.

lognaturel (approved):

mmarciniak90:
@grzesiek2010, there are issues in cases:

1. Fill answers for a minimal appearance
2. **Change answer** on first-level - state
3. Answers for County and City are not cleared
---
1. Fill answers for a minimal appearance
2. **Remove response** from first-level - state
3. Answers for County and City are not cleared

ezgif com-video-to-gif

grzesiek2010:
That whole clearing is faulty it seems to work well if we have a simple cascading select with questions on separate screens but it gets complicated:
- if we use groups putting answers on one screen
- if we have additional questions between cascading select levels what's also possible
- if we wrap particular levels into groups etc.

it might be also dangerous and try to clear what shouldn't be cleared. I think @lognaturel shares my concerns.

If it's the only bug let's file a separate issue to analyze that clearing mechanism and decide what to do.

mmarciniak90:
**Tested with success**

Verified cases:
- selectOneExternal form - cascading select with minimal appearance
- cascading select external itemsets form
- standard cascading select 
- move forward/backward
- jump from the hierarchy view
- removed response - new issue created
- change response - new issue created
- use search
- form edit
- light/dark mode
- RTL language
- rotate screen
- minimize app

kkrawczyk123:

PR #4243, opened 2020-11-25
Explicitly address repeat in field-list case

### LINKED ISSUE #4145: RuntimeException in FormController.java:922

lognaturel:
Closes #4145

This updates the message shown in case of a repeat in a field-list and makes sure the error is caught everywhere it may occur.

Tried it with https://docs.google.com/spreadsheets/d/159jGIYoZ5-f1JQTAWyfUFqk9-3k4mmbaM14KgbKbCbA/edit#gid=0, verified the new message is shown. This form does not result in the stack trace from the issue but it's similar.

@grzesiek2010 and I discussed several other options. He proposed using a snackbar generated in `FormController.getQuestionPrompts`. I don't love that this means having display code in a controller. His other idea was to replace the misplaced repeat by an inline message. I like this but it's a whole new concept. My thinking is that we'd like to support repeats in field lists at some point so we just need a temporary way to keep alerting users that it's not currently supported. Making the dialog message more helpful and making sure it always shows up seems like a good way to do this.

I introduced `FormDesignException` because this feels like a very particular kind of issue. Using a `RuntimeException` led us to the crash from the initial issue because calling sites didn't require that the exception be handled.

This should only provide more helpful messaging in the case of a repeat in a field-list. Regression risk should be low and all related to exceptions during form filling.

https://docs.google.com/spreadsheets/d/159jGIYoZ5-f1JQTAWyfUFqk9-3k4mmbaM14KgbKbCbA/edit#gid=0 is one.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -78,6 +78,7 @@ import org.odk.collect.android.analytics.Analytics;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.audio.AudioControllerView;
 import org.odk.collect.android.backgroundwork.FormSubmitManager;
+import org.odk.collect.android.forms.FormDesignException;
 import org.odk.collect.android.fragments.dialogs.CustomDatePickerDialog;
 import org.odk.collect.android.fragments.dialogs.CustomTimePickerDialog;
 import org.odk.collect.android.dao.FormsDao;
@@ -1091,9 +1092,9 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     }
                     return false;
                 }
-            } catch (JavaRosaException e) {
+            } catch (JavaRosaException | FormDesignException e) {
                 Timber.e(e);
-                createErrorDialog(e.getCause().getMessage(), DO_NOT_EXIT);
+                createErrorDialog(e.getMessage(), DO_NOT_EXIT);
                 return false;
             }
         }
@@ -1236,7 +1237,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     Timber.i("Created view for group %s %s",
                             groups.length > 0 ? groups[groups.length - 1].getLongText() : "[top]",
                             prompts.length > 0 ? prompts[0].getQuestionText() : "[no question]");
-                } catch (RuntimeException e) {
+                } catch (RuntimeException | FormDesignException e) {
                     Timber.e(e);
                     // this is badness to avoid a crash.
                     try {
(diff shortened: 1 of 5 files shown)


grzesiek2010 (approved):

mmarciniak90:
@lognaturel, I see a double space in error message: `not  supported` 

Verified cases:
- `field-list-repeat` form -> **user friendly error message is visible**

|PR|v1.27|
|---|---|
|Screenshot_20201126-174844|Screenshot_20201126-174648|

- `repeat_group_form` form
- `bad-instance-name` form
- rotation on error -> Form in background disappears. If error is closed, user is moved to main screen. The same behavior as before.
- adding groups
- edit submission
- jump from hierarchy view
- minimization
- light/dark mode
- RTL language

kkrawczyk123:
@lognaturel I have noticed a small issue: When close opened dialog and swipe back it is displayed again (what is fine as it works in the same way on 1.27 and master) but when I swipe twice fast the content of the error is doubled (this is not visible on 1.27 and master), see the gif:
ezgif com-video-to-gif (43)

lognaturel:
> I see a double space in error message: not supported

Ah, thanks. I went back and forth on putting "currently" in there and looks like I didn't clean up very well! Fixed.

> the content of the error is doubled 

I'm inclined to let this one go. I think what's going on is that the error is being generated from two different places. Do you think it's a big issue or might be hiding something more serious?

mmarciniak90:
> Do you think it's a big issue or might be hiding something more serious?

We see no danger for serious consequences here.

PR #4244, opened 2020-11-25
Allow disabling external recording feature

### LINKED ISSUE #4229: Admin can disable external recording setting
**Acceptance**

- [x] Given I have access to Admin Settings
When I navigate to "User settings"
And I uncheck "External app recording"
Then that setting is hidden in General Settings

- [x] Given I have access to admin settings
When I navigate to "User settings"
And I check "External app recording"
Then that setting is shown in General Settings

seadowg:
Closes #4229

Just using our standard pattern for tests and implementation here.

Just good to verify acceptance criteria here I think.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/AdminKeys.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/AdminKeys.java
@@ -140,7 +140,8 @@ public final class AdminKeys {
             KEY_HIGH_RESOLUTION,
             KEY_IMAGE_SIZE,
             KEY_GUIDANCE_HINT,
-            KEY_INSTANCE_FORM_SYNC
+            KEY_INSTANCE_FORM_SYNC,
+            KEY_EXTERNAL_APP_RECORDING
     );
 
     static Collection<String> userInterfaceKeys = Arrays.asList(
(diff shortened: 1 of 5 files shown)


lognaturel (approved):

mmarciniak90:
@seadowg, this option is not visible at all in General Settings when you uncheck all options except the `Use external app for audio recording` in Admin Settings.

ezgif com-video-to-gif

I have also a suggestion to move `Use external app for audio recording` option under `Show guidance for questions` to follow the same order as in the Form Management list.

mmarciniak90:

seadowg:
@mmarciniak90 oh wow I didn't realize that we hide sections if all the keys were disabled. What do you think about just not doing that? It's pretty easy to not realize and create the problem (like I just did).  We could add tests but I think you'd still need to remember to add a new test every time we move settings around or add/remove one. To me, it feels like a pretty risky feature with not a lot of value -worst case scenario the enumerator is able to click into a blank screen. cc/ @lognaturel.

mmarciniak90:
@seadowg, it's not new behavior. We checked a similar case to other options, eg. guidance. Displaying a blank screen in General Settings is possible.

seadowg:
@mmarciniak90 right. My feeling is it'd be great if none of us had to keep track of it! For the moment I can fix and I'll add an issue about removing the feature.

seadowg:
@mmarciniak90 fixed!

mmarciniak90:
@seadowg thanks! 

**Tested with success** 

Verified on Android versions: 7.0, 8.1, 10.0, and emulator with v5.0

Verified cases:
- `Use external app for audio recording` checked in Admin Settings
   - option is visible in the user's General Settings
- `Use external app for audio recording` unchecked in Admin Settings
   - option is not visible in the user's General Settings
   - option is still visible in admin's General Settings
- only `Use external app for audio recording` checked in Admin Settings
   - Form management with option is visible
- scan QR code with option disabled/enabled
- light/dark mode
- RTL language

PR #4251, opened 2020-11-30
Reorder strings to group by usage

### LINKED ISSUE #4072: Add context to existing strings, write up requirements for new strings

lognaturel:
Closes #4072

@seadowg @grzesiek2010 I found when translating Central that it can really help translators to have related strings grouped together. I also think we've had a hard time verifying that language is consistent and that there are no duplicates because this file has been so disorganized. 

I have loosely organized strings based on the UI but I wasn't religious about it. For certain things like geo features or the new audio work @seadowg has been working on I left strings grouped by feature. I think that as long as we have some organization that lets us relatively quickly find related strings it doesn't have to be perfect.

Sections are preceded by a comment block like this:
```
    <!--
     ##############################################
     # SECTION NAME
     ##############################################
     -->
```
I did it this way so it would stand out visually and would be easy to search. For example searching for "# FILL" gets you right to the "Fill Blank Form" section.

The sections are:
- Fill blank form
- Form features
- Geo
- Send finalized form
- Get blank form
- Delete saved form
- General settings
- Admin settings
- Google Drive, Google Sheets
- About
- Settings import
- Scoped Storage Migration
- Misc

I also introduced line breaks between subgroups of related strings.

There were lots of changes I was tempted to make as I worked through this but I held back to hopefully make this not too controversial. I'll come back at the beginning of the next release cycle with some proposed wording changes.

Ran lint to make sure the expected number of translations are there. Ran on a real device to spot check the string consolidation done in later commits.

I considered just adding the comments the forum poster had requested and moving on but this file has caused me enough pain that I wanted to take the time to clean it up some.

I considered splitting the initial commit up but I decided it wasn't very important because if any strings had gone missing the code wouldn't compile. It's going to be nearly impossible to verify the `strings.xml` diff but I think that's ok.

The biggest risk I can think of is that I somehow accidentally deleted or broke some of the text. I think anything like that with common strings would be caught in the regression pass. I'm not sure we need to QA this.

If someone has an actively-maintained fork in which some text is modified, this will be annoying. But it's easier to deal with than many of the recent refactors. Additionally, no maintainers of forks have been active members of the community recently. 

No.

I don't think it really needs an explicit mention in developer docs but let me know if you'd prefer it.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -898,14 +898,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             case RequestCodes.AUDIO_CHOOSER:
             case RequestCodes.VIDEO_CHOOSER:
             case RequestCodes.IMAGE_CHOOSER:
-                /*
-                 * We have a saved image somewhere, but we really want it to be in:
-                 * /sdcard/odk/instances/[current instnace]/something.jpg so we move
-                 * it there before inserting it into the content provider. Once the
-                 * android image capture bug gets fixed, (read, we move on from
-                 * Android 1.6) we want to handle images the audio and video
-                 */
-
                 ProgressDialogFragment progressDialog = new ProgressDialogFragment();
                 progressDialog.setMessage(getString(R.string.please_wait));
                 progressDialog.show(getSupportFragmentManager(), ProgressDialogFragment.COLLECT_PROGRESS_DIALOG_TAG);
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,10 @@ jobs:
       - store_test_results:
           path: collect_app/build/test-results
 
+      - run:
+          name: Assemble connected test build
+          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 assembleDebugAndroidTest
+
   test_instrumented:
     <<: *android_config
     environment:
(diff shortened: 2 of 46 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -898,14 +898,6 @@ protected void onActivityResult(int requestCode, int resultCode, final Intent in
             case RequestCodes.AUDIO_CHOOSER:
             case RequestCodes.VIDEO_CHOOSER:
             case RequestCodes.IMAGE_CHOOSER:
-                /*
```
seadowg, line 901: Guessing this just isn't relevant anymore?

seadowg (dismissed):
Definitely happy with this. I'd really like to just have multiple files for this but it seems like it'd be an absolute nightmare for translation (in Android Studio and for translators). I'm dreaming of a Gradle plugin that consolidates multiple string files into one as part of the build chain...

seadowg (changes requested):
I went to run this on Test Lab as I saw you hadn't checked the `checkAll` task off and it looks like there are some compilation failures for `:collect_app:compileDebugAndroidTestJavaWithJavac` (so when compiling Espresso tests).

As part of fixing could you add a step to `test_app` that builds the `collect_app` Espresso tests. Would be nice to catch that in CI if we can! I think we may have been doing that before but it must have got lost when I reordered things.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: That's right -- that code has changed a lot and we're way past Android 1.6!

lognaturel:
> add a step to test_app that builds the collect_app Espresso tests

Great idea, thank you! Done and fixed the resources I hadn't updated.

grzesiek2010 (approved):

seadowg (approved):

PR #4158, opened 2020-10-07
Downgraded commons-io dependency

grzesiek2010:
Closes #4157

#### What has been done to verify that this works as intended?
I tested performing migration.

#### Why is this the best possible solution? Were any other approaches considered?
It looks like newer versions of commons-io use [java.nio.file.CopyOption](https://developer.android.com/reference/java/nio/file/CopyOption) which in android is available on API 26+ only.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
We just need to test migration on various android versions.

#### Do we need any specific form for testing your changes? If so, please attach one.
No.

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/getodk/docs/issues/new) and include the link below.
No.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4252, opened 2020-12-01
Fixed closing cursor

### LINKED ISSUE #4159: Collect crashes when add 1000+ submissions manually on Androids 5.1-8.1
Collect v1.28 master.x, Android v5.1, 7.0, 7.1 and 8.1 device used...

#### Problem description
When testing #4158 I have noticed that Collect is crashing after loading 1000+ instances that were copied manually to instance folder. 

1. Copy form and instances manually to Collect folder: [backup1249.zip](https://github.com/getodk/collect/files/5346614/backup1249.zip)
2. Go to Collect open one of Edit Saved Forms/Send Finalized/ Delete Saved forms so submission will start to load
3. Wait until all submissions load: Collect crashes

#### Expected behavior
App should not crash, user should be able to continue

#### Other information 
- the issue does not occur on store version
- the issue does not occur on Android 9+
- the issue occurs before and after storage migration 
- none stacktrace seems to be visible 
- I know that it is not a very common scenario but this zip file has been used for testing purposes for a long time and it has never crashed before so it must be some kind of regression. 

@getodk-bot label "Specific to Android version"

grzesiek2010:
Closes #4159

I tested the fix manually.

We need to close cursor after using it and try-with-resources is a good way to do that.

It's a safe fix so using the zip folder attached to the issue and confirming that everything works well would be enough.

The zip folder which is attached to the issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/InstanceSyncTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/InstanceSyncTask.java
@@ -236,10 +236,11 @@ public class InstanceSyncTask extends AsyncTask<Void, String, String> {
                                          ContentValues values, InstancesDao instancesDao)
             throws EncryptionException, IOException {
 
-        Cursor instanceCursor = new InstancesDao().getInstancesCursorForFilePath(candidateInstance);
-        if (instanceCursor != null && instanceCursor.moveToFirst()) {
-            if (shouldInstanceBeEncrypted(formCursor)) {
-                encryptInstance(instanceCursor, candidateInstance, values, instancesDao);
+        try (Cursor instanceCursor = new InstancesDao().getInstancesCursorForFilePath(candidateInstance)) {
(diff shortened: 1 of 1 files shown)


grzesiek2010:
@kkrawczyk123 
I was able to reproduce the crash you probably experienced using older branches for example the one you mentioned #4158 but not on master so I think the issue no longer exists. I only noticed one problem with not closed cursors. It wasn't a crash but who knows maybe on some devices it could cause OutOfMemoryErrors or something like that.
Please just confirm that the crash you saw no longer exists.

seadowg (approved):

kkrawczyk123:
**Tested with success!**

Tested cases:
- scenario form issue
  - no crashes visible

- opened all main menu options when more 1000 sobissions added
- regression check on sending 1000 submissions to Aggregate, Google sheet and Central
  - no crashes visible

mmarciniak90:
Verified also on Androids: 7.0, 8.1, 10.0

PR #4255, opened 2020-12-01
Make external recording the default

### LINKED ISSUE #4254: External recording is the default
**Acceptance**

- [x] Given I install Collect
When I start recording with an audio question
Then it uses an external app to record

- [x] Given I install Collect
When I go to Form Entry settings
Then "Use external app for recording" is enabled

seadowg:
Closes #4254 

Changed tests and verified manually

Just reversed the default so no choices here really.

Should be able to skip QA for this as it's easy to verify when we're playing with the app.


lognaturel (approved):

lognaturel:
@getodk/testers The default audio mode will be `external` for this release so we don't affect users' workflows and we have an opportunity for more feedback before making internal recording the default. No need to verify this PR but you'll see that change when you see any audio questions. A form-specified quality like `voice-only` or `normal` still forces internal recording.

PR #4256, opened 2020-12-02
Add pause/unpause to recording

### LINKED ISSUE #4234: Can pause and unpause recording
**Acceptance**

- [x] Given I've started recording with the in-app recorder
When I click "Pause"
And wait for a few seconds
And press "Resume"
And then stop
Then the recording does not contain any audio from while the recording was paused
And the recording does include audio from after the recording was resumed

- [x] Given I've started recording with the in-app recorder
When I click "Pause"
And press "Stop"
Then the recording does not contain any audio from while the recording was paused

- [x] Given I'm using an SDK 23 device
Then I don't see the "Pause" button

seadowg:
Closes #4234
~~Blocked by #4241~~

The interface isn't the greatest right now (switches between Pause/Resume with same icon) but will iterate on this when we do design polish.

Just uses the Android `MediaRecorder` to pause so no big changes here. Annoyingly this doesn't work below API 24. Additional comments inline.

Definitely think we should leave QA on this until we have the "finished" (ready for release) feature set.

### DIFF
--- a/audiorecorder/src/main/java/org/odk/collect/audiorecorder/DaggerSetup.kt
+++ b/audiorecorder/src/main/java/org/odk/collect/audiorecorder/DaggerSetup.kt
@@ -10,9 +10,9 @@ import dagger.Provides
 import kotlinx.coroutines.Dispatchers
 import org.odk.collect.async.CoroutineScheduler
 import org.odk.collect.async.Scheduler
-import org.odk.collect.audiorecorder.recorder.MediaRecorderRecorder
-import org.odk.collect.audiorecorder.recorder.RealMediaRecorderWrapper
 import org.odk.collect.audiorecorder.recorder.Recorder
+import org.odk.collect.audiorecorder.recorder.RecordingResourceRecorder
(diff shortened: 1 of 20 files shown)


audiorecorder/src/main/java/org/odk/collect/audiorecorder/DaggerSetup.kt
```diff
@@ -65,7 +65,7 @@ internal open class AudioRecorderDependencyModule {
 
     @Provides
     open fun providesRecorder(application: Application): Recorder {
-        return MediaRecorderRecorder(application.cacheDir) { RealMediaRecorderWrapper(MediaRecorder()) }
+        return RecordingResourceRecorder(application.cacheDir) { MediaRecorderRecordingResource(MediaRecorder()) }
```
seadowg, line 68: Naming is hard. I'm attempting to make it clear that this could be implemented using something other than `MediaRecorder`. It's the direction I think we'd need to take if we look at a lossless codec.

lognaturel (approved):

PR #4261, opened 2020-12-02
Allow user to recover from multiple forms with same formid/version and prevent their future download

### LINKED ISSUE #4237: If two forms have the same form_id and version, soft delete prevents the user from recovering
Collect v1.28+

#### Problem description
If two forms have the same form_id and version, soft delete prevents the user from recovering. The user deletes one from the UI but it only gets soft deleted in the database so when the user tries to save a form with that form_id and version, there's a conflict.

[This forum thread](https://forum.getodk.org/t/error-about-two-different-forms-with-same-formid-and-version/31474) has a user running into it.

1. Push any form to a device
1. Copy that form definition and change ONLY the form title
1. Push the copy to the device
1. Go to delete saved form and verify there are two copies of the form, delete one
1. Fill out the one version of that form, get an error on attempt to finalize

#### Expected behavior
If the user deletes one of the forms with the same form_id/version, they should be able to keep using the one that was not deleted.

#### Other information 
Options:
- Use database ID when looking up blank form
- Prevent ingestion if the form_id/version is already in use
- When the user deletes a form, if another one exists with the same form_id/version combination, trust them and actually delete the form
- Filter out soft-deleted form definitions when doing uniqueness checks

lognaturel:
Closes #4237

Strongly recommend reviewing commit-by-commit.

If this were not a legacy project we'd want to use a hash to represent a unique form. An improvement we could make is linking instances and the forms they were filled from through database ids rather than through formid/version. However, that's an invasive change that I don't think we should take on so close to a planned release. It's probably not reasonable to take on at all before some of the mission-critical parts of the code have better test coverage.

Added automated tests for the changes to form downloads and delete behavior. I don't think there are low level tests to write for the changes to instance load and finalizing given the current code structure and I didn't want to get into a risky refactor. I think the problem is enough of an edge-case that I don't want to introduce Espresso tests. I verified those manually for now.

Cases
- adb push form with same formid/version and same filename. Collect should replace no matter the contents. Can get in a bad state if you have filled forms for one older version and there are schema changes. Won't be able to open the filled forms.
- adb push two identical forms (same md5 hash) but different filenames. Collect will treat these as different and this will cause the bad state from the issue.
- adb push same formid/version but the form contents are different (different md5 hashes). Collect will treat these as different and this will cause the bad state from the issue.
- try to download form with same formid/version and same filename. Collect should download and replace. There may or may not be new media attachments.
- try to download form with same formid/version/contents but different filenames. Collect should download but treat as different form. This will cause the bad state from the issue.
- try to download form with same formid/version/filename but different contents. Collect should prevent download.
- get a form on the device. Fill it out a few times. Delete it. Get a form with the same
- got into a bad state in v1.28 through one of the ways above, soft deleted one of the forms but was still suck
- got into a bad state in v1.28 through one of the ways above, had not yet deleted one of the forms

I think this is a good balance of not too invasive and pretty thorough. It both allows folks who ended up in a bad state in v1.28 to recover and limits the ways that the bad state can be reached in the future.

The first two commits filter out soft deleted forms when finalizing or opening filled forms. After those, it should be possible to recover from the bad state by deleting one of the blank forms. This helps folks who already ended up with the problem in v1.28 and soft-deleted one of the copies as instructed.

The third commit prevents downloading a form with the same formid/version as a form already on the device but with different contents. It still allows downloading a form with the same formid/version and the same contents because some servers will allow that to be combined with a media file update (Aggregate but not Central). This should make it harder for users to get into the bad state. There is still the possibility that the remote form would have the same contents but a different filename. We can't guard against that before downloading because the form XML endpoint may not include the filename. We'd have to do the check after downloading the file and I don't think this is common enough to add a special case. Plus, the two first commits will allow users to recover. **Note**: there should be a helpful message from manual download or automatic form update. However, in the match exactly form update mode, there will just be a generic "Failure" message, including if the refresh button is tapped manually.

This affects form load, form finalization, form download and form delete. There are opportunities for regression in any of these but I tried to make the changes easy to review to mitigate that. The challenge is that there are a lot of possible combinations to think of. I tried to list those above in the verification section.

I think there's also a fair amount of risk around the scenarios that we added soft delete for. That is, after deleting a form, its instances should still be editable and finalizable. I tried to verify all that manually and also `DeleteBlankFormTest` was useful.

No, any form will do.

I don't think so. It doesn't seem common enough to mention.


collect_app/src/main/java/org/odk/collect/android/database/DatabaseFormsRepository.java
```diff
@@ -48,6 +48,15 @@ public DatabaseFormsRepository() {
         }
     }
 
+    @Override
+    public List<Form> getAll(String jrFormId, @Nullable String jrVersion) {
```
lognaturel, line 52: I'm not sure about naming, here. There's `getByJrFormIdNotDeleted` above. But I can't think of a case where we'd want deleted forms so calling it `getAll` seemed right to me.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFormsRepository.java
```diff
@@ -57,10 +66,11 @@ public Form get(Long id) {
     @Nullable
     @Override
     public Form get(String jrFormId, @Nullable String jrVersion) {
-        if (jrVersion != null) {
-            return queryForForm(JR_FORM_ID + "=? AND " + JR_VERSION + "=?", new String[]{jrFormId, jrVersion});
+        List<Form> all = getAll(jrFormId, jrVersion);
+        if (!all.isEmpty()) {
+            return all.get(0);
```
lognaturel, line 71: This is a change in behavior because now deleted forms will be excluded. That seems like the desired behavior to me. Also note that if there is a formid/version conflict, we'll just get the first one. That hasn't changed.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -40,19 +43,31 @@
...
-        if (formOnDevice != null && formOnDevice.isDeleted()) {
-            formsRepository.restore(formOnDevice.getId());
+        if (formOnDevice != null) {
+            if (formOnDevice.isDeleted()) {
+                formsRepository.restore(formOnDevice.getId());
+            } else if (!(form.getHash().equals(formOnDevice.getMD5Hash()))) {
+                String formIdentifier = formOnDevice.getDisplayName() + " " + formOnDevice.getId();
+                String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
+                analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION, formIdHash);
+                throw new FormDownloadException("You've already downloaded a form with the same ID and version but with different contents. " +
```
lognaturel: I went back and forth on whether to localize the download message. I think it's uncommon enough that we don't need to. I've added analytics so that we can verify that claim. I also hope that if it is more common than I expect we'll hear from users on the forum and perhaps we'll choose to work on the wording. "please send all data you have collected with the existing form and delete the data and blank form" is admittedly awkward and I'm open to alternatives.

collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
```diff
@@ -271,12 +271,9 @@ public static EncryptedFormInformation getEncryptedFormInformation(Uri uri,
...
             if (InstanceColumns.CONTENT_ITEM_TYPE.equals(cr.getType(uri))) {
-                // chain back to the Form record...
-                String[] selectionArgs = null;
-                String selection = null;
-                Cursor instanceCursor = null;
-                try {
-                    instanceCursor = cr.query(uri, null, null, null, null);
+                String[] selectionArgs;
+                String selection = FormsColumns.DELETED_DATE + " IS NULL AND " + FormsColumns.JR_FORM_ID + " =? AND ";
+                try (Cursor instanceCursor = cr.query(uri, null, null, null, null)) {
```
lognaturel, line 277: I really wanted to update all this to use the instance/form repositories but it didn't feel worth the effort and risk at this time. I tried doing it for just this branch but there's a lot of code below that relies on having a `formCursor` so it becomes a lot to change.

A lot of this code is really odd so hopefully we'll have a good context for revisiting it soon. For starters, it seems unexpected that every form goes into `EncryptionUtils` on finalization even if it doesn't specify encryption. Feels like that check should happen earlier.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -594,46 +597,22 @@ private void loadFromIntent(Intent intent) {
...
+                        formInfo.getFormId())
+                                + ((formInfo.getFormVersion() == null) ? ""
+                                : "\n" + getString(R.string.version) + " " + formInfo.getFormVersion()),
                         EXIT);
                 return;
-            } else {
-                formPath = FormsDaoHelper.getFormPath(selection, selectionArgs);
-
-                /**
-                 * Still take the first entry, but warn that there are multiple rows. User will
```
lognaturel, line 629: This just wasn't true. The dialog errored out and quit the form so this wasn't a warning and the first entry was never used. Also it's no longer true that the SQLite database needs to be touched directly.

collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
```diff
@@ -271,40 +272,32 @@ public static EncryptedFormInformation getEncryptedFormInformation(Uri uri,
...
+                        selection += FormsColumns.JR_VERSION + " IS NULL";
                     }
                 }
 
                 formCursor = new FormsDao().getFormsCursor(selection, selectionArgs);
 
-                if (formCursor.getCount() != 1) {
+                // OK to finalize with form definition that was soft-deleted. OK if there are multiple
+                // forms with the same formid/version as long as only one is active (not deleted).
+                if (formCursor.getCount() == 0 || new DatabaseFormsRepository().getAllNotDeleted(formId, formVersion).size() > 1) {
```
lognaturel: Creating a new FormsRepository here locally felt much better than trying to juggle more raw cursors.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -594,46 +597,22 @@ private void loadFromIntent(Intent intent) {
...
-                 * need to hand-edit the SQLite database to fix it.
-                 */
-                if (formCount > 1) {
-                    createErrorDialog(getString(R.string.survey_multiple_forms_error), EXIT);
-                    return;
-                }
+            } else if (candidateForms.stream().filter(f -> !f.isDeleted()).count() > 1) {
```
seadowg, line 610: So we can't just use `FormsRepository#getAllNotDeleted` because we want a different error message here?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -594,46 +597,22 @@ private void loadFromIntent(Intent intent) {
             }
 
             instancePath = formInfo.getInstancePath();
+            List<Form> candidateForms = formsRepository.getAll(formInfo.getFormId(), formInfo.getFormVersion());
```
seadowg: Oh wow it feels like this is a big improvement. I'm excited to get this stuff out of the Activity...

collect_app/src/main/java/org/odk/collect/android/forms/FormsRepository.java
```diff
@@ -12,6 +12,8 @@
 
     List<Form> getAll();
 
+    List<Form> getAll(String jrFormId, @Nullable String jrVersion);
```
seadowg: Hmmm I think we probably want to be more consistent here and make this a `getBy...` that returns a list. Maybe the methods that return a single item should be `getOne...` or `findOne...`?

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFormsRepository.java
seadowg: I think we should add tests for these new repo methods so we then know the fake and real implementation have consistent behaviours.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
seadowg: I think I'm happy not localizing the method and with the text but I think it'd be good to separate it out as a display concern here. Could we maybe add a type to `FormDownloadException` (or a subtype) that we can use to match on and then display the right message in more UI focused code?

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
```diff
@@ -283,6 +284,59 @@ public void whenFormIsSoftDeleted_unDeletesForm() throws Exception {
...
+        try {
+            downloader.downloadForm(serverFormDetails, null, null);
+            fail("Expected exception");
+        } catch (FormDownloadException e) {
+            assertThat(e.getMessage(), notNullValue());
+        }
+    }
+
+    @Test
+    public void whenFormWithSameIdAndVersion_andSameContentsOnDevice_downloadsSuccessfully() throws Exception {
```
seadowg: Don't want to be too nitpicky about test names but just because this stuff is all kinda confusing I'm wondering if it should be `_replacesForm` or `_doesNotDuplicateForm`?

collect_app/src/main/java/org/odk/collect/android/forms/FormsRepository.java
lognaturel: `getAllByXYZ`, `getOneByXYZ`, `getPredicateByXYZ`

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: (@seadowg and I talked about this but answering for posterity and to double check my thinking)

That's right -- the first error case is that there are no forms at all (deleted or not) that match. The second error case is if there are more than one **non-deleted** that match the formid/version. Basically:
- we're happy to use a soft-deleted form if that's the only one that matches
- we're happy to have formid/version matches as long as exactly one is non-deleted

So initially we need to fetch all forms including soft-deleted ones. Then I could have gone back to the db with getAllNotDeleted but filtering the list seemed simpler/more performant.

collect_app/src/main/java/org/odk/collect/android/forms/FormsRepository.java
lognaturel: Thanks, this was a big improvement. Don't hesitate if you want to see more changes here to either add a commit or ask me. I think having a good interface will pay dividends, it was getting confusing in there.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
lognaturel: I find naming really important so not nitpicking at all!

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFormsRepository.java
lognaturel: Thanks for the nudge, that's an improvement.

collect_app/src/main/java/org/odk/collect/android/instances/Instance.java
```diff
@@ -91,7 +91,7 @@ public Builder canEditWhenComplete(boolean canEditWhenComplete) {
         }
 
         public Builder instanceFilePath(String instanceFilePath) {
-            this.instanceFilePath = new StoragePathProvider().getInstanceDbPath(instanceFilePath);
+            this.instanceFilePath = instanceFilePath;
```
lognaturel, line 94: This scares me a little bit. It's not in use by Collect code other than in tests and in the method that builds an `Instance` from a DB record which means the path is already sure to be the correct one. So I think it's safe. But we have to be really careful if we do use it that we pass in the correct path.

This mirrors what `Form` does. I suspect that this had to be changed by @seadowg to test the `FormsRepository`. Without this change, I always got a crash from this method call when trying to test `InstancesRepository`. This was not easy to hunt down.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoUpdateTaskSpec.java
```diff
@@ -86,7 +87,7 @@
                                         formDownloader.downloadForm(serverFormDetails, null, null);
                                         results.put(serverFormDetails, TranslationHandler.getString(context, R.string.success));
                                     } catch (FormDownloadException e) {
-                                        results.put(serverFormDetails, TranslationHandler.getString(context, R.string.failure));
+                                        results.put(serverFormDetails, new FormDownloadExceptionMapper(context).getMessage(e));
```
lognaturel, line 90: Note: there isn't an easy way to show this message in "match exactly" mode as far as I can tell because the `FormDownloadException` gets converted to a generic `FormSourceException`. I think those two should probably be consolidated at some point but now didn't feel like the right time.

collect_app/src/test/java/org/odk/collect/android/support/FormUtils.java
```diff
@@ -28,12 +28,12 @@ public static String createXForm(String id, String version) {
                 "</h:html>";
     }
 
-    public static Form.Builder buildForm(long id, String jrFormId, String jrVersion, String formFilesPath) {
-        return buildForm(id, jrFormId, jrVersion, formFilesPath, "blah");
+    public static Form.Builder buildForm(long id, String formId, String version, String formFilesPath) {
```
lognaturel, line 31: I have no idea why these values ended up being named `jrFormId` and `jrVersion`. The irony is that JavaRosa actually has no concept of them, they are only meaningful at the Collect level. They're clearly an ODK form spec thing at least at this point and the `jr` prefix is confusing.

collect_app/src/test/java/org/odk/collect/android/support/FormUtils.java
```diff
@@ -28,12 +28,12 @@ public static String createXForm(String id, String version) {
...
+        return buildForm(id, formId, version, formFilesPath, "blah");
     }
 
-    public static Form.Builder buildForm(long id, String jrFormId, String jrVersion, String formFilesPath, String xform) {
-        String fileName = jrFormId + "-" + jrVersion;
+    public static Form.Builder buildForm(long id, String formId, String version, String formFilesPath, String xform) {
+        String fileName = formId + "-" + version + "-" + Math.random();
```
lognaturel, line 36: This is so that we can represent the case where there are two forms with the same id and version.

lognaturel:
Only the last 4 commits have actually changed. The rest are updated because I rebased on `master`.

Unfortunately I'm going to need some help. There are two test failures that I can't reproduce locally and am out of ideas for. One of the new Robolectric tests I introduced for `InstancesRepository` [fails on CI with `java.lang.IllegalStateException: Illegal connection pointer 131`](https://app.circleci.com/pipelines/github/lognaturel/collect/121/workflows/e8c49eec-c6b8-4d1c-be71-3292812d0415/jobs/1524). It looks like some kind of pollution but I can't figure it out. I looked carefully at the corresponding `FormsRepository` implementation and tests and can't find meaningful differences.

The [automatic storage migration Espresso test fails](https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/8492847393201930510/executions/bs.7773b2eb496d1ccf/testcases/155/test-cases) systematically because the migration fails. I can't reproduce locally. I even manually installed v1.28.4 and then updated to this branch and the migration passed. This one I have absolutely no idea about. From the log, I believe the failure is:
```
12-03 16:40:29.568: E/SQLiteDatabase(6654): Failed to open database '/storage/emulated/0/Android/data/org.odk.collect.android/files/metadata/forms.db'.
12-03 16:40:29.568: E/SQLiteDatabase(6654): android.database.sqlite.SQLiteCantOpenDatabaseException: unknown error (code 14 SQLITE_CANTOPEN): Could not open database
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnection.nativeOpen(Native Method)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:211)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:195)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnectionPool.openConnectionLocked(SQLiteConnectionPool.java:503)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:204)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:196)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteDatabase.openInner(SQLiteDatabase.java:880)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteDatabase.open(SQLiteDatabase.java:865)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:766)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:714)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.database.FormsDatabaseHelper.databaseNeedsUpgrade(FormsDatabaseHelper.java:97)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.provider.FormsProvider.getDbHelper(FormsProvider.java:68)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.provider.FormsProvider.query(FormsProvider.java:119)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentProvider.query(ContentProvider.java:1055)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentProvider.query(ContentProvider.java:1147)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentProvider$Transport.query(ContentProvider.java:240)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentResolver.query(ContentResolver.java:802)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentResolver.query(ContentResolver.java:752)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at android.content.ContentResolver.query(ContentResolver.java:710)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.dao.FormsDao.getFormsCursor(FormsDao.java:53)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.dao.FormsDao.getFormsCursor(FormsDao.java:45)
12-03 16:40:29.568: E/SQLiteDatabase(6654): 	at org.odk.collect.android.storage.migration.StorageMigrator.migrateFormsDatabase(StorageMigrator.java:148)
```

I don't think any of the classes involved have been changed. I wonder whether it could also be a pollution issue of some kind. I don't see any changes to Espresso tests that would explain it, though.

seadowg:
Ok this seems to be happy now. I'm going to take 15 mins and see if I can extract the DB helpers that are causing problems to Dagger so it can manage them rather than using `static`. This would mean we automatically get new ones for every test.

seadowg:
Ok it looks like the above idea is going to be a little hard because of the storage migration stuff relying on being able to reinitialize the DB helper instance. I think we're good with the clearing in our Robolectric application fro the moment.

The underlying issue is discussed here: https://github.com/robolectric/robolectric/issues/2387.

I'll have a quick look  at the Espresso stuff later to see if anything rings a bell there.

seadowg (approved):

mmarciniak90:
@lognaturel, please take a look at this case:

1. I downloaded 2 forms from Aggregate.
2. Then I added media file to `All widgets` form on Aggregate server.
3. I opened `Get blank Form` page and I tried to download again the same forms.
4. Downloading was blocked. 

> try to download form with same formid/version and same filename. Collect should download and replace. There may or may not be new media attachments.

We can say that blocking the download of a form without any changes may be OK because exactly the same file exists. However, blocking the download in the case where only media file has been added, and update information is even available, looks like a bug.

ezgif com-video-to-gif(1)

lognaturel:
Thanks, @mmarciniak90! ~That's unexpected. I'm guessing Aggregate modifies the form file in that case in which case I don't think we should prevent downloads with the same formid/version, sadly.~ I just remembered that Aggregate makes you upload the form definition again to identify what form media belongs to. And then it changes the form definition contents on upload by adding an uploaded timestamp. That's really annoying and indeed, I think that means we have no way of differentiating between intentional and unintentional forms with same formid/version.
 
Where are you at with testing this otherwise? It's risky so the longer we can get it in users' hands in beta, the better. I'm tempted to remove the commit preventing download of forms with same formid/version and doing a beta.

lognaturel:
Ugggh. No, the problem is that I'm comparing the local and remote hash and only one has an `md5:` prefix so they're never the same.

lognaturel:
Thanks, @mmarciniak90. I have fixed that case and done various checks around it. It should now be possible to download the exact same form again as well as the same form with new media added.

mmarciniak90:
Thanks for quick fix! 

> Where are you at with testing this otherwise? 

As you mentioned, these changes are quite risky and after today's testing we are still not sure about all cases.
Hopefully we will complete it tomorrow.

lognaturel:
> Hopefully we will complete it tomorrow.

How about you finish it on master so it can be part of the beta?

mmarciniak90:
> How about you finish it on master so it can be part of the beta?

Sure!

mmarciniak90:
**Tested with success**

Results for cases mentioned in the description
* TC1: confirmed behavior. I was not able to open the filled submission
* TC2: 2 forms are visible in the Delete Saved Forms list. Toast and unfinalized submission is a consequence of this behavior.
* TC3: Toast and unfinalized submission is a consequence of this behavior.
* TC4: Identical form is downloaded and replaced
* TC5: Download is possible. Toast and unfinalized submission is a consequence of this behavior.
* TC6: Download is not possible
* TC7: Soft-delete occurred for the form
* TC8: After update -> Download is not possible, the problem is no longer visible
* TC9: After update -> Download is not possible when the first form is deleted, problem is no longer visible

lognaturel:
🙏 And I want to make sure that you saw I ended up backing out of blocking download for now.  https://github.com/getodk/collect/pull/4281

kkrawczyk123:
Verified on Androids 5.1, 7.1 and 9.0

PR #4270, opened 2020-12-04
Fix building select widgets

grzesiek2010:
I used the form mentioned below and added automated tests.

It's a follow up pr which fixes rest of the cases to allow mixing different select appearances. Cases that this pr fixes are marked in yellow https://docs.google.com/spreadsheets/d/1uA6NeUbE51WACiKUjErFUxYFeJdGGWglNGKivuzei7I/edit?usp=sharing

It fixes cases that were marked in red in the form mentioned below. I find this fix pretty safe so using the attached form for testing would be enough.

https://docs.google.com/spreadsheets/d/1uA6NeUbE51WACiKUjErFUxYFeJdGGWglNGKivuzei7I/edit?usp=sharing

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
@@ -205,13 +205,13 @@ public class WidgetFactory {
                 // This means normal appearances should be put before search().
                 if (appearance.contains(WidgetAppearanceUtils.MINIMAL)) {
                     questionWidget = new SelectMultiMinimalWidget(context, questionDetails, waitingForDataRegistry);
-                } else if (appearance.startsWith(WidgetAppearanceUtils.LIST_NO_LABEL)) {
+                } else if (appearance.contains(WidgetAppearanceUtils.LIST_NO_LABEL)) {
                     questionWidget = new ListMultiWidget(context, questionDetails, false);
-                } else if (appearance.startsWith(WidgetAppearanceUtils.LIST)) {
+                } else if (appearance.contains(WidgetAppearanceUtils.LIST)) {
                     questionWidget = new ListMultiWidget(context, questionDetails, true);
-                } else if (appearance.startsWith(WidgetAppearanceUtils.LABEL)) {
+                } else if (appearance.contains(WidgetAppearanceUtils.LABEL)) {
                     questionWidget = new LabelWidget(context, questionDetails);
-                } else if (appearance.startsWith(WidgetAppearanceUtils.IMAGE_MAP)) {
+                } else if (appearance.contains(WidgetAppearanceUtils.IMAGE_MAP)) {
                     questionWidget = new SelectMultiImageMapWidget(context, questionDetails);
                 } else {
                     questionWidget = new SelectMultiWidget(context, questionDetails);
@@ -270,7 +270,7 @@ public class WidgetFactory {
             questionWidget = new ListWidget(context, questionDetails, false, isQuick);
         } else if (appearance.contains(WidgetAppearanceUtils.LIST)) {
             questionWidget = new ListWidget(context, questionDetails, true, isQuick);
-        } else if (appearance.equals(WidgetAppearanceUtils.LABEL)) {
+        } else if (appearance.contains(WidgetAppearanceUtils.LABEL)) {
             questionWidget = new LabelWidget(context, questionDetails);
         } else if (appearance.contains(WidgetAppearanceUtils.IMAGE_MAP)) {
             questionWidget = new SelectOneImageMapWidget(context, questionDetails, isQuick);
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
@@ -202,7 +202,6 @@ public class WidgetFactory {
             case Constants.CONTROL_SELECT_MULTI:
                 // search() appearance/function (not part of XForms spec) added by SurveyCTO gets
                 // considered in each widget by calls to ExternalDataUtil.getSearchXPathExpression.
-                // This means normal appearances should be put before search().
                 if (appearance.contains(WidgetAppearanceUtils.MINIMAL)) {
                     questionWidget = new SelectMultiMinimalWidget(context, questionDetails, waitingForDataRegistry);
                 } else if (appearance.contains(WidgetAppearanceUtils.LIST_NO_LABEL)) {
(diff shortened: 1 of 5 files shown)


grzesiek2010:
@lognaturel looks like it's the last step and we will be able to mix all select appearances once we merge this pr. Would be good if you could review that form maybe to make sure I didn't miss anything. Maybe it's even worth going though this form during release testing?

collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
```diff
@@ -205,13 +205,13 @@ public QuestionWidget createWidgetFromPrompt(FormEntryPrompt prompt) {
                 // This means normal appearances should be put before search().
```
lognaturel: This comment should be removed. The documentation at https://xlsform.org/en/#dynamic-selects-from-pre-loaded-data also states that the `search()` appearance needs to come last and that can be updated.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
```diff
@@ -0,0 +1,345 @@
...
+                waitingForDataRegistry,
+                questionMediaManager,
+                null,
+                null,
+                activityAvailability,
+                null);
+    }
+
+    @Test
+    public void testCreatingSelectOneMinimalWidgetWithFastExternalChoices() {
```
lognaturel: This is a pretty bad case with two data sources. External itemsets "win" because they're first? Since this test is not about the two data source weirdness, I would strongly suggest removing it and just having either an external select or a search() select.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
```diff
@@ -0,0 +1,345 @@
...
+    @Test
+    public void testCreatingLikertWidgetWithFastExternalChoices() {
+        QuestionDef questionDef = mock(QuestionDef.class);
+        when(questionDef.getAdditionalAttribute(null, "query")).thenReturn("instance('counties')/root/item[state= /data/state ]");
+
+        FormEntryPrompt prompt = new MockFormEntryPromptBuilder()
+                .withControlType(Constants.CONTROL_INPUT)
+                .withDataType(Constants.DATATYPE_TEXT)
+                .withQuestion(questionDef)
+                .withAppearance("search(‘items’) LisT lIkErt something")
```
lognaturel: Same comment as above. Why use two external data features? If this is about verifying that appearances after `search()` are honored, no need to also have it be an external select.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
```diff
@@ -0,0 +1,345 @@
...
+    @Test
+    public void testCreatingSelectOneListNoLabelWidgetWithFastExternalChoices() {
+        QuestionDef questionDef = mock(QuestionDef.class);
+        when(questionDef.getAdditionalAttribute(null, "query")).thenReturn("instance('counties')/root/item[state= /data/state ]");
+
+        FormEntryPrompt prompt = new MockFormEntryPromptBuilder()
+                .withControlType(Constants.CONTROL_INPUT)
+                .withDataType(Constants.DATATYPE_TEXT)
+                .withQuestion(questionDef)
+                .withAppearance("search(‘items’) LisT-nOLabeL something")
```
lognaturel: These are really impractical appearance combinations. If you have a big enough dataset to user an external data file, why would you be building a grid?

lognaturel (changes requested):
I'm not really sure how practical the newly-supported combinations actually are but seems fine to offer some support for them where possible!

Could the tests be restructured to focus on cases? For example:
* `appearancesAfterSearchAppearanceFunction_areUsed`: `search('items') image-map` checks that an `image-map` widget is created. 
* `appearanceBeforeSearchAppearanceFunction_areUsed`: `image-map search('items')` checks that an `image-map` widget is created.
* `appearanceCase_isIgnored`: `miNimal` checks that a `minimal widget is created
* `unknownAppearances_areIgnored`: `minimal something` checks that a `minimal` widget is created

It's not worth trying to write exhaustive tests and I think focusing on a few common patterns that are important would be enough.

collect_app/src/main/java/org/odk/collect/android/widgets/WidgetFactory.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
grzesiek2010: I removed this tests since it was redundant.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
grzesiek2010: Same here this test is removed.

collect_app/src/test/java/org/odk/collect/android/widgets/WidgetFactoryTest.java
grzesiek2010: Same here this test is removed.

grzesiek2010:
I removed some redundant tests. When it comes to:
> Could the tests be restructured to focus on cases? For example:
appearancesAfterSearchAppearanceFunction_areUsed: search('items') image-map checks that an image-map widget is created.
appearanceBeforeSearchAppearanceFunction_areUsed: image-map search('items') checks that an image-map widget is created.
appearanceCase_isIgnored: miNimal checks that a `minimal widget is created
unknownAppearances_areIgnored: minimal something checks that a minimal widget is created

I don't think it's a good solution here because I don't need to tests cases like appearance before/after I just put the appearance I want to use in the middle (something like: `something minimal something`) in tests and it should be enough if such a test passes.
More important is to verify all available select widgets what I did.

lognaturel (approved):
Great, those tests look much better to me!

mmarciniak90:
@grzesiek2010 we finished testing this large form and we noticed that one question is marked as fixed in 4270 but we don't see the difference. Maybe this question was marked by mistake. Please take a look.

Screenshot from 2021-01-18 09-51-35

Verified cases:
- fixes for marked widgets in the attached sheet
- regression check to all questions in the attached form
- created and verified a shorter form that contains only fixed widgets
- remove response
- rotation
- light/dark mode
- English/Spanish language

grzesiek2010:
> we noticed that one question is marked as fixed in 4270 but we don't see the difference. Maybe this question was marked by mistake

you can't see any difference between master and my branch but it works as expected right? 
- Items are loaded from a csv file
- columns-pack appearance is used 
- no-buttons appearance is used
- autocomplete appearance is used and the searchbox is displayed
- quick appearance is ignored since it's select_multiple question

probably this case has been already fixed on master as a part of something else.

mmarciniak90:
> probably this case has been already fixed on master as a part of something else.

I checked it more closely and the same appearance is in version 1.29.0 so it was fixed before.

mmarciniak90:

grzesiek2010:
@lognaturel 
should we maybe get back to that form form time to time to confirm that our selects work as expected since we support so many combinations? During regression testing for example?

PR #4271, opened 2020-12-04
Audio recording design polish

seadowg:
This adds some tweaks made to the audio recording/playback features in form entry during design pairing. I've deleted the usual questions as pretty much every change is subjective.


collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
```diff
@@ -53,11 +55,13 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
             } else if (session.getFile() == null) {
                 binding.getRoot().setVisibility(VISIBLE);
 
+                binding.timeCode.setText(LengthFormatterKt.formatLength(session.getDuration()));
+
                 if (session.getPaused()) {
-                    binding.pauseRecording.setText(R.string.resume_recording);
+                    binding.pauseRecording.setIcon(ContextCompat.getDrawable(getContext(), R.drawable.ic_play_arrow_24dp));
```
lognaturel: I think this should be a red record circle.

seadowg: Nice touch. I like that!

seadowg: Actually after playing with this it makes it feel like we're recording while paused as there is a big red recording icon. Lets not change this at the moment.

lognaturel: I don’t think play is at all clear, though. Does it indicate ongoing playback? Playing back what you just recorded before stopping? Even knowing that it was supposed to be resume I wasn’t sure that’s what it would do. 

What do other recording apps do?  Maybe a blue or black circle? Or maybe the scale of the button is the challenge?

lognaturel:
I think it's important for the "resume recording" button to be evidently for recording before the beta goes out. Beyond that, the details can certainly be worked on over the next few days.

Here are a couple of concepts:
 

On the left, pause/record is red because that's the typical convention for recording. The challenge with that is that it makes it look dangerous when in fact it's the safer option.

The option on the right is what I currently prefer. Making the two buttons the same color gives them the same prominence which I think is what we want to aim for. The downside is that the record circle is not as recognizable without the red color. I think this is mitigated by the fact that the only way to see it is to first have paused recording. I don't understand why currently pause/record is gray and stop is blue.

One thing not being communicated or guarded against is that stop is the dangerous/final action. I could see an argument for making it red or for showing a confirmation dialog when it's tapped. Did you also consider the order of the two buttons? I think the outside button is easier to hit so perhaps putting the safe option there would be better. I'm guessing the current logic is that stop is the final action so it's furthest to the right.

seadowg:
> I don't understand why currently pause/record is gray and stop is blue.

The thinking here was to give them different precedence. Personally I'm happy with making them the same color but I'd like to pull @Ig-Rebollo in here as well. 

> One thing not being communicated or guarded against is that stop is the dangerous/final action.

Hmmm good point. There is a twist on the UI (now we have pause) where we only let the user stop once they've paused (maybe making stop more a "save" action). Again interested if @Ig-Rebollo has thoughts on the problem here.

ignatiusgr:
Hi @lognaturel @seadowg 

I suggested the blue/grey combination as it seems you are using only one colour (blue) with a greyscale palette. Usually the logic with that approach (which can be pretty effective) is to use the colour of choice to indicate the key action in each screen, and different shades of grey for all the other secondary or supporting actions. I can absolutely understand the point of safer vs 'dangerous' options and wanting to avoid mistakes, but from the logic of 'key action', stop(+ save) is ultimately the essential action, while pausing is more of a supporting one, hence the grey. 

Bear in mind that the default situation would be to start recording and then simply stop recording. In most cases users will not need to use the 'pause' option, and placing too much emphasis on it might lead to the safer but 'annoying' outcome of wanting to end the recording and yet keep pausing it.

Using a red/blue combination certainly reinforces the perception of safe vs dangerous options, but I'm not sure to what extent that is an accurate representation of the alternatives. Also, I still have to find solid evidence to support this, but my impression is that placing a button on the edge of a screen usually makes it slightly harder to tap when both holding and tapping with the thumb of the same hand, which certainly is a very common way to interact with these apps.

To mitigate potential errors, specially given how the circle indicating recording might be confusing, you could:
1. Re-incorporate the text (pause, resume, stop)
2. When tapping on 'stop', display a window asking for confirmation. Something along the lines: You are still recording. Are you sure you want to stop and save the recording, or do you want to continue recording? CONTINUE / STOP & SAVE. It would also be slightly annoying, but clearer.
3. Use (in addition to point 1 and/or 2) the same colours but make one filled with colour and the pause button an outline (similar to the delete button in the next screen, but blue). The [ODK site](https://getodk.org/) uses a very similar approach to this in the header (Contact/Demo)
4. Increase the size of the buttons to gain more attention. I actually did some additional research after the call with Callum and came across [this article](https://www.nngroup.com/articles/touch-target-size/). Essentially, the impact area of the typical thumb can be an average of 2.5cm (1 inch) wide. The recommendation is to use buttons of at least 1x1 cm of physical size, ideally larger. The translation would be around 50 or 60 px height to be on the safe side. It might also be worth considering separating them a bit more from each other, to avoid any potential tapping on the wrong button.

Let me know what you think. Of course these are only suggestions, I think your current concepts also work well enough already. These ideas may simply help making everything even clearer.

lognaturel:
Thanks, @Ig-Rebollo, really helpful to get some of your reasoning! Ultimately, the only thing I feel really strongly about is that record should not be represented by a play icon. That is unnecessarily confusing and I think really detracts from the interface. Imagine training people and saying "to resume recording, tap the play button."

> Re-incorporate the text (pause, resume, stop)

If we had a single-language interface, that might be ok (though it would wrap on some devices) but given that we have translations in many languages that have longer words for these actions, I don't think it's possible without a rethink of how space is used. Maybe one of them could have text but both won't fit.

ignatiusgr:
Absolutely, I agree that the play icon might not be the best in this context. Two alternatives:
1. You can probably use the [circle icon plus REC](https://image.shutterstock.com/image-vector/recording-sign-icon-red-logo-260nw-1586272831.jpg). and the stop icon with 'STOP', without necessarily having to translate to different languages, as those are already pretty widely used internationally. Being both so short, they would probably fit together. As per the pause button, you could simply use the icon with no text, as there's no confusion about what that icon means.
2. Use the icon only approach, and replace the play button for [this icon](https://www.pinclipart.com/picdir/middle/147-1471718_voice-icon-free-icons-record-icon-png-clipart.png)

In both cases the background colour vs outline styles would also be very effective, or alternatively, the ODK header approach I shared before.

lognaturel:
Thanks, @Ig-Rebollo, @seadowg! I'm liking the mic icon for record. The outline button also makes a lot more sense to me. I think this is great for the beta and we can keep chewing on some of @Ig-Rebollo's ideas to see if we want to incorporate any others.


Screen Shot 2020-12-07 at 11 50 13 AM

lognaturel (approved):

PR #4272, opened 2020-12-04
Improve remote logging

lognaturel:
First commit targets [this error](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/ddfe2f07d7b45192c590ee139c341b64?time=last-ninety-days&sessionEventKey=5FCA8F0B003D00012AAA09CA530E9285_1480954048439092828)

Second commit targets [this error](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/2b60a44dc09ff4f824495b6e82bb1758?time=last-ninety-days&sessionEventKey=5FCA930D033A00013CDB6E7D9164C555_1480942585699294808)

Third commit is a little bit of cleanup. There's a lot more to do but at least this is a start around where I was anyway.

Compiled and ran.

No big alternatives here.

No changes intended. There's a bit of refactoring in the task that loads a blank form and the code specifically that loads a form from cache.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FormDefCache.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FormDefCache.java
@@ -80,20 +80,21 @@ public class FormDefCache {
     public static FormDef readCache(File formXml) {
         final File cachedForm = getCacheFile(formXml);
         if (cachedForm.exists()) {
-            Timber.i("Attempting to load %s from cached file: %s.",
-                    formXml.getName(), cachedForm.getName());
+            Timber.i("Attempting to load %s from cached file: %s.", formXml.getName(), cachedForm.getName());
             final long start = System.currentTimeMillis();
-            final FormDef deserializedFormDef = deserializeFormDef(cachedForm);
-            if (deserializedFormDef != null) {
-                Timber.i("Loaded in %.3f seconds.", (System.currentTimeMillis() - start) / 1000F);
-                return deserializedFormDef;
-            }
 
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/FormDefCache.java
```diff
@@ -108,17 +109,11 @@ private static File getCacheFile(File formXml) {
                 FileUtils.getMd5Hash(formXml) + ".formdef");
     }
 
-    private static FormDef deserializeFormDef(File serializedFormDef) {
+    private static FormDef deserializeFormDef(File serializedFormDef) throws Exception {
```
grzesiek2010, line 112: Do we need readCache() and deserializeFormDef() looks like the same code is duplicated in javarosa `org.javarosa.benchmarks.utils.FormDefCache`?

lognaturel: Yes, this is client code. The package you point to is for benchmarks and that code was copied from Collect. You're right that the duplication is weird and I wonder whether we should rework the benchmark a bit on the JR side. I don't think there are changes to make here for now, though.

lognaturel: Well, I don't know. Maybe some of it should move down to JR. Seems like a consideration for another time, though!

grzesiek2010 (approved):

PR #4274, opened 2020-12-07
Fixed creating multiple files for recording selfie videos

### LINKED ISSUE #4268: Unexpected copy of selfie video is visible in main ODK directory if a form is filled after data migration

#### Problem description
Unexpected copy of selfie video is visible in main ODK directory if a form is filled after data migration

1. ODK data are migrated to internal storage
2. User fills a form with **Selfie video** widget
3. User records selfie video and finishes form

#### Actual result
* Copy of selfie video is visible in main ODK directory 
* Removing submission does not clear the copy of selfie video
* Uninstalling does not remove the copy of selfie video - this behavior does not always occur, the condition is undefined

Screenshot_20201204-135312

#### Expected behavior
Copy of selfie video should not be created

#### Other information 
The issue is not visible if data are not migrated

grzesiek2010:
Closes #4268 

I tested the fix manually.

The problem was that we created a new tmp file for every recording https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4268?expand=1#diff-f63fc4d1f7f735577c9fb1d0cb030deff5c0c4693005671cd7352cdf8a359a85L129 which we didn't remove. I decided to use the same solution we have for image files where we create a reusable tmp file in cache directory https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4268?expand=1#diff-305ec7a913eed2277d0af06a536383d929631be636da49be48e21df4bfb625d4R87

With the fix the app should use a reusable tmp file for recording selfie videos instead of creating a new file in the app directory.
It's a safe and isolate change so we can test just the behavior described in the issue.

I also implemented code improvements touching the DrawWidget so please make sure answers are saved witout regression (no need to test changing colors or something just saving output images)

Any form with selfie video widget like the demo All widgets form.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
@@ -79,11 +79,11 @@ public class StoragePathProvider {
                 : getUnscopedStorageRootDirPath();
     }
 
-    public String getTmpFilePath() {
+    public String getTmpImageFilePath() {
         return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.jpg";
     }
 
-    public String getTmpVideoPath() {
+    public String getTmpVideoFilePath() {
         return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.mp4";
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
@@ -87,6 +87,10 @@ public class StoragePathProvider {
         return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmpDraw.jpg";
     }
 
+    public String getTmpVideoPath() {
+        return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.mp4";
+    }
+
     public String getInstanceDbPath(String filePath) {
         return getDbPath(getDirPath(StorageSubdirectory.INSTANCES), filePath);
     }
--- a/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
@@ -83,10 +83,6 @@ public class StoragePathProvider {
         return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.jpg";
     }
 
-    public String getTmpDrawFilePath() {
-        return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmpDraw.jpg";
-    }
-
     public String getTmpVideoPath() {
         return getDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.mp4";
     }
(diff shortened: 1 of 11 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/CameraUtils.java
```diff
@@ -123,10 +124,8 @@ public static void savePhoto(String path, byte[] data) {
...
         }
     }
-  
-    public static String getVideoFilePath(Context context) {
-        final File dir = context.getExternalFilesDir(null);
-        return (dir == null ? "" : (dir.getAbsolutePath() + "/"))
-                + System.currentTimeMillis() + ".mp4";
+
+    public static String getVideoFilePath() {
+        return new StoragePathProvider().getTmpVideoPath();
```
lognaturel: If you're going to be creating a new `StoragePathProvider` here anyway I think you can do away with the indirection and just use `new StoragePathProvider().getTmpVideoPath()` inline.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
```diff
@@ -79,11 +79,11 @@ public String getStorageRootDirPath() {
                 : getUnscopedStorageRootDirPath();
     }
 
-    public String getTmpFilePath() {
+    public String getTmpImageFilePath() {
```
lognaturel, line 82: This is an existing issue but still want to highlight it. I don't think we know the image is a `jpg`. It could be a `png`, `bmp`, ... We should fix this at some point. Maybe a separate issue?

lognaturel (changes requested):
Nice. One other method to take out of `CameraUtils` and another issue to file, I think.

For QA, please make sure it's clear from the PR description that this also changes the drawing widgets.

collect_app/src/main/java/org/odk/collect/android/utilities/CameraUtils.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
grzesiek2010: You are right I'll file an issue.

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
grzesiek2010: Hmm unfortunately I think it's not possible... we pass our tmp file starting a camera app because it's a way to capture a full-size photo:
https://developer.android.com/training/camera/photobasics#TaskPath

we don't know what format such an external app is going to return so like in the documentation we use jpg. Then the captured photo is saved to our tmp file.

lognaturel: I see. Can we check the mime type and fix the extension later? I think all images are submitted with a jpg extension but maybe I’m remembering wrong.

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- save form
- finalize
- edit submission
- change video
- change video and ignore changes 
- remove submission
- regression checks on: image, audio, draw, signature, annotate, video, selfie, file widgets

mmarciniak90:

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
grzesiek2010: No really, I mean we can read mime type using the tmp file where an external app saves captured photos but it's pointless because it will always be jpeg.

PR #4281, opened 2020-12-08
Always allow download of form with same formid/version

### LINKED ISSUE #4276: Blocking download of forms with same formid/version doesn't work with Central draft testing
#4261 prevents download of form definitions with the same formid/version as an existing form on the device. However, when testing a Central draft this may be necessary. Currently, Collect has no way of knowing that we're dealing with a draft situation.

A few options:
- allow downloading forms with same formid/version
- allow downloading forms with same formid/version if there are no instances. I'm not sure this is all that helpful because it makes sense to fill out instances when testing
- allow downloading forms with same formid/version if `/draft` is in the path. This is pretty hacky.

lognaturel:
Reverts c59d6138dfbc37330aecc2816c1e4d4ecb11e085 and 0fabbdf766211f96e2e9c5f5582d37252e8719cc

Closes #4276

Iterated on a Central draft and verified revisions could be downloaded.

As I gave this more thought, I realized that we don't know enough about people's weird servers and workflows to prevent download of forms with same formid/version. I've left in analytics so we can find out how common it is. I think that to really solve all the problems around matching formid/version we should link instances to their forms through something truly unique like the database id or form hash. I added a note about this bad property of the system in state.md.

For now, reverting to the existing behavior feels like a pragmatic, safe approach. We've learned a lot through this experiment so I'm glad we thought through it but I think it's too aggressive for now.

This is intended to revert to the same behavior as in v1.28 and prior. It touches blank form download and that's where the regression risk is.

No.

No.


seadowg (approved):

kkrawczyk123:
**Tested with success!**

Tested cases:
- download the same form from Aggregate
- download the same form from Aggregate when submission fulfilled
- download the same form from Aggregate when media changed
- download the same form from Central
- download the same form from Central - draft version
- download the same form from Central when submission fulfilled
- download the same form from Google Drive
- download the same form from Google Drive when submission fulfilled

mmarciniak90:
Tested with success on Android versions: 7.0, 8.1, 10.0

lognaturel:
🎉

PR #4286, opened 2020-12-09
Handle process death during filling a form

### LINKED ISSUE #3873: Null formController object
#### Problem description
I have reproduced and solved many strange issues in Collect during past years and I even liked it but there is one bothering me since I remember, appearing in reports, being mentioned on the forum... it's null formController object.
Here is a sample crash report from v1.27.0:
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/aba5eacccf3e63937e507ae215d2a1fa?time=last-seven-days&sessionId=5ECCF34B00ED00016BE5AFD788389D67_DNE_2_v2

There are not many reports like that so the problem might seem not serious but it is. The reason why we don't have many reports like that one is the fact that in the past we didn't know how to fix the real cause so we used to add many null checks and now for example in FomrEntryActivity you can find a lot of them like:
```
if (formController != null) {
...
}
```
but it shouldn't be the case. If after loading a form we have that object (non null value) we shouldn't need to check it every time we use it.

I tried to reproduce the issue but to no avail so I had been waiting for a user that could reproduce it and describe and here you go: https://forum.getodk.org/t/backup-problem-at-the-end-of-the-form-v26-3/26564
Maybe the description is not perfectly detailed but still great taking into account it's rare.

The description that the mentioned user provided is different compared to what I had expected. He said the device is active and the problem occurs during filling the form. My assumption was rather that it happens when a user leaves a device with started form for a long time.

We keep formController object in the application class, not a good pattern but seems like we deal with it pretty well. I tried to learn more about how application class works hopping that maybe there is a magic case when such a class clears its data but as far as I know it's possible only when you leave your device for a long time (and the app is killed) what doesn't match the description our user provided and what I tested (How To Simulate The Application Being Killed http://www.developerphil.com/dont-store-data-in-the-application-object/) to no avail.

As I said we keep the formControler object in the application class using setFormController method to update it and according to my investigation we never set it to null I mean only here https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java#L531 temporarrely but then the form is loaded and we replace it with non-null object or close the form if it returns null formController object.

We could move the formController object to viewModel and keep it there but as long as I have no evidence it's cleared somehow because of the application class it would be just code refactoring not a fix.

I think it would be great to join forces and try to track down the root cause together once and for all @lognaturel @seadowg

### LINKED ISSUE #4250: Android process death might cause crashes if takes place during filling a form
Collect all versions

#### Problem description
Android process death might cause crashes if takes place during filling a form. We have been testing destroying activities using Don't keep activities option but this is something else we also should take into account. 

I was able to reproduce the issue in real life without using any developer options but that might be extremely difficult or even impossible for other developers/testers so here are steps you need to use to make sure you can reproduce the problem:
1. Go to Settings -> Developer Options -> Background process limit. Set this to "No background processes"
2. Open the demo form and navigate to the image widget.
3. Click take picture
4. Put the app into background and open any other app
5. Get back to the app and take picture
6. Try to navigate forward

the difference between the steps mentioned above and those I used in real life is that I didn't change anything in settings (step 1) and sometimes I had to perform step 4 and open more than one other app **but sometimes I didn't even have to put the app into background.**

#### Expected behavior
We should get back to the form and can continue working with it.

#### Other information 
Here is a good article that describes what's going on with android process death https://medium.com/inloopx/android-process-kill-and-the-big-implications-for-your-app-1ecbed4921cb

### LINKED ISSUE #4327: ViewModels sometimes have a null form controller even though it's set at the Application level
This is in v1.29.0 manifesting as crashes in `FormEntryViewModel.moveForward` ([Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/7793b2bc479705b76a05a38b55886a21?time=last-seven-days&versions=v1.29.0%20(4036)&sessionEventKey=5FDC7520002600015A7DDD7BA7D58223_1485970946505974935)), `FormEntryViewModel.moveBackward` ([Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8145577b551346217c3bbc8ad9b5392e?time=last-seven-days&versions=v1.29.0%20(4036)&sessionEventKey=5FDC79BB00F6000126ABDAD9E9AE3FE3_1485975707750444807)) and `FormEntryViewModel.openHierarchy` ([Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/00dbc308babc5ff267adc0e554377112?time=last-seven-days&versions=v1.29.0%20(4036)&sessionEventKey=5FDD1B7703A500011450D7A26E9130A6_1486153783798180265)). I believe I've seen other related logging in other non-fatal issues happening previously but I can't find it at the moment. I remember seeing an attempt to `getAuditEventLogger` from a null form controller.

grzesiek2010:
Closes #4250
Closes #4327
Closes #3873

I tested the fix manually.

It consists a couple of fixes and I can't come up with any better solution. generally our code was prepared for such cases but we have never tested process death so the code needed some improvements and fixes.

I hope it will fix the problem with null form controller that we have handled in many cases adding null checks.
Testing the issue in the way I described using the forms I mentioned would be enough.
Please additionally check the functionality with and without `don't keep activities` option. You can also test killing the app manually what you probably do more often)

**There is one thing**: during testing I noticed that sometimes we can get stuck in a state where savepoints are not saved/updated so if we add changes in a form and then kill the app those changes are lost. unfortunately I'm not able to reproduce it now so please keep an eye on it.

Any form with media files to reproduce the steps described in the issue, it might be the demo form.
Also it would be good to tests a form with multiple questions on one page, I used this one: 
[groupTestForm.txt](https://github.com/getodk/collect/files/5665530/groupTestForm.txt)
 
No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2315,6 +2315,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 if (pendingActivityResult) {
                     Timber.w("Calling onActivityResult from loadingComplete");
 
+                    formControllerAvailable(formController);
                     // set the current view to whatever group we were at...
                     onScreenRefresh();
                     // process the pending activity request...
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2317,9 +2317,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     Timber.w("Calling onActivityResult from loadingComplete");
 
                     formControllerAvailable(formController);
-                    // set the current view to whatever group we were at...
                     onScreenRefresh();
-                    // process the pending activity request...
                     onActivityResult(task.getRequestCode(), task.getResultCode(), task.getIntent());
                     return;
                 }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2645,7 +2645,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
     @Override
     public void widgetValueChanged(QuestionWidget changedWidget) {
-        FormController formController = Collect.getInstance().getFormController();
+        FormController formController = getFormController();
         if (formController == null) {
             // TODO: As usual, no idea if/how this is possible.
             return;
@@ -2688,7 +2688,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      */
     private void updateFieldListQuestions(FormIndex lastChangedIndex) throws FormDesignException {
         // Save the user-visible state for all questions in this field-list
-        FormEntryPrompt[] questionsBeforeSave = Collect.getInstance().getFormController().getQuestionPrompts();
+        FormEntryPrompt[] questionsBeforeSave = getFormController().getQuestionPrompts();
         List<ImmutableDisplayableQuestion> immutableQuestionsBeforeSave = new ArrayList<>();
         for (FormEntryPrompt questionBeforeSave : questionsBeforeSave) {
             immutableQuestionsBeforeSave.add(new ImmutableDisplayableQuestion(questionBeforeSave));
@@ -2696,7 +2696,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
         saveAnswersForCurrentScreen(questionsBeforeSave, immutableQuestionsBeforeSave);
 
-        FormEntryPrompt[] questionsAfterSave = Collect.getInstance().getFormController().getQuestionPrompts();
+        FormEntryPrompt[] questionsAfterSave = getFormController().getQuestionPrompts();
 
         Map<FormIndex, FormEntryPrompt> questionsAfterSaveByIndex = new HashMap<>();
         for (FormEntryPrompt question : questionsAfterSave) {
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -716,6 +716,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         }
     }
 
+    // This method may return null if called before form loading is finished
     @Nullable
     private FormController getFormController() {
         return Collect.getInstance().getFormController();
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -183,7 +183,7 @@ public class FormController {
      */
     public String getXPath(FormIndex index) {
         String value;
-        switch (getEvent()) {
+        switch (getEvent(index)) {
             case FormEntryController.EVENT_BEGINNING_OF_FORM:
                 value = "beginningOfForm";
                 break;
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
@@ -13,15 +13,12 @@ import org.javarosa.core.model.GroupDef;
 import org.javarosa.form.api.FormEntryController;
 import org.jetbrains.annotations.NotNull;
 import org.odk.collect.android.analytics.Analytics;
-import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.exception.JavaRosaException;
 import org.odk.collect.android.formentry.audit.AuditEvent;
 import org.odk.collect.android.javarosawrapper.FormController;
 import org.odk.collect.utilities.Clock;
 
 import javax.inject.Inject;
-
-import timber.log.Timber;
 
 import static org.odk.collect.android.javarosawrapper.FormIndexUtils.getRepeatGroupIndex;
 
@@ -133,8 +130,6 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
     }
 
     public void moveForward() {
-        ensureFormController();
-
         try {
             formController.stepToNextScreenEvent();
         } catch (JavaRosaException e) {
@@ -146,8 +141,6 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
     }
 
     public void moveBackward() {
-        ensureFormController();
-
         try {
             int event = formController.stepToPreviousScreenEvent();
 
@@ -164,7 +157,6 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
     }
 
     public void openHierarchy() {
-        ensureFormController();
         formController.getAuditEventLogger().logEvent(AuditEvent.AuditEventType.HIERARCHY, true, clock.getCurrentTime());
     }
 
@@ -177,13 +169,6 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
             return formController.getCurrentFormIdentifierHash();
         } else {
             return "";
-        }
-    }
-
-    private void ensureFormController() {
-        if (this.formController == null) {
-            Timber.e(new IllegalStateException("ensureFormController called before formLoaded"));
-            this.formController = Collect.getInstance().getFormController();
         }
     }
 


collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -183,7 +183,7 @@ public AuditEventLogger getAuditEventLogger() {
      */
     public String getXPath(FormIndex index) {
         String value;
-        switch (getEvent()) {
+        switch (getEvent(index)) {
```
grzesiek2010, line 186: The problem here was that if we were in a field-list group and we wanted to get XPath for a question to remember that that question is waiting for an answer it didn't work because we were passing it's index properly like 0,1 for example but then getEvent() (without any parameter) used our current index to return the XPath which in case of field-list groups is the index of the whole group not of that particular question.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2343,53 +2333,50 @@ public void denied() {
                         }
                     });
                 } else {
-                    Intent reqIntent = getIntent();
-                    boolean showFirst = reqIntent.getBooleanExtra("start", false);
```
grzesiek2010: In https://github.com/getodk/collect/pull/4286/commits/5b6ae1a3f2106b8395e3007964e65c3111b8a7e0 I just removed this check because we nowhere use `putBooleanExtra("start", false);` so this is always false. Please correct me if i made a mistake.
is it possible that external apps use this for something?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2343,53 +2333,50 @@ public void denied() {
...
-                                        }
-                                    }
-
-                                    formController.getAuditEventLogger().logEvent(AuditEvent.AuditEventType.FORM_RESUME, true, System.currentTimeMillis());
-                                    formController.getAuditEventLogger().logEvent(AuditEvent.AuditEventType.HIERARCHY, true, System.currentTimeMillis());
-                                    startActivityForResult(new Intent(this, FormHierarchyActivity.class), RequestCodes.HIERARCHY_ACTIVITY);
-                                }
-                            });
+                    // Returning to the app after process death
+                    if (formController.getFormIndex().isInForm()) {
```
grzesiek2010: If we are here (just after loading a form) and our index is in form (that mean it was saved and now restored) that means it was caused by process death. I hope there are no other cases :)

grzesiek2010:
There is one question that remains open: Do we need all those null checks we added to avoid using null formController object?
My answer is yes for now we need it because we start loading form in onCreate(), then it usually takes some time but other lifecycle methods like onStart(), onResume() are called and don't wait for form loading to be finished. In those methods we may try to use formController which at that point is null.

So this might be something for a separate pr. We could get rid of all those null checks if we added checks in lifecycle methods to make sure we skip executing them if a form is not ready.

Such cleaning would be great of course but I'm also a bit anxious, maybe there are other cases and removing those null checks would result in strange crashes? Hmm if we want take that risk and make the code a bit clearer it would be definitely something for an early beta.

@lognaturel @seadowg please share your thoughts.

seadowg:
I'll hopefully be able to take a look at this tomorrow!

grzesiek2010:
no rush it won't go in v1.29 and you probably have something more important.

grzesiek2010:
In onResume() we have code that my heart would love to remove but I feel anxious:

```java
if (formLoaderTask != null) {
            formLoaderTask.setFormLoaderListener(this);
            if (formController == null
                    && formLoaderTask.getStatus() == AsyncTask.Status.FINISHED) {
                FormController fec = formLoaderTask.getFormController();
                if (fec != null) {
                    if (!readPhoneStatePermissionRequestNeeded) {
                        loadingComplete(formLoaderTask, formLoaderTask.getFormDef(), null);
                    }
                } else {
                    DialogUtils.dismissDialog(FormLoadingDialogFragment.class, getSupportFragmentManager());
                    FormLoaderTask t = formLoaderTask;
                    formLoaderTask = null;
                    t.cancel(true);
                    t.destroy();
                    // there is no formController -- fire MainMenu activity?
                    startActivity(new Intent(this, MainMenuActivity.class));
                }
            }
        } else {
            if (formLoaderTask == null) {
                // there is no formController -- fire MainMenu activity?
                startActivity(new Intent(this, MainMenuActivity.class));
                finish();
                return;
            }
        }

```

Situation where we are in onResume() and  formLoaderTask is not null and is finished but formController is null seems unrealistic.
Another case where both formLoaderTask and formController are nulls also seems strange.

What about logging events to check if it happens?

grzesiek2010:
I'm still thinking about those null checks we have to avoid NPEs...
Doing it in a perfect way with our current code seems impossible, even if we introduce improvements in order to wait until form loading is finished and then remove all null checks it will be ok just for now but in the future it will be extremely difficult to remember about it to make sure we won't introduce NPEs.

A perfect solution could be separating form loading and displaying a form into two separate fragments. Thanks to that during form loading we wouldn't try to do all those things that are related to forms themselves and we wouldn't need all those null checks then.

Can you maybe thing of other solutions?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: I can't find anything on it and it doesn't look like it's documented (looking [here](https://docs.getodk.org/launch-collect-from-app/)) but I'm still a little worried it could be relied on somewhere out there? 

We can add a note to QA about this to see if they have any test cases for it. It doesn't look like the change would "break" anything if someone was using it - the user would just end up on the hierarchy.

seadowg: I can't find anything on it and it doesn't look like it's documented (looking [here](https://docs.getodk.org/launch-collect-from-app/)) but I'm still a little worried it could be relied on somewhere out there? 

We can add a note to QA about this to see if they have any test cases for it. It doesn't look like the change would "break" anything if someone was using it - the user would just end up on the hierarchy.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Could you add more details to the comment here? I'm not sure I follow exactly how we end up with an index in the form in this case but not if I kill the app and then open the form again? 

It'd also good to point out here that in the activity death scenario we don't need to reload the form so we don't end up here.

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
seadowg: Yeah this just seems cleaner as well!

seadowg (changes requested):

seadowg:
@grzesiek2010 looks like there is a conflict to fix here before handing this to QA.

grzesiek2010:
> looks like there is a conflict to fix here before handing this to QA.

Yes I expected conflicts after merging last fixes for v1.29.2 today I'm going to review and fix.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: It's not relevant anymore.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: I removed this change and I'll file a separate issue to discuss it.

grzesiek2010:
After merging last fixes for v1.29.2 we need fewer changes here.
- https://github.com/getodk/collect/pull/4286/commits/4f6685445ca65fc30d74a326f3a506ce44060895 is added because if `pendingActivityResult` is true in https://github.com/getodk/collect/pull/4286/commits/4f6685445ca65fc30d74a326f3a506ce44060895#diff-150be8209a860d77a445a11267821bfba08a71b0b8c581bbb115641f42dc7919R2314 return is called and formControllerAvailable() won't be called. The app wouldn't crash in this case because we  have this fix https://github.com/getodk/collect/commit/2e762a704389fd1fe3e26ead6b6b8200529ec24c but generally we shouldn't relay on it.
- https://github.com/getodk/collect/pull/4286/commits/35884cb2f39a1e98faec67bd0ed726ccb5e66c92 is added to fix returning to field-list group what we can test using the form I attached
and that's seems to be everything we need for now.

There is one open question about those null checks what I described https://github.com/getodk/collect/pull/4286#issuecomment-741744724 and in comments below.

seadowg (changes requested):
It looks like this change set breaks `IdentifyUserTest` (I think @lognaturel might have run into this before). We can probably do some rework on identify user or to get this to work in a simpler manner, but maybe we go with your original change (dealing with the pending activity result case specifically) for now?

grzesiek2010:
> It looks like this change set breaks

that small change breaks tests? I didn't expect it at all... I need to take a look to understand what's going on.

seadowg:
> that small change breaks tests? I didn't expect it at all... I need to take a look to understand what's going on.

@grzesiek2010 Yeah annoyingly. It feels like it's probably due to some over complication in the Identify User implementation. Your original solution didn't break those tests though.

lognaturel:
Yes, this is now the same as https://github.com/getodk/collect/pull/4326. To elaborate on the issue with the user identification, it's that the `formLoaded` method for that `ViewModel` requires the instance folder to exist which may not yet be the case at that point. I added the comment `Precondition: the instance directory must be ready so that the audit file can be created` on `formControllerAvailable` to hint at that.

I think this PR would ideally remove the hack introduced at https://github.com/getodk/collect/pull/4328. We could do it later but it would be a nice sanity check that the issue is addressed since we can easily reproduce the navigation case.

grzesiek2010:
Ok I fixed the problem and removed the fix implemented in #4328

seadowg (approved):

seadowg:

grzesiek2010:
There is one more issue that we may want to fix as a part of this pr:
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/f9991c3f867d61355cc17df3abee708c?time=last-thirty-days&versions=v1.29.2%20(4041)&sessionEventKey=5FF850BD00CA0001602398EAF3CBFBD3_1493812044950110121

I think it might share the same scenario with process death but additional condition is that null instance file.
What's interesting it's not a hard crash despite the fact it's NPE.

We could handle this case and close the form like we do in other similar cases but if it's not a hard crash I wonder why it doesn't cause other similar NPEs related to the instance file in other places...

seadowg:
@grzesiek2010 I'm thinking that might be good to fix as a follow-up PR (probably worth creating an issue)?

grzesiek2010:
Yes I agree especially that it's confusing. I reported it here: https://github.com/getodk/collect/issues/4340

mmarciniak90:
**Tested with success**

Verified on Android devices: 7.0, 8.1, 10.0

Verified cases:
- repro steps with image widget
- repro steps with video widget
- repro steps with audio widget
- form with multiple widgets on one page
- don't keep activities turned on and off
- verified image, selfie, annotate, new image widgets
- killing app manually

kkrawczyk123:

PR #4289, opened 2020-12-09
Make sure screen reloads when preferences are changed

### LINKED ISSUE #4212: Changes in audio recorders in experimental settings are not applied immediately when user sets it in the middle of the form
Collect master

#### Problem description
Changes in audio recorders in experimental settings are not applied immediately when user sets it in the middle of the form.

1. User starts form
2. User opens audio widget -> external recorder is opened
3. User opens `General Settings` in the middle of form
4. User goes to `Experimenta` settings
5. User unselects using an external app
6. User backs to the form and opens audio widget
7. External recorder is opened -> changes is settings wasn't applied

#### Expected behavior
The internal recorder is visible.
Changes in audio recorders are applied immediately even if user sets it in the middle of the form.

#### Notes
Related to #4190

seadowg:
Closes #4212 

New test and verified manually. There is one piece of tested code - comments inline.

I think it made sense to handle this generally rather than worrying about it for any new feature in form entry that can be configured.

The form "screen" will now reload whenever the user navigates to settings and back. It'd be good to check this doesn't create any problems (answers not retained for instance) and that there aren't any significant performance problems introduced.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -151,7 +151,6 @@ import org.odk.collect.android.utilities.MediaUtils;
 import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.utilities.PermissionUtils;
 import org.odk.collect.android.utilities.PlayServicesChecker;
-import org.odk.collect.android.utilities.QuestionFontSizeUtils;
 import org.odk.collect.android.utilities.ScreenContext;
 import org.odk.collect.android.utilities.SnackbarUtils;
 import org.odk.collect.android.utilities.SoftKeyboardUtils;
@@ -356,7 +355,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      * detection of location permissions changes while the activity is in the background.
      */
     private boolean locationPermissionsPreviouslyGranted;
-    private int questionFontSize;
 
     private BackgroundLocationViewModel backgroundLocationViewModel;
     private IdentityPromptViewModel identityPromptViewModel;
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -762,7 +762,7 @@ protected void onActivityResult(int requestCode, int resultCode, final Intent in
 
         // If we're coming back from the hierarchy view, the user has either tapped the back
         // button or another question to jump to so we need to rebuild the view.
-        if (requestCode == RequestCodes.HIERARCHY_ACTIVITY) {
+        if (requestCode == RequestCodes.HIERARCHY_ACTIVITY || requestCode == RequestCodes.CHANGE_SETTINGS) {
```
seadowg, line 761: This isn't tested. I played around with various Espresso tests but couldn't get something that worked nicely and actually drove this out. We still consider `FormEntryActivity` a no-go for Robolectric testing, so I didn't even look at that. 

The new [Activity Result APIs](https://developer.android.com/training/basics/intents/result) are really the answer here I think, but they are still in beta. Once that's stable we should be able to move our activity result code into components (like `FormEntryMenuDelegate`) and test the request and "response" in one place.

grzesiek2010, line 761: Can we do the same for font size? Probably we don't even need adjustFontSize() now if we call onScreenRefresh(); either way. Do you want to confirm? If you are in a hurry with this pr we can file a separate issue to make sure changes in settings are handled in the same way.

seadowg: Sounds like a good idea! Just to confirm: you're suggesting we get rid of the code that's dealing with the font size adjustment and see if this new solution works instead?

grzesiek2010: Yeah I guess it should because basically it does the same just calling onScreenRefresh();

grzesiek2010 (approved):

mmarciniak90:
**Tested with success**

Tested on Android 7.0, 8.1, 10.0

Verified cases:
- All settings in `General Settings`
- Theme and language work as before
- Guidance setting works immediately :+1: 
- `Default to finalized` setting is applied even if user opens General settings from last screen :+1:

kkrawczyk123:

PR #4290, opened 2020-12-10
Audio analytics

seadowg:
This adds analytics for the audio features. The different events and what we hope to track with them is detailed in `AnalyticsEvents`.

Ran tests.

Comments inline.

Shouldn't have any effect. This can be merged without QA.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -175,4 +175,9 @@ public class AnalyticsEvents {
      * Tracks how often questions are answered while an audio recording is being made
      */
     public static final String ANSWER_WHILE_RECORDING = "AnswerWhileRecording";
+
+    /**
+     * Tracks how many users have opted in to internal recording as their default
+     */
+    public static final String INTERNAL_RECORDING_OPT_IN = "InternalRecordingOptIn";
 }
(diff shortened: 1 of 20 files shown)


collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
```diff
@@ -153,4 +147,37 @@ private AnalyticsEvents() {
...
+     * audio question
+     */
+    public static final String AUDIO_RECORDING_INTERNAL = "AudioRecordingInternal";
+    public static final String AUDIO_RECORDING_EXTERNAL = "AudioRecordingExternal";
+    public static final String AUDIO_RECORDING_CHOOSE = "AudioRecordingChoose";
+
+    /**
+     * Tracks how often questions are answered while an audio recording is being made
+     */
+    public static final String ANSWER_WHILE_RECORDING = "AnswerWhileRecording";
```
seadowg, line 172: We initially talked abut an event for when an audio question (using internal) recording appeared in a field-list. I realized it was actually easier to just track the underlying behaviour we were interested in here.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -164,19 +156,34 @@ public void openHierarchy() {
         formController.getAuditEventLogger().logEvent(AuditEvent.AuditEventType.HIERARCHY, true, clock.getCurrentTime());
     }
 
+    public void logFormEvent(String event) {
```
seadowg, line 159: Rather than having to pass the (or grab statically) form identifier hash around it seemed nice to just encapsulate the logging on the view model here. I think we'd want a `FormEventLogger` interface if we were logging inside a widget (or anywhere it doesn't make sense to have a ViewModel) but I didn't run into that.

lognaturel (approved):

PR #4291, opened 2020-12-10
Fix crash when stop/pause pressed at the same time while recording

### LINKED ISSUE #4287: Collect crashes when use pause and stop button at the same time in the middle of recording
Collect v1.29.x, Android v all, device used...

#### Problem description
Collect is crashing when trying to click on pause and stop button in the same time, however, it is not easy to reproduce as error message display most often than crash. 
ezgif com-video-to-gif (46)

1. Start recording using an internal recorder 
2. Tap fast on pause and stop button
It might not be visible in all tries.

#### Expected behavior
- Collect should not crash
- Using pause and stop buttons at once should be blocked

#### Other information 
stacktrace of the crash:
`2020-12-09 11:30:10.012 19931-19931/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 19931
    java.lang.RuntimeException: Unable to start service org.odk.collect.audiorecorder.recording.internal.AudioRecorderService@1d67199 with Intent { act=STOP cmp=org.odk.collect.android/org.odk.collect.audiorecorder.recording.internal.AudioRecorderService }: java.lang.NullPointerException
        at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3722)
        at android.app.ActivityThread.access$1700(ActivityThread.java:200)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1704)
        at android.os.Looper.loop(Looper.java:201)
        at android.app.ActivityThread.main(ActivityThread.java:6810)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873)
     Caused by: java.lang.NullPointerException
        at org.odk.collect.audiorecorder.recorder.RecordingResourceRecorder.stop(RecordingResourceRecorder.kt:53)
        at org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.stopRecording(AudioRecorderService.kt:99)
        at org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.onStartCommand(AudioRecorderService.kt:62)
        at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3700)
        at android.app.ActivityThread.access$1700(ActivityThread.java:200) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1704) 
        at android.os.Looper.loop(Looper.java:201) 
        at android.app.ActivityThread.main(ActivityThread.java:6810) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873) 
2020-12-09 11:30:10.342 2031-2124/? E/InputDispatcher: channel '8595c0f org.odk.collect.android/org.odk.collect.android.activities.FormEntryActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
2020-12-09 11:30:10.343 2031-2124/? E/InputDispatcher: channel 'a3bec56 org.odk.collect.android/org.odk.collect.android.activities.MainMenuActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
2020-12-09 11:30:10.428 2031-4289/? E/ActivityTrigger: activityResumeTrigger: not whiteListedorg.odk.collect.android/org.odk.collect.android.activities.MainMenuActivity/4017
2020-12-09 11:30:10.430 541-572/? E/ANDR-PERF-MPCTL: Invalid profile no. 0, total profiles 0 only
2020-12-09 11:30:10.613 20740-20740/org.odk.collect.android E/collect.androi: Dex checksum does not match for dex: /system/framework/com.google.android.maps.jar.Expected: 3079070930, actual: 1845989642
`

seadowg:
Closes #4287

I wanted to make sure to deal with this in the `audiorecorder` framework as well as just using our standard multi click stuff. Feels risky to allow these sorts of crashes to happen in the framework if the UI doesn't protect it.

This should only fix the problem!

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/activities/MainActivityTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/activities/MainActivityTest.java
@@ -13,15 +13,10 @@ import org.robolectric.RobolectricTestRunner;
 import org.robolectric.shadows.ShadowActivity;
 import org.robolectric.shadows.ShadowEnvironment;
 import org.robolectric.shadows.ShadowIntent;
-import org.robolectric.shadows.ShadowSystemClock;
-
-import java.time.Duration;
 
 import static android.os.Environment.MEDIA_MOUNTED;
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertNotNull;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
 import static org.robolectric.Shadows.shadowOf;
 
 @RunWith(RobolectricTestRunner.class)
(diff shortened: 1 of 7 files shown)


collect_app/src/test/java/org/odk/collect/android/activities/MainActivityTest.java
```diff
@@ -82,54 +77,4 @@ public void manageFilesButtonTest() throws Exception {
         assertEquals(DeleteSavedFormActivity.class.getName(),
                 shadowIntent.getIntentClass().getName());
     }
-
-    @Test
-    public void shouldButtonsBeResistantToMulticlicking() {
```
seadowg, line 87: This test started failing when I disabled the multiclick guard in tests (so that other tests wouldn't fail). We've generally not been testing the multiclick guards (which I think makes sense) so personally I don't think we need this. Happy to chat about it though!

lognaturel (approved):
@grzesiek2010 should take a quick look at the multiclick test changes.

kkrawczyk123:
@seadowg I am not able to reproduce crash or even display error message what is great but I don't like the fact that after using pause button, stop button doesn't work in the first try - it must have been used for the second time. It's weird and a little confusing, see gif:
ezgif com-video-to-gif (48)

seadowg:
@kkrawczyk123 it looks like our multi click blocking needs you to wait a full second before another click will be allowed. It looks like you're beating this in the gif haha. As much as I'd usually expect a second to be short enough I think you just proved that it's not. I'll bring the time down to half a second and you can see how it feels (@grzesiek2010 please jump in if you think that's going to be a problem).

seadowg:
@kkrawczyk123 give that go!

grzesiek2010:
I don't like changing that 1s to 0.5s. We use it in many places and I'm pretty sure for some cases 0.5s was not enough that's why we decided to use 1s.

seadowg:
@grzesiek2010 thanks for weighing in. That makes a lot of sense. It'd be good to understand if we can use other methods to prevent multi/double clicks wait for a full second (progress dialogs, disabling buttons etc) but that's obviously not an investigation or set of changes we want to take on at this moment!

We might be able to actually leave the multi click guard out in this case as I think there may now be enough protection in the recording code. Let me investigate and see what I can come up with (cc/ @kkrawczyk123).

kkrawczyk123:
@seadowg I am still able to do it but only when clicking on the stop button right after clicking on pause so very quickly but when I click on them like a normal human being everything looks fine. I honestly think that nobody except us would do that @mmarciniak90 do you agree?

grzesiek2010:
Maybe create your own MultiClickSafeButton which would pass 0.5s calling MultiClickGuard.allowClick() to avoid regressions in all the other places we already use 1s.

seadowg:
@kkrawczyk123 I've tried just taking the multi click guards out and reverting the time change as @grzesiek2010 suggested. I'm preoccupied with whatever has happened to our CI so haven't had a chance to play on a real device but if you get a chance to try it out let me know if you can break it.

> Maybe create your own MultiClickSafeButton which would pass 0.5s calling MultiClickGuard.allowClick() to avoid regressions in all the other places we already use 1s.

That might be the best option. Hopefully we can get away without any UI tricks if the recording code holds up.

mmarciniak90:
**Tested with success**

Verified on Android versions: 5.1, 7.0, 7.1, 9.0, 8.1, 10.0

Verified cases:
- pause/stop buttons
- pause/record buttons
- error message
- crash does not occur

lognaturel:
We typically don't merge when `master` is broken but I'm merging now because we know there's some issue specific to Firebase Test Lab and I'd like to see whether anything has changed. Instead of re-running an existing build I think it makes sense to also get `master` up to date.

Issue #2834, opened 2019-01-29
Add tests for ContentResolverHelper


grzesiek2010:
#### Problem description
in #2823 we fixed a bug in `ContentResolverHelper` class. This class is not covered with tests at all. It would be good to add some.

grzesiek2010:
The mentioned class no longer exists so let's close the issue.

PR #4297, opened 2020-12-11
Do not ask for storage permissions when it's a fresh installation

### LINKED ISSUE #4288: Do not ask for storage permissions users who use scoped storage
#### Problem description
Using scoped storage we don't need storage permission because it's our own space.

#### Expected behavior
New users and those who have already migrated to scoped storage should not be asked to grant storage permission.

grzesiek2010:
Closes #4288

I tested the behavior.

I decided to call `requestStoragePermissions()` like before but add one trick which is here:
https://github.com/getodk/collect/pull/4297/files#diff-fe2bc5034d6dfe946ff681f198d0552c83038caed992865f5e7ce030447a276bR98

Another option would be to check if we need storage permission before calling `requestStoragePermissions()` but that would require adding more if/else statements and the code would be more cluttered.

Fresh installs should not ask for storage permission that's the only change. In order to confirm that everything works fine it would be good to go through some functionalities where normally we would be asked for permissions.

No.

No.


lognaturel (commented):
Would be great to get some testing. Maybe just a small test for `areStoragePermissionsGranted` that verifies that `true` is always returned if scoped storage is enabled?

I like what you've done with `PermissionUtils`. I'm wondering whether it's also a good opportunity to rename it to maybe `PermissionsRequester` or `PermissionsProvider`?

grzesiek2010:
> Would be great to get some testing

I thought the same at the end of my day yesterday. I'll try to add something valuable.

grzesiek2010:
> Would be great to get some testing. Maybe just a small test for areStoragePermissionsGranted that verifies that true is always returned if scoped storage is enabled?

I like what you've done with PermissionUtils. I'm wondering whether it's also a good opportunity to rename it to maybe PermissionsRequester or PermissionsProvider?

It's ready, I added some tests and implemented code refactoring.

collect_app/src/test/java/org/odk/collect/android/permissions/PermissionsProviderTest.java
```diff
@@ -0,0 +1,141 @@
...
+        when(storageStateProvider.isScopedStorageUsed()).thenReturn(false);
+        permissionsProvider = new PermissionsProvider(permissionsChecker, storageStateProvider);
+
+        assertThat(permissionsProvider.areStoragePermissionsGranted(), is(false));
+    }
+
+    @Test
+    public void whenStoragePermissionsNotGrantedAndScopedStorageUsed_shouldAreStoragePermissionsGrantedReturnTrue() {
+        when(permissionsChecker.isPermissionGranted(Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE)).thenReturn(false);
+        when(storageStateProvider.isScopedStorageUsed()).thenReturn(false);
```
lognaturel: This should be `true` so that the last assertion is `true`!

lognaturel (changes requested):
Looks good other than the main test... 🙃 Feel free to mark as `needs testing` once that's fixed and conflicts are resolved.

collect_app/src/test/java/org/odk/collect/android/permissions/PermissionsProviderTest.java
grzesiek2010: Ahh good catch, thanks!

kkrawczyk123:
@grzesiek2010 Not asking for storage permissions with a fresh install causes crashes when try to use widgets with adding files from device's storage - image, video, audio, annotate and file widget as those widgets need storage permission enabled. I can enable it in settings/ app permissions and then it works well but none of the users would actually do that. If you really want to remove asking for permissions right after installation perhaps adding another permission dialog after using the "Choose" buttons should be considered. 

grzesiek2010:
> Not asking for storage permissions with a fresh install causes crashes when try to use widgets with adding files from device's storage - image, video, audio, annotate and file widget as those widgets need storage permission enabled.

Does that depend on Android version or the source of files? I tested Api 29 and yeah the behavior is not consistent sometimes I get a message that I can't access a file that should be accessible but I wasn't able to reproduce a crash.

mmarciniak90:
We noticed slight differences depending on the version, e.g. Android 10 does not crash, it just shows the toast `Error Occured`. The easiest way to reproduce is to add an attachment from the IMAGES / AUDIO / VIDEOS directory.

grzesiek2010:
Ok everything seems to come down to Storage Access Framework:
https://developer.android.com/guide/topics/providers/document-provider
https://developer.android.com/training/data-storage/shared/documents-files

and the question I asked here: https://github.com/getodk/collect/pull/4351#issuecomment-761998944

looks like we not only don't need that nasty code (which I described in the question attached above) to retrieve files but we even have to stop using it.

grzesiek2010:
Ok I implemented some changes. @mmarciniak90 @kkrawczyk123 please test:
- capturing and attaching files in media widgets
- importing settings using a qr code from saved image (not scanning)
- setting custom splash screen images

did I miss something where we attach files?

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
```diff
@@ -76,20 +73,8 @@ public static String getFormPath(Uri uri) {
...
-        if (fileName != null && fileName.contains(".")) {
-            return fileName.substring(fileName.lastIndexOf('.') + 1);
-        } else {
-            return fileUri.getScheme() != null && fileUri.getScheme().equals(ContentResolver.SCHEME_CONTENT)
-                    ? MimeTypeMap.getSingleton().getExtensionFromMimeType(context.getContentResolver().getType(fileUri))
-                    : MimeTypeMap.getFileExtensionFromUrl(fileUri.toString());
-        }
+    public static String getFileExtensionFromUri(Uri fileUri) {
+        String mimeType = getContentResolver().getType(fileUri);
+        return mimeType.substring(mimeType.lastIndexOf("/") + 1);
```
lognaturel: This happens to work for some common types like `image/jpeg` but the subtype name is not guaranteed to be the extension. For example the SVG MIME type is `image/svg+xml` but its extension is `.svg`. Or `audio/amr-wb` and `.amr`. I think you have to use [MimeTypeMap#getExtensionFromMimeType](https://developer.android.com/reference/android/webkit/MimeTypeMap#getExtensionFromMimeType(java.lang.String)).

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -87,12 +86,6 @@ public static void releaseDatabaseHelper() {
 
     @Override
     public boolean onCreate() {
-
-        if (!areStoragePermissionsGranted(getContext())) {
```
lognaturel, line 91: I think this can all go away once we don't support files in `/sdcard/odk/` anymore but at the moment, removing it could result in a crash if someone's migration failed, they haven't yet granted permissions, and they try to access forms from an external app, right?

collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java
```diff
@@ -87,11 +86,6 @@ public static void releaseDatabaseHelper() {
 
     @Override
     public boolean onCreate() {
-        if (!areStoragePermissionsGranted(getContext())) {
```
lognaturel, line 90: Same question as for `FormsProvider`.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
```diff
@@ -82,362 +60,8 @@ public void openFile(Context context, File file) {
         }
     }
 
-    public String getMimeType(String url) {
+    private String getMimeType(String url) {
```
lognaturel: This seems unnecessary and likely to fail. It should be possible to use [`FileProvider#getType`](https://developer.android.com/reference/androidx/core/content/FileProvider#getType(android.net.Uri)).

lognaturel:
> @mmarciniak90 @kkrawczyk123 please test:

The really important things to check are around edge cases for file types, file extensions and file sources. For example, what if you have a jpg with the correct MIME type but no extension? What if you have a text file with extension `.fake` and the correct MIME type? What if you have files coming from Google Drive? Checking that for one place where files are attached should be sufficient.

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
```diff
@@ -144,40 +144,25 @@ public IAnswerData getAnswer() {
         }
     }
 
-    /**
```
lognaturel, line 147: `FormSaveViewModel.createAnswerFile` returns a `LiveData<Result<String>>` used to set the data from a file name. Have you tried this change with internal recording?

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
grzesiek2010: Generally you are right and I tried it but in some cases (which seem pretty obvious) it returns null those cases are like:
- audio/mp3
- text/csv

Here https://stackoverflow.com/questions/12473851/how-i-can-get-the-mime-type-of-a-file-having-its-uri/12473985 where that solution is marked as accepted people also complain about the same.

Maybe I should use what you mentioned first and then in case of null my: `mimeType.substring(mimeType.lastIndexOf("/") + 1);`

or better yet (as a second step) I can read filename that should contain file extension:
```
        Cursor cursor = getContentResolver().query(fileUri, null, null, null, null);
        if (cursor != null && cursor.moveToFirst()) {
            name = cursor.getString(cursor.getColumnIndex(MediaStore.MediaColumns.DISPLAY_NAME));
        }
        String extension = name != null ? name.substring(name.lastIndexOf(".") + 1) : "";
```

this is what this library does https://github.com/anggrayudi/SimpleStorage. I know it's new and didn't gain traction yet but i ran into it through this topic https://stackoverflow.com/questions/63080879/get-file-extension-using-android-storage-access-framework.

What do you think?

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
grzesiek2010: I tested it using collectTester app and:
- if you open list of blank form everything is fine, you are asked to grant permissions
- if you open list of saved forms **as above**
- if you open list of forms to submit **as above**
- if you open list of sent forms **as above**
- if you download forms passing ids **as above**
- if you upload forms passing ids **as above**

It only crashes if an external app tries to get list of forms or instances.
The problem was with injecting objects in content providers which are initialized before other app components but finally I fixed this issue thanks to https://stackoverflow.com/questions/23521083/inject-database-in-a-contentprovider-with-dagger
https://github.com/getodk/collect/pull/4297/commits/0c1ff84374c4ca7dc7c83067cea384473fcb1767

collect_app/src/main/java/org/odk/collect/android/provider/InstanceProvider.java
grzesiek2010: As above in case of FormsProvider

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
grzesiek2010: I tested the demo form with internal recording and everything seems fine.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
lognaturel: ANDROID! Why must it always make simple things hard. Good catch on `MimeTypeMap#getExtensionFromMimeType` not actually working all the time.

Doing that first and then falling back seems good. I think you might actually want to have both fallbacks since neither is perfect -- `MediaStore.MediaColumns.DISPLAY_NAME` is not necessarily a file name. Maybe first use `MimeTypeMap#getExtensionFromMimeType`, then try the display column and then use the subtype?

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
lognaturel: Interesting! @seadowg, would be good for you to at least see this change happening since it's to very recent audio code.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
lognaturel: Makes sense!

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
grzesiek2010: Ok I did it I even added another fourth aproach since I noticed that in case of selfie video only that last approach works well :scream:

lognaturel: Yikes! Is that maybe because selfie video gives back a file URI (not a content URI) without a MIME type set? Maybe that's worth looking at again after this PR.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java
grzesiek2010: I'm rolling back this. I tested that way and it doesn't work in all cases but the old one does at leas for the cases I tested. Oh Android you are so funny! 
I added some fallbacks just in case like in getFileExtensionFromUri() above.

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
seadowg: Looks like @lognaturel is right as `AudioRecordingTest` and `AudioWidgetTest` are failing. From debugging, it looks like `AudioWidget#setData` is still passed a `String`.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
```diff
@@ -76,20 +79,43 @@ public static String getFormPath(Uri uri) {
...
-        if (fileName != null && fileName.contains(".")) {
-            return fileName.substring(fileName.lastIndexOf('.') + 1);
-        } else {
-            return fileUri.getScheme() != null && fileUri.getScheme().equals(ContentResolver.SCHEME_CONTENT)
-                    ? MimeTypeMap.getSingleton().getExtensionFromMimeType(context.getContentResolver().getType(fileUri))
-                    : MimeTypeMap.getFileExtensionFromUrl(fileUri.toString());
+    public static String getFileExtensionFromUri(Uri fileUri) {
+        String mimeType = getContentResolver().getType(fileUri);
+
+        String extension = fileUri.getScheme() != null && fileUri.getScheme().equals(ContentResolver.SCHEME_CONTENT)
```
lognaturel, line 85: Wouldn't we always be dealing with content URIs here? In what cases would we have file URIs?

lognaturel: Or was that the selfie video case again?

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
lognaturel: Do you remember whether there was a specific reason you wanted to use a `String`, @seadowg? Looks like using a `LiveData<Result<File>>` works fine but want to make sure there isn't a strong reason against it.

grzesiek2010: Yeah I have just turned it into files so please answer if there was any reason.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
grzesiek2010: This was the fix added for a crash some time ago https://github.com/getodk/collect/commit/f98b98dddc3cc9af8e1db5144f4c9c7313854df1 so I wanted to keep it just in case seeing how other cases might be tricky.

grzesiek2010:
@lognaturel Generally I think it's ready for testing unless you see something important.

collect_app/src/main/java/org/odk/collect/android/widgets/AudioWidget.java
seadowg: I don't **think** it needs to be a `String`. It was just because it didn't need to be a file as the the other side could just use `getAnswerFile` with the name (the `String`) to access the actual file. I think if `File` is more consistent we can go with that!

collect_app/src/main/java/org/odk/collect/android/dao/helpers/ContentResolverHelper.java
lognaturel: Got it. Would be great to follow up with #2834 because this method has gone through lots of big changes during this PR that probably would have been avoided with tests.

lognaturel (approved):
This has turned into an adventure! Good to see so much code going away.

Lots of little things to check around the media selection questions like are temporary files cleaned up and all the [cases around MIME types and file extensions I listed earlier](https://github.com/getodk/collect/pull/4297#issuecomment-763264011).

grzesiek2010:
I added some improvements to handle `SecurityException` which is thrown when read permission is no granted to the returned uri. I was even able to reproduce it on an emulator Api 25 where I can't record video because apparently the default app doesn't grant read permission to returned uri. Please keep an eye on it during testing.

lognaturel:
> I was even able to reproduce it on an emulator Api 25

Nice. But does this mean that recording video with that default app would never work? I wonder if this is going to be a broader issue -- are media apps for older OS versions not going to work because they expect the blanket read permissions to be given?

grzesiek2010:
> Nice. But does this mean that recording video with that default app would never work? I wonder if this is going to be a broader issue -- are media apps for older OS versions not going to work because they expect the blanket read permissions to be given?

I tested emulators with apis 22-30 and noticed that problem on api 23 and 25. In those cases recording is not possible with default camera apps. Granting storage permission fixes the issue but I thought it was a bug in those camera apps. In the doc where recording videos is described https://developer.android.com/training/camera/videobasics there is nothing about storage permission and we use exactly that way.

Apparently they love to confuse people.

I also read https://developer.android.com/training/data-storage/shared/media and https://developer.android.com/training/data-storage/shared/documents-files and my understanding is that we shouldn't need that permissions.

mmarciniak90:
**Tested with success**

Verified cases:
- Image widgets
   - choose image from Image directory
   - choose image from Downloads
   - choose image from Google Drive
   - take new photo
   - Camera permissions
- Audio widget
   - choose audio file from Audio directory
   - choose audio file from Downloads
   - choose audio file from Google Drive
   - record new audio
   - Microphone permissions
- Video widget
   - choose video file from Video directory
   - choose video file from Downloads
   - choose video file from Google Drive
   - record new video
   - Camera and Microphone permissions
- File widget
   - choose any file for file widget from Images/Videos/Audio
   - choose any file for file widget from Downloads
   - choose any file for file widget from Google Drive
   - verified file types: amr, ogg, m4a, mp3, 3gp, gif, jpeg, jpg, png, pdf, doc, docx, txt, xls, xlsx, csv, xml, rtf, zip, rar files
- Location permissions
- Contact permissions
- Telephone permissions
- enabled storage permission manually
- fresh installation
- import settings using a qr code file
- import settings from .settings / json file
- custom splash screen images - file saved in /storage/emulated/0/Android/data/org.odk.collect.android/files 
- upgrade when storage permission is enabled
- upgrade when storage permission is disabled
   - if files are migrated, storage permissions are not required
   - if files are not migrated, storage permissions are required
- jpg with the correct MIME type but no extension (internal storage and GD) - verified on images widgets and file widget
- text file with extension .fake and the correct MIME type (internal storage and GD) - verified on file widget

kkrawczyk123:

grzesiek2010:
Regarding that problems with permissions: If it's not our fault bug external apps that do not grant uri permission we may need to consider adding something like: https://github.com/getodk/collect/commit/17431bbc5d663f01a349c43d40682674524dfc31

PR #4275, opened 2020-12-07
Add error handling for audio recording

seadowg:
Closes #4240

#### What has been done to verify that this works as intended?

New tests and verified manually.

#### Why is this the best possible solution? Were any other approaches considered?

Comments inline.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

The actual error cases here are pretty hard to reproduce, so I think the best way to test this is to try and throw everything at the audio recording and see if we can break it any way that we've not handled.

#### Do we need any specific form for testing your changes? If so, please attach one.

Any audio form.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4301, opened 2020-12-14
Show error when starting recording while mic in use

### LINKED ISSUE #4294: Crash when trying to record when another app already is
`master`

#### Problem description

The app crashes when starting a recording (using the internal recorder) if another app is already using the mic.

1. Start recording using another audio recorder app
2. Switch to Collect and open a form that an audio question (that has `quality` set to use the internal recorder )
3. Tap "Record sound"

#### Expected behavior

Recording should fail to start but the app should not crash. Ideally the user would see an error message telling them they are already using the mic somewhere else.

#### Other information 

Stacktrace:

```
Caused by: java.lang.IllegalStateException
        at android.media.MediaRecorder.start(Native Method)
        at org.odk.collect.audiorecorder.mediarecorder.MediaRecorderRecordingResource.start(MediaRecorderRecordingResource.kt:20)
        at org.odk.collect.audiorecorder.recorder.RecordingResourceRecorder.start(RecordingResourceRecorder.kt:39)
        at org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.startRecording(AudioRecorderService.kt:77)
        at org.odk.collect.audiorecorder.recording.internal.AudioRecorderService.onStartCommand(AudioRecorderService.kt:43)
```

seadowg:
Closes #4294

Not much to comment on really!

Don't think there is much risk here so just verifying the issue is fixed should be enough.


.circleci/config.yml
```diff
@@ -102,9 +102,22 @@ jobs:
       - store_test_results:
           path: collect_app/build/test-results
 
+  build_instrumented:
```
seadowg, line 105: This was adding a couple of mins onto our `test_app` tasks so made sense to pull it out so it could run in parallel and keep the PR build time down.

.circleci/config.yml
```diff
@@ -157,6 +170,10 @@ workflows:
       - check_quality
       - test_modules
       - test_app
+      - build_instrumented:
+          filters:
+            branches:
+              ignore: master
```
seadowg, line 180: Don't need to do this on `master` as we'll run `test_instrumented` there.

lognaturel (approved):

mmarciniak90:
I was able to reproduce this issue on Android 7.0 and 8.1.

@seadowg I see a new dialog but when I click on device back button instead of the OK button, I'm blocked.

ezgif com-video-to-gif-2

seadowg:
@mmarciniak90 huh. Will have a look into this!

seadowg:
@mmarciniak90 ah great catch. It looks like this affects the error handling we introduced in #4275 as well. Basically if you press "OK" everything is fine but pressing back creates the blocked case you found for both.

mmarciniak90:
Exactly! I've just discovered that the same behavior is when you click somewhere else than on active dialog.

seadowg:
@mmarciniak90 ok that should be fixed!

mmarciniak90:
@seadowg, thanks! I have no further comments.

**Tested with success**

As I mentioned before I'm able to reproduce this issue only on Android 7.0 and 8.1

Verified cases:
 - dialog with message `Microphone is already in use` appears - crash does not occur (Android 7.0, 8.1) 
 - closing dialog by OK button, device back button, click on an inactive area
 - audio widget with quality declared
 - audio widget with internal recorder set in general settings
 - light/dark mode
 - minimize
 - rotation
 - tried to record audio via an external app when recording via an internal app in ODK is in progress
 

PR #4309, opened 2020-12-15
Removed redundant ChangingDatabasesByForksTest

grzesiek2010:
Just reviewed implemented changes.

`ChangingDatabasesByForksTest` is sometimes failing on firebase as you have probably noticed. Since we migrated to scoped storage and we no longer share the same directory (and databases) with forks this test is redundant because other apps can't use and edit our databases.

It doesn't require testing.

No.

No.


grzesiek2010:
@lognaturel 
we could also get rid of:
https://github.com/getodk/collect/blob/022d8dcabf3940cc5978fb308b66d6955d59bea9/collect_app/src/main/java/org/odk/collect/android/database/FormsDatabaseHelper.java#L94

or maybe you would like to keep it just in case (i don't know what case exactly :smile:)?

lognaturel:
> we could also get rid of

I think we should as well as the corresponding `instance` method. But how about we do that at the beginning of v1.30 so it has a bit to get exercised? I think it's right that the only way the databases would be out of date would be if a fork touched them but it feels like there's a tiny bit of risk there.

grzesiek2010:
> I think we should as well as the corresponding instance method. But how about we do that at the beginning of v1.30 so it has a bit to get exercised?

I agree let's wait with this pr. I removed those redundant checks and added some refactoring as well.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/database/helpers/InstancesDatabaseHelperTest.java
```diff
@@ -29,7 +31,7 @@
 @RunWith(Parameterized.class)
 @Ignore("`Parameterized` causes problems for Firebase sharding. Probably need to replace this at JUnit level")
 public class InstancesDatabaseHelperTest extends SqlLiteHelperTest {
```
seadowg, line 36: It looks like these tests all fail. Should we rework or delete (if we don't need them) as part of this?

seadowg (changes requested):
Looks great! Just a question really about what to do with the ignored tests.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/database/helpers/InstancesDatabaseHelperTest.java
grzesiek2010: I think it would be good to have such tests that confirm that upgrading/downgrading databases work well so I would keep it.

seadowg: Totally agree that we want that coverage. With forms, it seems we have that covered in `FormDatabaseMigratorTest` so we could add a similar test for instances and remove this one (so replace the instrumented test with a Robolectric test). That said this PR definitely offers a bunch of good stuff, so we could just create an issue for that?

grzesiek2010: Yeah I agree.

seadowg (approved):

seadowg:

I think it'd be good to check upgrading and downgrading still as our core testing around that (for instances) is ignored at the moment.

mmarciniak90:
**Tested with success**

Verified cases:
- upgrade from v1.27 to version with PR
- upgrade from v1.28 to version with PR
- downgrade from version with PR to v1.27 - v1.27 refers to /odk directory and data from 1.29 are not available. I don't think this is a big deal because downgrading is not possible for the end user.
- downgrade from version with PR to v1.28
- settings are not restored to default
- bunch of files like forms, instances, layers available on device during each upgrade/downgrade

Manually generated apks were used to test because downgrade is not possible for released versions and upgrade is possible only if apks are signed by the same key

kkrawczyk123:

PR #4314, opened 2020-12-16
Add logging

lognaturel:
The first commit is to try to get more information on [this exception](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/dacbfe39d324ac95ff15529a8b7841bf?time=last-ninety-days) (needs auth). It happens on form save, generally triggered by `onPause` but sometimes also by manually exiting out of the form. The user experience is that a dialog is shown with the `NullPointerException` message. I believe that in case of a manual exit out of the form, the user is forced to tap "OK" and then their data is lost rather than saved. When this is triggered by `onPause`, it's possible that there would be a later opportunity to save again. Either way, we don't want this happening. It seems pretty clear it only affects v1.28.x and I suspect a JavaRosa issue though I can't see what it would be. Hopefully getting more information on the shape of the reference that causes this problem will help.

The second commit is for troubleshooting [this exception](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/1923b875080051dec14f1ebfaa6d716f?time=last-ninety-days). On instance save, in the code path where this is the first time the instance is being saved, there can be a crash on attempt to get the blank form corresponding to the new instance. One possible issue I saw is that the field `uri` was updated when a local variable should have been used. This looks like a very safe change so I made it but can also revert if preferred. I also added logging.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -516,8 +516,10 @@ public class FormController {
         try {
             return formEntryController.saveAnswer(index, data, true);
         } catch (Exception e) {
+            String dataType = data != null ? data.getClass().toString() : null;
+            String ref = index != null ? index.getReference().toString() : null;
             Timber.w("Error saving answer of type %s with ref %s for index %s",
-                    data.getClass().toString(), index.getReference(), index);
+                    dataType, ref, index);
 
             throw new JavaRosaException(e);
         }
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -516,6 +516,9 @@ public class FormController {
         try {
             return formEntryController.saveAnswer(index, data, true);
         } catch (Exception e) {
+            Timber.w("Error saving answer of type %s with ref %s for index %s",
+                    data.getClass().toString(), index.getReference(), index);
+
             throw new JavaRosaException(e);
         }
     }
--- a/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
+++ b/collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
@@ -94,7 +94,7 @@ public class PropertyManager implements IPropertyManager {
             putProperty(PROPMGR_DEVICE_ID,     "",         deviceDetailsProvider.getDeviceId());
             putProperty(PROPMGR_PHONE_NUMBER,  SCHEME_TEL,          deviceDetailsProvider.getLine1Number());
         } catch (SecurityException e) {
-            Timber.e(e);
+            Timber.i(e);
         }
 
         // User-defined properties. Will replace any above with the same PROPMGR_ name.
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
@@ -183,8 +183,7 @@ public class SaveFormToDisk {
         FormInstance formInstance = formController.getFormDef().getInstance();
 
         // If FormEntryActivity was started with an instance, update that instance
-        if (Collect.getInstance().getContentResolver().getType(uri).equals(
-                InstanceColumns.CONTENT_ITEM_TYPE)) {
+        if (Collect.getInstance().getContentResolver().getType(uri).equals(InstanceColumns.CONTENT_ITEM_TYPE)) {
             // TODO: reduce geometry duplication across three branches with different database queries
             String geometryXpath = getGeometryXpathForInstance(uri);
             ContentValues geometryContentValues = extractGeometryContentValues(formInstance, geometryXpath);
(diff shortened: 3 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
```diff
@@ -212,9 +210,9 @@ private void updateInstanceDatabase(boolean incomplete, boolean canEditAfterComp
             InstancesRepository instances = new DatabaseInstancesRepository();
             Instance instance = instances.getOneByPath(instancePath);
             if (instance != null) {
-                uri = Uri.withAppendedPath(InstanceColumns.CONTENT_URI, instance.getId().toString());
+                Uri instanceUri = Uri.withAppendedPath(InstanceColumns.CONTENT_URI, instance.getId().toString());
```
lognaturel: This should not have been reusing the field as it's just used locally. One idea I have for why this crash happens is that somehow there is an instance that matches the path in the DB but the update fails. Then the else branch below is hit but the uri is set to the instance uri rather than the form uri as it should be. This feels like a really safe fix to me but it would be good to get confirmation.

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -516,6 +516,9 @@ public boolean saveAnswer(FormIndex index, IAnswerData data) throws JavaRosaExce
         try {
             return formEntryController.saveAnswer(index, data, true);
         } catch (Exception e) {
+            Timber.w("Error saving answer of type %s with ref %s for index %s",
```
seadowg, line 521: Does this not need to be `Timber.e`?

collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
```diff
@@ -231,7 +229,11 @@ private void updateInstanceDatabase(boolean incomplete, boolean canEditAfterComp
             } else {
                 Timber.i("No instance found, creating");
                 try (Cursor c = Collect.getInstance().getContentResolver().query(uri, null, null, null, null)) {
-                    // retrieve the form definition...
+                    // retrieve the form definition
+                    if (c.getCount() != 1) {
+                        Timber.w("No matching form found for %s", uri);
```
seadowg, line 234: Again I thought these had to be `Timber.e` but I could be wrong.

collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
seadowg: Ach I'm in two minds. I think your explanation makes sense, but I'm a little worried about merging this in during regression testing especially as it's not tested. How do you feel about this being pulled out as a separate change (we can potentially as an issue) so we can add tests around it and merge later?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
seadowg: Oh no looks like I'm wrong (looking at `ApplicationInitializer`). Ignore me. I see now what the diff is.

lognaturel: Right, it's a little weird. `w` is for things that we don't want as a top-level issue but that we want to show up in the logs we see in Firebase to provide context. It feels like a little bit of an abuse of `w` but I don't see another way. We definitely don't want to send lower log levels to Firebase. 

Side note, it's not totally clear to me whether the top-level "non-fatal" crashes we see from Firebase are from `Timber.e` or from Firebase automatically intercepting exceptions even if they're caught or possibly both.

collect_app/src/main/java/org/odk/collect/android/tasks/SaveFormToDisk.java
lognaturel: Fair enough. Let me take it out and the logging will help us either confirm that's the issue or discard that hypothesis. Then we can either add the change in a point release or in v1.30.

seadowg (dismissed):

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -516,6 +516,9 @@ public boolean saveAnswer(FormIndex index, IAnswerData data) throws JavaRosaExce
...
+                    data.getClass().toString(), index.getReference(), index);
```
seadowg: Ok sorry now I'm being really paranoid, but I think it'd be good to make that we account for null `data` or `index` so there is no risk of ending up with a `NullPointerException` here.

seadowg (changes requested):
As comment.

lognaturel:
Going back to draft for now. We have another day of regression testing ahead so I may add more logging improvements during my day.

collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
```diff
@@ -94,7 +94,7 @@ public PropertyManager reload() {
             putProperty(PROPMGR_DEVICE_ID,     "",         deviceDetailsProvider.getDeviceId());
             putProperty(PROPMGR_PHONE_NUMBER,  SCHEME_TEL,          deviceDetailsProvider.getLine1Number());
         } catch (SecurityException e) {
-            Timber.e(e);
+            Timber.i(e);
```
lognaturel, line 97: This is the top reporting issue in Firebase but it's expected because there could be an attempt to get the phone number before permissions were prompted for.

seadowg (approved):

PR #4316, opened 2020-12-16
Make sure external recording is blocked when internal recording in progress

### LINKED ISSUE #4285: Audio widgets are not blocked while the internal recorder is recording if the external app is set as default
Collect master

#### Problem description
Audio widgets are not blocked while the internal recorder is recording if the external app is set as default

1. External app for recording is used as a default
2. User starts form where audio quality is set
3. User records audio with audio quality defined
4. Other audio widgets are not blocked

ezgif com-video-to-gif(1)

#### Expected behavior
All audio widgets are blocked

seadowg:
Closes #4285 

New tests.

Comments inline.

Good to just check the issue is fixed and that external and internal recording still behave as expected.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppRecordingRequester.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppRecordingRequester.java
@@ -6,6 +6,8 @@ import android.provider.MediaStore;
 import android.util.Pair;
 import android.widget.Toast;
 
+import androidx.lifecycle.LifecycleOwner;
+
 import org.javarosa.form.api.FormEntryPrompt;
 import org.odk.collect.android.R;
 import org.odk.collect.android.analytics.AnalyticsEvents;
@@ -14,6 +16,7 @@ import org.odk.collect.android.listeners.PermissionListener;
 import org.odk.collect.android.utilities.ActivityAvailability;
 import org.odk.collect.android.utilities.ApplicationConstants;
 import org.odk.collect.android.utilities.PermissionUtils;
+import org.odk.collect.audiorecorder.recording.AudioRecorderViewModel;
 
 import java.util.function.Consumer;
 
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppRecordingRequester.java
```diff
@@ -24,13 +27,24 @@
...
         this.activityAvailability = activityAvailability;
         this.waitingForDataRegistry = waitingForDataRegistry;
         this.formEntryViewModel = formEntryViewModel;
+        this.audioRecorderViewModel = audioRecorderViewModel;
+        this.lifecycleOwner = lifecycleOwner;
+    }
+
+    @Override
+    public void onIsRecordingBlocked(Consumer<Boolean> isRecordingBlockedListener) {
+        audioRecorderViewModel.getCurrentSession().observe(lifecycleOwner, session -> {
```
seadowg, line 45: I did consider creating an abstract class that both `Internal...` and `External...` could extend from with this implementation. However, I feel like the code here **happens** to be the same as opposed to being the same. It could very well be that internal recordings are blocked for different reasons than external ones down the line.

Could totally understand if that seems wrong though!

lognaturel:
This seems like enough of an edge case that maybe we can bump to 1.30? I think it's important to get the release out as soon as possible so there's as much time to react to issues before we head out for end of year breaks.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ExternalAppRecordingRequester.java
lognaturel: I'm not sure we're likely to add conditions but the principle makes sense to me and there isn't much more complexity one way or the other.

lognaturel (approved):

lognaturel:
How about this -- I'll take out of the milestone for now so it doesn't look urgent. If @getodk/testers are able to finish regression testing before the end of one of their days, they can verify this one. But we should not extend the QA period just for it.

mmarciniak90:
**Tested with success**

Verified cases
- all audio widgets are blocked while the internal recorder is recording if the external app is set
- all audio widgets are blocked while the internal recorder is recording if the internal app is set
- delete recorded audio while the internal recorder is recording - audio widget is blocked
- minimization
- lock screen
- rotation

PR #4318, opened 2020-12-16
Fixed generating QR codes

### LINKED ISSUE #4312: The admin password is included in QR code even if it was unselected during generating QR code
Collect v1.28.3 (older release points were not verified)
Collect v1.27.0 works properly

#### Problem description
The admin password is included in QR code even if it was unselected during generating QR code

1. User set Aggregate as a server
2. Username and password to Aggregate are declared
3. Admin password is set
4. User opens QR code
5. User clicks on Contains sensitive information -> User unselect BOTH options
6. QR code is scanned by another device
7. Admin password is expected

#### Expected behavior
None of the passwords should be saved

grzesiek2010:
Closes #4312

I added automated tests and also tested the fix manually.

It's a bug fix and I can't think of any better solution.

It's a pretty safe fix that shouldn't affect other functionalities so we can just test the scenario described in the issue.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ConfigureWithQRCodeTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/ConfigureWithQRCodeTest.java
@@ -14,11 +14,9 @@ import androidx.work.WorkManager;
 import com.google.zxing.BarcodeFormat;
 import com.google.zxing.Result;
 import com.google.zxing.ResultPoint;
-import com.google.zxing.WriterException;
 import com.journeyapps.barcodescanner.BarcodeResult;
 import com.journeyapps.barcodescanner.DecoratedBarcodeView;
 
-import org.json.JSONException;
 import org.junit.After;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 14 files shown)


seadowg (approved):
Awesome! The test first approach made this really easy to review, and I liked that you pulled out the JSON generation into its own thing-doer/method object.

A couple of things that could be nice to improve upon although I don't feel need to go in now unless it seems like a priority to you:

* The `JsonPreferencesGenerator` could take the general and admin `SharedPreference` objects in as dependencies at the constructor (or as method args if that makes sense). Then the tests and the object itself don't need to care at all about the `GeneralSharedPreference` and `AdminSharedPreference` objects.
* The assertions on json could use the `org.json` parsing so you could check that keys don't exist etc (rather than having to use `containsString`. Not a big deal for now really as the tests you wrote drove the change out. Maybe just a trick for another time. Another thing I've seen people use a lot is JSON Hamcrest matcher libraries.

These are really just thoughts I had rather than things that need to change!

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified on Android 7.0, 8.1, 9.0, 10.0

Verified cases:
- generated/scanned QR code with both passwords unselected
- generated/scanned QR code with only Admin password unselected
- generated/scanned QR code with only Server password unselected
- generated/scanned QR code with both passwords selected
- light/dark mode
- English/Spanish language

PR #4320, opened 2020-12-16
Catch exceptions that might take place during opening uris

### LINKED ISSUE #4319: Value of URL widget should be lowercased before passing to intent constructor
#### Problem description

https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/180c16234b393c9cafa478d82823aa53?time=last-seven-days&versions=v1.29.0-beta.2%20(4026)&sessionEventKey=5FD8F75B022F00016340020A0EBAE189_1484987790261810522

```
Fatal Exception: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=Https://google.com/... (has extras) }
       at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2067)
       at android.app.Instrumentation.execStartActivity(Instrumentation.java:1727)
       at android.app.Activity.startActivityForResult(Activity.java:5320)
       at androidx.fragment.app.FragmentActivity.startActivityForResult(FragmentActivity.java:675)
       at android.app.Activity.startActivityForResult(Activity.java:5278)
       at androidx.fragment.app.FragmentActivity.startActivityForResult(FragmentActivity.java:662)
       at android.app.Activity.startActivity(Activity.java:5664)
       at androidx.core.content.ContextCompat.startActivity(ContextCompat.java:251)
       at androidx.browser.customtabs.CustomTabsIntent.launchUrl(CustomTabsIntent.java:331)
       at org.odk.collect.android.utilities.CustomTabHelper.openUri(CustomTabHelper.java:99)
       at org.odk.collect.android.widgets.UrlWidget.onButtonClick(UrlWidget.java:87)
       at org.odk.collect.android.widgets.UrlWidget.lambda$onCreateAnswerView$0(UrlWidget.java:48)
       at org.odk.collect.android.widgets.UrlWidget.lambda$onCreateAnswerView$0$UrlWidget(UrlWidget.java)
       at org.odk.collect.android.widgets.-$$Lambda$UrlWidget$BgNotkrxtDDsnKOwingp922zQms.onClick(-.java:2)
       at android.view.View.performClick(View.java:7448)
```

grzesiek2010:
Closes #4319

I reviewed implemented changes and tested manually opening websites from Collect. 

Opening uris is tricky we were adding tr-catch blocks and looks like we need one more. In this case using try-catch is maybe even better because I removed [getPackageName()](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4319?expand=1#diff-61121fde4ced9fab134f17741cfed84d344268989cda8ab7faeef2ca0c387b48L66) which might be faulty and might be the cause of the issue.

It's a safe fix we could even merge it without testing.

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
@@ -1,0 +1,33 @@
+package org.odk.collect.android.utilities;
+
+import android.net.Uri;
+
+import androidx.test.core.app.ApplicationProvider;
+
+import junit.framework.TestCase;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+@RunWith(RobolectricTestRunner.class)
+public class CustomTabHelperTest extends TestCase {
+
+    @Test
+    public void uriShouldBeNormalized() {
+        CustomTabHelper customTabHelper = spy(new CustomTabHelper());
+        doNothing().when(customTabHelper).openUriInChromeTabs(any(), any());
+        doNothing().when(customTabHelper).openUriInWebView(any(), any());
+        doNothing().when(customTabHelper).openUriInExternalBrowser(any(), any());
+
+        Uri uri = mock(Uri.class);
+        customTabHelper.openUri(ApplicationProvider.getApplicationContext(), uri);
+        verify(uri).normalizeScheme();
+    }
+}
--- a/collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
@@ -4,8 +4,6 @@ import android.net.Uri;
 
 import androidx.test.core.app.ApplicationProvider;
 
-import junit.framework.TestCase;
-
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricTestRunner;
@@ -17,7 +15,7 @@ import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
 @RunWith(RobolectricTestRunner.class)
-public class CustomTabHelperTest extends TestCase {
+public class CustomTabHelperTest {
 
     @Test
     public void uriShouldBeNormalized() {
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/CustomTabHelper.java
@@ -1,22 +1,15 @@
 package org.odk.collect.android.utilities;
 
-import android.content.ActivityNotFoundException;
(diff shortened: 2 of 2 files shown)


lognaturel (approved):
I'm comfortable merging as-is without QA but perhaps @seadowg sees it differently. I also don't mind waiting for v1.30.

collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.robolectric.RobolectricTestRunner;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+@RunWith(RobolectricTestRunner.class)
+public class CustomTabHelperTest extends TestCase {
```
seadowg: Don't think this needs to extend `TestCase`.

collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    public void uriShouldBeNormalized() {
+        CustomTabHelper customTabHelper = spy(new CustomTabHelper());
+        doNothing().when(customTabHelper).openUriInChromeTabs(any(), any());
+        doNothing().when(customTabHelper).openUriInWebView(any(), any());
+        doNothing().when(customTabHelper).openUriInExternalBrowser(any(), any());
+
+        Uri uri = mock(Uri.class);
+        customTabHelper.openUri(ApplicationProvider.getApplicationContext(), uri);
+        verify(uri).normalizeScheme();
```
seadowg, line 29: This verifies that `normalizedScheme()` is called but there's no guarantee that we'll use the result of that call in our `ACTION_VIEW` intent. So here I could change `uri = uri.normalizeScheme();` in `CustomTabHelper` to just `uri.normalizeScheme();` and the test would still pass. 

We could test this without the `spy` by passing in a Uri that would get normalized (like `HTTP://www.android.com`) and then verifying that we see a normalized version in the resulting intent (using `shadowOf(application).getNextStartedActivity()`). The complexity here is that we'd need to have the `openUriInChromeTabs` route fail, so we fall through into the path we want to check. For this we'd need to pull out a component to represent that we could set up to fail.

That all seems like a lot of work, but I'd generally say that if something is hard to test it's a sign it needs rework. I'm also not clear on what's happening with the Chrome tab stuff here and why we're doing it.

All that said this is a crash fix, so it'd be nice to get into the release, so I don't want my pretentiousness to get in the way too much 😉. Maybe we open another PR for the test, and we can discuss/tweak there?

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
grzesiek2010: Right, I generated the test class and it was added automatically. Fixed.

collect_app/src/test/java/org/odk/collect/android/utilities/CustomTabHelperTest.java
grzesiek2010: Right it's a silly test but yeah maybe we can improve it later in a separate pr.

seadowg (approved):

PR #4322, opened 2020-12-17
Waveform visual tweaks

seadowg:
This makes the waveform a little chunkier which I think helps it fit in visually and work on larger screens. There is also a small tweak to fix the extra spacing in the audio controller.

Just verified visually.

Not a lot to discuss here. These are opinionated changes though so will understand if others want a less chunky waveform.

This can go in without QA (as long as at least one other person doesn't feel it makes anything look worse).

### DIFF
--- a/collect_app/src/main/res/layout/waveform_layout.xml
+++ b/collect_app/src/main/res/layout/waveform_layout.xml
@@ -5,10 +5,9 @@
     android:layout_height="match_parent"
     app:chunkAlignTo="bottom"
     app:chunkColor="?colorSecondary"
-    app:chunkMaxHeight="80dp"
     app:chunkMinHeight="2dp"
     app:chunkRoundedCorners="true"
     app:chunkSoftTransition="true"
     app:chunkSpace="1dp"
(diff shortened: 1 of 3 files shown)


collect_app/src/main/res/layout/waveform_layout.xml
```diff
@@ -5,10 +5,9 @@
     android:layout_height="match_parent"
     app:chunkAlignTo="bottom"
     app:chunkColor="?colorSecondary"
-    app:chunkMaxHeight="80dp"
```
seadowg, line 8: Weird to duplicate this value here as it meant changing the height in the parent had no effect. The value is now set in `onLayout()`.

lognaturel (approved):
I have no strong feelings about this and am happy to make the change. I'll let it linger for a little bit to see if anyone else wants to chime in.

lognaturel:
@grzesiek2010 I'll explicitly tag you in case you have an opinion! I'm going to merge in ~30mins if I don't get any comments. This is not a big deal either way and we can always change again if we need to.

Proposed:
thick

Current:
thin

PR #4325, opened 2020-12-18
Fix crash when media file download fails

### LINKED ISSUE #4324: Crash when downloading media file fails
Collect v1.29.0

#### Problem description

The app crashes if there is an HTTP error when attempting to fetch media files.

1. Set up a server where a media file request will fail (with a 500 etc)
2. Download the form with the media file using Get Blank Form

The app will crash

#### Expected behavior

The user should see a failure message rather than a crash.

#### Other information 

Looks like this is because `OkHttpConnection.executeGetRequest` returns `null` in  the event of an HTTP error and our `ServerFormDownloader` is not accounting for a nullable result.

seadowg:
Closes #4324 

New tests.

The crash made a lot of sense given how `OkHttpConnection` is tested and behaved so really just needed to add a test to drive out accounting for the failure case in `OpenRosaFormSource`.

Just checking the issue is fixed is probably enough.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaFormSource.java
+++ b/collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaFormSource.java
@@ -100,13 +100,27 @@ public class OpenRosaFormSource implements FormSource {
     }
 
     @Override
-    public @NotNull InputStream fetchForm(String formURL) throws FormSourceException {
-        return fetchFile(formURL);
+    @NotNull
+    public InputStream fetchForm(String formURL) throws FormSourceException {
+        InputStream formFile = mapException(() -> openRosaXMLFetcher.getFile(formURL, null));
+
+        if (formFile != null) {
+            return formFile;
+        } else {
+            throw new FormSourceException(FETCH_ERROR);
+        }
     }
 
     @Override
+    @NotNull
     public InputStream fetchMediaFile(String mediaFileURL) throws FormSourceException {
-        return mapException(() -> openRosaXMLFetcher.getFile(mediaFileURL, null));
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaFormSource.java
```diff
@@ -100,13 +100,27 @@ public ManifestFile fetchManifest(String manifestURL) throws FormSourceException
...
+        }
     }
 
     @Override
+    @NotNull
     public InputStream fetchMediaFile(String mediaFileURL) throws FormSourceException {
-        return mapException(() -> openRosaXMLFetcher.getFile(mediaFileURL, null));
+        InputStream mediaFile = mapException(() -> openRosaXMLFetcher.getFile(mediaFileURL, null));
+
+        if (mediaFile != null) {
```
grzesiek2010: Wouldn't it be better to check this in mapException and throw `new FormSourceException(FETCH_ERROR)` there? Then it would be in one place.

seadowg: Yeah, I think you're right. It feels like a good place for some defensive programming.

lognaturel (approved):

lognaturel:
I still don't think this is new, right? Do you think it should go in the point release, @seadowg? I'm leaning towards yes since it does look very safe.

seadowg:
@lognaturel I was assuming it was new and was introduced during the server code rework, but I could be totally wrong. Either way I think it's good to go in.

grzesiek2010 (approved):

PR #4326, opened 2020-12-18
Always update formController for viewmodels when singleton is updated

### LINKED ISSUE #4327: ViewModels sometimes have a null form controller even though it's set at the Application level

lognaturel:
Closes #4327 

Some users are getting a crash that suggests their singleton `formController` is set but the `ViewModel`s don't know about it.

I searched for `Collect.getInstance().setFormController` and made sure that when it's called, `formControllerAvailable` is always called with the same value.

I also added commits improving logging that I saw while troubleshooting this.

I can't reproduce the issue so all I have done is visual inspection.

It doesn't make sense that those two updates would be different so I think this is a change we have to make. I also updated some logging. The other approach I considered was passing in `formController` to `moveForward` and `moveBackward` and logging if the view model one is null but the one passed in is not. I think that would be fine and perhaps we can do that too but I think this change actually addresses the root cause.

I initially wanted to also null out the `ViewModel`s' `formController`s when `           Collect.getInstance().setFormController(null)` is called. However, that's not necessary for this fix and would probably require introducing new methods for those view models. It feels weird that there might be a stale form controller set but we can come back to it.

I believe this to be very low risk. We are calling `formControllerAvailable` in more cases which seems it would always be positive.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2300,7 +2300,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                         formController.setLanguage(newLanguage);
                     } catch (Exception e) {
                         // if somehow we end up with a bad language, set it to the default
-                        Timber.e(e, "Ended up with a bad language. %s", newLanguage);
+                        Timber.i("Ended up with a bad language. %s", newLanguage);
                         formController.setLanguage(defaultLanguage);
                     }
                     Timber.i("Done in %.3f seconds.", (System.currentTimeMillis() - start) / 1000F);
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -570,7 +570,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                     formControllerAvailable(formController);
                     onScreenRefresh();
                 } else {
-                    Timber.d("Reloading form and restoring state.");
+                    Timber.w("Reloading form and restoring state.");
                     formLoaderTask = new FormLoaderTask(instancePath, startingXPath, waitingXPath);
                     showIfNotShowing(FormLoadingDialogFragment.class, getSupportFragmentManager());
                     formLoaderTask.execute(formPath);
@@ -581,6 +581,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
             // Not a restart from a screen orientation change (or other).
             Collect.getInstance().setFormController(null);
+
             Intent intent = getIntent();
             if (intent != null) {
                 loadFromIntent(intent);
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2297,7 +2301,7 @@ public void denied() {
                         formController.setLanguage(newLanguage);
                     } catch (Exception e) {
                         // if somehow we end up with a bad language, set it to the default
-                        Timber.e("Ended up with a bad language. %s", newLanguage);
+                        Timber.e(e, "Ended up with a bad language. %s", newLanguage);
```
lognaturel: I see this in a lot of our logs and I can't tell if it's actually meaningful. Hopefully seeing the exception will give us some insights. I'm hoping it turns out to be expected and we can just remove the logging.

grzesiek2010: It takes place every time I open the demo form. I think it's caused by language code, we have just label but it should be label::English (en) https://xlsform.org/en/#multiple-language-support

lognaturel: Yeah, I saw the stacktrace locally and realized it's really nothing so I should take this change out. 

It doesn't have to do with language codes because those are ignored in JR/Collect for now. I think it's just that because no language was explicitly chosen from the menu the database column for language has a null value.

lognaturel: To be explicit, I think it's important to clarify/clean up this logging because it can waste our time when troubleshooting logs on Crashlytics.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2282,6 +2283,8 @@ public void denied() {
                 t.destroy();
 
                 Collect.getInstance().setFormController(formController);
+                formControllerAvailable(formController);
```
grzesiek2010, line 2286: This change makes sense. I wonder if we should also maybe dismiss the loading dialog later once we set everything because now we do that at the beginning of this method what unblocks the UI and if the rest of the job we do in this method takes some time it's dangerous.

PR #4328, opened 2020-12-19
Temporarily get global formController from navigation methods

lognaturel:
Temporary fix that should prevent crashes from #4327 while we figure out when the form controller should be set.

Ran tests. Instrumented run: https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/6070831307506522034 `when_storageMigrationNotPerformed_shouldBePerformedAutomatically` fails but we've seen it be flaky before so I can't imagine it's related.

I thought #4326 was a pretty good idea but it doesn't work with the way the identity logging works. This made me pretty nervous about any kind of significant change here. This solution should be truly very low risk so I think it's best for now.

The other area I've taken an interest in is `onResume`. It seems like there might be a scenario where the form controller is still set at the App level but the viewmodels had to be recreated. I think in that case `formControllerAvailable` would never be called. I thought about adding a `else if (formController != null) { formControllerAvailable(formController) }` to `if (formController == null && formLoaderTask.getStatus() == AsyncTask.Status.FINISHED) {` but I'm not confident enough that this would be a fix.

The only intentional change is preventing a crash when navigation happens. I can't think of any specific risk but everything in this code is tricky. Worst case would be some kind of bad behavior related to navigation.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1473,7 +1473,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             }
 
             if (!saveBeforeNextView(formController)) {
-                formEntryViewModel.moveForward();
+                formEntryViewModel.moveForward(formController);
                 formIndexAnimationHandler.handle(formController.getFormIndex());
                 return true;
             } else {
@@ -1489,7 +1489,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 saveAnswersForCurrentScreen(DO_NOT_EVALUATE_CONSTRAINTS);
             }
 
-            formEntryViewModel.moveBackward();
+            formEntryViewModel.moveBackward(formController);
             formIndexAnimationHandler.handle(formController.getFormIndex());
             return true;
         }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2309,6 +2309,8 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 boolean pendingActivityResult = task.hasPendingActivityResult();
 
                 if (pendingActivityResult) {
+                    Timber.w("Calling onActivityResult from loadingComplete");
+
                     // set the current view to whatever group we were at...
                     onScreenRefresh();
                     // process the pending activity request...
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1473,7 +1473,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             }
 
             if (!saveBeforeNextView(formController)) {
-                formEntryViewModel.moveForward(formController);
+                formEntryViewModel.moveForward();
                 formIndexAnimationHandler.handle(formController.getFormIndex());
                 return true;
             } else {
@@ -1489,7 +1489,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 saveAnswersForCurrentScreen(DO_NOT_EVALUATE_CONSTRAINTS);
             }
 
-            formEntryViewModel.moveBackward(formController);
+            formEntryViewModel.moveBackward();
             formIndexAnimationHandler.handle(formController.getFormIndex());
             return true;
         }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2100,6 +2100,10 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
         FormController formController = getFormController();
 
+        if (formController != null && !formEntryViewModel.isFormControllerSet()) {
+            Timber.e("FormController set in App but not ViewModel");
+        }
+
         if (formLoaderTask != null) {
             formLoaderTask.setFormLoaderListener(this);
             if (formController == null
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -125,7 +127,12 @@ public boolean canAddRepeat() {
...
 
-    public void moveForward() {
+    public void moveForward(FormController formController) {
+        if (this.formController == null) {
+            Timber.w("Null formController");
+            this.formController = formController;
+        }
+
         try {
             formController.stepToNextScreenEvent();
```
lognaturel, line 139: All the subsequent calls end up using the parameter value. I assume that the singleton form controller is more likely to be the true, authoritative one. It might be possible for the `ViewModel`s' form controllers to be set but not the same as the singleton (though I really hope not).

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -125,7 +127,12 @@ public boolean canAddRepeat() {
         }
     }
 
-    public void moveForward() {
+    public void moveForward(FormController formController) {
+        if (this.formController == null) {
```
seadowg: I'm thinking instead of passing `FormController` in here we could just grab the singleton and then wrap that in an `requireFormController` method. That way we're not having to change the signature so hopefully not leaving anything around we'll want to take out later.

In my head, if this all worked nicely the structure would actually be pretty similar anyway:  the form load and the resulting `FormController` would be grabbed on demand from a cache anyway (but keyed/bucketed by the form entry session rather than an application singleton) so I could see us ending up with a structure where any method that needs it in the `ViewModel` just loads it in lazily.

I'll have a go at adding that in and push to this, and then we can discuss later.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -168,6 +176,13 @@ private String getFormIdentifierHash() {
         }
     }
 
+    private void ensureFormController() {
+        if (this.formController == null) {
+            Timber.w("Null formController");
```
seadowg: I'm wondering if we want to add any more logging here. It'd be great to try and get as much detail as possible when this happens.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -168,6 +180,13 @@ private String getFormIdentifierHash() {
         }
     }
 
+    private void ensureFormController() {
```
grzesiek2010, line 183: I'm not sure if I like this...
As I understand it's added to avoid NPE's in cases where we forget to set formController right?
so in collect if we call (in FormEntryActivity):
```
Collect.getInstance().setFormController(formController);
formControllerAvailable(formController);
```
it's not needed. It's needed if we forget the second line?

I'm not sure if it makes a lot of sense. We also might forget to call the first line then: (used in this method)
`Collect.getInstance().getFormController();`
will return null.

so it's still not a perfect way to avoid human mistakes. Additionally it makes the code more coupled.

seadowg: Yeah agreed that it's not full proof and increases coupling. I think really the ViewModels should be able to grab the FormController from some shared component that doesn't just live on `Collect`. I could imagine there being some `FormSessionRepository` or something that lets the ViewModel get the `FormController` lazily (it's loaded if it's not already in memory for instance). I'd be really keen to discuss and try and put the beginnings of something like that (or something better) in #4286.

I do think this is good enough to get out there and fix the crash as the change is small and isolated but yeah, lets not get used to it.

seadowg: As you point out as well it doesn't prevent us making any further mistakes. I think that's a great indicator that we need to find something better.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2100,6 +2100,10 @@ protected void onResume() {
 
         FormController formController = getFormController();
 
+        if (formController != null && !formEntryViewModel.isFormControllerSet()) {
```
seadowg, line 2103: Is this not really likely on opening a form as `onResume` would usually happen before `loadingComplete` (as the latter only gets called after the background task is finished)?

seadowg (approved):
Not 100% sure I understand the last piece of logging added but not a blocker for this going out.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: Prior to this crash I would have thought it impossible but I’m not so sure anymore. Figure it can’t hurt to confirm.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
lognaturel: Agreed with all this. @grzesiek2010 I wish we could just set the viewmodels’ formController exactly where we set the App’s but hopefully my explanation makes sense that we can’t without further restructuring because of an extra precondition on formcontrolleravailable. This is definitely meant to be temporary.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Fair enough!

lognaturel: I think we're going to see this happen after the log for `Calling onActivityResult from loadingComplete`. Because I think that branch is going to call `onActivityResult`, then return from `loadingComplete` and then `onResume` will be called.

Basically I think we were honing in on manifestations of the same problem at different parts of the lifecycle. I guess we'll find out shortly!

PR #4329, opened 2020-12-21
Update dependencies

grzesiek2010:
I reviewed implemented changes and ran tests. 

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -350,7 +350,7 @@ dependencies {
     testImplementation "org.mockito:mockito-core:3.6.28"
 
     testImplementation project(path: ':testshared')
-    testImplementation "org.robolectric:robolectric:4.4"
+    testImplementation "org.robolectric:robolectric:4.3.1"
     testImplementation 'androidx.test.ext:junit:1.1.2'
     debugImplementation("androidx.fragment:fragment-testing:${rootProject.fragmentVersion}") {
         exclude group: 'androidx.test', module: 'monitor' //fixes issue https://github.com/android/android-test/issues/731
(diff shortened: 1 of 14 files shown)


collect_app/build.gradle
```diff
@@ -350,7 +350,7 @@ dependencies {
     testImplementation "org.mockito:mockito-core:3.6.28"
 
     testImplementation project(path: ':testshared')
-    testImplementation "org.robolectric:robolectric:4.4"
+    testImplementation "org.robolectric:robolectric:4.3.1"
```
grzesiek2010, line 353: Robolectric 4.4 was a massive release and a lot of our tests would fail. This is a good candidate for a separate pr so I'll file a separate issue.

config/pmd-ruleset.xml
```diff
@@ -64,6 +66,7 @@
     </rule>
 
     <rule ref="category/java/design.xml">
+        <exclude name="AbstractClassWithoutAnyMethod" />
```
seadowg, line 69: What's the reason for these extra exclusions? I could go either way on `LiteralsFirstInComparisons` but `UnusedAssignment` and `AbstractClassWithoutAnyMethod` look useful maybe?

seadowg (changes requested):

config/pmd-ruleset.xml
grzesiek2010: I updated pmd and those are newly added rules. Maybe they are useful maybe not I just did not investigate it since it is beyond the scope of this pr. I usually take look at such things later.

seadowg: Ah fair enough!

seadowg (approved):

PR #4330, opened 2020-12-21
Prevent crash when updating buttons on Main Menu

seadowg:
This reworks the main menu button code to not use deprecated Cursor requerying and Activity management. Instead, Cursors are just opened and closed as needed. 

I'd advise going through this commit by commit as the first two commits just add a way of preventing (and getting more info about) the crash where as the last two rework the code. If the rework is deemed to risky we could just go with the first two commits.

Verified manually and ran tests. The button counts are actually tested in quite a lot of places (finalizing forms, sending forms etc) which I confirmed by killing the code and then running tests (~10 failures). It'd be good to look at further refactor and tests (probably pulling the counts out into a ViewModel and working out whether we need a content observer here or not for example) but that feels like a job for another day.

Comment inline.

Should just prevent the crash which we don't have repro for, so I think checking the button counts update as expected should be the focus.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
@@ -466,7 +466,8 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
                 Timber.w("Cannot update \"View Sent\" button label since the database is closed. Perhaps the app is running in the background?");
             }
         } catch (SQLiteDiskIOException e) {
-            Timber.e(e);
+            boolean migrationBeingPerformed = storageMigrationRepository.isMigrationBeingPerformed();
+            Timber.e(e, "Storage migration being peformed: %s", migrationBeingPerformed);
         }
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
@@ -270,7 +270,6 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
     protected void onResume() {
         super.onResume();
 
-        countSavedForms();
         updateButtons();
         if (!storageMigrationRepository.isMigrationBeingPerformed()) {
             getContentResolver().registerContentObserver(InstanceColumns.CONTENT_URI, true, contentObserver);
@@ -423,6 +422,8 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
     }
 
     private void updateButtons() {
+        countSavedForms();
+
         try {
             if (finalizedCursor != null && !finalizedCursor.isClosed()) {
                 finalizedCursor.requery();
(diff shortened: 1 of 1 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -96,12 +96,7 @@
     private Button getFormsButton;
     private AlertDialog alertDialog;
     private MenuItem qrcodeScannerMenuItem;
-    private int completedCount;
```
seadowg, line 99: We don't need to keep any of this state in the Activity any longer.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -422,46 +374,50 @@ public void onClick(DialogInterface dialog, int i) {
     }
 
     private void updateButtons() {
-        if (finalizedCursor != null && !finalizedCursor.isClosed()) {
-            finalizedCursor.requery();
```
seadowg, line 426: `requery` is deprecated. The advice is to just open a new cursor.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -358,48 +352,6 @@ private void initToolbar() {
...
-
-        //count for view sent form
-        try {
-            viewSentCursor = instancesDao.getSentInstancesCursor();
-        } catch (Exception e) {
-            createErrorDialog(e.getMessage(), EXIT);
-            return;
-        }
-        if (viewSentCursor != null) {
-            startManagingCursor(viewSentCursor);
```
seadowg, line 398: `startManagingCursor` is deprecated. Instead of doing this I've switched it to using `try-with-resource` to manage opening/closing.

lognaturel (changes requested):
I'm totally with you for the first 3 commits. But I'm having a little trouble with the 4th. It means that if the migration were running and `updateButtons` were called, an error dialog would be shown to the user, right? Seems to me that you had the right idea before with silently catching that exception. I'd just let the counts be 0 in that case and next time the user gets back to the main menu the buttons would be updated using the new counts. I think this is the right thing to do for any exception that comes out of getting those counts because it's not essential functionality. We can do a `Timber.e` to track frequency if you want but even that doesn't seem critical. If there's a real db issue then it'll manifest when the user tries to get a form list, open a form, etc.

seadowg:
> I'm totally with you for the first 3 commits. But I'm having a little trouble with the 4th. It means that if the migration were running and updateButtons were called, an error dialog would be shown to the user, right?

Not my intention. I just moved the error handling over that was there before in `countSavedForms` when it did the db queries. I wouldn't expect the crash we were seeing to happen again as the 3rd commit removes the code that was holding onto `Cursors`.  Totally agree that `catch` clauses probably don't even need to be there though. Let's chat this through and work out which version we want to put out there.

lognaturel:
@seadowg and I talked about adding a (maybe overly) defensive `!storageMigrationRepository.isMigrationBeingPerformed()` wrap around `updateButtons` and also ignoring any exceptions and letting counts be 0.

lognaturel (approved):

PR #4333, opened 2021-01-06
Make sure form is created in correct place during download

### LINKED ISSUE #4332: Form showing after media files download failure
Collect v1.29.2 Android v9 (2020-11) LG K12+

#### Problem description

If the download of one, more or all of the media files fails, the form still appears in the list of blank forms to be filled out. 

Well... If the media files are important for filling out the electronic form, the electronic form should not be displayed if there is a connection failure during form and media files download.

Create a form with three large media files.
During media download, put your phone on airplane mode to receive a failure message.
Press Fill Blank Form.
Despite the connection failure, the incomplete form will appear.

#### Expected behavior

The electronic form should not be displayed if there is a connection failure during form and media files download.

#### Other information 

This should be treated just like a cancelled form download - remove the form and any associated media files.

seadowg:
Closes #4332

New tests.

~~I'm going to take the afternoon to try and improve the solution further as I'm not a fan of the fact that we have to clean up in the error case here. I think this solution is good enough to go in a point release though, so I don't want to wait on that (or worry about any changes/refactoring adding risk).~~

~~It looks like the original problem stems from the fact that the form file is created in Collect's forms dir rather than in the temp dir created for download which is too early. Really instead the form file should be created in the temp dir and then copied at the end. In that world we could avoid problems like this as the temp dir is always cleaned up.~~

EDIT: I ended up having to do all this anyway.

Good to check downloading forms with and without media files works as expected (in failure cases as well). I think paying special attention to the end state of the forms directory is important as well. 

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -142,10 +142,16 @@ public class ServerFormDownloader implements FormDownloader {
                 fileResult = downloadXform(fd.getFormName(), fd.getDownloadUrl(), stateListener, tempDir, formsDirPath);
 
                 if (fd.getManifest() != null) {
-                    finalMediaPath = FileUtils.constructMediaPath(
-                            fileResult.getFile().getAbsolutePath());
-                    String error = downloadManifestAndMediaFiles(tempMediaPath, finalMediaPath, fd,
-                            stateListener, fd.getManifest().getMediaFiles(), tempDir);
+                    finalMediaPath = FileUtils.constructMediaPath(formsDirPath + File.separator + fileResult.file.getName());
+                    String error = downloadManifestAndMediaFiles(
+                            tempMediaPath,
+                            finalMediaPath,
+                            fd,
+                            stateListener,
+                            fd.getManifest().getMediaFiles(),
+                            tempDir
+                    );
+
                     if (error != null && !error.isEmpty()) {
                         success = false;
                     }
@@ -155,11 +161,8 @@ public class ServerFormDownloader implements FormDownloader {
             } catch (InterruptedException e) {
                 Timber.i(e);
                 cleanUp(fileResult, tempMediaPath);
-
-                // do not download additional forms.
                 throw e;
             } catch (FormSourceException | IOException e) {
-                cleanUp(fileResult, tempMediaPath);
                 return false;
             }
 
@@ -212,18 +215,22 @@ public class ServerFormDownloader implements FormDownloader {
 
         boolean installEverything(String tempMediaPath, FileResult fileResult, Map<String, String> parsedFields, String formsDirPath) {
             UriResult uriResult = null;
+
             try {
-                uriResult = findExistingOrCreateNewUri(fileResult.file, parsedFields, formsDirPath);
-                if (uriResult != null) {
-                    // move the media files in the media folder
-                    if (tempMediaPath != null) {
-                        File formMediaPath = new File(uriResult.getMediaPath());
-                        FileUtils.moveMediaFiles(tempMediaPath, formMediaPath);
-                    }
-                    return true;
-                } else {
-                    Timber.w("Form uri = null");
+                // Copy form to forms dir
+                File formFile = new File(formsDirPath, fileResult.file.getName());
+                FileUtils.copyFile(fileResult.file, formFile);
+
+                // Save form in database
+                uriResult = findExistingOrCreateNewUri(formFile, parsedFields, formsDirPath);
+
+                // move the media files in the media folder
+                if (tempMediaPath != null) {
+                    File formMediaPath = new File(uriResult.getMediaPath());
+                    FileUtils.moveMediaFiles(tempMediaPath, formMediaPath);
                 }
+
+                return true;
             } catch (IOException e) {
                 Timber.e(e);
 
@@ -235,8 +242,9 @@ public class ServerFormDownloader implements FormDownloader {
                             null, null);
                     Timber.d("Deleted %d rows using uri %s", deletedCount, uri.toString());
                 }
+
+                return false;
             }
-            return false;
         }
 
         private void cleanUp(FileResult fileResult, String tempMediaPath) {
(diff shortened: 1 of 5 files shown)


collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
```diff
@@ -110,4 +112,16 @@ public void restore_marksDeletedAsFalse() {
         formsRepository.restore(1L);
         assertThat(formsRepository.get(1L).isDeleted(), is(false));
     }
-}
\ No newline at end of file
+
+    @Test
+    public void save_addsHashBasedOnFormFile() {
```
seadowg, line 126: This was needed for the new test to work properly.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
```diff
@@ -85,6 +86,49 @@ public void downloadsAndSavesForm() throws Exception {
         assertThat(new String(read(formFile)), is(xform));
     }
 
+    @Test
+    public void whenFormToDownloadIsUpdate_savesNewVersionAlongsideOldVersion() throws Exception {
```
seadowg, line 90: I was nervous about the update path not be covered here.

seadowg:
Putting this back to a draft as I think I've found an edge case that causes the form to get deleted. I want to add a test and try a simple fix but I think we might need a bit more of a rework than I hoped to get this in a good place.

seadowg:
Ok looks like the issue is existing (it's in v1.28.4). Because of the underlying problem if you cancel while redownloading (not an update) a form you'll end up deleting the existing one. On this branch there is an additional problem where if we redownload a form (again not an update) but the manifest or media file fetches fail we'll also end up deleting the form. 

As I wrote the above I realized that we need to just fix this properly which I'll continue working on. No quick hack here 😭. The issue in v1.28.4 is pretty recoverable, but it would be easy with the changes introduced in this branch to delete a form that you then wouldn't be able to download again (the failure could be because the manifest/media file endpoint is broken).

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -301,41 +311,25 @@ private Uri saveNewForm(Map<String, String> formInfo, File formFile, String medi
...
-            int i = 2;
-            File tempFormFile = new File(path);
-            while (tempFormFile.exists()) {
-                path = formsDirPath + File.separator + rootName + "_" + i + ".xml";
-                tempFormFile = new File(path);
-                i++;
-            }
-
-            InputStream file = formSource.fetchForm(url);
-            writeFile(tempFormFile, stateListener, file, tempDir);
```
seadowg, line 318: This was the key problem: `tempFormFile` is defined using the `formsDirPath` instead of `tempDir`.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -210,18 +215,22 @@ private boolean isSubmissionOk(Map<String, String> parsedFields) {
...
-                    // move the media files in the media folder
-                    if (tempMediaPath != null) {
-                        File formMediaPath = new File(uriResult.getMediaPath());
-                        FileUtils.moveMediaFiles(tempMediaPath, formMediaPath);
-                    }
-                    return true;
-                } else {
-                    Timber.w("Form uri = null");
+                // Copy form to forms dir
+                File formFile = new File(formsDirPath, fileResult.file.getName());
```
seadowg, line 221: We now need to copy the file at the end. This was missing before, but we got away with it because the form was created in the final destination instead of the temp directory.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -141,10 +142,16 @@ public boolean processOneForm(ServerFormDetails fd, FormDownloaderListener state
...
 
                 if (fd.getManifest() != null) {
-                    finalMediaPath = FileUtils.constructMediaPath(
-                            fileResult.getFile().getAbsolutePath());
-                    String error = downloadManifestAndMediaFiles(tempMediaPath, finalMediaPath, fd,
-                            stateListener, fd.getManifest().getMediaFiles(), tempDir);
+                    finalMediaPath = FileUtils.constructMediaPath(formsDirPath + File.separator + fileResult.file.getName());
+                    String error = downloadManifestAndMediaFiles(
+                            tempMediaPath,
+                            finalMediaPath,
```
lognaturel, line 148: Is this an example of where we might be able to have a `saveMediaFile` method on the `FormsRepository` that takes in a media filename, form_id, and file contents instead of building the path here? Then `downloadManifestAndMediaFiles` can take a `FormsRepository` instead.

seadowg: Yeah I think so! I'm thinking `downloadManifestAndMediaFiles` would be a lot simpler if it used the repositories.

lognaturel (approved):

mmarciniak90:
@seadowg I checked this fix for Central and everything works great, but then I started checking the same for Google Drive and the problem is still visible there. I see form on `Fill Blank Form` list even if download failed

seadowg:
@mmarciniak90 does the same thing happen in earlier versions?

mmarciniak90:
Yes (checked v1.28), this isn't regression. This fix improved the behavior for Central and Aggregate but Google Drive stayed unchanged. So the behavior is not identical for all server types.

lognaturel:
Since currently Google Drive uses a different code path, I think we should file the issue separately and not deal with it for the point release. Unless it’s already done, @seadowg!

seadowg:
Agreed! Would definitely want to look at that separately.

mmarciniak90:
**Tested with success**

Verified cases:
- manual downloading -> turning on flight mode -> form is not available
- match exactly -> turning on flight mode -> form is not available
- downloading from Central
- downloading from Aggregate
- downloading from Google Drive - issue is still visible and will be fixed separately 
- downloading form with media
- downloading forms without media
- re-download
- downloading the latest form version 
- cancel the download

PR #4334, opened 2021-01-06
Clean up ReferenceManager if form can't be parsed

### LINKED ISSUE #4315: Reset referenceManager after failed attempt to get form metadata
Collect v1.27 and prior

#### Problem description
If there's a crash in the call on FileUtils.getMetadataFromFormDefinition, the referenceManager is never reset and then forms with media can't be loaded until Collect is killed and restarted.

Try to load a form that Collect can't parse such as the one described in https://github.com/getodk/javarosa/issues/616

#### Expected behavior
The referenceManager should always be reset

lognaturel:
Closes #4315

Confirmed that the test failed prior to the fix. Tried it with the form in the issue.

I think I previously had left this aside because it didn't seem very common and I felt like some refactoring was in order. However, #4332 highlights that users really do run into issues downloading form attachments and that can potentially result in this bad state. I also got a private support request about the issue and saw some missing file errors in Crashlytics that made me think other users were hitting this state.

Since I think this should go in a point release, I went for as safe and straightforward of a solution as possible. This means a little bit of redundancy but I think it's worth it for the simplicity. 

This is only intended to fix the bug. The code path changed affects all the different ways there are to download form definitions (manually, auto-update, match exactly).

See issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMetadataParser.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMetadataParser.java
@@ -10,6 +10,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
+import timber.log.Timber;
+
 import static org.odk.collect.android.utilities.FileUtils.LAST_SAVED_FILENAME;
 import static org.odk.collect.android.utilities.FileUtils.STUB_XML;
 import static org.odk.collect.android.utilities.FileUtils.write;
@@ -31,7 +33,16 @@ public class FormMetadataParser {
         referenceManager.addReferenceFactory(new FileReferenceFactory(mediaDir.getAbsolutePath()));
         referenceManager.addSessionRootTranslator(new RootTranslator("jr://file-csv/", "jr://file/"));
 
-        HashMap<String, String> metadata = FileUtils.getMetadataFromFormDefinition(file);
+        HashMap<String, String> metadata;
+        try {
+            metadata = FileUtils.getMetadataFromFormDefinition(file);
+        } catch (Exception e) {
+            referenceManager.reset();
+            tmpLastSaved.delete();
+            Timber.e(e);
+
+            throw e;
+        }
         referenceManager.reset();
         tmpLastSaved.delete();
 
--- a/collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
@@ -103,7 +103,6 @@ public class FormMetadataParserTest {
     }
 
     @Test
-    @SuppressWarnings("PMD.JUnitUseExpected")
     public void cleansUpReferenceManager() throws Exception {
         File formXml = File.createTempFile("form", ".xml");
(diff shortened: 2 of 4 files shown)


collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -122,6 +122,27 @@ public void cleansUpReferenceManager() throws Exception {
...
 
+    @Test
+    @SuppressWarnings("PMD.JUnitUseExpected")
+    public void cleansUpReferenceManagerAfterFail() throws Exception {
+        File formXml = File.createTempFile("form", ".xml");
+
+        FormMetadataParser formMetadataParser = new FormMetadataParser(referenceManager);
+
+        try {
+            formMetadataParser.parse(formXml, mediaDir);
```
seadowg, line 131: Should be a `fail` at the end of the `try` I think. Otherwise, we could end up with a confusing failure later.

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -122,6 +122,27 @@ public void cleansUpReferenceManager() throws Exception {
         }
     }
 
+    @Test
+    @SuppressWarnings("PMD.JUnitUseExpected")
```
seadowg: If you agree I think we can suppress this for the whole project. `expected` is pretty useless as it doesn't let you test failure cases (like here) or cases where you need to check the details of an exception. It can also lead to incorrectly passing tests as you're not specifying where the exception should come from, just that the test should throw it at some point.

I didn't realize how much I disliked `expected` until I started writing this 😂

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMetadataParser.java
```diff
@@ -31,7 +33,16 @@ public FormMetadataParser(ReferenceManager referenceManager) {
         referenceManager.addReferenceFactory(new FileReferenceFactory(mediaDir.getAbsolutePath()));
         referenceManager.addSessionRootTranslator(new RootTranslator("jr://file-csv/", "jr://file/"));
 
-        HashMap<String, String> metadata = FileUtils.getMetadataFromFormDefinition(file);
+        HashMap<String, String> metadata;
+        try {
+            metadata = FileUtils.getMetadataFromFormDefinition(file);
+        } catch (Exception e) {
```
seadowg, line 39: This could just be a `try-finally` I think?

seadowg (changes requested):
Just a couple of little tweaks

lognaturel:
@seadowg, thanks -- force-pushed addressing your comments.

seadowg (approved):

mmarciniak90:
**Tested with success**

Before fix -> I tried to download `select-one-file` form and it failed, other forms were downloaded -> after that, media were not attached/applied to other forms.
After fix -> I tried to download `select-one-file` form and it failed, other forms were downloaded -> after that, media were available in other forms 

Verified cases:
- form image files
- form audio files
- form video files
- manual download forms with media
- auto-update form with new media
- match exactly
- downloaded from Central and Google Drive

PR #4337, opened 2021-01-08
Do not allow to perform rest of the code in onCreate() if not exactly one matching form was detected

### LINKED ISSUE #4336: NullPointerException in FormMapViewModel.getFormTitle(FormMapViewModel.java:43)
Collect v1.27+

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/27c36d6d10118a0367d013e9bf34592a?time=last-thirty-days&versions=v1.29.2%20(4041)&sessionEventKey=5FF7AFE902960001643FFBE9E19AF159_1493637213128445742

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.odk.collect.android.forms.Form.getDisplayName()' on a null object reference
       at org.odk.collect.android.activities.viewmodels.FormMapViewModel.getFormTitle(FormMapViewModel.java:43)
       at org.odk.collect.android.activities.FormMapActivity.onCreate(FormMapActivity.java:119)
       at android.app.Activity.performCreate(Activity.java:7963)
       at android.app.Activity.performCreate(Activity.java:7952)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1306)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3505)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3689)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:140)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:100)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2239)
```

grzesiek2010:
Closes #4336

I tested the behavior hardcoding false in https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4336?expand=1#diff-8dc5817ca05b86da0282112ffd78f381a37cc6e9061c194d4ac355e6dc5f4bcdR105

The problem was that calling finish() is not enough and we need to call return to avoid calling other code in a method.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -108,6 +108,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
             // This shouldn't be possible because the instance URI had to be used in the calling
             // activity but it has been logged to Crashlytics.
             finish();
+            return;
         }
 
         if (viewModelFactory == null) { // tests set their factories directly
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
@@ -104,7 +104,7 @@ public class FormMapActivity extends BaseGeoMapActivity {
         List<Form> forms = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData()));
         if (forms.size() == 1) {
             form = forms.get(0);
-        } else {
+        } else if (viewModelFactory == null) {
             // This shouldn't be possible because the instance URI had to be used in the calling
             // activity but it has been logged to Crashlytics.
             finish();


grzesiek2010:
@lognaturel I think we can add this fix to v1.29.3 without testing.
The only thing is that maybe we should consider adding a toast when this case takes place? Otherwise we click on the map icon and nothing happens what might be confusing.

lognaturel:
Is it possible maybe that this issue comes from a misunderstanding of the lifecycle of intents? It's making the assumption that the intent that the activity was launched by will continue to exist even after activity/process death. Is that maybe just not the case and we should be saving `form` in `savedInstanceState`?

EDIT: if that were the case I'd expect to be able to reproduce and I can't.

lognaturel:
> we click on the map icon and nothing happens what might be confusing

~I'm a confused by the stacktrace below `Instrumentation.callActivityOnCreate`. `Instrumentation`? Is this real users or some kind of automation that's hitting this?~ I guess that's how activities are launched? Still, are you sure that this problem shows up after clicking the map icon/ on first launch of the activity?

lognaturel:
Yikes, the logs are really strange. See [this one](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/27c36d6d10118a0367d013e9bf34592a?time=last-ninety-days&sessionEventKey=5F9C0F7502A2000120F36C3E41FE2DAA_1467844132319835769) or [this one](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/27c36d6d10118a0367d013e9bf34592a?time=last-ninety-days&sessionEventKey=5F9A940D0270000118B73AA6AE73E9AA_1467426972242572069) where the crash happens immediately after launch. It looks like no activity was even visited. I have no idea what that's about.

Or [this one](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/27c36d6d10118a0367d013e9bf34592a?time=last-ninety-days&sessionEventKey=5F99368601B800013738CD3920BC982E_1467042935523415634) or [this one](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/27c36d6d10118a0367d013e9bf34592a?time=last-ninety-days&sessionEventKey=5F97CCBD0153000145AE07F5C28EC483_1466653722562160001) where the activity displayed before the crash is not `FillBlankFormActivity`. 

I'm fine with adding the `return`. I don't think we should add a toast or anything.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -104,10 +104,11 @@
         List<Form> forms = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData()));
         if (forms.size() == 1) {
             form = forms.get(0);
-        } else {
+        } else if (viewModelFactory == null) {
```
lognaturel, line 107: Why add this condition? I think we want to exit in any other case. For example, it could be possible for no form to match the id.

lognaturel: `viewModelFactory` should only be `null` in tests and in that case it will always be `null`.

lognaturel (changes requested):
See inline.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
lognaturel: Oh, I see. It's `null` in the NON-test case and so you're avoiding this branch for tests. But can't you instead make sure the tests always have exactly one form so the `else` branch doesn't get hit?

lognaturel (approved):
Right, I see what's going on, now. The `else` branch is only relevant if we're not in a test context because in a test context the viewmodel and its form will have been manually supplied.

As you suggest, let's get it in the point release so we can see if that reveals another crash.

PR #4339, opened 2021-01-11
Log server and parse errors during Match Exactly sync

### LINKED ISSUE #4044: Add extra info to Match Exactly error analytics
In a lot of cases when there is failure during the Match Exactly "sync" we just log (to analytics) that there was a `FETCH_ERROR`. Adding more info here would let us get a better sense of what problems people are running into. To do this we need to investigate all the different situations that can cause a `FETCH_ERROR` (in `OpenRosaFormListApi`) and then decide on a set of new error types to add. 

Looking at a quick scan of what can go wrong it seems the errors can be:

* http errors: we get an error message (and probably a non 200 status code) back from the server. We already have an `AUTH_REQUIRED` error type that's specific to `401` responses.
* parse errors: something in the formlist or manifest response isn't right
* IO exceptions: this is mostly covered by catching `UnknownHostException` and `SSLException` but we'd need to investigate to see what other errors can come out of OkHttp

Given that I think it would make sense to add `SERVER_ERROR` and `PARSE_ERROR` types to `FormApiException` that we also log out to analytics and leave the other IO exceptions under the `FETCH_ERROR`  bucket for now. We should also log the status code with the `SERVER_ERROR`.

seadowg:
Closes #4044

This also includes so refactoring in `ServerFormDownloader` that hopefully makes it easier to read/work with. I'd advise reviewing comment by commit.

Ran tests.

We did consider switching away from the deprecated `log` method for analytics but doing so would make it harder to understand the events over time (as old and new versions of the app would be logging different events). For the moment we'll continue to use the deprecated method for old events.

Due to the changes to `ServerFormDownloader` it would be good to check downloading forms in general (with and without media files, error cases etc).


collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDetails.java
```diff
@@ -24,20 +24,16 @@
...
     private final String formVersion;
     private final String hash;
     private final boolean isNotOnDevice;
     private final boolean isUpdated;
     private final ManifestFile manifest;
 
-    public ServerFormDetails(String formName, String downloadUrl, String manifestUrl, String formID,
-                             String formVersion, String hash,
-                             boolean isNotOnDevice, boolean isUpdated, ManifestFile manifest) {
+    public ServerFormDetails(String formName, String downloadUrl, String formID, String formVersion, String hash, boolean isNotOnDevice, boolean isUpdated, ManifestFile manifest) {
```
seadowg, line 34: We don't need a manifest url in here any longer as the manifest is fetched before creating this object.

lognaturel (changes requested):
I think while we're here, "The server (`<url>`) returned status code XXX" and "The server (`<url>`) provided an invalid response" followed by the standard "if it continues, talk..." would be helpful in `FormSourceExceptionMapper`.

I'm happy other than that so please mark as `needs testing` once that's in. It will be hard to try the error cases in QA since that would require a server with bad responses. @seadowg's comment about generally doing regression checks on download seem right.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- download forms with success
- download forms with and without media
- download a lot of forms
- download failed because required credentials are missing
- download failed because of invalid credentials
- download failed because invalid URL
- plane mode turned on during downloading
    - Android 10 returned status code 0
- no internet connection
- downloading from Aggregate
- downloading from Central
    - status code 400 when URL has a letter at the end instead of project number
    - invalid project number
- downloading from KOBO
- downloading from ONA
- local server - download forms, send submissions, match exactly
   - HTTPS issue - expected error
   - wrong port number - status code 0
- regression check for Google Drive

kkrawczyk123:

lognaturel:
One thing I notice is that "status code 0" is not very helpful since that's not an HTTP code. Haven't looked in detail what an easy alternative would be. Maybe identifying 0 and throwing a `Unreachable` instead? ~(Side note, I notice that should really be `UnreachableError`)~

seadowg:
@lognaturel @mmarciniak90 hmm yeah I'll check out and try and fix those status 0 cases before merging.

seadowg:
@mmarciniak90 are you able to post reproduction steps for one of the status 0 cases? I'm having a little trouble making it happen.

mmarciniak90:
> are you able to post reproduction steps for one of the status 0 cases?

I needed to investigate a bit to define repro steps. @seadowg please, try these steps:

Steps:
1. Default server (where the anonymous user is allowed) is set and march exactly is turned on
2. Sync is finished with success
3. I changed the default server to https://agg-17.appspot.com where the anonymous user is NOT allowed. I added credentials.
4. I opened the `Fill Blank Form` page and clicked the refresh button, then I turned on plane mode

seadowg:
Looks like the problem was timeouts. I've fixed it, so you should now see the generic error message when this happens.

@mmarciniak90 Definitely useful if you could check the cases you found out again to make sure we're not seeing that 0 status code anywhere.

@lognaturel I think we may want an error message/analytics event for timeouts. What do you think? I think if so we can create an issue so as not to hold this up.

kkrawczyk123:
@seadowg I have re-tested:
- status code 0 not visible for both cases: wrong port number on local aggregate and airplane mode scenario
- quick regression check on other errors 

lognaturel:
> I think we may want an error message/analytics event for timeouts.

Yes! A separate issue sounds great.

lognaturel (approved):

PR #4350, opened 2021-01-13
Allow other appearances to combine with field-list

### LINKED ISSUE #4349: Allow field-list to combine with other appearances
All versions

#### Problem description
`field-list` can't combine with other appearances. There are no appearances that it can combine with that are recognized by Collect but it can be convenient to combine it with Enketo grid theme appearances for use on both platforms. For example a group with appearance `w6 field-list` would be rendered as a 6-wide row in the Enketo grid theme and a single screen in Collect.

Create a group with appearance `field-list` combined with others, e.g. [example.xml.txt](https://github.com/getodk/collect/files/5812391/example.xml.txt).

#### Expected behavior
`field-list` should be respected even if there are other appearances.

lognaturel:
Closes #4349

Wrote and ran automated tests. Tried form in issue.

No real alternatives considered.

The only intended change is to fix the limitation. Since I ended up pulling out some redundant code there's risk around any forms with field lists and especially forms with groups nested in field lists.

See issue. A form with groups nested in field lists would be good too.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -58,6 +58,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 
 import timber.log.Timber;
 
@@ -371,11 +372,11 @@ public class FormController {
     private boolean groupIsFieldList(FormIndex index) {
         // if this isn't a group, return right away
         IFormElement element = formEntryController.getModel().getForm().getChild(index);
-        if (!(element instanceof GroupDef)) {
+        if (!(element instanceof GroupDef) || element.getAppearanceAttr() == null) {
             return false;
         }
 
-        return ODKView.FIELD_LIST.equalsIgnoreCase(element.getAppearanceAttr());
+        return element.getAppearanceAttr().toLowerCase(Locale.ENGLISH).contains(ODKView.FIELD_LIST);
     }
 
     private boolean repeatIsFieldList(FormIndex index) {
@@ -437,8 +438,6 @@ public class FormController {
         } else if (event == FormEntryController.EVENT_REPEAT) {
             return repeatIsFieldList(index);
         } else {
-            // right now we only test Questions and Groups. Should we also handle
-            // repeats?
             return false;
         }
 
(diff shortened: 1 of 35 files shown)


collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -437,8 +438,6 @@ public boolean indexIsInFieldList(FormIndex index) {
         } else if (event == FormEntryController.EVENT_REPEAT) {
             return repeatIsFieldList(index);
         } else {
-            // right now we only test Questions and Groups. Should we also handle
```
lognaturel, line 440: This comment is a lie. Repeats are handled right above.

collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
```diff
@@ -371,11 +372,11 @@ private FormInstance getInstance() {
     private boolean groupIsFieldList(FormIndex index) {
         // if this isn't a group, return right away
         IFormElement element = formEntryController.getModel().getForm().getChild(index);
-        if (!(element instanceof GroupDef)) {
+        if (!(element instanceof GroupDef) || element.getAppearanceAttr() == null) {
             return false;
         }
 
-        return ODKView.FIELD_LIST.equalsIgnoreCase(element.getAppearanceAttr());
+        return element.getAppearanceAttr().toLowerCase(Locale.ENGLISH).contains(ODKView.FIELD_LIST);
```
grzesiek2010: It might be a good opportunity to move this const (FIELD_LIST) to `WidgetAppearanceUtils` where we store all appearances? That name `WidgetAppearanceUtils` might require refactoring.

lognaturel: Thanks! I did so and renamed to `Appearances` following the Java pattern from `Files`, `Collections`, etc. We're not very consistent with naming utilities and I also considered `AppearanceUtils`. I prefer `Appearances` because I think it's really nice when accessing the constants and pretty good for method calls, too.

grzesiek2010 (approved):

mmarciniak90:
@lognaturel we tested this issue but there are conflicting files. Please let us know if resolving conflicts may pose any risk.

Verified cases:
* number before and after `field-list`
* extra space before and after `field-list`
* double `field-list`
* special characters
* appearance="ield-list" - single question on page
* lowercase and uppercase
* appearance="field field-list"
* appearance="list field-list"
* appearance="w6 field-list"
* form with groups nested in field lists `NestedTest.txt`
* rank widgets in field list
* form with field-list where each question revealing the next when the former's value is set
* regression check on all forms with field-list 
* read-only with field-list

lognaturel:
Thank you! The commit that conflicted was the second one which renamed the `Appearances` class. It sometimes overlapped with the change from `PermissionsUtils` to `PermissionsProvider`. The diff was easy to deal with and feels really low risk.

kkrawczyk123:

Issue #4214, opened 2020-11-09
Add support for external binary question types


lognaturel:
https://github.com/getodk/collect/pull/4188 added support for binary files to be returned by external apps to intent field lists. This issue is to allow single-question media questions to call out to external applications. 

[Current documentation](https://docs.getodk.org/launch-apps-from-collect/#designing-an-external-app-to-return-a-single-value-to-collect) for external questions.

From [the specification document](https://docs.google.com/document/d/1JbKiwgBCuI2t3Cd8n5OnyrNce3woq20ncZbN3oGcPhU/edit):
If the field has a binary type, Collect will use the ClipData contents. If there’s more than one item in the ClipData, Collect will fail silently and clear the field.

Implementation notes:
* Currently each question type has its own widget class. Is it better to continue this pattern or could we pull everything to do with external applications out of the widget layer?
* Important to manage redundancy with the field-list implementation
* https://github.com/getodk/collect/pull/4208#issuecomment-724199226 has some cases to test

grzesiek2010:
I'm thinking about the UI and this is something we should discuss.
In case of non binary questions like text/integer/decimal if there is an appearance that indicates that an external app should be used we display a button and if trying to open such an app it turns out it's not available we display a text box like in case of normal text/integer/decimal question.
So now what about binary questions? There are options like (let's take the image widget as a sample):
- we can display one launch button instead of Take picture and Choose image buttons, then if the external app is not available display those buttons below. The behavior would be similar to what I described above in case of text widgets
- we can display all three buttons at once without checking if the external app is available
- we can override Take picture button's behavior and try to start an external app if it's available otherwise start the camera app
- we can also display just that one (Launch) button and not to allow taking pictures or choosing it from the phone

I would like to hear from you @lognaturel @seadowg

lognaturel:
The use case that led to this feature involves launching an app that goes through some process and then generates and passes back a file. My sense is that there's truly no reasonable backup if the app is missing. In the case of an integer value, I can imagine something like launching an app that takes some measurement but being able to fall back to a physical sensor if the app is not available. In the binary file case, it seems like if the desired app doesn't exist it shouldn't be possible to proceed. I like your last option the best.

seadowg:
> The use case that led to this feature involves launching an app that goes through some process and then generates and passes back a file.

Yeah based on this I'd agree with @lognaturel that the last option feels best. Maybe we could add fallbacks in the future.

PR #4351, opened 2021-01-15
Ex arbitrary file widget

grzesiek2010:
I tested the fix manually and added (and improved) automated tests.

It's a continuation of https://github.com/getodk/collect/pull/4188

We need to test widgets:
- **ExArbitraryFileWidget** - it's a completly new widget
- **ArbitraryFileWidget** - I had to refactor this widget to implement ExArbitraryFileWidget
- **ExStringWidgets** - I also touched these widgets to factor out the shared code so would be good to verify there is everything fine with them
and that should be enough when it comes to testing.

[externalForm.xlsx](https://github.com/getodk/collect/files/5820990/externalForm.xlsx)

I'll file an issue for all the external binary widgets one we implement all of them.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -145,6 +145,7 @@ import org.odk.collect.android.utilities.ActivityAvailability;
 import org.odk.collect.android.utilities.ApplicationConstants;
 import org.odk.collect.android.utilities.DestroyableLifecyleOwner;
 import org.odk.collect.android.utilities.DialogUtils;
+import org.odk.collect.android.utilities.ExternalAppIntentProvider;
 import org.odk.collect.android.utilities.FormNameUtils;
 import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.utilities.PlayServicesChecker;
@@ -350,6 +351,9 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
     @Inject
     ActivityAvailability activityAvailability;
+
+    @Inject
+    ExternalAppIntentProvider externalAppIntentProvider;
 
     private final LocationProvidersReceiver locationProvidersReceiver = new LocationProvidersReceiver();
 
@@ -819,13 +823,16 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             case RequestCodes.EX_STRING_CAPTURE:
             case RequestCodes.EX_INT_CAPTURE:
             case RequestCodes.EX_DECIMAL_CAPTURE:
-                Object externalValue = getValueFromExternalApp(intent);
+                Object externalValue = externalAppIntentProvider.getValueFromIntent(intent);
                 if (getCurrentViewIfODKView() != null) {
                     setWidgetData(externalValue);
                 }
                 break;
             case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
-                loadFile((Uri) getValueFromExternalApp(intent));
+                externalValue = externalAppIntentProvider.getValueFromIntent(intent);
+                if (externalValue instanceof Uri) {
+                    loadFile((Uri) externalValue);
+                }
                 break;
             case RequestCodes.EX_GROUP_CAPTURE:
                 try {
@@ -891,13 +898,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
             }
         });
-    }
-
-    @Nullable
-    private Object getValueFromExternalApp(Intent intent) {
-        return intent.getExtras().containsKey("value")
-                ? intent.getExtras().get("value")
-                : null;
     }
 
     public QuestionWidget getWidgetWaitingForBinaryData() {
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -829,9 +829,10 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 }
                 break;
             case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
-                externalValue = externalAppIntentProvider.getValueFromIntent(intent);
-                if (externalValue instanceof Uri) {
-                    loadFile((Uri) externalValue);
+                if (intent.getClipData().getItemCount() > 0 && intent.getClipData().getItemAt(0) != null) {
+                    loadFile(intent.getClipData().getItemAt(0).getUri());
+                } else {
+                    setWidgetData(null);
                 }
                 break;
             case RequestCodes.EX_GROUP_CAPTURE:
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -819,15 +819,13 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             case RequestCodes.EX_STRING_CAPTURE:
             case RequestCodes.EX_INT_CAPTURE:
             case RequestCodes.EX_DECIMAL_CAPTURE:
-            case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
-                String key = "value";
-                boolean exists = intent.getExtras().containsKey(key);
-                if (exists) {
-                    Object externalValue = intent.getExtras().get(key);
-                    if (getCurrentViewIfODKView() != null) {
-                        setWidgetData(externalValue);
-                    }
+                Object externalValue = getValueFromExternalApp(intent);
+                if (getCurrentViewIfODKView() != null) {
+                    setWidgetData(externalValue);
                 }
+                break;
+            case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
+                loadFile((Uri) getValueFromExternalApp(intent));
                 break;
             case RequestCodes.EX_GROUP_CAPTURE:
                 try {
@@ -893,6 +891,13 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
             }
         });
+    }
+
+    @Nullable
+    private Object getValueFromExternalApp(Intent intent) {
+        return intent.getExtras().containsKey("value")
+                ? intent.getExtras().get("value")
+                : null;
     }
 
     public QuestionWidget getWidgetWaitingForBinaryData() {
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -829,7 +829,9 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                 }
                 break;
             case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
-                if (intent.getClipData().getItemCount() > 0 && intent.getClipData().getItemAt(0) != null) {
+                if (intent.getClipData() != null
+                        && intent.getClipData().getItemCount() > 0
+                        && intent.getClipData().getItemAt(0) != null) {
                     loadFile(intent.getClipData().getItemAt(0).getUri());
                 } else {
                     setWidgetData(null);
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -819,6 +819,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             case RequestCodes.EX_STRING_CAPTURE:
             case RequestCodes.EX_INT_CAPTURE:
             case RequestCodes.EX_DECIMAL_CAPTURE:
+            case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
                 String key = "value";
                 boolean exists = intent.getExtras().containsKey(key);
                 if (exists) {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/BaseArbitraryFileWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/BaseArbitraryFileWidget.java
@@ -1,0 +1,71 @@
+package org.odk.collect.android.widgets;
+
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.core.model.data.StringData;
+import org.odk.collect.android.formentry.questions.QuestionDetails;
+import org.odk.collect.android.utilities.MediaUtils;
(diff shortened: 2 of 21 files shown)


grzesiek2010:
@lognaturel
I created this draft pr to give you a chance to take a look before I implement ex widgets for rest binary widgets so that you can let me know ff there is something wrong or something you particularly don't like.

I decided to create a separate pr for every widget because reviewing and testing will be easier especially taking into account that I will need to refactor some of the existed code first in some cases.

In https://github.com/getodk/collect/issues/4214 you wanted for the last option. That mean having one button which will launch an external app. Such a solution will require having different layouts so I think we should implement those ex widgets as separate widgets like I did in this pr.

grzesiek2010:
There is one thing related to this pr which is particularly confusing to me. Maybe @lognaturel or @seadowg one of us is more experienced in that area so let me explain:
In order to get files from uri we use this long and complex code [MediaUtils#getPath()](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.java#L109)
it contains separate fixes for different sources. Is it something that we still need or maybe it's an obsolete solution?
I'm asking because in case of launch intent group we use this code which also works well: [FileUtils#saveAnswerFileFromUri()](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java#L106)
That second solution is as I remember based on https://developer.android.com/training/secure-file-sharing/request-file
Does that mean we no longer need that first long method with separate conditions for different sources?

lognaturel:
I've only had a chance to take a quick look so far. I'm out most of Monday for Martin Luther King Day and wanted to at least give you quick impressions. Doing separate widget classes and separate PRs sounds good to me. It's not clear to me why you've used a call back structure with `OnIntentProvided`. It seems like it would be easier to understand if you just had a method that returns an `Intent` or throws an exception.

> MediaUtils#getPath()

I agree that it's really not clear why there's a need to ever convert the given URI to a path. It seems it's just used to get the extension and the MIME type. I believe the MIME type should always be discoverable from the `ContentResolver`. The extension could be figured out from a copy of the original file.

I'm confident that for the external widgets, getting a file descriptor and copying the file through it would be sufficient. It's somewhat less clear to me what happens with an explicit file picker that lets users pick from galleries and remote sources. I do think they're all just content URIs that could be treated the same way, though. Maybe you can try the much simpler approach for this "arbitrary file" hierarchy and we can ask QA to do a really careful check on the manual (non-external) question type.

grzesiek2010:
> It's not clear to me why you've used a call back structure with OnIntentProvided. It seems like it would be easier to understand if you just had a method that returns an Intent or throws an exception.

yes you are right simple method which throws an exception would suffice. The reason I did it in that way was that I was studying something (last week when I created this pr) and the topic was functional programming and my brain was like come on do that :) 

> I agree that it's really not clear why there's a need to ever convert the given URI to a path. It seems it's just used to get the extension and the MIME type. I believe the MIME type should always be discoverable from the ContentResolver. The extension could be figured out from a copy of the original file.

I'm confident that for the external widgets, getting a file descriptor and copying the file through it would be sufficient. It's somewhat less clear to me what happens with an explicit file picker that lets users pick from galleries and remote sources. I do think they're all just content URIs that could be treated the same way, though. Maybe you can try the much simpler approach for this "arbitrary file" hierarchy and we can ask QA to do a really careful check on the manual (non-external) question type.

It starts falling into place, please take a look at https://github.com/getodk/collect/pull/4297#issuecomment-762807706
I think we should implement and test it as a part of that pr and the we can use it everywhere even here.

collect_app/src/main/java/org/odk/collect/android/widgets/ExStringWidget.java
```diff
@@ -185,57 +178,17 @@ public void cancelLongPress() {
...
-                // ACTION_SENDTO used for sending text messages or emails doesn't require any results
-                if (ACTION_SENDTO.equals(i.getAction())) {
-                    getContext().startActivity(i);
-                } else {
-                    fireActivity(i);
-                }
-            } catch (ExternalParamsException | ActivityNotFoundException e) {
-                Timber.d(e);
-                onException(e.getMessage());
+        waitingForDataRegistry.waitForData(getFormEntryPrompt().getIndex());
```
lognaturel, line 181: I'm surprised to see this out of the `try`. It means that if the app launch fails, the registry will still be keeping track of the question as waiting for data. I think it doesn't actually matter in practice because nothing will read that it's waiting for data, is that right? 

Semantically it would kind of make more sense to me to have the call after the intent population as it was before but I guess that in practice it's really the starting of the activity that's more likely to fail so it's still not perfect? I think I've talked myself into thinking this is ok but would still be good for you to see whether my reasoning is sound.

collect_app/src/test/java/org/odk/collect/android/widgets/ArbitraryFileWidgetTest.java
```diff
@@ -51,36 +44,78 @@ public ArbitraryFileWidget createWidget() {
...
 
-        assertThat(getSpyWidget().chooseFileButton.getVisibility(), is(View.GONE));
+        ArbitraryFileWidget widget = getWidget();
+        assertThat(widget.binding.arbitraryFileAnswerText.getVisibility(), is(View.VISIBLE));
+        assertThat(widget.binding.arbitraryFileAnswerText.getText(), is(getInitialAnswer().getDisplayText()));
     }
 
     @Test
-    public void whenReadOnlyOverrideOptionIsUsed_shouldAllClickableElementsBeDisabled() {
+    public void whenClickingOnButton_shouldFilePickerByCalled() {
```
lognaturel: By -> Be

collect_app/src/test/java/org/odk/collect/android/widgets/ArbitraryFileWidgetTest.java
```diff
@@ -51,36 +44,78 @@ public ArbitraryFileWidget createWidget() {
...
+        getWidget().binding.arbitraryFileButton.performClick();
+        verify(mediaUtils).pickFile(activity, "*/*", ApplicationConstants.RequestCodes.ARBITRARY_FILE_CHOOSER);
+    }
+
+    @Test
+    public void whenClickingOnAnswer_shouldFileViewerByCalled() {
```
lognaturel, line 83: By -> Be

collect_app/src/test/java/org/odk/collect/android/widgets/ExArbitraryFileWidgetTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public void whenThereIsAnswer_shouldAnswerTextBeDisplayed() {
+        when(formEntryPrompt.getAnswerText()).thenReturn(getInitialAnswer().getDisplayText());
+
+        ExArbitraryFileWidget widget = getWidget();
+        assertThat(widget.binding.exArbitraryFileAnswerText.getVisibility(), is(View.VISIBLE));
+        assertThat(widget.binding.exArbitraryFileAnswerText.getText(), is(getInitialAnswer().getDisplayText()));
+    }
+
+    @Test
+    public void whenClickingOnButton_shouldFilePickerByCalled() throws ExternalParamsException, XPathSyntaxException {
```
lognaturel: `shouldFilePickerByCalled` -> `externalAppShouldBeLaunchedByIntent` or something like that

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -800,13 +804,15 @@ protected void onActivityResult(int requestCode, int resultCode, final Intent in
...
-                    if (getCurrentViewIfODKView() != null) {
-                        setWidgetData(externalValue);
-                    }
+                Object externalValue = externalAppIntentProvider.getValueFromIntent(intent);
+                if (getCurrentViewIfODKView() != null) {
+                    setWidgetData(externalValue);
+                }
+                break;
+            case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
+                externalValue = externalAppIntentProvider.getValueFromIntent(intent);
```
lognaturel: We'd said that since a `ClipData` needs to be provided anyway, we'll just take whatever is in the `ClipData` and use that, forgetting about any extras. From [the spec doc](https://docs.google.com/document/d/1JbKiwgBCuI2t3Cd8n5OnyrNce3woq20ncZbN3oGcPhU/edit): "Collect will read the ClipData contents and do the same as above, except that it will ignore any extras provided as well as the ClipData filename. Additionally, if there’s more than one item in the ClipData, Collect will fail silently and clear the field."

I'm not too picky about what to do in the case of multiple `ClipData` items as long as it's explicitly documented. If clearing the field is not practical, we could also just always take the first one. If no `ClipData` is provided, the ideal would be to clear the field if there's already an answer. Silently failing would be ok (though seems straightforward to clear the field?).

collect_app/src/main/java/org/odk/collect/android/widgets/BaseArbitraryFileWidget.java
```diff
@@ -0,0 +1,76 @@
...
+            if (answerFile.exists()) {
+                questionMediaManager.replaceAnswerFile(getFormEntryPrompt().getIndex().toString(), answerFile.getAbsolutePath());
+                showAnswerText();
+                widgetValueChanged();
+            } else {
+                answerFile = null;
+                Timber.e("Inserting Arbitrary file FAILED");
+            }
+        } else {
+            Timber.e("FileWidget's setBinaryData must receive a File or Uri object.");
```
lognaturel: Currently `Uri` objects aren't supported so this is misleading. In order for it to be actionable, we should log the actual file type received.

lognaturel (changes requested):
The structure looks good to me overall! A few things inline. Now that you've pulled out `ExternalAppIntentProvider`, it'd be great to get a little bit of testing on it.

collect_app/src/main/java/org/odk/collect/android/widgets/ExStringWidget.java
grzesiek2010: > I think it doesn't actually matter in practice because nothing will read that it's waiting for data, is that right?

Yes you are right we don't need to care about registering such questions because even if we register a question and then something wrong happens it won't cause any problems.

collect_app/src/test/java/org/odk/collect/android/widgets/ArbitraryFileWidgetTest.java
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/widgets/ArbitraryFileWidgetTest.java
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/widgets/ExArbitraryFileWidgetTest.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseArbitraryFileWidget.java
grzesiek2010: Good catch! I removed that `or Uri object`. However i'm not sure what you mean by:
> we should log the actual file type received

This else block is a result of `!(object instanceof File)` so what file types you do you want to log here? Or maybe you want to add such a log above where it's really a file?

lognaturel: I meant we could log `object.getClass()` so at least then we'd be able to track down where a bad call may have come from. In practice it feels like we can control the call site well enough to be confident everything coming in will be a `File` so hopefully this branch would never be hit.

grzesiek2010: Ok I see. I changed it and now I log object.getClass() if the object is not null because if it is null it most likely comes from our app (it's a way of removing the existing answer).

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Ok I changed the code and now it is using ClipData. I also added removing an existing answer if no ClipData is provided as you said.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -800,13 +804,18 @@ protected void onActivityResult(int requestCode, int resultCode, final Intent in
...
-                    if (getCurrentViewIfODKView() != null) {
-                        setWidgetData(externalValue);
-                    }
+                Object externalValue = externalAppIntentProvider.getValueFromIntent(intent);
+                if (getCurrentViewIfODKView() != null) {
+                    setWidgetData(externalValue);
+                }
+                break;
+            case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
+                if (intent.getClipData() != null
```
lognaturel, line 832: You'll probably want to pull this out somewhere once you have more external file widgets but this looks fine for now!

lognaturel (approved):

grzesiek2010:
> Now that you've pulled out ExternalAppIntentProvider, it'd be great to get a little bit of testing on it.

Agree but maybe as a part of one of other prs not to block this one.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseArbitraryFileWidget.java
lognaturel: > it's a way of removing the existing answer

It doesn't look to me like nulls would remove the existing answer currently.

grzesiek2010: Why? Please look at https://github.com/getodk/collect/pull/4351/files#diff-d4cabe3ff3c1faf046ef950df40de48cbf816cb2ab01e63578d5e8e897bef3e3R50

if there is an answer it is deleted and then below we check if the new answer is file or not and we skip rest of the code.

lognaturel: So if `null` is passed in, `answerFile` might still be non-null and in that case it would clear the answer out. Got it. I expected an explicit check on `object == null` but now I see why it's not needed.

grzesiek2010: There is even a test for that case.

kkrawczyk123:
**Tested with success!**

Tested cases:
- generated answer 
- RTL language
- used forward/backward navigation buttons
- remove response
- save/finalize form
- toast is visible when external app is not available
- launch counter app
- light/dark mode
- edit form
- don't keep activities
- rotation
- minimize 
- send submission to Central and Aggregate
- different font sizes
- regression check on ex string and numerical widgets

mmarciniak90:
Tested also on Android versions: 7.0, 8.1, 10.0

@getodk-bot unlabel "need testing"

mmarciniak90:

PR #4352, opened 2021-01-15
Use FragmentScenario to test DialogFragments

seadowg:
This updates most of the `DialogFragment` tests to use `FragmentScenario` now that we have a helper that handles setting up the themes required to make this work.

A few of the classes under test would need a bunch of refactoring to make this switch and so I decided to skip those to keep this timeboxed.

Ran tests.

`FragmentScenario` appears to be the new standard so it makes sense to move to it. It also means we are using Robolectric less (directly) which I think is a good thing given it's (always) dubious future.

Just test changes really. No need to go through QA.


collect_app/src/main/java/org/odk/collect/android/formentry/saving/SaveFormProgressDialogFragment.java
```diff
@@ -36,7 +36,6 @@ public void onAttach(@NonNull Context context) {
         ViewModelProvider.Factory factory = formSaveViewModelFactoryFactory.create(requireActivity(), null);
         viewModel = new ViewModelProvider(requireActivity(), factory).get(FormSaveViewModel.class);
 
-        setCancelable(false);
```
lognaturel: The default is true so this might change behavior. I see from `setCancellableFalse_makesTheDialogNotCancellable` that an argument can be set on launch. Is it always? And if it was ignored before, should we perhaps just make it always cancelable and forget about the arg?

seadowg: There is a test for this in `SaveProgressDialogFragmentTest`. I think you're right that the default is `true` for `DialogFragment` but in `ProgressDialogFragment` it looks like we make it `false` [here](https://github.com/getodk/collect/blob/5d4a1dae5520d4270a2c99c87a7420602559b2d8/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ProgressDialogFragment.java#L110).

lognaturel: Is it the case that `setupView` is always guaranteed to be called after `onAttach`? I thinkso? If so then you're right that the call on `onAttach` was always overridden.

seadowg: Yeah I think so. I think the test proves this is true for `SaveProgressDialogFragment`.

lognaturel: I don't think so? Wouldn't it only prove that if it set `cancellable` to `true`? Because previously `onAttach` already set `cancellable` to `false` so a param of `false` would have no effect.

lognaturel (approved):
Seems good to me. @grzesiek2010 heads up that fragments should use this pattern moving forward.

lognaturel (changes requested):
Can I not dismiss a review? I had a moment where I thought the cancelable thing was ok but now I want to at least finish the conversation. I think setting cancelable to false in onAttach might actually be the behavior we want. That is, a saveformprogressdialog is not cancelable by default.

lognaturel:
Can we make this change test-only? Then I think in a follow-up PR it would be great to reconsider `CANCELABLE`. ~rip out all this stuff about a CANCELABLE argument because I don't think it makes any sense. The CANCELABLE constant is set to `true`, so that's the bundle key name. There's never any attempt to set a value for it except for in the `BlankFormListFragment` instantiation. But that could just go in its `onAttach`, same as with `SaveProgressDialogFragment`.~ It does make sense in the case of a raw `ProgressDialogFragment` that needs to be modified to be cancelable like the delete dialog instantiated from `BlankFormListFragment`. For a subclass that always needs to be non-cancelable, using `onAttach` makes more sense.

seadowg:
@lognaturel Totally agree that the `CANCELLABLE` parameter is confusing stuff here -  I think you're right that that behaviour could just be set in `onAttach`. For the moment I'll make this test only as you suggest as there's a few too many things in flight for me to dive back in here.

lognaturel (approved):
I see, I think your change did end up working because `getBoolean` defaults to false if the key isn't present. And I do see there is a test to support it. Sorry about that. I do prefer the more explicit call and if we were going to rely on that fallback behavior for subclasses we should make it consistent (e.g. `FormLoadingDialogFragment` has the same `setCancelable(false)` call).

lognaturel:
> I do prefer the more explicit call

Ugh, except it does get overridden so that's misleading. Sigh. Anyway, moving on.

Issue #4344, opened 2021-01-12
Background audio is recorded when form requests it


seadowg:
~~Blocked on JavaRosa changes.~~

We can remove the background audio experimental setting as part of this.

**Acceptance**

- [x] Given my form has a `odk:recordaudio` action
When the form loads
Then I see the background recording overlay
When I finish filling it in
And I submit it
Then the submitted instance includes the recording in the action (question)

- [x] Given my form has multiple `odk:recordaudio` actions
When the form loads
Then I see the background recording overlay
When I finish filling it in
And I submit it
Then the submitted instance includes the recording in all the different actions (questions)

seadowg:
Discussion here: https://forum.getodk.org/t/form-spec-proposal-add-background-audio-recording/31889

Issue #4342, opened 2021-01-12
Can see background audio is being recorded


seadowg:
**Acceptance**

- [x] Given I've enabled "Background audio" in Experimental settings
When I start filling out a form
Then I see an overlay informing that there is a recording in progress
And I see how long the recording currently is
And there are no controls to stop/pause/start the recording

PR #4354, opened 2021-01-21
Add background recording for forms

### LINKED ISSUE #4341: Can record background audio while filling out form
**Acceptance**
When I fill out a form
Then there is a new `.amr` file containing audio of me filling out the form in the `recordings` directory (in the app's external storage)

seadowg:
Closes #4341 
~~Blocked by #4369~~

Implementing this forces us to let the user navigate the form while recording. I've ended up making this possible for both background recording and recording that happens as part of a question. Currently, the behaviour for audio questions is:

* You can swipe between form pages (or use next/back) while recording
* You can navigate the form using the hierarchy while recording
* You can add a new repeat using the ➕ button while recording
* Clicking "Save and exit" on the form end page will prompt you to stop recording
* Trying to navigate to settings will prompt you to stop recording
* Hitting the back button will prompt you to stop recording

Background recording only stops when save or ignore change to exit. Right now you can stop it with the stop button but the next issue will take care of that.

All of these behaviours are pretty easy to change around so if they don't make sense to anyone lets discuss. It could be that we find weird edge cases UX wise that force us to add the blocks on navigating back in but for the moment lets see how we go.

Comments inline.

Should just be the changes described above. I think it'd be good to check recording audio in general (and the new behaviours) to see if we find anything we might need to come back and fix.


audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderViewModel.kt
```diff
@@ -13,7 +14,7 @@ abstract class AudioRecorderViewModel : ViewModel() {
     abstract fun isRecording(): Boolean
     abstract fun getCurrentSession(): LiveData<RecordingSession?>
 
-    abstract fun start(sessionId: String, output: Output)
+    abstract fun start(sessionId: Serializable, output: Output)
```
seadowg, line 17: We need to use `Seriablizable` instead of `Object`/`Any` here as the identifier needs to be able to be passed as an argument to an Android `Service`.

audiorecorder/src/test/java/org/odk/collect/audiorecorder/recording/AudioRecorderViewModelTest.kt
```diff
@@ -153,4 +154,16 @@ abstract class AudioRecorderViewModelTest {
         runBackground()
         assertThat(session.value?.failedToStart, equalTo(null))
     }
+
+    @Test
+    fun supportsSerializableSessionIds() {
```
seadowg, line 159: Annoyingly we need a test to drive out some of the changes here as `String` is `Serializable`

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/BackgroundAudioRecordingTest.java
```diff
@@ -0,0 +1,86 @@
...
+        FormEndPage formEndPage = formEntryPage
+                .inputText("123")
+                .swipeToEndScreen();
+        assertThat(stubAudioRecorderViewModel.isRecording(), is(true));
+
+        formEndPage.clickSaveAndExit();
+        assertThat(stubAudioRecorderViewModel.isRecording(), is(false));
+
+        assertThat(stubAudioRecorderViewModel.getLastRecording(), notNullValue());
+        assertThat(stubAudioRecorderViewModel.getLastRecording().exists(), is(true));
```
seadowg, line 84: Right now we're just checking the recording exists on the disk. Once the form spec has been agreed on we can change this.

strings/src/main/res/values/strings.xml
```diff
@@ -988,4 +988,6 @@
     <string name="not_a_directory">%s exists, but is not a directory</string>
 
     <string name="select_odk_shortcut">Select ODK Shortcut</string>
+
+    <string name="background_audio_recording">Background audio recording</string>
```
lognaturel: We're not actually going to use this string in the final UI. Could you put it in untranslated so it doesn't go to translators, please?

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RecordingStatusProvider.java
```diff
@@ -0,0 +1,14 @@
...
+
+import android.util.Pair;
+
+import org.javarosa.form.api.FormEntryPrompt;
+
+import java.util.function.Consumer;
+
+public interface RecordingStatusProvider {
+
+    void onIsRecordingBlocked(Consumer<Boolean> isRecordingListener);
```
lognaturel: Is this idiomatic naming? That is, these methods don't feel like callbacks/handlers, they're setting up observers. I know this code was already there previously but it really jumped out at me this time, maybe because of the `Provider` naming. This doesn't feel like a provider. It feels almost like `RecordingStatusObserver` with `observeRecordingBlocked` and `observeRecordingInProgress`. Not perfect because `Observer` is a pattern but I'm bringing this up because my first reaction was that I didn't like the structure. I now think it's ok, I just can't reconcile it with the naming. Perhaps it's just an idiom I'm not familiar with.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -144,7 +145,8 @@
      * @param groups          the group hierarchy that this question or field list is in
      * @param advancingPage   whether this view is being created after a forward swipe through the
      */
-    public ODKView(ComponentActivity context, final FormEntryPrompt[] questionPrompts, FormEntryCaption[] groups, boolean advancingPage, QuestionMediaManager questionMediaManager, WaitingForDataRegistry waitingForDataRegistry, AudioPlayer audioPlayer, AudioRecorderViewModel audioRecorderViewModel, FormEntryViewModel formEntryViewModel) {
+    @SuppressWarnings("PMD.ExcessiveParameterList")
```
lognaturel, line 151: Since we seem to be going all in with passing in dependencies via constructors for classes that aren't Android classes, should we suppress this warning in our top-level config?

lognaturel (approved):
Overall looks good! Neat that ongoing recording didn't end up being a huge change.

@seadowg and I talked about this and agreed it'd be good to get preliminary QA feedback. The major things to look for are:
- does background recording work as expected?
- are there any unexpected cases around recording while navigating the form. If so, we would likely file those as issues and when the background audio feature is complete, decide whether to fix those cases or go back to blocking leaving a screen while recording. No matter what, we need to have some user conversations around whether this new behavior is desirable or not.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RecordingStatusProvider.java
lognaturel: Maybe `RecordingStatusHandler` and `onBlockedStatusChange` and `onRecordingStatusChange`?

I'm not sure this is even really worth discussing, I'm probably just not writing enough code and also getting old. 😄

strings/src/main/res/values/strings.xml
seadowg: Ah great idea. Will do

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/RecordingStatusProvider.java
seadowg: No I also felt weird about the naming once I broke it out but thought that might just be me.  I like your last suggestion, so I'll go with that for the moment, and we can see how it settles.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
seadowg: I'm in two minds about this. When I've added them in I've generally ended up being able to get rid of it somehow, and it's felt like a good piece of pressure. That said you're right that it's not really an actual problem to have a large constructor when the alternative is often grabbing things statically. Maybe let's leave this for now and if it comes up again we revise?

lognaturel: Sounds good!

lognaturel:
I think you’re holding onto some commits locally!

seadowg:
> I think you’re holding onto some commits locally!



I've just pushed the changes!

lognaturel (approved):

mmarciniak90:
@seadowg I've noticed that user is not asked for permissions.

seadowg:
@mmarciniak90 ah nice find. I'll create an issue for that as I don't think we need to block this (I've put a placeholder in the "Current release" project).

kkrawczyk123:
@seadowg I have noticed issues:
- The recording in question is continued after removing the group where the question is. I have compared this behavior to master and dialog preventing removing group used to display in such scenario. The exact steps:
1. Enable background recording.
2. Open form with repeated groups
3. Stop default recording 
4. Add groups
5. Start recording in question in the last added group
6. Remove group
gif:
ezgif com-video-to-gif (50)grupy

I was also able to reproduce when added only 1 group and remove that group.


- I can see inconsistency in behavior for default recording and recording in question when background recording is enabled:
1. For default recording when use Save and Exit I am able to leave the form and recording is stopped:
ezgif com-video-to-gif (50)_backgorund

2. For recording in question when I do the same thing I can't save and leave the form - Shouldn't the behavior be the same??
ezgif com-video-to-gif (50)singleq

mmarciniak90:
@seadowg here are my finds:

Is it expected that recorder audio is not permanently connected with the form?

1. User starts form -> Recording is started
2. User goes to General Settings and changes the language
3. User is moved to the main view -> there is no option to stop recording
4. User selects DIFFERENT form -> Recording is continued
5. User finishes and saves form -> Audio is saved as a part of the second form
6. User starts the first form from cache - new recording is started


 
---
The recording directory is not cleared. Even reset settings does not clear it.

---

App crashes on Android 8.1 and 7.1.
I'm not able to record video on a real device but I was able on an emulator with Android 8.1

1. User starts form -> Recording is started
2. User pauses recording
3. User goes to the end of the form
4. User clicks on `Save Form and Exit`



seadowg:
@mmarciniak90 thanks for all the investigating! Here are some thoughts on the findings:

> Is it expected that recorder audio is not permanently connected with the form?

For now, you'll just get a new recording every time you start form entry. Once the form spec has been agreed on only forms that ask for background recording will carry it out. The Settings > Change language is a really annoying case. I'd forgotten that some settings changes will boot you out of form entry. I'll have a look at how we can avoid that. We should solve this before moving on as it could be really confusing.

> The recording directory is not cleared. Even reset settings does not clear it.

Yeah, that's fine. Again once we have an agreed form spec this will be less weird. The recording will probably end up in the instance directory.

> App crashes on Android 8.1 and 7.1.

Hmmm. We'll be removing the pause button for background recordings in a later issue (#4342) but I'll look into this as it still shouldn't crash.

seadowg:
@kkrawczyk123 ooops missed your comment as well:

> The recording in question is continued after removing the group where the question is.

I think this is fine as long as we're not seeing crashes. Interested if you disagree @lognaturel. We could smooth it out down the line by not letting a user remove a group while recording etc. There are other cases like questions being hidden while recording because of form logic etc but I'm still feeling like nothing is bad enough to prevent us from allowing the user to navigate while recording.

> For recording in question when I do the same thing I can't save and leave the form - Shouldn't the behavior be the same??

I think it's what we want right now. For the background case we want the enumerator to be able to go through the form without any interruptions but when we're recording for a question we want to the enumerator to be in control - we make them stop recording before leaving the form.

mmarciniak90:
> The recording will probably end up in the instance directory.

Yes, it's an optimistic case. I asked about that because I'm worried about case where user stops recording and ignores changes. For now, record files take space in the `recordings` directory.

seadowg:
@mmarciniak90 sorry I don't think I was clear there. You're right that as it is this feature results in recordings piling up on the device. The recording will be being automatically deleted from the `recordings` directory and copied somewhere else in the final version of the feature (see [this placeholder issue](https://github.com/getodk/collect/issues/4344)) it's just that there's an ongoing discussion around exactly how it will be included in the form ([here](https://forum.getodk.org/t/form-spec-proposal-add-background-audio-recording/31889)). Once that discussion is finished we'll know what we're doing with the recording. 

It will of course be good to check we're not leaving stuff around but didn't want you to worry that we were going to be polluting the user's device with recordings.

seadowg:
@kkrawczyk123 @kkrawczyk123 this is good to test again. For the moment I've blocked settings while recording and the crash has been fixed.

kkrawczyk123:
@seadowg I have a question. Why exactly do we need this background recording starting by default when we open the form as we can not send it to any of the servers? Amr file is visible in the instance folder on a device but when send it to Central/Aggregate it's not recognized as a media file as it's not connected to any widget.

seadowg:
@kkrawczyk123 sorry I think I could have explained this better earlier! We're still discussing how exactly the form designer will add background recording into the form and that will determine how the recording gets included in the submission. In the meantime @lognaturel and I decided it was ok to go ahead and start implementing the actual background recording feature and just use the experimental preferences (which isn't available in release builds) to turn it on and off. 

It felt like adding the ability to navigate the form while recording could add some confusing situations and maybe some crashes, both of which you've managed to find. That's why I was interested in having this tested early before it's actually useful/finished.

Hopefully that makes sense?

In the future, I can try to have a more up front conversation around what exactly the PR adds and what it doesn't. For reference here are the discussions on [the feature](https://forum.getodk.org/t/create-automatic-background-recording-of-interview-in-collect/18601/20) and the [form spec](https://forum.getodk.org/t/form-spec-proposal-add-background-audio-recording/31889).

lognaturel:
> we can not send it to any of the servers?

@kkrawczyk123 This is a temporary state just to get the basic behavior working. #4344 will wire everything together once we have the form specification finalized.

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- background recording in form and in audio question
- swipe / navigation buttons
- add repeat group
- adding group via + icon 
- hierarchy view
- adding / deleting group from hierarchy view
- audio widget is blocked
- pause/stop recording
- finalize form -> recording is stopped
- open General settings from form view is not possible
- use external app widget
- use image, video, geo, file, draw, barcode widgets when audio is being recorded
- minimize app
- screen rotation
- RTL language
- readonly questions
- fieldlist question
- form with background location recording
- encrypted form
- save form
- form with voice-only, normal and low
- use audio widget when background recording disabled
- use back button
- moving backwards disabled
- required question: validation upon swipe and finalize

mmarciniak90:
Tested also on Android 7.0, 8.1, 10.0

PR #4356, opened 2021-01-22
Change download analytics to only log when forms with same formid/version but different hashes are requested

lognaturel:
The question we're trying to answer with this PR is whether we can make another attempt towards warning about or preventing downloads with the same formid/version and different hashes.

The `DownloadSameFormidVersion` event let us see that quite a lot of users download forms with the same formid and version. We suspect that these are typically identical forms and that people just select all in Get Blank Forms. This PR only logs analytics events when the form being downloaded has a different hash than the latest one with matching formid/version AND the download URL doesn't include `/draft.xml`. The latter condition is a rough way to avoid Central drafts since we know that Central draft workflows generally involve changing the form definition without changing the form version.

Used the debugger to confirm that the analytics event is reached when the form hashes are different and not reached when the form hashes are the same. In doing this, I noticed that when we get a form by id and version, we really should always get the latest one downloaded. I verified that change by adding an automated test. I also verified that if the form hashes are different but a Central draft is being used, the analytics event is not reached.

I think this should probably go through brief QA because of the change to `FormRepository`.

I didn't really consider any alternatives other than not making the testing change to remove id parameters. I put it in here because I got tripped up trying to write a test to make sure the date was used for ordering rather than the database ID. It turns out there's no point in trying to manually set dates or ids when building `Form` objects because the database implementation ignores them. It felt better to remove that confusion.

I think there are two ares of regression:
1. Ordering by date when matching a single form by formid and version. This gets used:
  - when deleting instances, to determine which form gets deleted when the last instance on the device is deleted. I think it would make more sense to delete all forms with matching formid/version but it seems like enough of an edge case that it's not a big deal.
  - when identifying whether instances of a form should be autosent or autodeleted.
2. Matching on hash instead of on formid/version to see if a form needs to be undeleted on form download. This affects cases where forms with the same formid/version and either the same or different contents as forms already on the device AND soft-deleted are downloaded. For a delete to be a soft delete, there need to still be instances matching the formid/version.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -33,7 +33,7 @@ import javax.annotation.Nullable;
 import timber.log.Timber;
 
 import static org.apache.commons.io.FileUtils.deleteDirectory;
-import static org.odk.collect.android.analytics.AnalyticsEvents.DOWNLOAD_SAME_FORMID_VERSION;
+import static org.odk.collect.android.analytics.AnalyticsEvents.DOWNLOAD_SAME_FORMID_VERSION_DIFFERENT_HASH;
 import static org.odk.collect.utilities.PathUtils.getAbsoluteFilePath;
 
 public class ServerFormDownloader implements FormDownloader {
@@ -63,9 +63,11 @@ public class ServerFormDownloader implements FormDownloader {
             if (formOnDevice.isDeleted()) {
                 formsRepository.restore(formOnDevice.getId());
             } else {
-                String formIdentifier = formOnDevice.getDisplayName() + " " + formOnDevice.getId();
-                String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
-                analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION, formIdHash);
+                if (!getMd5HashWithoutPrefix(form.getHash()).equals(formOnDevice.getMD5Hash())) {
+                    String formIdentifier = formOnDevice.getDisplayName() + " " + formOnDevice.getId();
+                    String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
+                    analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION_DIFFERENT_HASH, formIdHash);
+                }
             }
         }
 
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -62,12 +62,13 @@ public class ServerFormDownloader implements FormDownloader {
         if (formOnDevice != null) {
             if (formOnDevice.isDeleted()) {
                 formsRepository.restore(formOnDevice.getId());
-            } else {
-                if (!getMd5HashWithoutPrefix(form.getHash()).equals(formOnDevice.getMD5Hash()) && !form.getDownloadUrl().contains("/draft.xml")) {
-                    String formIdentifier = formOnDevice.getDisplayName() + " " + formOnDevice.getId();
-                    String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
-                    analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION_DIFFERENT_HASH, formIdHash);
-                }
+            }
+        } else {
+            List<Form> allSameFormIdVersion = formsRepository.getAllByFormIdAndVersion(form.getFormId(), form.getFormVersion());
+            if (!allSameFormIdVersion.isEmpty() && !form.getDownloadUrl().contains("/draft.xml")) {
+                String formIdentifier = form.getFormName() + " " + form.getFormId();
+                String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
+                analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION_DIFFERENT_HASH, formIdHash);
             }
         }
 
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -58,7 +58,7 @@ public class ServerFormDownloader implements FormDownloader {
 
     @Override
     public void downloadForm(ServerFormDetails form, @Nullable ProgressReporter progressReporter, @Nullable Supplier<Boolean> isCancelled) throws FormDownloadException, InterruptedException {
-        Form formOnDevice = formsRepository.getLatestByFormIdAndVersion(form.getFormId(), form.getFormVersion());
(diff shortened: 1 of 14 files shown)


collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
```diff
@@ -20,26 +20,43 @@
...
         assertThat(form, notNullValue());
         assertThat(form.getId(), is(1L));
     }
 
+    @Test
+    public void getLatestByFormIdAndVersion_whenMultipleExist_returnsLatest() throws InterruptedException {
+        FormsRepository formsRepository = buildSubject();
+        formsRepository.save(buildForm("1", "1", getFormFilesPath())
+                .build());
+        Thread.sleep(2); // ensure that the in-memory inserts don't all get the same timestamp
```
seadowg: We should be able to use `Clock` to avoid `sleep` here. As far as I can see the created at date is defined in Java rather than using the database (if I'm wrong about this then it's really hard to deal with though).

I'd look at injecting `Clock` into `FormsProvider` and just being a constructor parameter for `InMemFormsRepository` so the time can be set in the tests (with a fake implementation or with mockito).

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -58,14 +58,16 @@ public ServerFormDownloader(FormSource formSource, FormsRepository formsReposito
 
     @Override
     public void downloadForm(ServerFormDetails form, @Nullable ProgressReporter progressReporter, @Nullable Supplier<Boolean> isCancelled) throws FormDownloadException, InterruptedException {
-        Form formOnDevice = formsRepository.getOneByFormIdAndVersion(form.getFormId(), form.getFormVersion());
+        Form formOnDevice = formsRepository.getLatestByFormIdAndVersion(form.getFormId(), form.getFormVersion());
```
seadowg: I think maybe we need a test for this change in `ServerFormDownloaderTest`? This is needed before if we download a new form with the same version and id (but a different hash) we end up with two forms in the repository right?

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
lognaturel: Thanks, that is better. See what you think of my implementation. We can get away with overriding the app dependency module just for the test we need it because injection is delayed for providers. It's a little quirky but feels ok to me. I wanted to make sure the same mock was shared between the two repository implementations.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
lognaturel: Fair enough. In writing the test, I realized we have access to the hash there so there's no need to use a heuristic. I verified that everywhere else we use `getLatestByFormIdAndVersion` there's nothing better we can do because we're dealing with an instance.

I think we should really start tracking the form hash an instance was instantiated from.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseFormsRepositoryTest.java
```diff
@@ -28,6 +30,18 @@ public FormsRepository buildSubject() {
         return new DatabaseFormsRepository();
     }
 
+    @Override
+    public FormsRepository buildSubject(Clock clock) {
+        RobolectricHelpers.overrideAppDependencyModule(new AppDependencyModule() {
```
seadowg, line 35: This could also go in an `@Before` but I do think I prefer it here. It's nice that we end up with `Clock` being passed into the `buildSubject` so it's super clear to the implementers they should build their version wih it.

collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
```diff
@@ -125,12 +145,18 @@ public void save_addsId() {
...
 
         formsRepository.save(form);
 
         String expectedHash = FileUtils.getMd5Hash(new File(form.getFormFilePath()));
         assertThat(formsRepository.get(1L).getMD5Hash(), equalTo(expectedHash));
     }
+
+    public Clock getMockClock() {
+        Clock mockClock = mock(Clock.class);
+        when(mockClock.getCurrentTime()).thenReturn(2L, 3L, 1L);
```
seadowg: Such a minor tweak but I do think it'd be nicer if the mock values were set in the tests themselves. For instance, it's hard when reading `getLatestByFormIdAndVersion_whenMultipleExist_returnsLatest` to see which one is the latest.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -58,14 +58,16 @@ public ServerFormDownloader(FormSource formSource, FormsRepository formsReposito
...
-        Form formOnDevice = formsRepository.getOneByFormIdAndVersion(form.getFormId(), form.getFormVersion());
+        Form formOnDevice = formsRepository.getOneByMd5Hash(getMd5HashWithoutPrefix(form.getHash()));
         if (formOnDevice != null) {
             if (formOnDevice.isDeleted()) {
                 formsRepository.restore(formOnDevice.getId());
             } else {
-                String formIdentifier = formOnDevice.getDisplayName() + " " + formOnDevice.getId();
-                String formIdHash = FileUtils.getMd5Hash(new ByteArrayInputStream(formIdentifier.getBytes()));
-                analytics.logFormEvent(DOWNLOAD_SAME_FORMID_VERSION, formIdHash);
+                if (!getMd5HashWithoutPrefix(form.getHash()).equals(formOnDevice.getMD5Hash()) && !form.getDownloadUrl().contains("/draft.xml")) {
```
seadowg: I could be confused, but I don't think this will ever be `true` right? Here we're looking up `formOnDevice` using `form`'s hash and then in this else you'd be comparing them (so they'd always match and the `if` would be false). Maybe it's worth having a test for this analytics call as it's a bit fiddly.

Sorry if I'm totally wrong about this and just need to wake up a little more!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
lognaturel: Sigh. You're absolutely right. That was me not coming back to what I was actually trying to accomplish after addressing your feedback. It used to make sense before the condition for undeleting changed.

collect_app/src/test/java/org/odk/collect/android/forms/FormsRepositoryTest.java
lognaturel: Agreed. This was an artifact of initially using `@Before`. Being able to put this initialization inline is another nice side effect of only using the mock clock where needed.

seadowg (approved):

kkrawczyk123:
**Tested with success!**

Tested cases:
Central, Aggregate, Google Drive - download and send form
download all forms from server 
Central draft - download and send form
match exactly
match exactly when filled submission
get blank form manually
download the same form for a second time: there is issue visible, I can see that will be fixed in #4359
ordering forms and instances by date
autosend
autodelete
autosend and autodelete defined in form
update form version
form soft delete 
form with submission url defined 
refresh form list button
unassign/sign form and refresh form list

mmarciniak90:
Tested also on Android versions: 7.0, 8.1, 10.0

PR #4357, opened 2021-01-22
Make sure redownloading the same form doesn't break anything

### LINKED ISSUE #4355: Downloading a form with the same hash fails every other time
Collect v1.29.3, works in v1.29.2

#### Problem description
Downloading a form with the same formid and version fails every other time. The user reporting this uses match exactly which is particularly annoying because the forms will disappear and re-appear periodically. https://forum.getodk.org/t/document-has-no-root-element-error-message/32121

1. Download a form such as All Widgets from the default server, verify it opens
2. Go back to Get Blank Form, select the same form, download it
3. Go to Fill Blank Form, see that the form is no longer there
4. On the device in `/sdcard/Android/data/org.odk.collect.android/files/forms`, verify that the form was added a `.bad` extension and is an empty file

#### Expected behavior
It should be possible to request downloading a form with the same formid and version over and over again. Files (form or media) with the same hash should be skipped.

#### Other information 
I think the problem was introduced in #4333 and is that `ServerFormDownloader.downloadXform`'s result will have a file that is the destination form XML file if the form was already on the device. Then `installEverything` will try to copy the file onto itself and apparently the result is to empty the file.

@seadowg This is freshest for you! An easy solution would be to wrap the form def copy in `if (!formFile.getAbsolutePath().equals(fileResult.file.getAbsolutePath()))` in `installEverything`. Would be great to have a test and maybe you can think of a better approach. Perhaps `file` should be `null` in this case.

seadowg:
Closes #4355

Comments inline

It'd be good to check downloading, updating, redownloading etc.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -216,9 +216,16 @@ public class ServerFormDownloader implements FormDownloader {
             UriResult uriResult = null;
 
             try {
-                // Copy form to forms dir
-                File formFile = new File(formsDirPath, fileResult.file.getName());
-                FileUtils.copyFile(fileResult.file, formFile);
+                File formFile;
+
+                if (fileResult.isNew()) {
+                    // Copy form to forms dir
+                    formFile = new File(formsDirPath, fileResult.file.getName());
+                    FileUtils.copyFile(fileResult.file, formFile);
+                } else {
+                    formFile = fileResult.file;
+                }
+
 
                 // Save form in database
                 uriResult = findExistingOrCreateNewUri(formFile, parsedFields, formsDirPath);
--- a/collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
@@ -364,6 +364,50 @@ public class ServerFormDownloaderTest {
     }
 
     @Test
+    public void whenFormAlreadyDownloaded_formRemainsOnDevice() throws Exception {
+        String xform = createXForm("id", "version");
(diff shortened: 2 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -216,9 +216,16 @@ boolean installEverything(String tempMediaPath, FileResult fileResult, Map<Strin
             UriResult uriResult = null;
 
             try {
-                // Copy form to forms dir
-                File formFile = new File(formsDirPath, fileResult.file.getName());
-                FileUtils.copyFile(fileResult.file, formFile);
+                File formFile;
+
+                if (fileResult.isNew()) {
```
seadowg, line 221: For now, only copy the new form file if it's new to prevent anything weird happening when we copy a file into itself. I think we probably want to fix all the weirdness with the form file being new or not but it seems like it would really require ripping out any direct usage of the content provider (and switch to our newer repos) to make anything make sense.

As this is presumably going to be a hot fix and will have to be cherry-picked back onto master (which will involve some nasty merging) I just went with a quick hack solution.

lognaturel (approved):
Looks right to me. Would be good to check 3+ downloads with:
- same exact form
- same formId/version, different contents (eg Central draft)
- Updating version and contents

mmarciniak90:
**Tested with success** 

Verified cases:
- download using match exactly and then manually (Central, Aggregate, Kobo, ONA)
- download twice manually from the same server (Central, Aggregate, GD, Kobo, ONA)
- download twice manually from different servers (Central and then Aggregate)
- download the same and another version (Central)
- download when submission has been filled
- download twice Central draft
- same formId/version, different contents- Central draft -> error during finalizing when 2 forms with different content are available
- same formId/version, different media - Central draft -> only new media are downloaded
- download form with media
- download forms without media
- auto-update to the latest form version (match exactly; only previous forms; Central)

kkrawczyk123:

PR #4358, opened 2021-01-22
Handle faked locations

grzesiek2010:
I added tests and also tested the app manually.

It's a feature discussed on the forum: https://forum.getodk.org/t/restricting-fake-location-application-while-collecting-gps-points-with-collect-android-app/29231/12

When locations are faked the accuracy should be 0. You can use apps like https://play.google.com/store/apps/details?id=com.lexa.fakegps&hl=en&gl=US for testing.

Any form with geo widgets.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/location/client/MapboxLocationClient.java
+++ b/collect_app/src/main/java/org/odk/collect/android/location/client/MapboxLocationClient.java
@@ -1,0 +1,33 @@
+package org.odk.collect.android.location.client;
+
+import android.location.Location;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.gms.location.LocationListener;
+import com.mapbox.android.core.location.LocationEngineCallback;
+import com.mapbox.android.core.location.LocationEngineResult;
+
+public class MapboxLocationClient implements LocationEngineCallback<LocationEngineResult> {
+    @Nullable
+    private final LocationListener locationListener;
+
+    public MapboxLocationClient(@Nullable LocationListener locationListener) {
+        this.locationListener = locationListener;
+    }
+
+    @Override
+    public void onSuccess(LocationEngineResult result) {
+        Location location = result.getLastLocation();
+        if (locationListener != null && location != null) {
+            if (location.isFromMockProvider() || location.getAccuracy() < 0) {
+                location.setAccuracy(0);
+            }
+            locationListener.onLocationChanged(location);
+        }
+    }
+
+    @Override
+    public void onFailure(@NonNull Exception exception) { }
+}
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -348,6 +348,7 @@ dependencies {
     // Testing-only dependencies
     testImplementation "junit:junit:4.13.1"
     testImplementation "org.mockito:mockito-core:3.6.28"
+    testImplementation 'org.mockito:mockito-inline:3.7.7'
 
     testImplementation project(path: ':testshared')
     testImplementation "org.robolectric:robolectric:4.3.1"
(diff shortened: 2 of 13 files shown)


lognaturel (approved):
Beautiful. Those client abstractions are paying off.

mmarciniak90:
@grzesiek2010 I see that fake accuracy is cleared for Google Maps and OSM but when I use Mapbox I still see fake accuracy.

lognaturel:
Now I remember that the mapbox lib has its own abstraction similar to `LocationClient`. Hopefully there's a way to tie that in to ours. Maybe you'll need to add another layer that implements their interface but uses our implementation or something like that.

grzesiek2010:
> Now I remember that the mapbox lib has its own abstraction similar to LocationClient

Oh I missed that fact.

Ok now it should be handled as well. @lognaturel you may want to take a look at the implementation so I'm leaving this pr without `needs testing` flag for now.

collect_app/src/main/java/org/odk/collect/android/location/client/MapboxLocationClient.java
```diff
@@ -0,0 +1,33 @@
...
+import android.location.Location;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.gms.location.LocationListener;
+import com.mapbox.android.core.location.LocationEngineCallback;
+import com.mapbox.android.core.location.LocationEngineResult;
+
+public class MapboxLocationClient implements LocationEngineCallback<LocationEngineResult> {
```
lognaturel: I don't understand how the `LocationEngine` leak mentioned at https://docs.mapbox.com/android/core/guides/#requesting-location-updates would happen. Do you? Are you sure you're avoiding it here? It seems you still do have a strong reference to a Fragment here and so according to the example there it would be possible. It would be good to keep that link in the class comment since the structure is apparently important.

How about calling this `MapboxLocationCallback`? The `LocationClient` suffix to me implies it implements the interface of the same name. I think also augmenting the comment to say something like "Replace mock location accuracy with 0 as in `LocationClient` implementations since Mapbox uses its own location engine."

lognaturel: Oh, `engine.requestLocationUpdates` holds a reference to an instance of this callback class, right? So if that's a strong reference and the reference to `locationListener` is also a strong reference I think that means neither the activity nor the engine will ever be garbage collected.

lognaturel (changes requested):
Inline.

collect_app/src/main/java/org/odk/collect/android/location/client/MapboxLocationClient.java
grzesiek2010: Ok I added back WeakReference like it was in the original code. I think the issue is fixed now.

lognaturel (approved):

mmarciniak90:
@grzesiek2010 I noticed that accuracy appears on green bar when you rotate device. It happens for OSM, USGS, Stamen and Carto maps. Could you have a look at this case?

Additionally, If I try hard I'm able to save the fake location with accuracy value when I minimize app and come back to it plenty times - I know it's unlikely behavior.

grzesiek2010:
I think the problem is solved. The issue was that newly received locations were sanitized properly but just after starting recording location last known location is used first which wasn't sanitized.

mmarciniak90:
@grzesiek2010 thanks!

**Tested with success**

Verified cases:
- all GEO widgets 
- all maps
- offline layer
- audit with location
- background location
- map view with submissions
- green bar on map view with accuracy 0
- automatic and manual input method
- rotation -> accuracy doesn't appear on green bar 
- light/dark mode
- English/Spanish language
- simulate moving in Fake GPS app

kkrawczyk123:

PR #4359, opened 2021-01-26
Make sure redownloading the same form doesn't break anything [master]

### LINKED ISSUE #4355: Downloading a form with the same hash fails every other time

seadowg:
Closes #4355

@lognaturel fixed merged conflicts after #4356 was merged, ran the build and ran the tests.

#4357 was a simple fix for a point release. This is a more permanent one.

This makes changes to how forms are downloaded and moved so it's relatively risky. The test coverage is good which de-risks it but all of the same cases as https://github.com/getodk/collect/pull/4357#issuecomment-766981446 should be verified.

Any.

No.


lognaturel (approved):

mmarciniak90:
@seadowg something is wrong :cry: I can't download all forms from default server - I tried fresh instalation and reset settings.
When I try connect with Central, all forms are not downloaded (compared with master where all forms are visible)

You can try this user. 5 forms should be downloaded but I see only 2 forms. 
image

You can also try this Aggregate server. 8 forms should be downloaded.
collect-settings

mmarciniak90:

lognaturel:
Thank you, @mmarciniak90. This is a case of over-reliance on tests that use faulty doubles. The `InMemFormsRepository` always uses absolute paths for everything so we didn't see that this failed with scoped storage and relative paths in the db. I think we should run all tests with the equivalent of scoped storage now. I'm still working out the best way to do this and don't think it should block QA.

mmarciniak90:
**Tested with success**

Verified cases (followed list from  #4357):
- download using match exactly and then manually (Central, Aggregate, Kobo, ONA)
- download twice manually from the same server (Central, Aggregate, GD, Kobo, ONA)
- download twice manually from different servers (Central and then Aggregate)
- download the same and another version (Central)
- download when submission has been filled
- download twice Central draft
- same formId/version, different contents- Central draft -> error during finalizing when 2 forms with different content are available
- same formId/version, different media - Central draft -> only new media are downloaded
- download form with media
- download forms without media
- auto-update to the latest form version (match exactly; only previous forms; Central)
- cases from my previous comment

kkrawczyk123:

PR #4362, opened 2021-01-27
Allow crash reporting to be disabled

### LINKED ISSUE #4110: Allow disabling crash reporting
According to the [docs](https://docs.getodk.org/security-privacy/#odk-collect) disabling analytics does not disable crash reporting. This should not be the case.

As part of this it would be good to review if we can cut out any additional Firebase/Google data collection (like advertising ID) that we don't need to make our analytics as anonymous as possible.

**Acceptance**
 
- [x] Given I've disabled analytics
When a crash occurs
Then it does not appear in our crash reporting tool

- [x] Given I've enabled analytics
When a crash occurs
Then it does appear in our crash reporting tool

We should also create a docs issue once this is complete.

seadowg:
Closes #4110

Core changes here:

* Crash reporting is disabled when analytics is disabled by the user
* Analytics and crash reporting no longer start up with the app automatically - they become with enabled based on the user preference
* Performance monitoring has been removed - this will be easy to add back in when we want something specific
* Analytics/crash reporting code lives in its own module

I recommend running through this commit by commit as a lot of the changes are just imports.

Ran tests

This just changes analytics and crash reporting so can be merged without QA.


lognaturel (commented):
I tried this and did not see myself in the debug Firebase project. Did you get a chance to try it out, @seadowg? I sometimes have trouble with analytics showing up from my network so it could be that. The changes look ok but it would be good to have a sanity check.

seadowg:
@lognaturel I was able to see events using the debug steps (which I hadn't done for a while so looked up [here](https://firebase.google.com/docs/analytics/debugview)).

lognaturel (approved):

PR #4365, opened 2021-01-28
Workaround paused recording Android bug

### LINKED ISSUE #4364: Stopping recording while paused causes a crash when using `voice-only` quality
Collect v1.29.4, Android < 10

#### Problem description

Hitting the stop button (⏹) while recording while paused causes the app to hang before being killed by the system. This is due to an underlying issue in Android (logged [here](https://issuetracker.google.com/issues/178630865)).

This also causes the mic to be blocked, forcing the user to restart the device if they want to retry the recording.

1. Open a form with an audio question (using `voice-only`) and navigate to it
2. Start recording
3. Pause recording
4. Stop recording

The app will freeze and then Android will offer to kill it.

#### Expected behavior

Stop should work when paused.

seadowg:
Hot fix for #4364

Checked manually and wrote new tests.

Comments inline.

The only change here is to `voice-only` recordings, so we should focus any testing there. Good to check recording, pausing recordings etc on a few different devices if possible. We are now "resuming" recording before stopping for `voice-only` so it'd also be good to check that this doesn't introduce any artifacts into the recorded audio.

The `internal-audio-question.xml` form with a change to `voice-only` should do.

### DIFF
--- a/audiorecorder/src/main/java/org/odk/collect/audiorecorder/mediarecorder/MediaRecorderRecordingResource.kt
+++ b/audiorecorder/src/main/java/org/odk/collect/audiorecorder/mediarecorder/MediaRecorderRecordingResource.kt
@@ -1,9 +1,10 @@
 package org.odk.collect.audiorecorder.mediarecorder
 
+import android.annotation.SuppressLint
 import android.media.MediaRecorder
 import org.odk.collect.audiorecorder.recorder.RecordingResource
 
-internal abstract class MediaRecorderRecordingResource(private val mediaRecorder: MediaRecorder) : RecordingResource {
+internal abstract class MediaRecorderRecordingResource(private val mediaRecorder: MediaRecorder, private val sdk: Int) : RecordingResource {
 
     protected abstract fun beforePrepare(mediaRecorder: MediaRecorder)
 
@@ -22,13 +23,15 @@ internal abstract class MediaRecorderRecordingResource(private val mediaRecorder
     }
 
     override fun pause() {
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+        @SuppressLint("NewApi")
+        if (sdk >= android.os.Build.VERSION_CODES.N) {
             mediaRecorder.pause()
         }
     }
 
     override fun resume() {
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+        @SuppressLint("NewApi")
+        if (sdk >= android.os.Build.VERSION_CODES.N) {
             mediaRecorder.resume()
         }
     }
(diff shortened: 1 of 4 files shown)


audiorecorder/src/main/java/org/odk/collect/audiorecorder/mediarecorder/MediaRecorderRecordingResource.kt
```diff
@@ -22,13 +23,15 @@ internal abstract class MediaRecorderRecordingResource(private val mediaRecorder
     }
 
     override fun pause() {
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+        @SuppressLint("NewApi")
```
seadowg, line 26: The IDE warns us about this otherwise (even though we are performing the check).

audiorecorder/src/main/java/org/odk/collect/audiorecorder/mediarecorder/MediaRecorderRecordingResource.kt
```diff
@@ -66,4 +69,9 @@ internal class AMRRecordingResource(mediaRecorder: MediaRecorder) : MediaRecorde
         mediaRecorder.setAudioSamplingRate(8000)
         mediaRecorder.setAudioEncodingBitRate(12200)
     }
+
+    override fun stop() {
+        resume()
```
seadowg, line 74: As far as I can tell from docs and playing around calling `resume()` is safe (even when the recording isn't paused).

audiorecorder/src/test/java/org/odk/collect/audiorecorder/mediarecorder/AMRRecordingResourceTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Calls to stop() while paused hang which appears to be a problem in the Android
+     * framework. Resuming and the stopping does work however.
+     *
+     * @see <a href="https://gist.github.com/seadowg/e6cbab21e032fe2fcd9afb9b4f458ab9">Gist</a> for
+     * test demonstrating the issue in Android.
+     */
+    @Test
+    fun whenPaused_stop_callsResumeThenStopOnMediaRecorder() {
```
grzesiek2010: Don't we need another test but for lower android api?

grzesiek2010 (changes requested):

audiorecorder/src/test/java/org/odk/collect/audiorecorder/mediarecorder/AMRRecordingResourceTest.kt
seadowg: Do you mean for when `resume` isn't supported (below API 24)? I didn't need a test to drive out the change but I think you're right that it's important to check we don't call `MediaRecorder#resume` directly here - that would explode! I'll add the test.

grzesiek2010: Yes that's what I meant.

grzesiek2010 (approved):

mmarciniak90:
@seadowg I'm able to reproduce this issue only on Android 8.1. I can confirm that this fix works and crash doesn't longer occur.
During regression check, I noticed another crash related to `voice-only` quality - again, only Android 8.1.
App crashes when I click on stop and pause at the same time while recording. I'm going to create a separate issue for this.

Tested on Android versions: 8.1 (where the crash was repeatable), 7.0, 10.0

Verified cases:
- pause -> stop case
- voice-only quality
- normal quality
- low quality
- rotation
- click on notification -> pause -> stop
- stop recording before changing screen dialog:
   * swipe next/prev
   * click next/prev
   * new repeat group
   * jump screen
   * general settings
- blocking other audio widgets on the page
- light/dark mode

kkrawczyk123:
Verified on Android 7.1 and 9.0 and also did a regression check on Android 5.1. 
@seadowg I have also seen the crash that @mmarciniak90 mentioned but only on Android 7.1 device but not in all my tries.

PR #4367, opened 2021-01-28
Don't compute choice list to get underlying value

### LINKED ISSUE #4363: When going to a screen with a dynamic choice list, the choice list is computed twice
Collect v1.29; introduced when we added value changes to audit log

#### Problem description
Computing a dynamic choice list is an expensive operation because it involves evaluating an XPath expression over and over again as part of a linear search through all possible choices. Currently the list is computed twice when a screen flip occurs: once as part of building a select widget and another as part of logging an analytics event for that screen (even if values are not saved in the log!):
Screen Shot 2021-01-27 at 11 04 19 AM

Fill any form with a select with a choice filter (e.g. [cascade.xml.zip](https://github.com/getodk/collect/files/5883216/cascade.xml.zip)). Put a breakpoint at `FormDef.populateDynamicChoices` and see that it gets called twice when going to a question with a choice filter.

#### Expected behavior
The choice list should only be computed once.

lognaturel:
Closes #4363

Wrote and ran automated tests. Tried form from issue. Confirmed this halves the time to go to a question with a filtered choice list. @getodk/testers please confirm you also experience this on your device. https://drive.google.com/drive/u/0/folders/1QwX7E-u19jzXvzwkOpj2Y4X-xb9ihd1f is a good form for that (swiping to the settlement question).

There's a higher-level problem that choice lists get constantly recomputed whether anything that affects them has changed or not. I think we want to rethink how that works by e.g. caching `FormEntryPrompt`s instead of rebuilding them or caching the populated dynamic choices at another layer, etc. However, for this specific issue, I think it makes more sense that getting a node's underlying value would not require having a filtered choice list. There's just no use for it and it doesn't make sense.

I'm not completely happy with introducing a new method to `FormController` (`getAnswerUnderlyingValue`). However, I think it's the best place for it for now. I considered just making its calls inline in `AuditUtils` but I'd rather keep things that use JavaRosa classes in the `javarosawrapper` package and I like having a place for a longer comment. I'd like to replace `FormEntryPrompt.getAnswerText` but I'm not yet confident I understand the implications. We could introduce `getAnswerUnderlyingValue` in `FormEntryPrompt` and I'm happy to do that as a follow-up step that will require a JR update.

Test-wise, I like having an integration test for logging value changed audits. Prior to this we have no coverage at all. This brings a little bit of coverage to make sure new and old values are written, the `track-changes` attribute is respected, etc.

This feels low-risk to me. I think the tests give us pretty good coverage. The only code change is to registering a question value for the audit log. I think verifying various scenarios around that like setting a value for the first time, going back to a question and then leaving it again without changing the value, setting a value for the last question and then exiting the form would be sufficient. It should really only make a difference for select ones and select multiples but doing a quick check with e.g. strings would be good. Checking the actual audit log (e.g. on Central) would be good.

We've replaced code that computed the choice list and filtered out answer values that are no longer available choices with code that returns the raw answer value. However, I am quite confident that there's no way to change the answer without displaying the choice list which would also filter out answers that are no longer available. @getodk/testers something to experiment with is cases where you select some choice and then filter the list such that it is no longer available.

Any form with a select one, a select multiple and a question of another type that requests an audit with `track-changes`.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
@@ -56,7 +56,7 @@ public class AuditEventLogger {
      * Log a new event
      */
     public void logEvent(AuditEvent.AuditEventType eventType, FormIndex formIndex,
-                         boolean writeImmediatelyToDisk, String questionAnswer, long currentTime, String changeReason) {
+                         boolean writeImmediatelyToDisk, @Nullable String questionAnswer, long currentTime, String changeReason) {
         if (!isAuditEnabled() || shouldBeIgnored(eventType)) {
             return;
         }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditUtils.java
@@ -21,8 +21,8 @@ public class AuditUtils {
             try {
                 FormEntryPrompt[] prompts = formController.getQuestionPrompts();
                 for (FormEntryPrompt question : prompts) {
-                    String answer = question.getAnswerValue() != null ? question.getAnswerValue().getDisplayText() : null;
-                    auditEventLogger.logEvent(AuditEvent.AuditEventType.QUESTION, question.getIndex(), true, answer, currentTime, null);
+                    auditEventLogger.logEvent(AuditEvent.AuditEventType.QUESTION, question.getIndex(), true,
+                            formController.getAnswerUnderlyingValue(question.getIndex()), currentTime, null);
                 }
             } catch (FormDesignException e) {
                 throw new RuntimeException(e.getMessage(), e);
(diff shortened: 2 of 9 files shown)


seadowg (approved):

seadowg:
@lognaturel looks like there is a conflict to fix.

I'm really wondering if in future we should just remove the answer state from `FormEntryPrompt`. It's pretty scary that you can access the same data there but just in a massively inefficient manner. Anyway, not a change we could make now as far as I can tell.

lognaturel:
> we should just remove the answer state from FormEntryPrompt

I think I will be able to do some of that with changes I'm making to `ItemsetBinding`.

I've made more promising progress on that front than I thought I would. I'm thinking it will make more sense to add those changes into this PR before we QA.

lognaturel:
Given that we want to freeze by next Wednesday, I think we should go ahead with this low-risk change and consider the broader JR one for the next release.

lognaturel:
The JavaRosa update I'm working on will replace this so we can go straight with that change for v1.31.

PR #4371, opened 2021-01-29
Fixed centering texts in forms

### LINKED ISSUE #4370: Centered texts are not centered vertically
All versions

#### Problem description
Centered texts are not centered vertically
Screenshot_1611920658

Try the attached form: 
[centerText.xlsx](https://github.com/getodk/collect/files/5893894/centerText.xlsx)

#### Other information 
https://forum.getodk.org/t/how-to-align-text-at-the-center/32227/3

grzesiek2010:
Closes #4370

| Before  | After |
| Screenshot_1611920658  | Screenshot_1611924307  |

I added one test and also verified the fix manually.

The problem was that in some case [Html.fromHtml()](https://developer.android.com/reference/android/text/Html#fromHtml(java.lang.String,%20int)) might return values with new lines at the end for example. See:
https://stackoverflow.com/questions/9589381/remove-extra-line-breaks-after-html-fromhtml/9974562

It's a safe fix and we can just test the scenario described in the issue.

Yes I think we should add info about centering in https://docs.getodk.org/form-styling/ so that it's clear for our users.
https://github.com/getodk/docs/issues/1315

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
@@ -110,7 +110,7 @@ public class StringUtils {
     }
 
     public static CharSequence textToHtml(String text) {
-        return text == null ? "" : Html.fromHtml(markdownToHtml(text));
+        return text == null ? "" : trim(Html.fromHtml(markdownToHtml(text)));
     }
 
     public static String ellipsizeBeginning(String text) {
@@ -157,6 +157,26 @@ public class StringUtils {
         }
 
         return true;
+    }
+
+    private static CharSequence trim(CharSequence text) {
+        if (text == null || text.length() == 0) {
+            return text;
+        }
+
+        int len = text.length();
+        int start = 0;
+        int end = len - 1;
+        while (Character.isWhitespace(text.charAt(start)) && start < len) {
+            start++;
+        }
+        while (Character.isWhitespace(text.charAt(end)) && end > 0) {
+            end--;
+        }
+        if (end >= start) {
+            return text.subSequence(start, end + 1);
+        }
+        return text;
     }
 }
 
--- a/collect_app/src/test/java/org/odk/collect/android/utilities/StringUtilsTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/utilities/StringUtilsTest.java
@@ -20,6 +20,12 @@ public class StringUtilsTest {
     }
 
     @Test
+    public void textToHtml_shouldBeTrimmed() {
+        CharSequence observed = StringUtils.textToHtml("<p style=\"text-align:center\">Text</p>");
+        assertThat(observed.toString(), equalTo("Text"));
+    }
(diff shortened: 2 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
```diff
@@ -158,5 +158,25 @@ public static boolean isBlank(String string) {
 
         return true;
     }
+
+    private static CharSequence trim(CharSequence text) {
```
seadowg, line 162: Could we just use `String#trim` for this or is there an extra case we need covered? Looks like the test passes using `toString().trim()` but I might be missing something/test needs to be stricter.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
grzesiek2010: We can't  because we need to return `CharSequence` not string (to use it with `TextView.setText()`) otherwise html styling won't work

seadowg (approved):

mmarciniak90:
@grzesiek2010 I created form with multiselect [4371minimalTest.xml.txt](https://github.com/getodk/collect/files/5931676/4371minimalTest.xml.txt)

The appearance for multiple select should albo be improved.



grzesiek2010:
Unfortunately fixing that issue is not trivial. We concatenate all selected choices where every single choice might have its own styling and the result then is like:
`<p style="text-align:center">a</p>, <p style="text-align:center">b</p>, <p style="text-align:center">c</p>`
If you open any html editor and paste the text then the result will be the same like you noticed.

To fix the issue we would need to go through all the choices then if every choice should be centered remove the styling and add one general for the entire result. That's not easy and might be full of different cases if choices have different styling so it's not what we want here.

lognaturel:
I agree it's not worth fixing this specific case. @grzesiek2010 how about we add a note to the select multiple documentation and say that style will be applied to the list of selected choices and that this might have unexpected results so should be verified? We can even specifically call out centering as something that won't make sense with select multiple options.

grzesiek2010:
> how about we add a note to the select multiple documentation and say that style will be applied to the list of selected choices and that this might have unexpected results so should be verified?

shouldn't it be added to the [Form styling](https://docs.getodk.org/form-styling/#) section? I need to update it either way to let the users know centering is also possible.

mmarciniak90:
**Tested with success**

**Verified cases:**
- centered options in `select one` question
- centered options in `minimal` appearance
- centered question represented on `field-list`
- edit saved form
- send to Central
- light/dark mode
- English/Spanish language
- RTL language
- rotation
- minimization
- regression check select one/multi on All widget form

kkrawczyk123:
Verified also on Androids 5.1, 7.1 and 9.0. Answers are actually not centered on Android 5.1:
Screenshot from 2021-02-11 14-07-25

PR #4376, opened 2021-02-01
Hide stop/pause button while recording audio in the background

### LINKED ISSUE #4342: Can see background audio is being recorded

seadowg:
Closes #4342 
~~Blocked by #4354~~

Also:
* Adds waveform to recording overlay (as shown in [mock ups](https://forum.getodk.org/t/create-automatic-background-recording-of-interview-in-collect/18601/20))
* Refactor `AudioRecorderViewModel` to just `AudioRecorder` as it doesn't need to be a `ViewModel` (doesn't hold any state)
* Reverts an earlier change that let the user navigate the form while recording. This was pretty cool, but we're worried it's too big a change to the form flow to just sneak in with background audio.

Comments inline.

Only change should be the feature implementation. I think it makes sense to skip QA for this PR and do checks later once the background audio feature is more complete.


audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorder.kt
```diff
@@ -1,16 +1,15 @@
...
 import java.io.File
 import java.io.Serializable
 
 /**
- * Interface for a ViewModel that records audio. Can only record once session
+ * Interface for a component that records audio. Can only record once session
  * at a time.
  */
-abstract class AudioRecorderViewModel : ViewModel() {
+abstract class AudioRecorder {
```
seadowg, line 12: I'm getting a stronger and stronger feeling that we're going to want to wrap this in a `ViewModel` on the Collect side. I also keep confusing myself with lifecycle as this component's state is really at the singleton level and not related to the view/activity lifecycle.

lognaturel (approved):
Looks good overall. I'm going to merge now so you can easily build on top. Some things to look into:
- the back button is blocked while recording in the background but shouldn't be
- the waveform is not sensitive enough on any device I've tried. I think it makes sense to multiply the height by some constant and cap at the max. It's ok if it's pretty much always maxed out when someone is talking.
- on my Android 5.1 Galaxy Tab A, the waveform has an extra subset of a line:

device-2021-02-02-165607

PR #4377, opened 2021-02-02
Build debug code up front in CI

seadowg:
This adds an initial job to the CI pipeline that builds runs an `assembleDebug`. The output of this is then passed to the other jobs, so they don't have to recompile the (non-test) code. This should shorten the length of the `test_app` job as it won't have to build debug and test sources. It may also reduce the memory overhead, although it's unclear if that's related to the compilation work or not.

This shouldn't need QA as it's just changes to our CI config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,16 +43,6 @@ jobs:
       - attach_workspace:
           at: ~/work
 
-      - restore_cache:
-          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "collect_app/build.gradle" }}
-      - run:
-          name: Download dependencies
-          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 androidDependencies
-      - save_cache:
-          paths:
-            - ~/.gradle
-          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "collect_app/build.gradle" }}
-
       - run:
           name: Run code quality checks
           command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 checkCode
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -4,16 +4,13 @@
 version: 2
 
 references:
-  workspace: &workspace
-               ~/work
```
seadowg, line 8: Personally I find these references more confusing than helpful for smaller pieces of config like this. I considered getting rid of `android_config` as well, but that is used a bunch of different places.

.circleci/config.yml
```diff
@@ -30,6 +27,24 @@ jobs:
             - ~/.gradle
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "collect_app/build.gradle" }}
 
+      - run:
+          name: Compile code
+          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 assembleDebug
+      - persist_to_workspace:
+          root: ~/work
+          paths:
+            - .
```
seadowg, line 36: Here we're storing the whole workspace

.circleci/config.yml
```diff
@@ -30,6 +27,24 @@ jobs:
...
+
+  check_quality:
+    <<: *android_config
+    environment:
+      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx1024m" -Dkotlin.compiler.execution.strategy=in-process'
+    steps:
+      - attach_workspace:
```
seadowg, line 43: We don't need to do a `checkout` if we just attach the workspace.

.circleci/config.yml
```diff
@@ -165,16 +153,25 @@ jobs:
...
+          requires:
+            - compile
+      - test_modules:
+          requires:
+            - compile
+      - test_app:
+          requires:
+            - compile
       - build_instrumented:
+          requires:
```
lognaturel, line 173: Does this really require compile?

seadowg: It doesn't require is, but it does end up doing the same work. My thinking was to try and do the common step first and then let everything else run in parallel. It also makes these steps faster to rerun if we need to.

lognaturel (approved):

PR #4379, opened 2021-02-03
External video widget

grzesiek2010:
I tested the new widget manually and added automated tests.

It's a continuation of: https://github.com/getodk/collect/issues/4214

This time I didn't touch other widgets so we can focus on the new External video widget only and that would be enough.

[exTestForm.xlsx](https://github.com/getodk/collect/files/5919103/exTestForm.xlsx)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/ExVideoWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/ExVideoWidget.java
@@ -1,0 +1,145 @@
+package org.odk.collect.android.widgets;
+
+import android.annotation.SuppressLint;
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.util.TypedValue;
+import android.view.View;
+
+import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.core.model.data.StringData;
+import org.javarosa.form.api.FormEntryPrompt;
+import org.odk.collect.android.R;
+import org.odk.collect.android.databinding.ExVideoWidgetAnswerBinding;
+import org.odk.collect.android.formentry.questions.QuestionDetails;
+import org.odk.collect.android.utilities.ActivityAvailability;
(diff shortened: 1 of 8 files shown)


collect_app/src/main/java/org/odk/collect/android/widgets/ExVideoWidget.java
```diff
@@ -0,0 +1,148 @@
...
+            answerFile = (File) object;
+            if (answerFile.exists()) {
+                questionMediaManager.replaceAnswerFile(getFormEntryPrompt().getIndex().toString(), answerFile.getAbsolutePath());
+                binding.playVideoButton.setEnabled(true);
+                widgetValueChanged();
+            } else {
+                Timber.e("Inserting Video file FAILED");
+            }
+        } else if (object != null) {
+            Timber.e("ExVideoWidget's setBinaryData must receive a video file but received: %s", object.getClass());
```
lognaturel: I think if we do want to log something here we should log the MIME type. I don't really know if that's actionable on our end, though. It probably just means that a form called out to an app that can return files other than videos.

I don't think the user gets any feedback, though? Could we do a toast that says something like "Application returned an invalid file type"?

grzesiek2010: Ok I added it.

lognaturel (approved):

lognaturel (approved):

kkrawczyk123:
**Tested with success!**

Tested cases:
generated answer
RTL language
used forward/backward navigation buttons
remove response
save/finalize form
toast is visible when an external app is not available
launch counter app
light/dark mode
edit form
don't keep activities
rotation
minimize
send a submission to Central and Aggregate
different font sizes
regression check on ex string and numerical widgets

mmarciniak90:
Tested also with success on Android 7.0, 8.1, 10.0

grzesiek2010:
Closed in favor of https://github.com/getodk/collect/pull/4382

Issue #4374, opened 2021-01-29
Save/ignore dialog works during background recording


seadowg:
Clicking back while background recording (not foreground) shows save/ignore dialog. Save stops and saves recording. Ignore changes throws recording away.

**Acceptance**

- [x] Given I'm filling out a form with background recording
When I press the back button
And I click Save
Then the form saves and exits
And the recording is saved with the form

- [x] Given I'm filling out a form with background recording
When I press the back button
And I click Ignore
Then I exit the form
And the recording is not saved with the form

PR #4381, opened 2021-02-03
Allow enumerator to toggle background recording

### LINKED ISSUE #4343: Can toggle background audio recording
**Acceptance**

And I'm filling out a form
When I open the overflow menu
And I uncheck "Record audio"
And I confirm I want to stop recording
And I finish filling out the form
Then there is no new recording in the `recordings` directory (in the ODK external storage directory)

And I'm filling out a form
When I open the overflow menu
And I uncheck "Record audio"
Then the audio recording overlay disappears

And I'm filling out a form
And I've unchecked "Record audio"
When I open the overflow menu
And I check "Record audio"
And I finish filling out the form
Then there is a new recording in the `recordings` directory (in the ODK external storage directory) containing audio from when I checked 'Record audio"

And I'm filling out a form
And I've unchecked "Record audio"
When I open the overflow menu
And I check "Record audio"
Then the audio recording overlay appears

seadowg:
Closes #4343 

This also makes sure that the recording is saved **before** the form itself. This avoids any timing issues we might run into if the form saves before the recording can (and then the recording never will).

Comments inline.

It would be good to QA the [Change Reason Required feature](https://docs.getodk.org/form-audit-log/?highlight=audit#reason-for-changes)  as there were some changes under the hood to this area of the code. There are tests for this so hopefully we're covered.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/SaveAsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/SaveAsTest.java
@@ -1,0 +1,32 @@
+package org.odk.collect.android.feature.formentry;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.TestRuleChain;
+
+@RunWith(AndroidJUnit4.class)
+public class SaveAsTest {
+
+    public final CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public final RuleChain chain = TestRuleChain.chain()
+            .around(rule);
+
+    @Test
+    public void fillingFormNameAtEndOfForm_savesInstanceWithName() {
+        rule.mainMenu()
+                .copyForm("one-question.xml")
+                .startBlankForm("One Question")
+                .swipeToEndScreen()
+                .fillInFormName("My Favourite Form")
+                .clickSaveAndExit()
+                .clickSendFinalizedForm(1)
+                .assertText("My Favourite Form");
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -26,9 +26,6 @@ import android.media.MediaPlayer;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
-import android.text.Editable;
-import android.text.InputFilter;
-import android.text.TextWatcher;
 import android.view.ContextMenu;
 import android.view.ContextMenu.ContextMenuInfo;
 import android.view.KeyEvent;
@@ -36,12 +33,9 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.MotionEvent;
 import android.view.View;
-import android.view.View.OnClickListener;
 import android.view.ViewGroup.LayoutParams;
 import android.view.animation.Animation;
 import android.view.animation.Animation.AnimationListener;
-import android.widget.CheckBox;
-import android.widget.EditText;
 import android.widget.FrameLayout;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
@@ -84,6 +78,7 @@ import org.odk.collect.android.dao.helpers.InstancesDaoHelper;
 import org.odk.collect.android.events.ReadPhoneStatePermissionRxEvent;
 import org.odk.collect.android.events.RxEventBus;
 import org.odk.collect.android.exception.JavaRosaException;
+import org.odk.collect.android.formentry.FormEndView;
 import org.odk.collect.android.formentry.FormEntryMenuDelegate;
 import org.odk.collect.android.formentry.FormEntryViewModel;
 import org.odk.collect.android.formentry.FormIndexAnimationHandler;
@@ -146,7 +141,6 @@ import org.odk.collect.android.utilities.ApplicationConstants;
 import org.odk.collect.android.utilities.DestroyableLifecyleOwner;
 import org.odk.collect.android.utilities.DialogUtils;
 import org.odk.collect.android.utilities.ExternalAppIntentProvider;
-import org.odk.collect.android.utilities.FormNameUtils;
 import org.odk.collect.android.utilities.MultiClickGuard;
 import org.odk.collect.android.utilities.PlayServicesChecker;
 import org.odk.collect.android.utilities.ScreenContext;
@@ -218,9 +212,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         SelectMinimalDialog.SelectMinimalDialogListener, CustomDatePickerDialog.DateChangeListener,
         CustomTimePickerDialog.TimeChangeListener {
 
-    // Defines for FormEntryActivity
-    private static final boolean EXIT = true;
-    private static final boolean DO_NOT_EXIT = false;
     private static final boolean EVALUATE_CONSTRAINTS = true;
     public static final boolean DO_NOT_EVALUATE_CONSTRAINTS = false;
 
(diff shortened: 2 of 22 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/SaveAsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.TestRuleChain;
+
+@RunWith(AndroidJUnit4.class)
+public class SaveAsTest {
```
seadowg, line 13: There was no coverage of the "Instance naming". Added this in to give myself some while I was moving things around.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1284,70 +1250,47 @@ private View createViewForFormEnd(FormController formController) {
...
+                    showShortToast(R.string.save_as_error);
+                } else {
+                    formSaveViewModel.saveForm(getIntent().getData(), markAsFinalized, saveName, true);
                 }
-            });
-        } else {
+            }
+        });
+
+        if (!(boolean) AdminSharedPreferences.getInstance().get(AdminKeys.KEY_MARK_AS_FINALIZED)) {
```
seadowg, line 1285: I'd like to move more of the setup of this view into the class itself, but I think when we touch it next we can look at adding testing and doing that.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -35,14 +40,20 @@
...
         this.analytics = analytics;
+        this.preferencesProvider = preferencesProvider;
+        this.audioRecorder = audioRecorder;
     }
 
     @Override
     public void formLoaded(@NotNull FormController formController) {
         this.formController = formController;
+
+        if (hasBackgroundRecording()) {
```
seadowg: Felt like this made more sense as a place to start the recording. I'm also trying to hide away the fact that this is triggered by prefs as much as possible.

collect_app/src/main/java/org/odk/collect/android/formentry/audit/ChangesReasonPromptDialogFragment.java
```diff
@@ -56,10 +56,7 @@ public void afterTextChanged(Editable editable) {
         });
 
         toolbar.setOnMenuItemClickListener(item -> {
-            if (viewModel.saveReason()) {
-                dismiss();
-            }
-
+            viewModel.resumeSave();
```
seadowg, line 59: It felt weird to have two "interrupt save and then come back to it" methods

lognaturel (approved):
It took me a while to wrap my head around the sequencing for getting a change reason, saving audio and finally saving. It feels like it wants to be an event-driven state machine where something observes the save state and reacts to each state but instead `saveForm` and `resumeSave` end up explicitly checking for next states. Some of the code is duplicated but with slight differences (e.g. mutually exclusive branches vs. branches that return). Not a huge deal but maybe something to consider evolving.

seadowg:
@lognaturel totally agreed. I ended up going with what I have to avoid option paralysis, but I do feel like it's missing a trick. Hopefully we can make some improvements as part of the next few stories - #4374 will hopefully drive some out.

strings/src/main/res/values/strings.xml
```diff
@@ -990,4 +990,6 @@
     <string name="not_a_directory">%s exists, but is not a directory</string>
 
     <string name="select_odk_shortcut">Select ODK Shortcut</string>
+    <string name="record_audio">Record audio</string>
```
lognaturel: I spent too much time organizing this file for new strings to just pile up at the end!! 😄

Could you please move it to after the foreground recording strings? ~295. Would be great to get a comment too. `Text for a form entry setting to toggle background audio recording requested by the form.`

strings/src/main/res/values/strings.xml
```diff
@@ -990,4 +990,6 @@
...
+    <string name="stop_recording_confirmation">Do you really want to stop recording?</string>
```
lognaturel: This one can move with the string above. 

Maybe add more details with like "This form requests background audio recording. Disabling it will stop recording and discard existing audio. Are you sure you want to proceed?"

strings/src/main/res/values/strings.xml
seadowg: Ah sorry I keep meaning to come back to these but forgetting. Definitely continue telling me off for this!

mmarciniak90:
@seadowg I noticed that recording is started again when device is rotated or hierarchy view is opened even if `Record audio` was unchecked.  `Record audio` is automatically checked after those actions.

ezgif com-video-to-gif

Additionaly, I comapared behaviore of background recording with  location tracking in background. 
Maybe the behavior should be different and it's expected but when I uncheck tracking location I have to manually select it again. Audio recording is automatically select when I start a new form. 

ezgif com-video-to-gif(1)

seadowg:
> I noticed that recording is started again when device is rotated or hierarchy view is opened even if Record audio was unchecked.

Ah good catch.

> Additionaly, I comapared behaviore of background recording with location tracking in background.
Maybe the behavior should be different and it's expected but when I uncheck tracking location I have to manually select it again. Audio recording is automatically select when I start a new form.

Hmm good point. @lognaturel should disabling "Record audio" carry over between form filling sessions?

seadowg:
@mmarciniak90 ok I've fixed the rotation problem and made the behaviour consistent with track location. We've (@lognaturel and I) also decided to actually avoid restarting the recording when the user checks "Record Audio" (after they've unchecked it). This avoids complications under the hood that were causing us headaches. There is now a dialog instead that tells the user to reload the form if they want to start recording again.

mmarciniak90:
@seadowg there is still an issue connected with rotation and jumping from the hierarchy view. 
When I try to enable recording I see a clear message that I need to RE-OPEN form to record but recording starts when I rotate the device or jump from the hierarchy view.

ezgif com-video-to-gif

seadowg:
@mmarciniak90 ah sorry that will work right now. @lognaturel and I discussed the message thinking ahead for when the recording only starts when the form loads (which is introduced in #4385) but didn't take into account that it's confusing at this point. For the moment lets ignore that weirdness, and I'll add a note to the later PR. As long as the recording doesn't start after the user disables it we're all good.

kkrawczyk123:
@seadowg I have noticed issues:
- When recording in form is enabled and I am turning it off and start to use audio widget, pause and stop buttons do not show up so this recording doesn't really have any outcome. I am attaching gif, sorry about jumping in the hierarchy I couldn't find the Audio question XD when reproducing jump straight to audio question. 
ezgif com-video-to-gif (51)issue1

- The second issue is that when I am using recording on an audio widget I probably shouldn't be able to play around with enabling and disabling record audio in form. (in opposite record audio in question is disabled). When enable recording in the middle of recording using audio widget it's not that bad as I can see informing dialog that I have to reopen the form, but when I disable recording I don't have any option to save my recording on audio widget - user should not lose any data right? gif attached as my description might be confusing: 
ezgif com-video-to-gif (51)issue2


seadowg:
@kkrawczyk123 ah nice finds. We're certainly going to end up with weirdness when an audio widget is used as part of a form with background recording. I think maybe we should just disable the audio widget if a form has background recording. What do you think @lognaturel? We can add an issue for that if you agree. @mmarciniak90 @kkrawczyk123 if we pull that out as a separate issue are you happy for this to merge?

lognaturel:
Ah! Yes, disabling any foreground recording makes sense and agreed it should be a separate issue.

mmarciniak90:
> if we pull that out as a separate issue are you happy for this to merge?

I'm worried that I might have missed something but I know we will test this as a part of other PRs with this feature, so it sounds okay to me.

seadowg:
@mmarciniak90 yeah it's probably better to hold off and test once we have the form spec stuff in in #4385.

PR #4394, opened 2021-02-10
Show errors in recording overlay

### LINKED ISSUE #4383: Recording overlay shows errors
**Acceptance**

- [x] Shows recording start errors just for background (pop up is also there)

- [x] Shows warning when recording disabled

lognaturel:
Closes #4383

I wasn't able to finish this and am handing off to @seadowg. I believe the "Recording disabled" message fully works as intended. However, to show a message when recording failed, some additional state needs to be saved. I thought we could query the current session but that gets cleaned up after the message is shown, even in the background case.

The intended behavior for the "Record audio" toggle is that the "Recording disabled" message should always be shown when the toggle was off at some point in a form filling session. That is, even if it gets toggled back on, because the recording doesn't resume, the message should show. There's one edge case exception:
1. launch form with "Record audio" enabled
1. toggle "Record audio" off
1. toggle "Record audio" back on
1. rotate screen

In that case, the "recording disabled" message isn't visible anymore even though recording is still disabled for this form filling session. I think this is ok -- it's not very likely to happen and the resulting state is not so bad.

Added and ran automated tests. Tried manually with forms on QA server. 

For the message on failure to start recording in the background case, I changed the code to explicitly throw an exception in `AudioRecorderService.startRecording`. I don't know of another way to trigger the error.

There are two distinct cases for the "Record audio" toggle: when recording is already off at form launch and when recording gets toggled off during a form filling session. These are different because when `AudioRecordingControllerFragment` is first built, the form isn't loaded yet and so we have no way of knowing whether it requests background recording. I tried to combine them but couldn't come up with a way. With the failure to start case, I think the separated structure makes sense.

Right now the "recording disabled" message is shown if the form requests recording without regard of whether a recording action has actually been triggered. That's generally right for now because recording is typically triggered by any entry into a form. However, it's not strictly correct because it would be possible to manually edit form XML and trigger background recording on `odk-instance-first-load`. I think that it's unlikely enough that it doesn't matter and we can handle different triggering events or multiple recordings if/when we support multiple recordings.

Regression risks are centered around the recording overlay and what's displayed on it when.

Forms with and without background and foreground recording.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
@@ -18,6 +18,7 @@ import org.odk.collect.android.databinding.AudioRecordingControllerFragmentBindi
 import org.odk.collect.android.formentry.BackgroundAudioViewModel;
 import org.odk.collect.android.formentry.FormEntryViewModel;
 import org.odk.collect.android.injection.DaggerUtils;
+import org.odk.collect.android.utilities.TranslationHandler;
 import org.odk.collect.audiorecorder.recording.AudioRecorder;
 import org.odk.collect.strings.format.LengthFormatterKt;
 
@@ -70,7 +71,7 @@ public class AudioRecordingControllerFragment extends Fragment {
             } else if (session.getFile() == null) {
                 binding.getRoot().setVisibility(VISIBLE);
 
-                binding.timeCode.setText(LengthFormatterKt.formatLength(session.getDuration()));
+                binding.recordingStatusMessage.setText(LengthFormatterKt.formatLength(session.getDuration()));
                 binding.waveform.addAmplitude(session.getAmplitude());
 
                 if (session.getPaused()) {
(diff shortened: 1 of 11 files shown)


collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
```diff
@@ -105,5 +106,33 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         });
 
         binding.stopRecording.setOnClickListener(v -> audioRecorder.stop());
+
+        formEntryViewModel.isFormControllerSet().observe(getViewLifecycleOwner(), isSet -> {
+            if (formEntryViewModel.hasBackgroundRecording()) {
+                if (audioRecorder.getCurrentSession().getValue() != null && audioRecorder.getCurrentSession().getValue().getFailedToStart() != null) {
```
lognaturel, line 112: This unfortunately doesn't work because the session is cleared after a failure.

collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
```diff
@@ -105,5 +106,33 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         });
 
         binding.stopRecording.setOnClickListener(v -> audioRecorder.stop());
+
+        formEntryViewModel.isFormControllerSet().observe(getViewLifecycleOwner(), isSet -> {
```
lognaturel, line 110: It doesn’t feel great to set up an observer for something we know will only be set once but I can’t think of another way to change the view once the form controller is set.

PR #4395, opened 2021-02-10
Add audit event for disabling/enabling background recording

### LINKED ISSUE #4380: Log audit events for background recording
Follow on to #4343 

And I'm filling out a form
When I open the overflow menu
And I uncheck "Record audio"
Then a "background audio disabled" [event](https://docs.getodk.org/form-audit-log/#event-types) is added to the audit log

And I'm filling out a form
And I've unchecked "Record audio"
When I open the overflow menu
And I check "Record audio"
Then a "background audio enabled" [event](https://docs.getodk.org/form-audit-log/#event-types) is added to the audit log

seadowg:
Closes #4380 

Added tests.

Not a lot to comment on here.

Should just add the feature. Good to test that events appear in the audit log as expected.

Any form with background audio and auditing enabled.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -521,6 +521,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         identityPromptViewModel.formLoaded(formController);
         formEntryViewModel.formLoaded(formController);
         formSaveViewModel.formLoaded(formController);
+        backgroundAudioViewModel.formLoaded(formController);
     }
 
     private void setupFields(Bundle savedInstanceState) {
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioViewModel.java
@@ -5,6 +5,7 @@ import android.os.Handler;
 import android.os.Looper;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
 import androidx.lifecycle.ViewModel;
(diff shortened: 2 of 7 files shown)


lognaturel (approved):

mmarciniak90:
@seadowg I compared behavior for audio events with location events and I noticed that there is no event when you start a form (location event is visible). If you don't change the audio settings, there isn't any information about audio events in audit.

[audit-audio.csv.txt](https://github.com/getodk/collect/files/5965571/audit-audio.csv.txt)

Additionally, when permissions aren't granted, location events are visible but audio events are not.
[audit-permission.csv.txt](https://github.com/getodk/collect/files/5965574/audit-permission.csv.txt)

I used this form 
[AuditTestLocationBackgroundAudio.xml.txt](https://github.com/getodk/collect/files/5965603/AuditTestLocationBackgroundAudio.xml.txt)

seadowg:
> I noticed that there is no event when you start a form (location event is visible)

We never planned to have an event for this, but it's a great point as we should make sure to point out in docs what event the start of the recording corresponds to (probably "form start" and "form resume" looking at https://docs.getodk.org/form-audit-log/).

> Additionally, when permissions aren't granted, location events are visible but audio events are not.

Ah that's an interesting point. @lognaturel do you think we should add an event for if the record permission is denied so it's clear if that happens why the recording is missing?

lognaturel:
> point out in docs what event the start of the recording corresponds to

I've made a note of this.

Regarding permissions, this would be to handle the case where someone opens the form with granted audio permissions and then revokes the permission? When we had previously talked about this I didn't feel it was necessary given how unlikely that is but it's true that it gives a data collector an "out". @seadowg is it a big thing to add? Now I'm remembering there may be a relatively easy way to listen for those Android pref changes.

seadowg:
@lognaturel I was more thinking about the case where they open the form without audio permissions and then deny them. I guess the absence of a recording flags that that enumerators aren't granting permissions it feels like it might be nice to have something explicit in the log so analysts know what's gone wrong.

lognaturel:
> nice to have something explicit in the log so analysts know what's gone wrong.

I agree. If it's not a big lift, could you please see about adding for v1.30?

> open the form without audio permissions

I think you mean with and if so, that's what I was trying to express. It shouldn't currently be possible to open the form without audio permissions.

mmarciniak90:
Final results for this PR:
* Event `background audio enabled` is visible in the audit file
* Event `background audio disabled` is visible in the audit file
* If user doesn't change the audio settings, there isn't any information about audio events in the audit.
* If microphone permission is not granted, there is no possibility to fill form so the "permission" events in the audit are not required.

Tested on Android versions: 7.0, 7.1, 8.1, 9.0, 10.0

PR #4399, opened 2021-02-10
Show message when background audio recording is disabled

### LINKED ISSUE #4383: Recording overlay shows errors

seadowg:
Work towards #4383

Comments inline.

Should just add the feature. Good to check background audio in it's entirety when this and everything else are merged to master.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
@@ -79,6 +79,7 @@ public class AudioRecordingControllerFragment extends Fragment {
         if (session == null && hasBackgroundRecording && !isBackgroundRecordingEnabled) {
             binding.getRoot().setVisibility(VISIBLE);
             binding.timeCode.setText(TranslationHandler.getString(requireContext(), R.string.recording_disabled, "⋮"));
+            binding.waveform.setVisibility(GONE);
             binding.pauseRecording.setVisibility(GONE);
             binding.stopRecording.setVisibility(GONE);
         } else if (session == null) {
--- a/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
@@ -78,6 +78,7 @@ public class AudioRecordingControllerFragment extends Fragment {
     private void update(boolean hasBackgroundRecording, boolean isBackgroundRecordingEnabled, RecordingSession session) {
         if (session == null && hasBackgroundRecording && !isBackgroundRecordingEnabled) {
             binding.getRoot().setVisibility(VISIBLE);
+            binding.recordingIcon.setImageDrawable(ContextCompat.getDrawable(requireContext(), R.drawable.ic_baseline_mic_off_24));
             binding.timeCode.setText(TranslationHandler.getString(requireContext(), R.string.recording_disabled, "⋮"));
             binding.waveform.setVisibility(GONE);
             binding.pauseRecording.setVisibility(GONE);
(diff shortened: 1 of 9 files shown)


collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
```diff
@@ -61,49 +64,64 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
-                }
-
-                if (backgroundAudioViewModel.isBackgroundRecording()) {
-                    binding.pauseRecording.setVisibility(GONE);
-                    binding.stopRecording.setVisibility(GONE);
-                }
-            } else {
-                binding.getRoot().setVisibility(GONE);
-            }
+        zip3(formEntryViewModel.hasBackgroundRecording(), backgroundAudioViewModel.isBackgroundRecordingEnabled(), audioRecorder.getCurrentSession()).observe(getViewLifecycleOwner(), triple -> {
```
seadowg, line 67: I looked at removing the zip here and seperating it out to different observations but I ended up in trouble as I had to make some of the same decisions in each observation (the `null` session case needs to reach out to `hasBackgroundRecording` vice versa).

I also tried using a "React-esque" approach of setting fields on the fragment for each observation and then updating, but I ended up in trouble as the first observation would fire before the next and so with unset fields being using to make decisions.

lognaturel: Thanks for looking into it. Let's go with this.

seadowg: Let's pair on this when we can as I think it's probably a nice test of how we're going to use `LiveData`.

collect_app/src/test/java/org/odk/collect/android/audio/AudioRecordingControllerFragmentTest.java
```diff
@@ -193,4 +226,30 @@ public void whenThereIsAnErrorStartingRecording_showsErrorDialog() {
...
+            assertThat(fragment.binding.getRoot().getVisibility(), is(View.VISIBLE));
+            assertThat(fragment.binding.timeCode.getText(), is(fragment.getString(R.string.recording_disabled, "⋮")));
+            assertThat(fragment.binding.pauseRecording.getVisibility(), is(View.GONE));
+            assertThat(fragment.binding.stopRecording.getVisibility(), is(View.GONE));
+            assertThat(fragment.binding.waveform.getVisibility(), is(View.GONE));
+        });
+    }
+
+    @Test
+    public void whenFormDoesNotHaveBackgroundRecording_andBackgroundRecordingIsDisabled_showsThatRecordingIsDisabled() {
```
lognaturel: Is this really `doesNotShowThatRecordingIsDisabled`?

strings/src/main/res/values/strings.xml
```diff
@@ -196,7 +196,12 @@
 
     <!-- Error message displayed when copying media from a temporary directory fails -->
     <string name="answer_file_copy_failed_message">Media could not be attached to form but is available at %1$s</string>
-    
+
+    <string name="saving_audio_recording_failed">Saving audio recording failed</string>
```
lognaturel: I don't think this belongs here.

lognaturel (approved):

strings/src/main/res/values/strings.xml
seadowg: It felt related to the answer file copy failure to me. Where do you think it should go?

lognaturel: Isn't it a new string that's not actually used?!

seadowg: huh I see! Hmmm I'll fix, but I'm weirded out that lint didn't catch this...

mmarciniak90:
**Tested with success**

| Screenshot_20210211-085113|Screenshot_20210211-090912|
|---|---|

Verified cases:
- clear message is visible when recording is disabled
- different font size
- new form with disabled recording
- edit saved form
- light/dark mode
- English/Spanish language
- RTL language
- minimize
- rotation

kkrawczyk123:

PR #4401, opened 2021-02-10
Fully remove Firebase Performance to fix release builds

### LINKED ISSUE #3739: Add custom Firebase Performance traces or remove dependency
Firebase Performance monitoring was added some time ago. Currently we only collect the default traces and that information has not been actionable. At the same time, the dependency has caused issues for various contributors (e.g. https://github.com/opendatakit/collect/blob/master/docs/WindowsDevSetup.md#known-issues). For v1.27, we should either [add actionable custom traces](https://firebase.google.com/docs/perf-mon) or remove the dependency.

One actionable trace I can think of is registering form load time. We know that some folks are spending a long time loading forms. It would be great to know what the range of load times are. It would also be helpful to get separate traces for load time during initial metadata discovery, cache building time, and time to load form from cache. We won't know anything about the form definitions but we'll be able to answer questions like:
- How common is it to get a multi-minute load time?
- What kinds of devices are used that take the longest time?
- Does the cache generally help?

Some other:
- loading select lists
- loading the image map widget
- finalizing a submission
- loading the form submission map

lognaturel:
Closes #3739

Release builds were failing on the `transformClassesWithFirebasePerformancePluginForXXX` step. Removing the Firebase Performance dependency fixes this.

Made a self-signed release build, verified that it runs.

We no longer enable Firebase Performance anyway so removing the dependency makes sense. This has positive side effects like speeding up the build.

None.

No.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,6 @@ buildscript {
         classpath 'com.google.gms:google-services:4.3.4'
         classpath 'org.jacoco:org.jacoco.core:0.8.6'
         classpath 'com.google.firebase:firebase-crashlytics-gradle:2.4.1'
-        classpath 'com.google.firebase:perf-plugin:1.3.4'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jlleitschuh.gradle:ktlint-gradle:9.4.1"
     }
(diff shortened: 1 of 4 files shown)


grzesiek2010:
Looks like we can remove it from https://github.com/getodk/collect/blob/master/docs/WindowsDevSetup.md#current-workarounds too

lognaturel:
Thank you, @grzesiek2010!

grzesiek2010 (approved):

PR #4404, opened 2021-02-11
Make background audio confirmation dialog buttons more specific

### LINKED ISSUE #4393: Confirmation dialog for turning off Record Audio toggle should have cancel button
It's possible to dismiss the dialog by tapping outside of it but that's not intuitive.

lognaturel:
Closes #4393

While I was reviewing this behavior, I also made the following changes:
- Change confirmation button text from "OK" to "Disable Recording". It didn't feel very clear that "OK" would actually make the change since "OK" is often used as an acknowledgement that the text was understood.
- We really want users who are re-enabling background recording to read the notice about the recording not resuming for this session. For that reason, I made the dialog not cancelable.

Manual checks.

These are subjective user experience changes. I am prioritizing them because I think this whole override is pretty confusing and we want to be as precise as we can.

There's more we can do here. The dialog text when disabling recording, re-enabling it and then disabling it again is pretty confusing since recording is already disabled at that point. But that seems less likely to happen so I think it's ok for now.

No regression risk as this is just changing strings. Hopefully this makes the override slightly easier to understand and makes it less likely that a user will accidentally disable recording.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryMenuDelegate.java
@@ -145,12 +145,14 @@ public class FormEntryMenuDelegate implements MenuDelegate, RequiresFormControll
             if (!enabled) {
                 new MaterialAlertDialogBuilder(activity)
                         .setMessage(R.string.stop_recording_confirmation)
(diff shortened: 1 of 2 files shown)


seadowg (approved):

mmarciniak90:
**Tested with success**

**Verified cases:**
- click on `DISABLE RECORDING` on dialog
- clicks outside the dialog
- click on `CANCEL`
- minimization
- rotation
- light/dark mode

kkrawczyk123:

PR #4405, opened 2021-02-11
Increase sensitivity of overlay waveform

lognaturel:
We're considering a slight cosmetic change to how the waveform appears on the recording overlay. This is a minor, low-risk change that might end up coming in after regression testing is done. The problem we have right now is that by representing the full volume range in the waveform, the difference in volume as someone speaks is barely visible.

Folks who have experimented with a range of devices report that device microphones can reliably represent at most 90-100dB (https://stackoverflow.com/a/14870458/137744). That's a pretty big range (dB is a log scale) but it's bounded. Typical human speech is between 50 and 65 dB and shouting is around 80dB.

The raw max amplitude values are a linear scale with the highest sound pressure the mic can represent being 32767, the max value of a 16-bit unsigned int (https://stackoverflow.com/a/10954785/137744).

The library we're using takes those values and scales them down for display. To make the waveform more sensitive, I propose simply scaling up the source amplitude. This means we'll clip the upper range and not be able to differentiate between loud and extremely loud values. These are outside of human voice range so not relevant to us.

Ideally the waveform would show movement from a quiet speaking voice to a loudly projected voice when most devices are held at arm's reach. I think it's better to exaggerate the movement at the low end to provide feedback that something is happening. Even if a lot of speech is clipped off the top, we still need to breathe and pause so there'll be some visual feedback that something is going on.

I've landed on a multiplier of 6 after experimenting with some devices. It's not perfect but I do think it's a big improvement.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
@@ -92,7 +92,7 @@ public class AudioRecordingControllerFragment extends Fragment {
             binding.getRoot().setVisibility(VISIBLE);
 
             binding.timeCode.setText(LengthFormatterKt.formatLength(session.getDuration()));
-            binding.waveform.addAmplitude(session.getAmplitude());
+            binding.waveform.addAmplitude(6 * session.getAmplitude());
 
             if (session.getPaused()) {
                 binding.pauseRecording.setIcon(ContextCompat.getDrawable(requireContext(), R.drawable.ic_baseline_mic_24));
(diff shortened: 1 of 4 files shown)


lognaturel:
Questions:
- Does this look awful on any devices we have?
- Should we do this for the inline foreground recording waveform, too?

seadowg:
Yeah this looks great! Tried out in a few different scenarios, and it's good at signalling if there's a lot of background noise and also if you're talking too quietly/too far away from the mic. I'll have a look at the test failure and conflicts to see if I can push this along.

seadowg:
@lognaturel personally I'm not a fan of using this for the inline recording waveform. I think on that one you get a better sense of what's going on due to the extra height (and the "history"). That said, I only feel mildly about it - not strongly.

seadowg:
One last thing: I'm going to bundle some other design tweaks (like disabling general settings) in here.

seadowg:
@lognaturel re-ran test lab so this is good to go if you're happy with my changes

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Beta 1

The visual difference is very noticeable. Waveform reaches its maximum range even with the sound of the voice.

lognaturel:
> Waveform reaches its maximum range even with the sound of the voice.

Subjectively, did you find it more useful?

mmarciniak90:
Yes, it's more expressive and it attracts more attention.

lognaturel:
🎉 🎉 Goals achieved! That's helpful feedback because every device is going to look a little different.

mmarciniak90:

PR #4408, opened 2021-02-11
Catch nulls returned by getIndexFromXPath()

### LINKED ISSUE #4407: NPE in FormEntryModel.java:423
Collect all versions

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/5f0ad2af80b59bb59b4ff81a9e01e1f3?time=last-ninety-days&versions=v1.29.5%20(4056)&sessionEventKey=6025215403DB00013116CA2D3CCF78AC_1506425861413051808

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.javarosa.core.model.FormIndex.isInForm()' on a null object reference
       at org.javarosa.form.api.FormEntryModel.createModelIfNecessary(FormEntryModel.java:423)
       at org.javarosa.form.api.FormEntryModel.setQuestionIndex(FormEntryModel.java:281)
       at org.javarosa.form.api.FormEntryController.jumpToIndex(FormEntryController.java:230)
       at org.odk.collect.android.javarosawrapper.FormController.jumpToIndex(FormController.java:861)
       at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:208)
       at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:70)
```

grzesiek2010:
Closes #4407

I tested the fix with harcoded nulls. 
I was also able to reproduce the crash but I'm not sure if it's the same way our users do.
Steps are:
1. In developer settings set Background process limit to `no background processes`
2. Open any form with two questions and navigate to the second one
3. Put the app into the background
4. Replace the form you opened with the same one but without that second question.
5. Open the app

We use that method in two places:
- to set the last visited form index so that after returning to the app we are moved to the same place
- to set the index of a widget that is waiting for data

in both cases doing nothing if the value is null is totally fine and safe.

It should just catch nulls and I don't think this requires testing.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -233,7 +233,6 @@ public class FormController {
                     int event = stepToNextEvent(true);
                     while (event != FormEntryController.EVENT_END_OF_FORM) {
                         String candidateXPath = getXPath(getFormIndex());
-                        // Log.i(t, "xpath: " + candidateXPath);
                         if (candidateXPath.equals(xpath)) {
                             returned = getFormIndex();
                             break;
--- a/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
+++ b/collect_app/src/main/java/org/odk/collect/android/javarosawrapper/FormController.java
@@ -213,6 +213,7 @@ public class FormController {
         return value;
     }
 
+    @Nullable
     public FormIndex getIndexFromXPath(String xpath) {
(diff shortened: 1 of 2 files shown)


grzesiek2010:
@lognaturel @seadowg 
It's a fix for one of the most common crashes. Since it's pretty simple I think we can merge it without testing and add it to v1.30.

lognaturel (approved):

lognaturel:
🙏 🙏

PR #4416, opened 2021-02-15
Update spreadsheet locale to en_GB to avoid problems with date formats

### LINKED ISSUE #4415: Formatting dates in Google Sheets
Collect all versions

#### Problem description
Dates in spreadsheets might be interpreted using different formats what might cause problems like swapping days and months.

1. Create a simple form with a date question
2. Create a new spreadsheet and set its locale (File->Spreadsheet settings) to United States
3. Fill your form and upload it

#### Other information 
In Collect, we use `dd/mm/yy` format and strings that represent dates are uploaded using this format.
The problem has been originally reported on the forum https://forum.getodk.org/t/day-month-reversed-in-sheet/31629/

grzesiek2010:
Closes #4415

I tested the fix manually.

I analyzed the possible solutions and:
- it's not possible to specify single cell formats during uploading with the current implementation because we send a collection of plain strings.
- one of the other options would be to refactor our current implementation to use [AppendCellsRequest](https://developers.google.com/resources/api-libraries/documentation/sheets/v4/java/latest/com/google/api/services/sheets/v4/model/AppendCellsRequest.html) which allows uploading a collection of [CellData](https://developers.google.com/resources/api-libraries/documentation/sheets/v4/java/latest/com/google/api/services/sheets/v4/model/CellData.html) where we could specify formats of such cells. The problem is that we would not only need to refactor the code as I mentioned but also represent dates as numbers. It would require more work and would be much riskier!
- so I decided that the easiest possible solutions should be the best one here. We can set locale to United Kingdom when the sheet is initialized and that will solve the issue. If A user needs a different locale (and they do that consciously) it's possible to change it in settings after sending the first form and it won't be changed by ODK Collect then. Please notice that in most cases users probably do not need a different locale than en_GB, the problem stems from the fact that spreadsheets are generated for them with different locales based on their locations and they do not even notice it like in case of the user who reported the issue on the forum.

The change here is small and safe. The only difference compared to the master branch is that before sending data to an empty sheet the spreadsheet will be updated and its locale should be changed to `United Kingdom`. It should take place only once, so if we send the next forms it shouldn't happen again.

Any form with date questions.

Yes, I think it would be good to add a note about locales somewhere.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/gdrive/InstanceGoogleSheetsUploader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/gdrive/InstanceGoogleSheetsUploader.java
@@ -116,6 +116,7 @@ public class InstanceGoogleSheetsUploader extends InstanceUploader {
 
             TreeElement instanceElement = getInstanceElement(formFilePath, instanceFile);
             setUpSpreadsheet(spreadsheetUrl);
+            sheetsHelper.updateSpreadsheetLocaleForNewSpreadsheet(spreadsheet.getSpreadsheetId(), spreadsheet.getSheets().get(0).getProperties().getTitle());
             if (hasRepeatableGroups(instanceElement)) {
                 createSheetsIfNeeded(instanceElement);
             }
--- a/collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
@@ -26,10 +26,14 @@ import com.google.api.services.sheets.v4.model.UpdateSheetPropertiesRequest;
 import com.google.api.services.sheets.v4.model.UpdateSpreadsheetPropertiesRequest;
 import com.google.api.services.sheets.v4.model.ValueRange;
 
+import org.odk.collect.android.utilities.StringUtils;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+
+import timber.log.Timber;
 
 public class SheetsHelper {
 
@@ -79,6 +83,33 @@ public class SheetsHelper {
 
         // send the API request
         sheetsAPI.batchUpdate(spreadsheetId, requests);
+    }
+
+    public void updateSpreadsheetLocaleForNewSpreadsheet(String spreadsheetId, String mainSheetTitle) {
+       try {
+           if (!isNewSpreadsheet(spreadsheetId, mainSheetTitle)) {
+               return;
+           }
+
+           SpreadsheetProperties sheetProperties = new SpreadsheetProperties()
+                   .setLocale("en_GB");
+
+           List<Request> requests = new ArrayList<>();
+           requests.add(
+                   new Request().setUpdateSpreadsheetProperties(
+                           new UpdateSpreadsheetPropertiesRequest()
+                                   .setProperties(sheetProperties)
+                                   .setFields("locale")));
+
+           sheetsAPI.batchUpdate(spreadsheetId, requests);
+        } catch (IOException e) {
+            Timber.w(e);
+        }
+    }
+
+    private boolean isNewSpreadsheet(String spreadsheetId, String mainSheetTitle) throws IOException {
+        List<List<Object>> sheetCells = getSheetCells(spreadsheetId, StringUtils.ellipsizeBeginning(mainSheetTitle));
+        return sheetCells == null || sheetCells.isEmpty();
     }
 
     /**
--- a/collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
@@ -85,6 +85,7 @@ public class SheetsHelper {
         sheetsAPI.batchUpdate(spreadsheetId, requests);
     }
 
+    // Force a locale that correctly interprets dates sent by Collect, unlike en_US
     public void updateSpreadsheetLocaleForNewSpreadsheet(String spreadsheetId, String mainSheetTitle) {
        try {
            if (!isNewSpreadsheet(spreadsheetId, mainSheetTitle)) {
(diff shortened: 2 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/gdrive/InstanceGoogleSheetsUploader.java
```diff
@@ -116,6 +116,7 @@ public String uploadOneSubmission(Instance instance, String spreadsheetUrl) thro
 
             TreeElement instanceElement = getInstanceElement(formFilePath, instanceFile);
             setUpSpreadsheet(spreadsheetUrl);
+            sheetsHelper.updateSpreadsheetLocale(spreadsheet.getSpreadsheetId());
```
seadowg: Does this end up changing the locale for a sheet with existing submissions?

collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
```diff
@@ -81,6 +83,24 @@ public void resizeSpreadSheet(String spreadsheetId, int sheetId, int columnSize)
         sheetsAPI.batchUpdate(spreadsheetId, requests);
     }
 
+    public void updateSpreadsheetLocale(String spreadsheetId) {
+        SpreadsheetProperties sheetProperties = new SpreadsheetProperties()
```
seadowg: Feels like it would be good to have a comment here to explain why this is happening.

seadowg (changes requested):

seadowg:
Another thing I was wondering is if there's a more "universal" locale we can use (as opposed to "en_GB"). It's probably because I'm in the UK, but it makes me feel weird/guilty using the UK as the default here (although I guess we use "en_US" for pretty much everything else). Does "en" work or does that disappointingly give us the "en_US" date format?

collect_app/src/main/java/org/odk/collect/android/gdrive/InstanceGoogleSheetsUploader.java
grzesiek2010: That method checks if it's a new spreadsheet and only then updates the locale.

grzesiek2010:
> Another thing I was wondering is if there's a more "universal" locale we can use (as opposed to "en_GB"). It's probably because I'm in the UK, but it makes me feel weird/guilty using the UK as the default here (although I guess we use "en_US" for pretty much everything else). Does "en" work or does that disappointingly give us the "en_US" date format?

`en` doesn't work I decided to use `en_GB` because it's a default locale if I create a new spreadsheet and I'm in Poland so I guess it's a default value for many countries.

collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
grzesiek2010: Maybe... personally I don't like comments and prefer using git history but if you think it's valuable I can do that.

lognaturel: I understand the sentiment but I don't think our codebase is very well suited to this. It's really hard to find relevant commits when code has been refactored and in practice we move things around a lot. We'd also have to be more disciplined about rebasing to avoid noisy commits like "Refactored for readability" or "Fixed tests" or things like that which don't convey much information. As it is, we often have to go through many blame levels to find an explanation for a change. I don't find the Android Studio history tools particularly helpful in those cases or easy to access so I go to GitHub to find this type of information. I use a browser plugin to try to follow moved files but it doesn't always work and is of course not helpful if a method changed classes. If you have a tool or strategy that lets you get to explanations for changes quickly even through refactors, I'd love to see a demo. That would certainly change my thinking on this.

In this case, I think a simple note like "Force a locale that correctly interprets dates sent by Collect, unlike en_US" is a really helpful memory aide that doesn't require context switching.

lognaturel:
`en_GB` seems like the way to go to me. Given that we have this fix now and that it's so small, I think we should consider seeing with QA whether it makes sense to include in v1.30 given where they are with their regression pass.

collect_app/src/main/java/org/odk/collect/android/gdrive/sheets/SheetsHelper.java
grzesiek2010: Done.

grzesiek2010:
> I think we should consider seeing with QA whether it makes sense to include in v1.30 given where they are with their regression pass

Seems fine. It's an isolated fix that can't affect anything else so if we test it we can be sure that everything is ok. Given that we should have some time after finishing regression testing I'm ok with that but let's not jinx it.

collect_app/src/test/java/org/odk/collect/android/gdrive/SheetsHelperTest.java
```diff
@@ -99,6 +104,42 @@ public void getSpreadsheetTest() throws IOException {
         assertBatchUpdateCalled(1);
     }
 
+    @Test
+    public void whenNewSpreadsheetDetected_shouldBatchUpdateBeCalled() throws IOException {
+        doReturn(true).when(sheetsHelper).isNewSpreadsheet("spreadsheet_id", "Sheet1");
```
seadowg, line 109: It looks like this test file uses a spy to mock out parts of the object under test which y'all know I'm not a fan of. This is something we'd need to take some time to untangle (like headphones in a pocket) however as it's an existing problem. I think lets go with this for now!

By the way I do think these new tests have been structured really nicely. There's clear separation between what is testing `updateSpreadsheetLocaleForNewSpreadsheet` and what is testing `isNewSpreadsheet`. That'll make it a lot easier to pull apart later.

seadowg (approved):

mmarciniak90:
**Tested with success**

Tested on Android versions: 7.0, 8.1, 10.0

Verified cases:
* locale is changed to the United Kingdom in Spreadsheet settings 
* auto-change from The United States locale
* auto-change from Spain locale
* auto-change from Germany locale
* locale is not changed when Spreadsheet is not empty

kkrawczyk123:

PR #4432, opened 2021-02-16
Upgrade Google Services to get correct json config

lognaturel:
Same issue as https://github.com/getodk/collect/pull/4273; we upgraded Google Services again and hit the same bug.

Looked at [the release notes](https://firebase.google.com/support/release-notes/android#google-services_plugin_v4-3-5). Added a corrupt `google-services.json` file for the selfSignedRelease flavor. Confirmed that with v4.3.4 it was ignored and that with v4.3.5 the build fails.

The most important functionality that uses Google Services is location acquisition. I did a quick check to make sure it's still possible to get location indoors. This means we're getting the Fused location provider which comes from Google Services.

I considered downgrading but the changelog makes it look like the .5 point release was really about fixing this issue. I'd rather keep downgrading as an absolute last resort in general.

Risk is around anything that uses Play Services. I think it's low because the changelog for the .4 and .5 releases don't indicate many changes. I've verified the most important functionality which is around location.

I'm marking as `needs testing` just in case QA sees another reason for concern. I will include it in the next beta.

No.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:4.1.1'
-        // Do not upgrade until https://github.com/google/play-services-plugins/issues/163 is addressed
-        classpath 'com.google.gms:google-services:4.3.4'
+        classpath 'com.google.gms:google-services:4.3.5'
         classpath 'org.jacoco:org.jacoco.core:0.8.6'
         classpath 'com.google.firebase:firebase-crashlytics-gradle:2.4.1'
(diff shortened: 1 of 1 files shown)


mmarciniak90:
**Tested with success**

A quick check of the cases:
* reset settings
* download twice forms from the default server
* match exactly with the default server
* scan QR code with Central server (with match exactly)
* Aggregate server

mmarciniak90:
**Tested with success**

Verified on 1.30 Beta 1
I didn't notice any regression.

Verified cases:
- all Geo widgets on All widgets form
- fused location provider is visible
- all available maps
- fake location
- location in audit

kkrawczyk123:

PR #4437, opened 2021-02-17
Stop using unscoped storage + remove legacy settings

### LINKED ISSUE #4435: Stop using unscoped storage
#### Problem description
Starting from v1.31 we want to solely support scoped storage. That means we can get rid of:
- migration code
- requesting storage permissions
- all the if/else statements where we decided which storage should be used

### LINKED ISSUE #4440: Remove legacy settings import
We currently support importing settings from a `.json` or `.settings` file in the `odk` directory. This is not documented and also not really used by anyone so we should remove to prevent us from having to maintain these code paths.

grzesiek2010:
Closes #4435
Closes #4440

I reviewed implemented changes and performed some manual tests.

We just have to stop using unscoped storage and seems like the next release is a good time because we have given our users enough time to migrate.

I think we can merge this pr without testing. We are going to test it on a daily basis working on the app.

No.

No.


grzesiek2010:
It was sooo satisfying to remove almost 4k lines of code :smile: 

there is one more thing I would remove but it's not obvious so I wanted to discuss it first.
I'm talking about the fact that in many activities we check if odk dirs exist and if not we create them. I know it made sense when we used unscoped storage and it was easy to remove odk dirs during using the app or even other apps could have done that.
However, now we will use solely our dir not shared with forks etc. I don't think it still makes sense to keep our code cluttered and be overprotective. We could just initialize dirs when the app is initialized (ApplicationInitializer).
What do you think @lognaturel @seadowg?

seadowg:
@grzesiek2010 yeah I feel you're right. The user could delete the directories at any time so a lot of the checks on Activity launches still don't actually protect us. I think `ApplicationInitializer` makes sense. 

Another thing we might want to do with "projects" in mind is have something lazily create the dirs whenever they are needed. Ideally this would be the `DatabaseFormsRepository` but we haven't got it to a place where it's in control of the storage of the form files as well. My thinking is then it would be easier to be like "I need a form repository for project blah" and my storage is just created for me. That's probably getting ahead of things though? An alternative for right now could be to have `StoragePathProvider` create the dirs either on construction or when its first asked for a path.

grzesiek2010:
> Another thing we might want to do with "projects" in mind is have something lazily create the dirs whenever they are needed. Ideally this would be the DatabaseFormsRepository but we haven't got it to a place where it's in control of the storage of the form files as well.

The thing is that it's not only about forms but the problem might occur if you access forms, instances, settings, layers everything we store in our folder, so you would need to add it everywhere (in every repository etc) that seems pretty clunky again.

> An alternative for right now could be to have StoragePathProvider create the dirs either on construction or when its first asked for a path.

If we really want to be overprotective but clean the activities I think we could consider adding dirs initialization in [StoragePathProvider.getOdkRootDirPath() ](https://github.com/getodk/collect/pull/4437/files#diff-305ec7a913eed2277d0af06a536383d929631be636da49be48e21df4bfb625d4R18) since this is the root method used by all the other methods that return paths. 
To be honest, I'm not a fan of it, it's like fixing all disappearing dialogs after rotating a device even those that are not very important, etc. we talked about it some time ago.

seadowg:
> If we really want to be overprotective but clean the activities I think we could consider adding dirs initialization in StoragePathProvider.getOdkRootDirPath() since this is the root method used by all the other methods that return paths.
To be honest, I'm not a fan of it, it's like fixing all disappearing dialogs after rotating a device even those that are not very important, etc. we talked about it some time ago.

Yeah I think I agree. Thinking about it again we've already got that nice separation where `StoragePathProvider` knows where things are and `StorageInitialzer` actually does the work to setup the disk. Adding that setup to `StoragePathProvider` muddies the water somewhat. Happy to go ahead with your original idea!

grzesiek2010:
I included #4440 here since it's also about removing stuff.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.java
```diff
@@ -52,41 +47,23 @@ public void setup() {
...
+                                                                         SettingsPreferenceMigrator preferenceMigrator, PropertyManager propertyManager,
+                                                                         Analytics analytics, StorageInitializer storageInitializer) {
                 return applicationInitializer;
             }
         });
     }
 
     @Test
-    public void whenStoragePermissionGranted_createsODKDirectories() {
+    public void whenAppStarts_createsODKDirectories() {
```
seadowg: Could we just move this to a test for `ApplicationIntitializer`? Alternatively we could just think of it as being implicitly tested by our Espresso test. Your call!

seadowg (changes requested):
Just one small comment.

Another thing I was thinking about is: should we keep `IntentServiceIdlingResource` and `MaterialBanner`? These will probably be useful in the future. As I write that I feel silly though as we can always just get them out of history and why would we keep things we're not using around. Thoughts?

grzesiek2010:
> Another thing I was thinking about is: should we keep IntentServiceIdlingResource and MaterialBanner?

I was thinking about it as well especially about MaterialBanner. My thinking is that we don't know when we might need it again maybe in a few months, a year, or never. If we need something like that in a year will that still be the recommended solution (everything changes so quickly when it comes to UI/Material design)?
Even if we need it in a year and it's still the recommended component, we should remember that we had something like that and just restore it using git.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.java
grzesiek2010: I moved it to StorageInitializerTest.

seadowg (approved):

PR #4442, opened 2021-02-19
Fix background audio permissions flow

### LINKED ISSUE #4441: Collect crashes when user finalizes form if microphone permission was revoked and accepted while filling the form with background recording
Collect v1.30 Beta 1

#### Problem description
Collect crashes when user finalizes form if microphone permission was revoked and accepted while filling the form with background recording

Condition: 'Don't keep activity' is turned on

1. Open form with a background recording
2. Revoke microphone permission
3. Go back to form
4. Agree for microphone permission in form
5. Try to finalize form
6. Collect crashes 

stacktrace:
`2021-02-18 16:22:17.812 22910-22910/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: main Process: org.odk.collect.android, PID: 22910 java.util.NoSuchElementException at java.util.HashMap$HashIterator.nextEntry(HashMap.java:854) at java.util.HashMap$KeyIterator.next(HashMap.java:885) at org.odk.collect.android.formentry.RecordingHandler.handleBackgroundRecording(RecordingHandler.java:70) at org.odk.collect.android.formentry.RecordingHandler.lambda$handle$0$RecordingHandler(RecordingHandler.java:48) at org.odk.collect.android.formentry.-$$Lambda$RecordingHandler$V-ws2wEX8QL3nQl0XXD8NB2yn5M.onChanged(lambda) at androidx.lifecycle.LiveData.considerNotify(LiveData.java:131) at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:149) at androidx.lifecycle.LiveData.setValue(LiveData.java:307) at androidx.lifecycle.MutableLiveData.setValue(MutableLiveData.java:50) at org.odk.collect.android.formentry.saving.FormSaveViewModel.lambda$createAnswerFile$1$FormSaveViewModel(FormSaveViewModel.java:378) at org.odk.collect.android.formentry.saving.-$$Lambda$FormSaveViewModel$2uu05vBIQlrmlmrKOPHT6fFCUdQ.accept(lambda) at org.odk.collect.async.CoroutineScheduler$immediate$1.invokeSuspend(CoroutineScheduler.kt:17) at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106) at android.os.Handler.handleCallback(Handler.java:751) at android.os.Handler.dispatchMessage(Handler.java:95) at android.os.Looper.loop(Looper.java:154) at android.app.ActivityThread.main(ActivityThread.java:6165) at java.lang.reflect.Method.invoke(Native Method) at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888) at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:778)`

#### Expected behavior
Crash does not occur

#### Other information 
Noticed during testing #4434 
@getodk-bot label "Specific to Android version"

seadowg:
Closes  #4441

The user will now be returned to the main menu (with an error toast) if they end up in the scenario in the issue that would cause a crash. This will only happen in cases where the user starts a form with a savepoint (so it loads on the hierarchy page) with audio permissions denied and then grants audio permissions.

It's definitely not the best solution. There are definitely some things we could do to make this flow work without having to just bail out of the form. The two good ones (there are definitely some bad ones) that I've though of are:

1. Move the permission check and the state held about a recording while permission is being granted into the `audiorecorder` world. This would mean both internal and external recording could use the same mechanisms for handling the recording permissions. 
2. Use the `SavedStateHandle` in `BackgroundAudioViewModel` to store the state held about a recording while permission is being granted

I think **1** is the better solution here as it moves the state we need to hold onto while trying to start a recording into the lifecycle/realm of the recording rather than being subject to Android lifecycle. I'm becoming less and less interested in trying to deal with saved instance state as the whole thing is dealing with one big special case - rotation vs memory reclaim. 

Given implementing **1** just before release was probably a bad idea and that the crash feels like it's rare it made sense to me to flag the situation to the user and have them recover from it.

Should just fix the issue in question but I'd feel safer if we did some quick checks around background audio in general.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioPermissionDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioPermissionDialogFragment.java
@@ -3,6 +3,7 @@ package org.odk.collect.android.formentry;
 import android.app.Dialog;
 import android.content.Context;
 import android.os.Bundle;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -56,6 +57,11 @@ public class BackgroundAudioPermissionDialogFragment extends DialogFragment {
                 try {
                     viewModel.grantAudioPermission();
                 } catch (IllegalStateException e) {
+                    Toast.makeText(
+                            activity,
+                            "Could not start recording. Please reopen form.",
+                            Toast.LENGTH_LONG
+                    ).show();
                     activity.finish();
                 }
             }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioPermissionDialogFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/BackgroundAudioPermissionDialogFragment.java
@@ -44,18 +44,26 @@ public class BackgroundAudioPermissionDialogFragment extends DialogFragment {
         return new MaterialAlertDialogBuilder(requireContext())
                 .setMessage(R.string.background_audio_permission_explanation)
                 .setPositiveButton(R.string.ok, (dialog, which) -> {
-                    permissionsProvider.requestRecordAudioPermission(activity, new PermissionListener() {
-                        @Override
-                        public void granted() {
-                            viewModel.grantAudioPermission();
-                        }
-
-                        @Override
-                        public void denied() {
-                            activity.finish();
-                        }
-                    });
+                    onOKClicked(activity);
                 })
                 .create();
+    }
+
+    private void onOKClicked(FragmentActivity activity) {
+        permissionsProvider.requestRecordAudioPermission(activity, new PermissionListener() {
+            @Override
+            public void granted() {
+                try {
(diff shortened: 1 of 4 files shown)


grzesiek2010 (approved):
It's not a perfect solution but I'm ok with it. My approach has changed and I'm a bit tired of being overprotective and perfectly handle all cases where users do something strange so in this case I would put it in the same category with what I said about checking directories in activities or making sure that all dialog even those not important are recreated.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.0, 8.1, 9.0 and 10.

I agree that it is not the best solution but this an edge case so it might not be worth more work. 

Tested cases:
- scenario from the issue with don't keep activities enabled
- scenario from the issue with don't keep activities disabled
- background recording form
- background recording with audit form
- background recording with repeats form
- regression check on Android 5.1 and background recording in general 

Issue #563, opened 2017-03-10
Replace calling browser intent with custom chrome tabs.


getsanjeev:
Launching the browser is a heavy context switch and should not be practised much. Custom chrome tab is a better way to launch URLs. I think this would enhance user experience.
While launching ODK website in About section, we can replace browser intent call with chrome tabs.
Whats say @lognaturel ?

yanokwa:
@getsanjeev And what happens if Chrome isn't installed?

getsanjeev:
@yanokwa What I know is - In that case simply it will launch the URL in system browser(default). It would be good if you have some suggestions.

lakshyagupta21:
@getsanjeev You can use Webview in Activity to open About Section and store the webpage in cache so that if user opens the about section first it will retrieve the page from cache first, if page is not found in cache then it will load page from Internet , I think this would definitely enhance User Experience. @yanokwa Any suggestions ?

getsanjeev:
@lakshyagupta21  Sure it can be done.
The good thing about chrome custom tabs is that - without managing things we can provide customisation. Here we dont need to write bulky codes just to show a single webpage.

yanokwa:
[When should I use Chrome Custom Tabs vs WebView](https://developer.chrome.com/multidevice/android/customtabs#whentouse) suggests that this is a good optimization. 

Let's start slow, @getsanjeev. Replace the About browser launch in a branch. Then post in this issue what the experience is like on old devices with and without Chrome and on new devices with and without Chrome. 

Please use [LiceCap](http://www.cockos.com/licecap) or some other screen recorder so we can easily verify the behavior.

getsanjeev:
@yanokwa Sure. Thanks for your feedback.

shobhitagarwal1612:
@getsanjeev @yanokwa @lognaturel  I am a fan of `custom chrome tabs` as they make the web browsing experience in our own app much faster and stable. 

`Collect` calls for a browser intent only in the following cases:
- When visiting odk website from About Preferences
- When the user has to submit reviews and device doesn't have google play store app installed (but those devices might not have google chrome installed as well), also in the About Preferences

Since the scope of opening a browser is so little with such less use case, is it worth including another library in our app among already increasing number of external libraries? Also what will be the difference in the size of the apk after including the required gradle dependency for `custom chrome tabs`?

getsanjeev:
[...]

PR #4446, opened 2021-02-23
Add horizontal volume bar

seadowg:
This adds a horizontal volume bar to the recording overlay. It also adds a help button that links to the background audio recording docs (and is only shown during background recordings).

Comments inline.

Should just add the new features. Probably good to double-check background and normal recording to check nothing shows up when it shouldn't.


seadowg:
@lognaturel playing around with this some more it feels like we'd really need to segment it to get any extra value out of it - then trainers/designers could tell enumerators how many of blobs/pips they need etc. What do you think?

lognaturel:
We talked about this but for posterity:

> trainers/designers could tell enumerators how many of blobs/pips they need

I think it's really more of an aesthetic concern since we can't say much about the meaning of the volume indicator across devices. You make a valid argument in contexts where all the data collectors have the same device but I think that's increasingly rare because people use their personal devices.

lognaturel:
Some opinions:
- it's worth doing a quick dig into segmented bar libs to see what's possible.
- it's ok to just let it take whatever width is available, even with playback buttons. It's wide on a tablet in landscape but not terrible.
- segments should be of fixed width so going from portrait to landscape would mean more segments.
- the difference in sensitivity between it and the histogram is confusing for foreground recording. Maybe we should scale the histogram up just a bit?
- the volume indicator bar height should match the height of the timestamp text.

Less opinionated:
- it doesn't work on Galaxy Tab A, Android 5.1.

seadowg:
@lognaturel I've spent another hour on this and tried a couple of libraries (https://github.com/carlosmuvi/SegmentedProgressBar and https://github.com/rayzone107/SegmentedProgressBar). The former won't resolve as a dependency (really not sure what's going on) and the latter isn't licensed. I also had a quick look at https://github.com/muneikh/SegmentedProgressBar but it really doesn't look right for us.

I can definitely go ahead and build this, but I think it's probably at least a days work to get it working for different screen widths (with and without play buttons, screen orientations etc) and then more time with us tweaking back and forth. I'm going to assume you're OK with that cost and go ahead and build it but ping me if not!

lognaturel:
Let's do it but let's try hard to keep it to a day!

> get it working for different screen widths (with and without play buttons, screen orientations etc)

I'm hoping that with it taking the full width and the segments being of a fixed width it can be not too big a deal. I realize it may be simpler to use a fixed fraction for the segments and I think that's probably fine too (e.g. always divide whatever width you have in 12).

seadowg:
@lognaturel this is feeling ready to get put through its paces now. 

You'll see in `VolumeBar` there is a `getBestPipSize` method that's currently making the individual pips smaller when the width available drops (this number was to optimize the number of pips on a Nexus 5X while pause/stop was showing). I think if you're able to try on a tablet screen it might be worth adding a case for the pips to get larger as well (just go ahead and push to this branch). We should also consider what to do on really tiny devices.

lognaturel:
I don't have much to say about this -- it looks really good.

The Moto G5 has a pretty narrow screen with a low pixel density:
device-2021-03-02-225852

I think down to 3 segments would be fine. I don't think we need to do anything fancy and if anyone runs into issues they can report it.

The only tablet I have is the Galaxy Tab A which is more of a phablet. It looks maybe a little bit silly in landscape but not so bad (red is known issue):
device-2021-03-02-230219

I think if we want to do something about really wide screens maybe the thing to do is to cap the number of segments after all at something high like 20. Is that a big lift? Again, I think it's reasonable enough that we could leave it as-is and address any feedback that comes up on specific form factors.

Would be great to see the `?` at the far right for more details in the background case in this PR, too.

seadowg:
@lognaturel I like the idea of capping at 20. That's not a big deal at all. I'll look at that and then at the `?` button.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/BackgroundAudioRecordingTest.java
```diff
@@ -137,19 +137,6 @@ public void fillingOutForm_withMultipleRecordActions_recordsAudioOnceForAllOfThe
         assertThat(instanceXml, containsString("<recording2>" + recording.getName() + "</recording2>"));
     }
 
-    /**
-     * This could probably be tested at a lower level when the background recording implementation
-     * stabilizes.
-     */
-    @Test
-    public void fillingOutForm_doesntShowStopOrPauseButtons() {
```
seadowg, line 145: This is now covered at the Fragment test level

collect_app/src/main/java/org/odk/collect/android/utilities/ExternalWebPageHelper.java
```diff
@@ -14,7 +16,8 @@
 /**
  * Created by sanjeev on 17/3/17.
  */
-public class CustomTabHelper {
+public class ExternalWebPageHelper {
```
seadowg, line 19: It felt like this was a more appropriate name given it falls back from custom tabs

collect_app/src/main/java/org/odk/collect/android/utilities/ExternalWebPageHelper.java
```diff
@@ -59,16 +62,23 @@ public CustomTabsServiceConnection getServiceConnection() {
         return serviceConnection;
     }
 
-    public void openUri(Context context, Uri uri) {
+    public void openWebPageInCustomTab(Activity activity, Uri uri) {
```
seadowg, line 65: I didn't actually want this behaviour for the new help button. I'm also thinking we should stop using it for the website and forum links as it can prevent the user from multitasking between Collect and the web page (which they may need to do to read docs or post a problem on the forum). The custom tabs behaviour could useful for the `URLWidget` but honestly it still feels like a lot of code to maintain given the value it brings. Additionally, I just dislike the idea of using a special feature of the Chrome browser (that other browsers are then forced to implement) but that's just me being grouchy.

Interested in thoughts from @lognaturel and @grzesiek2010. It looks like the behaviour was introduced in https://github.com/getodk/collect/issues/563 (thanks for digging that up @grzesiek2010) and the docs on the feature are [here](https://developers.google.com/web/android/custom-tabs).

lognaturel:
I don't think the ? button should go to docs. That page does not seem appropriate for a data collector, it won't work offline and even with a reasonable connection that page is quite heavy. [In Slack](https://getodk.slack.com/archives/C35ENHA64/p1612384875054000?thread_ts=1612376697.039000&cid=C35ENHA64), we had discussed a message in a dialog. I think that achieves the goal of making it clear to the data collector what's going on. I'd shorten it to something like 
```
This form records the audio from your device's microphone. You can use the volume indicator to make sure the microphone is close enough to the sound you need to record. To stop the recording please exit this form. For more information please contact the person who asked you to collect data.
```
Once we have the basics in place we can play with the wording. We can also talk about this if you disagree.

collect_app/src/main/java/org/odk/collect/android/audio/VolumeBar.java
```diff
@@ -0,0 +1,121 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.odk.collect.android.R;
+
+import static androidx.core.content.res.ResourcesCompat.getDrawable;
+import static org.odk.collect.android.utilities.ViewUtils.dpFromPx;
+import static org.odk.collect.android.utilities.ViewUtils.pxFromDp;
+
+public class VolumeBar extends LinearLayout {
+
+    public static final int MAX_AMPLITUDE = 22760;
```
lognaturel: Could you please comment on this to make it clear where the value comes from? My understanding is that amplitude is a positive short so its range is between 0 and [Short.MAX_VALUE](https://docs.oracle.com/javase/7/docs/api/java/lang/Short.html#MAX_VALUE) (32767). If you did mean to use the highest possible value, I think I'd still create the constant for semantics but set it to Short.MAX_VALUE so it's really clear where it comes from.

collect_app/src/main/java/org/odk/collect/android/audio/VolumeBar.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+    }
+
+    public void addAmplitude(int amplitude) {
+        lastAmplitude = amplitude;
+
+        if (pips > 0) {
+            int segmentAmplitude = MAX_AMPLITUDE / pips;
+            int adjustedAmplitude = amplitude * 6;
```
lognaturel: Would be good to have a comment to explain the reasoning on this * 6 (scale up because the top of the range for most devices can't be reached by human voices. This makes the indicator more sensitive in the human voice range)

collect_app/src/main/java/org/odk/collect/android/utilities/ExternalWebPageHelper.java
lognaturel: I don't have a strong opinion and would rather let this go for now.

lognaturel (changes requested):
Looks good except for a mystery constant and the help behavior.

seadowg:
@lognaturel agreed about the help button. I don't think we explicitly talked about it (my bad, I should have clarified) and I got the idea of it going to docs stuck in my head.

collect_app/src/main/java/org/odk/collect/android/audio/VolumeBar.java
seadowg: Good catch. I'm using the max value that `AudioRecordView` rather than the max value I've realized. Definitely better to just use `Short.MAX_VALUE` here.

collect_app/src/main/java/org/odk/collect/android/audio/AudioRecordingControllerFragment.java
```diff
@@ -33,6 +34,8 @@
 
 public class AudioRecordingControllerFragment extends Fragment {
 
+    public static final String BACKGROUND_AUDIO_DOCS_URL = "https://docs.getodk.org/form-question-types/#background-audio-recording";
```
lognaturel: Remove.

collect_app/src/main/java/org/odk/collect/android/audio/VolumeBar.java
```diff
@@ -0,0 +1,129 @@
...
+
+import static androidx.core.content.res.ResourcesCompat.getDrawable;
+import static org.odk.collect.android.utilities.ViewUtils.dpFromPx;
+import static org.odk.collect.android.utilities.ViewUtils.pxFromDp;
+
+public class VolumeBar extends LinearLayout {
+
+    /**
+     * Amplitude is a reported by Android as a positive Short so
+     * the cannot be higher than the max value.
```
lognaturel: Some words missing? 😬

lognaturel (changes requested):
Couple of nit-picks inline.

I also feel like the help icon is a little too low. Is it exactly aligned with the microphone on the other side and exactly the same height? I'm talking if it is off it's just a couple points but it's oddly distracting. Maybe it's just because of the shape of each icon.

None of that blocks QA!

collect_app/src/main/java/org/odk/collect/android/audio/VolumeBar.java
seadowg: Wow yeah

mmarciniak90:
@seadowg I didn't notice nothing more than missing scroll bar in dialog under question mark icon.

|example with scroll bar|dialog without scroll bar|
|---|---|
Screenshot_20210309-151939|Screenshot_20210309-151954|

Tested on Android: 7.0, 8.1, 10.0

Verified cases:
* volume bar 
* vertical and horizontal device orientation with bar
* background recording
* audio widget
* question mark button
* rotation, scroll on dialog -> scroll bar is not visible
* light and dark mode
* RTL language

seadowg:
@mmarciniak90 that should be fixed now. Great catch by the way. It's going to be something to look out for as it looks like the settings dialogs get the scroll bars but that we have to do some extra work in other places for some reason. Most dialogs won't need them thankfully.

lognaturel:
Were you able to do a quick sanity check on the (?) icon position? From my review:

> I also feel like the help icon is a little too low. Is it exactly aligned with the microphone on the other side and exactly the same height? I'm talking if it is off it's just a couple points but it's oddly distracting. Maybe it's just because of the shape of each icon.

@mmarciniak90 didn't mention anything so that's probably a good sign. Would appreciate a quick check on your end in case it's something silly like an extra little bit of margin somewhere or something. I fully acknowledge it could just be my eyes.

seadowg:
@lognaturel sorry I totally missed that. Will check now.

seadowg:
@lognaturel it maybe does look slightly off, but I think it's just due to the shape of the icons rather than them actually out of alignment. Annoyingly there isn't an outlined mic or an unoutlined help.

mmarciniak90:
> that should be fixed now

@seadowg agree, scroll is available when it's needed.
Unfortunately, the last changes removed the margin in the dialog.

| scroll | margin|
|---|---|
Screenshot_20210311-174326|Screenshot_20210311-173951|

> I also feel like the help icon is a little too low

@lognaturel you have right. However, the difference is minimal. For me, the vertical line of icons is worse.

Screenshot_20210311-174746

lognaturel:
Thank you for that really useful drawing, @mmarciniak90! I think the ⋮ icon is deliberately right-aligned in the icon definition itself, for better or worse. I think we're better off leaving that one alone though I do agree it looks a little strange. 

If there's an easy way to deal with the couple of vertical points that would certainly make my eye twitch less but not worth spending much time on.

seadowg:
> If there's an easy way to deal with the couple of vertical points that would certainly make my eye twitch less but not worth spending much time on.

@lognaturel I think that would be hard without making it look off in some other way. Anything we do would technically move it out of alignment. Happy just to go with it?

lognaturel:
Did you consider stretching the stem of the mic by two points? Then yeah, that's my last suggestion and we can just go with it.

seadowg:
> Did you consider stretching the stem of the mic by two points?

@lognaturel do you mean by modifying the vector asset? I'd need to look into another tool for drawing a new asset or research how to modify the path data for that. I'm also not sure how I feel about us modifying the Material icons. 

I'm not personally seeing anything wrong with the alignment (without guidelines anyway), but it could be a screen size thing. Could you post a screenshot of a case where it looks really off to you?

lognaturel:
Let's not worry about it if there's no easy path. It's probably not an issue to people who are not me. I don't know why I notice single-pixel offsets and I wish it would stop.

The dialog margin is important, though.

mmarciniak90:
> The dialog margin is important, though.

seadowg:
@lognaturel cool I can have a look at the margin problem later today! @mmarciniak90 what OS version were you using when you saw it? I didn't notice any problems on Android 9 when I was working on it. I was using a longer test string to make sure  it ended up needing to scroll though.

mmarciniak90:
I have devices with Android 7.0, 8.1 and 10.0. On each of them, I see issue with a missing margin on dialog.

seadowg:
@mmarciniak90 ok that margin should be fixed now!

lognaturel (approved):

kkrawczyk123:
@seadowg there is one thing that I have mixed feelings about. The volume bar disappears after rotating the screen on a single question after pausing the recording. I wasn't able to record gif with rotation so I am attaching screenshots:
Paused recording on Audio widgets in All widget form in portrait:
Screenshot_2021-03-17-10-29-29-046_org odk collect android
Exactly the same view after rotating to landscape:
Screenshot_2021-03-17-10-29-45-668_org odk collect android
Both progress bars are gone. They are restored after starting recording again.

seadowg:
@kkrawczyk123 ah yeah I remember noticing this and just settling for it being consistent between the two volume indicators. We could do more work to make both stay on screen or retain their state - I don't think that feels like it's worth any time or effort though. It doesn't feel like a state many users are likely to encounter, and I'm also not sure if it really presents a problem.

mmarciniak90:
> ok that margin should be fixed now!

@seadowg thanks! Dialog looks good for me.

Finally tested with success on Android: 7.0, 8.1, 10.0

kkrawczyk123:
Verified also on Androids 5.1, 7.0 and 9.0.

PR #4459, opened 2021-03-05
Refactored shared prefernces

### LINKED ISSUE #4452: Make sure `PreferencesProvider` is always used to grab preferences
We want to use the `PreferencesProvider` everywhere so that it's easier to implement projects. This will probably involve:

* Removing the `GeneralSharedPreferences` and `AdminSharedPreferences` components (or implementing them with the `PreferencesProvider if that proves difficult)
* Replacing any raw accesses of shard preferences with ones going through the `PreferencesProvider`

grzesiek2010:
Closes #4452

I implemented one [PreferencesDataSource](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4452?expand=1#diff-9dc02f05139a5f9f4f30778d5a075bc2a13b7f56e10338f24426f62c24d99d50) it's a wrapper class around SharedPreferences which is going to work for all kinds of preferences General, Admin, Meta and any new if we need in the future. Before we had separate classes for every preference like [AdminSharedPreferences](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4452?expand=1#diff-e93432c4758d7a7e0eb44d0e7ca1732ea88f8ad4adea2a191669f569de1b9fee), [GeneralSharedPreferences](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4452?expand=1#diff-4fe891b2a5aef4ce599c567a0f5efe07244eaa16e47cdb8d755fdf4616c2331c).
Then there is one [PreferencesRepository](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4452?expand=1#diff-ae32b7b8ec2c7a87da6639a178f03935ed527b5e0c6461652aaeed5e981fc3f0) class that is responsibly for managing prefernces. In that class i added some improvements so it might be used with multiple projects and return their own preferences but this part may require some polishing depending on oure needs (we are at the beggining of implementing the feature).

I didn't implement the last point mentioned in the issue which was:
> Using the PreferencesProvider to set up the SharedPreference instances used by the PreferenceFragment implementations

beacause I think it should go in a separate pr taking into accoun how many classes I had to edit in this one.

There should be no difference and shared preferences should work as before, nevertheless I think this refactoring is riski taking into accoun t the amount of refactored classes. 

No.

No.


collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
```diff
@@ -0,0 +1,48 @@
+package org.odk.collect.android.preferences
+
+import android.content.Context
+import androidx.preference.PreferenceManager
+
+class PreferencesRepository(private val context: Context) {
```
seadowg: This still feels like it should be a `...Provider` to me. I do get that know that it takes a ` projectId` it could easily be thought of as a repository. The one thing that urks me is that it makes me think that this object is the interface for grabbing the preferences themselves. 

I'd vote for `PreferencesDataSourceProvider` personally.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+    @JvmOverloads
+    fun getMetaPreferences(projectId: String = ""): PreferencesDataSource {
+        val preferenceId = META_PREFS_NAME + projectId
+
+        if (!preferences.containsKey(preferenceId)) {
+            preferences[META_PREFS_NAME + projectId] = PreferencesDataSource(context.getSharedPreferences(preferenceId, Context.MODE_PRIVATE))
```
seadowg: I'm not sure we need this lazy loading/caching of the `PreferenceDataSource` objects here but it is a nice touch as it cuts down on any hidden object instantiation in calls to this code. Happy to leave it in.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+        val preferenceId = ADMIN_PREFS_NAME + projectId
+
+        if (!preferences.containsKey(preferenceId)) {
+            preferences[preferenceId] = PreferencesDataSource(context.getSharedPreferences(preferenceId, Context.MODE_PRIVATE), AdminKeys.getDefaults())
+        }
+        return preferences[preferenceId]!!
+    }
+
+    companion object {
+        private val preferences = mutableMapOf<String, PreferencesDataSource>()
```
seadowg: This doesn't need to be "static" really as we're using declaring this as an `@Singleton`. Could we just make this a normal field?

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+        val preferenceId = GENERAL_PREFS_NAME + projectId
+
+        if (!preferences.containsKey(preferenceId)) {
+            if (projectId.isBlank()) {
+                preferences[preferenceId] = PreferencesDataSource(PreferenceManager.getDefaultSharedPreferences(context), GeneralKeys.DEFAULTS)
+            } else {
+                preferences[preferenceId] = PreferencesDataSource(context.getSharedPreferences(preferenceId, Context.MODE_PRIVATE), GeneralKeys.DEFAULTS)
+            }
+        }
+        return preferences[preferenceId]!!
```
seadowg: We can avoid the `!!` here by flipping the order of operations:

```kotlin
val preferences = preferences[preferenceId]
return if (preferences == null) {
    newPreferences = PreferencesDataSource(context.getSharedPreferences(preferenceId, Context.MODE_PRIVATE), GeneralKeys.DEFAULTS)
    preferences[preferenceId] = newPreferences
    newPreferences
} else {
    preferences
}
```

Kotlin actually adds a [`getOrPut`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-put.html) extension function to let us to simplify that even further: 

```kotlin
return preferences.getOrPut(preferencesId) {
    PreferencesDataSource(context.getSharedPreferences(preferenceId, Context.MODE_PRIVATE), GeneralKeys.DEFAULTS)
}
```

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
```diff
@@ -0,0 +1,48 @@
+package org.odk.collect.android.preferences
+
+import android.content.Context
+import androidx.preference.PreferenceManager
+
+class PreferencesRepository(private val context: Context) {
+    @JvmOverloads
+    fun getMetaPreferences(projectId: String = ""): PreferencesDataSource {
+        val preferenceId = META_PREFS_NAME + projectId
```
seadowg: I think it'd be good to have tests for this `preferenceId` logic here. It's super simple but it feels like it would be good to cement it in now and then we know we can rely on calling it that way.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
```diff
@@ -0,0 +1,105 @@
+package org.odk.collect.android.preferences
+
+import android.content.SharedPreferences
+import java.util.Collections
+
+class PreferencesDataSource(private val sharedPreferences: SharedPreferences, private val defaultPreferences: Map<String, Any> = emptyMap()) {
```
seadowg: I think this should really be an interface, so we're able to create an InMem/Fake version in tests. Classes are `final` by default so can't be mocked/faked which is a nice nudge towards creating interfaces. Feels like `PreferencesDataSource` should be the interface and then there should be a `SharedPreferences...` implementation.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
```diff
@@ -0,0 +1,105 @@
...
+    fun contains(key: String): Boolean {
+        return sharedPreferences.contains(key)
+    }
+
+    fun getAll(): Map<String, *> {
+        return sharedPreferences.all
+    }
+
+    // Use only for General and Admin settings which have default values, otherwise it won't be possible to determine the type
+    fun get(key: String): Any? {
```
seadowg: It looks like we could replace this with `getAll` in the places where it's used externally and just use `defaultPreferences` directly in `loadDefaultPreferencesIfNotExist` (as we check for the key existing or not anyway). I think it would be nice to get rid of this method as it has some caveats (as your comment suggests) and makes the interface a little unclear (should I use this or `getX`?).

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
```diff
@@ -0,0 +1,105 @@
...
+        val defaultValue = (defaultPreferences[key] ?: 0f) as Float
+        return sharedPreferences.getFloat(key, defaultValue)
+    }
+
+    fun getStringSet(key: String): Set<String>? {
+        val defaultValue = (defaultPreferences[key] ?: emptySet<Any>()) as Set<String>
+        return sharedPreferences.getStringSet(key, defaultValue)
+    }
+
+    fun registerOnSharedPreferenceChangeListener(listener: SharedPreferences.OnSharedPreferenceChangeListener) {
```
seadowg: We probably don't want these listener methods to use any `SharedPreference` related types as it breaks the encapsulation we have. That's probably good to look at revising when we deal with the `PreferenceFragment` implementations though as that's where it's being used.

collect_app/src/main/java/org/odk/collect/utilities/TestPreferencesProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import org.odk.collect.android.application.Collect
+import org.odk.collect.android.injection.DaggerUtils
+import org.odk.collect.android.preferences.PreferencesDataSource
+import org.odk.collect.android.preferences.PreferencesRepository
+
+// Use just fo testing
+object TestPreferencesProvider {
```
seadowg, line 12: I feel like these helpers should definitely live in the `test` sources root if they're only intended to be used in tests. I'm thinking we could also just add them as static methods in `CollectHelpers` instead of adding a new object/class.

collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
```diff
@@ -70,6 +73,7 @@ public boolean isInstanceStateSaved() {
     @Override
     protected void attachBaseContext(Context base) {
         super.attachBaseContext(base);
+        DaggerUtils.getComponent(base).inject(this);
```
seadowg, line 76: Why does this move here?

seadowg (changes requested):
I haven't had a chance to go through every file but I've concentrated on the new components and structure and left some comments on some small changes I think we should consider.

Generally it's looking great though! I really like the middle ground you've found in `PreferencesDataSource` of giving us a nice wrapper but not trying to add too much fanciness on top that makes it significantly different.

collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
grzesiek2010: applyOverrideConfiguration() (-> updateConfigurationIfSupported()) which is one line below uses injected `preferencesRepository` so we had to setup dagger earlier.

seadowg: Ah ok makes sense!

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
seadowg: Actually thinking about this again I'm not sure this is the behaviour we'll want. For background jobs we'll need to pass the project ID but in most places we'll just want the no arg case to return the "current" project. As far as I can see the parameter is never passed in, so maybe we should just leave the projectId stuff out until we're actually switching between projects.

collect_app/src/main/java/org/odk/collect/utilities/TestPreferencesProvider.kt
grzesiek2010: You mean in testsshared module? Because this class is used by both instrumented and unit tests.

seadowg: Ah yeah if it is used in both places for sure `testshared` would be the right place.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
grzesiek2010: Ok PreferencesDataSourceProvider seems fine. Done.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
grzesiek2010: Right. Fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
grzesiek2010: Thanks, fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
grzesiek2010: Ok I implemented it.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
grzesiek2010: Right, fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
grzesiek2010: `getMetaPreferences()` is not the best example because in fact it should be unique and not to accept any `projectID`, I fixed it.
When it comes to general and admin preferences yes for background jobs that parameter will be used otherwise we probably should take current project id from meta preferences.

I'll add tests for this class but maybe a bit later one we decide for 100% what we need.

collect_app/src/main/java/org/odk/collect/utilities/TestPreferencesProvider.kt
grzesiek2010: Hmm but then `testshared` would need to know about classes from `collect_app` module (like `PreferencesDataSourceProvider` or `DaggerUtils`). That's not how the architecture should look like right?

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesDataSource.kt
grzesiek2010: Right, I implemented our own OnPreferenceChangeListener and now SharedPreferences.OnSharedPreferenceChangeListener is hidden.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferencesRepository.kt
grzesiek2010: I added it back because I had some failing tests... yes it's true that PreferencesDataSourceProvider and it's used like that in 99% of cases but there are 2 or 3 places where a new object is created. I should review those places maybe I can improve it.

collect_app/src/main/java/org/odk/collect/utilities/TestPreferencesProvider.kt
seadowg: Ach right. I had that wrong in my head: You're only sharing between two different test sources in `collect_app` not between multiple modules. I was wondering if anyone else had run into this and looks like we can create our own shared source set: https://stackoverflow.com/questions/47141095/organize-shared-code-between-androidtest-and-test. I gave this a little try, and it seemed to work!

seadowg (approved):
Looking good! I think it'd be cool to experiment with moving the `TestPreferencesProvider` into a new shared source set but that can easily happen as part of the next issue. I'll merge this now.

PR #4468, opened 2021-03-10
Implemented custom data stores for general and admin settings

### LINKED ISSUE #4461: PreferenceFragment implementations should use new preferences setup
The `PreferenceFragment` implementations should use the new preferences setup introduced in ##4459 rather than just the default `SharedPreferences` bucket. This most likely means exposing the names/ids for the preferences buckets from a provider.

grzesiek2010:
Closes #4461

I reviewed implemented changes, ran automated tests and played a bit with the app.

This is what's recommended in the docs https://developer.android.com/guide/topics/ui/settings/use-saved-values#custom-data-store

It doesn't require testing for now (since we decided to merge pr related to multi-project not to block us) but generally, general and admin settings should be tested.

No.

No.


collect_app/src/main/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStore.kt
```diff
@@ -0,0 +1,54 @@
+package org.odk.collect.android.preferences.stores
+
+import androidx.preference.PreferenceDataStore
+import org.odk.collect.android.preferences.PreferencesDataSourceProvider
+
+class AdminPreferencesDataStore(val preferencesDataSourceProvider: PreferencesDataSourceProvider) : PreferenceDataStore() {
```
seadowg: It feels we could just have a `PreferencesDataSourceDataStore` that takes a `PreferenceDataSource` in its constructor rather than specific implementations for admin and general. Is there a blocker for doing that?

seadowg: I'm also thinking we might want to rename our `PreferenceDataSource` interface to avoid any confusion. I guess `PreferenceDataStore` will only be used in the context of `PreferenceFragment` though.

collect_app/src/main/java/org/odk/collect/android/preferences/BaseAdminPreferencesFragment.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.inject.Inject;
+
+public abstract class BaseAdminPreferencesFragment extends BasePreferenceFragment {
+    @Inject
+    AdminPreferencesDataStore adminPreferencesDataStore;
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        getPreferenceManager().setPreferenceDataStore(adminPreferencesDataStore);
```
seadowg: If we did have one `PreferenceDataStore` implementation we could just inject the `PreferenceDataSourceProvider` here instead of the specific implementation.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    @Before
+    fun setup() {
+        preferencesDataSourceProvider = mock(PreferencesDataSourceProvider::class.java)
+        adminPreferences = mock(PreferencesDataSource::class.java)
+        `when`(preferencesDataSourceProvider.getAdminPreferences()).thenReturn(adminPreferences)
+        adminPreferencesDataStore = AdminPreferencesDataStore(preferencesDataSourceProvider)
+    }
+
+    @Test
+    fun `When putString() called, should save() be called on adminPreferences`() {
```
seadowg: I think for test names like this you can omit "when". I'd usually try and save it (or "given") for if I'm describing a condition. So for instance I'd name this test `putString saves string in adminPreferences`.  If the method exploded when you passed it a `null` key I'd write a test like `putString, when key is null, throws exception`. I'm also not including "should" here as it's not really needed. 

This is massively subjective so interested in your thoughts here! I think we have a nice opportunity to discuss different styles and then add some thoughts to our `CODE-GUIDELINES.md`.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        adminPreferencesDataStore.putString("Key", "value")
+        verify(adminPreferences, times(1)).save("Key", "value")
```
seadowg: Don't think you need the `times` in these tests.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    @Test
+    fun `When putStringSet() called, should save() be called on adminPreferences`() {
+        adminPreferencesDataStore.putStringSet("Key", setOf("x", "y", "z"))
+        verify(adminPreferences, times(1)).save("Key", setOf("x", "y", "z"))
+    }
+
+    @Test
+    fun `When getString() called, should getString() be called on adminPreferences`() {
+        adminPreferencesDataStore.getString("Key", "value")
+        verify(adminPreferences, times(1)).getString("Key")
```
seadowg: This doesn't actually verify that we are returning the right thing from `AdminPreferencesDataStore#getString`. We either use `when` to set up the return values and assert they come back or just create a fake (probably backed by a map) `PreferenceDataSource` implementation that we can add values to.  I think for simple cases like this the fake might be easier.

seadowg (changes requested):
A lot of moving around here, but it looks like a great set of changes! I've left comments on a couple of things I think we should change and on some tests I think it'd be good to discuss before moving on.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
grzesiek2010: Right, Fixed.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
grzesiek2010: I agree, Fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/BaseAdminPreferencesFragment.java
grzesiek2010: It's fixed now I inject `PreferencesDataStore`

collect_app/src/main/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStore.kt
grzesiek2010: Yeah I was thinking about this but somehow ended up with two implementation I can't recall the reason but now it's seems doable, what I did and now I have one PreferenceDataStore.

collect_app/src/test/java/org/odk/collect/android/preferences/stores/AdminPreferencesDataStoreTest.kt
grzesiek2010: Ok I improved that.

collect_app/src/test/java/org/odk/collect/android/preferences/source/PreferencesDataStoreTest.kt
```diff
@@ -12,78 +15,101 @@ class PreferencesDataStoreTest {
     @Before
     fun setup() {
         preferencesDataSource = mock(PreferencesDataSource::class.java)
+        `when`(preferencesDataSource.getString(KEY)).thenReturn(STRING_ANSWER)
```
seadowg: This is a bit nitpicky, but personally I'd try and avoid doing this kind of setup in a `Before`. I find when I'm reading a test it's nice to be able to see as much as possible of the state setup without scrolling or moving. I also find having constants set up like this harder to read than just having raw values or just local variables (all caps is probably the culprit). 

Another point is that I don't think we need to verify calls to `getString` etc as we're checking that that's happening through our state setup - we're testing the behaviour rather than the implementation.

I'd maybe suggest something like this for the `getString` read test:

```
@Test
fun `getString() reads string from preferences`() {
    `when`(preferencesDataSource.getString("key")).thenReturn("value")
    assertThat(preferencesDataStore.getString("key, "value"), `is`("value"))
}
```

Again, there are all just my opinions so lets discuss if you disagree with any of them!

collect_app/src/main/java/org/odk/collect/android/preferences/source/PreferencesDataStore.kt
```diff
@@ -0,0 +1,53 @@
+package org.odk.collect.android.preferences.source
+
+import androidx.preference.PreferenceDataStore
+
+class PreferencesDataStore(val preferencesDataSource: PreferencesDataSource) : PreferenceDataStore() {
```
seadowg: Ach my feedback has definitely pushed us towards some awkward naming here. I think this class should be named `<where data comes from>PreferenceDataStore` which would unfortunately mean calling this `PreferencesDataSourcePreferenceDataStore`. Yuck.

Maybe we should try and move away from using "preferences" for our own settings interfaces as Android seems to have done a good job of using every possible name using that word. Could we call our `PreferencesDataSource` something like `SettingsStore`, `ConfigStore`, `SettingsSource` etc?

Going to tag @lognaturel in as well because naming is hard.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/preferences/source/PreferencesDataStoreTest.kt
grzesiek2010: I tend to agree so I improved it according to what you said.

collect_app/src/main/java/org/odk/collect/android/preferences/source/PreferencesDataStore.kt
grzesiek2010: If class naming is the last thing maybe we could merge to avoid more conflicts and continue this discussion on slack?

lognaturel: I have a preference for resolving the issue in this PR if possible to try to keep history somewhat easier to navigate. Big renames can be really hard to follow. What are you seeing conflicts with, @grzesiek2010? Always helpful to make conflict concerns concrete so it's easier to evaluate! Looks like some of the renaming needs to happen in master anyway so maybe it's not a big deal in this case.

I think using the word `Settings` for Collect-specific classes is a good idea since that matches the UI. Some ideas:
`PreferencesDataSource` -> `Settings`
`defaultPreferences` -> `settingKeysToDefaults`
`loadDefaultPreferencesIfNotExist` -> `setDefaultForAllSettingsWithoutValues`
`OnPreferenceChangeListener` -> `OnSettingChangeListener`
`SharedPreferencesDataSource` -> `SharedPreferencesSettings`
`PreferencesDataSourceProvider` -> `SettingsProvider` (`getMetaSettings`, `getGeneralSettings`, etc)

seadowg: Yeah for me, it's easier to make the changes here as I'm going to have to rebase on top either way. I'm happy with the general change to "Settings".

grzesiek2010: Ok renamed those classes/methods.

seadowg (approved):

Issue #4454, opened 2021-02-25
Replace DAOs with Repositories


seadowg:
We want to replace all our accesses of the `FormsDao` and `InstancesDao` with their relevant repository. It might be that we end up with those objects still existing but private to the repository implementation.

If it becomes awkward to pass a `FormsRepository` (as an example) into a component to replace a `FormsDao` call, it might be worth just creating a `DatabaseFormsRepository` in place. The goal here is not to restructure everything but to consolidate how we are accessing forms and instances.

seadowg:
@getodk-bot hello there my little friend. Yes I'm still working on it.

![](https://media.giphy.com/media/3owzWkGtQ3us1pV0qc/giphy.gif)

PR #4469, opened 2021-03-10
Replace InstancesDao with InstancesRepository

seadowg:
Work towards #4454. There are still a few calls to that touch Instances going straight through the Content Resolver but this will be easier to clean up once we've go rid of calls to `FormsDao` as well.

I'd hope this pretty reviewable commit by commit as I tried to keep each step pretty darn small for my own sanity.

This PR is pretty rinse and repeat in its approach:

1. Find a Dao call
2. Work out if the code that it's part of is tested and try and add one if not
3. Add a new method to `InstancesRepository` if needed and then replace the dao call with one to the repo

I'll leave comments inline on any other changes I had to make.

I've tried to add tests for the core parts of everything I've changed but I of course could have missed things. I think it's worth waiting to swap out the forms calls as well before doing anything with QA but it feels like it might be worth doing something close to a regression pass (focused on forms/instances management) at that time.

For reviewers, I think it'd be good to look out for anywhere where the replacement is actually doing something different from the original dao call. Additionally, it would be good to think about if any changes have negative performance impacts.


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/QRCodePage.java
```diff
@@ -39,7 +40,13 @@ public QRCodePage clickScanFragment() {
     }
 
     public QRCodePage clickView() {
-        onView(withText(R.string.view_qr_code_fragment_title)).perform(click());
+        // Switching tabs doesn't seem to work sometimes
```
seadowg, line 43: This kept causing me problems in my test lab runs.

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
```diff
@@ -0,0 +1,153 @@
+package org.odk.collect.android.dao;
+
+import androidx.loader.content.CursorLoader;
+
+import org.odk.collect.android.application.Collect;
+import org.odk.collect.android.instances.Instance;
+import org.odk.collect.android.provider.InstanceProviderAPI;
+
+public class CursorLoaderFactory {
```
seadowg, line 9: The plan for the moment is to move all the `CursorLoader` calls here. Removing `CursorLoader` is probably going to be a job for sometime in the future so having them located in one easy to find place will hopefully make this easier.

collect_app/src/main/java/org/odk/collect/android/tasks/InstanceUploaderTask.java
```diff
@@ -60,68 +53,25 @@ protected void onPostExecute(Outcome outcome) {
...
+
+                    // The custom configuration from the third party app overrides
+                    // the app preferences set for delete after submission
+                    if (deleteInstanceAfterSubmission != null) {
+                        isFormAutoDeleteOptionEnabled = deleteInstanceAfterSubmission;
+                    } else {
+                        isFormAutoDeleteOptionEnabled = settingsProvider.getGeneralSettings().getBoolean(GeneralKeys.KEY_DELETE_AFTER_SEND);
                     }
+
+                    Stream<Instance> instancesToDelete = instanceIds.stream()
```
seadowg, line 68: This is less efficient in terms of the amount of I/O performed as it makes a call for each instance that has just been submitted. I did think about optimizing this but realized:

* in most cases the instance count will be fairly low and so not a big deal
* if the instance count is large we'd actually prefer to do a query per instance rather than load them all into memory

Given that I think this is fine. If we really want we could add a way to grab several Instances by id but I suspect we could optimize this by holding onto instance refs we supposedly load during submission.

collect_app/src/main/java/org/odk/collect/android/upload/InstanceUploader.java
```diff
@@ -49,55 +42,26 @@
...
-            selectionBuf.append(')');
-            String selection = selectionBuf.toString();
+        List<Instance> instances = new ArrayList<>();
 
-            Cursor c = dao.getInstancesCursor(selection, selectionArgs);
-            instancesToUpload.addAll(dao.getInstancesFromCursor(c));
-
-            counter++;
+        for (Long id : instanceDatabaseIds) {
+            instances.add(new DatabaseInstancesRepository().get(id));
```
seadowg, line 48: Again this could be more efficient but not sure it's a big deal given the instance count will usually be low.

collect_app/src/test/java/org/odk/collect/android/tasks/InstanceSyncTaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+@RunWith(AndroidJUnit4.class)
+public class InstanceSyncTaskTest {
+
+    @Test
+    public void whenAnInstanceFileNoLongerExists_deletesFromDatabase() {
```
seadowg, line 22: I mostly wrote this test to prove to myself what was happening.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
```diff
@@ -74,15 +74,12 @@ public static boolean isInstanceComplete(boolean end, boolean completedByDefault
...
         if (path != null) {
-            try (Cursor c = new InstancesDao().getInstancesCursorForFilePath(path)) {
-                if (c != null && c.getCount() > 0) {
-                    // should only be one...
-                    c.moveToFirst();
-                    String id = c.getString(c.getColumnIndex(InstanceColumns._ID));
-                    return Uri.withAppendedPath(InstanceColumns.CONTENT_URI, id);
-                }
+            Instance instance = new DatabaseInstancesRepository().getOneByPath(path);
+            if (instance != null && instance.getStatus().equals(Instance.STATUS_COMPLETE)) {
```
grzesiek2010: Why did you add this additional condition here (`&& instance.getStatus().equals(Instance.STATUS_COMPLETE)`)

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
grzesiek2010, line 9: Don't we need a test for this class?

collect_app/src/main/java/org/odk/collect/android/activities/InstanceChooserList.java
```diff
@@ -93,7 +93,7 @@ private void init() {
         setupAdapter();
         instanceSyncTask = new InstanceSyncTask(settingsProvider);
         instanceSyncTask.setDiskSyncListener(this);
-        instanceSyncTask.execute();
+//        instanceSyncTask.execute();
```
grzesiek2010, line 96: What's the next step here after disabling the task?

grzesiek2010 (changes requested):
Looks good!

collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
seadowg: Hmmm looks like a mistake. I feel like that should probably break something, but it doesn't seem to. I'll look into that.

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
seadowg: The code is being called from the same places but have now just moved into a new object. As far as I can recall there was no testing for the individual `CursorLoader` methods before, but they're all used as UI so should have a lot of implicit coverage from our Espresso tests 🤞. Is there any logic in particular here you felt might need a test?

collect_app/src/main/java/org/odk/collect/android/activities/InstanceChooserList.java
seadowg: There's an issue filed [here](https://github.com/getodk/collect/issues/4474). Sorry I should have mentioned that in the PR description.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
seadowg: OK I've removed the extra code. Looked like it wasn't causing problems because it was only used in places where the instance would be completed.

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
grzesiek2010: Since everything goes down to `getInstancesCursorLoader()` we could for example test that the method is called with proper parameters in all cases? but it's not a must I'm just thinking...

seadowg: Hmmmmm yeah especially for things like ordering it'd be nice to know we're calling the right thing. I do think because none of this has changed (the methods have just moved) and it's all things we want to move I wouldn't look at backfilling tests as a priority. I think it'd be hard to write tests for the specifics (like sort order) that wouldn't have to be changed if we remove cursor loaders. Does that make sense?

grzesiek2010: Yeah

grzesiek2010 (approved):

PR #4470, opened 2021-03-10
Fix discard changes dialog appearing on pressing back button in Geopoly Activity when original answer is same for GeoShape and GeoTrace widgets

### LINKED ISSUE #4179: Editing and then saving instances with shape|trace values results in invalid format
See Slack conversation: https://getodk.slack.com/archives/CB2UJJ6SF/p1603096069042400

From @TimonWeitkamp:

"All of the polygons are collected through the geoshape widget, and depending on the agricultural field they are sampling, either tap (if it is rectangle) or auto-record (if non-rectangular). None are prepopulated."

"I had a look at the csv from the manual download, and found this: there are two polygons near each other, collected by the same person on the same date, only half an hour difference and a few hundred meters from each other (see image).
The coordinates:
-25.041891420472602 33.6740592867136 0.0 0.0;-25.041891420472602 33.6740592867136 0.0 0.0;-25.041891420472602 33.6740592867136 0.0 0.0;-25.041891420472602 33.6740592867136 0.0 0.0;-25.042363462114455 33.674249053001404 0.0 0.0;-25.042071246074226 33.67517
and
-25.044823577656306 33.67414176464081 0.0 0.0; -25.044823577656306 33.67414176464081 0.0 0.0; -25.04530988431933 33.674252070486546 0.0 0.0; -25.045065364345714 33.67478650063276 0.0 0.0; -25.04461520331787 33.67461148649454 0.0 0.0; -25.044823577656306 3
The difference between these coordinates is the space after the semicolon, ";" in the first vs "; " in the second example, as mentioned above. Both polygons are tapped (they have perfectly straight lines), and no differences (except time). Could be a bug on the ODK collect side?
I found more of such examples in the csv file."

This really sounds like Collect is inconsistent about how it saves geo values. However, I can't reproduce. I see that `getDisplayText` put spaces between points but I can't see a place where that would determine the value. Maybe there's some sequence of events like going to the hierarchy or re-opening an instance that would lead us there.

@mmarciniak90 could you please work your magic and try to reproduce?

### LINKED ISSUE #4267: Discard changes dialog visible for GeoTrace and GeoShape read-only widgets

#### Problem description
When GeoTrace and GeoShape widgets have points in read-only user is able to view map but when try to leave map view using back button discard changes dialog is displayed what suggests that some changes have been made.
ezgif com-video-to-gif (44)

1. Open form with GeoTrace/GeoShape widget readonly with points
2. Open map view
3. Use back button to leave map

#### Expected behavior
Discard changes dialog should not be visible for read-only Geo widgets.

#### Other information 
Form that I have used: 
[4228AllWidgetsWithAnswersReadOnly.xml.txt](https://github.com/getodk/collect/files/5642595/4228AllWidgetsWithAnswersReadOnly.xml.txt)

### LINKED ISSUE #4451: Remove trailing semicolon from geotrace and geoshape serialized value
Currently traces and shapes always end in a trailing `;`. This is not very elegant but clients of data produced by Collect can typically easily be made robust to this quirk. However, it seems it's not practical to deal with it in Enketo: https://forum.getodk.org/t/problem-with-gps-shape-and-line-modification-in-enketo/32531/3.

`GeoPolyActivity.formatPoints` -- we can either explicitly deal with the fencepost problem or just trim the last `;` at the end.

SaumiaSinghal:
Closes #4267 and Closes #4451 and closes #4179

I tested it on Android 9, as well as on emulator running on Android 10

- I updated `formatPoint` method to remove last semi colon from the answer string and add spaces after points while concatenating.
- I used `parsePoints` to check if the points are actually modified or not, while pressing back button in `GeoPolyActivity`.

I just modified `formatPoint` method in `GeoPolyActivity`, and I have checked all possible cases for read only and non read only widgets.

form attached on the issue #4267 

No

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -282,7 +282,7 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
     }
 
     @Override public void onBackPressed() {
-        if (map != null && !formatPoints(map.getPolyPoints(featureId)).equals(originalAnswerString)) {
+        if (map != null && !parsePoints(originalAnswerString).equals(map.getPolyPoints(featureId))) {
             showBackDialog();
         } else {
             finish();
@@ -341,11 +341,13 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
         StringBuilder result = new StringBuilder();
         for (MapPoint point : points) {
             // TODO(ping): Remove excess precision when we're ready for the output to change.
-            result.append(String.format(Locale.US, "%s %s %s %s;",
+            result.append(String.format(Locale.US, "%s %s %s %s; ",
                     Double.toString(point.lat), Double.toString(point.lon),
                     Double.toString(point.alt), Float.toString((float) point.sd)));
         }
-        return result.toString().trim();
+
+        String answer = result.toString().trim();
+        return answer.isEmpty() ? answer : answer.substring(0, answer.length() - 1);
     }
 
     @Override
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -34,6 +34,7 @@ import org.odk.collect.android.geo.SettingsDialogFragment;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.preferences.MapsPreferences;
 import org.odk.collect.android.utilities.DialogUtils;
+import org.odk.collect.android.utilities.StringUtils;
 import org.odk.collect.android.utilities.ToastUtils;
 
 import java.util.ArrayList;
@@ -346,8 +347,7 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
                     Double.toString(point.alt), Float.toString((float) point.sd)));
         }
 
-        String answer = result.toString().trim();
-        return answer.isEmpty() ? answer : answer.substring(0, answer.length() - 1);
+        return StringUtils.removeEnd(result.toString().trim(), ";");
     }
 
     @Override
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -342,7 +342,7 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
         StringBuilder result = new StringBuilder();
         for (MapPoint point : points) {
             // TODO(ping): Remove excess precision when we're ready for the output to change.
-            result.append(String.format(Locale.US, "%s %s %s %s; ",
+            result.append(String.format(Locale.US, "%s %s %s %s;",
                     Double.toString(point.lat), Double.toString(point.lon),
                     Double.toString(point.alt), Float.toString((float) point.sd)));
         }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
@@ -34,12 +34,11 @@ import org.odk.collect.android.geo.SettingsDialogFragment;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.preferences.MapsPreferences;
 import org.odk.collect.android.utilities.DialogUtils;
-import org.odk.collect.android.utilities.StringUtils;
+import org.odk.collect.android.utilities.GeoUtils;
 import org.odk.collect.android.utilities.ToastUtils;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Locale;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -278,7 +277,7 @@ public class GeoPolyActivity extends BaseGeoMapActivity implements SettingsDialo
     private void finishWithResult() {
         List<MapPoint> points = map.getPolyPoints(featureId);
         setResult(RESULT_OK, new Intent().putExtra(
-            FormEntryActivity.ANSWER_KEY, formatPoints(points)));
+            FormEntryActivity.ANSWER_KEY, GeoUtils.formatPointsResultString(points, outputMode.equals(OutputMode.GEOSHAPE))));
         finish();
     }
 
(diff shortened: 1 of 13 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
```diff
@@ -341,11 +341,13 @@ private String formatPoints(List<MapPoint> points) {
...
             // TODO(ping): Remove excess precision when we're ready for the output to change.
-            result.append(String.format(Locale.US, "%s %s %s %s;",
+            result.append(String.format(Locale.US, "%s %s %s %s; ",
                     Double.toString(point.lat), Double.toString(point.lon),
                     Double.toString(point.alt), Float.toString((float) point.sd)));
         }
-        return result.toString().trim();
+
+        String answer = result.toString().trim();
+        return answer.isEmpty() ? answer : answer.substring(0, answer.length() - 1);
```
seadowg: @SaumiaSinghal 👋 heyo! What was the motivation behind this change? I'm a little worried changing the format of the answer could cause problems somewhere else?

SaumiaSinghal: Hi Callum! yeah I was concerned about this too, so I mentioned this in regression risks as well. Previously, I was trying to match the `formatPoints` output to the `answerString`, so I found that `formatPoint` method does not add up spaces when adding another point, and the last point that is added in the string should not have semicolon at the end, in order to match with the `answerString`. I found this mentioned in #4451 as well.

We use `formatPoints` while saving the answer in the form, and I checked that it is getting saved in the `Saved Forms` as we get the `answerString` in any new form that we load.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
```diff
@@ -341,11 +341,13 @@ private String formatPoints(List<MapPoint> points) {
         StringBuilder result = new StringBuilder();
         for (MapPoint point : points) {
             // TODO(ping): Remove excess precision when we're ready for the output to change.
-            result.append(String.format(Locale.US, "%s %s %s %s;",
+            result.append(String.format(Locale.US, "%s %s %s %s; ",
```
lognaturel: Please make sure your PR description is really specific and describes changes to output data in detail. There's huge risk in making these kinds of changes because they can break all downstream analysis.

I think this change comes from a misunderstanding of https://github.com/getodk/collect/issues/4179. There are NOT supposed to be spaces between points in shapes and traces. A user has reported that there are sometimes spaces but we haven't yet been able to reproduce. If you know of a place where spaces are introduced, that would be helpful. This change must be reverted.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
lognaturel: The general concept of trimming a semicolon is good. The PR should be updated to reflect that it also closes https://github.com/getodk/collect/issues/4451. I would strongly recommend using something like http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#removeEnd-java.lang.String-java.lang.String- to make it really explicit that a trailing ';' is removed.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
SaumiaSinghal: Oh alright @lognaturel, I'll revert this change. Yes that was a misunderstanding. Also, I noticed this on the form: 
[4228AllWidgetsWithAnswersReadOnly.xml.txt](https://github.com/getodk/collect/files/6143298/4228AllWidgetsWithAnswersReadOnly.xml.txt)

Screenshot_20210315_225542

More specifically, I noticed, If for the first time, I record points in `Geotrace` or `Geoshape` widget, the points appear in the form of string separated by semicolons and no spaces in between. However, if I kill and relaunch the app, there is an added space between them. Same is the case for saved forms.

lognaturel: There can be a difference between the underlying value and the display value, that's ok. It does feel to me like a display value with spaces is being saved somehow but we'll need to reproduce to fix.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPolyActivity.java
lognaturel: Some tests for the format of the underlying value would also be great! Something like `pointsInGeotraces_areSeparatedBySemicolons`, `geotraces_doNotHaveTrailingSemicolons`, `geoshapes_haveMatchingFirstAndLastPoints` or something like that. I'm surprised those aren't there already but if they were these changes would have caused test failures.

SaumiaSinghal: Alright, I'll work on the tests.

SaumiaSinghal: Hi @lognaturel! I was having some trouble as to where to add these tests, you mean for `GeoPolyActivity` test?

lognaturel: Right, the current structure does not make this easy. For now I would pull `GeoPolyActivity#formatPoints` out to `GeoUtils`. There's already `formatLocationResultString` there and I think you could rename the new static method to `formatPointsResultString(List<MapPoint> points, boolean isShape)` or something like that. You can then add tests on the static method in `GeoUtilsTest`. It's not great to introduce more static methods but I think it's better than leaving this functionality untested.

lognaturel:
I haven't reviewed in detail yet and I'd still love to see test for the underlying value but a round of QA would be helpful. Thanks for continuing to improve this, @SaumiaSinghal! 🙏

SaumiaSinghal:
Hi @lognaturel! Sorry for taking a bit long to add the tests, the tests are now in place

kkrawczyk123:
**Tested with success!**

Tested cases:
* spaces and semicolon in geotrace/geoshape widget:
   - kill app
   - change language
   - change theme
   - geotrace
   - geoshape
   - different numbers of points
   - add points
   - reduce number of points
   - hierarchy view
   - placement by tapping
   - manual mode
   - automatic mode 
   - don't keep activities turned on
   - don't keep activities turned off
   - Google, OSM and Mapbox
   - offline layer
   - edit saved form
   - minimize app
   - screen rotation
   - edit answer
   - remove response 
   - send submission to Central
   - download csv from Central
* read-only questions:
   - geoshape
   - geotrace
   - offline layer
   - Google, OSM and Mapbox 
* regression checks on geotrace and geoshape widgets

mmarciniak90:
Tested with success on Android 7.0, 8.1, 10.0

lognaturel (approved):
Thanks so much!

PR #4479, opened 2021-03-18
Implemented ProjectsRepository

### LINKED ISSUE #4478: Implement ProjectsRepository
Implement ProjectsRepository interface with on inMem or hardcoded implementation so that we can start working on the dialog (#4419) and use it as a source of data.

grzesiek2010:
Closes #4478

Those classes may evaluate for example we may not need all those fields I included in `Project` class and we may need more methods in `ProjectsRepository` but that's not a problem for know, it's enough to move forward.

It doesn't require testing.

No.

No.


seadowg:
@grzesiek2010 as we chatted about this we probably don't need hardcoded projects for the upcoming stories. I think for the moment we can just close this, so we're not adding any code before we need it. We can still copy-paste/pull it in from here if we do need it later.

PR #4481, opened 2021-03-19
New settings dialog

### LINKED ISSUE #4419: New settings dialog
This adds the new "Projects" dialog but just a way to get to settings.

**Acceptance**

- [ ] Given I'm on the Main Menu
When I click the ⚙ icon in the top right
And I click "General settings" in the dialog
Then I end up in General settings

- [ ] Given I'm on the Main Menu
When I click the ⚙ icon in the top right
And I click "Admin settings" in the dialog
Then I end up in Admin settings

grzesiek2010:
Closes #4419

I tested the changes manually and ran automated tests.

I decided to use the existed face icon temporarily to display the option.
The dialog is pretty small now what doesn't look very well but it will change so I think for now it's ok.
Screenshot_1616156937

No.

No.


collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -239,6 +242,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
 
         switch (item.getItemId()) {
+            case R.id.projects:
+                new ProjectSettingsDialog().show(getSupportFragmentManager(), ProjectSettingsDialog.TAG);
```
seadowg: I think this could use `DialogUtils.showIfNotShowing` for consistency.

collect_app/src/main/java/org/odk/collect/android/project/ProjectSettingsDialog.kt
```diff
@@ -0,0 +1,57 @@
...
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.fragment.app.DialogFragment
+import com.google.android.material.button.MaterialButton
+import org.odk.collect.android.R
+
+class ProjectSettingsDialog : DialogFragment() {
+    interface ProjectSettingsDialogListener {
```
seadowg: I don't think we need to use this listener pattern here - `ProjectsSettingDialog` can just call `requireActivity().startActivity(...` to launch general or admin settings.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
seadowg, line 2: Do you think we should start using `ConstraintLayout` for all new layouts? If I'm being honest I've been avoiding learning about it but I guess we should probably accept it's here to stay now...

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,61 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:background="@drawable/project_settings_dialog_background"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content">
+
+    <ImageView
```
seadowg, line 8: I noticed this doesn't have ripple touch feedback. I think it needs to use `?selectableItemBackground` as a background - definitely worth checking that's correct though as I might be wrong.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,61 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/projects"
+        style="@style/TextAppearance.Collect.Subtitle1"
+        app:layout_constraintBottom_toBottomOf="@+id/close_icon"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="@+id/close_icon" />
+
+    <com.google.android.material.button.MaterialButton
```
seadowg, line 28: I think we should avoid deviating from our theme on these buttons. @lognaturel I'm guessing you were just using placeholders from the default material theme in your mockups? If so, I think this can just use the `Widget.MaterialComponents.Button.OutlinedButton` style (again worth double-checking) which means we won't need `testAllCaps`, `cornerRadius`, `strokeColor`, `strokeWidth`, `backgroundTint` or `testColor`.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:background="@drawable/project_settings_dialog_background"
```
seadowg, line 4: @lognaturel how do you feel about us not using the super rounded shape from the mockups? I think we should just use the standard dialog shape here. As far as I understand shape can be customized at a theme level (https://material.io/design/shape/about-shape.html) and the extra rounded corners is probably part of Google's overall look and feel.

seadowg (changes requested):

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
lognaturel: The rounded shape is from https://material.io/components/dialogs#alert-dialog, https://www.figma.com/file/3UJK55PPAvNBdYC7pjbQ6r/%E2%9D%96-Material-Theming-%2B--Android-11-(Community)?node-id=1081%3A8808.

I don't feel strongly about it and agree we should have one style for dialogs as much as possible. If it's straightforward to make a blanket change for all dialogs, I think it's a small thing that will make the app feel more modern. But if that's unreasonable, using the current less-rounded dialog style would be fine.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
lognaturel: Right, I did just copy and paste the material theme and didn't intend for those to be prescriptive, sorry about not making that clear.

Using our standard outline buttons might be fine. My guess is that they will still look really big and out of place. I suggest either leaving it as-is or removing all the custom styles for now and refining once all the pieces are in place. I've also put a couple of other options [in the Figma file](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v1.31)?node-id=0%3A1) to consider. Thinking about it again, it may make more sense to use list items rather than buttons.

lognaturel:
This is exciting! It doesn't take much to make it feel like it's coming together. 👍

It's clear I didn't do a good job of communicating how I think about [the Figma file](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v1.31)?node-id=0%3A1) and I apologize for that. It's misleading because it looks so polished that it makes sense to think of it as an exact prescriptive spec.

My goal with it is to quickly put together something concrete that lets us reason through the interactions. I don't really know how to use Figma and am trying to not spend too much time on it so I copy and paste primarily from [this material theming template](https://www.figma.com/file/3UJK55PPAvNBdYC7pjbQ6r/%E2%9D%96-Material-Theming-%2B--Android-11-(Community)?node-id=1081%3A8808). It is FAR from pixel-perfect.

I think implementation goals should be to use Collect defaults as much as possible and to consider easier alternatives when something from the mockup is hard to implement. I also think that since you both have a good design sense, @grzesiek2010 and @seadowg, you should think a bit about what looks good to your eye. If you're thinking of something that's a big deviation, it's worth a quick chat. But if it's a quick thing, I think it's ok to put it out there and possibly do a bit of iteration during development. I also recommend continuing to refer to the patterns Google uses for their analogous dialogs. Maybe you'll come up with a slightly different interpretation than I did that makes implementation easier and that would be great.

I also expect that we'll have to do some polish when everything has come together at the end so it also feels fine to me to leave some things as placeholders that fall through to Collect defaults.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: I also didn't use it before but many times watching some tutorials I have run into something like:
`if you need nested layouts to achieve something it's more performant to use ConstraintLayout instead` so why not.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/project/ProjectSettingsDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: I removed those custom properties and used `style="@style/Widget.MaterialComponents.Button.OutlinedButton"` for now. It looks good and we can get back to it later if needed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: I think we should have a separate issue for styling all dialogs and then remove the custom styling I implemented.

seadowg: @grzesiek2010 that sounds good. Are you able to file?

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
seadowg: Yeah it definitely seems like they're concentrating on them over `RelativeLayout`. I'll try and remember to use `ConstraintLayout` instead of `RelativeLayout` but definitely tell me off if I forget!

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Sure: https://github.com/getodk/collect/issues/4482

collect_app/src/main/java/org/odk/collect/android/project/ProjectSettingsDialog.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.odk.collect.android.utilities.DialogUtils
+import javax.inject.Inject
+
+class ProjectSettingsDialog : DialogFragment() {
+
+    @Inject
+    lateinit var adminPasswordProvider: AdminPasswordProvider
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        DaggerUtils.getComponent(context).inject(this)
```
seadowg, line 30: This should really be in `onAttach(Context)` in Fragments (mentioned [here](https://developer.android.com/training/dependency-injection/dagger-android#injecting-activities)). Not a big deal but good to fix now.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,50 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="@+id/close_icon" />
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/general_settings_button"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        style="@style/Widget.MaterialComponents.Button.OutlinedButton"
+        android:layout_marginTop="@dimen/margin"
```
seadowg: This should be `margin_standard`. It looks `margin` has been declared in the color picker library and so annoyingly shows up in autocomplete. It happens to be the same right now but could change without us knowing

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -0,0 +1,50 @@
...
+        app:layout_constraintTop_toBottomOf="@+id/close_icon"/>
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/admin_settings_button"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        style="@style/Widget.MaterialComponents.Button.OutlinedButton"
+        android:layout_marginTop="@dimen/margin_small"
+        android:text="@string/admin_preferences"
+        android:textColor="@color/deep_purple_500"
```
seadowg, line 46: I thought these should really be black (`colorOnSurface`) but that also requires customization. We can come back to the color and style here when we've got more of the dialog in place.

seadowg (changes requested):
Thanks for the changes! Just a couple of tweaks that would be good to take care of now before we forget.

collect_app/src/main/java/org/odk/collect/android/project/ProjectSettingsDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Yeah let's leave it as-is now and discuss later.

seadowg (approved):
🙏

Issue #4453, opened 2021-02-25
Implement ContentProviders using repositories


seadowg:
~~Blocked by #4454~~

Right now our `ContentProvider` implementations (for forms and instances) interact directly with the database and are used by the repository implementations. We should switch this around so we don't have to use the Content URIs in our repository implementations.

## Notes

Things we need to do for this:

1. Remove any direct usage of content providers in the app (that aren't in the repos or in `CursorLoaderFactory`)
1. Write tests for the content provider behaviour (`query` and `delete` for Forms, ? for Instances)
1. Use direct database calls in repos rather than content provider
1. Replace database calls in content providers with repo calls (probably need to introduce `query` method for repos that won't work with in mem implementation)

seadowg:
![https://media.giphy.com/media/3ornk57KwDXf81rjWM/giphy.gif](https://media.giphy.com/media/3ornk57KwDXf81rjWM/giphy.gif)

@getodk-bot Hello there my little friend! Yes I'm still working on it.

PR #4483, opened 2021-03-22
Remove FormsDao

### LINKED ISSUE #4454: Replace DAOs with Repositories

seadowg:
Closes #4454

I'd advise reviewers to go through this commit by commit.

New test and verified manually.

Comments inline.


collect_app/src/main/AndroidManifest.xml
```diff
@@ -270,7 +270,7 @@ the specific language governing permissions and limitations under the License.
 
         <!-- Enable Shortcuts for Command Actions -->
         <activity
-            android:name=".activities.AndroidShortcuts"
+            android:name=".activities.AndroidShortcutsActivity"
```
seadowg, line 273: As far as  I understand it's OK to change this but let me know if that's not the case!

collect_app/src/main/java/org/odk/collect/android/activities/AndroidShortcutsActivity.java
```diff
@@ -14,44 +14,44 @@
...
 
     @Override
     public void onCreate(Bundle bundle) {
         super.onCreate(bundle);
-
-        final Intent intent = getIntent();
-        final String action = intent.getAction();
-
-        // The Android needs to know what shortcuts are available, generate the list
-        if (Intent.ACTION_CREATE_SHORTCUT.equals(action)) {
```
seadowg, line 52: I'm not sure why this filter was there seeing as that's the only reason the Activity would be opened.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/FormMapViewModel.java
```diff
@@ -43,8 +43,8 @@ public String getFormTitle() {
         return form.getDisplayName();
     }
 
-    public String getFormId() {
-        return form.getJrFormId();
+    public long getFormId() {
```
seadowg, line 46: This looks wrong in the changelist, but it's correct - other changes meant we needed a db id here not the JR form id.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
```diff
@@ -169,21 +167,9 @@ private boolean networkTypeMatchesAutoSendSetting(NetworkInfo currentNetworkInfo
...
-        try (Cursor cursor = dao.getFormsCursor()) {
-            List<Form> forms = dao.getFormsFromCursor(cursor);
-            for (Form form : forms) {
-                if (Boolean.valueOf(form.getAutoSend())) {
-                    return true;
-                }
-            }
-        }
-        return false;
+        return formsRepository.getAll().stream().anyMatch(form -> parseBoolean(form.getAutoSend()));
```
seadowg, line 172: This could be optimized but doesn't feel worth it given we generally assume the form list on device is pretty small.

collect_app/src/main/java/org/odk/collect/android/forms/Form.java
```diff
@@ -273,4 +275,12 @@ public boolean equals(Object other) {
     public int hashCode() {
         return md5Hash.hashCode();
     }
+
+    @Override
+    public String toString() {
```
seadowg, line 281: This was very useful in test failures.

collect_app/src/main/java/org/odk/collect/android/forms/MediaFileRepository.java
```diff
@@ -1,9 +0,0 @@
-package org.odk.collect.android.forms;
-
-import java.io.File;
-import java.util.List;
-
-public interface MediaFileRepository {
```
seadowg, line 6: When I got into it, it really felt like removing this and switching to absolute paths being returned from `FormRepository` was a lot simpler than what we had.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.java
lognaturel: Agreed the structure is fine. How do you decide when to do the filtering on a list of all forms vs. introducing a new repository method (e.g. `getAllWithAutoSend` or maybe even `hasAnyWithAutoSend`)? I see value to centralizing and hiding away as much of the filtering as possible. I also understand the value of keeping a tighter API so maybe that's what won out here.

seadowg: Good question. I find forcing myself to write tests for any new method is a nice way to find that balance. If it's something that's only used in one place and really driven out by a higher level test it means I probably don't need a repo method for it but if it's common then or nuanced (sorting etc) then it's nicer to have a repo method I think. That's a pretty fuzzy answer but does it make sense?

lognaturel: Yup! I'm somewhat surprised `getAllWithAutoSend` wouldn't get reused but it sounds like it was a deliberate choice and you do know you have coverage which are the things that matter.

seadowg: Yeah true. I'm honestly in two minds. We should potentially think about moving call sites to Kotlin, so we can start using default params. Then we'd just need a `getAllWith` method to cover most cases. 

For the moment I think this code is fine though. If we end up doing that operation twice I'd have no problem adding a method to the interface.

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceListCursorAdapter.java
```diff
@@ -57,26 +57,23 @@ public View getView(int position, View convertView, ViewGroup parent) {
         setUpSubtext(view);
 
         // Some form lists never contain disabled items; if so, we're done.
+        // Update: This only seems to be the case in Edit Saved Forms and it's not clear why...
         if (!shouldCheckDisabled) {
             return view;
         }
 
+        boolean formExists = true;
```
grzesiek2010: This still could be false by default the you wouldn't need that else statement adding just formExists - true below:
`isFormEncrypted = base64RSAPublicKey != null && !base64RSAPublicKey.isEmpty();`

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -186,12 +187,22 @@ public static String getCurrentFormIdentifierHash() {
...
+     * @param formId      id of a form
      * @param formVersion version of a form
      * @return md5 hash of the form title, a space, the form ID
      */
     public static String getFormIdentifierHash(String formId, String formVersion) {
-        String formIdentifier = new FormsDao().getFormTitleForFormIdAndFormVersion(formId, formVersion) + " " + formId;
+        Form form = new DatabaseFormsRepository().getLatestByFormIdAndVersion(formId, formVersion);
+
+        String formTitle;
+        if (form != null) {
```
grzesiek2010: Don't you like the ternary operator? I would use it in cases like this but maybe you don't like it?

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceListCursorAdapter.java
```diff
@@ -57,26 +57,23 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
-                    String base64RSAPublicKey = cursor.getString(base64RSAPublicKeyColumnIndex);
-                    isFormEncrypted = base64RSAPublicKey != null && !base64RSAPublicKey.isEmpty();
-                    formExists = true;
-                }
-            } finally {
-                cursor.close();
-            }
+        if (form != null) {
+            String base64RSAPublicKey = form.getBASE64RSAPublicKey();
+            isFormEncrypted = base64RSAPublicKey != null && !base64RSAPublicKey.isEmpty();
```
grzesiek2010: In some cases you check base64RSAPublicKey for null and length in other just for null. It would be good to be consistent.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -93,32 +100,21 @@
...
+    public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         DaggerUtils.getComponent(this).inject(this);
 
-        FormsDao dao = new FormsDao();
-        Form form = null;
-        List<Form> forms = dao.getFormsFromCursor(dao.getFormsCursor(getIntent().getData()));
-        if (forms.size() == 1) {
-            form = forms.get(0);
-        } else if (viewModelFactory == null) {
```
grzesiek2010, line 107: Don't we need this check anymore?

grzesiek2010 (changes requested):
Isn't it good time to remove deprecated FormInfo class?

another thing: There are still some classes where we use Collect.getInstance().getContentResolver() directly not using repositories like `SaveFormToDisk`, `EncryptionUtils` etc can't we use repositories there as well?

seadowg:
@grzesiek2010 yeah I'm actually removing `FormInfo` and the raw calls to the `ContentResolver` as part of #4453 (you'll see it's specified in there). Sorry I should have mentioned in the PR description that I just focused on getting rid of the daos here.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
seadowg: Nope that'd definitely be better here. Will change.

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceListCursorAdapter.java
seadowg: Yeah, that's probably me just copying logic from what was there before. There's no reason to check for an empty RSA key right?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
seadowg: I think for the form to not exist we'd have to have passed an incorrect ID in. Given that this Activity is only called from our app that'd be bad code rather than bad input, so I don't think we need to check for it. What do you think? 

I'm not sure I understand why the `else` is an `else if` with a check on `viewModelFactory` here.

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceListCursorAdapter.java
seadowg: Good point! I can clean that up.

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceListCursorAdapter.java
lognaturel: I would say no need to check for an empty RSA key.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
grzesiek2010: > I think for the form to not exist we'd have to have passed an incorrect ID in. Given that this Activity is only called from our app that'd be bad code rather than bad input, so I don't think we need to check for it. What do you think?

generally I agree ok let's try without that check. 

> I'm not sure I understand why the else is an else if with a check on viewModelFactory here.

It was added just to fix tests https://github.com/getodk/collect/pull/4337/commits/b396e9fc2fd378cc05fee71ddcf307fc1f2c3090

grzesiek2010 (approved):

PR #4484, opened 2021-03-23
About page is accessible from settings/projects dialog

### LINKED ISSUE #4460: About page is accessible from settings/projects dialog
~~Blocked by #4419~~

**Acceptance**

- [ ] Given I'm on the main menu
When I click the icon on top right to open the settings/projects dialog
And I click on About
Then I land on Collect's About menu

**Notes**

The "About" menu item should be below "Add projects" in the section below the project list.

grzesiek2010:
Closes #4460

Screenshot_1616424952

It's the only solution.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MainMenuTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MainMenuTest.java
@@ -36,8 +36,7 @@ public class MainMenuTest {
                 .clickOnString(R.string.qr_code)
                 .pressBack(new AdminSettingsPage(rule))
                 .pressBack(new MainMenuPage(rule))
-                .clickOnMenu()
-                .assertTextDoesNotExist(R.string.configure_via_qr_code);
+                .assertOverflowMenuDoesNotExist();
     }
 
 }
--- a/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
+++ b/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
@@ -9,6 +9,8 @@
         android:id="@+id/close_icon"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/margin_standard"
+        android:layout_marginTop="@dimen/margin_standard"
         android:src="@drawable/ic_close"
         android:background="?attr/selectableItemBackgroundBorderless"
         app:layout_constraintStart_toStartOf="parent"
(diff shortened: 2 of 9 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MainMenuTest.java
```diff
@@ -36,8 +36,7 @@ public void configureWithQRCanBeDisabled() {
                 .clickOnString(R.string.qr_code)
                 .pressBack(new AdminSettingsPage(rule))
                 .pressBack(new MainMenuPage(rule))
-                .clickOnMenu()
-                .assertTextDoesNotExist(R.string.configure_via_qr_code);
+                .assertOverflowMenuDoesNotExist();
```
seadowg, line 39: Actually I think we can just kill the overflow menu now right? We don't have an issue for it and would be nice to clean up while we're here.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -47,4 +50,28 @@
...
+        android:layout_marginStart="@dimen/margin_standard"
+        android:layout_marginTop="@dimen/margin_standard"
+        android:layout_marginBottom="@dimen/margin_standard"
+        android:background="?attr/selectableItemBackgroundBorderless"
+        android:gravity="center_vertical"
+        android:text="@string/about_preferences"
+        app:icon="@drawable/ic_outline_info_24"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/divider"/>
```
seadowg, line 76: It feels like so much work to create a basic linear flow with `ConstraintLayout`. It does seem like they don't want us using `LinearLayout` though: https://developer.android.com/guide/topics/ui/layout/linear. There seems to be some new "Flow" component but it's hard to find good docs on using it.

We can have a play this when we chat later.

seadowg (approved):
Looking good! I think it's worth removing the overflow menu while we're here, but no changes needed if you think that should be a separate PR.

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/MainMenuTest.java
grzesiek2010: We need it to keep `Configure via QR code` option. In this test the option is disabled in admin settings so that's why I assert it's not visible at all. I could remove the three dot's menu and display an icon for that option but it would be removed later either way so it doesn't make sense.

seadowg: I was thinking we'd just get rid of that feature now seeing as `Configure via QR code` is still available via "Admin Settings". As you say though the test is really about the ability to disable the feature so, we'd also need to remove that config. Let's merge this and discuss.

PR #4491, opened 2021-03-26
User can add project from settings dialog

### LINKED ISSUE #4420: User can add project from settings dialog
~~Blocked by #4419~~

**Acceptance**

- [x] Given I'm on the Setting dialog
When I click "Add project"
And I enter a project name
And I then I see the project listed in the Settings dialog

grzesiek2010:
Closes #4420

Screenshot_1616693539

I tested the change manually and ran automated tests.

The only thing that might need discussion is the fact that I put both `Add project` and `About` buttons in one row. We didn't have mockups for both buttons so I assumed I can make that decision by myself. It's also similar to what google dialogs look like:
image

 
No.

No.


seadowg (changes requested):
As for any feature, I think this needs tests. Generally for a new feature like this I'd try and write a UI/app level test (Espresso in our case) based on the first piece of acceptance I'm going to deal with. For this, I'd imagine we could have a test that creates a project via the 
"Add project" button and then checks it appears in the settings dialog.

I'm a big of the fact that you've added a `ProjectsRepository` with an `InMem...` implementation. I really like following a "persistence last" strategy like that to prevent us making any decisions around persistence until we need to. However I think to employ that approach we really need contract tests for the `ProjectsRepository` that check our implementation is correct. Otherwise, if we do want to move to a `SharedPreferences..` or `Database...` implementation it'll be hard to verify that our new and old implementation share the same behaviour. The `FormsRepositoryTest` (and the corresponding `Database...` and `InMem...` implementation tests) are a good example of that approach. There are a few frameworks for contract testing but in general with JUnit the simplest thing to do is have an abstract test class for the "contract" that declare a method to grab the subject implementation:

```kotlin
abstract class ProjectsRepositoryTest {

    abstract val subject: ProjectsRepository

    @Test
    fun testSomething() {
        // Test something about the repo
    }
}

class InMemProjectsRepositoryTest : ProjectsRepositoryTest {

    override val subject = InMemoryProjectsRepositoryTest()
}
```

And just to be clear I wouldn't usually think we'd need a unit/contract test for a simple feature like this (the UI test would probably do) - it's the likelihood we'll want a second implementation of `ProjectsRepository` that I think makes that important.

As always happy to discuss any of these approaches or if you don't think it makes sense!

grzesiek2010:
Right. I added those test.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import androidx.test.rule.ActivityTestRule
+import org.junit.Rule
+import org.junit.Test
+import org.odk.collect.android.activities.MainMenuActivity
+import org.odk.collect.android.support.pages.MainMenuPage
+
+class AddNewProjectTest {
+
+    @get:Rule var rule = ActivityTestRule(MainMenuActivity::class.java)
```
seadowg: Nice! Really cool to see these in Kotlin.

This should use `TestRuleChain` so we get all our state reset and idling resources:

```
public final CollectTestRule rule = new CollectTestRule();

    @Rule
    public final RuleChain chain = TestRuleChain.chain().around(rule);
```

Then you can just do `rule.mainMenu()` at the start of the test (rather than `MainMenuPage`).

seadowg (changes requested):
It looks like there is a problem on some OS levels (seeing this on API 28 and 24):

Screenshot_20210330_114215 

I've left another comment but haven't gone through everything yet. Thought I'd get that to you ASAP though!

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/MainMenuPage.java
```diff
@@ -51,6 +51,21 @@ public MainMenuPage openProjectSettingsDialog() {
...
+        onView(withId(R.id.add_project_button)).perform(click());
+        return this;
+    }
+
+    public MainMenuPage inputProjectName(String projectName) {
+        inputText(R.string.project_name, projectName);
+        return this;
+    }
+
+    public MainMenuPage addProject() {
```
seadowg: I was initially going to suggest that adding a project should return to the settings dialog, but actually I guess it does makes sense to return to the Main Menu for folks who want to import a project and then get going (downloading forms etc).

Either way I think the settings and add project dialogs should have their own `Page` objects and we should have `assertOnPage` calls in the methods that carry out the transitions between pages.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.hamcrest.Matchers.`is`
+import org.junit.Before
+import org.junit.Test
+
+abstract class ProjectsRepositoryTest {
+    lateinit var projectsRepository: ProjectsRepository
+
+    abstract fun buildSubject(): ProjectsRepository
+
+    @Before fun setup() {
```
seadowg: I think the `@Test` needs to go above the methods or hitting reformat in Android Studio will change the file.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
```diff
@@ -0,0 +1,25 @@
+package org.odk.collect.android.projects
+
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.`is`
+import org.junit.Before
+import org.junit.Test
+
+abstract class ProjectsRepositoryTest {
```
seadowg, line 10: Given we have implementation for them, it would be good to have tests for `get` and `delete`. We also want to check that each Project gets its own ID. IMO, this doesn't need to do anything to mathematically validate uniqueness. There's a lost example in `InstancesRepositoryTest#save_addsUniqueId`. 

My tip for (especially but not just) these kinds of objects and these kinds of tests is to use strict TDD and not write any code in the object until there is a failing test. That help keeps the coverage really tight and means there is documentation for every piece of behaviour we expect from the interface. Most importantly, I find I have an easier time working out what tests to write. I completely understand that TDD can be a weird way to work for a lot of people though so happy to pair on it or run through examples.

strings/src/main/res/values/strings.xml
```diff
@@ -982,4 +982,7 @@
     <string name="background_audio_help">This form records the audio from your device\'s microphone.\n\nYou can use the volume indicator to make sure the microphone is close enough to the sound you need to record.\n\nTo stop the recording please exit this form.\n\nFor more information please contact the person who asked you to collect data.</string>
 
     <string name="projects">Projects</string>
+    <string name="add_project">Add project</string>
```
seadowg, line 990: We should add comments for these strings really. Maybe we should add a "projects" group? @lognaturel did that for a bunch of other areas of the app.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -1,27 +1,40 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    android:background="@drawable/project_settings_dialog_background"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content">
+    android:layout_height="wrap_content"
+    android:background="@drawable/project_settings_dialog_background">
+
+    <androidx.constraintlayout.widget.Guideline
```
seadowg, line 8: Ooo this is really nice. Good move!

grzesiek2010:
I forgot to add background color. Both things should be fixed.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
```diff
@@ -0,0 +1,63 @@
...
+import android.view.View
+import android.view.ViewGroup
+import androidx.appcompat.widget.Toolbar
+import org.odk.collect.android.R
+import org.odk.collect.android.databinding.AddProjectDialogLayoutBinding
+import org.odk.collect.android.injection.DaggerUtils
+import org.odk.collect.material.MaterialFullScreenDialogFragment
+import javax.inject.Inject
+
+class AddProjectDialog : MaterialFullScreenDialogFragment() {
```
seadowg, line 15: I think we need tests to check that clicking "Cancel" or pressing back dismiss the dialog. As far as I can tell we should be able to do that with the `FragmentScenario` tests (`AudioRecordingFormErrorDialogFragmentTest` is a recent example).

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
grzesiek2010: You mean @Before? like:
```
@Before 
fun setup() {
...
}
```

seadowg: Yeah, exactly. If you do "Reformat Code" in Android Studio is that not what you automatically get?

grzesiek2010: Rights, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/MainMenuPage.java
grzesiek2010: I wanted to add it later once we have more methods that e could move but ok it can be a part of this pr as well.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
grzesiek2010: I added more tests.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
grzesiek2010: Ok, I added those tests.

strings/src/main/res/values/strings.xml
grzesiek2010: I added a group comment like in other cases:
```
<!--
 ##############################################
 # Projects
 ##############################################
-->
```

collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
```diff
@@ -0,0 +1,39 @@
...
+            assertThat(dialog?.isShowing, `is`(false))
+        }
+    }
+
+    @Test
+    fun clickingDeviceBackButton_dismissesDialog() {
+        val scenario = RobolectricHelpers.launchDialogFragment(AddProjectDialog::class.java)
+        scenario.onFragment { f: AddProjectDialog ->
+            val dialog = f.dialog
+            Espresso.onView(ViewMatchers.isRoot()).perform(ViewActions.pressBack())
```
seadowg, line 35: You can just use `f.onBackPressed()` here. I'm not sure what I prefer as it's hard to get a sense how seriously they want Robolectric and Espresso tests to look the same. This is totally fine though!

collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.odk.collect.android.support.RobolectricHelpers
+
+@RunWith(AndroidJUnit4::class)
+class AddProjectDialogTest {
+
+    @Test
+    fun clickingCancel_dismissesDialog() {
+        val scenario = RobolectricHelpers.launchDialogFragment(AddProjectDialog::class.java)
+        scenario.onFragment { f: AddProjectDialog ->
+            val dialog = f.dialog
```
seadowg, line 22: I'd be tempted to do `f.dialog!!` here to catch any problems early (same suggestion for the other test). I feel like I end up using `!!` in tests a lot but completely try to avoid it in non-test code.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    @Test
+    fun getAll_shouldReturnAllProjectsFromStorage() {
+        projectsRepository.add("ProjectX")
+        projectsRepository.add("ProjectY")
+        projectsRepository.add("ProjectZ")
+
+        val projects = projectsRepository.getAll()
+
+        assertThat(projects.size, `is`(3))
+        assertThat(projects[0].name, `is`("ProjectX"))
```
seadowg, line 29: We're implicitly testing order here. Not a problem right now but in the future we might just want to use the `contains` matcher if we don't care about ordering on the `getAll` results.

seadowg (approved):
I've left a few thoughts inline but nothing that needs to change!

PR #4495, opened 2021-03-29
Implement FormsProvider with FormsRepository

seadowg:
Work towards #4453 

Still to do here:

- [x] Need a test for "Hide old versions" feature
- [x] Write comments for/rename `FormsCountRepository`
- [x] How should we manage the DB helper/connection? (https://stackoverflow.com/questions/6608498/best-place-to-close-database-connection)
- [x] Need a test for deleting itemsets (only tested at ContentProvider level)
- [x] Check for unclosed `Cursor` objects
- [x] Where should `FormColumns` live?

Verified manually and added a bunch of tests.

As detailed in the issue I followed these steps for the `FormsProvider` and `FormsRepository`:

1. Remove any direct usage of content providers in the app (that aren't in the repos or in `CursorLoaderFactory`)
1. Write tests for the content provider behaviour (`query` and `delete` for Forms, ? for Instances)
1. Use direct database calls in repos rather than content provider
1. Replace database calls in content providers with repo calls (probably need to introduce `query` method for repos that won't work with in mem implementation)

I'll leave comments inline on anything that seems strange or should be discussed.

Should hopefully not change anything other than maybe some obscure usages of the `FormsProvider` (we've already started discussing those). We'll look at a full QA pass once we've done the same thing for Instances.


collect_app/src/main/java/org/odk/collect/android/formmanagement/InstancesCountRepository.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@Singleton
+public class InstancesCountRepository {
+
+    private final MutableLiveData<Integer> finalized = new MutableLiveData<>(0);
+    private final MutableLiveData<Integer> unsent = new MutableLiveData<>(0);
+    private final MutableLiveData<Integer> sent = new MutableLiveData<>(0);
+
+    public LiveData<Integer> getUnsent() {
+        update();
```
lognaturel: TODO H: every getter updates? This just gets the initial livedata for observation so maybe it's the right thing?

collect_app/src/main/java/org/odk/collect/android/forms/FormUtils.java
```diff
@@ -65,4 +87,86 @@ public static void addSessionRootTranslators(ReferenceManager referenceManager,
...
     }
+
+    public static ContentValues getValuesFromForm(Form form, StoragePathProvider storagePathProvider) {
+        ContentValues values = new ContentValues();
+        values.put(_ID, form.getId());
+        values.put(FormsProviderAPI.FormsColumns.DISPLAY_NAME, form.getDisplayName());
+        values.put(FormsProviderAPI.FormsColumns.DESCRIPTION, form.getDescription());
+        values.put(FormsProviderAPI.FormsColumns.JR_FORM_ID, form.getJrFormId());
+        values.put(FormsProviderAPI.FormsColumns.JR_VERSION, form.getJrVersion());
+        values.put(FormsProviderAPI.FormsColumns.FORM_FILE_PATH, storagePathProvider.getRelativeFormPath(form.getFormFilePath()));
```
lognaturel: H TODO: Does this mean repository doesn't handle being given an absolute path? Should it? Feels scary to still have to keep track of where there are relative and absolute paths. Ideally either could be passed in and the repository would sort it out.

collect_app/src/main/java/org/odk/collect/android/forms/FormUtils.java
```diff
@@ -65,4 +87,86 @@ public static void addSessionRootTranslators(ReferenceManager referenceManager,
...
+    }
+
+    public static Form getFormFromValues(ContentValues values, StoragePathProvider storagePathProvider) {
+        return new Form.Builder()
+                .id(values.getAsLong(_ID))
+                .displayName(values.getAsString(DISPLAY_NAME))
+                .description(values.getAsString(DESCRIPTION))
+                .jrFormId(values.getAsString(JR_FORM_ID))
+                .jrVersion(values.getAsString(JR_VERSION))
+                .formFilePath(storagePathProvider.getAbsoluteFormFilePath(values.getAsString(FORM_FILE_PATH)))
```
lognaturel, line 120: H TODO: should method deal with absolute vs. relative path?

collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
```diff
@@ -294,15 +289,8 @@ public static EncryptedFormInformation getEncryptedFormInformation(Uri uri, Inst
...
-        } else if (FormsColumns.CONTENT_ITEM_TYPE.equals(cr.getType(uri))) {
-            Cursor cursor = cr.query(uri, null, null, null, null);
-            if (cursor.getCount() != 1) {
-                String msg = TranslationHandler.getString(Collect.getInstance(), R.string.not_exactly_one_blank_form_for_this_form_id);
-                Timber.d(msg);
-                throw new EncryptionException(msg, null);
-            }
-
-            form = DatabaseFormsRepository.getFormsFromCursor(cursor, new StoragePathProvider()).get(0);
+        } else if (FormsColumns.CONTENT_ITEM_TYPE.equals(Collect.getInstance().getContentResolver().getType(uri))) {
```
lognaturel: Was this just not used (anymore)?

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
```diff
@@ -0,0 +1,478 @@
...
+                assertThat(cursor.getString(cursor.getColumnIndex(LANGUAGE)), is("English"));
+            } else {
+                assertThat(cursor.isNull(cursor.getColumnIndex(LANGUAGE)), is(true));
+            }
+        }
+    }
+
+    @Test
+    @Ignore("Not sure we need this")
+    public void update_withFormFilePath_deletesFiles_andUpdatesHashAndPaths() {
```
lognaturel: You wrote/ran these tests against the old code and they passed, right? 

For this one, I guess it's reasonably kind cleanup but the stakes feel low when it comes to having orphaned form definition files. Am I understanding the implications of not having this behavior correctly?

I hadn't thought about this before but when we used to store absolute file paths, an external app could have written files to some random place and had Collect use those, right? Now that we store relative paths, this is no longer possible?

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
```diff
@@ -0,0 +1,478 @@
...
+            assertThat(cursor.getString(cursor.getColumnIndex(MD5_HASH)), is(newHash));
+            assertThat(cursor.getString(cursor.getColumnIndex(FORM_FILE_PATH)), is(newFile.getAbsolutePath())); // This is broken - should be relative
+            assertThat(cursor.getString(cursor.getColumnIndex(JRCACHE_FILE_PATH)), is(newHash + ".formdef"));
+            assertThat(cursor.getString(cursor.getColumnIndex(FORM_MEDIA_PATH)), is(oldMediaPath)); // This is broken - hasn't updated
+        }
+    }
+
+    @Test
+    @Ignore("Not sure we need this")
+    public void update_withCachePath_deletesCache() {
```
lognaturel: This might have been a way for external apps to force cache regeneration? Doesn't seem necessary.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
```diff
@@ -0,0 +1,478 @@
...
+
+        assertThat(new File(getFormsDirPath() + oldFormPath).exists(), is(false));
+        assertThat(new File(externalFilesDir + File.separator + ".cache" + File.separator + oldCachePath).exists(), is(false));
+
+        try (Cursor cursor = contentResolver.query(formUri, null, null, null)) {
+            assertThat(cursor.getCount(), is(1));
+
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(MD5_HASH)), is(newHash));
+            assertThat(cursor.getString(cursor.getColumnIndex(FORM_FILE_PATH)), is(newFile.getAbsolutePath())); // This is broken - should be relative
```
lognaturel: Except it only really makes sense to hand Collect a filepath in a universe with absolute paths, I think.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
```diff
@@ -0,0 +1,478 @@
...
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(JRCACHE_FILE_PATH)), is("newCache"));
+        }
+    }
+
+    /**
+     * Theory: this is attempting to prevent an edge case where the file path is changed, a new cache path
+     * is generated and then that new cache path is deleted instead of the old one. This was based on
+     * comments in the code that indicated ordering was important.
+     */
```
lognaturel: Agreed. Does not seem like a useful case. Seems that the implications of not deleting the files are just that there are possibly orphaned files which doesn't seem like a big deal.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
```diff
@@ -0,0 +1,478 @@
...
+            assertThat(cursor.getCount(), is(1));
+
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(JRCACHE_FILE_PATH)), is(newHash + ".formdef"));
+        }
+    }
+
+    @Test
+    @Ignore("Not sure we need this")
+    public void update_withSelection_andFormFilePath_deletesFiles_andUpdatesHashAndPaths() {
```
lognaturel: It does feel important that an update to form file path would update the md5 hash.

As I've alluded to above, an important question is whether there's any value to passing in a form path now that only paths relative to Collect's storage can be used. Could apps write relative paths that reach their own dirs?

lognaturel: None of these deal with "fast external itemset" data, right?

lognaturel (commented):
Haven't had as much of a chance to think about possible missing tests as I'd like. What's there looks good and I've left reactions to the questionable tests.

I left some comments with H TODO that I intend to come back to. Feel free to respond/react if you want or just let me think about it some more and see if I really have a question/suggestion.

collect_app/src/main/java/org/odk/collect/android/formmanagement/InstancesCountRepository.java
seadowg: Yeah I don't think it's right. Ideally `update` would get called when the object is created and then by anything that updates the instances count. Right now `update` is called from `MainMenuActivity#onResume` via the `MainMenuViewModel` so I think I'll change this so `update` is never called by `InstancesCountRepository` as it really doesn't need to be.

Edit: Also there's no test (even UI test) that drives out those calls to `update` so I'm killing them out of principle.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
grzesiek2010: I agree, it doesn't seem important.

collect_app/src/main/java/org/odk/collect/android/forms/FormUtils.java
seadowg: It only stores relative paths in the DB and returns `Form` objects with absolute paths. However, you could save an object with an absolute or a relative path it'd still work because I think because of how `StoragePathProvider#getRelativeFormPath` works. 

It seems like we now deal in absolute paths everywhere in our code however (as everything is going through the repository) so we shouldn't really have to think about any longer outside of `DatabaseFormsRepository`.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
grzesiek2010: :+1:

collect_app/src/main/java/org/odk/collect/android/forms/FormUtils.java
seadowg: As in the other comment, it supports both but always returns a form with an absolute path. I guess the way to think about it is:

* `ContentValues` and `Cursor` objects always use relative paths for files as they represent data to be saved to a DB or data retrieved from one.
* `Form` always uses absolute paths for files as it is in mem/temporary so can safely use them (without us worrying about changing where the form lives in a later version of the app).

collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java
seadowg: Yeah as far as I could see this helper would never be passed a form URI. That has definitely been a piece of complexity that's held me up throughout this work: A `URI` can point at different data types and multiple parts of the code actually use that (see `SaveFormTask` that has a `uri` field that changes from pointing to a Form to pointing to an Instance).

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
seadowg: Yeah I wrote this test file in a branch off of `v1.30.1` (and everything passed) and then rebased it onto my `content-providers` branch so I definitely had a set of tests for the existing behaviour.

> For this one, I guess it's reasonably kind cleanup but the stakes feel low when it comes to having orphaned form definition files. Am I understanding the implications of not having this behavior correctly?

That's my understanding of it.

> I hadn't thought about this before but when we used to store absolute file paths, an external app could have written files to some random place and had Collect use those, right? Now that we store relative paths, this is no longer possible?

Yeah probably right? Although since the storage migration feature went in, and I'd be surprised if something didn't break given there was then a lot of assumptions around relative paths in the app.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
seadowg: Haha that's pretty unanimous. I'll go ahead and delete this one then.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
seadowg: I think the problem here is that if you update the form file path you end up with an absolute path stored in the DB (and returned to you from `query`) but when you inserted that form you would have ended up with a relative path stored in the DB (and again returned from `query`). As far as I can tell our code never actually updates the path like this though.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
seadowg: Cool! Will nuke this test as well.

collect_app/src/test/java/org/odk/collect/android/provider/FormsProviderTest.java
seadowg: > It does feel important that an update to form file path would update the md5 hash

Yeah agreed. There's actually a test for that in the `FormsRepositoryTest` so that is handled.

> As I've alluded to above, an important question is whether there's any value to passing in a form path now that only paths relative to Collect's storage can be used. Could apps write relative paths that reach their own dirs?

Maybe? I guess because of how `StoragePathProvider#getRelativePath` works if an external app gives us paths in their own dir they would be stored as absolute (as they don't start with our forms dir). I think Collect would break if it tries to access these forms though as it would prepend our forms dir path. If we want to support that kind of behaviour it would be pretty easy to write a test for it (just insert a form that lives somewhere else on the device and check Collect can open it for instance.

> None of these deal with "fast external itemset" data, right?

Nope that only got deleted as part of `delete`. Remember these are tests for the existing functionality so I haven't tried to alter or fix any behaviour.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFormColumns.java
```diff
@@ -0,0 +1,38 @@
+package org.odk.collect.android.database;
+
+import android.provider.BaseColumns;
+
+/**
+ * Columns for the Forms table.
+ */
+public final class DatabaseFormColumns implements BaseColumns {
```
seadowg: It would have been nice to also have a `FormsProviderAPI.FormsColumns` but it doesn't seem worth it unless we want to rename a column in the DB (but still return the same thing from the content provider).

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormDeleter.java
```diff
@@ -29,5 +33,7 @@ public void delete(Long id) {
         } else {
             formsRepository.softDelete(form.getId());
         }
+
+        fastExternalItemsetsRepository.deleteAllByCsvPath(form.getFormMediaPath() + File.separator + "itemsets.csv");
```
seadowg, line 37: It felt cleaner to avoid having the `FormsRepository` care about itemsets and have the clean up logic here.

collect_app/src/main/java/org/odk/collect/android/tasks/InstanceSyncTask.java
```diff
@@ -119,7 +119,7 @@ protected String doInBackground(Void... params) {
                 List<Instance> instances = new DatabaseInstancesRepository().getAllNotDeleted();
 
                 for (Instance instance : instances) {
-                    String instanceFilename = storagePathProvider.getAbsoluteInstanceFilePath(instance.getInstanceFilePath());
+                    String instanceFilename = instance.getInstanceFilePath();
```
grzesiek2010, line 122: Now in Instance we have two methods:
- getInstanceFilePath()
- getAbsoluteInstanceFilePath()
and basically both do the same but the second one always calls `new StoragePathProvider().getAbsoluteInstanceFilePath(instanceFilePath)` what's now redundant as I understand.
Please remove one of them. The one that you keep should be named `getAbsoluteInstanceFilePath` or `getInstanceFilePath` but then it would be good to have a comment to make it clear it returns absolute paths.

You could also rename the setter and the field itself `instanceFilePath` to make it clear that it's an absolute path.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -590,22 +590,22 @@ private void loadFromIntent(Intent intent) {
...
+            instancePath = instance.getInstanceFilePath();
+            List<Form> candidateForms = formsRepository.getAllByFormIdAndVersion(instance.getJrFormId(), instance.getJrVersion());
 
             if (candidateForms.isEmpty()) {
                 createErrorDialog(getString(
                         R.string.parent_form_not_present,
-                        formInfo.getFormId())
-                                + ((formInfo.getFormVersion() == null) ? ""
-                                : "\n" + getString(R.string.version) + " " + formInfo.getFormVersion()),
+                        instance.getJrFormId())
```
grzesiek2010: I know that it's not new but this is something I've never liked... why do we use that `Jr` in case of ids or version numbers? Could we get rid of it?

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFastExternalItemsetsRepository.kt
```diff
@@ -0,0 +1,13 @@
+package org.odk.collect.android.database
+
+import org.odk.collect.android.fastexternalitemset.ItemsetDbAdapter
+import org.odk.collect.android.itemsets.FastExternalItemsetsRepository
+
+class DatabaseFastExternalItemsetsRepository : FastExternalItemsetsRepository {
```
grzesiek2010, line 6: Why do you sometimes creating new classes use Kotlin like here or `FastExternalItemsetsRepository`
but in other cases it's java:
- DatabaseFormColumns
- FormsDatabaseProvider
- InstancesCountRepository
- ContentUriHelper
- plus some tests

I guess that the reason is that some of them are just factored out from java classes and you didn't want to convert them to kotlin but it's not always the case.

collect_app/src/main/java/org/odk/collect/android/utilities/ContentUriHelper.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.odk.collect.android.application.Collect;
+
+public final class ContentUriHelper {
+
+    private ContentUriHelper() {
+
+    }
+
+    public static Long getIdFromUri(Uri contentUri) {
```
grzesiek2010: There are probably other cases where this method could be used like `InstanceProvider` where we call:
`uri.getPathSegments().get(1))` in a few places. Since now we have this helper method probably it should be used everywhere.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -265,269 +169,102 @@ private void deleteFileOrDir(String fileName) {
...
-            }
+                    count = cursor.getCount();
+                }
+                break;
 
-            getContext().getContentResolver().notifyChange(uri, null);
-            getContext().getContentResolver().notifyChange(FormsColumns.CONTENT_NEWEST_FORMS_BY_FORMID_URI, null);
-        }
+            case FORM_ID:
+                Form form = formsRepository.get(ContentUriHelper.getIdFromUri(uri));
```
grzesiek2010, line 225: You may need a null check below?

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -265,269 +169,102 @@ private void deleteFileOrDir(String fileName) {
...
-                        }
-                    } finally {
-                        if (c != null) {
-                            c.close();
-                        }
+        deferDaggerInit();
+
+        int count;
+
+        FormDeleter formDeleter = new FormDeleter(formsRepository, instancesRepository, fastExternalItemsetsRepository);
```
grzesiek2010, line 176: Shouldn't we inject this like other dependencies?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/tasks/InstanceSyncTask.java
seadowg: Ah good point. I removed `StoragePathProvider` as a dependency from `Form` at some point but haven't done it for `Instance`. I'll take care of it.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: That's a good question. I'm guessing it stands for "Java Rosa"? Really it feels like it should be "Open Rosa" right? Maybe (for `Form` and `Instance`) we should swap where the specifier is and have `id` (String), `version` (String) and `dbId` (Long)? I'm interested to tag @lognaturel in here as she might have some context around the current naming.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseFastExternalItemsetsRepository.kt
seadowg: > I guess that the reason is that some of them are just factored out from java classes and you didn't want to convert them to kotlin but it's not always the case.

Yeah that's mostly what was happening, but really it was just that I wasn't really thinking about converting to Kotlin (just not in the habit yet). I'm happy to convert the 4 classes you referenced, and then I'll try and get in the habit of all new classes being in Kotlin.

collect_app/src/main/java/org/odk/collect/android/utilities/ContentUriHelper.java
seadowg: Good shout will look at that.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
seadowg: Hmmm good question. I'll look at what the previous behaviour was (if a non-existent ID was passed in) and write a test to make sure this implementation is consistent.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
seadowg: This is a question I'm always asking myself and I only really have fuzzy answers. There's always a trade-off around what "level" of dependency is inverted (and then injected) especially when you're using a DI framework as there is some overhead in adding or changing a dependency (changing the constructor for instance). 

Generally I try and have my tests tell me what should be explicitly exposed as a dependency and what should be constructed from other dependencies. Here we test the `FormsProvider` behaviour as an "opaque box" - we just test its behaviour rather than it's interactions with other components, so I never end up writing a test that wants to check that `FormDeleter` is called (which would force me to make it an explicit dependency). This means there's no test that guarantees I'm using `FormDeleter`, although the part of `FormDeleter`'s behaviour we want is tested (deleting itemsets). 

All that is mostly out loud thoughts rather than an answer or a solid opinion, but I guess I would say that I wouldn't bother making it an explicit dependency until it would be easier to write a test if it was one. Does that make sense or is just rambling 😅?

collect_app/src/main/java/org/odk/collect/android/utilities/ContentUriHelper.java
seadowg: Actually it looks like the only uses of `getPathSegments` outside this are in `InstancesProvider`. I'll change those in the next PR after I've added test coverage.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
lognaturel: I like using `formId` because that matches how it's referred to in documentation and other parts of the system. I have made that change when I've touched related code. I do prefer `dbId` to just `id` as well.

lognaturel:
Whoops, realize I didn't come back here as I thought I had. Let's remove all of the questionable update and delete tests for now. I doubt we'll bring any of that functionality back. Even if we do, it will be easier to deal with it separately.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: Great. I'll make that change while I'm here.

grzesiek2010: We still have columns in forms databse like:
const val JR_FORM_ID = "jrFormId"
const val JR_VERSION = "jrVersion"

should we rename them too? Probably not because external apps may relay on the existing naming?

seadowg: Good point. Yeah I was assuming we'd just leave those in place for now. Our own app shouldn't really have to care about the column names outside the `DatabaseFormsRepository` anyway.

grzesiek2010 (approved):

PR #4496, opened 2021-03-31
User can add a project with icon

### LINKED ISSUE #4422: User can add a project with icon
Blocked by #4420 

**Acceptance**

When I click "Add project"
And I enter a name
And I enter an icon (single character)
And I click "Add"
Then I see the project is added with my chosen name and icon (icon doesn't need to have a background color)

grzesiek2010:
Closes #4422

Screenshot_1617234951

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
@@ -20,10 +20,10 @@ class AddProjectDialogTest {
         val scenario = RobolectricHelpers.launchDialogFragment(AddProjectDialog::class.java)
         scenario.onFragment { f: AddProjectDialog ->
             val dialog = f.dialog
-            val cancelButton = dialog?.findViewById<MaterialButton>(R.id.cancel_button)
-            assertThat(cancelButton?.text, `is`(f.getString(R.string.cancel)))
-            cancelButton?.performClick()
-            assertThat(dialog?.isShowing, `is`(false))
+            val cancelButton = dialog!!.findViewById<MaterialButton>(R.id.cancel_button)
+            assertThat(cancelButton!!.text, `is`(f.getString(R.string.cancel)))
+            cancelButton.performClick()
+            assertThat(dialog.isShowing, `is`(false))
         }
     }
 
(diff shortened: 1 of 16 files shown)


collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
```diff
@@ -32,8 +39,22 @@ class AddProjectDialogTest {
...
         scenario.onFragment { f: AddProjectDialog ->
             val dialog = f.dialog
-            Espresso.onView(ViewMatchers.isRoot()).perform(ViewActions.pressBack())
+            onView(isRoot()).perform(ViewActions.pressBack())
             assertThat(dialog!!.isShowing, `is`(false))
         }
     }
+
+    @Test
+    fun oneCharacterOrEmojiOnly_shouldBeAcceptedAsProjectIcon() {
```
seadowg, line 50: I like this. Definitely good to check the emoji case here, so we don't accidentally break that! It'd also be good to add a test here that covers adding more than one character (it looks there is already implementation for that).

collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
```diff
@@ -32,8 +39,22 @@ class AddProjectDialogTest {
...
+        val scenario = RobolectricHelpers.launchDialogFragmentInContainer(AddProjectDialog::class.java)
+        scenario.onFragment {
+            // One character
+            onView(withHint(R.string.project_icon)).perform(ViewActions.typeText("XYZ"))
+            onView(Matchers.allOf(withText("X"), withEffectiveVisibility(Visibility.VISIBLE))).check(matches(isDisplayed()))
```
seadowg: I think using `withHint` again here would make the test a little easier to read (it'd be clearer that the same view is being checked). I just checked and doing `onView(allOf(withHint(R.string.project_icon), withText("X")))` passes and fails as expected.

seadowg: I'm definitely liking that the one character and emoji case are covered here, but I'm thinking it might be nice to split them, so it's clear the one character part is testing that entering **more than one** character is truncated.

seadowg (changes requested):
Looks great! Just one tweak in the tests would be nice before merge.

collect_app/src/test/java/org/odk/collect/android/projects/AddProjectDialogTest.kt
grzesiek2010: Ok, done.

seadowg (approved):
Ace! Thanks for indulging me there.

Issue #4497, opened 2021-03-31
Can select current project


seadowg:
**Acceptance**

- [x] Given I've added a project
When I click on that project in the project settings dialog
Then I see that project as my current project at the top of the project settings dialog ([mockup](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v1.31)?node-id=0%3A1))
And I do not see that project in the list of projects

PR #4501, opened 2021-04-02
User can add a Project with a color

### LINKED ISSUE #4422: User can add a project with icon

### LINKED ISSUE #4492: User can add a Project with a color
~~Blocked by #4422~~

**Acceptance**

When I click "Add project"
And I enter a name
And I enter an icon (single character)
And I enter a hex code in the "Color" field
And I click "Add"
Then I see the project is added with my chosen name and icon
And the icon has a background matching the hex code I entered

grzesiek2010:
Closes #4492

Screenshot_1617288416

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/DrawableMatcher.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/DrawableMatcher.java
@@ -1,5 +1,6 @@
 package org.odk.collect.android.support.matchers;
 
+import android.graphics.drawable.GradientDrawable;
 import android.view.View;
 import android.widget.ImageView;
 
@@ -23,6 +24,20 @@ public class DrawableMatcher  {
             @Override
             public boolean matchesSafely(ImageView imageView) {
                 return expectedResourceId == (Integer) imageView.getTag();
+            }
+        };
+    }
+
+    public static Matcher<View> withBackgroundColor(int color) {
+        return new BoundedMatcher<View, View>(View.class) {
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("View background color to be " + color);
+            }
+
+            @Override
+            public boolean matchesSafely(View view) {
+                return color == ((GradientDrawable) view.getBackground()).getColor().getDefaultColor();
             }
         };
     }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
@@ -22,6 +22,11 @@ internal class AddProjectDialogPage(rule: ActivityTestRule<*>) : Page<AddProject
         return this
     }
 
+    fun inputProjectColor(projectColor: String?): AddProjectDialogPage {
+        inputText(R.string.project_color, projectColor)
+        return this
+    }
+
     fun addProject(): MainMenuPage {
         Espresso.onView(withId(R.id.add_button)).perform(ViewActions.click())
         return MainMenuPage(rule).assertOnPage()
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
@@ -12,17 +12,17 @@ internal class AddProjectDialogPage(rule: ActivityTestRule<*>) : Page<AddProject
         return this
     }
 
-    fun inputProjectName(projectName: String?): AddProjectDialogPage {
+    fun inputProjectName(projectName: String): AddProjectDialogPage {
         inputText(R.string.project_name, projectName)
         return this
     }
 
-    fun inputProjectIcon(projectIcon: String?): AddProjectDialogPage {
+    fun inputProjectIcon(projectIcon: String): AddProjectDialogPage {
         inputText(R.string.project_icon, projectIcon)
         return this
     }
 
-    fun inputProjectColor(projectColor: String?): AddProjectDialogPage {
+    fun inputProjectColor(projectColor: String): AddProjectDialogPage {
         inputText(R.string.project_color, projectColor)
         return this
     }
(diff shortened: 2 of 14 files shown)


grzesiek2010:
This pr is based on #4422. We can close that one if there is not much to improve and keep this one or merge one after another. Whatever you prefer @seadowg

seadowg:
@grzesiek2010 is this meant to be closing #4497 rather than #4422 (which your earlier PR deals with)?

grzesiek2010:
> @grzesiek2010 is this meant to be closing #4497 rather than #4422 (which your earlier PR deals with)?

No, #4422 and and #4492 I know there is a separate pr for #4422 but I didn't want to wait so I implemented changes for #4492 on top of #4422 so this pr is for both and as I said you can review them separately or at once.

seadowg:
@grzesiek2010 right it's just that this also adds functionality around switching between projects. That looks like it's word towards #4497 to me. Maybe that's meant to be in a different branch?

grzesiek2010:
Ah that my mistake I also started working on #4497 and looks like I pushed here some commits from that branch, 4 last commits should be removed.
Ok, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/DrawableMatcher.java
```diff
@@ -26,4 +27,18 @@ public boolean matchesSafely(ImageView imageView) {
...
+    public static Matcher<View> withBackgroundColor(int color) {
+        return new BoundedMatcher<View, View>(View.class) {
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("View background color to be " + color);
+            }
+
+            @Override
+            public boolean matchesSafely(View view) {
+                return color == ((GradientDrawable) view.getBackground()).getColor().getDefaultColor();
```
seadowg, line 40: I was going to say that we didn't need to test color here (and should do it at a lower level) but you've found a really simple way to do it! Nice.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
```diff
@@ -22,6 +22,11 @@ internal class AddProjectDialogPage(rule: ActivityTestRule<*>) : Page<AddProject
         return this
     }
 
+    fun inputProjectColor(projectColor: String?): AddProjectDialogPage {
```
seadowg: This should just be `String` rather than `String?` I think.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
```diff
@@ -60,7 +60,7 @@ class AddProjectDialog : MaterialFullScreenDialogFragment() {
         }
 
         binding.addButton.setOnClickListener {
-            projectsRepository.add(getProjectName(), getProjectIcon())
+            projectsRepository.add(getProjectName(), getProjectIcon(), getProjectColor())
```
seadowg: It feels like it would be good at this point to have `ProjectsRepository#add` just take a `Project` as an argument instead of the individual components. What do you think?

seadowg (dismissed):

seadowg (changes requested):
As in comments

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AddProjectDialogPage.kt
grzesiek2010: Right, fixed.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
grzesiek2010: I agree, fixed.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
```diff
@@ -60,7 +64,7 @@ class AddProjectDialog : MaterialFullScreenDialogFragment() {
         }
 
         binding.addButton.setOnClickListener {
-            projectsRepository.add(getProjectName(), getProjectIcon(), getProjectColor())
+            projectsRepository.add(Project(uuid.generateUUID(), getProjectName(), getProjectIcon(), getProjectColor()))
```
seadowg: Ach sorry I'd assumed we'd keep the UUID generation in the repository (like how IDs are assigned in the Forms or Instance repository). That way it's ensured that any Project saved to the repo has a unique identifier and no outside caller can mess with that. Happy to discuss if that doesn't make sense.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
grzesiek2010: Like this https://github.com/getodk/collect/pull/4501/commits/348d7e8f6d75e13f04184ec47fe610867e7cf17a?

seadowg (approved):

seadowg:
@grzesiek2010 looks good but we have a test compile failure

collect_app/src/main/java/org/odk/collect/android/projects/Project.kt
```diff
@@ -1,7 +1,9 @@
...
 
 data class Project(
-    val uuid: String,
     val name: String,
-    val icon: String
-)
+    val icon: String,
+    val color: String
+) {
+    lateinit var uuid: String
```
seadowg, line 8: Oh I really wouldn't recommend using `lateinit` for values like this that may or may not exist. For me, `lateinit` is really only for cases like a JUnit test or `Activity` where part of the object's setup happens outside of the constructor (`@Before` or `onCreate`). Docs on `lateinit` actually reference a test [here](https://kotlinlang.org/docs/properties.html#late-initialized-properties-and-variables).

Here I'd say `uuid` should be a `String?` as it wouldn't be considered exceptional or problematic but makes sure any caller knows they have to check before using it (something we don't get from `lateinit`). 

This can all end up feeling a bit nasty in Kotlin as we might end up with several optional fields on an object (a creation date for instance). A nice way of modelling that if it becomes painful is to do something like this:

```kotlin
abstract class Project {
    abstract val name: String
    abstract val icon: String
    abstract val color: String

    data class New(val name: String, val icon: String, val color: String)
    data class Existing(val uuid: String, val name: String, val icon: String, val color: String)
}
```

seadowg (changes requested):

PR #4503, opened 2021-04-06
Resolves #3643 Test image widgets display correct image

### LINKED ISSUE #3643: Test image widgets display correct image
There aren't currently tests for the image widgets (widgets descending from `BaseImageWidget`) that ensures they display the correct image based on the file path in a prompt. We now have tests that they display an image default however (#3642) so we should add the "normal" case to these tests as well.

dimwight:
Closes #3643

This is my first attempt at a PR, I think the code is OK but not so sure about the process. 

First, my answers to standard questions.
All tests in affected files pass. Running All Tests for collect_app in AS gives 1803/1832 passed, the failures seem to be due to running on Windows eg if they depend on the value of the path separator

Discussed below 'Design decisions'. 

Users won't even notice?

No.

No. 

## Design decisions
Since this is my first go, maybe a good idea to show my workings. 

### BaseImageWidget
Now uses its QuestionMediaManager to find files instead of defining them itself. This key decision was pragmatic to start with, since in Android Studio on Windows the deprecated getInstanceFolder() returns C:/ and a current file cannot be created in getFile(). The change necessitated extra null checks, especially in addCurrentImageToLayout() to cope with QMM only returning existing files. Having been forced to this solution for getFile() I applied it throughout, was somewhat surprised to find that all tests still passed. 
#### createWidget()
Passes an anonymous subclass of FakeQuestionMediaManager to the constructor; I couldn't quite work out how to mock it and my old-fashioned approach seems to work. 
#### whenPromptHasDefaultAnswer_showsInImageView() 
Cosmetically polished, in particular to avoid the repeated string literal. 
#### whenPromptHasCurrentAnswer_showsInImageView() 
Derived from Default, depends on tweaked FakeQuestionMediaManager. 


dimwight (commented):
Oops! The first two commits result from mucking around and in the wrong branch at that.

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
```diff
@@ -30,7 +30,6 @@
 import android.widget.Toast;
```
dimwight, line 30: Deprecated getInstanceFolder() replaced by calls to field instance. 
Adds null checks in getFile(), addCurrentImageToLayout().

collect_app/src/test/java/org/odk/collect/android/widgets/DrawWidgetTest.java
```diff
@@ -19,14 +19,15 @@
 import org.odk.collect.android.activities.DrawActivity;
```
dimwight, line 18: createWidget() passes an anonymous subclass of FakeQuestionMediaManager to the constructor. 
Adds field for current file - required by fake QMM. Default is null,  will be initialized to real file only by Current test.
New field provides generalised prompt text for all tests.
Lightly polishes identifiers.

collect_app/src/test/java/org/odk/collect/android/widgets/DrawWidgetTest.java
```diff
@@ -123,4 +123,23 @@ public void whenPromptHasDefaultAnswer_showsInImageView() throws Exception {
         String loadedPath = shadowOf(((BitmapDrawable) drawable).getBitmap()).getCreatedFromPath();
         assertThat(loadedPath, equalTo(imagePath));
     }
+
+    @Test
+    public void whenPromptHasCurrentAnswer_showsInImageView() throws Exception {
+        String imagePath = File.createTempFile("current", ".bmp").getAbsolutePath();
```
dimwight, line 124: Closely derived from Default test.

dimwight (commented):
**Propagate Current test for subclasses** clones previous two commits to Tests for remaining subclasses of BaseImageWidget. 
Also where possible simplifies supplying string for getNextAnswer(), obviating override of setup().
CircleCI says '8 PMD rule violations were found'??

dimwight:
Sorry guys, seem to have made a mess of this. Any help will be greatly appreciated.

lognaturel:
I'm not seeing any mess! You can go to the ["Artifacts" tab](https://app.circleci.com/pipelines/github/getodk/collect/1308/workflows/e780c4fb-607c-4213-9e58-f7e9e0988812/jobs/12333/artifacts) in the CircleCI results to see the PMD complaints. It looks like they're all pretty small. You can also run `./gradlew checkCode` locally to get the same results.

Thanks so much for this, @dimwight! Contributors of tests are the true heroes of open source and it's a great way to get to know a code base.

dimwight:
I'm baffled. Thanks @lognaturel for the guidance on CircleCI check_quality, I made what seemed to be the required changes to four files. 
Committed one as b07494d which seemed to pass. Then the remaining three with identical changes as 9a117c7 which didn't. 
So I tried new commits of each file with trivial cosmetic changes, now they all fail. 
The Artifacts tab now shows three files, of which the pmd.html seems effectively blank and the others seem to list all files in the project.

lognaturel:
The empty `pmd` results file means you've satisfied `pmd`. The quality checks are run sequentially and fail early so passing `pmd` revealed some `checkstyle` issues. For better or worse, we have to use different tools for different kinds of checks. You're right that the results file shows every file in the project but it's sorted by number of issues decreasing and only the first few files have issues flagged. It looks like they're all whitespace and naming issues. They are indeed nitpicky -- we've found that having a convention that's consistently applied means no arguing about style and increased readability!

dimwight:
Thanks for the further guidance, I've no problem with picking nits once I know where to find them. 
A bit late in the evening here in Blighty, hopefully I'll get back onto the case tomorrow.

dimwight:
Finally ready for a human to look at? :roll_eyes:

dimwight:
Excuse my nagging @lognaturel but I'd really like some feedback on this!

lognaturel:
Thanks for your patience, @dimwight. Really appreciate you sticking with this, especially since I now realize there's a lot of context required. Thanks so much for getting ramped up and for the really great explanations around your changes!

The practical difference in the test coverage ends up being that the default variants go through `BaseImageWidget.getDefaultFilePath` to get the real path from the reference manager and the user specified/"normal" variants you've added go through `QuestionMediaManager.getAnswerFile`. 

I got tripped up understanding what the difference between the two variants are because of the reuse of `ANSWER_TEXT` which is a `jr://image` reference. It gets to the results you want because whatever the answer text is gets looked up in the `QuestionMediaManager` but it's confusing because in the "normal" case the answer text would be an unqualified filename. It's extra confusing because in the default test variants the answer value is not used because `currentFile` is `null`.

All that said, how about renaming `ANSWER_TEXT` to `DEFAULT_IMAGE_ANSWER` and also introducing `USER_SPECIFIED IMAGE_ANSWER = "current.bmp"`. Only `USER_SPECIFIED_IMAGE_ANSWER` would be mapped in your fake QMM since `DEFAULT_IMAGE_ANSWER` gets resolved by the `ReferenceManager`.

dimwight:
Thanks @lognaturel for the very helpful feedback. I've prototyped your naming suggestions (which incidentally help clarify the context for me).

collect_app/src/test/java/org/odk/collect/android/widgets/DrawWidgetTest.java
```diff
@@ -42,7 +42,8 @@
 @RunWith(RobolectricTestRunner.class)
 public class DrawWidgetTest extends FileWidgetTest<DrawWidget> {
 
-    private static final String ANSWER_TEXT = "jr://images/someURI";
+    private static final String DEFAULT_IMAGE_ANSWER = "jr://images/someURI";
```
dimwight: Could be local to `whenPromptHasDefaultAnswer_showsInImageView` but its usage parallels `USER_SPECIFIED_IMAGE_ANSWER ` which has to be at least an instance variable.

lognaturel: I like the symmetry as well.

lognaturel:
Looks great to me, @dimwight! Funny how much of an impact naming can have.

This is ready to merge as soon as you make the update across the tests. I think you could consider having the two constants defined in `DrawWidgetTest` and having all other tests reference them so they exist in just one place. If you're up for it, you could squash this down to one commit. I can also squash on merge.

I see that you've created an account on [the forum](https://forum.getodk.org/). I'd love to learn more about you and your ODK projects if you have a moment to [introduce yourself](https://forum.getodk.org/t/introduce-yourself-here/6671). If you're looking to get more into Collect development, we can also chat about some possible next issues to pick up [in Slack](http://slack.getodk.org/).

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
```diff
@@ -165,13 +164,19 @@ public void whenPromptHasCurrentAnswer_showsInImageView() throws Exception {
 
     @Test
     public void markupButtonShouldBeDisabledIfImageAbsent() throws Exception {
```
dimwight, line 166: Method tweaked to match (essentially cosmetic?) splitting of answer strings. On general principle it should test against both ways a bad path could arrive?

lognaturel (approved):
Thanks! This was not an easy first issue to tackle.

Since this does make (small) changes to non-test code, we'll do a quick QA pass so I've marked it as `needs testing`.

kkrawczyk123:
**Tested with success!**

Tested cases:
Regression checks on:
- image widget
- selfie widget
- draw widget
- signature widget
- annotate widget
- annotate and singature widgets with default images
- edit form and media
- attached big sized images

mmarciniak90:
Tested with success also on Android versions 7.0, 8.1 and 10.0

PR #4505, opened 2021-04-06
Can select current project

### LINKED ISSUE #4497: Can select current project

grzesiek2010:
Closes #4497

Screenshot_1617360712

No.

No.


collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/SwitchProjectTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    val rule = CollectTestRule()
+    private val projects = listOf(
+        Project("Turtle nesting", "T", "#00FF00", "1"),
+        Project("Polio - Banadir", "P", "#FF0000", "2")
+    )
+
+    @get:Rule
+    var chain: RuleChain = TestRuleChain
+        .chain()
+        .around(SaveProjectsRule(projects))
```
seadowg: Personally I think we should try and avoid creating data directly (not through the UI) in Espresso tests. There's a risk we bypass bugs that could exist in the full flow or accidentally automate steps that wouldn't really happen, and it also means our test has more opinions on the implementation than it needs to.

I can see that if we did create Projects through the UI in this test it might feel like we've got duplicate coverage between it and `AddProjectTest`. In that case I would maybe consider deleting `AddProjectTest` (assuming we all it's assertions end up being duplicated).

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ProjectMatcher.kt
```diff
@@ -0,0 +1,30 @@
...
+import android.widget.TextView
+import androidx.test.espresso.matcher.BoundedMatcher
+import androidx.test.espresso.matcher.ViewMatchers.assertThat
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.`is`
+import org.odk.collect.android.R
+import org.odk.collect.android.projects.Project
+
+object ProjectMatcher {
```
seadowg, line 13: I'm liking that you're creating a lot of custom matchers. They're really expressive but I for some reason have never managed to get into the habit of making them.

I'm wondering if this could be an alias rather than a full matcher though, so we can avoid using `findViewById` (again as it links us to the implementation in a small way). Could it just be:

```kotlin
fun withProject(name, icon) = allOf(hasChild(withText(name)), hasChild(withText(icon))
```

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ProjectSettingsDialogPage.kt
```diff
@@ -42,8 +42,13 @@ internal class ProjectSettingsDialogPage(rule: ActivityTestRule<*>) : Page<Proje
         return AddProjectDialogPage(rule).assertOnPage()
     }
 
-    fun assertProjectIconColor(color: String): ProjectSettingsDialogPage {
-        Espresso.onView(withId(R.id.project_icon)).check(matches(withBackgroundColor(Color.parseColor(color))))
+    fun assertCurrentProject(project: Project): ProjectSettingsDialogPage {
+        Espresso.onView(withId(R.id.current_project)).check(matches(withProject(project)))
```
seadowg, line 46: Again it would be nice to avoid `findViewById` in these methods, but it seems like a hard one. Sometimes that can be a hint that we've created an accessibility problem in our app - if we don't have a way to match uniquely on text (or content description) then it also means the screen would be ambiguous when read through a screen reader (it's interesting to think of our UI tests as using a screen reader as they need to do everything through text).

I just had a play around with the account settings dialog in Google's Contacts app with TalkBack enabled and when I clicked on the current account it said "Signed in as <my main account>" and when I tapped on one of the other accounts it said "Use <other account>". We could do something similar and have a `contentDescription` on the project item - for the current account it could be "Using <project name> project" and for others it could be "Switch to <project name> project". This way we'd have something to match on in Espresso (using `withContentDescription`) and also be supporting screen readers.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
```diff
@@ -30,9 +30,9 @@ abstract class ProjectsRepositoryTest {
         val projects = projectsRepository.getAll()
 
         assertThat(projects.size, `is`(3))
-        assertThat(projects[0], `is`(projectX))
-        assertThat(projects[1], `is`(projectY))
-        assertThat(projects[2], `is`(projectZ))
+        assertThat(projects[0], `is`(projectX.copy(uuid = projects[0].uuid)))
```
seadowg, line 33: Urgh I miss `copy` so much when I'm in Java!

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -511,4 +512,10 @@ public ProjectsRepository providesProjectsRepository(UUIDGenerator uuidGenerator
     public UUIDGenerator providesUUIDGenerator() {
         return new UUIDGenerator();
     }
+
+    @Provides
+    @Singleton
```
seadowg: This doesn't need to be a singleton right? Are you just thinking it is safe to be one and so it keeps the object count down?

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ProjectMatcher.kt
grzesiek2010: I'm not sure if I like it because such a function would just check if both texts are displayed but it doesn't check where, so theoretically we may end up with mixed icon and project name and it would still pass right?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ProjectSettingsDialogPage.kt
grzesiek2010: You mean replacing `withId` with `withContentDescription`? Should we start using `contentDescription` more often? I thought about it and I guessed it's not important in our app so I'm not sure.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/SwitchProjectTest.kt
grzesiek2010: I agree it's redundant. Fixed.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
grzesiek2010: Right, there is no need to add singleton annotation there, Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ProjectMatcher.kt
seadowg: It would check that both pieces of text are within the same parent view (`hasChild` only checks direct children I think) so I think it would protect against the case you're thinking of? My main goal was just to try and avoid view IDs or any knowledge of the view structure in the test.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ProjectSettingsDialogPage.kt
seadowg: Yeah there are other projects where I've just used `withContentDecription` as the main text matcher (instead of `withText`). It **usually** works exactly the same as you automatically get content descriptions for `TextView`s. For me, it's a win-win as it helps our app be more inclusive to people that use screen readers but also makes it easier to write tests for cases where we have icon buttons or list items that perform an action on click (neither of which would always be labelled with text).

collect_app/src/androidTest/java/org/odk/collect/android/support/matchers/ProjectMatcher.kt
grzesiek2010: > so I think it would protect against the case you're thinking of?

I don't think so. It would be called on a layout (project_list_item) which contains icon and project name (both TextViews) but if you don't use ids you can't be sure which element contains which value. You just confirm that both values exist.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ProjectSettingsDialogPage.kt
grzesiek2010: The drawback is that I would need to set it programmatically because if I added `android:contentDescription` to those layouts it's not visible for Espresso. Probably it works for single elements like textViews, imageViews in layouts only?

seadowg (approved):
We discussed separately, and I'll have a play with the Espresso tweaks outside this, so we can get this merged. Good to go once this has been rebased and the checks are green!

PR #4507, opened 2021-04-07
Rework Dagger setup for audiorecorder module

seadowg:
This came from a quick spike on trying to understand how we could set up Dagger for an extracted `projects` module. I realized it would be easier to just try a setup on this existing module. `audiorecorder` already had a Dagger setup but it was a little awkward and I wanted to get it closer to the [Android docs example](https://developer.android.com/training/dependency-injection/dagger-multi-module#implementation_with_dagger_subcomponents) (the one I actually like) as it felt like this would also be a good fit for `projects`.

In short, we end up with our module assuming that the application object (in our case `Collect`) will provide it with it's dependency component. The nice things is that this structure is it gives `Collect` the opportunity to override dependencies for the module if it wants (either by using Daggers "subcomponents" concept or by just overriding the "dependency module" like we do in tests). In my head it's the Dagger equivalent of offering a constructor with optional args - you can use the defaults if you don't care but also pass dependencies in if you do.

For historical context, I'm pretty sure I started with a setup like this for `audiorecorder` but then ran into problems making it work for Robolectric. I later (re)discovered the hidden ability to explicitly set an Application for Robolectric (rather than relying on the automagic Test<YourApplicationObject> thing) which I don't think I realized let me move back to this.

Covered above and in comments.

Should be safe to merge without QA. Generally we'd see a compile-time problem for dodgy Dagger setups.


audiorecorder/src/main/java/org/odk/collect/audiorecorder/DaggerSetup.kt
```diff
@@ -91,25 +79,14 @@ internal open class AudioRecorderDependencyModule {
...
-}
 
-internal fun RobolectricApplication.setupDependencies(module: AudioRecorderDependencyModule) {
-    _component = DaggerAudioRecorderDependencyComponent.builder()
-        .application(this)
-        .dependencyModule(module)
-        .build()
+    @Provides
+    @Singleton
+    internal open fun providesRecordingRepository(): RecordingRepository {
```
seadowg, line 89: Marking this `internal` means that a dependent application cannot override this dependency (we don't want them changing this probably).

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recorder/Recorder.kt
```diff
@@ -4,7 +4,7 @@ import org.odk.collect.audiorecorder.recording.MicInUseException
 import org.odk.collect.audiorecorder.recording.SetupException
 import java.io.File
 
-internal interface Recorder {
+interface Recorder {
```
seadowg, line 7: Types that can be override can't be `internal` any longer.

audiorecorder/src/main/java/org/odk/collect/audiorecorder/recording/AudioRecorderService.kt
```diff
@@ -1,17 +1,19 @@
...
+import org.odk.collect.audiorecorder.AudioRecorderDependencyComponentProvider
 import org.odk.collect.audiorecorder.recorder.Output
 import org.odk.collect.audiorecorder.recorder.Recorder
+import org.odk.collect.audiorecorder.recording.internal.RecordingForegroundServiceNotification
+import org.odk.collect.audiorecorder.recording.internal.RecordingRepository
 import java.io.Serializable
 import javax.inject.Inject
 
-internal class AudioRecorderService : Service() {
+class AudioRecorderService : Service() {
```
seadowg, line 16: Annoyingly things that we inject into can't be `internal` any longer because they appear in the component's `inject` methods. This is a weird side effect but not really a problem.

audiorecorder/src/test/java/org/odk/collect/audiorecorder/testsupport/RobolectricApplication.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.app.Application
+import org.odk.collect.audiorecorder.AudioRecorderDependencyComponent
+import org.odk.collect.audiorecorder.AudioRecorderDependencyComponentProvider
+import org.odk.collect.audiorecorder.AudioRecorderDependencyModule
+import org.odk.collect.audiorecorder.DaggerAudioRecorderDependencyComponent
+
+/**
+ * Used as the Application in tests in in the `test/src` root. This is setup in `robolectric.properties`
+ */
+internal class RobolectricApplication : Application(), AudioRecorderDependencyComponentProvider {
```
seadowg, line 12: I'm a big fan of how this setup works in test. We end up with (in my mind) a much more intuitive setup where the `RobolectricApplication` has to do exactly the same thing as `Collect` does.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -150,8 +155,17 @@ private void setupDagger() {
         applicationComponent = DaggerAppDependencyComponent.builder()
                 .application(this)
                 .build();
-
         applicationComponent.inject(this);
+
+        audioRecorderDependencyComponent = DaggerAudioRecorderDependencyComponent.builder()
```
seadowg, line 160: This is where we can override default module dependencies if we want to (using the `dependencyModule` method of the builder). For `projects` this would let us pass in a `ProjectsRepository` for instance (letting us share access to the projects with the module).

grzesiek2010 (approved):

PR #4510, opened 2021-04-09
Tweak projects form UI

### LINKED ISSUE #4506: Tweak projects form UI
- [x] Fix grey color on field highlight
- [x] All fields should be single line
- [ ] Enter button keyboard should go to next field in project form
- [x] Add/cancel bottom bar should float on top of keyboard

grzesiek2010:
Closes #4506

I improved things mentioned in #4506 and added some other small changes in UI.

No.

No.

### DIFF
--- a/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
+++ b/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
@@ -73,7 +73,8 @@
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="@dimen/margin_small"
-        android:background="?dividerColor"
+        android:background="?colorOnSurface"
+        android:alpha="0.2"
         app:layout_constraintStart_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/admin_settings_button" />
@@ -92,7 +93,8 @@
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="@dimen/margin_small"
-        android:background="?dividerColor"
+        android:background="?colorOnSurface"
+        android:alpha="0.2"
         app:layout_constraintStart_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/project_list" />
--- a/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
+++ b/collect_app/src/main/res/layout/project_settings_dialog_layout.xml
@@ -76,7 +76,7 @@
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="@dimen/margin_small"
-        android:background="?colorPrimary"
+        android:background="?dividerColor"
         app:layout_constraintStart_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/admin_settings_button" />
@@ -94,7 +94,7 @@
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="@dimen/margin_small"
-        android:background="?colorPrimary"
+        android:background="?dividerColor"
         app:layout_constraintStart_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/project_list" />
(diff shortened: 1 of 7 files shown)


collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -76,25 +75,26 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/top_divider" />
 
     <View
         android:id="@+id/bottom_divider"
         android:layout_width="match_parent"
         android:layout_height="1dp"
         android:layout_marginTop="@dimen/margin_small"
-        android:background="?colorPrimary"
+        android:background="?dividerColor"
```
seadowg, line 97: I think we should use a Material theme attribute here rather than `dividerColor` which is a custom attribute (I'd like to get rid of as well). This should be a `colorOnSurface` I think.

seadowg: I think we should use a Material theme attribute here rather than dividerColor which is a custom attribute (I'd like to get rid of as well). This should be a colorOnSurface I think.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -53,21 +53,20 @@
         android:id="@+id/general_settings_button"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        style="@style/Widget.MaterialComponents.Button.OutlinedButton"
+        style="@style/Widget.Collect.Button.OutlinedButton"
         android:layout_marginTop="@dimen/margin_small"
         android:text="@string/general_preferences"
-        android:textColor="@color/deep_purple_500"
+        android:textColor="?colorOnPrimary"
```
seadowg: This should be `colorOnSurface` instead of `colorPrimary`. In our light and dark themes they happen to match, but you'll see this doesn't work in Magenta where they are different. There's some okish docs on the colors and what they're used for [here](https://material.io/develop/android/theming/color). I'm also thinking this could just go in the style?

collect_app/src/main/res/values/styles.xml
```diff
@@ -37,4 +37,18 @@
     <style name="Widget.Collect.BottomSheet.Modal" parent="Widget.Design.BottomSheet.Modal">
         <item name="behavior_peekHeight">300dp</item>
     </style>
+
+    <style name="Widget.Collect.TextInputLayout.OutlinedBox" parent="Widget.MaterialComponents.TextInputLayout.OutlinedBox">
+        <item name="boxStrokeColor">?colorOnPrimary</item>
```
seadowg: Again this should be `colorOnSurface`.

seadowg (changes requested):
Looking good! Just a few tweaks to what color attrs we use is needed. I find testing in Magenta is a good way to check I'm using the right theme values as it has a starker difference between the underlying "primary" and "surface" colors.

collect_app/src/main/res/values/styles.xml
grzesiek2010: Fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: Right, fixed.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: I'm not sure about it, then it would be very thick:
Screenshot_1618224683
It doesn't look like the divider in Google dialogs.

grzesiek2010: maybe we should use it + transparency or something?

grzesiek2010: ok I did it using transparency https://github.com/getodk/collect/pull/4510/commits/7cf918af33d1bc1d771b128e92266e2397dd1ef3 and it looks fine.

collect_app/src/main/res/values/attrs.xml
```diff
@@ -2,7 +2,6 @@
 <resources>
     <declare-styleable name="Collect">
         <attr name="iconColor" format="color" />
-        <attr name="dividerColor" format="color" />
```
seadowg, line 5: Amazing! Thanks for also getting rid of it 🙌.

seadowg (approved):

PR #4512, opened 2021-04-12
Implement InstanceProvider using InstancesRepository

### LINKED ISSUE #4453: Implement ContentProviders using repositories

seadowg:
Closes #4453 

Again pretty much just follows the steps outlined in the issues so reviewing commit by commit is best. There are a few things I'm going to clean up in a follow-up PR that I'll timebox for myself (this afternoon and tomorrow morning). In that I'm going to try and:

1. Remove zero arg constructors from repository implementations
1. Reorganize the `database` package so it's consistent (there is no `DatabaseInstancesColumns` at the moment for example)
1. Move `forms` and `instances` into their own module (including any relevant tests)
1. Move content provider code into a module (including any relevant tests)

The timeboxing is really to stop me going down a rabbit hole on the last point there as it might involve moving a lot of code and is really just a nice to have. None of these things are essential but I think definitely worth spending a day on to do some cleaning up while we're in the neighbourhood.

This should be pretty similar to the `FormsProvider` PR so not sure there is anything to comment on here.

After this merges I think we want to do a QA pass on master. We can discuss that separately.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/database/DatabaseInstancesRepository.java
+++ b/collect_app/src/main/java/org/odk/collect/android/database/DatabaseInstancesRepository.java
@@ -44,7 +44,7 @@ public final class DatabaseInstancesRepository implements InstancesRepository {
         String selection = _ID + "=?";
         String[] selectionArgs = {Long.toString(databaseId)};
 
-        try (Cursor cursor = query(selection, selectionArgs)) {
+        try (Cursor cursor = query(null, selection, selectionArgs, null)) {
             List<Instance> result = getInstancesFromCursor(cursor);
             return !result.isEmpty() ? result.get(0) : null;
         }
@@ -54,7 +54,7 @@ public final class DatabaseInstancesRepository implements InstancesRepository {
     public Instance getOneByPath(String instancePath) {
         String selection = INSTANCE_FILE_PATH + "=?";
         String[] args = {new StoragePathProvider().getRelativeInstancePath(instancePath)};
-        try (Cursor cursor = query(selection, args)) {
+        try (Cursor cursor = query(null, selection, args, null)) {
             List<Instance> instances = getInstancesFromCursor(cursor);
             if (instances.size() == 1) {
                 return instances.get(0);
@@ -66,14 +66,14 @@ public final class DatabaseInstancesRepository implements InstancesRepository {
 
     @Override
     public List<Instance> getAll() {
-        try (Cursor cursor = query(null, null)) {
+        try (Cursor cursor = query(null, null, null, null)) {
             return getInstancesFromCursor(cursor);
         }
     }
 
     @Override
     public List<Instance> getAllNotDeleted() {
-        try (Cursor cursor = query(DELETED_DATE + " IS NULL ", null)) {
+        try (Cursor cursor = query(null, DELETED_DATE + " IS NULL ", null, null)) {
             return getInstancesFromCursor(cursor);
         }
     }
@@ -95,7 +95,7 @@ public final class DatabaseInstancesRepository implements InstancesRepository {
 
     @Override
     public List<Instance> getAllByFormId(String formId) {
-        try (Cursor c = query(JR_FORM_ID + " = ?", new String[]{formId})) {
+        try (Cursor c = query(null, JR_FORM_ID + " = ?", new String[]{formId}, null)) {
             return getInstancesFromCursor(c);
         }
     }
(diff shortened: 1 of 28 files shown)


collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
```diff
@@ -0,0 +1,343 @@
...
+        new StorageInitializer().createOdkDirsOnStorage();
+
+        Context context = ApplicationProvider.getApplicationContext();
+        contentResolver = context.getContentResolver();
+        externalFilesDir = context.getExternalFilesDir(null);
+    }
+
+    @Test
+    public void insert_addsInstance() {
+        ContentValues values = getContentValues("/blah", "External app form", "external_app_form", "1");
```
grzesiek2010, line 65: I think we do not store `instanceFilePath` with that slash so it's a bit misleading to use it in tests.

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
```diff
@@ -0,0 +1,343 @@
...
+        assertThat(contentResolver.getType(CONTENT_URI), is(CONTENT_TYPE));
+        assertThat(contentResolver.getType(Uri.withAppendedPath(CONTENT_URI, "1")), is(CONTENT_ITEM_TYPE));
+    }
+
+    private File createInstanceDirAndFile() {
+        return InstanceUtils.createInstanceDirAndFile(externalFilesDir + File.separator + "instances");
+    }
+
+    private Uri addInstanceToDb(String instanceFilePath, String displayName) {
+        ContentValues values1 = getContentValues(instanceFilePath, displayName, "external_app_form", "1");
```
grzesiek2010: Just `values` not `values1`.

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
```diff
@@ -0,0 +1,343 @@
...
+    @Test
+    public void insert_addsInstance() {
+        ContentValues values = getContentValues("/blah", "External app form", "external_app_form", "1");
+        contentResolver.insert(CONTENT_URI, values);
+
+        try (Cursor cursor = contentResolver.query(CONTENT_URI, null, null, null)) {
+            assertThat(cursor.getCount(), is(1));
+
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(INSTANCE_FILE_PATH)), is("/blah"));
```
grzesiek2010, line 72: Maybe it would be good to check the number of columns as well so that in the future if we add something this tests will fail and force us to assert that new columns as well.

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
```diff
@@ -0,0 +1,343 @@
...
+        addInstanceToDb("/blah3", "Matching instance");
+
+        try (Cursor cursor = contentResolver.query(CONTENT_URI, null, DISPLAY_NAME + "=?", new String[]{"Matching instance"}, null)) {
+            assertThat(cursor.getCount(), is(2));
+
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(INSTANCE_FILE_PATH)), is(isOneOf("/blah1", "/blah3")));
+
+            cursor.moveToNext();
+            assertThat(cursor.getString(cursor.getColumnIndex(INSTANCE_FILE_PATH)), is(isOneOf("/blah1", "/blah3")));
```
grzesiek2010: It would be good to assert that those values are different as well right because you can have two rows with `/blah1` or two with `/blah3` and this test will also pass.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/database/DatabaseInstancesRepository.java
```diff
@@ -2,104 +2,148 @@
...
- * Over time, we should consider redefining the responsibility split between
- * {@link org.odk.collect.android.provider.InstanceProvider}, {@link InstancesRepository} and
- * {@link InstancesDao}.
  */
 public final class DatabaseInstancesRepository implements InstancesRepository {
 
+    private final InstancesDatabaseProvider instancesDatabaseProvider;
+
+    public DatabaseInstancesRepository() {
+        instancesDatabaseProvider = DaggerUtils.getComponent(Collect.getInstance()).instancesDatabaseProvider();
```
grzesiek2010, line 44: Why don't you pass it as a parameter like we do in probably all the other cases like that?

seadowg: Yeah agreed this isn't great. I'm just cleaning this up in a follow up PR - I'm removing the zero arg constructors in the repo implementation. I just had them in to make it easier to construct the repos while concentrating on swapping out the cursor/content provider stuff.

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
seadowg: Right but at the content provider level we do deal in absolute paths - they just get converted in `DatabaseInstanceRepository` before we store them.  I'm worried switching to relative paths here means the test reader need more context to understand what the paths are relative to. Does that make sense?

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
seadowg: Yeah, that's a great idea. I'll do that for forms as well.

collect_app/src/test/java/org/odk/collect/android/provider/InstanceProviderTest.java
seadowg: Good point! I'll revise the test.

grzesiek2010 (approved):

PR #4514, opened 2021-04-14
Move forms and instances data code to module

seadowg:
A lot of this PR ends up being just moves and changes to imports which can be a little hard to follow, so I'd advise going commit by commit. I'll also lay out the key changes here:

* Introduced providers for the forms and instances repositories (lets them not have zero arg constructors and will be needed for projects)
* Move Form and Instances pure Java (non-Android) data code to a new `forms` module
* Move test helpers for the above (`FormsUtils` and `FormsRepositoryTest` for example) to a new `formstest` module
* Create a new `shared` module for common non-Android code (and renamed the old `shared` to `androidshared`)
* Reorganized the `databases` package so forms and instances have the same structure

There are some things I would have liked to do but didn't get time to (I gave myself a one-day timebox for this). Let me know if you think any of these should go in for completeness: 

1. Move the `database` package out to a new module
1. Move the forms and instances content provider implementations to their own Android library module

It's worth noting that these are in order as each change would depend on the last. Another possible positive change we could make now is to move `openrosa` (our open rosa based implementation of `FormSource`) out to a new module.

Ran tests.

I wanted to specifically move the abstractions and data classes around forms and instances out to give us scope to move other parts of the code out to their own modules (form management, content providers etc). Most of the other changes are simply facilitating that.

I think this is pretty safe as it's mostly just moving things around. Shouldn't need QA if all checks are passing.


seadowg:
I've added a `README.md` to `formstest` to try and explain what it's for.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
```diff
@@ -0,0 +1,121 @@
...
+        values.put(FORM_MEDIA_PATH, storagePathProvider.getRelativeFormPath(form.formMediaPath))
+        values.put(LANGUAGE, form.language)
+        values.put(AUTO_SEND, form.autoSend)
+        values.put(AUTO_DELETE, form.autoDelete)
+        values.put(GEOMETRY_XPATH, form.geometryXpath)
+        return values
+    }
+
+    @JvmStatic
+    fun getFormFromValues(values: ContentValues, storagePathProvider: StoragePathProvider): Form? {
```
grzesiek2010: Is this class really a good place for method like this one? This was a class for constants rather.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormsRepository.java
```diff
@@ -38,10 +38,10 @@
 public class DatabaseFormsRepository implements FormsRepository {
 
     private final StoragePathProvider storagePathProvider;
-    private final Clock clock;
+    private final Supplier<Long> clock;
     private final FormsDatabaseProvider formsDatabaseProvider;
 
-    public DatabaseFormsRepository(Clock clock, StoragePathProvider storagePathProvider, FormsDatabaseProvider formsDatabaseProvider) {
+    public DatabaseFormsRepository(Supplier<Long> clock, StoragePathProvider storagePathProvider, FormsDatabaseProvider formsDatabaseProvider) {
```
grzesiek2010, line 44: This change will work assuming that we create a repository object every time we want to use it. OK maybe now this is the case but what if at some point someone reuses an already created object? Isn't it a bit dangerous?
What's the advantage of such a change?

grzesiek2010 (changes requested):
We have more and more modules. It would be good to have a graphical representation of them so that we can just take a look and understand the dependencies between them.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
seadowg: It seemed to make sense to me as it you often end up needing both the columns and the conversion between `ContentValues`/`Cursor` in the same place. I do see what you mean though. You think another "helpers" or "utils" is in order?

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormsRepository.java
seadowg: The advantage for me is using a standard library abstraction rather than a custom one - we can probably get rid of `Clock` tbh. What is the worry around the object being reused?

grzesiek2010: The worry is that if you pass long value when you create an object it's a constant and if you reuse the object later that value is not you current timestamp anymore so wrong timestamps might be saved to the database?

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
grzesiek2010: Yeah to me it seems odd and like something I wouldn't do so it cough my attention.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormsRepository.java
grzesiek2010: Ahh my mistake that's not the way it works probably now you are wondering what I was talking about sorry it won't be outdated.

seadowg: Ah I see what you were worried about! Right yeah as you say, we still get a different value each time, so we're ok. Definitely a good one to look out for though.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
seadowg: Yeah happy to move a helper out if it feels wrong! I'll do that.

.circleci/config.yml
```diff
@@ -17,19 +17,23 @@ jobs:
     steps:
       - checkout
 
+      - run:
```
seadowg, line 20: Added some proper caching based on dependencies changing to help speed up CI. I was running into problems with timeouts so thought it would be food to fix while I'm here.

grzesiek2010 (approved):

PR #4517, opened 2021-04-15
Add first launch experience

### LINKED ISSUE #4498: First launch experience for projects
Blocked by #4497
Mockup is [here](https://www.figma.com/file/SkyAE9BZVmPQYnrPb30peg/First-launch-experience?node-id=0%3A1).

**Acceptance**

- [x] Given I've installed Collect for the first time on a device
When I launch the app
Then I see a screen asking whether I want to create a new project or use the demo one

- [x] Given I've installed Collect for the first time on a device and launched the app
When I click on create new project
And I fill in the project details
And I click "Add"
Then I land on the main menu
And my current project is the one I just added
And there are no other projects

- [x] Given I've installed Collect for the first time on a device and launched the app
When I click on use demo project
Then I land on the main menu
And my current project is called "Demo"

grzesiek2010:
Closes #4498

Screenshot_1618502597

No.


collect_app/src/test/java/org/odk/collect/android/projects/ProjectSettingsDialogTest.kt
```diff
@@ -68,6 +69,7 @@ class ProjectSettingsDialogTest {
         }
     }
 
+    @Ignore
```
grzesiek2010, line 72: I disabled this test because I had hard time fixing it (probably something with scrollable) since it's not very important for now I'll look into it one the whole dialog is done.

collect_app/src/main/res/layout/first_launch_dialog_layout.xml
```diff
@@ -0,0 +1,81 @@
...
+            style="@style/TextAppearance.Collect.Headline5"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="@dimen/margin_extra_large"
+            android:textColor="?colorOnSurface"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/logo" />
+
+        <com.google.android.material.button.MaterialButton
```
seadowg, line 39: I think we should just hide this button for the moment as it doesn't do anything.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -95,59 +83,14 @@ private void endSplashScreen() {
         finish();
     }
 
-    // decodes image and scales it to reduce memory consumption
-    private Bitmap decodeFile(File f) {
```
seadowg, line 99: Ah nice catch

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
```diff
@@ -66,9 +60,6 @@ private void init() {
 
         if (showSplash) {
             startSplashScreen(splashPath);
-
-            String splashPathHash = FileUtils.getMd5Hash(new ByteArrayInputStream(splashPath.getBytes()));
-            analytics.logEvent(SHOW_SPLASH_SCREEN, splashPathHash, "");
```
seadowg, line 87: Are we definitely happy to get rid of this? If so we should remove the event from `AnalyticsEvents` as well.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.kt
```diff
@@ -18,8 +18,7 @@ import android.os.Bundle
 import android.os.Handler
 import android.view.View
 import android.view.Window
-import android.widget.ImageView
-import android.widget.LinearLayout
+import kotlinx.android.synthetic.main.splash_screen.*
```
seadowg: I had a hunch that this was deprecated now (due to it conflicting a little with View Binding) and it annoyingly is (see [the announcement](https://android-developers.googleblog.com/2020/11/the-future-of-kotlin-android-extensions.html) and [migration docs](https://developer.android.com/topic/libraries/view-binding/migration)).

As a side note: urggghhh. The synthetic property stuff was a lot nicer than the View Binding library IMO.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.kt
```diff
@@ -63,6 +65,9 @@ class SplashScreenActivity : AppCompatActivity() {
             splash.visibility = View.VISIBLE
         }
 
-        Handler().postDelayed({ endSplashScreen() }, 2000)
+        lifecycleScope.launch() {
```
seadowg: Nice!

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+    val splashScreenLogoFile
+        get() = File(generalSettings.getString(GeneralKeys.KEY_SPLASH_PATH) ?: "")
+
+    val scaledSplashScreenLogoBitmap: Bitmap
+        get() = FileUtils.getBitmapScaledToDisplay(splashScreenLogoFile, ScreenUtils.getScreenHeight(), ScreenUtils.getScreenWidth())
+
+    val doesLogoFileExist
+        get() = splashScreenLogoFile.exists()
+
+    interface FactoryFactory {
```
seadowg: Maybe I'm missing something here but I don't think this needs to interact with any of the `SavedState` stuff. Could it not just work like the `MainMenuViewModel.Factory`?

collect_app/src/main/java/org/odk/collect/android/utilities/AppStateProvider.kt
```diff
@@ -0,0 +1,18 @@
+package org.odk.collect.android.utilities
+
+import android.content.Context
+
+class AppStateProvider(private val context: Context) {
+
+    fun isFreshInstall(): Boolean {
+        return try {
+            val firstInstallTime = context.packageManager.getPackageInfo(context.packageName, 0).firstInstallTime
```
seadowg: I think we need to check for there being no projects instead here right? Otherwise, if the user does "Clear Storage" they won't get taken back to the first launch.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.kt
```diff
@@ -43,10 +45,10 @@ class SplashScreenActivity : AppCompatActivity() {
     private fun init() {
         setContentView(R.layout.splash_screen)
 
-        if (viewModel.shouldDisplaySplashScreen) {
-            startSplashScreen()
-        } else {
-            endSplashScreen()
+        when {
```
seadowg, line 58: Oooo not sure how I feel about using `when` to replace an `if-else` (not pattern matching on a piece of data). It does look like what you're doing follows the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#if-versus-when) though so I'll defer to that

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.odk.collect.android.R
+import org.odk.collect.android.activities.ActivityUtils
+import org.odk.collect.android.activities.MainMenuActivity
+import org.odk.collect.material.MaterialFullScreenDialogFragment
+
+class FirstLaunchDialog : MaterialFullScreenDialogFragment() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(DialogFragment.STYLE_NO_TITLE, R.style.Theme_Collect_Light)
```
seadowg: I don't think we always want this to use the light theme right?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import org.odk.collect.android.preferences.keys.MetaKeys
+import org.odk.collect.android.preferences.source.Settings
+
+class ProjectImporter(
+    private val projectsRepository: ProjectsRepository,
+    private val metaSettings: Settings
+) {
+    fun importDemoProject() {
+        val project = Project("Demo project", "D", "#3e9fcc", "1")
```
seadowg: Why does this end up with a hardcoded ID (rather than just using a generated UUID)?

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
```diff
@@ -29,8 +29,11 @@ class SplashScreenViewModel(
     val doesLogoFileExist
         get() = splashScreenLogoFile.exists()
 
-    val shouldFirstLaunchDialogBeDisplayed
-        get() = appStateProvider.isFreshInstall()
+    fun shouldFirstLaunchDialogBeDisplayed(): Boolean {
+        val isFirstLaunch = !metaSettings.contains(MetaKeys.KEY_FIRST_LAUNCH)
```
seadowg, line 31: This could potentially just if there are projects or not? That would mean we wouldn't need an extra piece of state. Maybe you already though of that and there is a problem with it though!

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
```diff
@@ -45,6 +46,18 @@ abstract class ProjectsRepositoryTest {
...
     }
 
+    @Test
+    fun add_shouldAddUuidIfNotSpecified() {
+        projectsRepository.add(projectX)
+        assertThat(projectsRepository.getAll()[0].uuid, `is`(not(isEmptyString())))
+    }
+
+    @Test
+    fun add_shouldNotAddUuidIfSpecified() {
```
seadowg: Again, not sure we need this.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialogTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.odk.collect.android.projects.ProjectsRepository
+import org.odk.collect.android.support.RobolectricHelpers
+import org.odk.collect.android.utilities.TranslationHandler.getString
+import org.odk.collect.android.version.VersionInformation
+
+@RunWith(AndroidJUnit4::class)
+class FirstLaunchDialogTest {
+
+    @Test
+    fun clickingDeviceBackButton_shouldNotDismissDialog() {
```
seadowg: These tests could use the "backticks and spaces" style

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialogTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        val scenario = RobolectricHelpers.launchDialogFragment(FirstLaunchDialog::class.java)
+        scenario.onFragment { f: FirstLaunchDialog ->
+            val dialog = f.dialog
+            MatcherAssert.assertThat(dialog!!.isShowing, CoreMatchers.`is`(true))
+            Espresso.onView(ViewMatchers.isRoot()).perform(ViewActions.pressBack())
```
seadowg: Could we do static imports for all the matchers and Espresso helpers here? Also, I'm pretty sure we can just use `Espresso.pressBack` to avoid an `onView...`.

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModelTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Before
+    fun setup() {
+        generalSettings = Mockito.mock(SharedPreferencesSettings::class.java)
+        metaSettings = Mockito.mock(SharedPreferencesSettings::class.java)
+        splashScreenViewModel = SplashScreenViewModel(generalSettings, metaSettings)
+    }
+
+    @Test
+    fun whenSplashScreenDisabled_shouldShouldDisplaySplashScreenReturnFalse() {
```
seadowg: Again we could use the backtick style here.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        })
+    }
+
+    @Test
+    fun whenAppLaunchedForTheFistTime_showFirstLaunchScreen() {
+        doReturn(true).`when`(splashScreenViewModel).shouldFirstLaunchDialogBeDisplayed()
+
+        val scenario = ActivityScenario.launch(SplashScreenActivity::class.java)
+        scenario.onActivity { activity: SplashScreenActivity ->
+            MatcherAssert.assertThat(activity.supportFragmentManager.findFragmentByTag(FirstLaunchDialog::class.qualifiedName), Matchers.`is`(notNullValue()))
```
seadowg: There's a helper for this in `DialogUtils`

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun whenAppLaunchedNotForTheFistTimeAndSplashScreenEnabled_showMainMenuActivityAutomaticallyAfterTwoSeconds() {
+        doReturn(false).`when`(splashScreenViewModel).shouldFirstLaunchDialogBeDisplayed()
+        doReturn(true).`when`(splashScreenViewModel).shouldDisplaySplashScreen
+        doReturn(false).`when`(splashScreenViewModel).doesLogoFileExist
+
+        Intents.init()
+
+        val scenario = ActivityScenario.launch(SplashScreenActivity::class.java)
+        scenario.onActivity { activity: SplashScreenActivity ->
+            activity.lifecycleScope.launch {
```
seadowg: I'm worried this could be flakey (concurrency with timing). Do we have a way to switch out the dispatcher in tests? It looks like there are some examples of testing components that use coroutines [here](https://developer.android.com/codelabs/kotlin-coroutines?hl=uk#5).

One thing I might suggest is using the `Scheduler` interface for this and swapping it out in the test (we have fakes already) for something that just runs everything immediately (rather than using coroutines). The "delay" here feels like fuzzy behaviour that we might want to tweak and play around with without tests changing - we only really care about what happens after the delay.

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
```diff
@@ -61,7 +65,12 @@ class AddProjectDialog : MaterialFullScreenDialogFragment() {
 
         binding.addButton.setOnClickListener {
             projectsRepository.add(Project(getProjectName(), getProjectIcon(), getProjectColor()))
-            dismiss()
+            if (isDialogStartedFromFirstLaunchScreen()) {
```
seadowg: I'm thinking it might be nicer to do this with an "on finish" listener (the classic "see if activity implements the listener and call it if it does" pattern). Then the `AddProjectDialog` doesn't have to know about `CurrentProjectProvider` or `MainMenuActivity` (both would prevent it from being pulled out to a module).

seadowg (changes requested):
Flow looks good, but I have a few questions and comments inline. 

I think it would also be good to think about a feature test for these onboarding flows (adding a project and using the demo one). There's a lot of unit tests here which is good but adding a feature level test might let you lower the unit level coverage and also provides a simpler way to run a couple of tests that check everything has been glued together correctly. 

I also think we should now have our other feature tests start from the `SplashScreenActivity` rather than the `MainMenuActivity`. Right now all our tests skip this first launch logic which could result in us missing problems in the future. To do this we probably want our tests to just always click though the first launch (using the demo project).

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
grzesiek2010: Let's ask @lognaturel I think we kept that even long enough to collect data we and make a decision to keep or remove that screen.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialog.kt
grzesiek2010: This particular line is removed but generally the activity uses a theme that extends `Theme.Collect.BaseLight`. I think that's ok because it's our default them and if you install the app for the first time to display that dialog that light them should be used.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
grzesiek2010: I implemented it that way to identify the default project somehow. Thanks to that we can assume that if on the list of projects there is one with uuid 1 it's the default project.
This might help us for example to manage preferences. For all the other projects to access their preferences you would use their uuid but you can't do the same for the default project since it should use default preferences.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
grzesiek2010: Which one there is only getDialog() but it's depreceted?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
seadowg: Hmmm that's a good point. Maybe we should use a more human-readable string like "DEMO" then? Making it a constant is probably a good move as well.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
seadowg: Ach sorry it's a test helper - `RobolectricHelpers.getFragmentByClass`.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialogTest.kt
grzesiek2010: This doesn't work in all cases I don't know why. i tested it and it worked in 1/3 of my tests.

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModelTest.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialogTest.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
grzesiek2010: If you upgrade from v1.30 for example you also won't have any projects but you don't expect to see that first launch screen, you want your default project to be imported.

collect_app/src/main/java/org/odk/collect/android/utilities/AppStateProvider.kt
grzesiek2010: This class no longer exist I replaced this approach with `KEY_FIRST_LAUNCH` in `MetaKeys`.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.kt
grzesiek2010: Oh right, fixed.

collect_app/src/main/res/layout/first_launch_dialog_layout.xml
grzesiek2010: I would keep especially taking into account that this pr won't be tested as a single pr. it we don't have separate issues for the logo or version but I also added t here as a part of this pr.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
grzesiek2010: I'm not sure if I understand... "DEMO" just for project name? or what?

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialogTest.kt
seadowg: Huh. That's annoying!

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
seadowg: Sorry I meant for the ID.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
seadowg: Makes sense!

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
grzesiek2010: Ah yeah that's a good idea. Added.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
grzesiek2010: I think it's important to make sure that when we provide our own id it shouldn't be changed. It's important in the case of importing the demo project when we pass "DEMO" id to identify that project later (see our discussion above).

collect_app/src/main/java/org/odk/collect/android/projects/AddProjectDialog.kt
grzesiek2010: Makes sense. Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
grzesiek2010: Could I override it in tests like I do now or change programmatically?

collect_app/src/test/java/org/odk/collect/android/projects/ProjectsRepositoryTest.kt
seadowg: Sorry should have resolved this after we discussed that.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
grzesiek2010: Ok this is improved. I need to read more about testing such cases because as we discussed on slack it's not clear why some failures occur.

collect_app/src/main/java/org/odk/collect/android/activities/SplashScreenActivity.java
lognaturel: Yes, let's get rid of the analytics for now. The splash screen is not a hugely popular feature (~5% of users see a splash screen) but a few big projects seem to make use of it. I think it's worth maintaining for now and we can bring the analytics back again if we see a strong reason to remove the feature. @grzesiek2010 made the point that it might be kind of nice in a multi-project context because then each project can have its own splash screen which gives users an expectation of what context they're in before Collect starts.

seadowg: Sounds good! We should just delete the constant in `AnalyticsEvents` to finish the clean up then.

grzesiek2010: Ok I removed that event.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
grzesiek2010: Ok I improved it.

grzesiek2010:
@seadowg looks like everything is fixed. I know that you wanted one feature test but let me add it later because 
- it's not very important since I have a lot unit test
- I don't want to make conflicts with your pr https://github.com/getodk/collect/pull/4522 so I'll add it one it's merged

lognaturel:
Didn't see anything about this so for the record, @grzesiek2010 and I discussed the resolution of the image briefly. We're leaving as-is for now because we will likely do a little iteration to make the screen look more friendly ([forum discussion](https://forum.getodk.org/t/collect-prompt-for-server-configuration-on-first-launch/33051/4)). There may or may not be a clipboard there after that.

seadowg:
> @seadowg looks like everything is fixed. I know that you wanted one feature test but let me add it later because
>
> * it's not very important since I have a lot unit test
> * I don't want to make conflicts with your pr #4522 so I'll add it one it's merged

Fair enough! It was more a point for discussion and maybe something to think about for future features. The right balance for app level vs integration vs classic unit tests is a moving target for any new or existing piece of code. Maybe we can look at that for this in one of our catch ups.

> Didn't see anything about this so for the record, @grzesiek2010 and I discussed the resolution of the image briefly. We're leaving as-is for now because we will likely do a little iteration to make the screen look more friendly (forum discussion). There may or may not be a clipboard there after that.

Sounds good. I think avoiding trying to hit pixel perfection for these PRs is a good attitude. Getting the skeleton for these flows in gets the hard parts out of the way and we can tweak to our hearts content later.

seadowg (approved):

Issue #4474, opened 2021-03-15
Instances added/removed on disk should update in Collect


seadowg:
~~Blocked by #4417~~

The `InstanceSyncTask` not being run is causing a bug where the progress bar always appears on some screens (Edit Saved, Send Finalized, View Sent and Delete Saved).

## Acceptance

- [x] Given I've created a directory for an instance and copied an instance file into it
When I open the app
And I go to "Edit Saved Form"
And I click on the filled form (of the instance I copied)
Then the filled form opens as normal

- [x] Given I've filled out a form
And I've then deleted the instance directory
When I go to "Edit Saved Form"
And I click on the filled form (of the instance I deleted)
Then I don't see a crash

## Notes

In previous versions of Collect `InstanceSyncTask` would run whenever the user views an instance list in the app (Edit Saved Forms etc). This would do two things:

* Add any instances that are on disk but in the db to the db (and encrypt them if the form includes encryption)
* Remove any instances that are in the db but not on disk from the db

This is mostly undocumented and untested behaviour and there are some performance concerns around loading every instance into memory to do this every time we view an instance. Some alternative approaches could be:

~~1. Implement some form of optimized iteration through all the instances~~
2. Only add instances when viewing instance lists and show an error when trying to open an instance without a file on disk
~~3. Add a button in the app for performing this sync so it becomes on demand rather than automatic~~

lognaturel:
#3648 is related. I assume it would just incidentally be fixed because it doesn't make sense to look for `instanceId` anywhere other than where the spec says it should be.

seadowg:
@lognaturel coming back to this now, I'm thinking we go with option 2 (only add new instances on disk when syncing and make sure the app doesn't crash when opening instances with deleted files) and adding analytics to see if this ever actually happens. What do you think?

lognaturel:
Sounds good to me. 👍

seadowg:
@lognaturel great! I'll update acceptance. I might make it that the sync happens in one place (main menu for instance) so that this logic is a little more contained, but I'll play around with it and work out what's easiest.

seadowg:
~~@lognaturel do you think we still need to support adding encrypted instances via disk sync?~~

We already added analytics to track this, so clearly we are going to see!

PR #4521, opened 2021-04-20
Update form and instance lists after changes

### LINKED ISSUE #4516: Problems with form lists updating when something changes
Collect master

#### Problem description
Problems with form lists updating when something changes

Successfully sent submission does not disappear from `Send Finalized Form` list. You need to leave the view.

Submission state is not updated on `Send Finalized Form` list

Deleted forms are also visible on `Delete Saved Form` list.

#### Other information 
Related to #4439

seadowg:
Closes #4516

Ran (and added to) tests and verified manually.

I thought about expanding the `InstancesCountRepository` to handle this kind of data update but realized, after thinking about the ideal implementation, that my solution of using `updateAdapter` was probably enough: If these screens were implemented now we'd probably be using paginated `LiveData` exposed from a `ViewModel` for these lists. The change code (delete, send) would probably run in a coroutine inside the view model and we'd refresh the list of forms/instances after the work completed. We could use some kind of global app state but the state and change to that state are all contained in one screen, so there's not really any point.

I think my main concern is that there might be other cases like this that we've missed. It'd be good to check that the problem is fixed and also to explore if there are any other similar problems.


lognaturel (approved):

kkrawczyk123:
@seadowg 
- Progress bar is still running infinitely on all views except Fill Blank Form. I have noticed that it disappears after screen rotation.
- When send submission and cancel in right away - status is not updated. I have compared this behavior with the store version and it worked well there. see gif:
ezgif com-video-to-gif (56)

seadowg:
@kkrawczyk123 could you file a separate issue for the progress bar? I think I'll have to look at fixes in a pretty different area of the code. I can look at the quick cancel problem here though!

kkrawczyk123:
Sure! Done in #4525

seadowg:
@kkrawczyk123 are you able to generate a failed state when cancelling every time? I can't seem to reproduce and as far as I can see you'd need to cancel at the exact moment that a file is being uploaded. I'm going to try and write a test that reproduces but it would be good to know if you've got any tricks for making this happen.

kkrawczyk123:
@seadowg Actually, I am having it each time. If you want to slow things down you can try to add some big attachment for one of the media widgets or try to send more submissions by one it should help significantly.

seadowg:
> @seadowg Actually, I am having it each time. If you want to slow things down you can try to add some big attachment for one of the media widgets or try to send more submissions by one it should help significantly.

Good call on the big attachment. That makes sense.

seadowg:
So as far as I can see we have an accidental feature here - albeit a good one I think. Basically if a user cancels a submission while a request is in progress, the filled form will be marked as "Sending failed". This happens because the code guards against **any** failures during a request (it uses a `catch (Exception...`) and the cancellation interrupts running threads (`true` is passed to `AsyncTask#cancel`) but I don't think it's intentional. I think there is also another theoretical case where the user cancels, but the submission successfully uploads, and they then return to the "Send Finalized Form" screen where the filled form will still look like it was never sent.

I mainly needed to write that down for myself but given all that, I think we've really got two features of the app that have regressed in some way:

* When I cancel a submission and only part of the files were uploaded the submission should be flagged as "sending failed"
* When I cancel a submission and return to "Send Finalized Forms" then filled forms in the list should have up to date statuses (or be gone if they sent successfully). 

My assumption is that we want both these features, but I want to double-check that with @lognaturel. If that is the case I think we want to break this out to a separate issue, so we can have explicit acceptance (and get some testing in for it).

lognaturel:
> When I cancel a submission and only part of the files were uploaded the submission should be flagged as "sending failed"

Yes. This is still happening but if there really isn't a test for it already it would be good to add. I wouldn't call it unintentional. Perhaps lazy because it would be nice to have an explicit cancelation state but it's definitely expected and desired that there's a state change when a submission is attempted and canceled.

> When I cancel a submission and return to "Send Finalized Forms" then filled forms in the list should have up to date statuses (or be gone if they sent successfully).

Isn't it just "filled in forms in the list should have up to date status"? We can easily reproduce the lack of status update with a cancel but wouldn't it also be the case with a failure at the moment?

> another theoretical case where the user cancels, but the submission successfully uploads, and they then return to the "Send Finalized Form" screen where the filled form will still look like it was never sent.

The only case in which I know this can happen is when a submission must be split into parts and the cancel happens after one or more parts have been received by the server. We've talked about it and it's a truly terrible state. Fine if the user just resends immediately but if the user opens the filled form for edit, that will update the edit timestamp and the remaining part(s) will be rejected by the server.

Is there another case you can think of? Is there a timing issue because the cancelation is only checked for at certain points in the sending?

seadowg:
> Yes. This is still happening but if there really isn't a test for it already it would be good to add. I wouldn't call it unintentional. Perhaps lazy because it would be nice to have an explicit cancelation state but it's definitely expected and desired that there's a state change when a submission is attempted and canceled.

I just mean that the code doesn't explicitly handle this scenario, but I agree we probably always wanted it there. And yes it needs a test!

> Isn't it just "filled in forms in the list should have up to date status"? We can easily reproduce the lack of status update with a cancel but wouldn't it also be the case with a failure at the moment?

Success/failure and cancel are different because of the implementation: when you cancel we return to the list of filled forms immediately rather whereas in a success/failure case the task would be finished when we return. A lot of the complexity is wrapped up in it being an `AsyncTask`.

> Is there another case you can think of? Is there a timing issue because the cancelation is only checked for at certain points in the sending?

I think so? As far as I can tell you can cancel in one of three "states": before the upload starts, during an upload and after success. If you cancel during an upload it will either run to the end of `doInBackground` in the AsyncTask (and then not run `onPostExecute`) or it will throw an `InterruptedException` while a request is being made (which is what results in the failed state being marked).

I'm going to try and write some tests today to verify all this as it's confusing 🤯 and we can maybe have a chat about it later.

seadowg:
@kkrawczyk123 the issue should be fixed now!

collect_app/src/main/java/org/odk/collect/android/formmanagement/InstancesAppState.kt
```diff
@@ -1,28 +1,29 @@
...
 /**
- * Stores reactive state of current instances count with various different statuses. This (as a
- * singleton) can be read or updated by different parts of the app without needing reactive data
- * in the [InstancesRepository].
+ * Stores reactive state of instances. This (as a singleton) can be read or updated by
+ * different parts of the app without needing reactive data in the [InstancesRepository].
  */
 @Singleton
-class InstancesCountRepository {
+class InstancesAppState(private val context: Context) {
```
seadowg, line 17: Trying this naming on for size. In my head there is potentially a pattern where all our interactions from "Android" code go through state objects like this (you could imagine this could have a `instanceSubmitted` or a `finalizeInstance` method). That would basically give us a Redux-esque state system where these app state objects could expose both the state and the event/reducers. But you know, lets just hold off on reinventing that wheel for the moment.

collect_app/src/main/java/org/odk/collect/android/upload/InstanceUploader.java
```diff
@@ -17,13 +17,30 @@
...
 public abstract class InstanceUploader {
+
+    @Inject
+    InstancesRepositoryProvider instancesRepositoryProvider;
+
+    @Inject
+    InstancesAppState instancesAppState;
+
+    public InstanceUploader() {
+        DaggerUtils.getComponent(Collect.getInstance()).inject(this);
```
seadowg, line 41: Look, grabbing the Collect singleton and using for injection in an object like this is bad, and I'm a bad person. We've done some much restructuring in the last few weeks though and starting to rework the submission/send class hierarchy doesn't feel like a great idea at the moment.

collect_app/src/main/java/org/odk/collect/android/formmanagement/InstancesAppState.kt
lognaturel: Like it.

collect_app/src/main/java/org/odk/collect/android/upload/InstanceUploader.java
lognaturel: tumblr_24aa1561c944059f71984f454aa3af8b_12ca05de_250

lognaturel (approved):

mmarciniak90:
When I re-tested this PR I noticed one more case where the list is not updated when something changes.

1. `Exactly match server` is set with 15 minutes update frequency
2. User opens `Fill Blank Form` view and sees all forms
3. Server administrator unselects some form or adds a new form for this user
4. User waits 15 minutes on `Fill Blank Form` view
5. **Form list is not updated**. User needs to leave view and opens it again. 

This issue is not visible on the released 1.30.1 version but is visible on master like the main issue.
@seadowg should I report a new issue for this or do you prefer to look at this as a part of this PR?

seadowg:
@mmarciniak90 nice find. I think that can be fixed as part of this PR.

seadowg:
@mmarciniak90 that should be fixed now. I've also made sure form lists update after auto update runs as well.

mmarciniak90:
> that should be fixed now.

@seadowg, I'm worried it isn't 😞 @kkrawczyk123 also still sees this issue as well as me.

seadowg:
@mmarciniak90 oh sorry I think I know what I've done wrong. My bad.

seadowg:
@mmarciniak90 ok pretty sure I've actually fixed it this time 🙈

mmarciniak90:
You're right. This case is fixed now but we need to continue retesting.

kkrawczyk123:
@seadowg I have noticed today that instances copied manually to instance folder are not loaded at all on Edit saved form, send finalized form and on Delete Saved form. In that way we test sending 1000 submissions each release so it used to work.
ezgif com-gif-maker
That one submission visible on each view has been filled manually, I also waited for quite a long time to see if submissions will show up - they didin't. Could you please take a look?

seadowg:
@kkrawczyk123 I'm pretty sure this is covered here: https://github.com/getodk/collect/issues/4474

mmarciniak90:
**Tested with success!!**

Verified on Android 7.0, 8.1, 10.0, 11.0

**Verified cases:**
- _Fill Blank Form list_
   - auto-download forms when match exactly is selected
   - auto-update form version
- _Send Finalized Form list_
   - cancel sending and failure status
   - unsuccessful sending and failure status
   - successful sending and success status
   - sent form disappeared
   - displaying sent / unsent forms on the list
- _View Sent Form list_
   - auto-sent submissions appear when the view is opened all the time
   - auto-deleted submissions appear when the view is opened all the time
- _Delete Saved Form list_
   - deleted forms
   - deleted submissions
- sorting forms/submissions
- searching
- Select all/Clear all buttons
- form with defined auto send and auto-delete in XML
- encrypted forms
- light/dark mode
- English and Spanish language
- RTL

kkrawczyk123:

PR #4522, opened 2021-04-20
Changes to projects tests

seadowg:
This was motivated by discussions about the Espresso tests for projects. I'll leave comments inline to add additional detail to the changes.

Ran tests.

No QA needed as it's just test changes really.


collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/SwitchProjectTest.kt
```diff
@@ -1,59 +1,45 @@
...
     val rule = CollectTestRule()
 
     @get:Rule
     var chain: RuleChain = TestRuleChain
         .chain()
         .around(rule)
 
-    @Before
-    fun setup() {
-        val projectsRepository = DaggerUtils.getComponent(Collect.getInstance()).projectsRepository()
```
seadowg, line 31: I could have sworn I wrote a comment about this in a review, but apparently I didn't. My fault! 

Although it can speed up tests I strongly feel it's best to avoid interacting with our "data" layer in app level tests. There are a couple of reasons for this:

* It bypasses the "real" way state/data is set up when a user is interacting with the app. This means the thing you're testing could potentially only work because the code in your test setup is correct, whereas the flow may be broken in real life.
* It means the tests have to change if there are changes to the domain objects, repositories etc. This can look a little confusing in history, but more importantly means that someone changing the test would then have to have knowledge of how to update the setup code, so it creates a realistic simulation of the user flow.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ProjectSettingsDialogPage.kt
```diff
@@ -42,12 +46,12 @@ internal class ProjectSettingsDialogPage(rule: ActivityTestRule<*>) : Page<Proje
     }
 
     fun assertCurrentProject(name: String, icon: String): ProjectSettingsDialogPage {
-        onView(withId(R.id.current_project)).check(matches(withProject(name, icon)))
+        onView(allOf(withChild(withText(name)), withChild(withText(icon)), withContentDescription(getTranslatedString(R.string.using_project, name)))).check(matches(isDisplayed()))
```
seadowg: This of course isn't as exact as the previous `ProjectMatcher` as it checks that the view contains the two pieces of information we expect, but not that they are in the position we want them to be. I feel like the latter is better suited to a view test however. The next commit will deal with that.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectSettingsDialog.kt
```diff
@@ -109,6 +109,7 @@ class ProjectSettingsDialog : DialogFragment() {
             }
             projectView.findViewById<TextView>(R.id.project_name).text = project.name
 
+            projectView.contentDescription = getString(R.string.switch_to_project, project.name)
```
seadowg, line 107: This might feel a little strange as we're adding data to the view that's only used in tests. It can also be used by screen readers (which is really how we interact with the app in our tests interestingly). Adding this data feels like a win-win as it helps accessibility tools and lets us write simpler tests.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectListItemViewTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.odk.collect.android.R
+
+@RunWith(AndroidJUnit4::class)
+class ProjectListItemViewTest {
```
seadowg, line 17: I'm a big fan of this pattern of pulling out "presenter" style views that hold all the knowledge of how to display a domain object. It lets you keep presentation logic for that data out of places like the Activity/Fragment, allows you to write super precise tests at the view level while keeping fuzzy at the app level. It also means that as the minutia of the view itself changes we don't need to change our app level tests - they only need to be concerned with the headline items (I can select a project, I can create a project etc).

seadowg:
@lognaturel @grzesiek2010 tagged you both in on this as it's just a small set of changes, but I think good discussion points on testing that you might disagree with me on.

seadowg:
@grzesiek2010 this is probably a good time to sneak this in if you're happy with the changes. I've just rebased against master.

grzesiek2010:
> @grzesiek2010 this is probably a good time to sneak this in if you're happy with the changes. I've just rebased against master.

yeah, I was just thinking about the same. I'll review it today.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectListItemView.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.graphics.Color
+import android.graphics.drawable.GradientDrawable
+import android.util.AttributeSet
+import android.widget.FrameLayout
+import android.widget.TextView
+import org.odk.collect.android.R
+import org.odk.collect.projects.Project
+import kotlin.properties.Delegates
+
+class ProjectListItemView(context: Context, attrs: AttributeSet?) : FrameLayout(context, attrs) {
```
grzesiek2010, line 13: Nice!

grzesiek2010 (approved):

PR #4527, opened 2021-04-21
Upgrades create "Default" project

### LINKED ISSUE #4418: Upgrades create "Default" project
Blocked by #4497 

**Acceptance**

- [x] Given I'm using an older version of Collect
When I upgrade to the new version
Then I see my current project has the name 'Default' and the icon 'D'
And I never see a page offering to create a new project or use the demo one

grzesiek2010:
Closes #4418

Importing the existing project is pretty clear and obvious but I need to explain other changes I added here:
- [b14a7bb](https://github.com/getodk/collect/commit/b14a7bb2a34431bb3ba6ed552afb8bdb11516dab) it's just an improvement. I introduced a rule that we will be able to reuse in the future as recommended in https://developer.android.com/codelabs/kotlin-coroutines#5
- [16c2445](https://github.com/getodk/collect/commit/16c2445b98c1a3e4b4818b3e373798d3b1b45091) - I realized that I made a mistake my previous approach with sharedprefs to detect first launch was faulty because it retuned true for upgrades since the preference I used didn't exist in v1.30.
- [8eb2ef5](https://github.com/getodk/collect/commit/8eb2ef50668f9056d57d7ef7349e1e400b12141d) here I fixed managing preferences. I realized that the demo project should have empty string uuid so that we can by default use preferences that existed in v1.30. For example for admin preferences values were saved using `admin_prefs` key, with current implementation if we had two project: the default one and projectX (with uuid 1234) we can access those prefs just adding their uuid to `admin_prefs`:
  - for the demo project it would be `admin_prefs` + `"" `= just `admin_prefs` - existing admin prefs will be used
  - for the projectX it would be `admin_prefs` + `1234` = `admin_prefs1234`
   (see SettingsProvider class)

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
@@ -31,14 +31,14 @@ class SplashScreenViewModel(
     val doesLogoFileExist
         get() = splashScreenLogoFile.exists()
 
-    fun shouldFirstLaunchDialogBeDisplayed(): Boolean {
+    fun isFirstLaunch(): Boolean {
         val isFirstLaunch = !metaSettings.contains(MetaKeys.KEY_FIRST_LAUNCH)
         metaSettings.save(MetaKeys.KEY_FIRST_LAUNCH, false)
         return isFirstLaunch
     }
 
     fun importExistingProjectIfNeeded() {
-        if (!shouldFirstLaunchDialogBeDisplayed() && projectsRepository.getAll().isEmpty()) {
+        if (!isFirstLaunch() && projectsRepository.getAll().isEmpty()) {
             projectImporter.importExistingProject()
         }
     }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
@@ -5,14 +5,18 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import org.odk.collect.android.preferences.keys.GeneralKeys
 import org.odk.collect.android.preferences.keys.MetaKeys
+import org.odk.collect.android.projects.ProjectImporter
 import org.odk.collect.android.utilities.FileUtils
 import org.odk.collect.android.utilities.ScreenUtils
+import org.odk.collect.projects.ProjectsRepository
 import org.odk.collect.shared.Settings
 import java.io.File
 
 class SplashScreenViewModel(
     private val generalSettings: Settings,
     private val metaSettings: Settings,
+    private val projectsRepository: ProjectsRepository,
+    private val projectImporter: ProjectImporter
 ) : ViewModel() {
 
     val shouldDisplaySplashScreen
@@ -33,9 +37,20 @@ class SplashScreenViewModel(
         return isFirstLaunch
     }
 
-    open class Factory constructor(private val generalSettings: Settings, private val metaSettings: Settings) : ViewModelProvider.Factory {
+    fun importExistingProjectIfNeeded() {
+        if (!shouldFirstLaunchDialogBeDisplayed() && projectsRepository.getAll().isEmpty()) {
+            projectImporter.importExistingProject()
+        }
+    }
+
+    open class Factory constructor(
+        private val generalSettings: Settings,
+        private val metaSettings: Settings,
+        private val projectsRepository: ProjectsRepository,
+        private val projectImporter: ProjectImporter
+    ) : ViewModelProvider.Factory {
         override fun <T : ViewModel?> create(modelClass: Class<T>): T {
-            return SplashScreenViewModel(generalSettings, metaSettings) as T
+            return SplashScreenViewModel(generalSettings, metaSettings, projectsRepository, projectImporter) as T
         }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
@@ -4,8 +4,8 @@ import android.graphics.Bitmap
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import org.odk.collect.android.preferences.keys.GeneralKeys
-import org.odk.collect.android.preferences.keys.MetaKeys
 import org.odk.collect.android.projects.ProjectImporter
+import org.odk.collect.android.utilities.AppStateProvider
 import org.odk.collect.android.utilities.FileUtils
 import org.odk.collect.android.utilities.ScreenUtils
 import org.odk.collect.projects.ProjectsRepository
@@ -16,7 +16,8 @@ class SplashScreenViewModel(
     private val generalSettings: Settings,
     private val metaSettings: Settings,
     private val projectsRepository: ProjectsRepository,
-    private val projectImporter: ProjectImporter
+    private val projectImporter: ProjectImporter,
+    private val appStateProvider: AppStateProvider
 ) : ViewModel() {
 
     val shouldDisplaySplashScreen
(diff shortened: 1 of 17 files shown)


seadowg:
@grzesiek2010 looks like this will need a rebase

grzesiek2010:
>@grzesiek2010 looks like this will need a rebase

It's fixed.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
```diff
@@ -27,15 +31,23 @@ class SplashScreenViewModel(
...
         get() = splashScreenLogoFile.exists()
 
-    fun shouldFirstLaunchDialogBeDisplayed(): Boolean {
-        val isFirstLaunch = !metaSettings.contains(MetaKeys.KEY_FIRST_LAUNCH)
-        metaSettings.save(MetaKeys.KEY_FIRST_LAUNCH, false)
-        return isFirstLaunch
+    val isFirstLaunch
+        get() = appStateProvider.isFreshInstall()
+
+    fun importExistingProjectIfNeeded() {
```
seadowg: Would it make sense to have this happen in the `ApplicaitonInitializer` instead?

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
```diff
@@ -13,7 +15,7 @@ class SettingsProvider(private val context: Context) {
     }
 
     @JvmOverloads
-    fun getGeneralSettings(projectId: String = ""): Settings {
+    fun getGeneralSettings(projectId: String = NOT_SPECIFIED_UUID): Settings {
```
seadowg, line 18: Generally for optional params like this I'd use `null` rather than a constant. Is there any reason not to do that here?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
```diff
@@ -15,11 +15,19 @@ class ProjectImporter(
...
+        metaSettings.save(MetaKeys.CURRENT_PROJECT_ID, DEMO_PROJECT_ID)
     }
 
     companion object {
-        const val DEMO_PROJECT_ID = "DEMO"
+        /*
+        Should be empty to easily access existed settings (general and admin) that had be saved in versions
+        prior to the one that implemented "Projects" and treat them as those which belong to
+        the imported existed project.
+         */
```
seadowg, line 30: Ah I'd thought about this differently. In my head all projects would have a prefix (including the demo) now that we always have a current project. We'd need to migrate the settings (copy them) but that seems simpler to me long term, but I could be wrong?

collect_app/src/test/java/org/odk/collect/android/rules/MainCoroutineScopeRule.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.TestCoroutineScope
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@ExperimentalCoroutinesApi
+class MainCoroutineScopeRule(private val dispatcher: TestCoroutineDispatcher = TestCoroutineDispatcher()) :
```
seadowg, line 13: Yeah this is much nicer!

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
grzesiek2010: That's what I wanted to avoid - migrating preferences. I agree in the long term that would be clearer but I thought we agreed on the existing approach. Let's maybe create a separate issue for this one and decide?

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
grzesiek2010: I would need to make uuid in Project class nullable what I don't like because if I read projects from the database I shouldn't have null values since those ids are generated if not exist. So I would need to add some `!!` or `?.` in the code to handle it.
This also depends on what we decide about what you said below (whether to always use prefixes or not)

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModel.kt
grzesiek2010: Yeah, I agree. Fixed.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
seadowg: Hmmm maybe we did and I've just forgotten. It's already been a while. Agreed we can keep this for the PR and then we can discuss next week.

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
seadowg: Right I see. I think that's a signal that we shouldn't be using the same object both what a project looks like before it is "saved" and how it is stored. In my head there are two ways we could go that would avoid a constant for "not specified":

1. Make project an `abstract` with two inner subclasses: `New` and `Existing`. The former does not have an id field but the latter does. `ProjectsRepository#save` could take a `Project` (seeing as our repositories usually do insert and update from save) but return a `Projects.Existing`. `Project.Existing` objects would be stored using GSON.
2. Create a private class just for GSON in `SharedPreferencesProjectsRepository` that has a non null ID. That would let `Project` have `String?` id.

I've been thinking it would be a good idea to use approach 1 for the forms/instances repos as well so would probably lean towards that.

seadowg: The default param value here could still just be `null` though right? In `getSettingsId` you'd just do `if (projectId == null...`.

In terms of running into problems with `Project#uuid` being null it sounds like a sign that we probably shouldn't use the same class as the "domain" object as we do to store the project. I think there are two options which let us avoid a constant:

1. Use sub-types (or if you're fancy "algebraic data types") to represent the project before and after it's saved:

    ```kotlin
    abstract class Project {

        val name: String

        data class New(override val name: String) : Project
        data class Existing(val uuid: String, override val name: String) : Project
    }
    ```

    In this world `ProjectsRepository#save` could take a project (as we usually do insert and update from save in our other repos) and return `Projects.Existing`. You can also use a `sealed` class rather than just an `abstract` to get magic pattern matching etc: https://kotlinlang.org/docs/sealed-classes.html.

2. Use a private class just for GSON in `SharedPreferencesProjectsRepository` that has a non-null uuid.

While I think 2 is probably always a good idea (as it ensures your storage mechanism doesn't affect your domain object) it might be overkill here. I had been thinking we'd probably want to use the kind of approach laid out in 1 for forms and instances as we have a similar problem there with IDs being assigned at save time. What do you think?

seadowg (commented):
Replied to comments

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
grzesiek2010: What you said makes sense but I'm not totally convinced because:
- if we decide to migrate default preferences and always use prefixes we won't need that null and then both objects would be equal so I think it wouldn't make sense to keep both of them
- even with the current implementation the role of Project.New would be small since we build it just to pass to add() method in ProjectsRepository, we could either way pass just values and need only Project.Existing

given that I would vote for postponing this decision and make it after we decide what to do with default preferences. What do you think?

seadowg: > given that I would vote for postponing this decision and make it after we decide what to do with default preferences. What do you think?

Yeah fair. Let's get there and we can revise this.

seadowg (approved):

PR #4529, opened 2021-04-23
Display project icon in the main menu

### LINKED ISSUE #4423: User can see what project they're using
Blocked by #4497

**Acceptance**

- [x] Given I'm on the Main Menu
Then I see the icon for my project in the top right (instead of the ⚙ icon)

grzesiek2010:
Closes #4423

Screenshot_1619194051

No.

No.


lognaturel:
Turtles for everyone! 🐢

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
```diff
@@ -93,4 +94,9 @@ private void clearPrefs() {
         settingsProvider.getAdminSettings().setDefaultForAllSettingsWithoutValues();
         settingsProvider.getMetaSettings().clear();
     }
+
+    private void clearProjects() {
+        DaggerUtils.getComponent(InstrumentationRegistry.getInstrumentation().getTargetContext()).projectsRepository().deleteAll();
+        DaggerUtils.getComponent(InstrumentationRegistry.getInstrumentation().getTargetContext()).projectImporter().importDemoProject();
```
seadowg, line 100: Do we need the `deleteAll` given meta prefs is cleared or do you think it's better to be explicit here (and not assume the projects are stored in prefs)? I'm guessing we need to do the import again because `ApplicationInitializer` won't run? Maybe we need to think about running it here. That's an improvement we can make down the road though.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -251,4 +263,10 @@ public void onCorrectAdminPassword(Action action) {
     public void onIncorrectAdminPassword() {
         ToastUtils.showShortToast(R.string.admin_password_incorrect);
     }
+
+    @Override
+    public void onProjectSwitched() {
+        invalidateOptionsMenu();
```
seadowg: This might be a nice opportunity to use `LiveData` instead of listeners. This would have the advantage of having less calls for `ProjectSettingsDialog` to make when a project is selected. `ProjectSettingsDialog` and `MainMenuActivity` could both share a `CurrentProjectViewModel`:

```kotlin
class CurrentProjectViewModel {

    val currentProject: LiveData<Project>...

    fun selectProject(project: Project)...
}
```

Then in `MainMenuActivity` you can do:

```kotlin
currentProjectViewModel.getCurrentProject().observe(this) { project ->
    this.currentProject = project
    invalidateOptionsMenu()
}
```

What do you think?

collect_app/src/main/res/layout/project_icon_view.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/project_icon_text"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:background="@drawable/project_icon_background"
+    android:textAppearance="@style/TextAppearance.Collect.Subtitle1"
+    android:layout_marginVertical="@dimen/margin_small"
+    android:textColor="@android:color/white"
```
seadowg, line 9: We're going to have to revise this for different colors I think but that'll be a tweak best left ot the end.

seadowg (changes requested):
Some comments around a slightly different structure that would be good to discuss.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: I was thinking about using live data but decided that it doesn't make much sense here.
1. In `onPrepareOptionsMenu()` I would take the current project from `MainMenuViewModel` but then when it's switched from `CurrentProjectViewModel` what makes it a bit more complex.
2.  There is just one event that triggers this action I think livedata is more useful in more dynamic cases
3. It requires more code but doesn't make it much better in my opinion.

If you are strongly in favor of livedata we can add it to the list of improvements and discuss it later.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
grzesiek2010: Right, `deleteAll()` is not needed here so I removed it. When it comes to `ApplicationInitializer` it is called here but it imports the default project only if it's not fresh install but in tests, we may have different scenarios so I added it here to make sure it works well. I could also override `AppStateProvider` to return false always for `isFreshInstall()` but it might be bad for other tests that we want to add in the future.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
seadowg: > In onPrepareOptionsMenu() I would take the current project from MainMenuViewModel but then when it's switched from CurrentProjectViewModel what makes it a bit more complex.

I think it would be pretty much the same right? You would just do `this.currentProject` instead of `viewModel.getCurrentProject()` in the `onPrepareOptionsMenu`.

> There is just one event that triggers this action I think livedata is more useful in more dynamic cases.

Agreed that it's most useful in those cases. Using reactive values like `LiveData` is also beneficial when you have multiple things that deal with those values though. For example, if we wanted the title bar to display the project name, and we went with the current structure we'd need to change the title in two places (initial setup and when the value changes) whereas with `LiveData` we can just have one place (`observe`) that declares what happens with the current project. The way menus work in Android does take some of that away from us though (having to call `invalidateOptionsMenu`).

> It requires more code but doesn't make it much better in my opinion.

I'm not sure if I agree. Right now we need an interface for the listener, the code to set the listener up and an implementation of that listener which could all disappear if we used a view model. Yes we'd have the extra `ViewModel` but its code would be coming from `ProjectSettingsDialog` and `MainMenuViewModel`.

I do see how it's easy to argue "well we make that change when we add something else" but I think the `ViewModel`/`LiveData` pattern is pretty standard and given we're writing feature tests for this work we have the luxury of being able to refactor our "green"/working solutions as we go to anticipate changes. Personally, I'd advocate making this kind of change now rather than waiting, but it's hard to argue that that is definitively a better idea, so I'll let you make the call here.

grzesiek2010: Ok I implemented it using LiveData.
There is one problem I ran into: https://github.com/getodk/collect/pull/4529/commits/3478dacfb6edeb729c1ca1db868eac248e192fd7#diff-8d4bf37be6c4278258a2d1623c25136611a9d483689a18b2a7eb7ca91e667095R16

this is caused as I explained because of the fact that ResetStateRule is called too late. I know you wanted to look into tests and add some fixes so I handled it like that for now but it should be fixed soon.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -196,18 +206,37 @@ public void onClick(View v) {
...
+    @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        final MenuItem projectsMenuItem = menu.findItem(R.id.projects);
+
+        ProjectIconView projectIconView = (ProjectIconView) projectsMenuItem.getActionView();
+        projectIconView.setProject(currentProjectViewModel.getCurrentProject().getValue().getValue());
+        projectIconView.setOnClickListener(v -> onOptionsItemSelected(projectsMenuItem));
+
+        currentProjectViewModel.getCurrentProject().getValue().consume();
+        currentProjectViewModel.getCurrentProject().observe(this, project -> {
```
seadowg: The `observe` can just happen in `onCreate` I think. The current project could either be stored in a field in the Activity and then used in `onPrepareOptionsMenu` (as it gets called after `invalidateOptionsMenu`), or you could just call `currentProjectViewModel.getCurrentProject().getValue()` in `onPrepareOptionsMenu`.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -196,18 +206,37 @@ public void onClick(View v) {
...
+            if (!project.isConsumed()) {
+                invalidateOptionsMenu();
+                ToastUtils.showLongToast(getString(R.string.switched_project, project.getValue().getName()));
```
seadowg: I think we can avoid using `Consumable` here by just moving this toast call in `ProjectSettingsDialog` right? It seems like the `Consumable` ends up being awkward as you have to call `consume` to deal with the "initial" value.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import org.odk.collect.android.projects.CurrentProjectProvider
+import org.odk.collect.androidshared.data.Consumable
+import org.odk.collect.projects.Project
+
+class CurrentProjectViewModel(currentProjectProvider: CurrentProjectProvider) : ViewModel() {
+
+    private val _currentProject = MutableLiveData(Consumable(currentProjectProvider.getCurrentProject()!!))
```
seadowg: I think we should probably make `getCurrentProject` return `String` rather than `String?`. We really don't expect it to be null. I think a check and throw in `getCurrentProject` (or just a `!!` in there) would be nice.

collect_app/src/main/java/org/odk/collect/android/projects/CurrentProjectProvider.kt
```diff
@@ -13,6 +13,11 @@ class CurrentProjectProvider(private val settingsProvider: SettingsProvider, pri
     }
 
     fun getCurrentProject(): Project? {
+        // This should be removed. It was added just to fix tests because ResetStateRule which should be responsible for it is not always called first
+        if (projectsRepository.getAll().isEmpty()) {
```
seadowg, line 17: I'll play around with this in my test tweaks to see if I can get rid of it 👍

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModelTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    fun `Initial current project should be set`() {
+        assertThat(currentProjectViewModel.currentProject.value!!.value, `is`(Project("Project X", "X", "#cccccc")))
+    }
+
+    @Test
+    fun `setCurrentProject() should add new project to live data`() {
+        val project = Project("Project Y", "Y", "#ffffff")
+
+        currentProjectViewModel.setCurrentProject(project)
+        assertThat(getOrAwaitValue(currentProjectViewModel.currentProject).value, `is`(project))
```
seadowg: I don't think you need `getOrAwaitValue` here. I gave it a try and the test worked without it. The `InstanceTaskExecutor` or just using Robolectric should take care of the weirdness in `LiveData` (although the former is better as the test is faster).

seadowg (changes requested):
Thanks so much for making the switch to a reactive value from a listener for the current project. I know I'm probably being [pernickety](https://www.google.com/search?q=pernickety+definition), but I feel heard, and I like where the code is going for projects in general a lot.

There's a couple of last tweaks I think we can make to simplify our lives a little but nothing major.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: > or you could just call currentProjectViewModel.getCurrentProject().getValue() in onPrepareOptionsMenu

I do that. The reason why I put it here is to avoid performing what's between those brackets when the activity is created. Once it's called to setup the icon (a few lines above) it's marked as consumed and not used anymore.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: but still `invalidateOptionsMenu();` would be called for no reason. Not harmful but...

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
seadowg: Yeah sorry this and the comment on `Consumable` should be one really as they are very related (as you point out).

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModel.kt
grzesiek2010: It returns `Project?` not `String?` but I agree this shouldn't return null. I just had problems fixing it now because of those tests but I'll do that as a part of the next pr.

seadowg: Ach sorry I'm getting distracted by the `?` and missing the important part there.

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModelTest.kt
grzesiek2010: Are you sure? I've just tried and it doesn't work without that utility method

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
seadowg: Personally that feels fine to me. We've done that in other places (see `BlankFormListMenuDelegate` for example). If we're worried about a double render we could skip calling it the first time, but it doesn't feel necessary to me.

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModelTest.kt
seadowg: Yeah if I remove `getOrAwait` so it's just:

```kotlin
assertThat(currentProjectViewModel.currentProject.value!!.value, `is`(project))
```

it works.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/CurrentProjectViewModelTest.kt
grzesiek2010: Hmm magic I tried the same and it failed now it works... nevermind. Fixed.

seadowg (approved):

PR #4530, opened 2021-04-26
Make sure instance update time changes when form status changes

### LINKED ISSUE #4511: The time of finalization a form is not updated after the form edition
Collect master
Collect v1.30.1 works correctly 

#### Problem description
The time of finalization a form is not updated after the form edition

1. User fills some form and finalizes it - pay attention to the finalization time
2. User waits few minutes and edits finalized form
3. User sees that finalization time isn't updated and the progress bar is visible all the time (also on Send Finalized Form, View Send Form and Delete Saved Form)

#### Expected behavior
The latest finalization time is visible, it is updated after each saved edition.

seadowg:
Closes #4511

Ran (corrected) tests and verified manually. Also added some new tests to the repository to account for changes there.

I guess the only thing to point out here is that it's still important to add tests to the repository as well as the provider. This is because it's likely we'll kill the mutation methods on the provider but still want the same behaviour in the repo.

Should just fix the issue in question.

### DIFF
--- a/formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
+++ b/formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
@@ -244,7 +244,7 @@ public abstract class InstancesRepositoryTest {
     }
 
     @Test
-    public void save_whenInstanceSoftDeleted_doesNotUpdateLastChangesStatusDate() {
+    public void save_whenInstanceHasDeletedDate_doesNotUpdateLastChangesStatusDate() {
         Supplier<Long> clock = mock(Supplier.class);
         when(clock.get()).thenReturn(123L);
 
@@ -254,23 +254,23 @@ public abstract class InstancesRepositoryTest {
         Long originalInstanceDbId = originalInstance.getDbId();
 
         when(clock.get()).thenReturn(456L);
-        instancesRepository.softDelete(originalInstanceDbId);
+        instancesRepository.deleteWithLogging(originalInstanceDbId);
         instancesRepository.save(instancesRepository.get(originalInstanceDbId));
 
         assertThat(instancesRepository.get(originalInstanceDbId).getLastStatusChangeDate(), is(123L));
     }
 
     @Test
-    public void softDelete_setsDeletedDate() {
+    public void deleteWithLogging_setsDeletedDate() {
         InstancesRepository instancesRepository = buildSubject();
         Instance instance = instancesRepository.save(InstanceUtils.buildInstance("formid", "1", getInstancesDir()).build());
 
-        instancesRepository.softDelete(instance.getDbId());
+        instancesRepository.deleteWithLogging(instance.getDbId());
         assertThat(instancesRepository.get(instance.getDbId()).getDeletedDate(), is(notNullValue()));
     }
 
     @Test
-    public void softDelete_deletesInstanceDir() {
+    public void deleteWithLogging_deletesInstanceDir() {
         InstancesRepository instancesRepository = buildSubject();
         Instance instance = instancesRepository.save(InstanceUtils.buildInstance("formid", "1", getInstancesDir()).build());
 
@@ -278,19 +278,19 @@ public abstract class InstancesRepositoryTest {
         assertThat(instanceDir.exists(), is(true));
         assertThat(instanceDir.isDirectory(), is(true));
 
-        instancesRepository.softDelete(instance.getDbId());
+        instancesRepository.deleteWithLogging(instance.getDbId());
         assertThat(instanceDir.exists(), is(false));
     }
 
     @Test
-    public void softDelete_clearsGeometryData() {
+    public void deleteWithLogging_clearsGeometryData() {
         InstancesRepository instancesRepository = buildSubject();
         Instance instance = instancesRepository.save(InstanceUtils.buildInstance("formid", "1", getInstancesDir())
                 .geometry("blah")
                 .geometryType("blah")
                 .build());
 
-        instancesRepository.softDelete(instance.getDbId());
+        instancesRepository.deleteWithLogging(instance.getDbId());
         assertThat(instancesRepository.get(instance.getDbId()).getGeometry(), is(nullValue()));
         assertThat(instancesRepository.get(instance.getDbId()).getGeometryType(), is(nullValue()));
     }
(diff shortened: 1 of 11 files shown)


grzesiek2010 (approved):

kkrawczyk123:
@seadowg I have noticed an issue - on view sent form view when you delete already sent submission date of deletion is also visible in sent date even you sent it earlier. Both dates should differ. I have compared it with the store version and the issue does not occur there. 
Reproduction steps:
1. Fill the submission
2. Send it
3. Wait at least 1 minute
4. Delete submission
5. Open view sent form: see that sent date and delete date are the same but they shouldn't.

ezgif com-video-to-gif (58)


I have noticed one more interesting case which might not occur anymore after merging #4521 and this pr but I will write it down so you can confirm that it won't be visible. 
Reproduction steps:
1. Send finalized submission
2. Wait at least 1 minute
3. Change view to display send and unset forms
4. Send the same submission once again

I have compared 3 aps:
store version: time of sending is updated immediately
pr/4521: time of sending is not updated at all
this pr: time of sending does not update immediately but it is updated after leave and open view again 

What do you think? 

ezgif com-video-to-gif (57)

seadowg:
@kkrawczyk123 the second one should be fixed by #4521 I think. Let me look into the first problem here though.

seadowg:
The problem here is that we clear geometry data before doing the soft delete so the instances "last update" time is updated. I can take that code out and run instrumentation tests and nothing fails. @lognaturel @grzesiek2010 why do we clear geometry data when soft deleting?

As a side note I think there is a problem with when we update the "last update" on a instance. It's named `LAST_STATUS_CHANGE_DATE` but it gets updated when you edit a finalized instance and keep it finalized (so there is no state change) so currently it just updates whenever the instance is.

lognaturel:
> why do we clear geometry data when soft deleting

Because it’s sensitive. A major reason users delete filled forms is to clear out sensitive data from their devices and so it’s important that location goes.

But you say soft deleting instances? When does that happen? I thought only form definitions got soft deleted?

seadowg:
@lognaturel instances get soft deleted if they're deleted after being submitted (I guess so they can still appear in "View sent forms"). The sensitive information reason makes sense as we also delete the actual submission file. I guess in that case it does make sense to also delete the geometry data as part of `softDelete` on the repo.

seadowg:
@kkrawczyk123 ok that's fixed

lognaturel:
> @lognaturel instances get soft deleted if they're deleted after being submitted

I wouldn't call that soft delete but I see what you mean. It's more like everything is deleted except for the db record with non-sensitive data which acts like a kind of log. It's the responsibility of the project designer to make sure the `instanceName` balances usefulness and privacy for their scenario. The only user-visible things that should remain are instanceName and deleted date, I believe (probably createdAt, whether it's encrypted, form version and other bookkeeping things that aren't exposed).

formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
```diff
@@ -281,6 +282,19 @@ public void softDelete_deletesInstanceDir() {
         assertThat(instanceDir.exists(), is(false));
     }
 
+    @Test
+    public void softDelete_clearsGeometryData() {
```
lognaturel: I don't think we should use "soft delete" when talking about instances anywhere. There is only one kind of delete and it deletes everything other than metadata. Is that used anywhere else or is it just introduced here?

lognaturel: Ah, I see the tests above use it too. I really think it's misleading.

formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
```diff
@@ -281,6 +282,19 @@ public void softDelete_deletesInstanceDir() {
...
+        InstancesRepository instancesRepository = buildSubject();
+        Instance instance = instancesRepository.save(InstanceUtils.buildInstance("formid", "1", getInstancesDir())
+                .geometry("blah")
+                .geometryType("blah")
+                .build());
+
+        instancesRepository.softDelete(instance.getDbId());
```
lognaturel: Maybe this language got introduced when pulling things out of the `InstancesProvider`? Could be `deleteWithLogging` or something? Is the behavior that we only maintain the log if the instance was submitted and otherwise truly delete everything?

formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
seadowg: You can also hard delete the Instance if it's not submitted though (using `#delete`). I'm not sure if I can think of a better name for this but open to suggestions!

formstest/src/main/java/org/odk/collect/formstest/InstancesRepositoryTest.java
seadowg: Yeah. To me, it felt analogous to what we do with forms, but I can see the argument for that not being correct.

lognaturel: "Soft delete" to me means something very specific which implies that the thing that was soft deleted can be recovered. That's not the case here which is why the language feels pretty misleading to me, does that make sense?

Sorry, I commented as I was thinking so it's a little disjoint. Something like `deleteWithLogging` seems like it captures what we want.

lognaturel (approved):
Works for me! There are still a few uses of softDelete in test names that I would change but I think the intent is now clearer and that'd just be me being pedantic.

lognaturel:
> As a side note I think there is a problem with when we update the "last update" on a instance. It's named LAST_STATUS_CHANGE_DATE but it gets updated when you edit a finalized instance and keep it finalized (so there is no state change) so currently it just updates whenever the instance is.

All things around the "finalized"/"completed" state is full on legacy. I think there's a lot of rethinking we could do in this area but ideally we can do it as one project. For example, "mark as finalized" is really a weird concept to expose on the last form filling screen. Maybe it should be "send now"/"save as draft" and maybe "encrypt now"/"save as draft" for encrypted forms. Agreed the timestamp is weird. Etc.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1 and 11.

Tested cases:
- edit form
- few edits of the same submission 
- cancel edits 
- change status from finalized to saved 
- save and finalize
- sent form
- sent the same submission again
- sent and delete submission 
- display sent and unsent forms
- cancel sending and send again 
- edit saved form view
- send finalized form view
- view sent form view
- delete saved forms

mmarciniak90:
Tested with success also on Android versions 8.1 and 10.0.

PR #4534, opened 2021-04-29
Use larger machine to speed up Circle CI builds

yanokwa:
These changes (especially @seadowg's) reduce the build from 19m to about 10m. We can get more improvement if we run tests by package name (or by timing) so that the slowest part of the build (testDebug) can be run in parallel.

The change to gradle.properties might result in faster or slower local tests depending what the defaults everyone has on their machine, but I don't think that should be a blocker to merging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,15 +9,14 @@ references:
     docker:
       - image: circleci/android:api-30
     resource_class: xlarge
+    environment:
+      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-XX:+UseContainerSupport -XX:MaxRAMPercentage=90.0" -Dorg.gradle.parallel="true" -Dkotlin.compiler.execution.strategy=in-process'
 
 jobs:
   compile:
     <<: *android_config
-    environment:
-      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx12g" -Dorg.gradle.parallel="true" -Dkotlin.compiler.execution.strategy=in-process'
     steps:
       - checkout
-
       - run:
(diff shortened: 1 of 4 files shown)


seadowg:
@yanokwa just added some tweaks to make sure the Robolectric tests pick up on the RAM change as well. Circle CI actually has great docs on this stuff (which annoyingly Gradle doesn't). I primarily used [this](https://circleci.com/docs/2.0/java-oom/) and [this](https://support.circleci.com/hc/en-us/articles/360021812453) as guides.

seadowg:
One last thing this needs before merge: a default heap size for running locally (that will go in `gradle.properties`) as whatever Gradle gives me on my machine (which has 32GB of RAM) is too low to run lint. I need to double check that whatever we set will get overriden on Circle CI though. I can look at that tomorrow if no one else gets to it.

seadowg:
As far as I can see from the Gradle docs ([here](https://docs.gradle.org/current/userguide/build_environment.html)) it looks like whatever we put in `gradle.properties` will override `GRADLE_OPTS` annoyingly. 

It does look like we can override a project's `gradle.properties` with the "machine"'s `gradle.properties` (found at `~/.gradle/gradle.properties` on macOS and Linux) so I think we probably need to change our strategy to have a `gradle.properties` for CI (that sets a heap size appropriate for machines there) that we copy at the beginning of CI and a project `gradle.properties` that sets a sensible heap size for most people (1 or 2GB probably). This also means it's easy for us to override these settings if we want for our own machines.

seadowg:
Ok this is probably good to go if all checks pass.

PR #4535, opened 2021-04-29
User can change project details

### LINKED ISSUE #4429: User can change project details
Blocked by #4497 

**Acceptance**

When I click "Admin settings"
And I change the value of "Project Name"
And I change the value of "Project Icon"
And I navigate back to the Projects dialog
Then I see the updated versions of Name and Icon

grzesiek2010:
Closes #4429

I recommend going through the list of commits one by one.

Screenshot_1619693747

No.

No.

### DIFF
--- a/androidshared/src/main/java/org/odk/collect/androidshared/OneSignTextWatcher.kt
+++ b/androidshared/src/main/java/org/odk/collect/androidshared/OneSignTextWatcher.kt
@@ -1,0 +1,27 @@
+package org.odk.collect.androidshared
+
+import android.text.Editable
+import android.text.TextWatcher
+import android.widget.EditText
+
+class OneSignTextWatcher(private val editText: EditText) : TextWatcher {
+    lateinit var oldTextString: String
+
+    override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {
+        oldTextString = s.toString()
+    }
+
+    override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
+    }
+
+    override fun afterTextChanged(editable: Editable?) {
+        var newTextString = editable.toString()
+        if (oldTextString != newTextString) {
+            if (Character.codePointCount(newTextString, 0, newTextString.length) > 1) {
+                newTextString = oldTextString
+            }
+            editText.setText(newTextString)
+            editText.setSelection(newTextString.length)
+        }
+    }
+}
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/UpdateProjectTest.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/UpdateProjectTest.kt
@@ -1,0 +1,39 @@
+package org.odk.collect.android.feature.projects
+
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions.pressBack
+import androidx.test.espresso.matcher.ViewMatchers
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.RuleChain
+import org.odk.collect.android.support.CollectTestRule
+import org.odk.collect.android.support.TestRuleChain
+
+class UpdateProjectTest {
+
+    val rule = CollectTestRule()
+
+    @get:Rule
+    var chain: RuleChain = TestRuleChain
+        .chain()
+        .around(rule)
+
+    @Test
+    fun updateProjectTest() {
+        rule.mainMenu()
+            .assertProjectIcon("D", "#3e9fcc")
+            .openProjectSettingsDialog()
+            .assertCurrentProject("Demo project")
+            .clickAdminSettings()
+            .setProjectName("Project X")
+            .setProjectIcon("X")
+            .setProjectColor("#cccccc")
+
+        onView(ViewMatchers.isRoot()).perform(pressBack())
+
+        rule.mainMenu()
+            .assertProjectIcon("X", "#cccccc")
+            .openProjectSettingsDialog()
+            .assertCurrentProject("Project X")
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
@@ -63,10 +63,6 @@ public class AdminPreferencesFragment extends BaseAdminPreferencesFragment
     public void onAttach(@NonNull Context context) {
         super.onAttach(context);
         DaggerUtils.getComponent(context).inject(this);
-
-        settingsProvider.getAdminSettings().save(PROJECT_NAME_KEY, currentProjectProvider.getCurrentProject().getName());
-        settingsProvider.getAdminSettings().save(PROJECT_ICON_KEY, currentProjectProvider.getCurrentProject().getIcon());
-        settingsProvider.getAdminSettings().save(PROJECT_COLOR_KEY, currentProjectProvider.getCurrentProject().getColor());
     }
 
     @Override
@@ -81,13 +77,19 @@ public class AdminPreferencesFragment extends BaseAdminPreferencesFragment
         findPreference("user_settings").setOnPreferenceClickListener(this);
         findPreference("form_entry").setOnPreferenceClickListener(this);
 
+        findPreference(PROJECT_NAME_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_NAME_KEY, currentProjectProvider));
+        findPreference(PROJECT_ICON_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_ICON_KEY, currentProjectProvider));
+        findPreference(PROJECT_COLOR_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_COLOR_KEY, currentProjectProvider));
+
         findPreference(PROJECT_NAME_KEY).setOnPreferenceChangeListener(this);
         findPreference(PROJECT_ICON_KEY).setOnPreferenceChangeListener(this);
         findPreference(PROJECT_COLOR_KEY).setOnPreferenceChangeListener(this);
 
-        ((EditTextPreference) findPreference(PROJECT_ICON_KEY)).setOnBindEditTextListener(editText -> editText.addTextChangedListener(new OneSignTextWatcher(editText)));
+        ((EditTextPreference) findPreference(PROJECT_NAME_KEY)).setText(currentProjectProvider.getCurrentProject().getName());
+        ((EditTextPreference) findPreference(PROJECT_ICON_KEY)).setText(currentProjectProvider.getCurrentProject().getIcon());
+        ((EditTextPreference) findPreference(PROJECT_COLOR_KEY)).setText(currentProjectProvider.getCurrentProject().getColor());
 
-        setProjectColorSummary();
+        ((EditTextPreference) findPreference(PROJECT_ICON_KEY)).setOnBindEditTextListener(editText -> editText.addTextChangedListener(new OneSignTextWatcher(editText)));
     }
 
     @Override
@@ -116,11 +118,9 @@ public class AdminPreferencesFragment extends BaseAdminPreferencesFragment
                     intent.putExtra(INTENT_KEY_ADMIN_MODE, true);
                     startActivity(intent);
                     break;
-
                 case KEY_CHANGE_ADMIN_PASSWORD:
                     DialogUtils.showIfNotShowing(ChangeAdminPasswordDialog.class, getActivity().getSupportFragmentManager());
                     break;
-
                 case KEY_IMPORT_SETTINGS:
                     Intent pref = new Intent(getActivity(), QRCodeTabsActivity.class);
                     startActivity(pref);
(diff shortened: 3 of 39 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/UpdateProjectTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        rule.mainMenu()
+            .assertProjectIcon("D", "#3e9fcc")
+            .openProjectSettingsDialog()
+            .assertCurrentProject("Demo project")
+            .clickAdminSettings()
+            .setProjectName("Project X")
+            .setProjectIcon("X")
+            .setProjectColor("#cccccc")
+
+        onView(ViewMatchers.isRoot()).perform(pressBack())
```
grzesiek2010, line 32: I could use `Espresso.pressBack()` we also have the` pressBack()` method in our Page class but it doesn't work. I ran into the same problem again and spent some time investigating it turns out it doesn't work only in kotlin tests... strange but I don't know what the cause is.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -127,79 +170,15 @@ private void displayPreferences(Fragment fragment) {
...
     public void preventOtherWaysOfEditingForm() {
-        FormEntryAccessPreferences fragment = (FormEntryAccessPreferences) getFragmentManager().findFragmentById(R.id.preferences_fragment_container);
+        FormEntryAccessPreferencesFragment fragment = (FormEntryAccessPreferencesFragment) getFragmentManager().findFragmentById(R.id.preferences_fragment_container);
         fragment.preventOtherWaysOfEditingForm();
     }
+
+    private void setProjectColorSummary() {
+        Spannable summary = new SpannableString("■");
+        summary.setSpan(new ForegroundColorSpan(Color.parseColor(currentProjectProvider.getCurrentProject().getColor())), 0, summary.length(), 0);
+        findPreference(PROJECT_COLOR_KEY).setSummary("");
```
grzesiek2010: It might seem weird that I clear the summary first but it's a must. I'm not sure but I think that android uses some kind of caching and since the summary is always the same square just with different colors we need to clear it first.

collect_app/src/main/res/xml/admin_preferences.xml
```diff
@@ -18,6 +18,13 @@
         app:allowDividerAbove="false"
         app:allowDividerBelow="false"
         app:iconSpaceReserved="false">
+
+        <EditTextPreference
+            android:key="project_name"
```
seadowg, line 23: I'm thinking this should probably have `app:persistent="false"` (like we do with Device ID in `form_metadata_preferences.xml`).

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -51,6 +54,18 @@
...
+
+    @Inject
+    CurrentProjectProvider currentProjectProvider;
+
+    @Override
+    public void onAttach(@NonNull Context context) {
+        super.onAttach(context);
+        DaggerUtils.getComponent(context).inject(this);
+
+        settingsProvider.getAdminSettings().save("project_name", currentProjectProvider.getCurrentProject().getName());
```
seadowg: These could use a `Preference.SummaryProvider` to just display the values (from `CurrentProjectProvider`) I think? Again, `FormMetadataPreferenceFragment` is a good example of this.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -133,4 +139,10 @@ public void preventOtherWaysOfEditingForm() {
         FormEntryAccessPreferencesFragment fragment = (FormEntryAccessPreferencesFragment) getFragmentManager().findFragmentById(R.id.preferences_fragment_container);
         fragment.preventOtherWaysOfEditingForm();
     }
+
+    private void setProjectColorSummary() {
+        Spannable summary = new SpannableString("■");
+        summary.setSpan(new ForegroundColorSpan(Color.parseColor(currentProjectProvider.getCurrentProject().getColor())), 0, summary.length(), 0);
+        findPreference("project_color").setSummary(summary);
```
seadowg: This is very cool. Again, I think it could work with a `SummaryProvider` though.

projects/src/main/java/org/odk/collect/projects/ProjectsRepository.kt
```diff
@@ -7,6 +7,8 @@ interface ProjectsRepository {
 
     fun add(project: Project)
 
+    fun update(project: Project)
```
seadowg: For the other repositories we've used just a `save` and then either added or updated based on whether the data object has an ID or not. Could we follow that pattern here, so it's consistent? I don't think one or the other is better, but it'd more work to reverse that decision for forms/instances than changing this now I think.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -69,6 +74,22 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
         findPreference("user_settings").setOnPreferenceClickListener(this);
         findPreference("form_entry").setOnPreferenceClickListener(this);
 
+        findPreference("project_name").setOnPreferenceChangeListener((preference, newValue) -> {
+            Project currentProject = currentProjectProvider.getCurrentProject();
+            projectsRepository.update(new Project(currentProject.getName(), currentProject.getIcon(), currentProject.getColor(), currentProject.getUuid()));
+            return true;
```
seadowg: If we do go with a `SummaryProvider` and `persistent=false` these listeners could return `false` to stop the value from getting saved to shared prefs.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -76,17 +76,17 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
 
         findPreference("project_name").setOnPreferenceChangeListener((preference, newValue) -> {
             Project currentProject = currentProjectProvider.getCurrentProject();
```
seadowg: This can just use `copy` right?

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/UpdateProjectTest.kt
seadowg: Oh wow that sucks. Do you want to just change the implementation of `pressBack` on `Page` then so we can still chain with it?

androidshared/src/main/java/org/odk/collect/androidshared/OneSignTextWatcher.kt
```diff
@@ -0,0 +1,27 @@
+package org.odk.collect.androidshared
+
+import android.text.Editable
+import android.text.TextWatcher
+import android.widget.EditText
+
+class OneSignTextWatcher(private val editText: EditText) : TextWatcher {
```
seadowg, line 7: Hah as I was going through the commits I thought to myself "we could definitely extract a `TextWatcher` to share that logic" and then bam there it is. Great minds 🧠

seadowg (changes requested):
Mostly looking great! I'm really liking the breakdown in tests here: the Espresso test checks the core flow and teh effect the user would expect from it and the fragment test checks the details around where things are displayed etc (rather than just being on screen).

I've left a few comments mainly focused on using a different approach for displaying the prefs. I think using a [`SummaryProvider`](https://developer.android.com/reference/androidx/preference/Preference.SummaryProvider) would let us display these preferences without actually having to persist them (which I don't think we want to do).

There's also one comment about switching `add`/`update` to `save` on the projects repo.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/UpdateProjectTest.kt
grzesiek2010: I tried it but the same line that works here doesn't work if it's in java class (and called here from kotlin). Probably I would need to convert those page classes to kotlin but that's what I don't want to do as a part of this or.

There is something weird between kotlin and java here.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: I do this save not only to handle summary but also to display proper values in EditText when the dialog is opened since I didn't create custom dialogs and can't set that value programmatically. This is not something that I can handle using SummaryProvider right?

projects/src/main/java/org/odk/collect/projects/ProjectsRepository.kt
grzesiek2010: Ok. Done.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: Ok I did that but it's java, not kotlin (where I can update one specific parameter) so does that really matter?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
seadowg: I think you can use `OnBindEditTextListener` to customize the `EditText` for a setup with `SummaryProvider`. Or, I think if its `persistent` is `false` you can just do `setText` for the value you want on the Preference itself. My goal here is just to avoid writing anything to shared preferences as I'm worried that might become confusing.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
seadowg: Hah I'd become so used to reading Kotlin I didn't notice. Yeah doesn't make sense to `copy` then.

grzesiek2010: It's even less readable in java right? I think I should remove this change...

grzesiek2010: Ok I removed that change I don't like it in java.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: Ok done.

collect_app/src/main/res/xml/admin_preferences.xml
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: I have both but can't change that value to false because the summary is not updated.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: Now as I implemented SummaryProvider this is no longer the case.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
seadowg: Hmmm I can have a play with it on Monday if you push what you have?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -62,6 +76,20 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
         findPreference("main_menu").setOnPreferenceClickListener(this);
         findPreference("user_settings").setOnPreferenceClickListener(this);
         findPreference("form_entry").setOnPreferenceClickListener(this);
+
+        findPreference(PROJECT_NAME_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_NAME_KEY, currentProjectProvider));
+        findPreference(PROJECT_ICON_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_ICON_KEY, currentProjectProvider));
+        findPreference(PROJECT_COLOR_KEY).setSummaryProvider(new ProjectDetailsSummaryProvider(PROJECT_COLOR_KEY, currentProjectProvider));
```
seadowg, line 82: I'm wondering if we can just use a `SimpleSummaryProvider` if we're also calling `setText` on these, but we can easily check that later.

seadowg (approved):

PR #4519, opened 2021-04-16
Extract projects module

seadowg:
Here's a first sketch of pulling projects out to its own module. The next steps would be:

- Move material toolbar layout to `material`
- Move any custom attrs we want to keep using the in the add project dialog layout to `androidshared`
- Move `AddProjectDialog` (and its layout and test) to `projects`
- Work out if `Settings` should have its own module (with implementations/tests) or just remain in `shared`

#### What has been done to verify that this works as intended?

Ran tests.

#### Why is this the best possible solution? Were any other approaches considered?

As mentioned above there is more we can do, but I think this is a good first step.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Shouldn't be any changes in behaviour. This can go in without QA.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4543, opened 2021-05-03
Move AddProjectsDialog to projects module

seadowg:
This was something we discussed as a follow-up to #4519. I ended up moving a few helpers and shared resources android (mostly to `androidshared` and `testshared`) so it's probably worth going commit by commit on this one.

Only potentially interesting thing here is the Dagger setup I've added for `projects` but this just follows the pattern we introduced for `audiorecorder`.

This can skip QA as it just moves some code around.


collect_app/build.gradle
```diff
@@ -361,8 +361,12 @@ dependencies {
     testImplementation "org.mockito:mockito-core:3.7.7"
     testImplementation 'org.mockito:mockito-inline:3.7.7'
 
-    testImplementation project(path: ':testshared')
-    testImplementation "org.robolectric:robolectric:4.3.1"
+    testImplementation(project(path: ':testshared')) {
+        exclude group: 'org.robolectric' // Some tests in `collect_app` don't work with newer Robolectric
```
seadowg, line 365: This problem is best left to our next "update all the libraries" cycle as it will involve reworking a few unrelated tests.

collect_app/src/main/res/layout/material_full_screen_dialog_toolbar.xml
```diff
@@ -1,17 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<com.google.android.material.appbar.AppBarLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
grzesiek2010, line 2: Why did you remove this layout? Couldn't you just change that style here?

grzesiek2010 (changes requested):

grzesiek2010:
Please fix conflicts.

collect_app/src/main/res/layout/material_full_screen_dialog_toolbar.xml
seadowg: I played around with moving it around to a shared module but in the end it seemed simpler to just have a shared style (`Widget.AndroidShared.Toolbar.Primary.CloseIcon`). I find `includes` can be a little awkward as you can't then customize/override attributes of the shared layout from where they are being used.

grzesiek2010 (approved):

PR #4544, opened 2021-05-04
Implement project switching

### LINKED ISSUE #4417: Switching between projects changes settings and forms
Blocked by #4435, #4453, #4467, #4420, #4498, #4418

## Acceptance

- [x] Given I've changed some settings in Collect
And downloaded and filled out some forms
When I navigate to the settings/projects dialog
And I add a project
And I select it
Then the settings I changed are now the default again
And there are no downloaded or filled out forms
When I switch back to the previous project
Then I see the forms and settings as they were originally

- [x] Given I've installed Collect for the first time on a device
When I opt to use the demo project
Then my server is the demo one (sandbox)
And all my settings are the default

- [x] Given I'm using an older version of Collect
When I upgrade to the new version
Then my settings (including server) are the same as they were before
And my forms are still the same as they were before

## Notes

App data should now be stored by project. The directory structure for a one project setup (or a demo/migrated setup) should look like this:

```
org.odk.collect.android
|-  files
    |-  projects 
        |-  <project uuid>
            |-  forms
            |-  instances
            |-  layers
            |-  metadata
```

`recordings` (and any other ancillary data dirs) will still live at the top level (in `org.odk.collect.android/files`).

seadowg:
Closes #4417

This ended being larger than expected as I had to rework the connected tests (a couple of times in the end) and also got caught up in some subtler issues around `SharedPreferences` listeners and database connections that I was hoping to avoid until we deal with background jobs. The latter I've managed to add a "quick fix" for which I'll be removing. I'd advise going through this commit by commit as a lot of file changes are renames in test helpers and string changes.

Verified the acceptance manually and wrote new tests.

This is mostly based on the architecture discussed previously and on the foundations laid by @grzesiek2010 so not a lot of new approaches to discuss here I don't think.

This could definitely cause bugs, but I think we should merge to master and then QA once we've dealt with background jobs.


collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
```diff
@@ -25,13 +26,19 @@
 
     private final AppDependencyModule appDependencyModule;
     private final SettingsProvider settingsProvider = TestSettingsProvider.getSettingsProvider();
+    private final boolean firstLaunch;
```
grzesiek2010: This is always true so do we even need this field? Maybe there will be cases in the future where it can be false?

collect_app/src/main/java/org/odk/collect/android/utilities/AppStateProvider.kt
```diff
@@ -19,6 +21,6 @@ class AppStateProvider(private val packageInfo: PackageInfo, private val metaSet
 
     companion object {
         @JvmField
-        var alwaysFresh: Boolean = false
+        var overrideFresh: Boolean? = null
```
grzesiek2010: Why not false by default? Then you wouldn't have to check if it's null first for example in `CollectTestRule`:
```
if (overrideFresh != null && overrideFresh && skipLaunchScreen) {
                    onView(withText(R.string.configure_later)).perform(click());
                }
```

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -532,7 +530,7 @@ public InstancesAppState providesInstancesAppState(Application application) {
 
     @Provides
     @Singleton
-    public FormsDatabaseProvider providesFormsDatabaseProvider() {
+    public FormsDatabaseProvider providesFormsDatabaseProvider(CurrentProjectProvider currentProjectProvider) {
```
grzesiek2010: Looks like we don't need this change.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -568,8 +566,8 @@ public InstancesRepositoryProvider providesInstancesRepositoryProvider() {
     }
 
     @Provides
-    public ProjectImporter providesProjectImporter(ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
-        return new ProjectImporter(projectsRepository, settingsProvider.getMetaSettings());
+    public ProjectImporter providesProjectImporter(ProjectsRepository projectsRepository, CurrentProjectProvider currentProjectProvider, StorageInitializer storageInitializer, StoragePathProvider storagePathProvider, Context context, SettingsProvider settingsProvider) {
```
grzesiek2010: CurrentProjectProvider and StorageInitializer are not used here.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalSelectsTest.java
```diff
@@ -0,0 +1,37 @@
...
+            .around(rule);
+
+    @Test
+    public void displaysAllChoicesFromItemsetsCSV() {
+        new MainMenuPage()
+                .startBlankForm("selectOneExternal")
+
+                // Check all states in `itemsets.csv` are displayed
+                .assertText("Texas")
+                .assertText("Washington");
```
grzesiek2010: Why did you reduce this test like that? Now it checks only the first question which comes from the xml file not csv so it doesn't even touch that external file.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalSecondaryInstanceTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public RuleChain copyFormChain = RuleChain
+            .outerRule(new ResetStateRule())
+            .around(rule);
+
+    @Test
+    public void displaysAllOptionsFromSecondaryInstance() {
```
grzesiek2010, line 36: Here as well you reduced the test and I think it's something that should be discussed with the QA team because those test cases come from their regression check list.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectSettingsDialog.kt
```diff
@@ -37,10 +39,10 @@ class ProjectSettingsDialog : DialogFragment() {
     lateinit var projectsRepository: ProjectsRepository
 
     @Inject
-    lateinit var currentProjectProvider: CurrentProjectProvider
+    lateinit var currentProjectViewModelFactory: CurrentProjectViewModel.Factory
 
     @Inject
-    lateinit var currentProjectViewModelFactory: CurrentProjectViewModel.Factory
+    lateinit var storageInitializer: StorageInitializer
```
grzesiek2010: We don't need this.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -151,8 +155,10 @@ private void initializeMapFrameworks() {
 
     private void importExistingProjectIfNeeded() {
         if (!appStateProvider.isFreshInstall() && !metaSettings.getBoolean(MetaKeys.EXISTING_PROJECT_IMPORTED)) {
-            projectImporter.importExistingProject();
+            Project.Saved existingProject = projectImporter.importExistingProject();
+            currentProjectProvider.setCurrentProject(existingProject.getUuid());
```
grzesiek2010, line 159: What's the benefit of moving this out?

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
```diff
@@ -20,7 +18,7 @@ class SettingsProvider(private val context: Context) {
     }
 
     @JvmOverloads
-    fun getGeneralSettings(projectId: String = NOT_SPECIFIED_UUID): Settings {
+    fun getGeneralSettings(projectId: String? = null): Settings {
         val settingsId = getSettingsId(GENERAL_SETTINGS_NAME, projectId)
 
         return settings.getOrPut(settingsId) {
```
grzesiek2010, line 23: Can't we improve this by removing that `SharedPreferencesSettings(PreferenceManager.getDefaultSharedPreferences(context), GeneralKeys.DEFAULTS)`
If you migrated settings now it should be needed only to read those default settings during migration and once this operation is done we shouldn't use it right? It should work like in the case of adminSettings. If I'm right you could remove it from here to avoid confusion and use that `getDefaultSharedPreferences` directly in the code where you perform that migration or have a separate method here in this class that would return default settings but has a comment or a good descriptive name.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
```diff
@@ -1,29 +1,78 @@
...
+
+    public String getOdkDirPath(StorageSubdirectory subdirectory, Project.Saved project) {
+        switch (subdirectory) {
+            case FORMS:
+            case INSTANCES:
+            case CACHE:
+            case METADATA:
+            case LAYERS:
+                return getProjectRootDirPath(project) + File.separator + subdirectory.getDirectoryName();
+            case SETTINGS:
```
grzesiek2010, line 70: Maybe I missed the discussion about having one shared directory for settings so could you tell me something more? Did you think about it? What pros/cons did you take into account?

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
```diff
@@ -1,29 +1,78 @@
...
+            case PROJECTS:
+                return getOdkRootDirPath() + File.separator + subdirectory.getDirectoryName();
+            default:
+                throw new IllegalStateException("Unexpected value: " + subdirectory);
+        }
     }
 
     public String getCustomSplashScreenImagePath() {
```
grzesiek2010, line 79: I think we need to have a separate issue for this case. If we store all splash screen images in the same place and using the same name it's no possible to set different images for different projects. I'm not even sure what the expected behavior is but definitely it's something we should think through.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
seadowg: Oh I think you're right that we just don't need this any longer! I'll take it out.

collect_app/src/main/java/org/odk/collect/android/utilities/AppStateProvider.kt
seadowg: Yeah, I think like the above comment this is out of date now. I'll look at cleaning it up.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalSelectsTest.java
seadowg: Nice catch! Looks live I've misinterpreted the form as the states are in the `itemsets.csv` as well, but you're right that the state select is just a normal select. I'll fix this. I might try and simplify the form a little as well.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalSecondaryInstanceTest.java
seadowg: Yeah good point. I have a note to chat to QA about "Issue number NODK-377 Test case 2" (the other test). I'm thinking we probably don't need a specific automated test for normal/internal selects at this level given it's tested at the widget level (and doesn't involve "external files" like the external select tests. What do you think @mmarciniak90 @kkrawczyk123?

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: You mean the `CurrentProjectProvider` out of `ProjectImporter`?  I was just trying to have the `...Importer` focus on the setup/creation of the project. I was finding it a bit overwhelming having one component that created the project and switched to it. That might get simpler once we've done some rework for background jobs I guess. It could also get worse though 😕

collect_app/src/main/java/org/odk/collect/android/preferences/source/SettingsProvider.kt
seadowg: Ah, yeah nice idea! The project importer currently grabs the default shared prefs and admin prefs directly (as it knows that's exactly what it needs) so we can remove that complexity here now.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
seadowg: This is a mistake on my part. `settings` just stores QR codes as far as I can see so that should definitely be in the project directory. I'll fix it.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.java
seadowg: ~~Hmmmm. I think because of the new [splash screen API](https://developer.android.com/about/versions/12/features/splash-screen) (which looks like it only supports hardcoded splash icons) we probably want to deprecate the custom splash screen image sometime soon.~~

There's an [issue for just removing the splash screen configuration](https://github.com/getodk/collect/issues/254) that we probably want to schedule within the next couple of releases anyway. @lognaturel do you think we should go ahead and stick this in this release so we don't have to worry about the custom splash?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
```diff
@@ -1,33 +1,67 @@
...
-        metaSettings.save(MetaKeys.CURRENT_PROJECT_ID, DEMO_PROJECT_ID)
+        setupProject(project)
     }
 
-    // Now it does the same like importDemoProject() but it should be changed later
-    fun importExistingProject() {
-        val project = Project("Demo project", "D", "#3e9fcc", DEMO_PROJECT_ID)
-        projectsRepository.save(project)
-        metaSettings.save(MetaKeys.CURRENT_PROJECT_ID, DEMO_PROJECT_ID)
+    fun importExistingProject(): Saved {
```
grzesiek2010, line 27: I feel uneasy with this solution a user can have a lot of files so this operation might be time-consuming. Shouldn't this be a service like we did in the case of migrating files between storages?

seadowg: This had crossed my mind as well - we might just always get away with it as it's a move rather than a copy but I do feel uncomfortable doing IO on the UI thread like this. It would be good for us to test this out with a large amount of files. How about I add an issue to optimize this with a "needs discussion" on it (with a note for us to test this)?

grzesiek2010: Yeah we have just talked about it with @lognaturel and having an issue for testing this case seems fine. If the result's are no satisfying we will optimize it.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
grzesiek2010: Ok I see, I'm not totally convinced but also not against so let's leave it as is.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
seadowg: Ok issue [here](https://github.com/getodk/collect/issues/4578). I realized when writing it that we might be able to just run `ApplicationInitializer` off the UI thread to make these kinds of things easier.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ExternalSecondaryInstanceTest.java
mmarciniak90: @seadowg, I agree. In NODK-377, we checked both types for comparison. It's not necessary for automated tests.

grzesiek2010:
@seadowg you have some conflicts

collect_app/src/main/java/org/odk/collect/android/projects/ProjectSettingsDialog.kt
```diff
@@ -110,6 +116,10 @@ class ProjectSettingsDialog : DialogFragment() {
         return binding.root
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        (requireDialog() as AlertDialog).setView(view)
```
seadowg: @grzesiek2010 I ended up in a weird place here as I was using `viewLifecycleOwner` which scarily doesn't work in `onCreateDialog`. After playing with a few different options it seems like the right thing to do is to override `onCreateDialog` to create the dialog, `onCreateView` to create the view and `onViewCreated` to set the view on the dialog.

In the [docs](https://developer.android.com/guide/fragments/dialogs#custom) it suggests that we shouldn't have to do this and that just overrding `onCreateView` should do the trick. Trying this out the problem ends up being that we end up with a `Dialog` instead of an `AlertDialog` and then if we override `onCreateDialog` we use the magic of having the view automatically set on the dialog 🤯

seadowg:
@grzesiek2010 ok that's ready to go again

collect_app/src/main/java/org/odk/collect/android/projects/ProjectSettingsDialog.kt
grzesiek2010: but you still should be able to initialize binding in `onCreateDialog()`:

```
    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
        binding = ProjectSettingsDialogLayoutBinding.inflate(LayoutInflater.from(context))

        return AlertDialog.Builder(requireActivity())
            .setView(binding.root)
            .create()
    }
```

then you could remove `onViewCreated()` and it would be a bit less confusing why we need to use as many lifecycle methods I think.

seadowg: Ah wait also looking at this again I've realized that we shouldn't use `viewLifecycleOwner` anyway in DialogFragments. This is in a note in the [docs](https://developer.android.com/guide/fragments/dialogs):

> Note: When subscribing to lifecycle-aware components such as LiveData, you should never use viewLifecycleOwner as the LifecycleOwner in a DialogFragment that uses Dialogs. Instead, use the DialogFragment itself, or if you're using Jetpack Navigation, use the NavBackStackEntry.

So I can definitely simplify this.

EDIT: I've filed an issue with Android on this being confusing: https://issuetracker.google.com/issues/189250614

grzesiek2010 (approved):

PR #4547, opened 2021-05-05
Fixed starting First Launch Screen

### LINKED ISSUE #4538: First Launch Screen appears every time the app is started if the app is newly installed (not upgraded)
Master branch

#### Problem description
First Launch Screen appears every time the app is started if the app is newly installed (not upgraded)

grzesiek2010:
Closes #4538 

The previous approach that we used in `AppStateProvider` didn't work when the app was newly installed (`packageInfo.firstInstallTime == packageInfo.lastUpdateTime`) so I mixed it with shared preferences to make sure it works in all the cases.

It's just a bug fix that should solve the problem described in the issue. Nothing else might be affected.

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/utilities/AppStateProviderTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/utilities/AppStateProviderTest.kt
@@ -1,4 +1,50 @@
 package org.odk.collect.android.utilities
 
+import android.content.pm.PackageInfo
+import org.hamcrest.CoreMatchers.`is`
+import org.hamcrest.MatcherAssert.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mockito.`when`
+import org.mockito.Mockito.mock
+import org.odk.collect.android.preferences.keys.MetaKeys
+import org.odk.collect.shared.Settings
+
 class AppStateProviderTest {
+    private lateinit var appStateProvider: AppStateProvider
+    private lateinit var packageInfo: PackageInfo
+    private lateinit var metaSettings: Settings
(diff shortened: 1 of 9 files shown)


collect_app/src/test/java/org/odk/collect/android/utilities/AppStateProviderTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        packageInfo = PackageInfo()
+        metaSettings = mock(Settings::class.java)
+        appStateProvider = AppStateProvider(packageInfo, metaSettings)
+    }
+
+    @Test
+    fun `When app is newly installed isFreshInstall() should return true when opened for the first time`() {
+        packageInfo.firstInstallTime = 100
+        packageInfo.lastUpdateTime = 100
+        `when`(metaSettings.contains(MetaKeys.FIRST_LAUNCH)).thenReturn(false)
```
seadowg: We can start using `mockito-kotlin` now to avoid the back ticks: https://github.com/mockito/mockito-kotlin/wiki/Mocking-and-verifying

seadowg (changes requested):
Looks good! I noticed this problem yesterday but saw you'd already started on a fix. Just one comment about moving to mockito-kotlin for new tests.

collect_app/src/test/java/org/odk/collect/android/utilities/AppStateProviderTest.kt
grzesiek2010: Done.

seadowg (approved):
Will merge after checks

PR #4550, opened 2021-05-06
Change machine to large to reduce credit use

yanokwa:


lognaturel (approved):
Merging now so further CI experiments don't use excessively large and expensive machines.

PR #4552, opened 2021-05-06
CI tweaks

seadowg:
Follow up from #4549.

Ran tests/CI.

This gets us away from having all the complex bash in the circle ci config and also ensures we're actually running all the tests but not much to discuss here.

This can go in without QA.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,7 +121,11 @@ jobs:
 
       - run:
           name: Generate list of tests for this fork
-          command: cat collect_app_test_classes.txt | circleci tests split > fork_test_classes.txt
+          command: |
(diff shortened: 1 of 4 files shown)


yanokwa:
Not related to this specific PR, but since we are tweaking...Is there a reason you limited parallelism to 50? We can have up to 83 (which is oddly specific) on our plan. 

Screen Shot 2021-05-06 at 12 01 53

I'm sure sure if that slows builds on other repos, but I don't think so?

seadowg:
@yanokwa yeah I was just upping the number as I played around with different splits, but I can boost it TO THE MAX.

grzesiek2010 (approved):

PR #4553, opened 2021-05-10
User can delete projects

### LINKED ISSUE #4424: User can delete projects
Blocked by #4497

**Acceptance**

- [x] Given I'm on the settings/projects dialog
When I click "Admin settings"
And I click "Delete project"
And I confirm
Then the project I was in no longer appears in the app

grzesiek2010:
Closes #4424

No.

No.


collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesActivity.java
```diff
@@ -70,4 +73,10 @@ public void preventOtherWaysOfEditingForm() {
     public void onDialogClosed() {
         startActivityAndCloseAllOthers(this, MainMenuActivity.class);
     }
+
+    @Override
+    public void deleteProject() {
+        AdminPreferencesFragment fragment = (AdminPreferencesFragment) getSupportFragmentManager().findFragmentByTag(TAG);
```
seadowg: It feels a little tangled to me to have the dialog call back to the Activity and then have the Activity have to grab a Fragment and call it to make the deletion happen. Could the delete logic (currentily in `AdminPreferencesFragment#deleteProject` not just live here or am I missing something?

Alternatively we could just use a `MaterialAlertDialogBuilder` in `AdminPreferencesFragment` to keep all the logic in there. We probably don't need this dialog to stay on screen during a rotation or other config change so I don't see a problem with it being "inline" instead of a separate `DialogFragment`.

projects/src/main/java/org/odk/collect/projects/DeleteProjectDialog.kt
```diff
@@ -0,0 +1,37 @@
...
+        if (context is DeleteProjectListener) {
+            listener = context
+        }
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        return AlertDialog.Builder(requireActivity())
+            .setTitle(R.string.delete_project_confirm_message)
+            .setNegativeButton(R.string.delete_project_no) { _: DialogInterface?, _: Int ->
+                dismiss()
```
seadowg: We should have a test for this flow

collect_app/src/main/java/org/odk/collect/android/preferences/source/SharedPreferencesSettings.kt
```diff
@@ -92,6 +92,8 @@ class SharedPreferencesSettings(private val sharedPreferences: SharedPreferences
     }
 
     override fun unregisterOnSettingChangeListener(listener: Settings.OnSettingChangeListener) {
-        sharedPreferences.unregisterOnSharedPreferenceChangeListener(sharedPreferencesListener)
+        if (this::sharedPreferencesListener.isInitialized) {
```
seadowg: I think we can just make `sharedPreferencesListener` a `?` type rather than `lateinit` here instead.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -150,9 +150,10 @@ private void initializeMapFrameworks() {
     }
 
     private void importExistingProjectIfNeeded() {
-        if (!appStateProvider.isFreshInstall() && projectsRepository.getAll().isEmpty()) {
+        if (!appStateProvider.isFreshInstall() && !metaSettings.getBoolean(MetaKeys.ALREADY_TRIED_TO_IMPORT_EXISTING_PROJECT)) {
```
seadowg: I'm not sure I fully understand why this is needed? I'm guessing we end up with a problematic state if not?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/source/SharedPreferencesSettings.kt
grzesiek2010: This was removed in one of the other commits.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
grzesiek2010: Because if you remove all projects you are moved to the `First Launch Screen`. Then a user can kill the app and open it again what would trigger importing the existing project what I want to avoid (I want that importing to be triggered just once during upgrading to the version with projects). Is that clear?

seadowg: Right that makes sense. Thanks. Could we change the name to be `EXISTING_PROJECT_IMPORTED`? I think the `TRIED` made me think it might be related to an earlier failure or something.

grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesActivity.java
grzesiek2010: Fixed.

projects/src/main/java/org/odk/collect/projects/DeleteProjectDialog.kt
grzesiek2010: This class does not exist anymore.

seadowg (approved):

PR #4556, opened 2021-05-11
Upgrade Robolectric

seadowg:
We've been running into issues with Robolectric recently (mostly around it not being able to download JARs on-the-fly) so I really think it's time to do this. Hopefully that will get Robolectric back to a stabler place. A bunch of tests break after upgrading to `4.5.1` but it mostly looks like it's to do with the recent [looper changes](http://robolectric.org/blog/2019/06/04/paused-looper/). These can be fixes by either: 

* Adding calls to `RobolectricHelpers.runLooper()` before assertions that are failing (probably after clicks, showing fragments etc)
* Adding `@LooperMode(LooperMode.Mode.LEGACY)` to the test class. This is only really advisable if the test seems to complex to fix with `runLooper` or if it uses things like `Robolectric.getBackgroundThreadScheduler()` and so needs completely reworked at some point.

Todo here:

- [x] Fix tests that are broken because of upgrade
- [x] Upgrade to newest Android Gradle plugin


grzesiek2010:
Ok I think this is done. We ended up with just three tests with `@LooperMode(LooperMode.Mode.LEGACY`) what seems fine.

I added some additional improvements like:
- using the same runner across the project https://github.com/getodk/collect/pull/4556/commits/aeddd960627e3b0f7de1fdccaea8d014efa52738
- using `ApplicationProvider.getApplicationContext()` instead of deprecated `RuntimeEnvironment.application` https://github.com/getodk/collect/pull/4556/commits/969902cff5833b203f8c002b84d1f860dbb752d9
- removed redundant `@LooperMode(PAUSED)` annotations because since robolectric 4.4 it's used by default https://github.com/getodk/collect/pull/4556/commits/394d1df84b73dc30df490aa8a36bfcecc47cd5ba

lognaturel (approved):
@grzesiek2010 I assume you've tagged me for review not for my extensive Robolectric knowledge (😂) but because this is better to merge as soon as possible and the changes were pretty mechanical overall. I scrolled through to make sure nothing looked likely to change what was being tested. I can't see anything dangerous.

Please squash on merge.

seadowg:
@grzesiek2010 this looks great. Thanks for taking over and also going the extra mile with the clean-up on runners and `ApplicationProvider` etc.

PR #4558, opened 2021-05-11
Install Robolectric dependencies manually

seadowg:
This provides a fix for the `Unable to resolve artifact: Missing` error during Robolectric tests we're seeing occasionally locally (it was blocking me today) and quite a lot on CI. This uses the very undocumented `robolectric-deps.properties` to reference locally downloaded JARs rather than relying on Robolectric's very broken on-the-fly downloads.

Ran tests locally and on CI.

I've looked at a bunch of different solutions and I think we're running into two closed Robolectric issues in Robolectric:

* https://github.com/robolectric/robolectric/issues/5456
* https://github.com/robolectric/robolectric/issues/5305

This is most likely because we're using an older Robolectric version or because our Android Gradle plugin is out of date (or the issues aren't actually fixed). 

For the moment I think providing a work around that we can use and can be used on CI is the best way to go. I've also created another PR for upgrading Robolectric (and to hopefully get us away from those issues) at #4556.

This can go in without QA.


.circleci/config.yml
```diff
@@ -93,7 +97,7 @@ jobs:
 
   test_app:
     <<: *android_config
-    parallelism: 83
+    parallelism: 50
```
seadowg, line 100: I think pushing `parellelism` TO THE MAX was causing our parallel jobs to get queued.

grzesiek2010 (approved):

PR #4562, opened 2021-05-12
Implement and expose Content Provider for Projects

### LINKED ISSUE #4471: Implement and expose Content Provider for Projects
~~Blocked by #4420~~

**Acceptance**

- [ ] External apps should be able to use a [Content Provider](https://developer.android.com/guide/topics/providers/content-providers) to retrieve a list of Projects (ids and name are the only details that need to be included).

## Notes

We don't currently have any need to allow external apps to create, update or delete Projects so only `query` needs to be implemented. Other `ContentProviders` should return appropriate empty/zero values.

grzesiek2010:
Closes #4471

I added tests and changes to [collectTester](https://github.com/grzesiek2010/collectTester/pull/33) so that we can also verify this change using an external app.

This is the only solution.

No.

No.

### DIFF
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -181,6 +181,10 @@ the specific language governing permissions and limitations under the License.
             android:name=".provider.InstanceProvider"
             android:authorities="org.odk.collect.android.provider.odk.instances"
             android:exported="true" />
+        <provider
+            android:name=".provider.ProjectsProvider"
+            android:authorities="org.odk.collect.android.provider.odk.projects"
+            android:exported="true" />
 
         <activity-alias
             android:name=".activities.FormEntryActivity"
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -182,7 +182,7 @@ the specific language governing permissions and limitations under the License.
             android:authorities="org.odk.collect.android.provider.odk.instances"
             android:exported="true" />
         <provider
-            android:name=".provider.ProjectsProvider"
+            android:name="org.odk.collect.projects.providers.ProjectsProvider"
             android:authorities="org.odk.collect.android.provider.odk.projects"
             android:exported="true" />
 
--- a/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
@@ -1,0 +1,48 @@
+package org.odk.collect.android.provider
+
+import android.content.ContentProvider
+import android.content.ContentValues
+import android.database.Cursor
+import android.database.MatrixCursor
+import android.net.Uri
+import org.odk.collect.android.injection.DaggerUtils
+import org.odk.collect.projects.ProjectsRepository
+import javax.inject.Inject
+
+class ProjectsProvider : ContentProvider() {
+
+    @Inject
+    lateinit var projectsRepository: ProjectsRepository
+
+    // Do not call it in onCreate() https://stackoverflow.com/questions/23521083/inject-database-in-a-contentprovider-with-dagger
+    private fun deferDaggerInit() {
+        DaggerUtils.getComponent(context).inject(this)
+    }
+
+    override fun onCreate(): Boolean {
+        return true
+    }
+
+    override fun query(uri: Uri, projection: Array<out String>?, selection: String?, selectionArgs: Array<out String>?, sortOrder: String?): Cursor? {
+        deferDaggerInit()
+
+        val cursor = MatrixCursor(arrayOf(PROJECT_UUID_COLUMN_NAME, PROJECT_NAME_COLUMN_NAME))
+        projectsRepository.getAll().forEach {
+            cursor.addRow(arrayOf<Any>(it.uuid, it.name))
+        }
+        return cursor
+    }
+
+    override fun getType(uri: Uri) = ProjectsProviderAPI.CONTENT_TYPE
+
+    override fun insert(uri: Uri, values: ContentValues?) = throw UnsupportedOperationException()
+
+    override fun delete(uri: Uri, selection: String?, selectionArgs: Array<out String>?) = throw UnsupportedOperationException()
+
+    override fun update(uri: Uri, values: ContentValues?, selection: String?, selectionArgs: Array<out String>?) = throw UnsupportedOperationException()
+
+    companion object {
+        const val PROJECT_UUID_COLUMN_NAME = "uuid"
+        const val PROJECT_NAME_COLUMN_NAME = "name"
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
@@ -26,7 +26,7 @@ class ProjectsProvider : ContentProvider() {
     override fun query(uri: Uri, projection: Array<out String>?, selection: String?, selectionArgs: Array<out String>?, sortOrder: String?): Cursor? {
         deferDaggerInit()
 
-        val cursor = MatrixCursor(arrayOf(PROJECT_UUID_COLUMN_NAME, PROJECT_NAME_COLUMN_NAME))
+        val cursor = MatrixCursor(arrayOf(PROJECT_UUID, PROJECT_NAME))
         projectsRepository.getAll().forEach {
             cursor.addRow(arrayOf<Any>(it.uuid, it.name))
         }
@@ -42,7 +42,7 @@ class ProjectsProvider : ContentProvider() {
     override fun update(uri: Uri, values: ContentValues?, selection: String?, selectionArgs: Array<out String>?) = throw UnsupportedOperationException()
 
     companion object {
-        const val PROJECT_UUID_COLUMN_NAME = "uuid"
-        const val PROJECT_NAME_COLUMN_NAME = "name"
+        const val PROJECT_UUID = "uuid"
+        const val PROJECT_NAME = "name"
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
@@ -16,7 +16,7 @@ class ProjectsProvider : ContentProvider() {
     lateinit var projectsRepository: ProjectsRepository
 
     // Do not call it in onCreate() https://stackoverflow.com/questions/23521083/inject-database-in-a-contentprovider-with-dagger
-    private fun deferDaggerInit() {
+    private fun daggerInit() {
         DaggerUtils.getComponent(context).inject(this)
     }
 
@@ -25,7 +25,7 @@ class ProjectsProvider : ContentProvider() {
     }
 
     override fun query(uri: Uri, projection: Array<out String>?, selection: String?, selectionArgs: Array<out String>?, sortOrder: String?): Cursor? {
-        deferDaggerInit()
+        daggerInit()
 
         val cursor = MatrixCursor(arrayOf(BaseColumns._ID, PROJECT_UUID, PROJECT_NAME))
         var index = 1
--- a/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
@@ -5,6 +5,7 @@ import android.content.ContentValues
 import android.database.Cursor
 import android.database.MatrixCursor
 import android.net.Uri
+import android.provider.BaseColumns
 import org.odk.collect.android.injection.DaggerUtils
 import org.odk.collect.projects.ProjectsRepository
 import javax.inject.Inject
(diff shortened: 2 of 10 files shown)


collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.odk.collect.projects.ProjectsRepository
+import javax.inject.Inject
+
+class ProjectsProvider : ContentProvider() {
+
+    @Inject
+    lateinit var projectsRepository: ProjectsRepository
+
+    // Do not call it in onCreate() https://stackoverflow.com/questions/23521083/inject-database-in-a-contentprovider-with-dagger
+    private fun deferDaggerInit() {
```
seadowg: The naming of this threw me off a little as "defer" makes me think "do dagger init later" rather than "we are doing it later than expected". This could maybe be named "deferredDaggerInit"? Even simpler, it could just be `daggerInit` as I think your comment gives enough context.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    override fun getType(uri: Uri) = ProjectsProviderAPI.CONTENT_TYPE
+
+    override fun insert(uri: Uri, values: ContentValues?) = throw UnsupportedOperationException()
+
+    override fun delete(uri: Uri, selection: String?, selectionArgs: Array<out String>?) = throw UnsupportedOperationException()
+
+    override fun update(uri: Uri, values: ContentValues?, selection: String?, selectionArgs: Array<out String>?) = throw UnsupportedOperationException()
+
+    companion object {
```
seadowg: Should these go in `ProjectProviderAPI` seeing as they are part of the "API" of the content provider. I think it would be nice if we could point people to one class to see how to interact with it (it'll be interesting to think about actually documenting some of this).

seadowg (changes requested):
This is looking good! One change I would suggest is moving `ProjectsProvider`, `ProjectsProviderTest` and `ProjectsProviderAPI` to the `projects` module so all our project specific code stays in there.

Good to also get @lognaturel's input on this as it's externally facing.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
```diff
@@ -0,0 +1,11 @@
+package org.odk.collect.android.provider
+
+import android.net.Uri
+
+object ProjectsProviderAPI {
+    const val AUTHORITY = "org.odk.collect.android.provider.odk.projects"
```
lognaturel: This naming convention seems strange to me. I would expect a single authority of `org.odk.collect.android.provider` shared between all providers -- see https://developer.android.com/guide/topics/providers/content-provider-creating. 

This does match `org.odk.collect.android.provider.odk.forms` and `org.odk.collect.android.provider.odk.instances`, though. I'm ok with matching the existing convention but let's make sure that's an explicit decision you both like, @grzesiek2010, @seadowg. I have a slight preference for going with `org.odk.collect.android.provider` in the hopes that we could one day correct the errors of the past but it's not strong.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+    // Do not call it in onCreate() https://stackoverflow.com/questions/23521083/inject-database-in-a-contentprovider-with-dagger
+    private fun deferDaggerInit() {
+        DaggerUtils.getComponent(context).inject(this)
+    }
+
+    override fun onCreate(): Boolean {
+        return true
+    }
+
+    override fun query(uri: Uri, projection: Array<out String>?, selection: String?, selectionArgs: Array<out String>?, sortOrder: String?): Cursor? {
```
lognaturel: It feels weird to me to ignore all of the parameters here. Is there another story to use them? If not, should there be some kind of error/warning if they're included?

Eventually the other project details will be included, right? I'm specifically thinking it will be important to include the server URI because that's a stable value that can be matched on. Imagine you want to write a companion app that reads form and submission metadata from a specific project only. You're going to need some way of consistently identifying the project you need. Doing so by server URI (and maybe also username) seems like the most reliable option (recognizing that there could be more than one project with the same server URI). I'm not exactly sure yet how this works with App User tokens.

I also wonder whether we should support the id convention so that folks can easily display a list of projects in their companion apps (https://developer.android.com/guide/topics/providers/content-provider-creating). Or is your idea that you'd like people to explicitly request this before we build it? (fine by me)

collect_app/src/main/AndroidManifest.xml
```diff
@@ -181,6 +181,10 @@ the specific language governing permissions and limitations under the License.
             android:name=".provider.InstanceProvider"
             android:authorities="org.odk.collect.android.provider.odk.instances"
             android:exported="true" />
+        <provider
+            android:name=".provider.ProjectsProvider"
+            android:authorities="org.odk.collect.android.provider.odk.projects"
+            android:exported="true" />
```
lognaturel, line 187: We could use this new provider as an opportunity to introduce [permissions](https://developer.android.com/guide/topics/providers/content-provider-creating). In fact, we may want to do this for all providers with this release. On one hand, there isn't much sensitive exposed by any of these anymore. On the other, project URI, instance_name, geometry actually could be quite sensitive.

I would go with a single read/write provider-level permission. E.g. `org.odk.collect.android.provider.permission.READ_WRITE_PERMISSION`. Ideally it would be shared between all providers, I think.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
seadowg: Yeah, I think agree after reading those docs that we should just have one authority for the app. It seems that we want to think about Collect as a "content provider" and that the `ContentProvider` implementations come together to form that. I think here it would be a good idea to change that everywhere and just use `org.odk.collect.android.provider`. We could still support `.forms` and `.instances` by keeping them in the relevant `ContentProvider` implementation's uri matcher. What do you folks think?

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
seadowg: I think adding warnings and supporting other parameters should be an extra issue. I've added a discussion item to our catch-up tomorrow as then we walk through some different workflows that we might want to support.

 I'm definitely on board with a lot of this: I think adding projects to Collect's content provider could potentially allow for better integration with Collect (setting up projects for "workflow" apps) which might be a nice way to avoid people needing to fork the app.

collect_app/src/main/AndroidManifest.xml
seadowg: I think we should plan this as a follow-up issue, so we can discuss properly. Again, I've added an item to our catch-up.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
seadowg: Actually scratch all that. Looking at the `<provider>` documentation it does seem like we want different authorities for different kinds of data ("A list of one or more URI authorities that identify data offered by the content provider" and "Typically, it's the name of the ContentProvider subclass that implements the provider"). This seems to make even more sense to me if we end up with different permissions.

grzesiek2010: First of all, changing anything in our current pattern would break all the apps that use this option. Is this something we would accept in the name of introducing such small adjustments?

collect_app/src/main/AndroidManifest.xml
grzesiek2010: I agree it should be a separate issue.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
grzesiek2010: Yes, this should be a separate issue. We can discuss it once we are sure what project details should be included.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProvider.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
lognaturel: > We could still support .forms and .instances by keeping them in the relevant ContentProvider implementation's uri matcher.

Is that confirmed possible? If so, that might be a nice way to transition folks as needed.

To be clear, I'm not suggesting we change the other providers' authorities unless we can have the kind of migration path described above. I'm wondering about the convention we would use in an ideal world. To me it feels like we're exposing different tables of connected data. From the perspective of an integrating app, it would seem more palatable to have a single authority with different paths. The Android docs are contradictory on this point so I think we should do what feels right.

> if we end up with different permissions.

Do you see a reason for this? Seems projects, forms and instances are tightly connected and should be accessible with the same permission.

lognaturel:
Fine by me for all the high-level questions I raised to be addressed in a followup issue. That said, I don't see a whole lot of benefit to merging this before they're addressed since this is only externally-facing. Up to y'all!

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
seadowg: > Is that confirmed possible? If so, that might be a nice way to transition folks as needed. 

It definietly works wit `UriMatcher` but then when I started thinking about the `<provider>` tag I realized that I was confused as to how Android would know which `ContentProvider` implementation to target if they all shared one authority. it turns out that this doesn't work: https://stackoverflow.com/questions/11975215/how-to-address-multiple-content-providers. Given that, the authorities we're using do make sense I think. 

@grzesiek2010 I think a change we should make given that we know we **need** a different authority for a provider is to change the URI we use (just for projects at the moment) to make it clear that projects are at the "root" of the provider. By that I mean the `CONTENT_URI` should probably be `"content://$AUTHORITY/"` instead of `"content://$AUTHORITY/projects"`. Can we make change here before this goes in?

seadowg (changes requested):
As comment

collect_app/src/main/java/org/odk/collect/android/provider/ProjectsProviderAPI.kt
grzesiek2010: Yes, I agree it was strange to me too that we end up with uris with doubled forms/instances/projects at the end but just wanted to follow the same pattern, but ok I can change it for projects.

grzesiek2010: Done!

lognaturel: &*$#@? Why don't the Android docs just include the one crucial piece of information instead of making us run around?!

See also https://stackoverflow.com/questions/11975215/how-to-address-multiple-content-providers#comment15967300_11975215. I think a more sensible organization for the data we're providing would be a single provider/authority with different paths.

lognaturel:
Thanks for doing this, @grzesiek2010, it sure had been informative. I thought this would be simple but clearly there's weirdness in the interface of the existing providers and a lack of clarity around just what we should expose for projects.

I've done more meditating on all of it and I think we should go back to one of our earlier proposals and not expose a project provider for now. I briefly discussed with @seadowg who seemed on board. That would mean keeping the same URI convention for instance and form providers and only giving access to the first project added chronologically (if we don't store enough state to determine that, yet, we should). This means:
- For folks who upgrade existing Collect installs and have an integrating app, nothing will change.
- In general, for folks who don't use multiple projects, there will be no change in behavior. 
- For folks who want to use multiple projects but only need to integrate with one, they'll need to be mindful of the order they add those projects in. If they get it wrong, they can always delete some projects and recreate them or uninstall and reinstall Collect.
- For folks who want to use multiple projects and integrate with more than one of them, we'll ask that they clearly describe their need. This is not functionality that exists today so there's no regression in behavior.

Some advantages of this approach:
- We'll get analytics from the other providers so we'll have a sense of what kind of usage there is currently.
- If no one needs to integrate with multiple projects, we won't need to create or maintaining this extra external-facing behavior.
- If people do need to integrate with multiple projects, we'll get a sense of their needs and be able to design the API more thoughtfully (e.g. do we expose the server URI? The username? What else...?)

I hate to back out of work that's already been done but I think the learning was really valuable so no loss. Does that sound ok, @grzesiek2010, @seadowg?

seadowg:
@lognaturel yeah this makes sense. We have an issue for updating the content providers for projects and I can update that.

grzesiek2010:
Hmm what you said makes some sense and since this functionality doesn't seem to be very important at the moment I tend to agree to postpone implementing it...

lognaturel:
> what you said makes some sense

🥳🥳 😂

~Could you please push the branch to this repo so it's easy to find once we come back to this?~ Oh, it already is here. So let's just keep it around for now.

PR #4567, opened 2021-05-17
Create one style for dialogs

### LINKED ISSUE #4482: Create one style for dialogs
#### Problem description
In https://github.com/getodk/collect/pull/4481 we added a dialog with rounded corners.
It would be good to have one style for dialogs which would define properties like rounded corners so that we can remove custom backgrounds like the one added for the project settings.

grzesiek2010:
Closes #4482 

I tested the app manually checking different dialogs.

I decided to get rid of MaterialAlertDialogs and use AlertDialogs from androidX because that second type of dialogs is used under the hood in preferences, so if we want to make our dialogs look the same across the app we can't use other dialog types in the rest of our codebase https://github.com/getodk/collect/commit/adfeee8d5f565755bb6d9c0bbe05a3c885d7242f.

Then I fixed my custom project settings dialog by using an alert dialog there as well: https://github.com/getodk/collect/commit/235c6c9607afa15ba14380f086fad08ec9328030 and thanks to that I don't need [the custom background view](https://github.com/getodk/collect/commit/235c6c9607afa15ba14380f086fad08ec9328030#diff-e91520b6e127ceb35d5fec81d1a442956d2f97e4344b44b558ecc17e73b108d8L1) anymore. Fortunately, it was the only dialog fragment that didn't use alert dialog so every other dialog fragment should also have rounded corners.

Some screenshots:

| Before  | After | Before  | After |
| ------------- | ------------- | ------------- | ------------- |
| Screenshot_1621254376  | Screenshot_1621254309  | Screenshot_1621254366 | Screenshot_1621254290 |
| Screenshot_1621254401 | Screenshot_1621254447 | Screenshot_1621254384  | Screenshot_1621254319  |
| Screenshot_1621344836 | Screenshot_1621344696 | Screenshot_1621344829 | Screenshot_1621344683

Let me know if you think that we need the corners even more rounded.

It should add rounded corners to our dialogs and make them look similar no matter if it's 
- a dialog in preferences 
- an alert dialog we use to confirm some actions 
- custom dialog fragments

No.

No.

### DIFF
--- a/collect_app/src/main/res/values/magenta_theme.xml
+++ b/collect_app/src/main/res/values/magenta_theme.xml
@@ -49,7 +49,7 @@
         <!-- Used as divider in many places. Not sure why. -->
         <item name="dividerCompat">@android:drawable/divider_horizontal_bright</item>
 
-        <item name="alertDialogTheme">@style/Theme.Collect.Magenta.Dialog</item>
+        <item name="alertDialogTheme">@style/Theme.Collect.Magenta.Dialog.Alert</item>
 
         <item name="searchViewStyle">@style/Widget.Collect.SearchView</item>
 
@@ -60,7 +60,7 @@
         <item name="android:textDirection">locale</item>
     </style>
 
-    <style name="Theme.Collect.Magenta.Dialog" parent="Theme.AppCompat.Light.Dialog.Alert">
+    <style name="Theme.Collect.Magenta.Dialog.Alert" parent="Theme.AppCompat.Light.Dialog.Alert">
         <item name="colorAccent">?attr/colorSecondary</item>
         <item name="buttonBarNegativeButtonStyle">@style/NegativeButtonStyle</item>
         <item name="buttonBarPositiveButtonStyle">@style/PositiveButtonStyle</item>
--- a/collect_app/src/main/res/values/magenta_theme.xml
+++ b/collect_app/src/main/res/values/magenta_theme.xml
@@ -49,8 +49,6 @@
         <!-- Used as divider in many places. Not sure why. -->
         <item name="dividerCompat">@android:drawable/divider_horizontal_bright</item>
 
-        <item name="alertDialogTheme">@style/Theme.Collect.Magenta.Dialog.Alert</item>
-
         <item name="searchViewStyle">@style/Widget.Collect.SearchView</item>
 
         <!-- Always show scrollbars for a11y reasons -->
@@ -58,12 +56,5 @@
 
         <!-- Use locale for text direction in views -->
         <item name="android:textDirection">locale</item>
-    </style>
-
-    <style name="Theme.Collect.Magenta.Dialog.Alert" parent="Theme.AppCompat.Light.Dialog.Alert">
-        <item name="colorAccent">?attr/colorSecondary</item>
-        <item name="buttonBarNegativeButtonStyle">@style/NegativeButtonStyle</item>
-        <item name="buttonBarPositiveButtonStyle">@style/PositiveButtonStyle</item>
-        <item name="dialogCornerRadius">8dp</item>
     </style>
 </resources>
--- a/collect_app/src/main/res/values/magenta_theme.xml
+++ b/collect_app/src/main/res/values/magenta_theme.xml
@@ -51,17 +51,13 @@
 
         <item name="alertDialogTheme">@style/Theme.Collect.Magenta.Dialog</item>
 
-        <item name="searchViewStyle">@style/Widget.Collect.SearchView.Magenta.Theme</item>
+        <item name="searchViewStyle">@style/Widget.Collect.SearchView</item>
 
         <!-- Always show scrollbars for a11y reasons -->
         <item name="android:fadeScrollbars">false</item>
 
         <!-- Use locale for text direction in views -->
         <item name="android:textDirection">locale</item>
-    </style>
-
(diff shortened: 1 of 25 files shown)


grzesiek2010:
@lognaturel @seadowg 
I attached both of you as reviewers. Of course, just one person can do that but I think both of you should at least take a look at the screenshots and share their thoughts. 

@seadowg 
Please take a look at the last commit where I fixed `ProjectSettingsDialogTest` it turns out that if we use `AlertDialogs` in `DialogFragments` we need to use `FragmentScenario.launch()` not `FragmentScenario.launchInContainer()` (when I use the second one `onCreateDialog()` is not even called, it's even in the documentation that we should use the first method https://developer.android.com/guide/fragments/test#dialog) and then we can't use espresso... am I missing something?

lognaturel:
Looks good to me! Always funny to see such a trivial-sounding change make the app actually feel more modern. Things change quickly in the world of design conventions.

lognaturel:
Over to you, @seadowg, so QA can take a look? I think their queue is otherwise empty.

collect_app/src/main/res/values/magenta_theme.xml
```diff
@@ -58,8 +60,10 @@
         <item name="android:textDirection">locale</item>
     </style>
 
-    <style name="Widget.Collect.SearchView.Magenta.Theme" parent="Widget.AppCompat.SearchView">
-        <item name="searchHintIcon">@null</item>
+    <style name="Theme.Collect.Magenta.Dialog.Alert" parent="Theme.AppCompat.Light.Dialog.Alert">
```
seadowg: I think we should avoid making any changes to Magenta here as we probably want to look at using [Shape theming](https://material.io/develop/android/theming/shape) for Material Components themes. This would also mean switching back to `MaterialAlertDialogBuilder` but I think the switch to `AlertDialog.Builder` makes our lives much easier for the moment (and is easy to unwind).

I think for me, the next step with Magenta is to use less extreme colors, switch it to being a `DayNight` theme so we get automatic light/dark switching based on the system, add shape theming (to mimic what you've done here) and then replace our light/dark themes entirely. I'll add an issue for this. Could you just remove these changes to Magenta for the moment?

seadowg (changes requested):
This looks ace! Just one comment about removing changes to Magenta for the moment that we should do before merging. This is ready for QA but there is no to test changes in the Magenta theme.

mmarciniak90:
@grzesiek2010, we noticed a few dialogs with an old style.

|Screenshot_20210524-165540 |Screenshot_20210524-165550 | Screenshot_20210524-165659
|---|---|---|
|Screenshot_20210524-165756|Screenshot_20210524-165800|Screenshot_20210524-170048|

grzesiek2010:
It looks like ProgressDialogs don't take the same theme by default I'll try to fix it.
When it comes to the rest of the dialogs those are not mine so I doubt we can fix them, for example, we tried to fix that problem with the dark google account dialog to no avail but I'll take a quick look.

collect_app/src/main/res/values/magenta_theme.xml
grzesiek2010: Ok it's fixed.

grzesiek2010:
@mmarciniak90 @kkrawczyk123 

Ok, so I was able to fix progress dialogs and the dialogs that are displayed to provide more info after denying permissions.

We can't customize permission dialogs so it's not fixed (those dialogs not only have a different shape but also colors):

>Note: Your app cannot customize the dialog that appears when you call requestPermissions(). The text in the system permission dialog references a permission group, but this permission grouping is designed for system ease-of-use. Your app shouldn't rely on permissions being within or outside of a specific permission group.

https://developer.android.com/training/permissions/requesting.html


...and I'm also not able to fix that google account picker dialog.

mmarciniak90:
**Tested with success**

Verified on Android versions: 5.1, 8.1, 10.0, 11.0

Verified cases:

_Admin Settings_
- admin password
- project name
- project icon
- project color
- reset application
- reset results
- delete project
- moving backward disabled/enabled

_General Settings_
- server type
- server url
- username
- password
- form list path
- submission path
- fallback submission URL
- Google account not set dialog
- theme
- language
- test font size
- navigation
- Maps/source
- map style dialog
- offline layer dialog
- blank form update mode
- automatic update frequency dialog
- auto send dialog
- constraint processing behavior dialog
- image size dialog
- show guidance for questions
- form metadata/username
- form metadata/phone number
- form metadata/email address

_Main options_
- connecting to server dialog
- download results dialog
- sending form
- upload result dialog 
- change view dialog
- delete selected dialog

_Dialogs in form_
- saving form dialog
- save and exit dialog 
- change language in form dialog 
- bearing widget
- Range widget with number picker
- exit sketch image widget
- set color
- exit gather signature
- all date widgets 
- location widgets
- input methods on Geotrace and Geoshape dialog
- offline layer on map view
- discard changes on Geotrace and Geoshape
- clear the feature on Geotrace and Geoshape
- rank widgets
- add group dialog
- remove group
- remove response dialog
- audio recording in background

_Other_
- enter admin password
- auto-send results dialog
- edit encrypted form
- field-list not supported in repeats error 
- form list download failed dialog
- server require authentication dialog 
- screen rotation
- light/ dark mode

seadowg (approved):

Issue #4428, opened 2021-02-15
User can add server credentials when adding project


seadowg:
~~Blocked by #4422~~
Mockup [here](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=37%3A219)

### Acceptance

- [x] Given I'm adding a project by manually entering details
When enter a URL
And a username/password
And click "Add"
Then the project shows the username and server host under the project name
And the project has a generated name (see notes)
And the project has a generated icon (see notes)

- [x] Given I'm adding a project by manually entering details
When enter a URL
And I don't enter a username or password
And click "Add"
Then the project shows the server host under the project name
And the project has a generated name (see notes)
And the project has a generated icon (see notes)

### Notes

* Server host should show without protocol, path or query but with subdomain so `https://www.example.com/something?id=123` would show as `www.example.com`.
* The auto-generated project name should be everything in the host name up until the path. For example, `https://www.example.com/something?id=123` would be `www.exampe.com`.
* The auto-generated project icon would just be the first letter of the project name.
* For the moment the project color can be anything. Probably best to just use the default blue we've been using for demo/existing.

seadowg:
@grzesiek2010 right now on master it doesn't look like the settings change when switching between projects - if you change the server in one project and then switch to another you'll see the same server for example. That's fine as that's what I'm working on in #4544, so I've just changed the acceptance here slightly to focus more on the UI around entering project details and then seeing those details in the project list (which will work using `SettingsProvider#getGeneralSettings` with the project ID I imagine).

grzesiek2010:
I think it must be something small because when you switch projects we save `CURRENT_PROJECT_ID` in meta settings then whenever `SettingsProvider` is used it takes that current project id to return proper settings. That should be enough...

grzesiek2010:
ahhh taht's because `providesGeneralSettingsStore` and `providesAdminSettingsStore` in `AppDependencyModule` are singletons, that's the bug.

seadowg:
@grzesiek2010 ah yeah! Looks like I made that change in #4544: https://github.com/getodk/collect/pull/4544/commits/33126eb1ae7c7d1cc41757746e27ed69decf1264#diff-327c65e40a12e80150d6fd6ca4113fc600545c80e650202fdb633e84bd9c32d7L493. That PR can take care of it then.

lognaturel:
[...]

PR #4573, opened 2021-05-24
Collect 4428

grzesiek2010:
Closes #

- [ ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)


grzesiek2010:
@seadowg 
Generally, this pr would be ready but there is one thing I would like to refactor (inconsistency between old project details like url/username/password and new details like project name/icon/color

I keep all those details using Project class:

```
data class Project(
    val url: String,
    val username: String,
    val password: String,
    val name: String,
    val icon: String,
    val color: String,
    val uuid: String = NOT_SPECIFIED_UUID,
)
```

the new details (url/username/password) are not saved to preferences but the old ones are because it was the original behavior. Now I think it would be better to make it consistent and stop storing url/usernam/password in preferences and use ProjectsRepository in both cases.
It will require some refactoring so I wanted to discuss it first. What do you think?

seadowg:
@grzesiek2010 I was thinking we'd avoid having the server details be part of the project itself and just leave them in general settings. It seems simpler to have the server details just live in one place and I think a lot of the code interacting with the server details won't (and shouldn't need to) care about projects.

grzesiek2010:
> @grzesiek2010 I was thinking we'd avoid having the server details be part of the project itself and just leave them in general settings. It seems simpler to have the server details just live in one place and I think a lot of the code interacting with the server details won't (and shouldn't need to) care about projects.

That was my initial idea too but then I needed to for example pass those values together with other details in places like:
- when a new project is added https://github.com/getodk/collect/pull/4573/files#diff-87208d60709e96fbf33666bec56ab9b05b05c7ebc3afeeffcba78219dc800a84R290
- when we pass values to ProjectListItemView https://github.com/getodk/collect/pull/4573/files#diff-bdad71faa755e5bbfd4f0dc9eb421477fedd2cd37bee6b9c4b7534713ac38c22R23
- and maybe other places I don't remember now about or that not exist for now but will be implemented

it's strange to pass a Project and url/username/password separately don't you think? It's more convenient to keep it in one object.

seadowg:
Right I see. Maybe we need to move `AddProjectDialog` back to `collect_app` now that it cares about server stuff? In https://github.com/getodk/collect/pull/4544 I've moved towards `ProjectImporter` handling the storage setup for a project so (once we rebase this on top of that) we could have `ProjectImporter#setupProject` take in the project details (name etc and server details), create the project and set the relevant settings. Does that make sense?

grzesiek2010:
> Does that make sense?

I don't know I need to review your changes first to wrap my head around it but when it comes to moving `AddProjectDialog` back to `collect_app` I also thought about it but then I realized that I need that `onProjectAdded` listener either way (to handle moving from SplashScreenActivity to MainMenuActivity) so passing additional parameters is not a problem (and it's not important where I perform those operations - in AdProjectDialog vs classes that implement that interface).
Even if we do that there will be probably other places where we need to pass both project and url/username/password together.
So you are rather strong against having those values in Project class and think I should don't do that like that at least for now?

seadowg:
> So you are rather strong against having those values in Project class and think I should don't do that like that at least for now?

Yeah, I'm not convinced about moving the server details to live in the `Project` instead and I do feel that it would be more complex to have them in both settings and the project (although I know this isn't what you're suggesting). If I understand correctly,  you're thinking that for the moment we can just have `onProjectAdded` have a signature like:

```kotlin
fun onProjectAdded(Project project, String serverUrl, String serverUsername, String serverPassword)`
```

Is that right? If so, I like that for the moment! The listener implementations can then the project up in Collect - create directories, set settings etc (that's what `ProjectImporter#setupProject` is doing in #4544 really).

grzesiek2010:
Yes, basically that's what I thought. I'll leave it for now and review your pr so that it's merged and we don't have conflicts with this pr.

seadowg:
@grzesiek2010 that sounds great! After we get that in we can hopefully get away from the conflict risks again

lognaturel:
I've been meditating on all this after writing up https://github.com/getodk/collect/issues/4428#issuecomment-848261115 and thinking about what I said at https://github.com/getodk/collect/issues/4579#issuecomment-848114962 about most users not changing the default way a project is presented. 

Running this PR made me realize I made a mistake by not mocking the screen up with the keyboard. Without the keyboard, server info + project display configuration is somewhat digestible, and I didn't realize that it's a lot of prompts without context. I believe that @grzesiek2010 you chose not to pop up the keyboard when the screen first loads to give users a sense of what can be configured. That makes sense but it does mean an extra tap into the URL field which feels unexpected and awkward to me. And then depending on the device, the keyboard likely hides fields that get auto-populated which is extra confusing once the user scrolls down to see them. Or if you're unlucky like me, the keyboard exposes exactly the first three fields and it's very easy to forget there's more below.

I also think you may have decided to do the auto-population of the fields in an attempt to provide some context on what the project configuration values should be. That helps but it still feels surprisingly confusing to me. Where will the name be shown? The icon? Why a color? I said most users won't change these defaults and I really believe they won't have any reason to.

This was very informative to actually try. Having seen it, I propose we only ask for URL/username/password. That's similar to what e.g. Discourse does -- when you create a new account, the first letter of your username is used as your avatar along with some color and then you can choose to configure your avatar differently in settings if you want. That seems like a better user experience. I also mocked it up [here](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=37%3A219) and another advantage is that it gives us a lot more options for Google Drive config, I think.

Happy to hop on a call if y'all aren't sure about this change and/or what I described around fallback/default values.

grzesiek2010:
Hmm I'm not against that change... It would be simpler to have just those three most important fields and then project details generated automatically and editable in admin settings. What about you @seadowg ?

seadowg:
@lognaturel @grzesiek2010 this idea was just popping into my head after I wrote my reply on #4428 (which I've just deleted because it doesn't make any sense anymore) so I think I was on the same journey. @grzesiek2010 I'm going to do the following things:

* Close this PR as I think it's confusing (certainly for me) to have a discussion about a  feature spread across an issue and a PR.
* Move @lognaturel's comment over to the issue as a quote, so it's still clear how we got to where we got
* Rework the issue based on the idea of only having three fields

PR #4581, opened 2021-05-27
User can add server credentials when adding project

### LINKED ISSUE #4428: User can add server credentials when adding project

grzesiek2010:
Closes #4428

No.

No.


projects/src/main/java/org/odk/collect/projects/AddProjectDialog.kt
```diff
@@ -39,17 +40,28 @@ class AddProjectDialog : MaterialFullScreenDialogFragment() {
...
-        binding.projectIconInputText.addTextChangedListener(OneSignTextWatcher(binding.projectIconInputText))
+        binding.urlInputText.doOnTextChanged { text, _, _, _ ->
+            if (Validator.isUrlValid(text.toString())) {
+                binding.urlInputText.error = null
+                binding.addButton.isEnabled = true
+                binding.addButton.alpha = 1f
+            } else {
+                binding.urlInputText.error = getString(R.string.url_error)
+                binding.addButton.isEnabled = false
+                binding.addButton.alpha = 0.3f
```
grzesiek2010: I tried to do that using styling but somehow wasn't able. I think there might be something wrong with it but @seadowg if you want to play with it a bit you can do that.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
```diff
@@ -18,12 +18,11 @@ class AddNewProjectTest {
...
-            .inputProjectName("Project 1")
-            .inputProjectIcon("X")
-            .inputProjectColor("#0000FF")
+            .inputUrl("https://my-server.com")
+            .inputUsername("John")
             .addProject()
 
             .openProjectSettings()
-            .assertInactiveProject("Project 1")
+            .assertInactiveProject("my-server.com", "John / https://my-server.com")
```
seadowg, line 26: I think the URL displayed should really just be "my-server.com" - without the protocol. This is what we have in the mock-ups, and it also doesn't really feel like we need the protocol here for anything.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectListItemView.kt
```diff
@@ -16,10 +15,15 @@ class ProjectListItemView(context: Context, attrs: AttributeSet?) : FrameLayout(
         inflate(context, R.layout.project_list_item, this)
     }
 
-    var project: Project? by Delegates.observable(null) { _, _, new ->
-        if (new != null) {
-            findViewById<ProjectIconView>(R.id.project_icon).project = new
-            findViewById<TextView>(R.id.project_name).text = new.name
+    fun setupView(project: Project?, username: String, url: String) {
```
seadowg: Maybe this method could take in `Project` and `Settings` instead of explicitly taking the username and url? That way the knowledge that we have to grab the username and url out of `Settings` only lives in one place (in `setupView`). What do you think?

seadowg: Oh also, I'm not sure why `project` is a `Project?` here?

projects/src/main/java/org/odk/collect/projects/AddProjectDialog.kt
```diff
@@ -39,17 +40,28 @@ class AddProjectDialog : MaterialFullScreenDialogFragment() {
...
         }
 
         binding.addButton.setOnClickListener {
-            val newProject = Project.New(getProjectName(), getProjectIcon(), getProjectColor())
+            val generatedProjectDetails = ProjectDetailsGenerator.generateProjectDetails(getUrl())
+            val newProject = Project.New(generatedProjectDetails.projectName, generatedProjectDetails.projectIcon, generatedProjectDetails.projectColor)
             val savedProject = projectsRepository.save(newProject)
 
-            listener?.onProjectAdded(savedProject)
+            listener?.onProjectAdded(savedProject, getUrl(), getUsername(), getPassword())
```
seadowg, line 54: I'm wondering if it makes sense to move `AddProjectDialog` back into `collect_app`. That way we can have the code that saves the settings in here as well (I'm sure it will get pulled out somewhere else eventually) and simplifies what has to happen in the listeners (I think just the `MainMenuActivity` would need to set the current project). What do you think?

projects/src/main/java/org/odk/collect/projects/ProjectDetailsGenerator.kt
```diff
@@ -0,0 +1,23 @@
+package org.odk.collect.projects
+
+import java.net.URL
+
+object ProjectDetailsGenerator {
+    fun generateProjectDetails(urlString: String): GeneratedProjectDetails {
```
seadowg: Could this not return a `Project.New` instead? Maybe it's nice to have a different data type here.

projects/src/test/java/org/odk/collect/projects/AddProjectDialogTest.kt
```diff
@@ -50,45 +65,57 @@ class AddProjectDialogTest {
...
-        val application = ApplicationProvider.getApplicationContext<RobolectricApplication>()
-        application.projectsDependencyComponent = DaggerProjectsDependencyComponent.builder()
-            .projectsDependencyModule(object : ProjectsDependencyModule() {
-                override fun providesProjectsRepository(): ProjectsRepository {
-                    return projectsRepository
-                }
-            })
-            .build()
-
+    fun `The 'Add' button should be disabled when url is invalid`() {
```
seadowg: I'm a little uncomfortable with this feature. If our validation is wrong (or just out of date) in any way we could block a user from entering a URL that would work from their device. My preference would be to just show the error but not disable "Add". Disabling when empty feels fine. 

I'm also a little uncomfortable about blocking the user from entering an "invalid" (according to us) email for form metadata. But maybe we should create a separate issue for that.

I'm going to pull in @lognaturel on this as I could be being very paranoid here. Generally I've always felt safe using the assumption that all email and URL validation are wrong no matter how smart or simple.

seadowg (changes requested):
There are some tweaks to the code I'd like us to make here, but it's all feeling really smooth. Liking this flow a lot. It's all starting to feel very real!

projects/src/main/java/org/odk/collect/projects/ProjectDetailsGenerator.kt
grzesiek2010: It doesn't return `Project.New` it returns `GeneratedProjectDetails` so what do you mean?

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
grzesiek2010: So basically this should display the same like project name (if generated automatically)?

lognaturel: Yes, I agree with using just the domain in its entirety. That's the same rule we use for automatically generating project name, exactly.

projects/src/test/java/org/odk/collect/projects/AddProjectDialogTest.kt
lognaturel: > assumption that all email and URL validation are wrong

I unfortunately have to agree with that. I think we're going to move towards doing an actual network request to the server when the URL is entered but we shouldn't do this for the current release. At that point we'll be able to meaningfully reject.

grzesiek2010: So what should I do with the validation here? Validate and display error but not block adding a project as @seadowg said?

lognaturel: I’m happy with either that or just no validation. No validation seems simpler?

grzesiek2010: Yeah probably it doesn't make sense to display that error if we are going to allow using it either way so I'll remove it at all.

Done!

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
grzesiek2010: Ok done.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectListItemView.kt
grzesiek2010: Fixed.

lognaturel:
Sooooo good! 🎉 I agree, the flow is feeling really good. 

Small thing and we can do it as part of polish but I think it'd be even smoother if the keyboard were displayed right away and we didn't have to tap into the URL field.

projects/src/main/java/org/odk/collect/projects/ProjectDetailsGenerator.kt
seadowg: I mean the function could return a `Project.New` instead of a `GeneratedProjectDetails` and then we wouldn't need `GeneratedProjectDetails` at all. Just a thought but I could see there being advantages to the separation (having a different data class for the "generated" details) here.

projects/src/main/java/org/odk/collect/projects/AddProjectDialog.kt
grzesiek2010: You know what... I was thinking about it but I'm not sure I would leave it as it is for now since:
1. it's not something very important
2. it is covered with tests (so even if it's not a perfect solution at least it's testable)

let's see what other prs require maybe it will make it clearer.

projects/src/main/java/org/odk/collect/projects/ProjectDetailsGenerator.kt
grzesiek2010: Oh I see. Yeah maybe it would be easier to return `Project.New`.
Fixed.

projects/src/main/java/org/odk/collect/projects/AddProjectDialog.kt
seadowg: That sounds like it makes sense

seadowg (approved):

PR #4587, opened 2021-05-31
User can override project settings with QR code

### LINKED ISSUE #4426: User can override project settings with QR code
~~Blocked by #4417~~ 

**Acceptance**

When I click "Admin settings"
And I click "Project"
And I click "Configure with QR code"
And I scan a QR code from another project (on a different device)
Then the project settings match those in the QR code

grzesiek2010:
Closes #4426

It has been implemented according to what's described in the issue.

No.

No.


collect_app/src/main/java/org/odk/collect/android/configure/qr/JsonPreferencesGenerator.java
```diff
@@ -36,13 +37,15 @@ private JSONObject getPrefsAsJson(Collection<String> includedPasswordKeys) throw
...
-        for (String key : DEFAULTS.keySet()) {
+        Map<String, ?> generalSettings = settingsProvider.getGeneralSettings().getAll();
+        Map<String, ?> defaultSettings = DEFAULTS;
+        for (String key : defaultSettings.keySet()) {
             if (key.equals(KEY_PASSWORD) && !includedPasswordKeys.contains(KEY_PASSWORD)) {
                 continue;
             }
 
-            Object defaultValue = DEFAULTS.get(key);
-            Object value = settingsProvider.getGeneralSettings().getAll().get(key);
```
seadowg, line 45: Hah took me a minute to spot but yeah, definitely better 😆

collect_app/src/main/java/org/odk/collect/android/preferences/keys/GeneralKeys.java
```diff
@@ -95,7 +95,7 @@
     public static final String KEY_MAGENTA_THEME = "magenta";
     public static final String KEY_EXTERNAL_APP_RECORDING = "external_app_recording";
 
-    private static HashMap<String, Object> getHashMap() {
+    public static HashMap<String, Object> getDefaults() {
```
seadowg, line 98: Definitely prefer this 👍

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
```diff
@@ -66,4 +67,63 @@ class JsonPreferencesGeneratorTest : TestCase() {
...
     }
+
+    @Test
+    fun `Only saved settings should be included in json`() {
+        val generalPrefs = buildMap<String, Any> {
+            put(GeneralKeys.KEY_DELETE_AFTER_SEND, true)
+            put(GeneralKeys.KEY_APP_THEME, "dark_theme")
+        }
+
+        val adminPrefs = buildMap<String, Any> {
```
seadowg: You can just use `mapOf` here instead to avoid the need for the new experimental API:

```kotlin
mapOf(AdminKeys.KEY_GET_BLANK to false, AdminKeys.KEY_DELETE_SAVED to false)
```

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
```diff
@@ -114,6 +118,46 @@ class JsonPreferencesGeneratorTest {
...
+            put(GeneralKeys.KEY_DELETE_AFTER_SEND, true)
+            put(GeneralKeys.KEY_APP_THEME, "dark_theme")
+        }
+
+        val adminPrefsForDemoProject = buildMap<String, Any> {
+            put(AdminKeys.KEY_GET_BLANK, false)
+            put(AdminKeys.KEY_DELETE_SAVED, false)
+        }
+
+        settingsProvider.getGeneralSettings().saveAll(generalPrefsForDemoProject)
```
seadowg, line 125: I'm thinking it might be simpler for `JsonPreferencesGenerator` to take general settings and admin settings in at the constructor rather than a `SettingsProvider` this test class can just create new settings objects (they don't need to be the real ones) and I'm sure you'd need this test.

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
```diff
@@ -66,4 +67,63 @@ class JsonPreferencesGeneratorTest : TestCase() {
         assertThat(jsonPrefs, not(containsString("password")))
         assertThat(jsonPrefs, not(containsString("123456")))
     }
+
+    @Test
+    fun `Only saved settings should be included in json`() {
```
seadowg, line 73: Not to suggest they're not useful, but I'm interested in why you decided to write these tests?

seadowg (changes requested):
A couple of points for discussion but looking good! Did anything actually change behaviourally or is this just code improvements?

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
grzesiek2010: As we discussed during our meeting those tests are for the fix from https://github.com/getodk/collect/pull/4587/commits/7a22d846f9309612b9c54ff5ffe5a2c560fa0694

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
grzesiek2010: Yes but I like more providing `SettingsProvider` because then even if we make a mistake and for example make `JsonPreferencesGenerator` a singleton (it's not necessary a mistake it might be a feature) then it still would work because `JsonPreferencesGenerator` will take proper settings (for the current project, not settings for the project which was the current one when the object was generated)

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
grzesiek2010: Right, fixed.

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
seadowg: Yeah, I take you point. I think it still feels more complex than it needs to be, but it isn't a big deal and as you say could be useful in the future, so I'm happy for it to stick like this.

seadowg (approved):

Issue #4430, opened 2021-02-15
Background jobs for all projects continue to run


seadowg:
Blocked by #4417 

- [x] Given I've set up match exactly in a Project
When I switch to another project
Then the both background jobs continue to work as expected

- [x] Given I've set up match exactly in two Projects
Then the both background jobs (from both projects) continue to work as expected

- [x] Given I've set up auto update in a Project
When I switch to another project
Then the auto update continue to work as expected

- [x] Given I've set up auto update in two Projects
Then the auto update (in both projects) continue to work as expected

- [x] Given I've set up auto send in a project
And I've filled out some forms that haven't sent yet
And have disconnected from the connection that lets auto send happen
When I change projects
And then reconnect to the connection for auto send
Then the forms are sent as expected

- [x] Given I've set up auto update or match exactly in a previous version of Collect
When I upgrade
Then the jobs continue to work as expected

- [x] Given I've set up a project with auto update or match exactly
When I delete the project
Then those jobs don't continue to run in the background

- [x] Given I've set up a project with auto send
And I've filled out some forms that haven't sent yet
And have disconnected from the connection that lets auto send happen
When I delete the project
And then reconnect to the connection for auto send
Then the auto send job doesn't run

- [x] Given I'm launching the app for the first time
When I add a project via a QR code
And the project config includes match exactly
Then the forms update as expected once the project is set up

- [x] Given I'm launching the app for the first time
When I add a project via a QR code
And the project config includes auto update
Then the forms update as expected once the project is set up

## Notes

We could use input data or the tag for to get the project ID to the WorkManager job. We'll also need to make sure that when we cancel jobs they only cancel for that project.

Autosend can run for all projects (that have autosend enabled) whenever it's scheduled if that's easier.

Existing work manager jobs will need to be migrated. It's actually important we do this whenever we change **any** of the background worker classes, so it might just be worth something that always reschedules background work on upgrade  (and cancels old jobs).

lognaturel:
To consider during implementation: should submission and project delete be allowed while auto-send is in progress? 

QA had also highlighted project switching during background jobs as a possible source of weirdness but with the ways we've talked about implementing these I don't think they'd be an issue.

PR #4588, opened 2021-06-01
Fix auto update job so it works with projects

seadowg:
Work towards #4430
~~Blocked by #4587~~

I tried to keep commits small, so it should be possible to go through one by one.

New tests and verified that "Previously Downloaded" only form management worked with multiple projects.

There's probably a couple of major decisions going on here that will hopefully explain the rationale behind some of these changes:

1. **Using `inputData` rather than the tag for specifying the project** - `WorkManager` other tags to the job as well as the ones you set yourself, so it would be some extra malarkey to filter these out and find ours with the embedded ID. Adding input data to our `Scheduler` interface also makes this more extendable for other kinds of "data" that our jobs might need.

2. **Keeping form management use cases away from projects** - I chose to keep objects like `ServerFormDetailsFetcher` and `FormsDirDiskSynchronizer` oblivious of projects and just get passed repositories, directories and settings they need that come from the project. This was purely as it felt silly to propagate the added complexity down to these use case/logic/thing doer pieces of code. This meant moving them out of Dagger and making them more "implementation" details of other parts of the app, but I think that makes sense - they're the builidng blocks of other pieces of logic.

We'll be testing this all later so no reason for a QA pass now.


seadowg:
This is ready for review but includes changes from #4587 (as fixes there made it easier to test this on device).

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.odk.collect.android.storage.StoragePathProvider
+import org.odk.collect.android.storage.StorageSubdirectory
+import org.odk.collect.android.utilities.FormsDirDiskFormsSynchronizer
+import org.odk.collect.android.utilities.FormsRepositoryProvider
+import org.odk.collect.android.utilities.TranslationHandler
+import org.odk.collect.forms.FormSourceException
+import java.io.File
+import java.util.stream.Collectors
+
+class FormUpdateChecker(
```
seadowg, line 21: As discussed on Slack and calls, I'd like to expand on this for other jobs and make it a `FormService`. This would also replace `SyncStatusAppState`.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialog.kt
```diff
@@ -66,7 +65,6 @@ class FirstLaunchDialog : MaterialFullScreenDialogFragment() {
         binding.configureLaterButton.setOnClickListener {
             projectImporter.importDemoProject()
             currentProjectProvider.setCurrentProject(ProjectImporter.DEMO_PROJECT_ID)
-            Collect.resetDatabaseConnections()
```
grzesiek2010, line 69: So we no longer need this? I didn't test it but I continued using it after switching projects in https://github.com/getodk/collect/pull/4591

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
```diff
@@ -0,0 +1,124 @@
...
+            newVersion
+        )
+    }
+}
+
+/**
+ * Allows creating an [SQLiteOpenHelper] that references a `.db` somewhere other than the standard
+ * Android database path (by passing this as [Context] to the constructor).
+ */
+class AltDatabasePathContext(private val path: String, context: Context) : ContextWrapper(context) {
```
grzesiek2010: Why did you decide to keep it here now and not in a separate class like before?

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
```diff
@@ -0,0 +1,100 @@
...
+                    notifier.onUpdatesDownloaded(results)
+                } else {
+                    notifier.onUpdatesAvailable(updatedForms)
+                }
+            }
+
+            context.contentResolver.notifyChange(FormsProviderAPI.CONTENT_URI, null)
+            true
+        } catch (e: FormSourceException) {
+            true
```
grzesiek2010: Shouldn't it return false if something goes wrong?

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
```diff
@@ -0,0 +1,100 @@
...
+    private val notifier: Notifier,
+    private val analytics: Analytics,
+    private val changeLock: ChangeLock,
+    private val storagePathProvider: StoragePathProvider,
+    private val settingsProvider: SettingsProvider,
+    private val formsRepositoryProvider: FormsRepositoryProvider,
+    private val formSourceProvider: FormSourceProvider
+) {
+
+    fun checkForUpdates(projectId: String): Boolean {
```
grzesiek2010: This name is kind of confusing taking into account that this method might just check for updates but also download them. A method that is called like that shouldn't do anything more than just checking if updates are available.
I think it would be better to split it into smaller use cases like `checkForUpdates` and `downloadUpdates` and use them depending on our settings not like now ran the job and then at the last moment decide what to do. Did you think about it?

grzesiek2010 (dismissed):

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FirstLaunchDialog.kt
seadowg: You should be able to get rid of it now in normal code. It's still in tests so that we clean up connections during a full suite run rather than collecting thousands of them.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
seadowg: I originally intended to make this private as I thought it was only used by `DatabaseConnection`, but it does actually get used in a couple of other places (`ItemsetsDbAdapter` and `ExternalSqliteOpenHelper`). I was planning to make it, so those classes use `DatabaseConnection` instead, but in case that isn't something I get to I'll move this back to its own file for the moment.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
seadowg: Yeah, this confused me a little. The existing code returns `true` here though. I think it's because returning `false` in the `TaskSpec` would make the job run again which we don't do for auto update - even if there is a failure we count the job itself as "successful". It is confusing to have it here though. I'll rework this so that `checkForUpdates` is void as it always returns `true` at the moment.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
seadowg: The problem is that they're not additive - we don't check for update **then** download updates, we do one or the other. We could just have two different methods and call different ones from `AutoUpdateTaskSpec` based on the setting, but I'd like to have that decision outside an "Android" component and I don't personally feel like it makes sense to add another layer.

I do think the naming is a bit misleading though (although there are tests for the method that describe the behaviour). I'm thinking I'll rename it `downloadUpdates` (as that's the "primary" use case) and add some JavaDoc.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
lognaturel: I've looked at the way things are structured in this PR, read [the Slack thread](https://getodk.slack.com/archives/C35ENHA64/p1622453471012200) about this, reviewed [the Miro board](https://miro.com/app/board/o9J_lBsLESw=/) and briefly discussed with @seadowg. I think the big questions are whether the various existing functionality actually belongs together, whether there's a risk to introducing a class that could end up as a "dumping ground" (see `FormController`) and whether a `Service` could be confusing given the existence of Android `Service`s.

I see the value of replacing disparate classes that do related things with a single class that has all the right dependencies so I'm in favor of moving in that direction. I think it'd be really helpful to try to draw bounds around what this thing will do either in comments, naming or both. Maybe it's `FormUpdateManager` so it's clearly about the various kinds of updates we support and not about anything that touches forms. @seadowg you mentioned this new abstraction possibly managing the list of forms on the device. I think a `latestLocalFormList` or whatever makes sense in something like a `FormUpdateManager`.

lognaturel (approved):
I've only very quickly gone through this, mostly around [this architectural question](https://github.com/getodk/collect/pull/4588#discussion_r644981160). I've verified that @grzesiek2010 change requests were addressed in what looked like a reasonable way to me.

Let's merge for now so PRs don't need to stack and @grzesiek2010 please take another quick look when you're back and file an issue if you see any problems remaining!

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormUpdateChecker.kt
seadowg: All good points. I'll try and move towards this kind of approach in the next PR (that will look at match exactly) but will stay away from the `XService` because of both the reasons you mention.

lognaturel: Sounds great. One thing I want to bring up explicitly so you can have it in your brain as you do that is that I continue to find the use of `sync` in code for the "match exactly" form update mode confusing. It's not a big deal but perhaps this will be an opportunity to make it more clear that "match exactly" is really nothing other than a type of form update. It's not two directional, it doesn't involve submissions, it doesn't involve settings, etc.

Issue #4584, opened 2021-05-28
Use project name as app title, remove app icon, move Collect version to bottom


lognaturel:
See [forum discussion](https://forum.getodk.org/t/collect-multiple-projects-support/23340/40?u=ln) for discussion. The primary goal is to explicitly place users in the context the most relevant to them -- their project. Secondary goals include moving towards Collect feeling like a system app and reducing the need to fork by making project branding more central.

- [x] Currently there's the Collect/clipboard icon shown in the app bar at the top left. That should be removed.

- [x] Currently the app bar title is ODK Collect vx.y.z. That should be replaced by the current project's name.

- [x] ODK Collect vyyyy.rr.pp should be displayed at the bottom below the buttons and above the hash in case of a non-tagged commit. 

- [ ] Consider removing project switching toast (https://github.com/getodk/collect/pull/4591#discussion_r647616226)

See [mockup](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=462%3A530).

Issue #4579, opened 2021-05-25
Generate project details for existing project


seadowg:
## Acceptance

- [x] Given I've upgraded the app from an older version
When I go to the projects settings screen
Then the current project name has been generated based on the project's server
And the current project icon is the first letter of the project name

## Note

The generated project name should work the same way as the "Add project" form.

seadowg:
@lognaturel we could possibly give everyone an emoji icon (either a fixed one or a random one from a small set) to make it clear that you can use emojis as the icon. Definitely a bit of extra delight but potentially too playful?

lognaturel:
Project name should be the server's full domain name including any subdomains but no paths:
- `https://test.getodk.org/v1/key/jfkdlsajfaioehafnelsajfkdsljfs;jfsa/projects/17` => `test.getodk.org`
- `https://demo.getodk.org` => `demo.getodk.org`
- `https://ona.io/somebody` => `ona.io`

> give everyone an emoji icon

I don't think users will typically change their defaults and we want the default to provide some kind of value. I think it's important to use a letter for that reason:
- `test.getodk.org` => `T`
- `demo.getodk.org` => `D`
- `ona.io` => O

Is default color captured somewhere else?

seadowg:
> Project name should be the server's full domain name including any subdomains but no paths

That's actually a little different from how we described it in https://github.com/getodk/collect/issues/4428. Easy to change though. Maybe better to discuss there as it's where we're first adding the server URL to name mapping in.

> I don't think users will typically change their defaults and we want the default to provide some kind of value.

Yeah convinced! The emoji is nice looking but would be fairly useless for them.

> Is default color captured somewhere else?

I posted another question about that in https://github.com/getodk/collect/issues/4493.

seadowg:
We probably shouldn't work on this yet as it's going to end up conflicting with some of the upgrade code I'm playing with.

PR #4591, opened 2021-06-02
User can add project with QR code

### LINKED ISSUE #4427: User can add project with QR code
~~Blocked by #4422~~

**Acceptance**

- [x] Given I'm on the first launch screen
When I click "Configure with QR code"
And I scan a QR code
Then a new project with the QR code settings is created
And the profile name matchers the server domain
And the profile character is the first character of the name
And the color is the default one

When I click "Add project" 
And I scan a QR code
Then a new project with the QR code settings is created
And the profile name matchers the server domain
And the profile character is the first character of the name
And the color is the default one

When I click "Add project" 
And I click "Manually enter server details"
Then I can create a new project by entering the details manually

grzesiek2010:
Closes #4427 

Screenshot_1623146693

No.

No.


grzesiek2010:
@seadowg as i said I have to review and probably improve some tests but generally the implementation itself should be ready so you can take a look if you want to avoid conflicts in other prs.

collect_app/src/test/java/org/odk/collect/android/projects/AutomaticProjectCreatorDialogTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.odk.collect.android.fragments.support.DialogFragmentHelpers
+import org.odk.collect.android.support.CollectHelpers
+import org.odk.collect.android.support.TestScreenContextActivity
+import org.odk.collect.android.utilities.DialogUtils
+import org.odk.collect.testshared.RobolectricHelpers
+
+@RunWith(AndroidJUnit4::class)
+class AutomaticProjectCreatorDialogTest {
```
grzesiek2010, line 24: This test doesn't use `FragmentScenario` like `ManualProjectCreatorDialogTest` because then something doesn't work. From what I've noticed it's because of the fact that the dialog needs activity https://github.com/getodk/collect/pull/4591/files#diff-78a8fdb05c2a21720fa00d2097e76a82f212c91fa10693d078bd6c9338b0c601R131 and somehow the one provided 

and somehow the one provided by `FragmentScenario` had problems with that.

collect_app/build.gradle
```diff
@@ -189,6 +189,10 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
+    kotlinOptions {
```
seadowg, line 192: What's this for?

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
```diff
@@ -3,26 +3,59 @@ package org.odk.collect.android.feature.projects
...
+    @Test
+    fun addingProjectAutomatically_addsNewProject() {
+        val page = rule.startAtMainMenu()
+            .openProjectSettings()
+            .clickAddProject()
+
+        stubBarcodeViewDecoder.scan("{\"general\":{\"server_url\":\"https:\\/\\/my-server.com\",\"username\":\"adam\",\"password\":\"1234\"},\"admin\":{}}")
+        page.checkIsToastWithMessageDisplayed(R.string.new_project_created)
+
+        MainMenuPage()
```
seadowg, line 56: This needs an `assertOnPage` call I think to make sure we're in the right place

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
```diff
@@ -3,26 +3,59 @@ package org.odk.collect.android.feature.projects
...
     }
+
+    @Test
+    fun addingProjectAutomatically_addsNewProject() {
+        val page = rule.startAtMainMenu()
+            .openProjectSettings()
+            .clickAddProject()
+
+        stubBarcodeViewDecoder.scan("{\"general\":{\"server_url\":\"https:\\/\\/my-server.com\",\"username\":\"adam\",\"password\":\"1234\"},\"admin\":{}}")
+        page.checkIsToastWithMessageDisplayed(R.string.new_project_created)
```
seadowg, line 54: @lognaturel as a side-note it's probably good for us to think about whether we want toasts when creating or switching projects.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/LaunchScreenTest.kt
```diff
@@ -40,4 +55,17 @@ class LaunchScreenTest {
...
+
+    @Test
+    fun clickingAutomaticallyEnterProjectDetails_andScanningQRCode_setsAppUpWithProjectDetails() {
+        val page = rule.startAtFirstLaunch()
+            .clickAutomaticallyEnterProjectDetails()
+
+        stubBarcodeViewDecoder.scan("{\"general\":{\"server_url\":\"https:\\/\\/my-server.com\",\"username\":\"adam\",\"password\":\"1234\"},\"admin\":{}}")
+        page.checkIsToastWithMessageDisplayed(R.string.new_project_created)
+
+        MainMenuPage()
```
seadowg, line 67: This also needs an `assertOnPage`.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
```diff
@@ -49,15 +55,11 @@ class ProjectImporter(
...
-        return project
-    }
-
-    fun setupProject(project: Saved) {
         createProjectDirs(project)
+        return project
     }
 
-    private fun createProjectDirs(project: Saved) {
+    fun createProjectDirs(project: Saved) {
```
seadowg: I think this should be private right? It looks like a test helper calls it, but it could just use `importNewProject` instead I think.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectCreator.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val projectsRepository: ProjectsRepository,
+    private val currentProjectProvider: CurrentProjectProvider,
+    private val settingsImporter: SettingsImporter,
+) {
+
+    fun createNewProject(settingsJson: String): Boolean {
+        val newProject = getNewProject(settingsJson)
+        val savedProject = projectImporter.importNewProject(newProject)
+
+        val settingsImportedSuccessfully = settingsImporter.fromJSON(settingsJson, savedProject.uuid)
```
seadowg, line 22: I'm having a little trouble with the seam between `ProjectCreator` and `ProjectImporter`. My reading is:

* `ProjectCreator` is responsible for creating a project based on JSON and setting it up as the current project if needed
* `ProjectImporter` is responsible for setting up all the auxiliary things about a project (directories etc)

Is that right? If that's the case, I feel that `ProjectCreator` should probably perform the `ProjectsRepository#save` and then pass the project along with the JSON settings to the project importer, which should then import the settings from JSON and create directories. I think in that case we'd actually only have a `importProject` method on `ProjectImporter` and `ProjectCreator` could have methods with hardcoded values for creating a demo project and existing project. Now I'm looking at all this, I think the `importExistingProject` code should probably live in a different object (`ExistingProjectImporter` for instance) as that's only really needed once and will go away. 

How does that kind of structure sound? I'll pause my review here, as that might move things around a little.

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
lognaturel: I think the "project successfully added" feedback is nice. Since we're not switching to the project that was just created, it is unclear whether anything happened. Probably what we'd want to work towards is a snackbar with a "Switch" action or something like that but I think for now we can go with this toast.

I'm less convinced by the project switching toast. I think once the app bar has the project title (#4584) it will feel redundant. Let's leave it for now and see how it feels with #4584.

collect_app/build.gradle
grzesiek2010: I added this to use `doOnTextChanged` in https://github.com/getodk/collect/pull/4591/files#diff-651f8f96000eb2c3ca961c1f82e7a5300b5cc835f97d7ee77762fe83048eeebeR48

which is an extension function and a nice replacement for `addTextChangedListener` (`beforeTextChanged`, `onTextChanged`, `afterTextChanged`) which requires overriding three methods what doesn't make sense if you need just one.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectCreator.kt
grzesiek2010: > Is that right?

mostly yeah... `ProjectCreator` does something more than `ProjectImporter`, it's responsible for:
1. Creating new project (project details) based on its url where `ProjectImporter` just takes it as a parameter
2. Calling ProjectImporter.importNewProject() passing the generated project
3. Saving settings for the new project
4. Setting the current project if needed.

I'm not sure which structure we should follow. I would like to leave it as is and get back to it implementing https://github.com/getodk/collect/issues/4579 because then we are going to implement generating details for all cases:
- importing existing project
- importing demo project
- importing projects scanned via QR code
- importing projects added manually

so then there will be no cases where we just save a project to the repository and create dirs (like in those two first cases listed above), we will always need to do pretty the same:
- generate project details
- save project to the repository
- create dirs
- import settings if needed
- sett current project id if needed

so then we I think we will need just one creator/importer do those classes will be merged or one of them removed but this is something I would rather think later as I said working on https://github.com/getodk/collect/issues/4579

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/AddNewProjectTest.kt
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/LaunchScreenTest.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectCreator.kt
seadowg: Hmmm yeah that makes sense. I think I'll end up pulling out the migration stuff and needing to tweak settings import as part of the background jobs (for importing configs with match exactly for instance and migrating existing jobs) so that probably moves closer to this all as well. I'll make sure to rebase on top of this so that we're working in the same direction!

collect_app/build.gradle
seadowg: Ah sorry I didn't see what you mean at first but you needed it because you can't use an `inline` extension for Java 8 code while targeting 6 with Kotlin right?

collect_app/src/test/java/org/odk/collect/android/configure/SettingsImporterTest.java
```diff
@@ -48,7 +48,7 @@ public void setup() {
         settingsValidator = mock(SettingsValidator.class);
         when(settingsValidator.isValid(any())).thenReturn(true);
 
-        importer = new SettingsImporter(generalSettings, adminSettings, (Settings generalSettings, Settings adminSettings) -> {}, settingsValidator, generalDefaults, adminDefaults, (key, newValue) -> {});
+        importer = new SettingsImporter(settingsProvider, (Settings generalSettings, Settings adminSettings) -> {}, settingsValidator, generalDefaults, adminDefaults, (key, newValue) -> {});
```
seadowg, line 51: Ah this is a nice change as we'll need the project ID for setting up jobs on import.

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
```diff
@@ -0,0 +1,79 @@
...
+    private val settingsProvider: SettingsProvider,
+    private val preferenceMigrator: SettingsPreferenceMigrator,
+    private val settingsValidator: SettingsValidator,
+    private val generalDefaults: Map<String, Any>,
+    private val adminDefaults: Map<String, Any>,
+    private val settingsChangedHandler: SettingsChangeHandler
+) {
+
+    @JvmOverloads
+    fun fromJSON(json: String, uuid: String? = null): Boolean {
```
seadowg: I think we should use `projectId` here rather than `uuid` so we know what the ID refers to. I've been think it also makes sense to rename the field on project, but holding off on doing that. The fact that the ID is a UUID isn't really as important outside the repo itself.

collect_app/src/main/java/org/odk/collect/android/projects/AutomaticProjectCreatorDialog.kt
```diff
@@ -0,0 +1,163 @@
...
+        return binding.toolbar
+    }
+
+    private fun setUpToolbar() {
+        toolbar.setTitle(R.string.add_project)
+        toolbar.navigationIcon = null
+    }
+
+    private fun startScanning(savedInstanceState: Bundle?) {
+        val intent = IntentIntegrator(requireActivity())
```
seadowg: Are we able to wrap all the shared (between this and `BarcodeScannerFragment`) setup for the barcode scanning? Even if it's not all tested, it would be great to have it isolated, so it's easy to debug inspect in one place.

collect_app/src/test/java/org/odk/collect/android/projects/AutomaticProjectCreatorDialogTest.kt
seadowg: Actually extracting the barcode setup stuff (like I suggest in another comment) might allow you to then stub it here so that you don't have to worry about the barcode integration in tests.

collect_app/src/main/res/layout/automatic_project_creator_dialog_layout.xml
```diff
@@ -0,0 +1,104 @@
...
+                android:layout_height="wrap_content"
+                android:text="@string/or"
+                android:layout_marginTop="@dimen/margin_standard"
+                style="@style/TextAppearance.Collect.Subtitle1"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@+id/barcode_view" />
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/configure_manually_button"
```
seadowg, line 72: Could we introduce a [theme overlay](https://medium.com/androiddevelopers/android-styling-themes-overlay-1ffd57745207) for `OutlinedButton` (like the ones in `material_theme_polyfill`) to solve the color problems here? When we switch to a MaterialComponents theme/proper colors it will be more awkward if we have color customizations on individual views.

seadowg (changes requested):
OK I took a look at the rest now. Thanks for making the changes I already suggested! Just a couple of tweaks and one thing I think we should clean up while we're here.

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
grzesiek2010: Fixed.

collect_app/src/main/res/layout/automatic_project_creator_dialog_layout.xml
grzesiek2010: Actually that color is not needed and I removed it.

collect_app/src/main/java/org/odk/collect/android/projects/AutomaticProjectCreatorDialog.kt
grzesiek2010: Done: https://github.com/getodk/collect/pull/4591/commits/a3a164b142efe1e10a2119a9fdf2e58c89a6c367

collect_app/build.gradle
grzesiek2010: Yes

seadowg (approved):

PR #4590, opened 2021-06-02
Cancel auto update job when deleting project

seadowg:
Work towards #4430
~~Blocked by #4588~~ 

#### What has been done to verify that this works as intended?

New tests.

#### Why is this the best possible solution? Were any other approaches considered?

Nothing to discuss here really. Cancellation lives next to scheduling and I had to pull out a small use case for project deletion to let me write a unit test to drive the cancellation call out.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Like most of these PRs we'll skip QA on this and come back to it later.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

PR #4593, opened 2021-06-04
Fix match exactly for projects

seadowg:
Work towards #4430
~~Blocked by #4590~~

This makes it possible to use Match Exactly with projects. Deleting a project will also clean up the match exactly job.

This is mostly pretty similar to the previous changes made for auto-update/previously downloaded. The headline changes are:

* The change locks we use for forms are now project specific. This means that match-exactly/previously downloaded jobs won't run at the same time for the same project, but can run at the same time for different projects. This change will be made for instances change locks when we look at auto send.
* The `FormUpdateChecker` is now the `FormUpdater` and handles both the match-exactly and previously downloaded logic. I thought about dissolving the `SyncStatusAppState` class into `FormUpdater` as well but I think we're still going down a path of separated logic/state for this grander "app" layer of Collect so wanted to avoid a sudden about turn there. If it seems better to folks we could now just separate the `FormUpdater` into two objects and just share the dependency management logic it has (for grabbing project related deps). I think we should see where go with auto send before making that call though.

We'll look at QA for all this later.


collect_app/src/main/java/org/odk/collect/android/formmanagement/FormsUpdater.kt
```diff
@@ -0,0 +1,182 @@
...
+        changeLockProvider,
+        formSourceProvider
+    )
+}
+
+/**
+ * Provides all the basic/building block dependencies needed when performing logic inside a
+ * project.
+ */
+private class ProjectSandbox(
```
seadowg, line 133: I could see this ending up being a convenience we share in a few places (I think we'll need it almost immediately for auto-send for instance). This could also be implemented with a Dagger component (that we just build on the fly). But personally I think I'm happy to avoid Dagger in places it's not absolutely needed (i.e. where we have a constructor).

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormsUpdater.kt
```diff
@@ -0,0 +1,182 @@
...
+        val serverFormsSynchronizer = ServerFormsSynchronizer(
+            serverFormsDetailsFetcher,
+            sandbox.formsRepository,
+            sandbox.instancesRepository,
+            formDownloader
+        )
+
+        return sandbox.formsLock.withLock { acquiredLock ->
+            if (acquiredLock) {
+                syncStatusAppState.startSync()
```
seadowg, line 97: I realized that we need to rework this so that the "syncing" state works on a per-project basis. I've added this as work to do post beta. The worst case right now is that someone won't be able to manually refresh their project while another one is syncing in the background - fine for beta, but we need to fix before launch.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormsUpdater.kt
```diff
@@ -0,0 +1,182 @@
...
+import org.odk.collect.android.utilities.ChangeLockProvider
+import org.odk.collect.android.utilities.FormsDirDiskFormsSynchronizer
+import org.odk.collect.android.utilities.FormsRepositoryProvider
+import org.odk.collect.android.utilities.InstancesRepositoryProvider
+import org.odk.collect.android.utilities.TranslationHandler
+import org.odk.collect.forms.FormSourceException
+import java.io.File
+import java.util.stream.Collectors
+
+class FormsUpdater(
```
lognaturel, line 25: Really like how this turned out and the naming seems so obvious now seeing it! 🎉

lognaturel:
@seadowg asked me to think a little bit about the locks and whether making them per-project made the most sense. I think so and I'm glad you've run with it @seadowg, sorry about not giving quicker feedback. Good catch regarding manual updates in one project being blocked by automatic updates for another project.

The alternative mentioned was only letting one automatic job run at once and ensuring that jobs that get blocked for this reason can rerun later. As a user, this would feel unexpected to me. I expect each project to be independent from others and that would include updates possibly happening at the same time if that's how they end up being scheduled. That said, I don't think the difference would be very visible to users. From a developer perspective I think it's also more consistent for projects to act independently in this way.

The one possible issue I see with per-project locks is in a case where multiple projects are configured against the same server. Per-project locks mean a single device could make multiple parallel requests to the same server. Multiply that across a bunch of data collectors and it could be more than the server can handle. I brought this up in the context of issues around `ReferenceManager` being a singleton (to be addressed after beta) but @seadowg didn't seem particularly concerned. I've come around to feeling like it's not a very big deal. The server will just eventually drop requests and the updates will retry. I believe in that process they'll then become staggered for the next time (sound right, @seadowg?). Regardless, it's an issue we can deal with if we find out it's one users run into.

Only letting one automatic job run at once seems like it could help with the `ReferenceManager` issue but it wouldn't because `ReferenceManager` must be set for form entry as well as for form updates. So we have to deal with that separately no matter what.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    val customSplashScreenImagePath: String
+        get() = odkRootDirPath + File.separator + "customSplashScreenImage.jpg"
+    val tmpImageFilePath: String
+        get() = getOdkDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.jpg"
+    val tmpVideoFilePath: String
+        get() = getOdkDirPath(StorageSubdirectory.CACHE) + File.separator + "tmp.mp4"
+
+    fun getRelativeInstancePath(filePath: String?): String? {
```
grzesiek2010: Those are single-expression functions so you could omit brackets. Don't you like it?

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+                projectId
+            ) + File.separator + subdirectory.directoryName
+            StorageSubdirectory.PROJECTS -> odkRootDirPath + File.separator + subdirectory.directoryName
+            else -> throw IllegalStateException("Unexpected value: $subdirectory")
+        }
+    }
+
+    val customSplashScreenImagePath: String
+        get() = odkRootDirPath + File.separator + "customSplashScreenImage.jpg"
+    val tmpImageFilePath: String
```
grzesiek2010: I would rather have blank lines between those vals it's more difficult to look at them when it's one block of code.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+        return getOdkDirPath(subdirectory, null)
+    }
+
+    fun getOdkDirPath(subdirectory: StorageSubdirectory, projectId: String?): String {
+        return when (subdirectory) {
+            StorageSubdirectory.FORMS, StorageSubdirectory.INSTANCES, StorageSubdirectory.CACHE, StorageSubdirectory.METADATA, StorageSubdirectory.LAYERS, StorageSubdirectory.SETTINGS -> getProjectRootDirPath(
+                projectId
+            ) + File.separator + subdirectory.directoryName
+            StorageSubdirectory.PROJECTS -> odkRootDirPath + File.separator + subdirectory.directoryName
+            else -> throw IllegalStateException("Unexpected value: $subdirectory")
```
grzesiek2010: You don't need this else since you listed all elements from `StorageSubdirectory` which is enum.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    fun getOdkDirPath(subdirectory: StorageSubdirectory): String {
+        return getOdkDirPath(subdirectory, null)
+    }
+
+    fun getOdkDirPath(subdirectory: StorageSubdirectory, projectId: String?): String {
+        return when (subdirectory) {
+            StorageSubdirectory.FORMS, StorageSubdirectory.INSTANCES, StorageSubdirectory.CACHE, StorageSubdirectory.METADATA, StorageSubdirectory.LAYERS, StorageSubdirectory.SETTINGS -> getProjectRootDirPath(
```
grzesiek2010: This is difficult to read. I would rather format it like:
```kotlin
            StorageSubdirectory.FORMS, 
            StorageSubdirectory.INSTANCES, 
            StorageSubdirectory.CACHE, 
            StorageSubdirectory.METADATA, 
            StorageSubdirectory.LAYERS, 
            StorageSubdirectory.SETTINGS -> getProjectRootDirPath(projectId) + File.separator + subdirectory.directoryName
            StorageSubdirectory.PROJECTS -> odkRootDirPath + File.separator + subdirectory.directoryName
```

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+    fun getProjectRootDirPath(projectId: String?): String {
+        return if (projectId == null) {
+            val currentProjectId = currentProjectProvider.getCurrentProject().uuid
+            getOdkDirPath(StorageSubdirectory.PROJECTS) + File.separator + currentProjectId
+        } else {
+            getOdkDirPath(StorageSubdirectory.PROJECTS) + File.separator + projectId
+        }
+    }
+
+    fun getOdkDirPath(subdirectory: StorageSubdirectory): String {
```
grzesiek2010: You can remove this method adding default null to the one below (+@JvmOverloads)

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+        return when (subdirectory) {
+            StorageSubdirectory.FORMS, StorageSubdirectory.INSTANCES, StorageSubdirectory.CACHE, StorageSubdirectory.METADATA, StorageSubdirectory.LAYERS, StorageSubdirectory.SETTINGS -> getProjectRootDirPath(
+                projectId
+            ) + File.separator + subdirectory.directoryName
+            StorageSubdirectory.PROJECTS -> odkRootDirPath + File.separator + subdirectory.directoryName
+            else -> throw IllegalStateException("Unexpected value: $subdirectory")
+        }
+    }
+
+    val customSplashScreenImagePath: String
```
grzesiek2010: I would move those vals above so that they are not mixed with functions because now we have some vals then functions then vals and again functions.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseObjectMapper.kt
```diff
@@ -20,11 +22,17 @@ object DatabaseObjectMapper {
         values.put(DatabaseFormColumns.DESCRIPTION, form.description)
         values.put(DatabaseFormColumns.JR_FORM_ID, form.formId)
         values.put(DatabaseFormColumns.JR_VERSION, form.version)
-        values.put(DatabaseFormColumns.FORM_FILE_PATH, StoragePathProvider.getRelativeFilePath(formsPath, form.formFilePath))
+        values.put(
```
grzesiek2010: I don't know if you like it or not but I usually put such pure code formatting in separate commits named `code formatting` / `code improvements`. I think it's easier for reviewers if pure formatting is separated from more relevant changes. Otherwise you might end up (what in fact is a common situation) creating a commit with 100 lines of formatting and 5 lines of really important changes and it's easy to ignore those changes.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtil.java
```diff
@@ -1,22 +0,0 @@
-package org.odk.collect.android.utilities;
-
-import java.io.File;
-import java.util.ArrayList;
-import java.util.List;
-
-import static java.util.Arrays.asList;
-
-public class FileUtil {
```
grzesiek2010, line 9: Finally :smile:

shared/src/main/java/org/odk/collect/shared/PathUtils.kt
```diff
@@ -0,0 +1,22 @@
+package org.odk.collect.shared
+
+import java.io.File
+
+object PathUtils {
+
+    @JvmStatic
+    fun getRelativeFilePath(dirPath: String, filePath: String?): String? {
+        if (filePath == null || filePath.isEmpty()) {
```
grzesiek2010: This should not happen but maybe we should use `isBlank()` instead of `isEmpty()` to make this method even more secure? 
(the same below)

config/pmd-ruleset.xml
```diff
@@ -92,6 +92,7 @@
         <exclude name="TooManyMethods" />
         <exclude name="UseObjectForClearerAPI" />
         <exclude name="AvoidUncheckedExceptionsInSignatures" />
+        <exclude name="ExcessiveParameterList" />
```
grzesiek2010, line 95: If you exclude this rule you can remove all `@SuppressWarnings("PMD.ExcessiveParameterList")` that we have in our codebase.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormDeleter.java
```diff
@@ -33,7 +29,5 @@ public void delete(Long id) {
         } else {
             formsRepository.softDelete(form.getDbId());
         }
-
-        fastExternalItemsetsRepository.deleteAllByCsvPath(form.getFormMediaPath() + File.separator + "itemsets.csv");
```
grzesiek2010, line 37: Are we going to handle this in the future? Any issue for that?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
seadowg: You mean just using the `=` operator to denote the function body? I don't really have a strong preference between those two styles. With this function, the body would always end up on the next line anyway as auto formatting would push it there due to the line length, so I guess I feel like having and curly braces and an explicit type helps readability?

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
seadowg: Yeah this code is auto converted to Kotlin and I definitely should have changed this. Will fix.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseObjectMapper.kt
seadowg: Yeah I agree, and I'll try and separate formatting existing code from new things in my commits more.

One thing to note is I'm just using Android Studio's "Reformat Code", (CMD+OPTION+L on Mac) which I basically hit after I write any code from muscle memory now. The default settings (which come from the [Kotlin Style Guide](https://kotlinlang.org/docs/coding-conventions.html)) seem to be much more aggressive than the Java ones, so there is much likelihood of adding a small piece of code to a file, autoformatting and seeing a lot of change if the file wasn't formatted correctly before. It seems the main culprit is max line length which Android Studio automatically reformats but ktLint seem to care about. I'll have a quick look at adding a rule for ktLint to fail on max line length, so it's less likely that we commit or merge incorrectly formatted code.

seadowg: Ah ok. The problem is that we're not using the [Android Kotlin Style Guide](https://developer.android.com/kotlin/style-guide) with ktlint but we can easily configure it use that instead of the Jetbrains style guide. Then it should match what Android Studio auto formats to (by default). I'll make another PR for that as it'll introduce a bunch of formatting.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
seadowg: Interestingly, it wants to format them like that. I really don't like the property syntax anyway so I'll just remove the `get()` or make them functions.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtil.java
seadowg: Haha yeah I'm so glad I had the opportunity to fix this.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormDeleter.java
seadowg: We could create an issue, but when @lognaturel and I talked it about it, it seemed like it would be a "Won't fix" anyway. @lognaturel is this something we'd prioritize in the next couple of years?

config/pmd-ruleset.xml
seadowg: Good point. I'll do that now.

shared/src/main/java/org/odk/collect/shared/PathUtils.kt
seadowg: Looking at it now, I'm actually wondering if we should just remove the check. I think it would a coding error for us to pass `null` or `""` into this, and I can't see anywhere where we'd genuinely want to return `null` from either of these (that's also the case for the `StoragePathProvider` functions that make heavy use of these helpers). I'd imagine if the logic in this method doesn't explode, returning `null` will (like passing it to a `File` constructor). 

What do you think? Should I go ahead and make them return `String` rather than `String?`.

seadowg: I've made the change but can easily revert the last two commits if you don't think it makes sense.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormDeleter.java
lognaturel: I’d be surprised if it’s something we three prioritize but it’s not a bad idea to have an issue so that if a user runs into problems it’s easy to find and so that a contributor can possibly take it on.

seadowg: Ok issue [here](https://github.com/getodk/collect/issues/4599).

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
grzesiek2010: Yes that's what i meant but we can leave it as-is I'm not strongly in favor of either approach.

grzesiek2010 (approved):

PR #4595, opened 2021-06-08
Use correct signature for openOrCreateDatabase to prevent crashes

seadowg:
This fixes an issue where some older versions of Android pass an optional value as `null` to a method we'd overriden but made the parameter non-null.

@grzesiek2010 can you verify this fixes the issue on your device?

Annoyingly there isn't really a way to ensure this doesn't come up again. I'll file a bug with the Android team around `ContextWrapper` not having correct annotations on `openOrCreateDatabase` (which would prevent Kotlin conversion going wrong here).

Just fixes the issue. No need for QA.

### DIFF
--- a/download-robolectric-deps.sh
+++ b/download-robolectric-deps.sh
@@ -1,3 +1,5 @@
+set -e
+
 wget -nc https://repo1.maven.org/maven2/org/robolectric/android-all/6.0.1_r3-robolectric-r1/android-all-6.0.1_r3-robolectric-r1.jar -P robolectric-deps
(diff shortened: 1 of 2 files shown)


seadowg:
Android issue is here: https://issuetracker.google.com/issues/190428215

download-robolectric-deps.sh
```diff
@@ -1,3 +1,5 @@
+set -e
```
seadowg, line 1: We were getting test failures in Robolectric because downloading these JARs failed (due to the outages across the internet during Europe's AM) but didn't stop the build at the `compile` stage.

grzesiek2010 (approved):

PR #4596, opened 2021-06-08
Fix auto send for projects and allow importing of QR codes with form update mode set

seadowg:
Work towards #4430
~~Blocked by #4593~~

For auto-send, the approach I've taken has really ended up being a spike. There is a lot of work to be done in the instance submission logic to pull out testable interfaces and I don't think this is the right time to embark on that journey given the sheer amount of rework we've had to do already to get projects working. Auto send has coverage at the Espresso level and I think we'll just need to highlight this feature as a risk to QA. My hope is that we can spend some time restructuring the auto-send logic when we inevitably need to remove the `AsyncTask` it's wrapped in.

No big decisions here other than the test coverage discussed above. Auto-follows largely the same approach as we've used in the other background jobs, and the import changes were pretty minimal - we just needed to add a project ID to our change handler.

We'll look at QA for all these PRs later.


collect_app/src/main/java/org/odk/collect/android/preferences/screens/BaseAdminPreferencesFragment.java
```diff
@@ -17,6 +18,9 @@
     @Named("ADMIN_SETTINGS_STORE")
     SettingsStore adminSettingsStore;
 
+    @Inject
+    CurrentProjectProvider currentProjectProvider;
```
grzesiek2010: You can move it one level up to `BasePreferencesFragment` since it's used in both `BaseAdminPreferencesFragment` and `BaseGeneralPreferencesFragment`.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/FormUpdateScheduler.java
```diff
@@ -2,7 +2,7 @@
 
 public interface FormUpdateScheduler {
 
-    void scheduleUpdates();
+    void scheduleUpdates(String s);
```
grzesiek2010: Why `s` not `projectId`?

collect_app/src/main/java/org/odk/collect/android/backgroundwork/FormUpdateScheduler.java
```diff
@@ -2,7 +2,7 @@
...
 
-    void cancelUpdates();
+    void cancelUpdates(String s);
```
grzesiek2010: Same as above.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/backgroundwork/FormUpdateScheduler.java
seadowg: Ooops. Probably added this with refactoring tools but forgot to rename.

grzesiek2010:
@seadowg you have some conflicts

grzesiek2010 (approved):

PR #4597, opened 2021-06-09
Remove support for form list responses without form hashes and without OpenRosa headers

### LINKED ISSUE #4450: Remove support for legacy formList API and for incomplete OpenRosa formList responses
Revert the patch from https://github.com/getodk/collect/pull/4449 and show an error message on `NullPointerException` from hash.

Remove `OpenRosaFormSource.parseLegacyFormList` and all related code. Show a generic error about an unrecognized server response. In the 28 days Jan 27 - Feb 23, 29 users made 99 requests with legacy responses.

lognaturel:
Closes #4450 
As announced [on the forum](https://forum.getodk.org/t/breaking-changes-coming-to-collect-v2021-2/33169).

This would be good to get in any beta we do to give folks a chance to see the change in action.

Usage of the legacy API has gone up over the past few months ([analytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/events/~2Freport~2FLegacyFormList%3Ft%3D1623196424583&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_r..eventId%253DLegacyFormList%2526_u.dateOption%253Dlast90Days&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1)). That means this change will be more disruptive than we'd hoped but it's still a very small fraction of users. I see this increase in usage as a sign that we need to make the deprecation sooner than later or else folks will keep on developing against the legacy API.

Added and ran tests. Manually verified the happy path against the demo server.

- I initially wasn't going to keep the smoke test for the missing hash case but I decided to do so because it's the only place we can test that the message is correctly shown in the manual form download case.
- It's really important to show a specific message in the missing hash case. I decided to do this by exposing a message from `FormDownloadException` if one exists. It's not the cleanest thing but this is an area we know we need to come back to sooner than later because we just show the unhelpful "Failure" for every other failure case. What I did seems like a reasonable interim solution. I only exposed it for the manual download case because I imagine folks with custom servers that don't follow ODK development don't use update features.
- I decided to remove analytics for the legacy API. We're committed to removing support for it and I don't think tracking attempts to use it would affect our behavior.

I think this is relatively low-risk overall. The biggest source of risk I can think of is if I didn't think of something related to background form update cases. For example, I was initially thinking of passing on the `IllegalArgumentException` from `getByMd5Hash` for the missing hash case but that would have required handling that exception type for all three paths to downloading forms (and I think `FormDownloadException` is more appropriate after all.

No. I don't think this should go through QA since it needs a badly-behaved server to verify. The non-error cases will be verified as we continue development and will be looked at in the regression pass.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
@@ -15,7 +15,6 @@ import org.odk.collect.android.openrosa.OpenRosaHttpInterface;
 import org.odk.collect.android.support.CollectTestRule;
 import org.odk.collect.android.support.ResetStateRule;
 import org.odk.collect.android.support.StubOpenRosaServer;
-import org.odk.collect.android.support.pages.MainMenuPage;
 import org.odk.collect.utilities.UserAgentProvider;
 
 @RunWith(AndroidJUnit4.class)
@@ -42,7 +41,9 @@ public class BadServerTest {
             .around(rule);
 
     @Test
-    public void whenHashNotIncludedInFormList_canGetBlankForm_fillItIn_andSubmit() {
+    // The hash from the form list wasn't used for a long time so some server implementations omitted it even though
+    // it's required by the spec. Now we explicitly show an error.
+    public void whenHashNotIncludedInFormList_showError() {
         server.removeHashInFormList();
         server.addForm("One Question", "one-question", "1", "one-question.xml");
 
@@ -50,16 +51,6 @@ public class BadServerTest {
                 .setServer(server.getURL())
                 .clickGetBlankForm()
                 .clickGetSelected()
-                .assertText("One Question (Version:: 1 ID: one-question) - Success")
-                .clickOK(new MainMenuPage())
-
-                .startBlankForm("One Question")
-                .swipeToEndScreen()
-                .clickSaveAndExit()
-
-                .clickSendFinalizedForm(1)
-                .clickOnForm("One Question")
-                .clickSendSelected()
-                .assertText("One Question - Success");
+                .assertText("One Question (Version:: 1 ID: one-question) - Missing form hash. ODK-compatible servers must include form hashes in their form lists. Please talk to the person who asked you to collect data.");
     }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/StubOpenRosaServer.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/StubOpenRosaServer.java
@@ -140,7 +140,6 @@ public class StubOpenRosaServer implements OpenRosaHttpInterface {
         noHashInFormList = true;
     }
 
-
     public String getURL() {
         return "https://" + HOST;
     }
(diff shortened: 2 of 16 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
```diff
@@ -42,24 +41,16 @@ public OpenRosaHttpInterface provideHttpInterface(MimeTypeMap mimeTypeMap, UserA
...
-
-                .startBlankForm("One Question")
-                .swipeToEndScreen()
-                .clickSaveAndExit()
-
-                .clickSendFinalizedForm(1)
-                .clickOnForm("One Question")
-                .clickSendSelected()
-                .assertText("One Question - Success");
+                .assertText("One Question (Version:: 1 ID: one-question) - Missing form hash. ODK-compatible servers must include form hashes in their form lists. Please talk to the person who asked you to collect data.");
```
seadowg, line 54: This should probably use the (translated) string resource for the assertion so that the test doesn't fail if there is a tweak to the language.

collect_app/src/androidTest/java/org/odk/collect/android/support/StubOpenRosaServer.java
```diff
@@ -140,6 +141,10 @@ public void removeHashInFormList() {
         noHashInFormList = true;
     }
 
+    public void omitOpenRosaHeader() {
```
seadowg: This isn't being used anywhere as far as I can see?

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
lognaturel: https://github.com/getodk/collect/pull/4597/files#diff-bf920b275657bdd44fb5321255809b7ab723c2a2fe66b89706efd95ac5f3a7e9R82 -- it's not translated. I meant to bring this up in the PR description but now see it's only in the commit details. This is going to be short lived and it targets just a couple of rogue servers so I'd rather not have translators deal with it. This is especially important since it's really a message aimed at server developers. "form hash" probably shouldn't be translated, for example.

collect_app/src/androidTest/java/org/odk/collect/android/support/StubOpenRosaServer.java
lognaturel: Whoops, thank you! I incompletely backed out of having an Espresso test for the legacy form list case. Will fix.

collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
lognaturel: For the legacy formlist request, I used the translated string for "Please talk..." mostly because that one will be more common. I can do this here if you prefer.

seadowg: Ah right sorry I missed that in the PR. Thinking about this a bit more, I feel a bit uncomfortable with the `DatabaseFormRepository` declaring the message. We'll need to readdress this whenever we move the `DatabaseFormsRepository` over to Kotlin as the method will no longer accept `null` at all after that, and that will push the message up. I think I'm happy to leave this in right now and avoid making changes to the use cases where we're more likely to see conflicts.

lognaturel: Sounds good. We're going to have to rethink all of this sooner than later anyway because the generic `Failure` message in all other cases is causing user pain. I agree this is temporary.

seadowg (approved):

PR #4598, opened 2021-06-09
Log analytics on return from OpenMapKit app

### LINKED ISSUE #4466: Add analytics for OSM question type
Follow up from conversation on #4033. We don't know how much this functionality is used. I think the ideal is to track every time that a value is returned from the OSM app to make sure it's a real usage. We should track the form hash with that.

lognaturel:
Closes #4466

Nothing. 😬

The other (likely more) logical place to put this would be `OSMWidget.setData`. However, this is not mission-critical and might not stay in for very long so I'd prefer to just get it in there without adding new dependencies to the widget or worrying about testing. I think that's reasonable but if you don't, let me know.

No risk to users. Worst thing that can happen is that the analytics we want don't actually get logged.

No. I don't think we should verify this at all beyond making sure the code makes sense.

No.


lognaturel:
Not sure why the CircleCI run isn't showing up. https://app.circleci.com/pipelines/github/lognaturel/collect/220/workflows/ed1c55bb-71c2-4dcb-8397-0736606669a5

seadowg (approved):

PR #4600, opened 2021-06-10
Refine project color

### LINKED ISSUE #4493: Refine project color
Mock up for color picker dialog is [here](https://user-images.githubusercontent.com/967540/119382207-93fe2000-bc76-11eb-969e-56a1b35ca0c5.jpg)
Figma with project color palette is [here](https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=454%3A60).

## Acceptance

- [ ] Given I've added a project already
When I add another project with the same server domain (path and protocol can be different)
Then the new project has the same icon color
And the color is one of the project color palette colors

- [x] Given I've added a project already
When I navigate to admin settings
And click on "Project color"
Then I see a dialog with a hex value and a preview of the color
And when I change the hex value
The preview color changes

- [x] Given I've added a project already
When I navigate to admin settings
And click on "Project color"
Then I see a dialog with a color palette (of the project colors)
And when I click on one of the colors
Then the hex value changes (along with the color preview)

## Notes

Color generation based can be very simple. @lognaturel suggested something like `domain.hashCode() % numColorsInPalette` would be enough.

It probably makes sense for the color picker dialog to live in a new module (or `androidshared`) seeing as it's a potentially reusable component and shouldn't need any of collect specific logic in it. 

As much as possible we should just follow material design on the layout, so if something from the mock-up doesn't make sense it's fine to tweak it.

### Assumptions

- A lot of people won’t care what color the project is
- Some people will want to change the color (to work with their emoji icon or to differentiate from other projects that happen to be the same)
- Some people will want to use an exact color to match a brand
- People who share devices will want the color of a project to be the same across devices
- A limited set of color options is helpful for people who don't have a preference (but still want it to look good)

grzesiek2010:
Closes #4493

Screenshot_1623343551

No.

No.


collect_app/src/main/res/layout/color_picker_dialog_layout.xml
```diff
@@ -0,0 +1,359 @@
...
+            app:layout_constraintBottom_toBottomOf="@+id/current_color"
+            app:layout_constraintStart_toStartOf="@+id/color14"
+            app:layout_constraintEnd_toEndOf="parent">
+
+            <com.google.android.material.textfield.TextInputEditText
+                android:id="@+id/hex_color"
+                style="@style/TextAppearance.Collect.Headline5"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:maxLength="6"/>
```
grzesiek2010: I added this because hex colors consist of 6 chars. The thing is that users might ignore that `#` we have before this editText and try to add it. I thought that maybe we could resolve it adding a `TextWatcher` that would automatically remove it. Or we can change this value to 7 (or even remove at all) and try to use both formats (with and without #).

@lognaturel what do you thing? It's probably something for polishing after beta but I wanted to bring it up.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+import androidx.core.widget.doOnTextChanged
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import org.odk.collect.android.R
+import org.odk.collect.android.databinding.ColorPickerDialogLayoutBinding
+import java.lang.Exception
+
+class ColorPickerDialog : DialogFragment() {
```
seadowg: I really think this should live in a separate module (probably `colorpicker`) or at least in Android shared as it doesn't have any dependencies (or shouldn't) on any Collect components and splitting it out helps with our build time and parallelizing our test runs. 

The color values could either live in the other module and be exposed through a helper, or could be passed as an arg to the dialog when opened.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AdminSettingsPage.java
```diff
@@ -66,14 +71,15 @@ public AdminSettingsPage setProjectName(String projectName) {
...
+        onView(withClassName(endsWith("EditText"))).perform(replaceText(""));
+        onView(withClassName(endsWith("EditText"))).perform(typeText(projectIcon));
         clickOKOnDialog();
         return this;
     }
 
     public AdminSettingsPage setProjectColor(String projectColor) {
         clickOnString(R.string.project_color);
-        inputText(projectColor);
+        onView(withId(R.id.hex_color)).perform(replaceText(projectColor));
```
seadowg: Are we not able to use a hint here?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/MainMenuPage.java
```diff
@@ -201,6 +206,8 @@ public DeleteSavedFormPage clickDeleteSavedForm() {
 
     public MainMenuPage assertProjectIcon(String projectIcon, String expectedBackgroundColor) {
         onView(allOf(hasDescendant(withText(projectIcon)), withId(R.id.projects))).check(matches(isDisplayed()));
+        int backgroundColor = ((GradientDrawable) getActivity().findViewById(R.id.project_icon_text).getBackground()).getColor().getDefaultColor();
```
seadowg: It's nice to be able to do this, but we do have the color tested in `ProjectIconTest` so I'd tend to not add these assertions up here. I guess in my head I want to check that the project icon is on screen but the view tests can check the details of what that looks like.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ColorPickerDialog.kt
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    companion object {
+        const val CURRENT_COLOR = "CURRENT_COLOR"
+    }
+}
+
+class ColorPickerViewModel : ViewModel() {
+    val color: MutableLiveData<String> = MutableLiveData()
```
seadowg: I'm liking this set up where the dialog exposes a view model for the other components to use. I'm just wondering about a different interface:

```kotlin
class ColorPickerViewModel : ViewModel() {
    private val _pickedColor = MutableLiveData()
    val pickedColor: LiveData<String> = _pickedColor

    fun pickColor(color: String) {
        _pickedColor.value = color   
    }
}
```

There's two things I'm thinking would be different here:

1. The `MutableLiveData` isn't exposed so only the `pickColor` function changes it.
2. It's just a name change but I'm think have a `pickedColor` that starts `null` and is non-null is simpler for the `ColorPickerDialog`. That way it doesn't need to set the view model value before the color itself is picked. The `AdminPreferencesFragment` can just have a null check in the case where the color hasn't been picked yet.

What do you think?

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -501,8 +502,14 @@ public ProjectsRepository providesProjectsRepository(UUIDGenerator uuidGenerator
...
     public ProjectCreator providesProjectCreator(ProjectImporter projectImporter, ProjectsRepository projectsRepository,
-                                                    CurrentProjectProvider currentProjectProvider, SettingsImporter settingsImporter) {
-        return new ProjectCreator(projectImporter, projectsRepository, currentProjectProvider, settingsImporter);
+                                                    CurrentProjectProvider currentProjectProvider, SettingsImporter settingsImporter, ProjectDetailsCreator projectDetailsCreator) {
+        return new ProjectCreator(projectImporter, projectsRepository, currentProjectProvider, settingsImporter, projectDetailsCreator);
+    }
+
+    @Provides
+    @Singleton
+    public ProjectDetailsCreator providesProjectDetailsCreator(Context context) {
```
seadowg: I don't think we need this to have its own provider as it's only used by the project creator. I think just new-ing it up in the provider for `ProjectCreator` would be fine, and that way the `ProjectDetailsCreator` isn't exposed to every component.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        assertThat(projectDetailsCreator.getProject("https://my-project.com").icon, `is`("M"))
+        assertThat(projectDetailsCreator.getProject("https://your-project.com/one").icon, `is`("Y"))
+        assertThat(projectDetailsCreator.getProject("http://www.my-project.com").icon, `is`("W"))
+    }
+
+    @Test
+    fun `Generated project color should be the same for identical project names`() {
+        assertThat(projectDetailsCreator.getProject("https://my-project.com").color, `is`(projectDetailsCreator.getProject("https://my-project.com").color))
+        assertThat(projectDetailsCreator.getProject("https://your-project.com/one").color, `is`(projectDetailsCreator.getProject("https://your-project.com/one").color))
+        assertThat(projectDetailsCreator.getProject("http://www.my-project.com").color, `is`(projectDetailsCreator.getProject("http://www.my-project.com").color))
```
seadowg, line 34: I really like these tests. Super loose, but get to the core of what we want. It'd be really easy to tweak how we hash the colors, but we'd have confidence that our tests cover the really important part (same color for same domain).

seadowg (changes requested):
This looks really great, and I'm liking the flow even more now I've played with it. One tweak I might make is to remove the "Hex" text? Now I'm playing with it, I'm not sure why it's there. What do you think @lognaturel and @grzesiek2010? I'm also a little worried that it's not clear that the circle at the bottom is the "current" or "picked" color, but I think we should have people play with this and see what their reaction is.

I think there's also a couple of tweaks and a structural change we should make before moving on (inline in comments).

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ColorPickerDialog.kt
grzesiek2010: but then it's not possible to change that value from `ColorPickerDialog` I would need to create a setter method for that right?

seadowg: Sorry the formatting was completely broken on that comment. You could call `pickColor` from `ColorPickerDialog`.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ColorPickerDialog.kt
grzesiek2010: Ok, done.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/AdminSettingsPage.java
grzesiek2010: We don't have hint there but I changed it and now it uses `contentDescription` as I remember you liked it more.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ColorPickerDialog.kt
grzesiek2010: Ok I moved it to `androidshared`. It required moving some resources there too but I think it's ok we probably should move there even more but I didn't want to do that as a part of this pr.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/MainMenuPage.java
grzesiek2010: Ok, I removed it.

grzesiek2010:
> One tweak I might make is to remove the "Hex" text? Now I'm playing with it, I'm not sure why it's there. What do you think

I don't know I left it there for now, @lognaturel?

seadowg (approved):
Really liking the changes. One thing though: I don't think we need to move our styles, typography and themes files over to `androidshared`.

lognaturel:
This looks fabulous!!

I filed an issue at https://github.com/getodk/collect/issues/4606 for another quick pass after the beta. The idea for "Hex" was to provide a label for the text box so users have something to look for if they want to add a custom color but aren't familiar with color representations. I realize the spacing in my sketch really lost that intent and it's not clear it's so helpful after all. I'll try to take that polish pass on post-beta.

lognaturel:
One thing we should probably do before the beta is to prevent crashes when an illegal hex code is provided. Maybe it's just that 'ok' can't be tapped on the color selection dialog for now. The problem is that if a user tries that case, the admin settings for that project can't be accessed anymore.

collect_app/src/main/res/layout/color_picker_dialog_layout.xml
lognaturel: I don't think we need to specifically worry about the `#`. I do think that before beta we should prevent a permanent crash when characters outside the hex range are provided (https://github.com/getodk/collect/pull/4600#issuecomment-861016973). I've put some notes for a longer-term approach at https://github.com/getodk/collect/issues/4606.

PR #4601, opened 2021-06-11
Rework first launch screen

### LINKED ISSUE #4531: Polish first launch dialog
See [mockup](https://www.figma.com/file/SkyAE9BZVmPQYnrPb30peg/First-launch-experience?node-id=104%3A00) discussed in [forum thread](https://forum.getodk.org/t/collect-prompt-for-server-configuration-on-first-launch/33051) and by TAB.

- [x] Update first launch screen following rough guidance to start with:
- Use outline edit icon
- 60x60 [ODK logo](https://getodk.org/assets/zip/odk-brand-assets.zip)
- 20sp button text, medium weight (got feedback that initial mockups were too small/light)
- 35sp tagline text
- 18sp version text
- 14sp "try Collect" text
- My general sense of the middle part is that it's all in a container that hugs contents. The width is determined by the width of the widest button. The two buttons are the same width. The whole container is centered horizontally and vertically.
- I don't know how the button icon and text should be laid out (centered, left aligned). I'm hoping there's a sensible default that can just be used.
- I also don't really know about margins and vertical spacing between elements. I'm hoping there are also some defaults that can be relied on there.

These are all starting points. We should check in about anything that's hard and see if there are easier alternatives. We'll likely want to do another pass of adjusting font sizes once we can look at it on different screens.

- [x] Change first launch screen to `Activity`. The `SplashScreenActivity` will be going away soon. There's some glitchy navigation and backstack manipulation going on right now. (https://github.com/getodk/collect/issues/4531#issuecomment-858689395)

- [x] Leave QR code config dialog when manual config dialog is launched from it instead of manipulating the back stack. (https://github.com/getodk/collect/issues/4531#issuecomment-861619917)

lognaturel:
Closes #4531, see issue for desired changes and expected behavior.

Y'all didn't tell me that working with constraint layouts is pretty reasonable and relaxing!

Some things to pay attention to (and maybe we come back to these after beta):
- Did I get the odk logo png scaling for different resolutions correctly?
- I like the explicitness of "Manually enter project details" but it's pretty long. Should we consider "Configure manually"?
- I increased the vertical spacing between the buttons but maybe I went too far?
- Is the gray for the Collect version too light in the light theme? Is there maybe an existing "color role" we can use for this?

Ran on Moto G5, 1080 x 1920 pixels, 16:9 ratio (~424 ppi density), Samsung Galaxy Tab A 800 x 1280 pixels, 16:10 ratio (~216 ppi density) and various different form factors in the Android Studio design preview. All this is very subjective and I welcome all feedback either now or post-beta.

| Moto G5 	| Galaxy Tab A | 
|---------	|--------------	|
| motog5 | tabA |    

Collect is used on so many different devices, it's hard to know what to design for. I think for now we want to continue with a single layout rather than trying to support different form factors differently.

I used the Figma file and notes from the issues as starting points and stuck to the Material [type scale](https://material.io/design/typography/the-type-system.html#type-scale) (which I have some issues with. material.io doesn't even follow it). I'm fairly happy with where I've landed. It all feels a little bit small on a tablet but I think it's acceptable. If all the whitespace feels too bothersome, one idea is to cap the height of the outermost container.

Currently the tagline uses H4 (34sp) and the button text uses Body1 (16sp). The other thing I considered is a tagline of H3 (48sp) and buttons of H6 (20sp). It looks good on a tablet but cramped on a more typical device and would result in wrapping on some of the smallest devices:

| Moto G5 | 480x800 | 
|---------	|--------------	|
| g5-h3tagline-h6buttons | 480x800-h3tagline-h6buttons|    

I looked for reasons why the first launch screen might have been a dialog but I don't see any discussion at https://github.com/getodk/collect/pull/4517.

These are mostly layout changes so risk is low.

No.


collect_app/src/main/res/layout/automatic_project_creator_dialog_layout.xml
```diff
@@ -75,7 +75,8 @@
                 android:layout_height="wrap_content"
                 android:layout_marginTop="@dimen/margin_standard"
                 android:text="@string/configure_manually"
-                app:icon="@drawable/ic_edit"
+                android:textAppearance="?attr/textAppearanceBody1"
```
lognaturel, line 78: Is this ok or should it be wrapped in something Collect-specific like the corresponding style?

collect_app/src/main/res/values/colors.xml
```diff
@@ -20,6 +20,9 @@ the License.
     <color name="green_500">#4CAF50</color>
     <color name="deep_purple_500">#673AB7</color>
 
+    <color name="website_blue">#3e9fcc</color>
```
lognaturel, line 23: These two new colors look fine in both light and dark modes. I don't really know what they represent yet so giving them names based on their colors and putting them here seemed ok.

collect_app/src/main/res/values/dimens.xml
```diff
@@ -1,38 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- These can likely all be replaced by Material resources -->
```
lognaturel, line 2: I got pretty confused about what these were and how they were being used so I figured it'd be a good opportunity to clean up the unused ones. These were introduced at some point on the way to greater standardization but now we can use material resources.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FirstLaunchActivityTest.kt
```diff
@@ -0,0 +1,111 @@
...
+                    projectsRepository: ProjectsRepository,
+                    storagePathProvider: StoragePathProvider,
+                    context: Context,
+                    settingsProvider: SettingsProvider
+            ): ProjectImporter? {
+                return projectImporter
+            }
+        })
+
+        val scenario = ActivityScenario.launch(FirstLaunchActivity::class.java)
```
lognaturel, line 62: I'd prefer to use a rule rather than explicitly building the scenario every time but I can't figure out how to do the dependency override in that case. Ok to leave like this for now?

collect_app/src/test/java/org/odk/collect/android/activities/FirstLaunchActivityTest.kt
```diff
@@ -1,57 +1,54 @@
...
+class FirstLaunchActivityTest {
     @Test
-    fun `The dialog should not be dismissed after clicking on a device back button`() {
-        val scenario = RobolectricHelpers.launchDialogFragment(FirstLaunchDialog::class.java, R.style.Theme_Collect_Light)
-        scenario.onFragment {
-            assertThat(it.isVisible, `is`(true))
-            onView(isRoot()).perform(pressBack())
-            assertThat(it.isVisible, `is`(true))
+    fun `The main menu should be started after clicking on the 'Try a demo' link`() {
+        val scenario = ActivityScenario.launch(FirstLaunchActivity::class.java)
```
lognaturel, line 39: I'd prefer to use a rule rather than explicitly building the scenario every time but I can't figure out how to do the dependency override in that case. Ok to leave like this for now?

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
```diff
@@ -67,56 +64,10 @@ class SplashScreenActivityTest {
         })
     }
 
-    @Test
-    fun `SplashScreenActivity should implement ProjectAddedListener`() {
```
lognaturel, line 78: This test and the following further convinced me that the first launch screen should be its own activity.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
```diff
@@ -67,56 +64,10 @@ class SplashScreenActivityTest {
...
-        }
-    }
-
     @Test
     fun `The Splash screen should be displayed with the default logo if it's enabled and no other logo is set`() {
-        doReturn(false).`when`(splashScreenViewModel).shouldFirstLaunchScreenBeDisplayed
-        doReturn(true).`when`(splashScreenViewModel).shouldDisplaySplashScreen
-        doReturn(false).`when`(splashScreenViewModel).doesLogoFileExist
+        `when`(splashScreenViewModel.shouldDisplaySplashScreen).thenReturn(true)
+        `when`(splashScreenViewModel.doesLogoFileExist).thenReturn(false)
```
lognaturel: I changed all these to use when/thenReturn which feels more readable to me. `doReturn` is typically only recommended if necessary: https://www.oreilly.com/library/view/test-driven-development-with/9781783283293/ch06s10.html

I also removed the configuration that was not relevant to the tests. For example, it makes no difference to what this test is verifying whether the first launch screen should be displayed or not.

collect_app/src/main/java/org/odk/collect/android/projects/AutomaticProjectCreatorDialog.kt
```diff
@@ -143,15 +139,11 @@ class AutomaticProjectCreatorDialog : MaterialFullScreenDialogFragment() {
         val projectCreatedSuccessfully = projectCreator.createNewProject(CompressionUtils.decompress(result.text))
 
         if (projectCreatedSuccessfully) {
+            ActivityUtils.startActivityAndCloseAllOthers(activity, MainMenuActivity::class.java)
             ToastUtils.showLongToast(getString(R.string.new_project_created))
-            listener?.onProjectAdded()
```
lognaturel, line 147: It's hard to tell but the logic here is actually the same. The call on `onProjectAdded` made the `SplashScreenActivity` launch the `MainMenuActivity` and close all others.

grzesiek2010:
Generally it looks nice. I can spend more time once you make it ready for review but I think nothing big will need to be changed.

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
```diff
@@ -129,7 +118,12 @@ class AutomaticProjectCreatorDialog : MaterialFullScreenDialogFragment() {
         barcodeViewDecoder.waitForBarcode(binding.barcodeView).observe(
             viewLifecycleOwner,
             { barcodeResult: BarcodeResult ->
-                beepManager.playBeepSoundAndVibrate()
+                try {
```
lognaturel, line 121: In general, ignoring any exception here seems like a fine approach because the beep and vibration is not mission-critical and we certainly don't want to prevent project creation because it fails in some way.

What drove adding this now is that `beepManager` is a dependency managed by the dialog. Robolectric provides some kind of shadow media manager but it's missing some kind of configuration and crashes.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FirstLaunchActivityPage.kt
```diff
@@ -2,12 +2,13 @@ package org.odk.collect.android.support.pages
 
 import org.odk.collect.android.R
 
-internal class FirstLaunchDialogPage : Page<FirstLaunchDialogPage>() {
+internal class FirstLaunchActivityPage : Page<FirstLaunchActivityPage>() {
```
grzesiek2010: Now you have `FirstLaunchActivityPage` and `FirstLaunchPage` it should be just one class right?

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
grzesiek2010, line 121: Do we really need this? In `BarCodeScannerFragment` it's called without try/catch and it doesn't seem to cause problems.

collect_app/src/main/res/drawable/ic_outline_edit_24.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
grzesiek2010, line 1: Do we need this new icon? The one used before `ic_edit` looked almost the same.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
grzesiek2010: Can we use org.mockito.kotlin.whenever to avoid those backticks?

grzesiek2010 (changes requested):

collect_app/src/main/res/drawable/ic_outline_edit_24.xml
lognaturel: Yes, using the outline icon was an explicit goal in the issue.

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
lognaturel: Whoops, https://github.com/getodk/collect/pull/4601#discussion_r652156792 was stuck in a review. Yes, it’s required for the test.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FirstLaunchActivityPage.kt
lognaturel: `FirstLaunchActivityPage` was used in just one place so I removed it.

collect_app/src/test/java/org/odk/collect/android/activities/SplashScreenActivityTest.kt
lognaturel: Sure. 👍

grzesiek2010 (approved):

PR #4602, opened 2021-06-11
Fix upgrading to projects with existing admin settings

seadowg:
This fixes a problem that QA found where admin settings from an existing installation were not retained when updating the app.

This ended up being a bit of a rabbit hole as I ran into a few related problems on my way (I've left comments inline about that). I'd initially intended to work on an implementation of migrating our background jobs on upgrade, but realized we needed a clearer sense of if the app was being upgraded or not, so decided to rework the code that performs the "existing project" upgrade to drive out a new upgrade hook and fix the bug in the process (so this isn't just purely a refactor). That lets us review the new structure without any "new" behaviour.

Comments inline.

I think we can continue to skip QA on this, but we need to make sure we pay special attention to upgrades later.


collect_app/src/androidTest/java/org/odk/collect/android/support/CopyFormRule.java
```diff
@@ -71,6 +73,7 @@ public void evaluate() throws Throwable {
             } catch (IllegalStateException e) {
                 component.projectImporter().importDemoProject();
                 component.currentProjectProvider().setCurrentProject(ProjectImporter.DEMO_PROJECT_ID);
+                projectCreated = true;
```
seadowg, line 76: Ideally we'd just get rid of `CopyFormRule` but that would be a lot of energy right now.

collect_app/src/main/java/org/odk/collect/android/application/initialization/AppUpgrader.kt
```diff
@@ -0,0 +1,26 @@
+package org.odk.collect.android.application.initialization
+
+import org.odk.collect.shared.Settings
+
+class AppUpgrader(private val metaSettings: Settings, private val upgrades: List<Upgrade>) {
```
seadowg: I've potentially gone overboard on building out a small framework for calling upgrade code, but I found it really hard to slot the logic I needed in. My goal here was to separate the decision-making around whether we are upgrading or not from the different things that need to happen when we are.

collect_app/src/main/java/org/odk/collect/android/utilities/AppStateProvider.kt
```diff
@@ -1,25 +0,0 @@
...
-
-    fun isFreshInstall(): Boolean {
-        return !isUpdatedVersion() && !metaSettings.contains(MetaKeys.FIRST_LAUNCH)
-    }
-
-    private fun isUpdatedVersion(): Boolean {
-        return if (alwaysFresh) {
-            false
-        } else {
-            packageInfo.firstInstallTime != packageInfo.lastUpdateTime
```
seadowg, line 17: This logic wouldn't have worked in cases where we upgrade and then clear state - we'd really want to treat that as a "fresh install". Again I could probably have held back on addressing this case here, but I found it very hard to keep my mind straight on everything else knowing about these little edge cases.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -19,70 +19,57 @@
...
-    private void initializeStorage() {
-        storageInitializer.createOdkDirsOnStorage();
+    private void performUpgradeIfNeeded() {
+        if (launchState.isUpgradedFirstLaunch()) {
+            appUpgrader.upgrade();
+        }
     }
 
-    private void initializePreferences() {
-        performMigrations();
```
seadowg, line 84: This needs to move into the new upgrade hook. I've created an issue for this: https://github.com/getodk/collect/issues/4604. Where this call was, it wasn't actually achieving anything as the "existing" project hadn't been imported yet.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -19,70 +19,57 @@
...
-        reloadSharedPreferences();
```
seadowg, line 85: There is no test to drive this out and I think it might just be a relic at this point? Our `SharedPreferencesSettings` returns defaults if values are missing, so there is no reason to write them to the prefs themselves (as far as I understand).

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -95,11 +82,10 @@ private void initializeFrameworks() {
     }
 
     private void initializeAnalytics() {
-        boolean isAnalyticsEnabled = generalSettings.getBoolean(GeneralKeys.KEY_ANALYTICS);
-        analytics.setAnalyticsCollectionEnabled(isAnalyticsEnabled);
+        analytics.setAnalyticsCollectionEnabled(false);
 
-        FormUpdateMode formUpdateMode = getFormUpdateMode(context, generalSettings);
-        analytics.setUserProperty("FormUpdateMode", formUpdateMode.getValue(context));
+//        FormUpdateMode formUpdateMode = getFormUpdateMode(context, generalSettings);
```
seadowg, line 88: I've created an issue for this as well: https://github.com/getodk/collect/issues/4605

async/src/main/java/org/odk/collect/async/Scheduler.kt
```diff
@@ -72,4 +72,6 @@ interface Scheduler {
      * @return object that allows task to be cancelled
      */
     fun repeat(foreground: Runnable, repeatPeriod: Long): Cancellable
+
+    fun cancelAllDeferred()
```
grzesiek2010, line 76: If it's unused why did you decide to add it as a part of this pr? Will you need this later or what?

collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.odk.collect.android.preferences.keys.MetaKeys
+import org.odk.collect.android.preferences.source.SettingsProvider
+import org.odk.collect.android.projects.CurrentProjectProvider
+import org.odk.collect.android.storage.StoragePathProvider
+import org.odk.collect.projects.Project
+import org.odk.collect.projects.ProjectsRepository
+import java.io.File
+import java.io.FileNotFoundException
+
+class ExistingProjectMigrator(
```
grzesiek2010, line 22: I think adding a comment would be valuable here. You know what's going on I know too but for other people it might be not clear (maybe even for us in a few months). So I would clarify that it imports the project that existed before migrating to the version with multiple projects.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
grzesiek2010: Yes it shouldn't be needed as we return our defaults if a preference doesn't exist. I wonder if we need that `setDefaultForAllSettingsWithoutValues()` at all... this is something we need to investigate. New issue?

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -16,73 +16,61 @@
...
-        this.projectImporter = projectImporter;
-        this.currentProjectProvider = currentProjectProvider;
+        this.launchState = launchState;
+        this.appUpgrader = appUpgrader;
     }
 
     public void initialize() {
         initializeStorage();
-        initializePreferences();
+        performUpgradeIfNeeded();
```
grzesiek2010, line 59: Is there any reason you do this earlier than the old `importExistingProjectIfNeeded()`? I mean before `initializeFrameworks()` and `initializeLocale()`

collect_app/src/main/java/org/odk/collect/android/utilities/LaunchState.kt
```diff
@@ -0,0 +1,34 @@
...
+    private val context: Context,
+    private val metaSettings: Settings,
+    private val currentVersion: Int
+) {
+
+    fun isUpgradedFirstLaunch(): Boolean {
+        return if (metaSettings.contains(MetaKeys.LAST_LAUNCHED)) {
+            metaSettings.getInt(MetaKeys.LAST_LAUNCHED) < currentVersion
+        } else {
+            val legacyMetadataDir = File(context.getExternalFilesDir(null), "metadata")
```
grzesiek2010: So this checks if it's upgraded from version without projects and above all the future cases? If so it would be good to comment it somehow as well or create two separate methods with descriptive names.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -581,4 +563,16 @@ public InstanceAutoSender providesInstanceAutoSender(Context context, ChangeLock
...
     }
+
+    @Provides
+    public ExistingProjectMigrator providesExistingProjectMigrator(Context context, StoragePathProvider storagePathProvider, ProjectsRepository projectsRepository, SettingsProvider settingsProvider, CurrentProjectProvider currentProjectProvider) {
+        return new ExistingProjectMigrator(context, storagePathProvider, projectsRepository, settingsProvider, currentProjectProvider);
+    }
+
+    @Provides
+    public AppUpgrader providesAppUpgrader(SettingsProvider settingsProvider, ExistingProjectMigrator existingProjectMigrator) {
+        return new AppUpgrader(settingsProvider.getMetaSettings(), asList(
```
grzesiek2010: You can use `singletonList` instead of `asList`

collect_app/src/main/java/org/odk/collect/android/application/initialization/upgrade/AppUpgrader.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import org.odk.collect.shared.Settings
+
+class AppUpgrader(private val metaSettings: Settings, private val upgrades: List<Upgrade>) {
+
+    fun upgrade() {
+        upgrades.forEach {
+            val key = it.key()
+
+            if (key == null) {
```
grzesiek2010, line 11: Is this case possible?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: I could just delete it as part of this PR I guess?

async/src/main/java/org/odk/collect/async/Scheduler.kt
seadowg: Yeah it's meant to be in the next piece of work. I thought moved the commit but accidentally left it here. I'll reset back one commit and push again.

seadowg: Oops no it is the first commit. Let's just leave this in, as I will be using it next.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
seadowg: Great point. Will do.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: We need it earlier than anything that references `Settings` but I guess for the moment it doesn't need to be that early. Any reason to move it back?

collect_app/src/main/java/org/odk/collect/android/utilities/LaunchState.kt
seadowg: I put a comment on the tests but I think you're right that a descriptive method would be nice here.

collect_app/src/main/java/org/odk/collect/android/application/initialization/upgrade/AppUpgrader.kt
seadowg: My thinking is we'll want to use the `AppUpgrader` for settings and work manager migrations, which should run on every upgrade (unlike the existing project which only runs once). For some reason, to get the logic covering those different kinds of cases out of my head, so I was sure the structure we had in place would actually work moving forward. Does that make sense?

async/src/main/java/org/odk/collect/async/Scheduler.kt
grzesiek2010: Ok.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.odk.collect.android.projects.CurrentProjectProvider
+import org.odk.collect.android.storage.StoragePathProvider
+import org.odk.collect.projects.Project
+import org.odk.collect.projects.ProjectsRepository
+import java.io.File
+import java.io.FileNotFoundException
+
+/**
+ * This migrates from a version before Projects were introduced. Any data and settings will be
+ * used used to create a new project which is then set as the current one This means the user never
```
grzesiek2010: Double `used`.

collect_app/src/main/java/org/odk/collect/android/utilities/LaunchState.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    fun isUpgradedFirstLaunch(): Boolean {
+        return if (metaSettings.contains(MetaKeys.LAST_LAUNCHED)) {
+            metaSettings.getInt(MetaKeys.LAST_LAUNCHED) < currentVersion
+        } else {
+            return legacyInstallDetected()
+        }
+    }
+
+    // Check if it looks like  is an install from before when `LAST_LAUNCHED` was introduced
```
grzesiek2010: Double space after like.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
grzesiek2010: No i was just curious. It makes sense to do that earlier.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
grzesiek2010: Actually there are cases where loading defaults is important... ah that's confusing!
For example take a look at `SettingsImporter` where we load defaults after importing new settings (we don't use that `setDefaultForAllSettingsWithoutValues()` method, maybe we should to remove duplicated code but generally it does the same)
We need loading defaults because next step it to set up `SettingsChangeHandler` for every setting so that it can perform operations after changing some settings not only for those preferences that were included in json but for all settings (those just cleared as well - taht's why we need to save defaults to loop through them as well).
I'm not sure if in our code base we have settings that really need to react to clearing them but at least theoretically it should work like that.

seadowg: Hmmm we can come back to this later then! I think hopefully we can get away with just using the default behaviour in `SharedPreferencesSettings` eventually.

grzesiek2010: So maybe we should have an issue for investigating it - when we need loading defaults?

grzesiek2010 (approved):

PR #4607, opened 2021-06-14
Add import/export of project details via QR

### LINKED ISSUE #4582: Add import/export of project details via QR
## Acceptance

- [x] Given I'm viewing a project's QR code in settings on my first device
And have Collect open on a second device
When I click "Add project" on the second device
And scan the QR code displayed on the first device
Then a new project is added to the second device
And the project details (name, icon etc) are the same as the project on the first device
And the settings for the project are the same as the project on the first device

- [x] Given I'm viewing a project's QR code in settings on my first device
And have Collect open at the first launch screen on a second device
When I click "Configure with QR code" on the second device
And scan the QR code displayed on the first device
Then a project is set up to the second device
And the project details (name, icon etc) are the same as the project on the first device
And the settings for the project are the same as the project on the first device

- [x] Given I'm viewing a project's QR code in settings on my first device
And have Collect open at Admin Settings
When I click "Configure with QR code" on the second device
And scan the QR code displayed on the first device
Then the project is updated on the second device
And the project details (name, icon etc) are the same as the project on the first device
And the settings for the project are the same as the project on the first device

## Notes

The project details could live in either `admin` or in a new `project` object in the config JSON depending on whatever is easiest for implementation.

grzesiek2010:
Closes #4582

No.

No.


collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
```diff
@@ -142,11 +153,11 @@ class JsonPreferencesGeneratorTest {
         settingsProvider.getAdminSettings("2").saveAll(adminPrefsForProjectX)
 
         // Verify the demo project
-        verifyJsonContent(jsonPreferencesGenerator.getJSONFromPreferences(), generalPrefsForDemoProject, adminPrefsForDemoProject)
+        verifyJsonContent(jsonPreferencesGenerator.getJSONFromPreferences(), generalPrefsForDemoProject, adminPrefsForDemoProject, projectDetails)
```
seadowg: This all really confused me as the test is using a real `SettingsProvider` which means it uses the real `CurrentProjectProvider` but then the actual object under test uses a mock `CurrentProjectProvider`. I think we need to clean that up as the boundaries of this test aren't quite defined.

One idea is that we create a test settings provider implementation (and pull out an interface to sit above the real and test one) that just uses `InMemSettings` under the hood and doesn't use the `CurrentProjectProvider` (it just returns different settings objects for different project Ids and a default one for no project Id). This would be really useful in a lot of places as we could use it instead of the real `SettingsProvider` and potentially avoid Robolectric as well (as we need it for `SharedPreferences`).

collect_app/src/test/java/org/odk/collect/android/configure/SettingsImporterTest.java
```diff
@@ -45,6 +50,10 @@
 
     @Before
     public void setup() {
+        getComponent(ApplicationProvider.<Collect>getApplicationContext()).projectImporter().importDemoProject();
```
seadowg, line 58: Instead of setting up a current project, each test could just pass in a random project ID to `getGeneralSettings` etc and `fromJSON`. There's no need to actually have the project setup, just the same id used for the import and assertions. This is another place where a test version of `SettingsProvider` would be useful.

seadowg: Instead of setting up a current project, each test could just pass in a random project ID to `getGeneralSettings` etc and `fromJSON`. There's no need to actually have the project setup, just the same id used for the import and assertions. This is another place where a test version of `SettingsProvider` would be useful.

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
seadowg: This all really confused me as the test is using a real `SettingsProvider` which means it uses the real `CurrentProjectProvider` but then the actual object under test uses a mock `CurrentProjectProvider`. I think we need to clean that up as the boundaries of this test aren't quite defined.

One idea is that we create a test settings provider implementation (and pull out an interface to sit above the real and test one) that just uses `InMemSettings` under the hood and doesn't use the `CurrentProjectProvider` (it just returns different settings objects for different project Ids and a default one for no project Id). This would be really useful in a lot of places as we could use it instead of the real `SettingsProvider` and potentially avoid Robolectric as well (as we need it for `SharedPreferences`).

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
```diff
@@ -75,4 +84,20 @@ class SettingsImporter(
             }
         }
     }
+
+    private fun importProjectDetails(projectJson: JSONObject, projectId: String) {
+        val project = projectsRepository.get(projectId)!!
```
seadowg: Could we pass a `Project.Saved` into `fromJSON` to avoid the `!!` here? We should always have a project when doing an import.

collect_app/src/test/java/org/odk/collect/android/projects/ManualProjectCreatorDialogTest.kt
```diff
@@ -112,7 +112,7 @@ class ManualProjectCreatorDialogTest {
             onView(withHint(R.string.password)).perform(replaceText("1234"))
 
             onView(withText(R.string.add)).perform(click())
-            verify(projectCreator).createNewProject("{\"general\":{\"server_url\":\"https:\\/\\/my-server.com\",\"username\":\"adam\",\"password\":\"1234\"},\"admin\":{}}")
+            verify(projectCreator).createNewProject("{\"general\":{\"server_url\":\"https:\\/\\/my-server.com\",\"username\":\"adam\",\"password\":\"1234\"},\"admin\":{},\"project\":{}}")
```
seadowg, line 115: Why does there need to be a `project` object in the JSON here?

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/projects/ManualProjectCreatorDialogTest.kt
grzesiek2010: Because now it's a part of the json generated by our app. And like `admin` section which also in this case is empty it exists there.

seadowg: Right I was just checking it doesn't **need** to be there (that we still support QR codes with `project`) but I've double-checked and it looks like we're ok!

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
grzesiek2010: Yes we can do that. Done!

collect_app/src/test/java/org/odk/collect/android/preferences/JsonPreferencesGeneratorTest.kt
grzesiek2010: > which means it uses the real CurrentProjectProvider but then the actual object under test uses a mock CurrentProjectProvider

why? I think there is no real `CurrentProjectProvider` just a mock.

I thought this test had a broader responsibility (testing generating jsons with real settings). It's not a test for one class (SettingsImporter for example) which then (I would agree) should rather have mocked settings.

seadowg: > why? I think there is no real CurrentProjectProvider just a mock.

The test interacts with a real `CurrentProjectProvider` through Dagger:

```
// Verify the 'Project X' project
        DaggerUtils.getComponent(ApplicationProvider.getApplicationContext<Collect>()).currentProjectProvider().setCurrentProject("2")
```

The `SettingsProvider` is also interacting with this real `CurrentProjectProvider` as well, whereas the `AppConfigurationGenerator` is interacting with a mock one. I think it's confusing that we need to provide setup for the `SettingsProvider` here which I think we trust (and have tested) to return the right settings objects. To me, it feels like a fake/stub `SettingsProvider` could be used here instead and that would make it clearer what the `AppConfigurationGenerator` is responsible for. The stub could return `SharedPreferencesSettings` if that feels like a more thorough test, but I'm not sure we'd need that. Does that make sense?

grzesiek2010: Hmm yes I think I confused this test (which is about exporting prefs to json) with the one which decodes json and my thinking was wrong.
Ok I think mocking `SettingsProvider` makes sense:
https://github.com/getodk/collect/pull/4607/commits/d5dbd8160b995400c799d0f16a669fecf9d127b5
Do you like this version?

seadowg: Ah right! Yeah this looks great.

collect_app/src/test/java/org/odk/collect/android/preferences/AppConfigurationGeneratorTest.kt
```diff
@@ -1,32 +1,51 @@
...
+        AppConfigurationKeys.PROJECT_NAME to "Project X",
+        AppConfigurationKeys.PROJECT_ICON to "X",
+        AppConfigurationKeys.PROJECT_COLOR to "#cccccc"
+    )
+
+    private lateinit var appConfigurationGenerator: AppConfigurationGenerator
 
     @Before
     fun setup() {
         settingsProvider.getGeneralSettings().clear()
```
seadowg: I don't think you'll need these `clear()` calls here.

seadowg (changes requested):
Looks great. Just one little tweak before we merge.

collect_app/src/test/java/org/odk/collect/android/configure/SettingsImporterTest.java
grzesiek2010: I solved this in a similar way like `AppConfigurationGeneratorTest`

collect_app/src/test/java/org/odk/collect/android/preferences/AppConfigurationGeneratorTest.kt
grzesiek2010: Fixed.

seadowg (approved):

PR #4609, opened 2021-06-15
Make sure form update jobs are rescheduled on upgrade

### LINKED ISSUE #4430: Background jobs for all projects continue to run

seadowg:
Closes #4430
~~Blocked by #4602~~

This makes sure that, when the app is upgraded, existing background jobs are cancelled and then form updates are rescheduled.

New tests and verified manually for match exactly (configured for 1.30.1 and then updated to current dev version).

Not a lot to chat about here, as the structure of how upgrade work was defined and discussed in an earlier PR. I've left a couple of comments inline.

We'll skip QA on this for the moment and test background jobs as a whole later.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
@@ -26,8 +26,6 @@ import org.odk.collect.utilities.UserAgentProvider;
 
 import java.util.Locale;
 
-import javax.inject.Inject;
-
 import timber.log.Timber;
 
 public class ApplicationInitializer {
@@ -40,7 +38,6 @@ public class ApplicationInitializer {
     private final LaunchState launchState;
     private final AppUpgrader appUpgrader;
 
-    @Inject
     public ApplicationInitializer(Application context, UserAgentProvider userAgentProvider,
                                   PropertyManager propertyManager, Analytics analytics,
                                   StorageInitializer storageInitializer, LaunchState launchState,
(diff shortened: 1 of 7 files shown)


collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -40,7 +38,6 @@
     private final LaunchState launchState;
     private final AppUpgrader appUpgrader;
 
-    @Inject
```
seadowg, line 43: I got rid of this as we generally don't use constructor `@Inject`s, and I only had it as a convenience for myself while I messed around with the signature.

grzesiek2010 (approved):

Issue #4572, opened 2021-05-21
Switch to MaterialComponents theme


seadowg:
We should replace our `Light` and `Dark` themes with one Material `DayNight` theme. The groundwork for this has been laid with our `Magenta` theme already but next steps to get this work into production is:

- [x] Pick new primary and secondary colors for the app (dark and light)
- [x] Use `Magenta` as the theme for the app
- [x] Switch `Magenta` to being `DayNight` using the new colors
- [x] Make main menu buttons grey again
- [x] Remove poyfills in `material_theme_polyfills.xml`
- [x] Add shape theming to `Magenta` based on the current dialog shape in `Light`/`Dark`
- [x] Make light/dark preference work again (using [`AppCompatDelegate.setDefaultNightMode()`](https://developer.android.com/reference/androidx/appcompat/app/AppCompatDelegate#setDefaultNightMode(int)) to switch between themes)
- [x] Get rid of `colors.xml` (as much as possible)
- [x] Work out form state icons colors
- [x] Fix dark colors
- [x] Check date/time picker dialogs look right
- [x] Get rid of unneeded theme/style files
- [x] Work out map/geo colors
- [x] Update dev docs to reference Material Theming documentation (MaterialAlertDialogBuilder)

seadowg:
Primary and secondary colors are going to just be: #3e9fcc and #bd006b (the "website" blue we are using and magenta from Central). This is demo'ed in the Material Color Tool (and text/dark/light colors can be exported from) [here](https://material.io/resources/color/#!/?view.left=0&view.right=0&primary.color=3e9fcc&secondary.color=bd006b&primary.text.color=000000&secondary.text.color=ffffff). We'll be using a surface color app bar rather than a primary color one.

seadowg:
The date/time picker dialogs have some messed up colors because of course they do 🤯

seadowg:
I've created #4806 as I think we're going to need to do that long term to maintain any sanity (these things are pretty much unmaintained as far as I can tell).

PR #4610, opened 2021-06-16
Show project name in app bar

### LINKED ISSUE #4584: Use project name as app title, remove app icon, move Collect version to bottom

lognaturel:
Closes #4584

Manually tried it.

I could be convinced to make the app name and version black. I used a dark gray to match the first launch screen version introduced at https://github.com/getodk/collect/pull/4601. Otherwise, I matched the spirit of the mockups and the items we indicated in the issue with as minimal of changes as possible for now.

I don't think there are any code-related risks. There is a risk that the behavior change will be confusing or disorienting to users. This is why we want it to be in the beta so we can get feedback as soon as possible.

No.

https://github.com/getodk/docs/issues/1365

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
@@ -92,6 +92,7 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
         currentProjectViewModel = new ViewModelProvider(this, currentProjectViewModelFactory).get(CurrentProjectViewModel.class);
         currentProjectViewModel.getCurrentProject().observe(this, project -> {
             invalidateOptionsMenu();
+            setTitle(project.getName());
         });
 
         initToolbar();
@@ -265,7 +266,6 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
 
     private void initToolbar() {
         Toolbar toolbar = findViewById(R.id.toolbar);
-        setTitle(String.format("%s %s", getString(R.string.app_name), mainMenuViewModel.getVersion()));
         setSupportActionBar(toolbar);
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
@@ -182,6 +182,9 @@ public class MainMenuActivity extends CollectAbstractActivity implements AdminPa
             }
         });
 
+        TextView appName = findViewById(R.id.app_name);
+        appName.setText(String.format("%s %s", getString(R.string.app_name), mainMenuViewModel.getVersion()));
+
         TextView versionSHAView = findViewById(R.id.version_sha);
         String versionSHA = mainMenuViewModel.getVersionCommitDescription();
         if (versionSHA != null) {
--- a/collect_app/src/main/res/layout/main_menu.xml
+++ b/collect_app/src/main/res/layout/main_menu.xml
@@ -91,17 +91,17 @@
 
                 <TextView
                     android:id="@+id/app_name"
-                    style="@style/TextAppearance.Collect.Body1"
+                    android:textAppearance="?textAppearanceBody1"
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
                     android:layout_gravity="center_horizontal"
                     android:layout_marginTop="@dimen/margin_standard"
-                    android:textColor="@color/dark_gray"
+                    android:textColor="@color/color_on_surface_medium_emphasis"
                     tools:text="ODK Collect v2022.3" />
 
                 <TextView
                     android:id="@+id/version_sha"
-                    style="@style/TextAppearance.Collect.Caption"
+                    android:textAppearance="?textAppearanceCaption"
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
                     android:layout_gravity="center_horizontal"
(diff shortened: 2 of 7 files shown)


collect_app/src/main/res/layout/main_menu.xml
```diff
@@ -89,13 +89,22 @@
                     android:layout_marginTop="@dimen/margin_extra_small"
                     tools:text="@string/manage_files" />
 
+                <TextView
+                    android:id="@+id/app_name"
+                    style="@style/TextAppearance.Collect.Body1"
```
seadowg: This should be `?textAppearanceBody1` so we're referencing the theme attribute rather than our style explicitly. Changing the version sha to do the same thing would be good as well.

collect_app/src/main/res/layout/main_menu.xml
```diff
@@ -89,13 +89,22 @@
...
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_gravity="center_horizontal"
+                    android:layout_marginTop="@dimen/margin_standard"
+                    android:textColor="@color/dark_gray"
```
seadowg: No problem with this being gray but it needs to use theme colors rather than an explicit colors so it changes with theme. I think this should probably be `@color/color_on_surface_medium_emphasis` (which is just applies an alpha to `?colorOnSurface`). There is also low and high emphasis versions if that doesn't look right.

collect_app/src/main/res/values/colors.xml
```diff
@@ -22,6 +22,8 @@ the License.
 
     <color name="highContrastHighlight">#ff8c00</color>
 
+    <color name="dark_gray">#999999</color>
```
seadowg: We should just delete this color once we're no longer using for the text.

collect_app/src/main/java/org/odk/collect/android/activities/MainMenuActivity.java
```diff
@@ -92,6 +92,7 @@ protected void onCreate(Bundle savedInstanceState) {
         currentProjectViewModel = new ViewModelProvider(this, currentProjectViewModelFactory).get(CurrentProjectViewModel.class);
         currentProjectViewModel.getCurrentProject().observe(this, project -> {
             invalidateOptionsMenu();
+            setTitle(project.getName());
```
seadowg, line 95: We need a test to drive this out. Currently, the main menu is asserted on using the app name (which is still on screen) so I think it either makes sense to add something to the Espresso tests in `projects` (probably to an existing test rather than a whole new one) or to add a test at the activity level.

seadowg (changes requested):

collect_app/src/main/res/layout/main_menu.xml
lognaturel: Great, thanks. I had that question at #4601 because I was not finding the "color role" I wanted.

collect_app/src/main/res/values/typography.xml
```diff
@@ -32,7 +32,4 @@
         <item name="android:textAllCaps">false</item>
         <item name="android:letterSpacing">0.04</item>
     </style>
-
-    <style name="TextAppearance.Collect.Caption" parent="TextAppearance.MaterialComponents.Caption">
```
lognaturel, line 36: We should only be introducing new styles where we need to override the Material defaults, right? Like I said in https://github.com/getodk/collect/pull/4611#discussion_r652898601, there's quite a bit of chaos around where styles are introduced, where styles are used instead of theme attributes, etc. Might be worth a short narrative somewhere? I don't see anything currently but maybe it already exists in Android docs or something.

collect_app/src/main/res/layout/main_menu.xml
seadowg: Ah right I hadn't had a chance to look at the PR. It's very likely that new blue you've introduced should be our "primary" color. I've added a note to try and kill `colors.xml` as part of https://github.com/getodk/collect/issues/4572.

collect_app/src/main/res/values/typography.xml
seadowg: Yeah you're right, there is no reason for us to have a text appearance style with no customization. For typography, it is actually pretty well documented at https://material.io/develop/android/theming/typography. There is also docs on color and shape theming there. I've done a bad job of enforcing this in reviews because my brain is reading the attribute and the explicit reference as the same thing. It only really becomes a huge deal when you move the component to another module where it won't have access to that style any longer.

I think we can maybe add to the `STATE.md` that we're using Material Theming but ultimately as we're using Material Components and will eventually switch to a Material Theme it's just the "correct" way of doing things within those systems, so it comes down to following the framework docs more than anything else. I've added a note to https://github.com/getodk/collect/issues/4572 try and get a reference in to our dev docs.

seadowg (approved):

PR #4611, opened 2021-06-16
Group settings buttons with current project and add margins between project list projects

lognaturel:
We've gotten feedback that it's not clear what "general settings" and "admin settings" connects to. This PR attempts to address that.

Manual verification.

The alignment of the settings button isn't perfect because it's missing margin. The goal for now is just to see whether the general approach is acceptable and we can do a final cleanup pass after beta.

These are just layout changes so risk is very low. 

No.

No.

### DIFF
--- a/collect_app/src/main/res/drawable/project_icon_background.xml
+++ b/collect_app/src/main/res/drawable/project_icon_background.xml
@@ -2,5 +2,5 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="oval">
     <solid android:color="@android:color/black"/>
-    <size android:width="@dimen/project_list_icon_size" android:height="@dimen/project_list_icon_size"/>
+    <size android:width="30dp" android:height="30dp"/>
 </shape>
--- a/collect_app/src/main/res/drawable/project_icon_background.xml
+++ b/collect_app/src/main/res/drawable/project_icon_background.xml
@@ -2,5 +2,5 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="oval">
     <solid android:color="@android:color/black"/>
-    <size android:width="30dp" android:height="30dp"/>
+    <size android:width="@dimen/project_list_icon_size" android:height="@dimen/project_list_icon_size"/>
 </shape>
--- a/collect_app/src/main/res/layout/project_list_item.xml
+++ b/collect_app/src/main/res/layout/project_list_item.xml
@@ -27,7 +27,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_marginStart="@dimen/margin_small"
-            android:textAppearance="@style/TextAppearance.Collect.Subtitle1"
+            android:textAppearance="?textAppearanceSubtitle1"
             tools:text="A project"/>
 
         <TextView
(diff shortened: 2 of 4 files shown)


collect_app/src/main/res/layout/project_list_item.xml
```diff
@@ -24,13 +27,15 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_marginStart="@dimen/margin_small"
-            android:textAppearance="@style/TextAppearance.Collect.Subtitle1" />
+            android:textAppearance="@style/TextAppearance.Collect.Subtitle1"
```
seadowg: These `textAppearance` attribute values should use the theme attributes instead.

collect_app/src/main/res/drawable/project_icon_background.xml
```diff
@@ -2,5 +2,5 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="oval">
     <solid android:color="@android:color/black"/>
-    <size android:width="30dp" android:height="30dp"/>
+    <size android:width="@dimen/project_list_icon_size" android:height="@dimen/project_list_icon_size"/>
```
seadowg: Personally I prefer just having these values inline as they're not going to be shared with anything else. Do you think they really need to go into `dimens.xml`?

seadowg (changes requested):
The change is really subtle but it looks better to me! There are a couple of tweaks I think we should make here.

collect_app/src/main/res/drawable/project_icon_background.xml
lognaturel: They're currently used as `marginStart` for the two settings buttons which is kind of weird. It's something I hope we come back to sooner than later but this might just be good enough that we leave it as-is. I thought adding the resource would be a useful breadcrumb for when we return to it.

collect_app/src/main/res/layout/project_list_item.xml
lognaturel: Made this change. Note it's not something I touched here, the diff was just the `/>`. I did have a related question at #4601 where I used theme attributes even though it went against the established patterns in the file I was in.

I've left the `@style/TextAppearance.Collect.Body2.MediumEmphasis` below for now since there isn't a corresponding theme attribute and the style is used in a few places. There isn't a lot of consistency across layouts currently and I think it's something that can be better addressed in a separate pass.

collect_app/src/main/res/drawable/project_icon_background.xml
seadowg: That feels like a "happens to be the same" rather than "is the same thing" to me, so I'd rather just inline it everywhere I think. The two buttons could anchor to a guideline instead of having individual margins on them as well. I'm know I'm being a bit of a stickler with all this, but these values just sit in a global namespace that's super hard to organize.

If you really think they need a shared value then I think the value should at least move to `collect_app`'s `dimens.xml` rather than have it in `androidshared`.

seadowg (changes requested):

collect_app/src/main/res/drawable/project_icon_background.xml
lognaturel: I introduced the resource because it felt helpful to connect the icon size to the margin but if it's not, then let's just use constants. That also lets me do a margin of 38 which is icon_size + margin and better aligns. 

As far as I know, the best we could do with guidelines is introduce one at 38dp which I don't think is an improvement over the two margins. The project icon is in a separate layout which is what makes doing this dynamically a trickier thing.

Hopefully this seems ok to you! (Re: dimens files, I didn't even realize I was in the global one, thanks)

seadowg: Yeah that seems good! Thanks for indulging me here.

seadowg (approved):

PR #4612, opened 2021-06-16
Run androidshared tests on CI

seadowg:
This adds `androidshared`'s test run to CI and also adds quality checks to the same module. 

In the process of getting this to work, I needed to create a theme that would let me run tests for `ColorPickerDialog` using the always `FragmentScenario` as we were relying on the Collect "Light" theme before, which can't be accessed from `androidshared`. That then led me down a short rabbit hole that where I was able to work out what was going on with us needing `launchFragment` for some dialog fragments and `launchFragmentInContainer` for others. In the end, I've added a new package that we can pull in to test fragments that brings in `androidx.fragment:fragment-testing` along with a new `FragmentTest` helper object that gives us everything we actually need to test dialog fragments. In the end this is a really positive step!

Ran tests.

I could have replaced all our usage of `RobolectricHelpers` for dialog fragment or fragment tests, but I thought deprecating a linking to the new method would be better for the moment, so I'm not spending lots of time on something we could do progressively.

No need for QA as this is just test changes.


collect_app/build.gradle
```diff
@@ -278,6 +278,7 @@ dependencies {
     implementation("org.getodk:javarosa:3.2.0-SNAPSHOT") {
         exclude group: 'joda-time'
         exclude group: 'org.slf4j'
+        exclude group: 'org.hamcrest', module: 'hamcrest-all'
```
seadowg, line 281: Not sure why JavaRosa exposes this. We definitely need to look at why we end up with so many excludes here.

fragmentstest/src/main/java/org/odk/collect/fragmentstest/FragmentsTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        )
+    }
+
+    /**
+     * Slightly easier way to perform actions/assertions against a [DialogFragment]'s UI. Using
+     * [onView] normally will match on a root view outside of where the dialog is displayed.
+     */
+    @JvmStatic
+    fun onViewInDialog(viewMatcher: Matcher<View>): ViewInteraction {
+        return onView(viewMatcher).inRoot(isDialog())
```
seadowg: This solution was hidden in a Stack Overflow rather than in Android docs (of course). The recent updates to the fragment testing docs definitely helped me, but there is still a lot that they haven't explained properly.

fragmentstest/src/main/res/values/dialog_fragment_test_theme.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="Theme.DialogFragmentTest" parent="Theme.MaterialComponents">
+        <!--
+        For some reason the default value of `alertDialogTheme` (which is
+        `ThemeOverlay.MaterialComponents.Dialog.Alert`  ) will cause `launchFragment` to
+        explode. This is most likely a bug in FragmentScenario (or Robolectric itself)
```
seadowg, line 8: This is pretty grim. I'm not even sure where to file a bug. I'll take all this stuff and play around with it sometime soon so I can understand which problems we're having with Robolectric are holes in documentation, which are bugs in androidx test and which are bugs in Robolectric itself.

.circleci/config.yml
```diff
@@ -69,6 +69,9 @@ jobs:
       - run:
           name: Run shared unit tests
           command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 shared:test
+      - run:
+          name: Run androidshared unit tests
+          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 androidshared:testDebug
```
grzesiek2010, line 74: Why for some modules we use `testDebug` but for the others it's `test`?

fragmentstest/src/main/java/org/odk/collect/fragmentstest/FragmentsTest.kt
grzesiek2010: But it's a solution to what? because it's unused at this moment.

fragmentstest/src/main/java/org/odk/collect/fragmentstest/FragmentsTest.kt
```diff
@@ -0,0 +1,47 @@
...
+ * less troublesome.
+ */
+object FragmentsTest {
+
+    /**
+     * Creates a FragmentScenario using a test theme that supports AppCompat and MaterialComponents
+     * attributes but also won't explode (like MaterialComponents themes do).
+     *
+     * This uses [FragmentScenario.launch] here instead of [FragmentScenario.launchInContainer] as
+     * per docs at https://developer.android.com/guide/fragments/test#dialog.
```
grzesiek2010: It's true that dialog fragments should be started using FragmentScenario.launch but when it comes to normal fragments using FragmentScenario.launchInContainer() is also possible. This class however is not just for dialogs right? but for all fragments, if so it should also have implementation for `launchInContainer()`.

fragmentstest/src/main/res/values/dialog_fragment_test_theme.xml
grzesiek2010: You have 2 spaces before `)`

grzesiek2010 (changes requested):
A few comments/questions and there is also one empty class named `consumer-rules.pro`

.circleci/config.yml
seadowg: `testDebug` is an Android Application/Library task, whereas `test` is just a vanilla Gradle one. If we just run `test` on an Android Application/Library we'll run every build variant (`release`, `selfSignedRelease` etc) which we don't want.

fragmentstest/src/main/java/org/odk/collect/fragmentstest/FragmentsTest.kt
seadowg: Ah my bad. I thought I had a couple of places using this, but I must have stashed them while working on this. You need to us `inRoot` to do `onView` with `DialogFragment` when using `launch` rather than `launchInContainer` and `launchInContainer` only works for `DialogFragment` classes that don't use `onCreateDialog` and just use `onCreateView` (as you're testing them as a normal fragment, not a dialog fragment).

I'll make sure this is used somewhere as an example and update the comment to be a little clearer.

fragmentstest/src/main/java/org/odk/collect/fragmentstest/FragmentsTest.kt
seadowg: I was thinking for those we could just use `FragmentScenario` directly, but I could see that it might be nice to encapsulate everything into this object. What do you think?

grzesiek2010: Yes it would be clearer if we had everything here and with comments.

seadowg: I gave it a try with `launch` and `launchInContainer` and it ended up feeling a bit unnecessary, as I don't think there is any reason not to use `FragmentScenario` directly for normal fragments. What I've done instead is renamed the helper object, so it's (hopefully) clear it's just for `DialogFragment`.

grzesiek2010 (approved):

PR #4616, opened 2021-06-17
Migrate settings on upgrade

### LINKED ISSUE #4604: Existing project settings should be migrated when upgrading from versions that need it
An upgrade from 1.30 to 2021.2 won't actually suffer any consequences of this, but upgrades from older versions (like 1.28) with different settings structures will end up with their existing project having invalid/missing settings.

For 2021.2 we can get away with only solving this for the "existing" project, but moving forward we'll need to migrate settings for all projects in the app. I personally think just implementing the latter will be simpler to think about.

seadowg:
Closes #4604 

New tests and verified manually (went back to a version that would actually apply a settings migration)

This just adds another `Upgrade` to handle the settings migration like we've doing for existing project and work manager upgrades.

I did notice as part of this that we weirdly migrate meta settings when we migrate any other settings, which now makes even less sense (due to multiple settings because of projects). Probably something to tidy up later, but no need to address right now.

We'll start looking at QA for all this soon but no need for a pass before merging.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/SettingsMigrator.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/SettingsMigrator.java
@@ -2,7 +2,7 @@ package org.odk.collect.android.application.initialization;
 
 import org.odk.collect.shared.Settings;
 
-public interface SettingsPreferenceMigrator {
+public interface SettingsMigrator {
 
     void migrate(Settings generalSettings, Settings adminSettings);
 }
--- a/collect_app/src/test/java/org/odk/collect/android/application/initialization/ExistingSettingsMigratorTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/application/initialization/ExistingSettingsMigratorTest.kt
@@ -1,0 +1,49 @@
+package org.odk.collect.android.application.initialization
+
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.`is`
+import org.hamcrest.Matchers.nullValue
+import org.junit.Test
+import org.mockito.kotlin.doReturn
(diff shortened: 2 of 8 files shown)


collect_app/src/test/java/org/odk/collect/android/application/initialization/ExistingSettingsMigratorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+        val projectsRepository = InMemProjectsRepository()
+        val project1 = projectsRepository.save(Project.New("1", "1", "#ffffff"))
+        val project2 = projectsRepository.save(Project.New("2", "2", "#ffffff"))
+
+        val project1GeneralSettings = InMemSettings()
+        val project1AdminSettings = InMemSettings()
+        val project2GeneralSettings = InMemSettings()
+        val project2AdminSettings = InMemSettings()
+
+        val settingsProvider = mock<SettingsProvider> {
```
seadowg, line 28: I think it'd be nice to have an `InMemSettingsProvider` that does the work for me here, as the test setup is pretty verbose. There are a few other tests that we could clean up with that as well, so I'll follow up with a PR for that rather than holding this up.

collect_app/src/main/java/org/odk/collect/android/application/initialization/SettingsMigrator.java
```diff
@@ -2,7 +2,7 @@
 
 import org.odk.collect.shared.Settings;
 
-public interface SettingsPreferenceMigrator {
```
seadowg, line 5: I realized this name didn't really make sense any more.

lognaturel (approved):
Thanks, I had totally forgotten about settings migration. Good thing you caught it!

Issue #4605, opened 2021-06-14
Update analytics behaviour for projects


seadowg:
- [x] Analytics should be enabled and disabled based on the project setting
- [x] Analytics should always be enabled for beta versions
~~- [ ] Form update mode user property should be set on app launch for all projects~~

PR #4619, opened 2021-06-17
Always enable analytics in beta

seadowg:
We're going to make analytics enable and disable based on the current project (#4605) but until then we'll just force it on at an app level for beta releases. I've made it so analytics and crash reporting is enabled on app start for beta releases. Also, the setting will be disabled for the user, and I've added a note to the setting summary, so it's (hopefully) clear what's happening.

Verified manually.  

Not a lot to discuss here. We already had a lot of the machinery in place for checking if the app is a beta or not so I just used that.

Good for someone else to sanity check this by making a beta build and checking they see analytics in the real time dashboard. I just hardcoded the app to be a beta to check the settings all looked correct.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
@@ -22,6 +22,7 @@ import org.odk.collect.android.logic.PropertyManager;
 import org.odk.collect.android.logic.actions.setgeopoint.CollectSetGeopointActionHandler;
 import org.odk.collect.android.storage.StorageInitializer;
 import org.odk.collect.android.utilities.LaunchState;
+import org.odk.collect.android.version.VersionInformation;
 import org.odk.collect.utilities.UserAgentProvider;
 
 import java.util.Locale;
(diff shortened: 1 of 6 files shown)


collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -80,10 +83,7 @@ private void initializeFrameworks() {
     }
 
     private void initializeAnalytics() {
-        analytics.setAnalyticsCollectionEnabled(false);
-
-//        FormUpdateMode formUpdateMode = getFormUpdateMode(context, generalSettings);
```
lognaturel, line 85: Leave this in uncommented below instead?

lognaturel (approved):
Looks great to me. I trust your verification process and I have trouble verifying analytics on my connection anyway (for Reasons) so I won't attempt it.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: I'm worried it'll produce confusing results, so I added a note in to the analytics issue (#4605) to fix it properly.

lognaturel: Got it, great.

PR #4625, opened 2021-06-18
Add icon text in preview icon

### LINKED ISSUE #4606: Polish project color dialog
- [x] Add icon text in preview icon. Ideally this would be the real icon text if it's set with a fallback to `A` but always showing `A` is fine too.

- [x] Make sure hex code text box doesn't wrap with a narrow screen (screenshot from Moto G5)
wrapping hex box

- [x] Rethink hex label. Right now it looks really random. The intent was for it to provide context for the text box so users know what to look for. Maybe "hex color code" would better achieve that. At the very least, it should be more clearly visually linked with the text box. Maybe it's not necessary.

- [x] Add validation to the hex code text box. Currently the number of characters is capped to 6. if characters outside of the hex range are included, the color just stays as it was last set with no other feedback. As soon as 6 valid hex characters are included, the color of the preview changes. One idea is to show "invalid hex code" and prevent tapping 'ok' until 6 valid digits are provided. This likely could be a separate issue handled later.

grzesiek2010:
Closes #4606

Screenshot_1624018206

No.

No.

### DIFF
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'kotlin-android'
+    id 'kotlin-kapt'
 }
 
 apply from: '../config/quality.gradle'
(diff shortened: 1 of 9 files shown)


seadowg (changes requested):
Looks good! This is definitely a nicer experience.

I think I'd prefer if the `initColor` and `icon` were Fragment args rather than data in the view model. We only set that data when opening the dialog anyway, and it doesn't feel like data the ViewModel really needs to hold on to. Is there a reason I'm missing to have those in the ViewModel?

grzesiek2010:
> I think I'd prefer if the initColor and icon were Fragment args rather than data in the view model. We only set that data when opening the dialog anyway, and it doesn't feel like data the ViewModel really needs to hold on to. Is there a reason I'm missing to have those in the ViewModel?

Ok I'm not strongly in favor of one of those solutions so ok we can do what you want.

seadowg (approved):
@grzesiek2010 thanks for making the changes!

PR #4629, opened 2021-06-21
Prevents editing of form following failed submission

### LINKED ISSUE #4589: Prevent editing of submissions that failed to send
We've been getting more and more reports of users getting submission errors because the server has already received a submission with the current submission's id but with different XML. From Central, the error message is 

> A submission already exists with this ID, but with different XML. Resubmissions to attach additional multimedia must resubmit an identical xml_submission_file.

See [this forum post](https://forum.getodk.org/t/odk-central-upload-error-a-submission-already-exists-with-this-id-but-with-different-xml/33596) for an example.

I think the majority of these errors would be prevented if we didn't allow editing of submissions that failed to send. I can't think of a downside of doing this -- it really is unexpected that those could be edited. My guess is that data collectors open failed submissions to try to see what could have happened and then save them when they exit which would change the value of an end timestamp if there is one.

Longer-term I think we should stop allowing edits of any finalized forms but that's a much more disruptive change so I'd like us to consider the narrower proposal above first.

dimwight:
Closes #4589

Tested with new `FormResubmissionTest`.

See **Design Details** below.

Small change to behaviour from user perspective. 
No obvious regression risks. 

No form required. 

https://github.com/getodk/docs/issues/1367
****
## Design Details
Please excuse this rather elaborate account; writing it did help me refine the code significantly. 
### `CursorLoaderFactory`
Just one tiny mod seems to do the job: `createSavedInstancesCursorLoader()` adds `STATUS_SUBMISSION_FAILED` to the filter along with `STATUS_SUBMITTED`.
* For demo and WIP purposes the change is switched with class variable `beforeUpdate`.
	
### `FormResubmissionTest`
Closely derived from `SendFinalizedFormTest`, with two flags setting the following server behaviours:  
* `noHttpPostResult` causes a series of calls to `Thread.sleep()` - simulating a connection drop-out.
* `rejectResubmission` sets a limit of a single call for submissions passing the same file - simulating the response to resubmission after unintended editing.

The pair of `whenFailedForm…Resubmission()` methods simulate/test 'before' and 'after' behaviours. Both methods
* initialise the appropriate value of `CursorLoaderFactory.beforeUpdate`
* depend on `handleFlag` detailed below first to set `STATUS_SUBMISSION_FAILED` and then to reject or accept resubmission.
* need manual intervention as I haven't been able to work out how to simulate clicking on a CANCEL dialog. 
* call the slightly tweaked `MainMenuPage.clickViewSentForm()`
* share preliminary code defined in `createAndSubmitFormWithFailure()`

`handleFlag()` defines server responses to the states of the above-mentioned flags. 
* Called from the override of `uploadSubmissionAndFiles()` in the anonymous `StubOpenRosaServer` passed to the `TestDependencies` in the `RuleChain`.
This approach necessitated constructors for `TestDependencies` that set the final `StubOpenRosaServer` member.
****

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/FormResubmissionTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/FormResubmissionTest.java
@@ -1,0 +1,151 @@
+package org.odk.collect.android.feature.instancemanagement;
+
+import android.Manifest;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.rule.GrantPermissionRule;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.RecordedIntentsRule;
+import org.odk.collect.android.dao.CursorLoaderFactory;
+import org.odk.collect.android.openrosa.HttpCredentialsInterface;
+import org.odk.collect.android.openrosa.HttpPostResult;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.StubOpenRosaServer;
+import org.odk.collect.android.support.TestDependencies;
+import org.odk.collect.android.support.TestRuleChain;
+import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.android.support.pages.SendFinalizedFormPage;
+
+import java.io.File;
+import java.net.URI;
+import java.util.List;
+
+import timber.log.Timber;
+
+@RunWith(AndroidJUnit4.class)
+public class FormResubmissionTest {
+
+    public static final String _FORM_NAME = "One Question";
+    public static final String _FORM_XML = "one-question.xml";
+    public static final String _QUESTION = "what is your age";
+    public static final String _ANSWER = "123";
+
+    private boolean noHttpPostResult;
+    private boolean rejectResubmission;
+    private File submissionFile;
+
+    private final StubOpenRosaServer server = new StubOpenRosaServer() {
+        @NonNull
+        @Override
+        public HttpPostResult uploadSubmissionAndFiles(@NonNull File submissionFile,
+                                                       @NonNull List<File> fileList,
+                                                       @NonNull URI uri,
+                                                       @Nullable HttpCredentialsInterface credentials,
+                                                      long contentLength) throws Exception {
+            if(noHttpPostResult||rejectResubmission){
+                boolean doReject = handleFlag(submissionFile);
+                if(doReject){
+                    return new HttpPostResult("", 500, "Resubmission not permitted for " + submissionFile.getName());
+                }
+            }
+
+            return super.uploadSubmissionAndFiles(submissionFile,
+                    fileList,
+                    uri,
+                    credentials,
+                    contentLength);
+        }
+    };
+
+    private boolean handleFlag(@NotNull File submissionFile) throws InterruptedException {
+        if (noHttpPostResult) {
+            this.submissionFile = submissionFile;
+            int timeOutMs = 1000;
+            int timeOuts = 60*60;
+            Timber.i("sleeping for %s sec", timeOutMs * timeOuts / 1000);
+            for (int timeOut = 1; timeOut <= timeOuts; timeOut++) {
+                Thread.sleep(timeOutMs);
+                Timber.i("slept for %s ms", timeOut * timeOutMs);
+            }
+        } else{
+            return rejectResubmission
+                    && this.submissionFile.equals(submissionFile) ;
+        }
+        return false;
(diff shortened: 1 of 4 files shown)


dimwight:
On further consideration, perhaps I should also update the two-parameter version of `createSavedInstancesCursorLoader`? 
And (the tricky part again) work out how to test it?

dimwight:
Particularly as it looks to be called only when constructing a filtered cursor, of which I can't find any examples in the current application?

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
```diff
@@ -13,13 +13,30 @@
 
 public class CursorLoaderFactory {
 
-    public CursorLoader createSentInstancesCursorLoader(CharSequence charSequence, String sortOrder) {
+    public static boolean beforeUpdate;
```
lognaturel: This is a good idea for a sanity check but will need to be removed before final review since we don't want that old behavior at all anymore.

collect_app/src/main/java/org/odk/collect/android/dao/CursorLoaderFactory.java
```diff
@@ -31,27 +48,13 @@ public CursorLoader createSentInstancesCursorLoader(CharSequence charSequence, S
         return cursorLoader;
     }
 
-    public CursorLoader createSentInstancesCursorLoader(String sortOrder) {
```
lognaturel, line 34: I believe the only thing that changed in this section of the diff is method order but it's not displayed very well. Is that right?

lognaturel:
> should also update the two-parameter version of createSavedInstancesCursorLoader

Yes. I believe this is used when searching the instance list for editing but this should be verified. It's important that submissions that failed to send are not included there or that would give a path to editing them. Regardless, those two methods should have the same behavior when it comes to which statuses they include.

lognaturel:
Thanks so much for taking this on and my apologies for letting it sit so long, time flies. The behavior change looks good except for, as you pointed out, extending it to both variants. We also don't want to maintain the old behavior with the `beforeUpdate` flag.

The test doesn't need to do anything server-related other than maybe getting to a failed state. The important thing to test is that if there is a submission in a failed state, it does NOT appear in the "Edit Saved Form" listing whether or not there is a search filter applied.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/FormResubmissionTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+            return super.uploadSubmissionAndFiles(submissionFile,
+                    fileList,
+                    uri,
+                    credentials,
+                    contentLength);
+        }
+    };
+
+    private boolean handleFlag(@NotNull File submissionFile) throws InterruptedException {
```
lognaturel: I don't think any of this is needed. You just need any error state, not necessarily a timeout. See https://github.com/getodk/collect/issues/4589#issuecomment-857008731 for some ideas.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/FormResubmissionTest.java
```diff
@@ -0,0 +1,151 @@
...
+                .clickOnForm(TEXT_FORM_NAME)
+                .clickSendSelected()
+//               clickOnText("CANCEL")
+                .pressBack(new MainMenuPage())
+                .clickViewSentForm(0)
+                .assertTextDoesNotExist(TEXT_FORM_NAME)
+                .pressBack(new MainMenuPage());
+    }
+
+    @Test
```
lognaturel: This one isn't needed since we don't want this old behavior anymore at all.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/FormResubmissionTest.java
```diff
@@ -0,0 +1,151 @@
...
+    public void whenFailedFormCannotBeEdited_ServerAcceptsResubmission() {
+        CursorLoaderFactory.beforeUpdate = false;
+        rejectResubmission = false;
+        MainMenuPage mainMenuPage = createAndSubmitFormWithFailure();
+        noHttpPostResult = false;
+        mainMenuPage
+                .clickEditSavedForm(1)
+                .assertTextDoesNotExist(TEXT_FORM_NAME)
```
lognaturel, line 95: I think the test can stop here! The rest is really testing the behavior of the stub server which is not behavior Collect is responsible for. Maybe the thing to add in this test is a search filter. I believe that currently if you search for the value of `TEST_FORM_NAME` you will see the submission appear even though it is in a failed state. I think it makes most sense to add that to this test but it could possibly be separated.

dimwight:
Not sure how this happened, I hadn't thought of the consequences of messing around with my commits. 

I'll take the opportunity to redo it in a less convoluted way. Should be back tomorrow 🤔

lognaturel:
> Not sure how this happened 

Did the PR autoclose? I think it's because you "merged it"... into itself! I've never seen that before!

PR #4630, opened 2021-06-21
Always fall back to the "Demo Project" name when the domain is demo.getodk.org and no project name is set

### LINKED ISSUE #4626: Always fall back to the "Demo Project" name when the domain is demo.getodk.org and no project name is set
Tapping on "Try a demo" provisions a project with name "Demo project" and corresponding icon and color. There should be the same behavior in the following cases:
- [x] manually configuring "https://demo.getodk.org" as the URL
- [x] upgrading from a prior version with the server configured as "https://demo.getodk.org"
- [x] scanning a QR code with "https://demo.getodk.org" as the URL and without a project name
- [x] scanning a QR code no (or empty) URL and without a project name

I'm hoping there's a single place where the fallback project name is defined that this special case can be added to.

grzesiek2010:
Closes #4626

I tested this change manually and improved automated tests.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/forms/FormNavigationTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/forms/FormNavigationTest.java
@@ -31,13 +31,13 @@ import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.injection.config.AppDependencyComponent;
 import org.odk.collect.android.javarosawrapper.FormController;
 import org.odk.collect.android.listeners.FormLoaderListener;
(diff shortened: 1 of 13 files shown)


seadowg (commented):
This looks good, but one of the cases in the issue isn't quite solved. Will discuss there.

seadowg (changes requested):
Ok so as per the discussion in #4626 we also need to handle empty URL cases for QR codes as that's what we'll get when scanning demo QRs from older versions.

grzesiek2010:
> Ok so as per the discussion in #4626 we also need to handle empty URL cases for QR codes as that's what we'll get when scanning demo QRs from older versions.

Done.

seadowg (approved):
Done!

PR #4635, opened 2021-06-23
Add check to stop project being deleted when it has unsent instances

### LINKED ISSUE #4577: Add check to stop project being deleted when it has unsent instances
~~Blocked by #4417~~

## Acceptance

- [x] Given I have at least one unsent filled form
When I go to Admin settings
And click "Delete project"
Then I see a message prompting me to send or delete filled forms
And the project does not get deleted

## Note

The message prompting the user should probably just be a dialog with an "OK" button (similar to an error dialog).

grzesiek2010:
Closes #4577 

This is the only solution.

No.

No.


strings/src/main/res/values/strings.xml
```diff
@@ -996,6 +996,8 @@
     <string name="delete_project_confirm_message">Are you sure you want to delete the current project?</string>
     <string name="delete_project_yes">Yes</string>
     <string name="delete_project_no">No</string>
+    <string name="cannot_delete_project_title">Cannot delete the current project!</string>
+    <string name="cannot_delete_project_message">You have unsent instances. Please send them or delete first.</string>
```
seadowg: I'm thinking we probably want to call these "filled" or "saved" forms? I also don't think we should have the "!" on the title, and we can probably drop "current". cc/ @lognaturel.

collect_app/src/androidTest/java/org/odk/collect/android/feature/projects/DeleteProjectTest.kt
```diff
@@ -41,4 +42,32 @@ class DeleteProjectTest {
             .clickAdminSettings()
             .deleteLastProject()
     }
+
+    @Test
+    fun deleteProjectWithInstancesTest() {
```
seadowg: For a test like this I think we can have a name that references the precondition and the result of that. An example could be `whenProjectHasUnsentInstances_cannotDeleteProject` or `whenProjectHasUnsentInstances_clickingDelete_showsMessage` or even just `cantDeleteProjectWithUnsentInstances`.

seadowg (changes requested):
Just a couple of tweaks before merging I think.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -153,12 +157,20 @@ public boolean onPreferenceClick(Preference preference) {
                     startActivity(pref);
                     break;
                 case DELETE_PROJECT_KEY:
-                    new AlertDialog.Builder(requireActivity())
-                            .setTitle(R.string.delete_project_confirm_message)
-                            .setNegativeButton(R.string.delete_project_no, (dialog, which) -> {
-                            })
-                            .setPositiveButton(R.string.delete_project_yes, (dialog, which) -> deleteProject())
-                            .show();
+                    if (instancesRepositoryProvider.get().getAll().isEmpty()) {
```
seadowg: Actually, I just realized: as we still want to be able to delete projects with sent instances, this isn't quite right. 

Maybe it's worth having a helper method for if you can delete projects or alternatively return a specific error from `ProjectDeleter#deleteCurrentProject` so you don't have to test both deleted and sent projects at the Espresso level?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: > Actually, I just realized: as we still want to be able to delete projects with sent instances, this isn't quite right.

Ahh right good catch!

grzesiek2010: Ok I came up with a new implementation could you take a look?

I have one problem... Have you tried using espresso with preference fragments? I can't seem to get it working. I wanted to perform simple tests that I click that `Delete project` option and check displayed dialogs using
`onView(withText(R.string.delete_project)).perform(scrollTo(), click())`
doesn't work. I tried different themes to no avail.

collect_app/src/main/java/org/odk/collect/android/database/instances/DatabaseInstancesRepository.java
```diff
@@ -96,6 +96,11 @@ public Instance getOneByPath(String instancePath) {
         }
     }
 
+    @Override
+    public List<Instance> getAllUnsent() {
```
seadowg: Could we not just use `getAllByStatus` here to avoid adding a new method to the interface that needs maintained/tested?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
```diff
@@ -2,29 +2,41 @@ package org.odk.collect.android.projects
...
+                DeleteProjectResult.DeletedSuccessfully(null)
+            }
         }
     }
 }
+
+sealed class DeleteProjectResult {
+    object UnsentInstances : DeleteProjectResult()
+
+    data class DeletedSuccessfully(val project: Project.Saved?) : DeleteProjectResult()
```
seadowg: Lets name this field `newCurrentProject` just to make it clearer which project is being returned.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
```diff
@@ -1,32 +1,114 @@
...
+import org.odk.collect.projects.ProjectsRepository
 
 class ProjectDeleterTest {
+    private val project1 = Project.Saved("1", "1", "1", "#ffffff")
+    private val project2 = Project.Saved("2", "2", "2", "#cccccc")
+    private val currentProjectProvider = mock<CurrentProjectProvider> {
+        on { getCurrentProject() } doReturn project1
+    }
+    private val instancesRepository = mock<DatabaseInstancesRepository> {
+        on { allUnsent } doReturn emptyList()
```
seadowg: I think it might be nicer to just use `InMemInstancesRepository` as it means there is less enforcing of the implementation (we just set up the state rather than enforce what method is called). What do you think?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
```diff
@@ -235,14 +231,22 @@ public CharSequence provideSummary(Preference preference) {
...
-        Project.Saved newProject = projectDeleter.deleteCurrentProject();
-
-        if (newProject != null) {
-            ActivityUtils.startActivityAndCloseAllOthers(requireActivity(), MainMenuActivity.class);
-            ToastUtils.showLongToast(getString(R.string.switched_project, newProject.getName()));
-        } else {
-            ActivityUtils.startActivityAndCloseAllOthers(requireActivity(), SplashScreenActivity.class);
+        DeleteProjectResult deleteProjectResult = projectDeleter.deleteCurrentProject();
+
+        if (deleteProjectResult instanceof DeleteProjectResult.UnsentInstances) {
```
seadowg: Maybe worth a Kotlin convert here to allow us to use `when`?

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
seadowg: Oh I think the preferences fragments might use `RecyclerView` under the hood so you probably need to use the specific actions it needs. We have `Page#scrollToRecyclerViewItemAndClickText` for that. Maybe it's worth making an abstract `Page` subclass for "settings" pages with helpers like `clickSetting` or something, so we only have to update the helpers in one place when they inevitably introduce a new list view...

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: I like converting into kotlin so I thought about it, the problem is that it extends another class (Java class) so I thought you wouldnt like it but if you are ok with it I can do that

seadowg: Should be fine? I think if it causes problems then yeah we can make it a separate PR, but I'm happy for a subclass of a Java class to use Kotlin

grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
grzesiek2010: Yeas, done.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/database/instances/DatabaseInstancesRepository.java
grzesiek2010: Yes we can do that, done.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/AdminPreferencesFragment.java
grzesiek2010: First of all it's in `collect_app/androidTest` so we need to move it to `testshared` probably and put in a package that will be shared between both types of tests.
When it comes to that method I tried it but to no avail. I think we need to create separate issues for that, one for moving those espresso helpers and another one for investigating how to use preference fragments in fragment scenario.
What do you think?

strings/src/main/res/values/strings.xml
grzesiek2010: I updated those strings so that now they look more like those from https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=0%3A1

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
```diff
@@ -1,32 +1,158 @@
...
+            mock(),
+            currentProjectProvider,
+            formUpdateManager,
+            instanceSubmitScheduler,
+            instancesRepository
+        )
+
+        deleter.deleteCurrentProject()
+
+        verify(instancesRepository)
```
seadowg: We don't need to use a spy here. For me, mockito spies can be really confusing as it allows `when` implementation stubbing on real objects which can make the test boundaries really blurry, so I'd prefer to avoid them at all costs.

Instead, we can just add instances with the statuses and check that it doesn't delete in those cases (like you do in `If there are unsent instances the project should not be deleted`). You could just have a test for each kind of instance that counts as unsent, like:

* `If there are incomplete instances the project should not be deleted`
* `If there are complete instances the project should not be deleted`
* `If there are submission failed instances the project should not be deleted`

This approach also has the advantage (in my mind) of taking a state/behaviour approach: we set up the state and check what the behaviour is. That for me always feels clearer and more protective than verify what interactions get made (although there are of course different schools of thought on this).

Alternatively, we can of course use a mock just in one test, and keep the class field a fake, but we still need a test to verify the return value when there are unsent forms for at least one of the instance status types.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
grzesiek2010: Ok, fixed.

seadowg (approved):

PR #4637, opened 2021-06-23
Add instances from disk

### LINKED ISSUE #4474: Instances added/removed on disk should update in Collect

seadowg:
Closes #4474

This makes it, so instances added on disk are added to the app. The app checks for new instances on disk whenever the user views the main menu.

This also adds analytics to see if this feature is ever used at all, so we can gauge whether to continue to maintain it or not.

Added new tests to check the cases described in the issues.

For right now I've just taken the existing `AsyncTask`, converted it to synchronous code and called it from a coroutine that's triggered when the main menu's `onResume` is called. I kept this confined to the main menu, so there would be only one place where the task is triggered. It also fits here as it's already where we update our counts of instances on the device (which is the closest thing we have to an instance "refresh").

There's definitely a lot more refactoring and clean up that we could do here, and I think even some redesign of how this feature works, but I don't think any of that is worth it until we're sure people are actually using it. For right now, it feels like having the core flow tested and the code in one place is enough.

Should remove the weird problem with progress bars never going away and add the features described in the issue! I don't think there is any reason to QA this specifically.


collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/InstancesAdbTest.kt
```diff
@@ -0,0 +1,52 @@
...
+            .swipeToEndScreen()
+            .clickSaveAndExit()
+
+        val instancesDir = testDependencies.storagePathProvider.getOdkDirPath(StorageSubdirectory.INSTANCES)
+        val instanceDeleted = File(instancesDir).listFiles()[0].deleteRecursively()
+        Assert.assertTrue(instanceDeleted)
+
+        mainMenuPage
+            .clickEditSavedForm(1)
+            .clickOnText("One Question")
```
seadowg, line 50: Interestingly, what happens here is that the filled form loads fine, but then has no answers in it. It also appears you can save the form again. We could put more work in to show a specific error message, but I think that should be a follow-up, so we can discuss what to do there.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
```diff
@@ -88,11 +89,11 @@ private void resetDagger() {
...
     }
 
-    private void clearPrefs() {
-        settingsProvider.getGeneralSettings().clear();
-        settingsProvider.getGeneralSettings().setDefaultForAllSettingsWithoutValues();
-        settingsProvider.getAdminSettings().clear();
-        settingsProvider.getAdminSettings().setDefaultForAllSettingsWithoutValues();
-        settingsProvider.getMetaSettings().clear();
+    private void clearPrefs(Context context) {
+        settingsProvider.clearAll();
```
seadowg, line 93: I ran into problems where some state from the old version of the app would lie around and realized we needed some more aggressive housekeeping here.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/InstancesAdbTest.kt
lognaturel: That seems like it could be really confusing. We have reason to believe it would never happen so maybe it doesn't matter. Is it a state we could detect and track analytics for?

lognaturel (approved):
Going to think more about the deletion case but the rest looks good to me.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/InstancesAdbTest.kt
seadowg: Yeah we could! We'd need to dig into detecting that in `FormEntryActivity` so it's about as much work as showing an error in the same case.

seadowg:
@lognaturel I'm going to merge this, and we can follow up with any tweaks we think we need to the deletion case.

PR #4639, opened 2021-06-24
Add validation for project details import

### LINKED ISSUE #4632: Add validation for project details import
## Acceptance

- [x] Given I've generated a QR code with project details
And the project icon value is something longer than one character
When I try to add a project using the QR code
Then the project imports successfully
And the project icon is just the first character of the value in the QR code

- [x] Given I've generated a QR code with project details
And the project color value is not a valid color hex code
When I try to add a project using the QR code
Then the project imports successfully
And the project color is based on the name

- [x] Given I've generated a QR code with project details
And only the project name is included
When I try to add a project using the QR code
Then the project imports successfully
And the project icon is based on the name (not the server URL)
And the project color is based on the name (not the server URL)

grzesiek2010:
Closes #4632 

I tested this change manually and improved automated tests.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
@@ -22,7 +22,7 @@ class ProjectDetailsCreator(private val context: Context) {
             if (Character.codePointCount(icon, 0, icon.length) == 1) {
                 icon.toUpperCase(Locale.US)
             } else {
-                getFirstSing(icon)
+                getFirstSign(icon)
             }
         } else {
             projectName.first().toUpperCase().toString()
@@ -68,11 +68,11 @@ class ProjectDetailsCreator(private val context: Context) {
             .matches()
     }
 
-    private fun getFirstSing(value: String): String {
+    private fun getFirstSign(value: String): String {
         return if (Character.codePointCount(value, 0, value.length) == 1) {
             value
         } else {
-            getFirstSing(value.substring(0, value.length - 1))
+            getFirstSign(value.substring(0, value.length - 1))
         }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
@@ -5,32 +5,74 @@ import androidx.core.content.ContextCompat
 import org.odk.collect.android.R
 import org.odk.collect.projects.Project
 import java.net.URL
+import java.util.Locale
+import java.util.regex.Pattern
 import kotlin.math.abs
 
 class ProjectDetailsCreator(private val context: Context) {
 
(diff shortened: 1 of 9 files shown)


collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
```diff
@@ -5,32 +5,74 @@ import androidx.core.content.ContextCompat
...
     }
+
+    private fun isProjectColorValid(hexColor: String): Boolean {
+        return Pattern
+            .compile("^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})\$", Pattern.CASE_INSENSITIVE)
+            .matcher(hexColor)
+            .matches()
+    }
+
+    private fun getFirstSing(value: String): String {
```
seadowg: I think the method name is a typo here?

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
```diff
@@ -15,43 +15,87 @@ class ProjectDetailsCreatorTest {
...
 
     @Test
-    fun `Project icon should be generated from url`() {
-        assertThat(projectDetailsCreator.getProject("https://my-project.com").icon, `is`("M"))
-        assertThat(projectDetailsCreator.getProject("https://your-project.com/one").icon, `is`("Y"))
-        assertThat(projectDetailsCreator.getProject("http://www.my-project.com").icon, `is`("W"))
-        assertThat(projectDetailsCreator.getProject("").icon, `is`("P")) // default project icon for invalid urls
-        assertThat(projectDetailsCreator.getProject("something").icon, `is`("P")) // default project icon for invalid urls
+    fun `If project name is not included in project details or empty should be generated based on url`() {
+        // Case1: not included
```
seadowg: I think we should pull all these "cases" out in to separate tests

seadowg (changes requested):
Just a couple of tweaks here, but all looks great. Was able to try it out by hacking the QR generation code.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
grzesiek2010: You mean never have two cases in one test or what?

seadowg: I just think if we feel like it's worth the comments, then we can just pull them out into separate tests and use the test names instead.

grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectDetailsCreator.kt
grzesiek2010: Done.

seadowg (approved):

PR #4642, opened 2021-06-24
Fix multiclicking in ManualProjectCreatorDialog 

### LINKED ISSUE #4641: Able to multiclick on Add button on manually enter project details view
Collect v2021.2, Android v all device used...

#### Problem description
I have noticed that I can click on Add button on manually enter project details view multiple times and add the same project multiple times. 
ezgif com-gif-maker (2)
(I am not sure why this video has such bad quality - I used the same app as always)

1. Click on Add project
2. Click on Manually enter project details
3. Input URL
4. Tap multiple times on Add button

#### Expected behavior
User should not be able to Add multiple project by multi-clicking on Add button

#### Other information 

grzesiek2010:
Closes #4641

It's the only solution.

No.

No.

### DIFF
--- a/collect_app/src/main/res/layout/manual_project_creator_dialog_layout.xml
+++ b/collect_app/src/main/res/layout/manual_project_creator_dialog_layout.xml
@@ -162,7 +162,7 @@
         android:src="@drawable/shadow_up"
         app:layout_constraintBottom_toTopOf="@+id/cancel_button" />
 
-    <com.google.android.material.button.MaterialButton
+    <org.odk.collect.android.views.MultiClickSafeButton
         android:id="@+id/cancel_button"
         style="@style/Widget.MaterialComponents.Button.TextButton"
         android:layout_width="wrap_content"
@@ -173,7 +173,7 @@
         app:layout_constraintEnd_toStartOf="@+id/add_button"
         app:layout_constraintTop_toTopOf="@+id/add_button" />
 
-    <com.google.android.material.button.MaterialButton
+    <org.odk.collect.android.views.MultiClickSafeButton
         android:id="@+id/add_button"
         style="@style/Widget.MaterialComponents.Button.TextButton"
         android:layout_width="wrap_content"


collect_app/src/test/java/org/odk/collect/android/projects/ManualProjectCreatorDialogTest.kt
```diff
@@ -121,4 +123,13 @@ class ManualProjectCreatorDialogTest {
             Intents.release()
         }
     }
+
+    @Test
+    fun `MultiClickSafeButton should be used to protect from multiclicking`() {
```
seadowg: OK, this is very opinionated and a bit hand wavy: I honestly wouldn't bother with tests like these that just check types. It ends up feeling like the test and the implementation are exact mirrors of each other and don't really enforce behaviour. We have to trust that `MultiClickSafeButton` does the right thing for this test to really mean anything, so it doesn't really give us much confidence or protection, I'd argue.

Interested in your thoughts!

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/projects/ManualProjectCreatorDialogTest.kt
grzesiek2010: `MultiClickSafeButton` is widely used so I think confirming it's present like in this test in enough. I thought we had tests for that button (for `MultiClickSafeImageButton` and `MultiClickSafeTextInputEditText` too) but it looks like we don't. So I would rather add tests for those views and in other places only check if they are used. I don't think it should block this issue because we have believed it worked so far (in like 10 classes) we can do that now too and create a separate issue for adding tests.

seadowg: Oh sorry, I didn't mean we needed more tests, I meant we can just delete these tests. I feel like just making the button a `MultiClickSafeButton` is enough, as we "trust" it (like you say). This is one of the few times I'm saying we can just get rid of a test rather than asking for more haha.

grzesiek2010: Ok but then we need something more I think so I created this issue https://github.com/getodk/collect/issues/4645

seadowg (approved):

PR #4643, opened 2021-06-25
Don't configure ReferenceManager when doing form discovery parse

### LINKED ISSUE #4634: Possible ReferenceManager conflict when form filling and/or background form update(s) happen concurrently
@seadowg pointed out when fixing background form updates for multiple projects that the singleton `ReferenceManager` is a problem if two different projects begin form update jobs at the same time (https://github.com/getodk/collect/pull/4593#issuecomment-856260937) and/or form filling and a form update job start at the same time. I believe this is an existing issue that was introduced by https://github.com/getodk/collect/pull/3523 which added `ReferenceManager` configuration to the initial form discovery pass. Even within one project, it would be possible to launch form filling and have another form automatically start updating at the same time. In that case, one would have the final say on the `ReferenceManager` configuration and the other would fail.

The introduction of projects makes this a little more obvious and likely because it's now possible for multiple automatic form update jobs to start at the same time.

Some options to explore:
- Avoid the need to configure the `ReferenceManager` in form discovery by addressing https://github.com/getodk/javarosa/issues/444
- Have the `ReferenceManager` manage several forms' paths
- Offer a JavaRosa code path that bypasses the `ReferenceManager` for building a `FormDef` by passing in secondary instance paths (this sounds really annoying now that I write it out)

Thoughts on desired behavior:
- Ideally the user does get feedback at time of form download if a secondary instance is missing (though other media attachments can be missing with no feedback until form filling so maybe not?) Related: https://github.com/getodk/collect/issues/4489

lognaturel:
Closes #4634 

The goal is to make it so `ReferenceManager` is only configured when a user loads a form for filling. I did a search through the codebase for `ReferenceManager.` and made sure all calls are access-only. I removed configuration from the disk sync task and metadata parsing (incidentally, there was some redundancy there so this is a good change overall).

Allowing form parses when a secondary instance is required without a configured `ReferenceManager` required JavaRosa changes. Those are at https://github.com/getodk/javarosa/pull/635 and include some new tests. The JR snapshot has now been updated to include those changes.

Manually tried the following cases:
- Download a form that requires a CSV external secondary instance [external-select-csv.xml.zip](https://github.com/getodk/collect/files/6712840/external-select-csv.xml.zip) without attaching the CSV. The download succeeds. Opening the form succeeds but there are no choices available.
-  Download a form that requires a CSV external secondary instance used in a select and attach a CSV with missing `label` and `name` columns. The download fails with `Failure`, as it does on master.
- Download a form that requires a CSV external secondary instance without attaching the CSV. Then attach a CSV with missing `label` and `name` columns and download the update. The update succeeds. Empty choices are displayed. Selecting a choice and moving to the next question shows an error.
- Download a form in one project, switch projects to a project with a form that has media, verify that the media is visible (e.g. All Widgets and Likert widget images)
- Download a form with a last-saved reference, then open it and verify last-saved is used
- Verify forms with media work on first and second open and with switches between projects
- adb push forms that have a last-saved call, that require an external CSV without pushing that CSV, verify they are correctly discovered. Verify adding the external CSV makes the choices appear.

It's relatively simple and improves Collect code quality.

I tried to add a helpful message when the file is missing (https://github.com/getodk/collect/pull/4643/commits/b953ef946eb1767d7ba4e60184a31a289c19e9f5) but I backed out of that because we can't currently tell the difference between a missing external secondary instance and a choice filter that doesn't return any items. Adding a message made a common case worse: having a cascade of selects in a field-list. In that case, you want all the lower-level questions to show no options until selections are made further up the chain.

This is pretty risky. Everything that touches media is affected. Everything that discovers forms is affected (automatic form downloads, disk sync, manual form download). The biggest risk is that media doesn't show up somewhere it should, especially when switching projects.

See cases above. Forms with last-saved an XML and CSV external secondary instances.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
@@ -28,17 +28,13 @@ public class FormUtils {
      * directory with name matching the name of the directory represented by {@code formMediaDir}.
      * <p>
      * E.g. if /foo/bar/baz is passed in as {@code formMediaDir}, jr:// URIs will be resolved to
-     * /odk/root/forms/baz.
+     * projectRoot/forms/baz.
      */
     public static void setupReferenceManagerForForm(ReferenceManager referenceManager, File formMediaDir) {
-        // Clear mappings to the media dir for the previous form that was configured
-        referenceManager.clearSession();
+        referenceManager.reset();
 
-        // This should get moved to the Application Class
-        if (referenceManager.getFactories().length == 0) {
-            // Always build URIs against the ODK root, regardless of the absolute path of formMediaDir
-            referenceManager.addReferenceFactory(new FileReferenceFactory(new StoragePathProvider().getProjectRootDirPath()));
-        }
+        // Always build URIs against the project root, regardless of the absolute path of formMediaDir
+        referenceManager.addReferenceFactory(new FileReferenceFactory(new StoragePathProvider().getProjectRootDirPath()));
 
         addSessionRootTranslators(referenceManager,
                 buildSessionRootTranslators(formMediaDir.getName(), enumerateHostStrings()));
@@ -50,7 +46,7 @@ public class FormUtils {
 
     public static List<RootTranslator> buildSessionRootTranslators(String formMediaDir, String[] hostStrings) {
         List<RootTranslator> rootTranslators = new ArrayList<>();
-        // Set jr://... to point to /sdcard/odk/forms/formBasename-media/
+        // Set jr://... to point to <projectRoot>/forms/formBasename-media/
         final String translatedPrefix = String.format("jr://file/forms/" + formMediaDir + "/");
         for (String t : hostStrings) {
             rootTranslators.add(new RootTranslator(String.format("jr://%s/", t), translatedPrefix));
--- a/collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
@@ -2,8 +2,6 @@ package org.odk.collect.android.formmanagement;
 
 import com.google.common.io.Files;
 
-import org.javarosa.core.reference.InvalidReferenceException;
-import org.javarosa.core.reference.ReferenceManager;
 import org.junit.Before;
 import org.junit.Test;
 import org.odk.collect.android.utilities.FileUtils;
@@ -13,18 +11,13 @@ import java.util.Map;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.fail;
 
 public class FormMetadataParserTest {
 
     private File mediaDir;
-    private ReferenceManager referenceManager;
 
     @Before
     public void setup() {
-        referenceManager = ReferenceManager.instance();
-        referenceManager.reset();
-
         mediaDir = Files.createTempDir();
     }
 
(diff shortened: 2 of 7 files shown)


lognaturel:
@seadowg we'd talked about me merging and doing another beta but since the last beta we published is still in review for unclear reasons I'd rather wait. I think it would also be a good idea for you to do a quick sanity check to try to identify any cases I may have still missed.

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -60,43 +59,17 @@ public void canParseFormWithLastSaved() throws Exception {
         File formXml = File.createTempFile("form", ".xml");
         FileUtils.write(formXml, LAST_SAVED.getBytes());
 
-        FormMetadataParser formMetadataParser = new FormMetadataParser(referenceManager);
+        FormMetadataParser formMetadataParser = new FormMetadataParser();
         Map<String, String> metaData = formMetadataParser.parse(formXml, mediaDir);
         assertThat(metaData.get(FileUtils.FORMID), is("basic-last-saved"));
     }
 
-    /**
```
lognaturel, line 68: This will now parse with the placeholder instance but there will be no choices to pick from. That case is tested in JavaRosa. It feels esoteric enough that it doesn't need to be tested again.

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
```diff
@@ -28,17 +28,13 @@ private FormUtils() {
...
-     * /odk/root/forms/baz.
+     * projectRoot/forms/baz.
      */
     public static void setupReferenceManagerForForm(ReferenceManager referenceManager, File formMediaDir) {
-        // Clear mappings to the media dir for the previous form that was configured
-        referenceManager.clearSession();
+        referenceManager.reset();
 
-        // This should get moved to the Application Class
-        if (referenceManager.getFactories().length == 0) {
```
seadowg, line 38: When I was in here recently (fixing the `StoragePathProvider` call) I didn't think the check made sense at all - why would we care if there are already configured factories? I'm guessing our thinking lines up here, but I just wanted to check.

collect_app/src/test/java/org/odk/collect/android/formmanagement/FormMetadataParserTest.java
```diff
@@ -37,7 +36,7 @@ public void canParseFormWithExternalSecondaryInstance() throws Exception {
         File externalInstance = new File(mediaDir, "external-data.xml");
         FileUtils.write(externalInstance, EXTERNAL_INSTANCE.getBytes());
 
-        FormMetadataParser formMetadataParser = new FormMetadataParser(referenceManager);
```
seadowg, line 40: I think we also want to remove the `referenceManager` field from the test class as well right? There are tests that use it that happen to pass right now but they're definitely misleading.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
lognaturel: It used to work out because there was only one factory to add -- the one on the ODK root. I think at that point without the check there would just have been lots of factories with the same root. Now we use a whole different context per project so I thought it was clearer to just always reset and configure a fresh factory (as opposed to detecting whether we've changed projects and only reconfiguring if we have).

There's a lot of weirdness here in general. I don't know why JR decided to have different prefixes for different file types. This whole code here is set up to erase that distinction -- we basically say no matter what the prefix is, first pretend it's jr://file and then resolve that.

lognaturel: Oh and just another note since I was also confused by all this. There used to be a call on `referenceManager.clearSession()` which did NOT reset the factories. That's why the test had value and why without it there'd be lots of duplicate factories.

seadowg (approved):

PR #4648, opened 2021-06-28
Use first project as default in Content Providers

### LINKED ISSUE #4462: Content providers should work with projects
~~Blocked by #4417~~

**Acceptance**

- [x] Interacting with forms from the content provider should work as if using the first created (chronologically) project
- [x] Interacting with instances from the content provider should work as if using the first created (chronologically) project

seadowg:
Closes #4462

New tests and verified nothing in Collect is broken manually.

Comments inline.

I think it'd be good for QA to look at this to see if this resolved other issues we were seeing with CollectTester. I'll mark this as needing testing (cc/ @mmarciniak90 @kkrawczyk123).

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
@@ -5,7 +5,6 @@ import android.content.Intent;
 import androidx.test.core.app.ActivityScenario;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
@@ -29,7 +28,6 @@ public class AndroidShortcutsTest {
             .around(rule);
 
     @Test
-    @Ignore("See comment in #pickAndLaunchShortcutForForm")
     public void canFillOutFormFromShortcut() {
         rule.startAtMainMenu()
                 .copyForm("one-question.xml")
@@ -46,9 +44,9 @@ public class AndroidShortcutsTest {
         Intent shortcutIntent = resultData.getParcelableExtra(Intent.EXTRA_SHORTCUT_INTENT);
 
          /*
-        This launch call doesn't work because of: https://github.com/android/android-test/issues/496
+        This can't use ActivityScenario.launch because of: https://github.com/android/android-test/issues/496
          */
-        ActivityScenario.launch(shortcutIntent);
+        scenario.onActivity(activity -> activity.startActivity(shortcutIntent));
         return new FormEntryPage(formName);
     }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
@@ -3,6 +3,7 @@ package org.odk.collect.android.feature.external;
 import android.content.Intent;
 
 import androidx.test.core.app.ActivityScenario;
+import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import org.junit.Rule;
@@ -46,7 +47,8 @@ public class AndroidShortcutsTest {
          /*
         This can't use ActivityScenario.launch because of: https://github.com/android/android-test/issues/496
          */
-        scenario.onActivity(activity -> activity.startActivity(shortcutIntent));
+        shortcutIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        ApplicationProvider.getApplicationContext().startActivity(shortcutIntent);
         return new FormEntryPage(formName);
     }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/forms/FormNavigationTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/instrumented/forms/FormNavigationTest.java
@@ -65,7 +65,7 @@ public class FormNavigationTest {
             .around(new RunnableRule(() -> {
                 // Set up demo project
                 AppDependencyComponent component = DaggerUtils.getComponent(ApplicationProvider.<Application>getApplicationContext());
-                component.projectImporter().importDemoProject();
+                component.projectImporter().importNewProject(Project.Companion.getDEMO_PROJECT());
                 component.currentProjectProvider().setCurrentProject(Project.DEMO_PROJECT_ID);
             }));
 
--- a/collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
+++ b/collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
@@ -14,11 +14,8 @@
 
 package org.odk.collect.android.dao.helpers;
 
-import android.net.Uri;
-
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.javarosawrapper.FormController;
-import org.odk.collect.android.provider.InstanceProviderAPI;
 import org.odk.collect.android.utilities.InstancesRepositoryProvider;
 import org.odk.collect.forms.instances.Instance;
 import org.odk.collect.forms.instances.InstancesRepository;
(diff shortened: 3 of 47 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
```diff
@@ -29,7 +29,6 @@
             .around(rule);
 
     @Test
-    @Ignore("See comment in #pickAndLaunchShortcutForForm")
```
seadowg, line 32: I definitely wanted this running while making these changes so managed to find a workaround.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/forms/FormNavigationTest.java
```diff
@@ -65,7 +65,7 @@
             .around(new RunnableRule(() -> {
                 // Set up demo project
                 AppDependencyComponent component = DaggerUtils.getComponent(ApplicationProvider.<Application>getApplicationContext());
-                component.projectImporter().importDemoProject();
+                component.projectImporter().importNewProject(Project.Companion.getDEMO_PROJECT());
```
seadowg, line 68: For some reason I couldn't get the companion field to be static (even with `@JvmStatic`)

collect_app/src/main/java/org/odk/collect/android/projects/ProjectImporter.kt
```diff
@@ -16,12 +16,6 @@ class ProjectImporter(
         return savedProject
     }
 
-    fun importDemoProject() {
```
seadowg, line 19: Felt like it made sense just to inline this to trim down the interface.

collect_app/src/main/java/org/odk/collect/android/dao/helpers/InstancesDaoHelper.java
```diff
@@ -69,19 +66,6 @@ public static boolean isInstanceComplete(boolean end, boolean completedByDefault
         return complete;
     }
 
-    // TODO: replace with method in {@link org.odk.collect.android.instances.InstancesRepository}
-    // that returns an {@link Instance} object from a path.
-    public static Uri getLastInstanceUri(String path) {
```
seadowg, line 74: It looks like the repository method was added, so there wasn't really a need for this helper any longer.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
```diff
@@ -24,7 +24,6 @@ object DatabaseFormColumns : BaseColumns {
     const val DISPLAY_SUBTEXT = "displaySubtext" // not used in the newest database version
     const val MD5_HASH = "md5Hash"
     const val DATE = "date"
-    const val MAX_DATE = "MAX(date)" // used only to get latest forms for each form_id
```
seadowg, line 27: This was confusing me (again) so I got rid of it. It's only used for one query so definitely shouldn't be referenced globally.

collect_app/src/test/java/org/odk/collect/android/activities/FirstLaunchActivityTest.kt
```diff
@@ -24,52 +21,23 @@ import org.odk.collect.android.R
...
 import org.odk.collect.android.projects.QrCodeProjectCreatorDialog
-import org.odk.collect.android.storage.StoragePathProvider
 import org.odk.collect.android.support.CollectHelpers
 import org.odk.collect.android.utilities.TranslationHandler.getString
 import org.odk.collect.android.version.VersionInformation
-import org.odk.collect.projects.ProjectsRepository
 
 @RunWith(AndroidJUnit4::class)
 class FirstLaunchActivityTest {
-    @Test
```
seadowg, line 37: This behaviour is all covered by our `LaunchScreenTest`

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
lognaturel: Huh! I vaguely remember we didn't like this but couldn't come up with an alternative at the time. This works with the adapter? Are there sufficient tests around this behavior that you're confident about the change?

@getodk/testers there's a change to how the setting to only show the latest version of each form definition that feels risky to me. I haven't had a chance to look for automatic test coverage but regardless, this would be a good one to verify manually.

seadowg: Yeah we have a `HideOldVersionsTest` that checks this. I've not changed how the query works (it's still doing a `GROUP BY` and using `MAX`), I've just moved all of that into the provider and then used a projection map so that the rest of the app doesn't have to know about it.

kkrawczyk123:
@seadowg  this is a list of verified cases, please let me know if there's something more you can think of it would be good to verify here:
- use all views in CollectTester: no crashes detected
- using CollectTester with diffetent projects

server defined in Collect: server in CollectTester is empty
- download form
- fill submission
- send submission
- submit instance: I am getting error message "Host name may not be null" and I am unable to send submission. I have compared with 1.30.1 and other older versions and this issue is also visible there so it looks like it's not connected to that PR, maybe it's connected to Collecttester and it has never been implemented. I can't see such test case on our regression checklist so I am not even sure if it has ever worked at all.

server defined in CollectTester:
- download form
- fill submission
- send submission 
- submit instance: submissions are sent successfully 
- server requires authentication dialog
- server that does not require credentials
- dialog when try to send not existing submission 


- quick regression check on counter app and external app

 form update:
 - download form updates
- hide old version of form selected and unselected
- fill and send both versions

I have noticed weird behavior on Android 11 - after downloading form update in Fill Blank Form I can still see the older version. On other devices, I can see correct behavior and the newest version is visible (Androids 5.1, 8.1 and 10) I was able to reproduce it for both Aggregate and Central. 
see gif: 
ezgif com-gif-maker (3)

seadowg:
@kkrawczyk123 huh so the "hide old version" isn't working but only in Android 11? Very strange. I'll look into that.

kkrawczyk123:
@seadowg yes, exactly.

seadowg:
@kkrawczyk123 are you able to double-check that it works fine on Android 11 for 1.30.1 as well as me if you get a chance?

kkrawczyk123:
@seadowg I've just confirmed that it works well on 1.30.1.

seadowg:
@kkrawczyk123 yeah can confirm that we are seeing `HideOldVersionsTest#whenHideOldVersionsEnabled_onlyTheNewestVersionOfAFormShowsInFormList` fail on Android 11. I have zero ideas about what the problem could be at the moment, but will dig into it. I might just need to revert some of my changes, but that'll be a shame, as it will move some nastiness back into the codebase.

seadowg:
@kkrawczyk123 ok fixed! This was a really weird one where the feature only happened to work on other versions of Android. I think Android 11 must have an update to SQLite that changes the default order that rows come back in or something, and that   exposed the problem. Yuck.

collect_app/src/main/java/org/odk/collect/android/database/forms/DatabaseFormColumns.kt
lognaturel: Great!

collect_app/src/main/java/org/odk/collect/android/provider/FormsProviderAPI.java
```diff
@@ -56,9 +56,6 @@ public static Uri getContentUri(String projectId) {
      * @deprecated This should be implemented as part of the UI/app code and should not be
```
lognaturel: Remove comment

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
```diff
@@ -113,25 +112,46 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
...
 
             case NEWEST_FORMS_BY_FORM_ID:
-                Collection<String> allColumns = new HashSet<>(PROJECTION_MAP.values());
-                allColumns.remove(DATE);
-                allColumns.add("MAX(date)");
-
-                Map<String, String> maxDateProjectionMap = new HashMap<>(PROJECTION_MAP);
-                maxDateProjectionMap.put("MAX(date)", DATE);
+                Set<String> maxDateColumns = new HashSet<>();
+                maxDateColumns.add(_ID);
```
lognaturel, line 120: And there's no way to get a projection map for all columns, is there? Or a set of all column names or something that would keep us from having to ever touch this again?

lognaturel (approved):
Thanks for making the history really easy to follow, that really helped.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProviderAPI.java
seadowg: This is still valid, I think?

collect_app/src/main/java/org/odk/collect/android/provider/FormsProvider.java
seadowg: Not right now. We could reorg the way the columns are defined to set that up for ourselves, but given this whole way of accessing data is deprecated, I'm not sure if that makes sense right now.

collect_app/src/main/java/org/odk/collect/android/provider/FormsProviderAPI.java
lognaturel: I guess the parts about where things should go are? I thought there was specific mention of constants that adds removed. But anyway, not very important.

kkrawczyk123:
**Tested with success!**

re-tested form update:
- download form updates
- hide old version of form selected and unselected
- fill and send both versions
- issue from Android 11 has been fixed
- automatic update
- notification about form update
- Aggregate and Central

PR #4650, opened 2021-06-28
Fix crash when opening QR scanner

### LINKED ISSUE #4649: App crashes on Android 8 when adding project via QR code
The app crashes when clicking "Configure with QR code" on Android 8 with the following exception:

```java
Fatal Exception: java.lang.RuntimeException: Failure delivering result ResultInfo{who=@android:requestPermissions:, request=42, result=-1, data=Intent { act=android.content.pm.action.REQUEST_PERMISSIONS (has extras) }} to activity {org.odk.collect.android/com.karumi.dexter.DexterActivity}: java.lang.IllegalStateException: Only fullscreen activities can request orientation
       at android.app.ActivityThread.deliverResults(ActivityThread.java:4496)
       at android.app.ActivityThread.handleSendResult(ActivityThread.java:4539)
       at android.app.ActivityThread.-wrap20(ActivityThread.java)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1757)
       at android.os.Handler.dispatchMessage(Handler.java:105)
       at android.app.ActivityThread.main(ActivityThread.java:6949)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
```

It also looks like the first launch screen does not respect device orientation on some devices/API levels. This is most likely due to the theme being used and is probably worth fixing at the same time.

## Notes
[Firebase crash report](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8a8f12acd04b9cb7d90cac6546e121a9?time=last-seven-days&versions=v2021.2-beta.0%20(4182)&sessionEventKey=60D1A19303900001596CB061690271D5_1554980593949325369)

seadowg:
Closes #4649

This PR fixes the crash but it will also allow barcode scanning screens to be rotated which we may not want.

These changes feel pretty "configuration level" to me, so I only validated manually. If we feel like we want to add orientation locking back in then I think we should have tests for that, however.

Comments inline.

I think we need to play with this to check we're OK with the behaviour change and if not we can look at adding orientation locks back in but in a more controlled way. Another option is just to let the bug on the launch screen (going to scan a QR code and then being locked in an orientation) continue to exist and look at fixing that later.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FirstLaunchActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FirstLaunchActivity.kt
@@ -28,6 +28,8 @@ class FirstLaunchActivity : CollectAbstractActivity() {
 
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        setTheme(R.style.Theme_Collect_Light)
+
         binding = FirstLaunchLayoutBinding.inflate(layoutInflater)
         setContentView(binding.root)
         DaggerUtils.getComponent(this).inject(this)
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/CodeCaptureManagerFactory.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/CodeCaptureManagerFactory.kt
@@ -27,6 +27,7 @@ object CodeCaptureManagerFactory {
         return IntentIntegrator(activity)
             .setDesiredBarcodeFormats(supportedFormats)
             .setPrompt(prompt)
+            .setOrientationLocked(false) // Let UI control orientation lock
             .addExtra(Intents.Scan.SCAN_TYPE, Intents.Scan.MIXED_SCAN)
             .createScanIntent()
     }
--- a/collect_app/src/main/res/layout/first_launch_layout.xml
+++ b/collect_app/src/main/res/layout/first_launch_layout.xml
@@ -32,7 +32,7 @@
 
         <TextView
             android:id="@+id/tagline"
-            style="@style/TextAppearance.Collect.Headline4"
+            style="?textAppearanceHeadline4"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:lines="2"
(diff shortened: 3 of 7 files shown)


collect_app/src/main/res/layout/first_launch_layout.xml
```diff
@@ -32,7 +32,7 @@
 
         <TextView
             android:id="@+id/tagline"
-            style="@style/TextAppearance.Collect.Headline4"
+            style="?textAppearanceHeadline4"
```
seadowg, line 35: These can just use theme attributes rather than explicitly relying on styles.

collect_app/src/main/java/org/odk/collect/android/utilities/CodeCaptureManagerFactory.kt
```diff
@@ -27,6 +27,7 @@ object CodeCaptureManagerFactory {
         return IntentIntegrator(activity)
             .setDesiredBarcodeFormats(supportedFormats)
             .setPrompt(prompt)
+            .setOrientationLocked(false) // Let UI control orientation lock
```
seadowg, line 30: I don't think we should be doing this as it locks the orientation for the activity even after whatever is using it closes (to see this hit "Configure with QR code" and then press back and rotate).

We can definitely add smarter locks in at the fragment/activity level if it feels like we need it.

collect_app/src/main/java/org/odk/collect/android/activities/FirstLaunchActivity.kt
```diff
@@ -28,6 +28,8 @@ class FirstLaunchActivity : CollectAbstractActivity() {
 
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        setTheme(R.style.Theme_Collect_Light)
```
seadowg, line 31: This initially fixed the issue (no crash when using a non-translucent theme), but I ended up also disabling the orientation lock as well. It feels like this activity should use one of our themes though to prevent any other issues so don't see a reason not to include this change.

lognaturel:
There were some advantages to the orientation lock but overall it seems more consistent to rotate everywhere (and especially not to force orientation lock forever after certain sequences of actions). I think we merge this and do a beta asap so there's a chance for feedback in case we're missing a case.

lognaturel (approved):

PR #4655, opened 2021-06-30
Prevent editing of submissions that failed to send

### LINKED ISSUE #4589: Prevent editing of submissions that failed to send

dimwight:
Closes #4589

Revives (with improvements) #4629 closed accidentally

Tested with new `FormResubmissionTest`.

See design notes as updated below.

Small change to behaviour from user perspective. 
No obvious regression risks. 

No form required. 

https://github.com/getodk/docs/issues/1367

-----

#### _`CursorLoaderFactory`_
The single-argument version of `createSavedInstancesCursorLoader` adds `STATUS_SUBMISSION_FAILED` to the filter along with `STATUS_SUBMITTED`.

The two-argument version remains unchanged as it's not obvious how to test it - maybe leave for now?

Preceding methods have been reordered to match the subsequent ones. 
	
#### _`FormResubmissionTest`_
Closely derived from `SendFinalizedFormTest`.
Setting flag `noHttpPostResult` causes a series of calls to `Thread.sleep()`, simulating a connection drop-out.

`whenFailedForm…Resubmission` simulates the server behaviour and tests the updated application response to this. The method
- depends on the `StubOpenRosaServer` detailed below first to set `STATUS_SUBMISSION_FAILED` and then to accept resubmission.
- needs manual intervention as I haven't been able to work out how to simulate clicking on a _CANCEL_ dialog. 
- calls the slightly tweaked `clickViewSentForm` in `MainMenuPage`.

Server response is simulated in the override of `uploadSubmissionAndFiles` in the anonymous `StubOpenRosaServer` passed to the `TestDependencies` in the `RuleChain`.

This necessitated constructors for `TestDependencies` that set the final `StubOpenRosaServer` member.


lognaturel:
Thanks! I think some of my inline comments on the prior PR got lost so I'm going to go ahead and push some updates addressing them to this branch.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/EditSavedFormTest.java
```diff
@@ -0,0 +1,81 @@
...
+                .clickOnForm(TEXT_FORM_NAME)
+                .clickSendSelected()
+                .clickOK(new SendFinalizedFormPage())
+                .pressBack(new MainMenuPage())
+
+                .assertNumberOfEditableForms(0)
+                .clickEditSavedForm()
+                .assertTextDoesNotExist(TEXT_FORM_NAME)
+
+                .clickMenuFilter()
```
lognaturel, line 48: This block tests the `createEditableInstancesCursorLoader` with a `charSequence`.

lognaturel:
@dimwight thanks again for getting this started and do let me know if you have any questions about what I added on. There ended up being three areas that needed to be changed and tested:
- excluding failed submissions from the edit saved form menu without search
- excluding failed submissions from the edit saved form menu with search
- excluding failed submissions from the edit saved form count of editable forms

dimwight:
Fascinating! So much I hadn't even considered in my ignorance, hopefully my input was of some use.

lognaturel:
Yes, thank you! There are a lot of moving pieces and I apologize for underestimating the context required for this one.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/EditSavedFormTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.TestDependencies;
+import org.odk.collect.android.support.TestRuleChain;
+import org.odk.collect.android.support.pages.MainMenuPage;
+import org.odk.collect.android.support.pages.SendFinalizedFormPage;
+
+@RunWith(AndroidJUnit4.class)
+public class EditSavedFormTest {
+
+    public static final String TEXT_FORM_NAME = "One Question";
```
seadowg: (Very) Minor nitpick here, but I actually find moving test data into constants like this makes the test harder to read. I think in general, the more of the test that is colocated, the easier it can be comprehended as you don't have to jump anywhere else to see what the input data is. 

Another disadvantage in mind is that it can enforce a consistency that isn't really required. What if I need to use a different form in one of these tests? Do I change all of them?

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/EditSavedFormTest.java
seadowg: Could we add a comment to the two blocks to highlight that these could probably be replaced at a lower level when we're moving this data fetching here away from `CursorLoader`? My hope is that whoever is doing that will be running these tests and a comment will make them more likely to transfer that coverage.

seadowg: A reply to myself: an alternative would be to move these assertions to a separate test (maybe in a new `EditSavedFormTest`) that checks both cases and also has a comment about it probably being a test that could probably live further down the stack.

seadowg (approved):
Some little nitpicks around tests (of course) but this looks good and I think it's ready for QA!

kkrawczyk123:
**Tested with success!**

Verified cases:
- Cancel sending very fast
- Cancel sending when big media attached to form
- sending failed when the server's url invalid
- sending filed when server not working
- Closed-form on Central 
- fill submissions after failed sending
- send submission when first sending try failed
- Aggregate, Central, Google sheet
- missing submission url, invalid submission url
- turn off the internet connection while sending

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/EditSavedFormTest.java
lognaturel: I went with comments for now. I initially wanted to separate them but these tests take so long that I just can't bring myself to do it for something like this! I agree it should be lower. I do think it's important to include since the initial implementation had a bug around the search behavior.

seadowg (approved):

Issue #4247, opened 2020-11-29
Should home screen widget form shortcut should link to latest version or pinned version?


florianm:
## Software and hardware versions 
Collect v1.29.0 beta

#### Problem description
To help our enumerators remember which form to fill in their survey workflow, we use the ODK 1x1 widget to create shortcuts on the devices' home screen in a particular order.

Collect allows to remotely maintain devices through settings "exactly match server" and "hide old form versions". 
Currently, the widgets link to the latest form versions at the time of creation, and still do so even if the form has been replaced with newer versions.

This requires us to manually update each widget on each device whenever forms get updated.

#### Steps to reproduce the problem
* Publish a form through ODK Central. The form has a title, an ID, and an optional version.
* Create an ODK 1x1 widget on a device's home screen. The widget is named after the form title.
* Set ODK Collect to "exactly match server" and "hide old form versions".
* Publish an updated form version on  ODK Central. The updated form must have the same ID, an updated version string, and possibly an updated title.
* The widget remains named after the original form title and still opens the old form version.

#### Expected behavior
There's probably a good reason behind pinning form versions rather than form IDs in the ODK 1x1 widget.
However, for the use case "fully remote form updates", the widgets should link to the latest respective version (eg. using form ID) of each form instead of a pinned form version.

lognaturel:
I haven’t confirmed yet but I’m guessing that the shortcut is based on a database ID that represents a specific formid/version combination. We may not have control over that. 

Is this as necessary with forms exactly matched from the server? Would a shortcut to the form list screen help? Can you name your forms with a numeric prefix to identify order? (E.g “1 - intake”, “2 - first encounter”, etc) The form list screen can be sorted by name. 

We do want to eventually make form order server, configurable but that is not on the immediate roadmap.

florianm:
Based on the error messages I get when opening a shortcut to a freshly removed form, it's indeed some kind of integer ID linking a specific form.

We lay out our shortcuts on the home screen in a flowchart like pattern to bridge the mental gap between "what did I encounter now" and "what form to fill". We have a survey start and a survey end form that we show as first and last form on the screen (otherwise enumerators will inevitably forget to fill them out). Our forms are filled in an ever changing order (we have one for each entity type we encounter), and different campaigns use different set of forms, so numbering forms (and hiding some versions) is the analogue of releasing three fully greased up pigs numbered 1, 2, and 4 in a busy marketplace.

[...]

PR #4657, opened 2021-06-30
Use current project for shortcuts UI

### LINKED ISSUE #4640: Fix shortcuts for projects
## Notes

* The user sees the current project's forms when a setting a shortcut up.
* Shortcuts should work regardless of the current project.
* We should have analytics on this.

seadowg:
Closes #4640 
~~Blocked by #4462~~

I initially thought I'd have to do more work to get this working properly, as the shortcuts weren't respecting the "Hide old versions" setting. Looking at the old implementation ([here](https://github.com/getodk/collect/blob/d539ef61b151ed66307cb40351dd40802c5afe36/collect_app/src/main/java/org/odk/collect/android/activities/AndroidShortcuts.java#L69)) in 1.30.1 however, it looks like we always returned all forms rather than just the newest versions. Noticing that, I also realized that shortcuts will not work with form updates, which seems like a problem for their use with things like Match Exactly. I didn't try to fix either of these problems here, but it's probably something to keep in mind for the future.

EDIT: It looks like the above points are also discussed in [this issue](https://github.com/getodk/collect/issues/4247).

I chose to use the `BlankFormsListViewModel` here, as I can imagine we'll eventually end up using it for the "Fill Blank Form" (or whatever might replace it) and "Delete Blank Form" pages in future. I think we'll be able to use and expand on the new `getForms` and `BlankForm` class there as well.

This can go in without QA I think as it'll be covered in a regression pass.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.java
@@ -2,22 +2,15 @@ package org.odk.collect.android.feature.external;
 
 import android.content.Intent;
 
-import androidx.test.core.app.ActivityScenario;
-import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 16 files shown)


lognaturel (approved):
Signing off so I'll let someone else merge and watch the build!

Issue #4464, opened 2021-03-05
Reset application does not remove maps cache


yanokwa:
#### Software and hardware versions 
Collect v1.30.1

#### Problem description
Resetting Collect does not reset the maps cache

#### Steps to reproduce the problem
1. Uninstall Collect.
2. Reinstall Collect.
3. Get blank form and download all widgets form
4. Fill blank form using the map view. Zoom in on your area so tiles cache.
5. Admin Settings > Reset application > Select map layers and OSM tile cache
6. Put phone in airplane mode.
7. Fill blank form using the map view. Zoom in on your area and you'll see the tiles
8. Look in `/sdcard/Android/Data/org.odk.collect.android/cache` and there will still be files there.

#### Expected behavior
Reset removes `/sdcard/Android/Data/org.odk.collect.android/cache` (or its contents).

lognaturel:
> removes /sdcard/Android/Data/org.odk.collect.android/cache (or its contents)

Are you absolutely sure you saw map tiles in that folder? It's the default Android cache folder and we don't use it. I can't reproduce any tiles ending up there and @grzesiek2010 says at #4659 that they're stored in the default location which is `/sdcard/Android/data/org.odk.collect.android/files/osmdroid/tiles`.

I think what happened is that Collect used to set a custom tile cache path to be able to control that but when we reworked the geo functionality a couple years ago that probably went away. Given that no one has asked for this and that I can't even remember why we wanted to clear the map tile cache, I think @grzesiek2010's suggestion of removing that functionality is a good one.

grzesiek2010:
> @grzesiek2010 says at #4659 that they're stored in the default location which is /sdcard/Android/data/org.odk.collect.android/files/osmdroid/tiles

I didn't say that I saw anything there I just said that this is the value returned if by: `Configuration.getInstance().getOsmdroidTileCache()` (which we tried to clear during resetting)

but whether it's really used... I don't know for sure.

lognaturel:
I evaluated `Configuration.getInstance().getOsmdroidTileCache()` and got `/sdcard/osmdroid/tiles` which makes a lot of sense and does contain tiles. We can't touch that folder and shouldn't anyway.

It may be that we could set our own cache folder but I don't think it's very important and would bring up a number of annoying design issues. @yanokwa agrees with the decision we made to remove this resetting.

More info about osmdroid tile caching at https://osmdroid.github.io/osmdroid/Tile-Caching.html

yanokwa:
[...]

PR #4659, opened 2021-06-30
User can reset a project

### LINKED ISSUE #4425: User can reset a project
~~Blocked by #4417~~

**Acceptance**

When I click "Admin settings"
And I click "Project"
And I click "Reset project"
And I confirm
Then the project settings return to the Collect defaults

grzesiek2010:
Closes #4425

I tested resetting projects manually and improved automated tests.

This is the only solution.

No.

No.


grzesiek2010:
I came across two problems related to resetting projects:
**The first one (more serious):** we keep credentials in one static hashmap:
```
private static final Map<String, HttpCredentialsInterface> HOST_CREDENTIALS = new HashMap<>();
```

that means if we clear them in preferences we do that for all projects.
What should we do? Keep them separately for every projects and then clear only those that were added for the current project?

**The second issue (less important):** The last option is clearing OSM tile cache. The problem is that it's not our folder and we are not responsible for it. That means it's a shared folder for all projects and if we clear it in one projects everything is removed.
I wonder if we need to that step at all... maybe we could just remove it?

@lognaturel what do you think?

lognaturel:
> that means if we clear them in preferences we do that for all projects.

What are the actual implications of this? e.g. does it affect background jobs? I would expect the map to be cleared when switching projects.

If it does make sense to keep those credentials in memory for some reason, perhaps `webCredentialsUtils.clearCredentials(url)` could be used to only clear the current project's?

> it's not our folder

I don't really remember how this works. Do we specify a root that the cache folder gets put in? Or does the library just pick the folder for us? It looks like this behavior is already broken, probably since using scoped storage -- https://github.com/getodk/collect/issues/4464. It would be good to explore whether we have options like having each project set its own cache folder but if not, I think removing the option is ok.

grzesiek2010:
> What are the actual implications of this? e.g. does it affect background jobs? I would expect the map to be cleared when switching projects.

Oh I think I misunderstood how we use that static hashmap it's rather for temporary credentials and even if we clear it we should be able to take username/password from settings:

```
    public @Nullable HttpCredentialsInterface getCredentials(@NonNull URI url) {
        String host = url.getHost();
        String serverPrefsUrl = getServerUrlFromPreferences();
        String prefsServerHost = (serverPrefsUrl == null) ? null : Uri.parse(serverPrefsUrl).getHost();

        // URL host is the same as the host in preferences
        if (prefsServerHost != null && prefsServerHost.equalsIgnoreCase(host)) {
            // Use the temporary credentials if they exist, otherwise use the credentials saved to preferences
            if (HOST_CREDENTIALS.containsKey(host)) {
                return HOST_CREDENTIALS.get(host);
            } else {
                return new HttpCredentials(getUserNameFromPreferences(), getPasswordFromPreferences());
            }
        } else {
            return HOST_CREDENTIALS.get(host);
        }
    }
```

but here we have another issue. We read `getUserNameFromPreferences()` and `getPasswordFromPreferences()` from the current project but if it's a background job that was created by an inactive project then it won't work.
It's not a problem when it comes to Central because there we don't use username/password but it might be a problem for aggregate or other servers like that.

@seadowg did you test background jobs with credentials or just with ODK Central?

grzesiek2010:
> What are the actual implications of this? e.g. does it affect background jobs? I would expect the map to be cleared when switching projects.

Oh I think I misunderstood how we use that static hashmap it's rather for temporary credentials and even if we clear it we should be able to take username/password from settings. I tested it and it works well.

> I don't really remember how this works. Do we specify a root that the cache folder gets put in? Or does the library just pick the folder for us? It looks like this behavior is already broken, probably since using scoped storage -- #4464. It would be good to explore whether we have options like having each project set its own cache folder but if not, I think removing the option is ok.

when I debug it shows that the path is:
`/data/user/0/org.odk.collect.android/files/osmdroid/tiles`

We don't create that folder and the only thing we can do is setting a different location:
`DefaultConfigurationProvider.setOsmdroidTileCache(File osmdroidTileCache)`
but we would need to do that for every project, I mean after switching projects we would need to set that default osmdroid tile cache dir path.

I can create a separate issue for investigating this but for now I would just remove that option from resetting because I don't think it's important.

Do you agree?

lognaturel:
>  I tested it and it works well.

🎉

> for now I would just remove that option from resetting

That sounds good to me. There was some reason why we added clearing that cache but @yanokwa would have to remind us. Maybe it had to do with space concerns and those are now less significant with modern devices.

lognaturel (approved):

lognaturel:
I meant to review this yesterday and apparently got confused. Sorry about that! 😭

PR #4660, opened 2021-07-01
Add Google Drive project from manual config dialog

### LINKED ISSUE #4575: Add Google account support to project setup
~~Blocked by #4428~~
Mockup: https://www.figma.com/file/nDsMAYQDIflyl7rznEhwnN/ODK-Collect-Multiple-Project-Support-(v2021.2)?node-id=37%3A219

## Acceptance

- [x] Given I've navigated to the manual project setup screen
When I click "Configure" next to "My project uses Google Drive
And I choose my account
Then my project is added as my google account name (email)
And the project icon is the first letter of my account name
And the in the project list it shows "`<google account name>` / Google Drive" under the project name

- [x] Given I've navigated to the manual project setup screen
And I have a Google account on the phone with forms set up
When I click "Configure" next to "My project uses Google Drive
And I choose my account
And I navigate to Get Blank Forms
Then I can download the forms I have in my account

- [x] Given I've navigated to the manual project setup screen
When I click "Configure" next to "My project uses Google Drive
And I click cancel
Then I return to the manual setup screen

- [x] Fix font size for "My project uses Google Drive" text

lognaturel:
Closes #4575

Added an end-to-end test which passes. Tried manually.

Behavior-wise, it's consistent with configuring a GDrive project from settings and is the simplest approach I could come up with.

I've made a small change to use whatever the URL or Google account is as the project name if it can't be parsed as a URL. Previously if the value couldn't be parsed as a URL there was a special fallback of "Project" which I think was confusing.

I switched out `onActivityResult` for the AndroidX Activity Result APIs. Maybe it's my Bad Attitude(tm) but I'm not impressed. The thing that I find most problematic about `onActivityResult` is that it feels disconnected from the activity launch. I don't feel like the new APIs makes much of an improvement because the  `registerForActivityResult` call has to happen before `Fragment` creation (any attempt to put it elsewhere leads to exception with `Fragments must call registerForActivityResult() before they are created`). I also had a terrible time naming the `ActivityResultLauncher` and if anyone has a suggestion that's better than `googleAccountResultLauncher` I'd love to hear it. The docs call it `getContent` which I find particularly terrible (seems like a getter).

Side note: I had a terrible time getting all the dependencies updated and resolving (Android Studio had caching issues).

This is an additive change that's isolated so the worst case is that Google Drive configuration doesn't actually work. There's also some risk associated with 4722ece7 if I used the wrong key.

No.

I don't think so. We don't really talk about Google Drive integration in docs currently since it's not recommended.


grzesiek2010:
> Testing-wise, I'd like to have a test that verifies project creation similar to "Server project creation should be triggered after clicking on the 'Add' button". I don't currently have any great ideas for doing this other than with an Espresso integration test because of the account picker dialog. Is there a way to build some kind of mock intent that always immediately returns its result?

As I understand you want to add that test in `ManualProjectCreatorDialogTest` which already uses espresso so you should be able to stub intents as well https://developer.android.com/training/testing/espresso/intents?authuser=1

The existing `FakeGoogleAccountPicker` also might be helpful it depends on what you want to achieve.

am I missing something?

lognaturel:
> which already uses espresso

I think you meant that it uses the Espresso matchers so could also use further intent matchers but you got me into an existential crisis over at [this Slack thread](https://getodk.slack.com/archives/C35ENHA64/p1625156729134200)!

My challenge ended up being more about what style of test(s) to write. I would have liked to write lower-level tests but I don't really see anything meaningful to do there. I had started added a `Fragment`/Robolectric test to verify that the account selection dialog was being launched. But then testing the project creation step at that level felt strange. It would either have ended up as a single end-to-end test or 2-3 disjoint tests that didn't test the full flow. I was looking at stubbing the intent at the start of another test but the separation of the different tests didn't feel good (I also hadn't realized I could use the Espresso intent APIs). I've gone with an end-to-end test and I think it generally makes sense to run those on a real device even though that makes them a lot more time-consuming to run. More about this high-level question of balance between types of test in the Slack thread.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
```diff
@@ -12,21 +12,35 @@ import org.odk.collect.projects.Project
 @RunWith(AndroidJUnit4::class)
 class ProjectDetailsCreatorTest {
 
-    val projectDetailsCreator = ProjectDetailsCreator(ApplicationProvider.getApplicationContext())
+    private val projectDetailsCreator = ProjectDetailsCreator(ApplicationProvider.getApplicationContext())
```
lognaturel, line 15: This test wouldn't need Robolectric if the context didn't have to be passed in here, right?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ManualProjectCreatorDialogPage.kt
```diff
@@ -26,8 +34,18 @@ class ManualProjectCreatorDialogPage : Page<ManualProjectCreatorDialogPage>() {
         return this
     }
 
+    fun setGoogleAccount(googleAccount: String): MainMenuPage {
```
seadowg: It confused me a little what this was doing (stubbing vs an action etc). Could we call this something like `openGooglePickerAndSelect` or something to make it clear the gdrive button is getting hit?

collect_app/src/androidTest/java/org/odk/collect/android/support/TestDependencies.java
```diff
@@ -29,6 +30,8 @@
     public final FakeGoogleApi googleApi = new FakeGoogleApi();
     public final FakeGoogleAccountPicker googleAccountPicker = new FakeGoogleAccountPicker();
     public final StoragePathProvider storagePathProvider = new StoragePathProvider();
+    public final StubBarcodeViewDecoder stubBarcodeViewDecoder = new StubBarcodeViewDecoder();
```
seadowg, line 33: Yeah good shout. I guess we're never going to want a real one in test really.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ManualProjectCreatorDialogPage.kt
```diff
@@ -26,8 +34,18 @@ class ManualProjectCreatorDialogPage : Page<ManualProjectCreatorDialogPage>() {
...
+        val activityResult = Instrumentation.ActivityResult(Activity.RESULT_OK, data)
+        intending(hasAction("PICK_GOOGLE_ACCOUNT")).respondWith(activityResult)
+
+        onView(withText(R.string.gdrive_configure)).perform(scrollTo(), click())
+        return MainMenuPage().assertOnPage()
+    }
+
     fun addProject(): MainMenuPage {
-        Espresso.onView(withId(R.id.add_button)).perform(ViewActions.click())
+        onView(withId(R.id.add_button)).perform(click())
```
seadowg: I know this was already an ID but can we switch to `clickOnText` or `clickOnContentDescription` here?

collect_app/src/main/java/org/odk/collect/android/projects/ManualProjectCreatorDialog.kt
```diff
@@ -34,8 +39,28 @@ class ManualProjectCreatorDialog : MaterialFullScreenDialogFragment() {
...
+    @Inject
+    lateinit var permissionsProvider: PermissionsProvider
+
+    @Inject
+    lateinit var googleAccountsManager: GoogleAccountsManager
+
     private lateinit var binding: ManualProjectCreatorDialogLayoutBinding
 
+    val googleAccountResultLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result: ActivityResult ->
+        if (result.resultCode == Activity.RESULT_OK && result.data != null && result.data!!.extras != null) {
```
seadowg: If you assign `result.data` to a `val` here or use [`let`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/let.html) you won't need all the `!!`s.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
seadowg: Yeah, exactly. We could rework it so that the colours get passed in here rather than grabbed from `Context` which would remove the need for Robolectric. I'll let you make a call on whether you want to do that or not 😀.

config/lint.xml
```diff
@@ -57,4 +57,6 @@
     <!-- This check crashes if request interval is passed as a param :S -->
     <issue id="InvalidPeriodicWorkRequestInterval" severity="ignore" />
     <issue id="AlwaysShowAction" severity="ignore" />
+
+    <issue id="InvalidFragmentVersionForActivityResult" severity="ignore" />
```
seadowg, line 61: What happens without this?

collect_app/src/main/java/org/odk/collect/android/configure/qr/AppConfigurationGenerator.kt
```diff
@@ -25,6 +27,20 @@ class AppConfigurationGenerator(
         }.toString()
     }
 
+    fun getAppConfigurationAsJsonWithGoogleDriveDetails(googleAccount: String?): String {
+        val generalSettings = JSONObject().apply {
+            put(GeneralKeys.KEY_PROTOCOL, Collect.getInstance().getString(R.string.protocol_google_sheets))
```
grzesiek2010: I don't like using resources here and `Collect.getInstance()`. I also ran into a similar problems when I had to use a value from `untranslated.xml`. Wouldn't it be better to move those untranslated strings to a java/kotlin class so that we don't have to read them from resources?

collect_app/src/main/java/org/odk/collect/android/projects/ProjectListItemView.kt
```diff
@@ -25,18 +26,27 @@ class ProjectListItemView(context: Context, attrs: AttributeSet?) : FrameLayout(
     }
 
     private fun getSubtext(generalSettings: Settings): String {
-        val username = generalSettings.getString(GeneralKeys.KEY_USERNAME) ?: ""
-        var url = generalSettings.getString(GeneralKeys.KEY_SERVER_URL) ?: ""
+        val username = if (generalSettings.getString(GeneralKeys.KEY_PROTOCOL).equals(Collect.getInstance().getString(R.string.protocol_google_sheets))) {
```
grzesiek2010: I think ProjectListItemViewTest should be updated because this new case is not tested.

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
```diff
@@ -50,10 +52,17 @@ class SettingsImporter(
             } else {
                 JSONObject()
             }
+
+            val connectionIdentifier = if (generalSettings.getString(GeneralKeys.KEY_PROTOCOL).equals(Collect.getInstance().getString(R.string.protocol_google_sheets))) {
```
grzesiek2010: We have `projectDetailsShouldBeImportedIfIncludedInJson` test but now as we have the second case maybe it would be good to add a test for it as well?

collect_app/src/main/java/org/odk/collect/android/configure/qr/AppConfigurationGenerator.kt
```diff
@@ -25,6 +27,20 @@ class AppConfigurationGenerator(
         }.toString()
     }
 
+    fun getAppConfigurationAsJsonWithGoogleDriveDetails(googleAccount: String?): String {
```
grzesiek2010, line 28: There is a test for `getAppConfigurationAsJsonWithServerDetails` which is above but not for this method.

grzesiek2010 (changes requested):

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
grzesiek2010: Right.

grzesiek2010:
I'm also not sure about the visibility of that option. On my emulator it's hidden.
Screenshot_1625223014

Did you maybe consider other options? Maybe a google icon in toolbar or something?
image

collect_app/src/main/java/org/odk/collect/android/projects/ProjectListItemView.kt
```diff
@@ -25,18 +26,27 @@ class ProjectListItemView(context: Context, attrs: AttributeSet?) : FrameLayout(
...
+            generalSettings.getString(GeneralKeys.KEY_SELECTED_GOOGLE_ACCOUNT) ?: ""
```
seadowg, line 29: It actually looks like both these `:? ""` cases are untested to me. Could we add that to `ProjectLIstItemViewTest`?

seadowg (changes requested):

config/lint.xml
lognaturel: Lint check fails. See https://github.com/getodk/collect/pull/4660/commits/809b312d5ca9a43bb6d51275dd915c16597e02c3, looks like a false positive to me. Nothing would work otherwise...

collect_app/src/main/java/org/odk/collect/android/configure/qr/AppConfigurationGenerator.kt
lognaturel: Feels a little out of scope but it's bothered me before so I introduced the values in `GeneralKeys`. This follows the pattern other values have. It does mean the corresponding array (arrays.xml) has a value that's separately defined but I think that's ok.

Would be good to carefully review 4722ece to make sure I didn't use the wrong value anywhere.

collect_app/src/main/java/org/odk/collect/android/configure/SettingsImporter.kt
```diff
@@ -98,11 +98,17 @@ class SettingsImporter(
     }
 
     private fun clearUnknownKeys(preferences: Settings, defaults: Map<String, Any>) {
+        val toRemove = ArrayList<String>()
```
lognaturel, line 101: Can't remove things from a collection that's being iterated over.

lognaturel:
> I'm also not sure about the visibility of that option. On my emulator it's hidden.

We don't want to encourage using Google Drive so it's intentionally out of the way and not very discoverable. It is below the keyboard on my primary test device. I think that someone who really wants to find it will see the scrollbar and take that as a cue to find out what's lower on the screen.

seadowg (approved):

grzesiek2010 (approved):

PR #4664, opened 2021-07-01
Don't rotate images for display, instead preserve and use EXIF tag

### LINKED ISSUE #4571: Incorrect rotation or EXIF orientation of portrait photos captured with collect app
ODK Collect v1.30.1
Android 9
Xiaomi Mi A1

#### Problem description

Photos captured in portrait format with ODK Collect are not saved or rendred in the right direction. 

Portrait format with ODK Collect:
Portrait_ODK_Collect

Portrait format with camera app:
Portrait_Camera app

Landscape format with ODK Collect:
Landscape_ODK_Collect

Landscape format with camera app:
Landscape_Camera app

1. Create a simple form containing a photo uploader
2. Using ODK Collect, take a photo on a device that produces an EXIF Orientation tag for portrait photos and upload it via the form
3. Download the photo from KoBoToolbox or access the file directly in the folder : "Internal/Android/data/org.odk.collect.android/files/instances/<instance directory>/" 
4. Open on any system that understands EXIF Orientation and notice that landscape photos render correctly, but portrait photos are sideways
5. Compare with the same photo taken directly with the camera app of the device where portrait photos are displayed vertical as expected

#### Other information 

This issue was initially posted on : 
https://github.com/kobotoolbox/kobocat/issues/707

abhishekab:
- Remove rotation methods from ImageConvertor
  (Don't apply explicit rotation if EXIF rotation present)
- Copy image after applying EXIF rotation for Annotate Widget
- Remove ImageConvertor rotate tests
- Add FileUtils rotate tests

Closes #4571 

Manual Testing done. And unit tests added
There were two main approaches to solve this:
- Remove the explicit bitmap transformation which rotates the image, and preserve EXIF orientation. And make the ImageView displaying the image in ImageWidget sensitive to EXIF just for presentation.

- Let the explicit bitmap transformation as it is and let it rotate the image, and at the same time remove the orientation from EXIF before saving the transformed image.
Got inclined toward the former one, as using EXIF as it's supposed to be used and not rewriting the image seemed more apt.

N/A
N/A
N/A


seadowg:
@abhishekab sorry for the delay in a review. We're putting together a Collect release at the moment, but after that someone should be able to take a look!

abhishekab:
No worries :)

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
```diff
@@ -69,6 +77,7 @@
     private final MediaUtils mediaUtils;
     protected final String tmpImageFilePath;
 
+
```
grzesiek2010: Please remove this doubled empty line.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -174,6 +180,51 @@ public static Bitmap getBitmapScaledToDisplay(File file, int screenHeight, int s
         return bitmap;
     }
 
+
+    /**
+     * While copying the file, apply the exif rotation of sourceFile to destinationFile
+     * so that sourceFile with EXIF has same orientation as destinationFile without EXIF
+     */
+    public static void copyImageAndApplyExifRotation(File sourceFile, File destFile) {
```
grzesiek2010: `FileUtils` is growing maybe it would be better to have `ImageFileUtils` for utility methods that operate only on images and then `GeneralFileUtils`? We don't need to add it here but maybe in a separate pr.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/ImageConverterTest.java
```diff
@@ -419,7 +338,7 @@ public void scaleImageToNewWidthTest() {
     }
 
     @Test
-    public void keepExifTest1AfterScaleAndRotation() {
+    public void keepExifTest1AfterScaling() {
```
grzesiek2010: A better name would be just `keepExifAfterScaling`

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
```diff
@@ -123,45 +117,6 @@ public void whenReadOnlyOverrideOptionIsUsed_shouldAllClickableElementsBeDisable
         assertThat(getSpyWidget().annotateButton.getVisibility(), is(View.GONE));
     }
 
-    @Test
-    public void whenPromptHasDefaultAnswer_showsInImageView() throws Exception {
```
grzesiek2010, line 127: Why did you remove these tests?

grzesiek2010 (changes requested):

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
abhishek-wiai: So these tests were relevant when the answer images were being loaded synchronously via code, now that we are using Glide to load the image into imageView, I am not sure if we need them.

abhishekab: So these tests were relevant when the answer images were being loaded synchronously via code, now that we are using Glide to load the image into imageView, I am not sure if we need them.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/ImageConverterTest.java
abhishekab: Ah ok, I just reused the naming convention how it was. Will fix this

collect_app/src/main/java/org/odk/collect/android/widgets/BaseImageWidget.java
abhishekab: Will do.

collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
abhishekab: Agreed. Will create a separate issue followed by a PR after this one.

seadowg:
@abhishekab are you still planning to work on this?

abhishekab:
yes, yes. Sorry was a little occupied will update the PR in couple of days

seadowg:
@abhishekab no worries! Thanks so much for looking at it.

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
grzesiek2010: Those tests were added to verify that adding [default answers](https://docs.getodk.org/form-logic/#static-defaults) (even like in this case default media answers works well) and the fact that we use Glide or not has nothing to do here. So I think we should keep them.

grzesiek2010 (changes requested):
Some relevant tests were removed.

grzesiek2010:
@abhishekab what do you think about those tests? As I said I am pretty sure they are still needed.

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
abhishekab: Sorry for my limited understanding, but isn't the getFile() function in BaseImageWidget responsible for returning the file path for default answer, do we need to test this explicitly in an integrated manner to check that the default image is loaded into the image view in this scenario?

abhishekab: If yes, may be if you could guide me if I should explore [this](https://lcdsmao.dev/synchronize-glide-loading-in-test-with-idlingresource/) to synchronize loading to ensure the test passes, or if you can suggest any other methodology

grzesiek2010: > but isn't the getFile() function in BaseImageWidget responsible for returning the file path for default answer

yes but we need test to make sure the image is really displayed.

> If yes, may be if you could guide me if I should explore this to synchronize loading to ensure the test passes, or if you can suggest any other methodology

ahh so that's the issue... maybe you could try to implement abstraction something like described here https://stackoverflow.com/a/46286760
the goal would be to use Glide in production code but in tests load images like before.

seadowg:
Going to close this for the moment but feel free to reopen if you get a chance to come back to it @abhishekab!

grzesiek2010:
@abhishekab could you reopen the pull request but keep the fix only and remove the part with switching to Glide what caused problems with our tests? If you don't have time let us know.

collect_app/src/test/java/org/odk/collect/android/widgets/AnnotateWidgetTest.java
grzesiek2010: I reverted that change with using Glide for now and added the tests back.

grzesiek2010:
@seadowg please take a look.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/ImageConverterTest.java
```diff
@@ -461,6 +380,18 @@ public void keepExifTest1AfterScaleAndRotation() {
         }
     }
 
+    @Test
+    public void verifyNoRotationAppliedForExifRotation() {
+        Map<String, String> attributes = new HashMap<>();
+        attributes.put(ExifInterface.TAG_ORIENTATION, String.valueOf(ExifInterface.ORIENTATION_ROTATE_90));
+        saveTestBitmap(3000, 4000, attributes);
+        ImageConverter.execute(testImagePath, getTestImageWidget(), Collect.getInstance(), IMAGE_SIZE_ORIGINAL);
```
seadowg: We can use `ApplicationProvier.getApplication()` here instead of `Collect.getInstance()`

collect_app/src/test/java/org/odk/collect/android/utilities/CopyImageAndApplyExifRotationTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Test
+    fun copyAndRotateImageOneEighty() {
+        attributes[ExifInterface.TAG_ORIENTATION] = ExifInterface.ORIENTATION_ROTATE_180.toString()
+        saveTestBitmapToFile(sourceFile.absolutePath, 3000, 4000, attributes)
+        FileUtils.copyImageAndApplyExifRotation(sourceFile, destinationFile)
+        val image = FileUtils.getBitmap(destinationFile.absolutePath, BitmapFactory.Options())
+
+        assertEquals(3000, image.width)
```
seadowg: This test still passes if I change the `ORIENTATION_ROTATE_180` branch in the implementation to perform a rotation of `0` - 180 is annoying as the image dimensions end up the same. We technically have a similar problem for 90 and 270 as they could be flipped and the tests would still pass. 

Are we able to verify these by creating a Bitmap with a single filled pixel and checking where it ends up? I can imagine that might not be worth it wanted to discuss. Maybe there is a better way that you can think of.

collect_app/src/test/java/org/odk/collect/android/utilities/CopyImageAndApplyExifRotationTest.kt
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import timber.log.Timber
+import java.io.File
+import java.io.IOException
+
+@RunWith(AndroidJUnit4::class)
+class CopyImageAndApplyExifRotationTest {
```
seadowg: I think we should have these tests in `FileUtilsTest` or in a file prefixed with `FileUtils`. Otherwise, tools like IntelliJ's "switch to test" won't be able to find the test from the implementation and vice versa.

seadowg (changes requested):

seadowg:
Marking this as "needs testing" as all requested changes are in tests.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/utilities/ImageConverterTest.java
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/utilities/CopyImageAndApplyExifRotationTest.kt
grzesiek2010: I agree and I improved that test. It required moving it to `androidTest` because when you create/open bitmaps in unit test it returns empty bitmaps (the size is correct but its empty/transparent).

collect_app/src/test/java/org/odk/collect/android/utilities/CopyImageAndApplyExifRotationTest.kt
grzesiek2010: I factored out `ImageUtils` class because `FileUtils` was big already, then I rename that class to `ImageUtilsTest` and merged with `BitmapScaledToDisplayTest` which also tested only methods from the same class.

seadowg (approved):

grzesiek2010:
@kkrawczyk123 I was able to find a camera app whcih you can use to reproduce the issue: https://play.google.com/store/apps/details?id=com.jeyluta.timestampcamerafree&gl=PL

Unfortunately there is one problem... images are saved correctly in the storage but they are displayed incorrectly (rotated) in Collect. It would be confusing for users that use app like the one mentioned above (that provides orientation attribute in exif) so I think we need to improve the solution. Seems like using glide is a must here.

grzesiek2010:
Ok I fixed that problem so it's ready for testing. I need to disable a few tests that were broken because of Glide and I will try to bring them back after merging https://github.com/getodk/collect/pull/4957 or even as a part of that pr since that pr introduces a new module fro loading images and doing that should be easier then.

seadowg:
@grzesiek2010 what's motivating the change to Glide here? This probably need re-reviewed right?

grzesiek2010:
> what's motivating the change to Glide here

Glide handles orientations in EXIF correctly without this images displayed in Collect might be rotated, saved correctly but rotated in our app.

> what's motivating the change to Glide here

It's not a big change but if you want yeah but just 2 last commits.

kkrawczyk123:
**Tested with success!**
Verified on Androids 8.1 and 5.1

@grzesiek2010 thanks for sharing this app, it helped me with reproducing the issue!

Verified cases:
- take a picture, verify image attached to an instance
- choose image, verify image attached to an instance
- check orientation using Photo exif app
- image, annotate, selfie widgets
- display images in collect
- add big image
- add small image 
- add image from internal storage
- add image from Google Drive

seadowg (approved):

srujner:
**Tested with success!**

grzesiek2010 (approved):

PR #4665, opened 2021-07-02
Require protocol on manually-entered URLs and always log analytics on URL change

lognaturel:
I noticed two things: configuring a URL without a protocol results in a crash with a very unhelpful message on form download and we aren't getting analytics about all URLs configured. This PR addresses both and does a little bit of renaming for clarity.

No tests for analytics. I set a breakpoint on the analytics call and tried the various ways a URL can be configured. It would be good to have tests to validate that the handler is called from various places but now doesn't seem like the time.

I added a fragment test for the URL "validation". With some restructuring, I think this could be tested at a lower level but I think this test fits well with how the manual project creation code is currently tested.

I like that this is identical to the behavior for manually entering a URL in settings. I considered doing the same URL validation for QR codes but I couldn't immediately see how to plumb through a meaningful message for users. I also think that because of the way QR codes are constructed, it's a lot less likely they'll contain URLs without protocols.

I also considered entirely removing the requirement for a protocol prefix. We could fall back to https in that case. That felt like a riskier change that I'd like to think about some more. Similarly, I also considered changing "Sorry, invalid URL!" to be a more useful message but have decided to leave it for now. I think we should get this in to avoid the cryptic crash and follow up with a more robust improvement like making a preemptive request if connectivity is available or something like that.

I did remove some code that I didn't understand (see inline) so there might be some risk there.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsUtils.java
@@ -6,8 +6,10 @@ import org.odk.collect.shared.Settings;
 import org.odk.collect.shared.strings.Md5;
 
 import java.io.ByteArrayInputStream;
+import java.util.Locale;
 
 import static java.lang.String.format;
+import static org.odk.collect.android.analytics.AnalyticsEvents.SET_SERVER;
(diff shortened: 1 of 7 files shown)


grzesiek2010 (approved):

grzesiek2010:
@lognaturel you have some conflicts.

PR #4670, opened 2021-07-02
Protect against issues that may come up when removing a project while a background job is running

### LINKED ISSUE #4654: Protect against issues that may come up when removing a project while a background job is running
## Acceptance

- [x] Given I've set up match exactly for my project
And the background job is currently running
When I navigate to project settings
And click "Delete Project"
Then I see a warning that background jobs are running and to try again in a moment

- [x] Given I've set up previously downloaded updates for my project
And the background job is currently running
When I navigate to project settings
And click "Delete Project"
Then I see a warning that background jobs are running and to try again in a moment

## Notes

We're already protected from project deletion and auto send interacting weirdly as we don't allow deleting projects with unsent instances.

The simplest way to implement this is to probably try to acquire the forms `ChangeLock` as part of deletion (rather than trying to detect the background job running).

grzesiek2010:
Closes #4654

I extended the existed `ProjectDeleter` which is the best option because this class is already responsible for checking other states that should block deleting projects like unsent instances. I think there is no other solution here.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
@@ -2,6 +2,7 @@ package org.odk.collect.android.projects
 
 import org.odk.collect.android.backgroundwork.FormUpdateScheduler
 import org.odk.collect.android.backgroundwork.InstanceSubmitScheduler
+import org.odk.collect.android.utilities.ChangeLockProvider
 import org.odk.collect.forms.instances.Instance
 import org.odk.collect.forms.instances.InstancesRepository
 import org.odk.collect.projects.Project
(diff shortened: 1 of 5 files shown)


collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
```diff
@@ -14,39 +15,60 @@ class ProjectDeleter(
...
-                Instance.STATUS_SUBMISSION_FAILED
-            ).isNotEmpty()
-        ) {
-            return DeleteProjectResult.UnsentInstances
-        } else {
-            val currentProject = currentProjectProvider.getCurrentProject()
+        return when {
+            unsentInstancesDetected() -> DeleteProjectResult.UnsentInstances
+            runningBackgroundJobsDetected() -> DeleteProjectResult.RunningBackgroundJobs
+            else -> performProjectDeletion()
```
seadowg, line 25: Minor nitpick: this can just match `DeleteProjectResult.DeletedSuccessfully` instead. Then if you add a new subtype the `when` will fail to compile (which is a nice piece of protection).

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
```diff
@@ -105,14 +116,57 @@ class ProjectDeleterTest {
...
+            "",
+            changeLockProvider
+        )
+
+        val result = deleter.deleteCurrentProject()
+        assertThat(result, instanceOf(DeleteProjectResult.RunningBackgroundJobs::class.java))
+    }
+
+    @Test
+    fun `If there are running background jobs that use saved forms the project should not be deleted`() {
```
seadowg, line 162: I don't think we actually need to protect ourselves from this, but at the same time it does nicer code-wise to reference both change locks.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
```diff
@@ -105,14 +116,57 @@ class ProjectDeleterTest {
...
+            currentProjectProvider,
+            formUpdateManager,
+            instanceSubmitScheduler,
+            instancesRepository,
+            "",
+            changeLockProvider
+        )
+
+        val result = deleter.deleteCurrentProject()
+        assertThat(result, instanceOf(DeleteProjectResult.RunningBackgroundJobs::class.java))
```
seadowg, line 157: Both these tests should also assert that the project still exists - probably just checking the repository is enough.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
grzesiek2010: Done.

seadowg (approved):

Issue #4258, opened 2020-12-02
Answers in related lower-level questions in "fast external itemsets" with the minimal appearance are not cleared when the upper-level answer is removed or changed


mmarciniak90:
#### Software and hardware versions 
Collect with #4242 

#### Problem description
Answers in related lower-level questions in "fast external itemsets" with the minimal appearance are not cleared when the upper-level answer is removed or changed

#### Steps to reproduce the problem
[selectOneExternal.zip](https://github.com/getodk/collect/files/5630004/selectOneExternal.zip)

Use case 1:
1. Fill answers for a minimal appearance 
2. Change answer on first-level - state
3. Answers for County and City are not cleared

Use case 2:
1. Fill answers for a minimal appearance
2. Remove response from first-level - state
3. Answers for County and City are not cleared

![100866015-62ab5100-3498-11eb-9cea-2f0acaa837c8](https://user-images.githubusercontent.com/3907554/100885976-8cbe3c80-34b3-11eb-994f-b2821ef74549.gif)

#### Expected behavior
Answers for County and City are cleared

PR #4673, opened 2021-07-06
Resolves #4500 autocomplete minimal…

### LINKED ISSUE #4500: autocomplete minimal not working as expected with select_one_external
Collect v1.30.1, Android vx.x.x, device used Samsung Galaxy TAB 4

#### Problem description
Using cascading select with select_one_external with appearance autocomplete minimal
Once selected  the first level, I select the second and all works fine.
Going back to change the first level, the second is still selected, not reset.

I'm instead expecting the reset of the second level (filtered to the new first level)
#### Other information 
It works well with normal select_one but slower due to the large number of items in the form (see atached file)

[Entrevista_Nacimientos_Etnicos_filtro.xlsx](https://github.com/getodk/collect/files/6246387/Entrevista_Nacimientos_Etnicos_filtro.xlsx)

dimwight:
As delicately suggested by @lognaturel, here goes...

Resolves #4500 (I hope)

Tested with new `SelectOneResetTest`.

See design notes below.

Improvement to behaviour from user perspective. 
No obvious regression risks. 

_[selectOneReset.zip](https://github.com/getodk/collect/files/6767992/selectOneReset.zip)_

No documentation changes (behaviour now as expected)

-----
# Design notes
A lot of stuff here, but both [the issue](https://github.com/getodk/collect/issues/4500) and my proposed resolution have turned out to be quite complex. I thought more than once that my reach had exceeded my grasp!

## The issue
A shout out to @aurdipas for a [form](https://github.com/getodk/collect/files/6246387/Entrevista_Nacimientos_Etnicos_filtro.xlsx) that demonstrates clearly the undesired behaviour of selects with fast external itemsets. 
- Most obviously, members of cascades using fast external itemsets often don't reset when preceding members are updated, in constrast to their behaviour with internal itemsets. 

- This behaviour is most glaringly apparent with minimal appearances, especially in field lists - plain selects reset anyway when they can’t match an existing value in their current choices.

Desirable if less obvious behaviours also implied by the form are that

- Members should reset even when unrelated fields are interpolated in the cascade.
- Members hidden when non-relevant should be reset when made visible.

And all this should work inside groups and especially field lists.

Not only may the current behaviour confuse the user, it permits saving of incoherent data.

Partly for this reason it seems important that a resolution to this issue should focus on resetting the answers for a complete cascade, rather than just ensuring that widgets reset when rendered. 

#### _1_  Preparatory changes
`clearFollowingItemsetWidgets` in `SelectOneWidget`
 sort of resets succeding cascade members, but hitherto there’s been no equivalent in `SelectOneMinimalWidget`. 

So to start with I recycled `clearFollowingItemsetWidgets`:
-  Pulled it up to `ItemsWidget`
-  Called it from `SelectOneMinimalWidget` by analogy with `SelectOneWidget`

#### _2_  Basic improvement
`SelectOneWidgetUtils` now has `checkFastExternalCascade`, called (for now) from `clearFollowingItemsetWidgets`
- Detects fast external cascade members from their query strings
- Skips over interpolated questions - including unrelated selects as in the issue form
- Returns at non-question - assumes cascades aren't split between groups/repeats	

#### _3_  Field lists improvement
The basic fix aborts silently when the cascade is in a field list as in the issue form.
In `FormEntryActivity`, `updateFieldListQuestions` currently resets only the next cascade member (possibly as a side effect?). 

In this improvement it calls `checkFastExternalCascadeInFieldList`, also living in `SelectOneWidgetUtils` as the functionality is identical.

#### _4_  Relevance reset (on unhide)
For field lists, a tweak to `updateFieldListQuestions` itself resets any cascade member controlled by **relevant** when it's added to the view.  

This adapts existing code of some complexity - I don't see any danger of regression, but maybe the most recent maintainer should consider?

For standard views, reset on relevance ought to be logically impossible but happens somehow in the widget view, though not in the hierarchy. More puzzling still, behaviour before the update was correct but isn't now. 

## Testing
The issue appears resolved as outlined above when changes are tested manually against the issue form (when [slightly modified ](https://github.com/getodk/collect/files/6768015/Entrevista_Nacimientos_Etnicos_filtro%2B.zip) with no cascade splitting between groups).

For formal testing there are numerous use cases, representing combinations of:
- 2 types of itemset: internal (as a baseline) and fast external
- 4 appearances: plain, minimal, minimal autocomplete and just autocomplete (never used in practice?)
- 5 contexts: top-level; simple and nested groups; simple and nested field lists

### Test form
The [form](https://github.com/getodk/collect/files/6767992/selectOneReset.zip) derives from _[selectOneExternal.zip]( https://github.com/getodk/collect/files/5630004/selectOneExternal.zip)_ from #4258 - thanks @mmarciniak90. It covers
- four cascade levels - seems the maximum in practice
- interpolated fields before fourth levels - actually relevance selects as in the issue form
- all the use cases, arranged in a pragmatic order

The workbook includes these supporting sheets:

- The _variants_ sheet defines section numbers in terms of the two itemset types and the two essentially different appearances.
- The _asserts_ sheet records results of manual testing both before and after the update.
- The _master_ sheet defines five functionally identical context blocks:
  - _A_ - top level
  - _B_ - simple group
  - _C_ - nested group
  - _D_ - field list
  - _E_ - nested field list
- It includes two alternative **type** and four **appearance** columns, and prefills **defaults** ready for testing of reset behaviour.
- These are propagated in the _survey_ sheet into sections testing the various combinations of appearances and itemsets.
- For this draft PR the form just implements the four core variants:
  - 0 and 1 - internal itemsets with plain and minimal appearances
  - 2 and 3 - fast external itemsets

### Test class
`SelectOneResetTest` borrows extensively from `FieldListUpdateTest`. In each block it makes changes and asserts expected outcomes as set out in the workbook _asserts_ sheet. 

- The `SectionVariant` `enum` combines an `itemset` based on the **type** of its section with an `appearance` from the **appearance**.  

- The `Block` `enum` provides the appropriate strings for each `SectionVariant`.

The single test method iterates through the variants, thus visiting all sections of the prefilled form, and runs test code parameterized to the section for each variant.  

The code makes use of new convenience methods in
`FormEntryPage` that wrap useful formulations found in `FieldListUpdateTest`. 

`SelectOneResetTest` can be validated by running against the current codebase and checking that it fails (mostly) where expected. 

It's currently somewhat of a work in progress.


dimwight:
I guess it's unlikely anyone will have time and attention to spare in the immediate future to check the above out. But if I'm wrong, please hold off for now: there are major improvements coming up in both application and test code.

lognaturel:
Exciting! Was hoping to carve out time early this week but I'll wait for your signal. 👍

dimwight:
Now superseded by #4712

Issue #4675, opened 2021-07-07
Move location access to Foreground Service


seadowg:
This is an output of #4499.

There are several places in Collect where we access the user's location in "the background" (i.e. while the `Activity` is not on screen and there is no active Foreground Service). Android 11 does not allow this.

Location usages in Collect

- Audit events can have lat/long included. Collect gets location updates and passes this to the `AuditEventLogger` so the latest one can always be used in the next audit event. This happens in the "background", as it's not cleaned up in `onPause`.
- The `odk:setgeopoint` (only really used by `start-geopoint`) records a single location/point which gets written as an answer. To do this, it grabs location updates for a set amount of time and then uses the most accurate one. This could happen in the background if the user locks the screen after opening the form.
- Geoshape and Geotrace record points at an interval and this needs to work while the phone is locked (for walking around a field for instance)

What we need to do:

- [ ] Move any location clean-up in an `onStart`/`onStop` cycle to `onResume`/`onPause`
- [ ] Use a foreground service to manage location client for geoshape/geotrace

PR #4677, opened 2021-07-07
Move location tracking to foreground lifecycle

seadowg:
Work towards #4675

This makes it so that location tracking that we were currently performing with the "visible" lifecycle (start/stop) happens within the "foreground" lifecycle (resume/pause). This will make it so that we aren't breaching any of the new Play Store regulations, but will mean that geoshape and gotrace won't work while the phone is locked (or the app is in the background). That will be fixed in a follow-up that closes the referenced issue.

It also moves the location client code to a new package, which was a helpful exercise to understand its structure.

Ran tests.

We had talked about moving all location access to a foreground service, but it looks like that will be a lot of work. We'll do that to enable geoshape/trace to work while locked, but leave everything else as it is.

Other than the mentioned regression, there shouldn't really be any noticeable changes in behaviour here. It'd be good to play around with the geoshape and geotrace widgets (using different map providers), the audit log with location and `start-geopoint`. Geoshape and geotrace should still work as expected while the app is open (things like the map center moving with the user) but won't update while the app is in the background.


seadowg:
Marking this as "needs testing" before review, as it feels like the real reason to break the issue into two PRs is to QA this change.

collect_app/src/main/java/org/odk/collect/android/location/client/MaxAccuracyWithinTimeoutLocationClientWrapper.java
```diff
@@ -16,7 +18,7 @@
  *
  * New requests reset the timeout and the highest accuracy.
  */
-public class MaxAccuracyWithinTimeoutLocationClient implements LocationClient.LocationClientListener, LocationListener {
```
seadowg, line 19: This name really threw me off, so decided to make it clear this didn't implement `LocationClient`.

collect_app/src/test/java/org/odk/collect/android/location/client/MaxAccuracyWithinTimeoutLocationClientWrapperTest.java
```diff
@@ -4,33 +4,37 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.odk.collect.testshared.LocationTestUtils.createLocation;
 import static org.robolectric.shadows.ShadowSystemClock.advanceBy;
 
 @RunWith(AndroidJUnit4.class)
-public class MaxAccuracyWithinTimeoutLocationClientTest {
+public class MaxAccuracyWithinTimeoutLocationClientWrapperTest {
     private FakeLocationClient fakeLocationClient;
-    private TestLocationListener testLocationListener;
```
seadowg, line 24: This was already a "mock" test double, but it felt easier to just use mockito.

strings/src/main/res/values/untranslated.xml
```diff
@@ -7,7 +7,7 @@ OR CONDITIONS OF ANY KIND, either express or implied. See the License for
 the specific language governing permissions and limitations under the License. -->
 <resources>
     <!-- strings that should not be translated -->
-    <string name="app_name" translatable="false">ODK Collect</string>
```
seadowg, line 10: Another library also had an `app_name` and for some reason this was now overriding ours. This should prevent that happening in the future.

kkrawczyk123:
**Tested with success!**

Tested cases:
- regression checks on all geo widgets
- Google map, Mapbox, OSM
- offline layers
- forms with recording location background: audit, repeats, set location in xml
- display submissions on map
- automatic and manual modes

location/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    kotlinOptions {
+        jvmTarget = '1.8'
+    }
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'androidx.core:core-ktx:1.6.0'
```
lognaturel, line 48: I've started noticing versions drifting apart. Should we maybe always have a version constant for dependencies that show up in more than one module?

lognaturel (approved):

location/build.gradle
seadowg: This is an interesting one because although it doesn't feel quite right, I'm not sure if it really matters unless we see a conflict at compile time. Maybe we should add an investigation issue to look into dependency versions in multimodule setups?

lognaturel: Investigation issue works for me! I don't expect mismatched versions would generally cause a problem but there might be occasional subtle issues that drive us crazy.

I spun my wheels for a while when I upgraded the AndroidX fragment version. It turned out to need to be upgraded across three modules and the error states before I figured that out were really hard to understand.

seadowg: Done: https://github.com/getodk/collect/issues/4691

PR #4681, opened 2021-07-08
Add subtext to existing settings and use outline icons for all settings sections

### LINKED ISSUE #4666: Add subtext to existing settings and use outline icons for all settings sections
This is a first quick step towards [settings improvements](https://forum.getodk.org/t/collect-combine-general-and-admin-settings/34149).

The idea is to update the names of existing settings based on the mockups and add the subtext. This means translators can start on those strings.

Another small change that should be easy to group with the string update is to change existing General Settings section icons to all use the outline versions (some currently use the filled version). It may also be reasonable/helpful to add icons for the admin settings that are going to stay, I'm not sure.

grzesiek2010:
Closes #4666

I reviewed implemented changes and ran automated tests.

|   |  |
| Screenshot_1625743454  | Screenshot_1625781265  |

It doesn't need testing.

No.

No.


lognaturel:
Looks good! Would be great to get the icons and subtext for the three User access control subcategories as well.

This PR renames a lot of resources. If the text is changing anyway, that is a great opportunity to do the rename. But if the text isn't changing, the translations will still be lost because Transifex won't be able to detect the rename. Is this something you considered? That makes it seem not worth it to me, what do you think?

grzesiek2010:
> Would be great to get the icons and subtext for the three User access control subcategories as well.

You mean for:
- Main menu settings
- User settings
- form entry settings

if so what icons and subtext should I use because it's not visible in your mockups so I didn't change anything there.

> This PR renames a lot of resources. If the text is changing anyway, that is a great opportunity to do the rename. But if the text isn't changing, the translations will still be lost because Transifex won't be able to detect the rename. Is this something you considered? That makes it seem not worth it to me, what do you think?

Oh i didn't know it would be a problem for Transifex, in this case I agree that it doesn't make sense to rename some strings. I removed most of them.

lognaturel:
Thanks!

https://www.figma.com/file/qOwdGX7QgNOAzQ10TVxH6h/ODK-Collect-settings?node-id=6%3A2619 has the icons and text for those three. But I also don't mind merging this own and coming back to it!

grzesiek2010:
> https://www.figma.com/file/qOwdGX7QgNOAzQ10TVxH6h/ODK-Collect-settings?node-id=6%3A2619 has the icons and text for those three. But I also don't mind merging this own and coming back to it!

ahh I missed it thanks. Ok It's done.

lognaturel (approved):

PR #4685, opened 2021-07-12
Spike out location tracker service for geo trace/shape 

seadowg:
Work towards #4675 
~~Blocked by #4677~~

This adds a new experimental setting ("Use location tracker service") that switches our location tracking over to an Android Foreground Service for the geoshape/trace widgets. This is just a spike at the moment that should allow people to test this out in the field.

Verified manually.

This is just a spike, so going with the easiest thing to add at the moment.

Only changes behaviour when the experimental setting is enabled, so not a lot of risk here. No need for QA.


lognaturel:
This makes it look easy!

Behavior all looks as expected. It'd be nice if tapping on the location notification brought Collect to the foreground but not essential.

lognaturel (approved):
Merging now so we can tag a beta with this off `master`.

PR #4687, opened 2021-07-13
Provide UI for entering admin password when an admin password is set

### LINKED ISSUE #4683: Provide UI for entering admin password when an admin password is set
~Blocked by #4682~

- [ ] Given an admin password is set
When I go to project settings
Then I see a locked icon in the upper right side of the app bar
And all of the settings categories in the "Protected" section are disabled

- [ ] Given an admin password is set
When I tap on the locked icon in the upper right side of the app bar
Then I see a dialog prompting me to enter the admin password 
[see Figma](https://www.figma.com/file/qOwdGX7QgNOAzQ10TVxH6h/ODK-Collect-settings?node-id=29%3A687) (note, this could reuse the existing dialog as a first step. Let's only rebuild it with the new design if it's really easy)

- [ ] Given an admin password is set
When I go to project settings
And I tap on a disabled settings category in the "Protected" section
Then I see the same dialog prompting me to enter the admin password as when I tap on the lock

- [ ] Given I have been prompted for the project admin password
When I enter an incorrect password
Then I see a toast (same behavior as now)

- [ ] Given I have been prompted for the project admin password
When I enter a correct password
Then "Project settings" refreshes without any access controls (same behavior as Admin Settings > General Settings today)
And all categories in the "Protected" section are enabled
And the lock icon in the upper right side of the app bar is shown as open

- [ ] Given I have successfully unlocked admin settings
When I exit the settings screen and then go to it again
Then settings are locked again

- [ ] Given no admin password is set
When I go to project settings
Then I do not see anything in the upper right side of the app bar
And access controls are applied
And settings in the "Protected" section are enabled

### LINKED ISSUE #4695: Add protected setting to hide Project display category
- [ ] When I go to project settings
And I go to "Access control" > "User settings"
Then I see "Project display" below "Server" and above "Theme"

- [ ] When I uncheck "Project display" in "Access control" > "User settings"
And I go back to "Project settings"
Then I don't see the "Project display" category

### LINKED ISSUE #4699: Don't show active protected settings when admin password is set but has not been entered
As of https://github.com/getodk/collect/pull/4687, when there is an admin password set but it has not been entered yet, protected settings are all displayed and active. Clicking on them prompts the user to enter the admin password and then refreshes the settings, possibly changing what is displayed.

The original idea behind showing protected settings when admin password is set but has not been entered was to help guide users towards administrative functionality since the lock is fairly subtle. But the current flow feels pretty confusing. 

Some alternatives:
1. Hide the whole protected settings section and hope users eventually find the lock
2. Disable protected settings so that they're visible to give users a strong hint that there's some action to take on that screen. I'm not sure how practical this is but an idea would be to modify the section heading in that case to say "Protected (tap lock in upper right to access)"
3. Show a list item that when tapped shows the password dialog. [Figma](https://www.figma.com/file/qOwdGX7QgNOAzQ10TVxH6h/ODK-Collect-settings?node-id=82%3A140)

2 and 3 both seem fine to me. Perhaps the person who does the implementation can weigh level of effort and user experience to pick.

I don't think it's critical to get this in before regression testing.

grzesiek2010:
Closes #4683 
Closes #4695  
Closes #4699

I tested the fix manually and added/improved tests.

The biggest change I implemented and based my changes on are states.

Before if admin password was set we were asked to enter it before opening settings so there were only two states:
- settings in normal mode
- settings in admin mode

Now we are able to open settings without entering password (we ill be asked for it if we try to access protected settings).
This required implementing new states:
- LOCKED, // Admin password is set but not entered
- UNLOCKED, // Admin password is set and entered
- NOT_PROTECTED // Admin password is not set

All settings should be tested. I would be rather focused on visibility/accessibility because the functionality itself shouldn't be changed.

No.

No.


lognaturel:
Looking really good!

On Android 7.1, I'm not seeing any visual indication that the items in the protected section are disabled when an admin password is set and hasn't been entered. Maybe it's an issue specific to this Android version?

Now that I see it in action, I wonder whether not applying access controls when the admin password has been entered is too confusing. Doesn't have to be resolved in this PR but something to think about and consider.

I also think that at some point (doesn't even have to be this release) we should give some behavior to tapping on the open lock. It feels unexpected that it doesn't do anything. Maybe it could show a confirmation dialog to lock settings and apply access controls or something like that.

When there's no admin password set, the access controls are applied which is great. It looks like when access control is changed to hide something, it refreshes immediately. But when access control is changed to show something, it does not.
- Access control > User settings > Uncheck server
- Go back to Project settings, Server isn't there 👍
- Access control > User settings > Check server
- Go back to Project settings, Server still isn't there
- Rotate the screen, Server appears

grzesiek2010:
> On Android 7.1, I'm not seeing any visual indication that the items in the protected section are disabled when an admin password is set and hasn't been entered. Maybe it's an issue specific to this Android version?

No it's not not implemented. I can't make those prefs disabled because then they are not clickable so I would just need to set the colors somehow programmatically but does that make sense. Isn't the padlock icon enough? 

> When there's no admin password set, the access controls are applied which is great. It looks like when access control is changed to hide something, it refreshes immediately. But when access control is changed to show something, it does not.

Now it should be fixed.

I still want to add more automated tests and I will do that tomorrow, but if you think the implementation is good enough we can ask Kasia to test it.

lognaturel:
> I can't make those prefs disabled because then they are not clickable so I would just need to set the colors somehow programmatically but does that make sense. Isn't the padlock icon enough?

Ah, I see. It feels a lot weirder to me than I expected. That is, if I have a password set, when I tap on "Set Admin Password", I don't expect to be prompted for the existing password because there's no indication that it will do anything other than let me set the password. It's also very disorienting that the screen refreshes and scrolls to the top but doesn't seem to have done the action I expected.

Maybe a simple answer is to hide the whole protected section if there's an admin password and it hasn't been entered? I'm not sure why I didn't just do that in the first place... 😭 

> ask Kasia to test it.

Yes, that would be good. I still do want to do another pass looking at a couple quick implementation things but I'll mark as `needs testing`. We can decide whether the changes described above should go in this PR or a follow-on.

lognaturel:
> I'm not sure why I didn't just do that in the first place... 😭

Going back to the Figma file reminded me that I was worried that the lock wouldn't be intuitive to folks who already have an admin password set and are upgrading. They might be looking for admin settings and feel really lost so showing some familiar settings but disabled seemed like a way to guide upgrading users towards what they're looking for. 

I'm not sure, though, what do you think? Maybe the lock is clear enough? Would it be straightforward to add a tip like [this](https://www.figma.com/file/qOwdGX7QgNOAzQ10TVxH6h/ODK-Collect-settings?node-id=29%3A611)?

grzesiek2010:
> Maybe the lock is clear enough?

I guess it should be enough so i would rather just hide those protected prefs.

The second scenario would require creating our custom preference to display a message like that, it's doable but I'm not sure if it makes sense.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/ProjectPreferencesFragment.kt
```diff
@@ -39,90 +45,167 @@ class ProjectPreferencesFragment :
...
+        setHasOptionsMenu(true)
+
+        val changeAdminPasswordViewModel = ViewModelProvider(requireActivity()).get(
+            ChangeAdminPasswordViewModel::class.java
+        )
+        changeAdminPasswordViewModel.passwordSet.observe(
+            this,
+            { isPasswordSet: Boolean ->
+                if (isPasswordSet) {
+                    projectPreferencesViewModel.setStateUnlocked()
```
seadowg: I think we can potentially simplify this. Right now, we have two ViewModels interacting via observations from Fragments. 

Instead, could we share the `ProjectPreferencesViewModel` between the `ProjectPreferencesFragment` and the `AdminPasswordDialogFragment`. I'm imagining the ViewModel could have an `attemptPassword` method that checks the password and then sets the locked/unlocked state based on that. The locked state could then also be a `LiveData` exposed by the `ProjectPreferencesViewModel` and the fragment could decide how it renders based on an observation of that. 

For showing Toasts etc I think exposing a separate `LiveData` for "errors" or "failed attempts" could work. This could use a `LiveData<Consumable>` instead of a `SingleEventLiveData`. I prefer this as it allows observers to make their own decision about what to do with the "event" rather than changing the semantics of `LiveData`. 

Does that make sense @grzesiek2010 @lognaturel?

grzesiek2010: Like this https://github.com/getodk/collect/pull/4687/commits/a2ba0a754b27b5bcaf58cf8ed78ab48478612dab?
Yeah it makes the code a bit cleaner. I forgot about `LiveData<Consumable>` I can use it but I'm still not sure if I like it more than that `SingleEventLiveData`...

kkrawczyk123:
**Tested with success!**

Verified cases:
- light mode/dark mode
- screen rotation
- minimizing app
- set admin password
- edit admin password
- access settings when admin password set
- create a project through qr code when admin password set 
- create a project through qr code when settings hidden
- reset settings when admin password selected 
- incorrect admin password toast
- show and hide settings when locked/unlocked
- set admin password on few projects
- hide/unhide project display
- create project without admin password set but with hidden settings
- hide all settings options


@lognaturel I have noticed that now when general settings and admin settings are gathered together I am able to jump to admin settings and edit them in the middle of filling form. I haven't noticed anything especially weird as those changes are not updated immediately and I have to leave form (ex. turn on and off moving backwards) but I want to make sure that you are aware of that.

lognaturel:
I like the latest change! Great to see that QA didn't reveal anything serious. I'd like to see this merged before we make any final behavior updates.

> I am able to jump to admin settings and edit them in the middle of filling form

Ah yes, thanks for highlighting this, @kkrawczyk123. It's not really ideal but since you're not finding any problems with it, I think we go with it for now.

grzesiek2010:
@lognaturel so what about those protected prefs? Can I just hide them?

lognaturel:
> what about those protected prefs?

Let's leave for now and keep thinking about the best approach. I'll file a separate issue. It will be isolated enough that I think it's fine to add after/during regression testing.

lognaturel:
I've filed #4699. Like I've said, fine to do after you've merged. Or if one of the two ideas I am leaning towards sounds good to you and is straightforward to implement, you could do it here. Up to you!

grzesiek2010:
I've just added a fix for #4699 to this pr (https://github.com/getodk/collect/pull/4687/commits/747ad31a2409c8b701cb38950735d21bdd2b7f04) and also added a nice improvement I think (showing/hiding prefs instead of removing them and adding again https://github.com/getodk/collect/pull/4687/commits/baf2f05febdb34d798ee96ac197a022a6df7cbc1) what makes the experience much better @kkrawczyk123 also noticed that the refreshing was just ugly.

@kkrawczyk123 please take a look at it when you are free.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/ProjectPreferencesFragment.kt
seadowg: Yup, liking the new structure a lot more! I'm wondering if we could get rid of the `Consumable` entirely, but I think there is some awkwardness with the prefs lifecycle and `LiveData` that we'd need to get right, so probably a tweak for another day!

seadowg (approved):

kkrawczyk123:
@grzesiek2010 I have noticed weird behavior when followed scenario:
1. Open form
2. Jump to Settings
3. Change ex. font size
4. Try to unlock protected settings
After input admin password they do not get unlocked, see gif:
ezgif com-gif-maker (6)

When followed the same scenario but without any edit on the settings side, it works well.

lognaturel:
And this does NOT happen when launching settings from the main menu?

I think we should merge, start regression testing (Monday I guess), and explore in parallel.

lognaturel:
I checked and it's not related to launching settings from form entry, it happens regardless. So if you ~change non-protected settings~ navigate to any other settings screen and then unlock, the UI doesn't reflect the unlock and there doesn't seem to be any way to make it do so. Maybe a listener is getting lost or something?

That's serious enough that it should be fixed before merge.

grzesiek2010:
I'm going to investigate it

grzesiek2010:
This small change fixes the issue https://github.com/getodk/collect/pull/4687/commits/3e2fac0dc77d612fe83b26ab048f503bce213c29
Don't ask my why because I'm still wondering what was wrong...
I debugged the project and everything was called properly so @lognaturel 

> Maybe a listener is getting lost or something?

it wasn't the case.

The same method I moved down in `onViewCreated` (`updatePreferencesVisibility()`) is called either way for example in the described case after entering password so why does that change in `onViewCreated` do the job???

lognaturel:
Wowww... that is... interesting. Going to look at it a little bit more to see if it might be hiding some other weird behavior. Thanks for getting a fix in so fast!

lognaturel:
I can't really explain why order mattered in that case. ~But I think it would be preferable to entirely remove the `onViewCreated` implementation. The effect it has now is that on view creation `updatePreferencesVisibility` is called twice. There will always be a state from the viewmodel that will not have been consumed once the view is created so you can rely on the observation to do the correct refresh.~ No no, that won't work on rotation. It /is/ weird that the refresh happens twice when the view is first created but let's go with it for now.

@kkrawczyk123 please verify that set of steps again on master and file an issue if you still see anything unusual!

PR #4694, opened 2021-07-14
Add Location Tracker service

### LINKED ISSUE #4675: Move location access to Foreground Service

seadowg:
Closes #4675. This follows up #4685 with a proper implementation driven out with tests.

This version has tests and I also verified the geotrace behaviour manually.

The main new thing to consider here is the `AppState` concept I introduced. I've added docs there (in `AppState.kt`) that hopefully describe my thinking on why I wanted it and what it's for. I've also gone ahead and used it in `audiorecorder` as well, so it's not a disparate pattern only used in one place.

Only a couple of behavioural tweaks between this and the spike version, so not much to test here. I think the reviewer should have a play to see if they'd like to tweak anything before moving on, but no need for QA as these widgets/flows will go through regression testing very soon!

Any form with geotrace/geoshape.


androidshared/src/main/java/org/odk/collect/androidshared/data/AppState.kt
```diff
@@ -0,0 +1,53 @@
+package org.odk.collect.androidshared.data
+
+import android.app.Activity
+import android.app.Application
+import android.app.Service
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.ViewModel
+
+/**
+ * Can [AppState] can be used as a shared store of state that lives at an "app"/"in-memory" level
```
lognaturel: Extra "Can"

This seems nice and clear, thanks. 👍

testshared/src/main/java/org/odk/collect/testshared/ServiceScenario.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.fragment.app.testing.FragmentScenario
+import androidx.lifecycle.Lifecycle
+import androidx.test.core.app.ActivityScenario
+import org.robolectric.Robolectric
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.android.controller.ServiceController
+import java.util.function.Consumer
+
+/**
+ * Mimics [ActivityScenario]/[FragmentScenario] to provide a way to test [Service] in a similar
```
lognaturel, line 16: Today I learned about KDoc.

Bummer that this isn't part of the framework.

lognaturel (approved):

lognaturel:
We talked about `AppState` and it seems reasonable to me. Up to you whether you want to fix the small typo or not before merging!

lognaturel:
This is a very small thing but ideally tapping the geotrace/shape automatic capture pause button would stop location updates. It looks like currently the service is only stopped when leaving the activity. Doesn't really matter but the notification is a little misleading since the tracked location isn't being accessed at that point.

lognaturel:
I turned on automatic geotrace capture every 20s with min accuracy of 20m, switched to some other apps, locked the screen and then drove to a dr appointment for ~20mins. Pixel 3a, Android 11, no SIM. 

I got 39 points that do accurately represent my route. There were not a lot of points captured while I was on the highway and I think that makes sense — no WiFi and Android was probably limiting how much GPS access there was because of the screen lock. 

On the way back, I did the same collection but with the screen unlocked. I got 71 points that way.

@getodk/testers I don’t think this feature area will need much extra verification beyond that.

PR #4697, opened 2021-07-15
Prompt for confirmation when duplicate project configured

lognaturel:
When multiple projects with identical connection settings are configured, it's possible to get in bad states like having some submissions in one duplicate project that never get sent. This PR checks for matching connection settings and prompts the user to either switch to the existing project or actually create a duplicate.

Note: the QR code scanner scans continuously. This leads to weird behavior when scanning an invalid code or a duplicate project -- the same code is scanned over and over again. This is existing behavior and seems ok for these relatively rare cases.

Google Drive is currently untested.

Added and ran automated tests.

I am not very satisfied with the control flow but couldn't come up with anything I liked better. I wanted to use some kind of view model but I don't see a reasonable way to communicate UI actions back to the configuration dialogs. There's also more redundancy than I'd like but it seems pragmatic for now.

Everything around project creation is at risk. I think the worst outcome would be that users are prompted for confirmation to create non-duplicate projects.

No.

Yes. https://github.com/getodk/docs/issues/1365

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
+++ b/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
@@ -93,6 +93,7 @@ import org.odk.collect.android.projects.ProjectCreator;
 import org.odk.collect.android.projects.ProjectDeleter;
 import org.odk.collect.android.projects.ProjectDetailsCreator;
 import org.odk.collect.android.projects.ProjectImporter;
+import org.odk.collect.android.projects.SettingsConnectionMatcher;
 import org.odk.collect.android.storage.StorageInitializer;
 import org.odk.collect.android.storage.StoragePathProvider;
 import org.odk.collect.android.storage.StorageSubdirectory;
@@ -602,5 +603,10 @@ public class AppDependencyModule {
     @Provides
     public ProjectResetter providesProjectResetter(StoragePathProvider storagePathProvider, PropertyManager propertyManager, SettingsProvider settingsProvider, InstancesRepositoryProvider instancesRepositoryProvider, FormsRepositoryProvider formsRepositoryProvider) {
         return new ProjectResetter(storagePathProvider, propertyManager, settingsProvider, instancesRepositoryProvider, formsRepositoryProvider);
+    }
+
+    @Provides
+    public SettingsConnectionMatcher providesSettingsConnectionMatcher(ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
+        return new SettingsConnectionMatcher(projectsRepository, settingsProvider);
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
+++ b/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
@@ -93,7 +93,6 @@ import org.odk.collect.android.projects.ProjectCreator;
 import org.odk.collect.android.projects.ProjectDeleter;
 import org.odk.collect.android.projects.ProjectDetailsCreator;
 import org.odk.collect.android.projects.ProjectImporter;
-import org.odk.collect.android.projects.SettingsConnectionMatcher;
 import org.odk.collect.android.storage.StorageInitializer;
 import org.odk.collect.android.storage.StoragePathProvider;
 import org.odk.collect.android.storage.StorageSubdirectory;
@@ -603,10 +602,5 @@ public class AppDependencyModule {
     @Provides
     public ProjectResetter providesProjectResetter(StoragePathProvider storagePathProvider, PropertyManager propertyManager, SettingsProvider settingsProvider, InstancesRepositoryProvider instancesRepositoryProvider, FormsRepositoryProvider formsRepositoryProvider) {
         return new ProjectResetter(storagePathProvider, propertyManager, settingsProvider, instancesRepositoryProvider, formsRepositoryProvider);
-    }
-
-    @Provides
-    public SettingsConnectionMatcher providesSettingsConnectionMatcher(ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
-        return new SettingsConnectionMatcher(projectsRepository, settingsProvider);
     }
 }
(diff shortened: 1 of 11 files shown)


collect_app/src/test/java/org/odk/collect/android/projects/SettingsConnectionMatcherTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.odk.collect.projects.Project
+
+class SettingsConnectionMatcherTest {
+    private val inMemProjectsRepository = InMemProjectsRepository()
+    private val inMemSettingsProvider = InMemSettingsProvider()
+    private val settingsConnectionMatcher = SettingsConnectionMatcher(inMemProjectsRepository, inMemSettingsProvider)
+
+    @Before
+    fun setUp() {
+        inMemProjectsRepository.deleteAll()
```
seadowg: Every test gets a new object so you don't these these clean ups.

collect_app/src/test/java/org/odk/collect/android/projects/SettingsConnectionMatcherTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    fun `returns null when no projects exist`() {
+        val jsonSettings = getServerSettingsJson("https://demo.getodk.org")
+
+        assertThat(settingsConnectionMatcher.getProjectWithMatchingConnection(jsonSettings), `is`(nullValue()))
+    }
+
+    @Test
+    fun `returns a match when urls match`() {
```
seadowg: `returns matching project...`

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -603,4 +604,9 @@ public ProjectDeleter providesProjectDeleter(ProjectsRepository projectsReposito
     public ProjectResetter providesProjectResetter(StoragePathProvider storagePathProvider, PropertyManager propertyManager, SettingsProvider settingsProvider, InstancesRepositoryProvider instancesRepositoryProvider, FormsRepositoryProvider formsRepositoryProvider) {
         return new ProjectResetter(storagePathProvider, propertyManager, settingsProvider, instancesRepositoryProvider, formsRepositoryProvider);
     }
+
+    @Provides
+    public SettingsConnectionMatcher providesSettingsConnectionMatcher(ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
```
seadowg: Don't think you need this?

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -603,4 +604,9 @@ public ProjectDeleter providesProjectDeleter(ProjectsRepository projectsReposito
     public ProjectResetter providesProjectResetter(StoragePathProvider storagePathProvider, PropertyManager propertyManager, SettingsProvider settingsProvider, InstancesRepositoryProvider instancesRepositoryProvider, FormsRepositoryProvider formsRepositoryProvider) {
         return new ProjectResetter(storagePathProvider, propertyManager, settingsProvider, instancesRepositoryProvider, formsRepositoryProvider);
     }
+
+    @Provides
```
lognaturel: Don't need to inject because I ended up not including any Robolectric tests

collect_app/src/main/java/org/odk/collect/android/projects/DuplicateProjectConfirmationDialog.kt
```diff
@@ -0,0 +1,23 @@
...
+import android.content.DialogInterface
+import android.os.Bundle
+import androidx.fragment.app.DialogFragment
+import org.odk.collect.android.R
+
+class DuplicateProjectConfirmationDialog : DialogFragment() {
+    lateinit var listener: DialogInterface.OnClickListener
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        listener = parentFragment as DialogInterface.OnClickListener
```
seadowg: I think this would be nicer with a specific interface

collect_app/src/main/java/org/odk/collect/android/projects/DuplicateProjectConfirmationDialog.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import android.app.AlertDialog
+import android.app.Dialog
+import android.content.DialogInterface
+import android.os.Bundle
+import androidx.fragment.app.DialogFragment
+import org.odk.collect.android.R
+
+class DuplicateProjectConfirmationDialog : DialogFragment() {
+    lateinit var listener: DialogInterface.OnClickListener
```
lognaturel: Use a specific interface instead of this generic one

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
```diff
@@ -128,16 +137,24 @@ class QrCodeProjectCreatorDialog : MaterialFullScreenDialogFragment() {
                 }
 
                 try {
-                    handleScanningResult(barcodeResult)
+                    val settingsJson = CompressionUtils.decompress(barcodeResult.text)
+                    lastScannedJson = settingsJson
+
+                    settingsUniquenessChecker.getProjectWithMatchingConnection(settingsJson)?.let { uuid ->
+                        lastMatchingUuid = uuid
+                        DialogUtils.showIfNotShowing(DuplicateProjectConfirmationDialog::class.java, childFragmentManager)
```
seadowg: You could pass the `lastScannedJson` and `lastMatchingUuid` into the dialog here using as args. `showIfNotShowing` can take args in as a `Bundle`.

seadowg: And then pass those values back on clicks.

seadowg (changes requested):

seadowg (approved):

PR #4702, opened 2021-07-17
Don't allow shortcuts to launch forms outside the current project

### LINKED ISSUE #4668: Unable to save form filled through shortcut when other project selected on Collect
Collect v2021.2 Beta 2, Android v all device used...

#### Problem description
After creating form shortcut from project 1 I am able to fill and save the form. After changing project to another on Collect I am able to use shortcut, fill the form but I am unable to save the finalized form. Toast with error message displays:
Screenshot (Jul 2, 2021 1_21_44 PM)

1. Create form shortcut
2. Change project in Collect
3. Try to fill and save form through created shortcut

#### Expected behavior
Created shortcuts should work whatever project you have selected.

seadowg:
Closes #4668

This fixes the bug as described in my comment in the issue. I moved some files around - mainly creating an `external` package for shortcuts and the content providers, which meant changing the name of the old `external` package (which is really only for external data). This makes the changeset appear larger than it really is, so I'd advise going through commit by commit.

Verified manually and wrote new tests.

The big decision here was to add a "proxy" activity to handle form `VIEW` and `EDIT` actions. This keeps `FormEntryActivity` clear of the URI sanitizing and error handling for these cases. In the future, it might also allow us to simplify how forms load in `FormEntryActivity`. Imagine if it didn't need to care about content URIs eh?

I've also added timestamps to projects to allow them to be sorted. We don't technically need them now (as projects are added and returned in creation order) but it seems like a change we probably want to make before release. I've made it so the timestamp is not visible outside `ProjectsRepository` however, as this would have required more changes across the code base.

Should just fix the bug. It'd be good to check form loading (especially from other apps) but instead of specifically looking at that for this PR, that will take place in general regression for this release.


collect_app/src/main/AndroidManifest.xml
```diff
@@ -280,5 +279,19 @@ the specific language governing permissions and limitations under the License.
...
         </activity>
 
+        <activity
+            android:name=".external.FormUriActivity"
+            android:noHistory="true"
+            tools:ignore="AppLinkUrlError">
+
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <action android:name="android.intent.action.EDIT" />
```
seadowg, line 295: I'm not sure why we support `EDIT` for forms. It looks like we use `EDIT` as the action to open a form for entry from the "Fill Blank Form" page, but I don't think there is really any need for that - it could launch the Activity directly. Anyway, that's a battle for another day.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -630,8 +630,7 @@ private void loadFromIntent(Intent intent) {
 
             formPath = candidateForms.get(0).getFormFilePath();
         } else if (uriMimeType != null && uriMimeType.equals(FormsProviderAPI.CONTENT_ITEM_TYPE)) {
-            String projectId = uri.getQueryParameter("projectId");
```
seadowg, line 633: No need to care about anything other than the current project in here any longer. It'd be a lot of extra work to get `FormEntryActivity` to actually function properly with a form outside the current project, so I thought leaving this in felt inconsistent.

collect_app/src/main/AndroidManifest.xml
lognaturel: I think what you’re saying is that view makes no sense for a form — there’s no read-only view for a blank form. Also internally, there’s no point in using actions. 

Using the EDIT action for a form definition for external clients makes sense to me. I think in general we don’t want external clients using activities directly. Using actions is all that is documented https://docs.getodk.org/launch-collect-from-app/#using-a-uri-to-edit-a-form-or-instance)

collect_app/src/main/AndroidManifest.xml
```diff
@@ -174,11 +174,11 @@ the specific language governing permissions and limitations under the License.
          -->
 
         <provider
-            android:name=".provider.FormsProvider"
+            android:name=".external.FormsProvider"
```
lognaturel, line 180: Does the package change have implications for apps that currently use these providers? I think they’d just access them through the content URIs but I’m not confident.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -192,7 +192,6 @@ the specific language governing permissions and limitations under the License.
 
                 <category android:name="android.intent.category.DEFAULT" />
 
-                <data android:mimeType="vnd.android.cursor.item/vnd.odk.form" />
```
lognaturel, line 195: Does this result a breaking change in what external clients can do?

strings/src/main/res/values/strings.xml
```diff
@@ -1065,4 +1065,7 @@
     <!-- Preceded by "Don't have a project yet?" question -->
     <string name="try_demo">Try a demo</string>
     <string name="scan_configuration_qr_code">Scan configuration QR code</string>
+
+    <!-- Error shown when a user tries to open a form that's in a different project -->
```
lognaturel, line 1069: Expressing this is tricky! An alternative — “To open this shortcut, you must first open ODK Collect and switch to the project that contains this form.”

lognaturel (commented):
Forgive me if my questions don’t make total sense, some of these external integration paths are pretty confusing to me. 

I think this all makes sense and looks good for shortcuts. Did actually switching projects seem too risky? Realistically I don’t think the error message will be helpful to most users but it’s certainly better than allowing actions that result in uncertain state. 

I believe this change also addresses the issue for folks who are launching blank forms from external apps. But I think folks who launch edits of existing submissions from external apps will continue to have problems. I believe this is still possible by using the instance content provider to get URIs for instances in the first project. But if there is another project and Collect is switched to it, then the external app request for a submission edit will launch the FormEntryActivity but in the context of the wrong project. Does that sound right? In other words, I’m not following why the proxy activity was introduced just for blank forms. Like I put inline, it’s also not clear to me whether or not this introduction results in a breaking change.

collect_app/src/main/AndroidManifest.xml
seadowg: Ah nice, I had incorrectly assumed there wasn't docs for this. Yeah, I think ultimately I don't mind whether we use `EDIT` or `VIEW` to mean "fill a blank form" but it doesn't make sense to support both. Given the docs, I think I should change the tests to focus on the `EDIT` action rather than `VIEW`. I could also change shortcuts to use `EDIT`. What do you think?

collect_app/src/main/AndroidManifest.xml
seadowg: It shouldn't, as you can only access content providers via a URI as far as I'm aware.

collect_app/src/main/AndroidManifest.xml
seadowg: I don't think so. It just moves the target to `FormUriActivity`. Both worked with the `ViewFormActionTest`.

seadowg:
> Forgive me if my questions don’t make total sense, some of these external integration paths are pretty confusing to me.

You and me both!

> I think this all makes sense and looks good for shortcuts. Did actually switching projects seem too risky? Realistically I don’t think the error message will be helpful to most users but it’s certainly better than allowing actions that result in uncertain state.

Yeah it should be doable, but it did raise more questions in my head than just showing an error. I think if people complain, we can put the (slight) extra effort in.

strings/src/main/res/values/strings.xml
seadowg: This message could be shown from an external app, not just a shortcut. I'm struggling to think of better text if I'm being honest!

lognaturel: “To open this form, you must first open ODK Collect and switch to the project that contains it.”

collect_app/src/main/AndroidManifest.xml
lognaturel: It will definitely work with the actions. My question is whether integrations could previously have been using the class name. If so, this is breaking.

lognaturel: We’ve talked about this and decided to leave as-is. We now document the actions. The class names were never documented and we want to move more in the direction of them not being considered part of the API. 

For folks who upgrade to this and already have an integration directly with the activity, it will leave things in a bad state if the project isn’t first switched. They’ll get no indication of this.

lognaturel (approved):

PR #4707, opened 2021-07-20
Write an empty file with project name inside of each project folder

### LINKED ISSUE #4678: Write an empty file with project name inside of each project folder
Project folders on the filesystem are named according to the project UUID. To make it easier to navigate ODK's folders, we will add an empty file whose name matches the project name directly inside each project folder.

- [x] Given I've accessed the ODK projects folder (`/sdcard/Android/data/org.odk.collect.android/files/projects/`)
When I go into a specific project's folder
And I list all files
Then I see an empty file with the project's name

- [x] Given I've changed the project name either manually through settings or by scanning a new QR code within the project
When I go into that project's folder
And I list all files
Then I see an empty file with the project's new name

grzesiek2010:
Closes #4678 

I create a new file with project name if new project is created. If it's a demo project I don't do that for two reasons:
- the code responsible for creating demo projects is somewhere else so it would require more changes
- it doesn't make much sense because demo project has directory with demo name
but if a user changes the name in setting then such a file is created.

We need to verify that the file with project name is present after:
- creating new projects manually
- creating new projects with qr code
- setting project name in settings
- reconfiguring a project with qr code

No.

No.


collect_app/src/main/java/org/odk/collect/android/configure/qr/QRCodeScannerFragment.kt
```diff
@@ -30,19 +32,30 @@ class QRCodeScannerFragment : BarCodeScannerFragment() {
...
+        val oldProjectName = currentProjectProvider.getCurrentProject().name
+
         val importSuccess = settingsImporter.fromJSON(
             CompressionUtils.decompress(result.text),
             currentProjectProvider.getCurrentProject()
         )
         val settingsHash = getMd5Hash(ByteArrayInputStream(result.text.toByteArray()))
         if (importSuccess) {
+            val newProjectName = currentProjectProvider.getCurrentProject().name
+            if (newProjectName != oldProjectName) {
```
lognaturel, line 54: Is this going to compare string references? Should be .equals, no?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/ProjectDisplayPreferencesFragment.kt
```diff
@@ -125,11 +130,16 @@ class ProjectDisplayPreferencesFragment :
     override fun onPreferenceChange(preference: Preference, newValue: Any): Boolean {
         val (uuid, name, icon, color) = currentProjectProvider.getCurrentProject()
         when (preference.key) {
-            PROJECT_NAME_KEY -> projectsRepository.save(
-                Project.Saved(
-                    uuid, newValue.toString(), icon, color
+            PROJECT_NAME_KEY -> {
```
lognaturel, line 133: Could all this be done in `SettingsChangeHandler` instead so that the work only has to be done in one place for all settings paths?

grzesiek2010: No because we don't store project details in settings so `SettingsChangeHandler` is not called as well.

lognaturel: Ahhh, I see. That's really too bad. This feels dangerous because  it's the kind of thing where we could add additional paths. But let's go with it for now. At least it's not critical behavior.

grzesiek2010: Even if we stored those details in settings it would be a problem because in `SettingsChangeHandler` we have just access to the new value but I also need the old one to delete the existed file.

collect_app/src/main/java/org/odk/collect/android/configure/qr/QRCodeScannerFragment.kt
grzesiek2010: It's kotlin and in kotlin reference equality is `===` not `==`

lognaturel: > It's kotlin

Sigh. It's tricky switching back and forth!

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified cases:
* file is created for Central project scanned from QR code
* file is created for Aggregate project created manually
* file is created for Aggregate project scanned from QR code
* file is created for Google Drive project, manually and scanned QR code from another device
* file is changed when user reconfigured project with qr code
* file with long name
* file is changed when user updated project details in settings
* lowercase and uppercase name of the file
* special characters -> #4766 
* file is not created for DEMO project
* removed, renamed file manually

PR #4708, opened 2021-07-20
Fix external app integrations for projects

### LINKED ISSUE #4703: Don't allow opening instances outside the current project with `EDIT` action
Covers behaviour documented at: https://docs.getodk.org/launch-collect-from-app/#using-a-uri-to-edit-a-form-or-instance. We need to make changes because of multi project setups.

- [x] `EDIT` for an instance with a URI without project ID works if the currently select project is the first (chronologically created) project
- [x] `EDIT` for an instance with a URI without a project ID shows a warning and closes if the currently select project is **NOT** the first (chronologically created) project
- [x] `EDIT` for an instance with a URI with a project ID works if the project is the currently selected one
- [x] `EDIT` for an instance with a URI with a project ID shows a warning and closes if the project is **NOT** the currently selected one

### LINKED ISSUE #4704: `PICK` action should return form and instance URIs with project ID
- [x] `PICK` for forms should return a URI with the `projectId` query parameter
- [x] `PICK` for instances should return a URI with the `projectId` query parameter

seadowg:
Closes #4703
Closes #4704

This ended up being pretty easy as forms and instances can very simply share the implementation for `EDIT` actions and it turns out `PICK` was already working due to us removing the old `CONTENT_URI` constants from the instances and forms contracts.

Interestingly, it looks like we don't actually support `PICK` for instances. Instead of fixing this, I propose we just amend our docs.

New tests.

I'll leave comments inline on anything that stands out.

One thing to point out is there are two commits here dedicated to speeding up tests. I noticed that tests that added a project were running really slowly and found that, other than the disk write time, the slowness was caused by two things:

1. Waiting for the QR code scanner view to initialize
2. Repeatedly asking Android for the phone number while importing settings

It turns out that 1 can be fixed simply by making the view invisible in test (we stub out the decoder anyway) and that 2 could be solved by caching the phone number statically.

Again, we'll test this as part of general regression.

https://github.com/getodk/docs/issues/1375

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
@@ -1,5 +1,7 @@
 package org.odk.collect.android.support;
 
+import android.app.Activity;
+import android.app.Instrumentation;
 import android.content.Intent;
 
 import androidx.test.core.app.ActivityScenario;
@@ -15,6 +17,8 @@ import org.odk.collect.android.support.pages.FirstLaunchPage;
 import org.odk.collect.android.support.pages.MainMenuPage;
 import org.odk.collect.android.support.pages.Page;
 import org.odk.collect.android.support.pages.ShortcutsPage;
+
+import java.util.function.Consumer;
 
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.action.ViewActions.click;
@@ -73,5 +77,12 @@ public class CollectTestRule implements TestRule {
         intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
         ApplicationProvider.getApplicationContext().startActivity(intent);
         return destination.assertOnPage();
+    }
+
+    public <T extends Page<T>> Instrumentation.ActivityResult launchForResult(Intent intent, T destination, Consumer<T> actions) {
+        ActivityScenario<Activity> scenario = ActivityScenario.launch(intent);
+        destination.assertOnPage();
+        actions.accept(destination);
+        return scenario.getResult();
     }
 }
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
@@ -5,14 +5,13 @@ import android.app.Instrumentation;
 import android.content.Intent;
 
 import androidx.test.core.app.ActivityScenario;
-import androidx.test.core.app.ApplicationProvider;
 
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 import org.odk.collect.android.R;
-import org.odk.collect.android.external.AndroidShortcutsActivity;
 import org.odk.collect.android.activities.SplashScreenActivity;
+import org.odk.collect.android.external.AndroidShortcutsActivity;
 import org.odk.collect.android.support.pages.FirstLaunchPage;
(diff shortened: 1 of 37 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
```diff
@@ -67,11 +71,14 @@ public ShortcutsPage launchShortcuts() {
     }
 
     public <T extends Page<T>> T launch(Intent intent, T destination) {
-        /*
-        This can't use ActivityScenario.launch because of: https://github.com/android/android-test/issues/496
-         */
-        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-        ApplicationProvider.getApplicationContext().startActivity(intent);
+        ActivityScenario.launch(intent);
```
seadowg, line 73: Looks like they fixed this.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -188,15 +188,8 @@ the specific language governing permissions and limitations under the License.
...
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-                <action android:name="android.intent.action.EDIT" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-
-                <data android:mimeType="vnd.android.cursor.item/vnd.odk.instance" />
-            </intent-filter>
+            tools:ignore="AppLinkUrlError"
+            android:exported="true">
```
seadowg, line 192: We need `exported` here now that there isn't an intent filter.

lognaturel:
>  it looks like we don't actually support PICK for instances

I'm quite confident this is a regression -- systems that wrote instances to disk as a means of communication with Collect used this. Also, [showing the instance list](https://docs.getodk.org/launch-collect-from-app/#launching-the-form-list-or-instance-list-activity) is useless without picking. Did you verify that it doesn't work in v1.30.1? This would have been tried with ODK Collect Tester so it would surprise me.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -282,7 +275,7 @@ the specific language governing permissions and limitations under the License.
             </intent-filter>
         </activity>
 
-        <!-- Supports VIEW and EDIT actions for forms (documented at
+        <!-- Supports VIEW and EDIT actions for forms and instances (documented at
         https://docs.getodk.org/launch-collect-from-app/#using-a-uri-to-edit-a-form-or-instance).
         Do not use the activity directly as it's name/package might change. -->
```
lognaturel: `it's` to `its` while you're here? 🤓

seadowg:
@lognaturel I could definitely be missing something, but the `AndroidManifest.xml` for 1.30.1 doesn't declare `PICK` action for instances (file [here](https://github.com/getodk/collect/blob/d539ef61b151ed66307cb40351dd40802c5afe36/collect_app/src/main/AndroidManifest.xml)). The `VIEW` and `EDIT` actions are supported for instances (from `InstanceChooserList`) but not `PICK`.

seadowg:
@lognaturel it does look like the `InstanceChooserList` expects it (in the current version) though:

https://github.com/getodk/collect/blob/11ae4d29fa4c71c0777ffe20c61ccbd7fd52fc5b/collect_app/src/main/java/org/odk/collect/android/activities/InstanceChooserList.java#L108

seadowg:
@lognaturel ok digging even deeper it looks like this has never been supported but `InstanceChooserList` does seem to include the code that would make it work, so it's likely people are using the activity directly (which we don't want). I can add the `PICK` action here so people can migrate.

lognaturel (approved):

Issue #4438, opened 2021-02-18
Analytics for projects


lognaturel:
General usage questions:
- [x] How many projects do users typically have?
- [x] How frequently do users add projects?
- [x] How frequently do users delete projects?
- [x] How frequently do users reconfigure an existing project via QR code?

Specific feature usage:
- [x] How frequently do users switch projects?
- [x] Is adding a project by QR code or manually more common?
- [x] How frequently do users try a demo from the initial landing screen?
- [x] How frequently do users configure a Google Sheets project from the manual configuration screen?
- [x] How frequently do users manually set project name? project character? project color?
- [x] How often are forms added on disk (via disk sync)?

PR #4711, opened 2021-07-21
Add menu option to create project from QR code image

### LINKED ISSUE #4709: Add overflow menu with QR code import from file to initial project configuration screen
We know from analytics and talking to users that some people configure projects from QR code images. This is required for workflows in which data collectors are remote or there are devices without cameras/with broken cameras.

This functionality is still available from Settings > Project Management > Reconfigure with QR code but that's pretty buried and it's a big change from v1.30 that would require retraining.

- [ ] When I access the "Add project" screen either from the first launch screen or from "Add project" from the project dialog
 And I tap on the overflow menu
 And I tap on "Import QR code"
 I see a file picker and the same experience that I can currently access from Settings > Project Management > Reconfigure with QR code

lognaturel:
Closes #4709

Some ways to proceed:
- Someone else takes over to write that Espresso test I couldn't get working
- Merge as-is and I keep looking at testing over the next couple days
- Try to consolidate some of the different QR code paths now. I think this would likely require someone else to take over

Manual verification only. I verified that the overflow menu appears when accessing the qr code configuration screen both when first configuring Collect or when adding a new project. I tried a valid qr code, an invalid qr code and an image without a qr code. I also verified the "reconfigure with QR code" tab interface for regressions.

I also decided to remove QR code related analytics here. I don't think they're in the right places or have the right names to be actionable anymore. I apologize, @seadowg, this will likely give you merge conflicts since you've started on #4438. We want to answer related questions and I think it's important we use new events so that we capture behavior in v2021.2 and after only.

I really wanted to get an Espresso test for this but had to back out. I can successfully get the option menu to be clicked but then I can't click "Import QR code".

I started out wanting to use `QRCodeMenuDelegate` and `QRCodeActivityResultDelegate`. I backed out due to time but it's something that should be come back to. There's now disparate QR code handling code in a bunch of places. The main blocker I had was the difference between creating a brand new project vs. populating an existing project. There's some weird flow around that.

This is intended to be purely additive. Risk is concentrated around the QR code scanning dialog for adding new projects. Another area of risk is that there's duplication between the QR code scanning dialog for adding new projects and reconfiguring projects so that behavior could be inconsistent.

No.

Part of https://github.com/getodk/docs/issues/1365


collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
```diff
@@ -88,6 +125,26 @@ class QrCodeProjectCreatorDialog : MaterialFullScreenDialogFragment(), Duplicate
         return binding.root
     }
 
+    private fun configureMenu() {
+        binding.toolbar.menu.removeItem(R.id.menu_item_share)
+
+        binding.toolbar.setOnMenuItemClickListener {
+            when (it.itemId) {
+                R.id.menu_item_scan_sd_card -> {
```
grzesiek2010, line 133: This the only option in the tootlbar. It would be better to have an icon for it instead of hiding it. It might be a separate issue since if we decide to add such and icon we need to do the same for 'Reconfigure with qr code' screen.

seadowg:
@lognaturel I can have a think about testing for this today and we can chat about it later.

seadowg (changes requested):
It feels to me like this could be implemented by adapting `QrCodeOptionsMenuDelegate` and `QrCodeActivityResultDelegate` to be more general and then sharing them between `QrCodeProjectCreatorDialog` and `QrCodeTabsActivity` (or potentially a more generalized component). Personally, I'm feeling like it makes sense to push back on rushing this and potentially thinking about it as an enhancement we can release as a patch rather than essential for release?

We're already feeling that the QR and project creation paths are needlessly disparate, and I'm worried we're adding to debt we'd need to clean up here. As far as I'm aware, people are still able to set projects up with a file (it's just a bit awkward) but more importantly, existing single-project installs won't feel any drawbacks from not having this.

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
lognaturel: Not displaying it is intentional. When we previously considered it we wanted to make sure it wasn’t eye-catching and that we had an opportunity to use text to explain what it does. It still feels like the right decision to me.

lognaturel:
Unfortunately I think it had to be in for release. Remember that users have no control over when they get the update so they could have trained their staff and then by the time the staff gets to the field they have the new version. Even if it’s not a huge number of users who rely on this path, I’d rather not force them to a dead end. 

I understand your point but we also have the same problem on the scanning side that this helped expose. I don’t think we gain much by delaying — this duplication is isolated and easy to keep track of.

seadowg:
> Even if it’s not a huge number of users who rely on this path, I’d rather not force them to a dead end.

I still think my preference here would be to document the path they can take (manual or demo and then import via settings) and hold off on implementation as a patch. However, it feels like a compromise would to create an issue to use `QrCodeOptionsMenuDelegate`/`QrCodeActivityResultDelegate` in both places and prioritize it as part of 2021.3 (or even a 2021.2.1) and then move forward with this implementation tested by QA. How does that sound?

lognaturel:
> create an issue to use QrCodeOptionsMenuDelegate/QrCodeActivityResultDelegate in both places and prioritize it as part of 2021.3 (or even a 2021.2.1) and then move forward with this implementation tested by QA.

Yes, let’s do that. Did a sanity check with @yanokwa and he also finds it important not to disrupt this path.

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
lognaturel: Sorry, was going quickly so I think that came across as more aggressive than I intended! Fundamentally, you're right that displaying a single item in an overflow menu is an anti-pattern. I think here it's kind of the "least bad" alternative. Most importantly, it matches the prior workflow so meets our most important goal of helping users who already have a process built around this functionality.

lognaturel:
Apologies all around for the process here, I know it wasn't ideal. Talked to @seadowg and he confirmed he's ok for a merge as-is with some time spent at the beginning of the next release to untangle all the places where there's QR code functionality.

I'm going to merge now so that I can get a (hopefully last) beta out.

PR #4713, opened 2021-07-21
Project delete does not fully clear state

### LINKED ISSUE #4710: Project delete does not fully clear state
- If the last project is deleted, the meta setting for current project id is still set. The implication is that getting the current project results in a different exception which is not the end of the world. I ran into this working on #4707 but I don't really know how important it is.

- There seems to be some settings retained. I don't yet have exact repro steps but I did something like create a demo project from the special link from the first launch screen, reconfigure that project, delete that project, then manually add the demo project URL and ended up with the reconfigured settings.

grzesiek2010:
Closes #4710 

I added clearing settings for the project that we remove in `ProjectDeleter` which is responsible for deleting. This is the best place and I can't think of any better solution.

It's a pretty safe and isolated change so we just need to verify that the bug does not exist and that's it.

No.

No.

### DIFF
--- a/collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
@@ -13,6 +13,10 @@ import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.odk.collect.android.backgroundwork.FormUpdateScheduler
 import org.odk.collect.android.backgroundwork.InstanceSubmitScheduler
+import org.odk.collect.android.preferences.keys.AdminKeys
+import org.odk.collect.android.preferences.keys.GeneralKeys
+import org.odk.collect.android.preferences.keys.MetaKeys
+import org.odk.collect.android.support.InMemSettingsProvider
 import org.odk.collect.android.utilities.ChangeLockProvider
 import org.odk.collect.forms.instances.Instance
 import org.odk.collect.formstest.InMemInstancesRepository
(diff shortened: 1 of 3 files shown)


seadowg (changes requested):
I'm wondering if a simpler implementation (which would also need less testing) would be to just have the `ProjectDeleter` clear the general and admin settings for the project directly. That way we only need to make changes to `ProjectDeleter`. It could also be tested pretty easily using the `InMemSettingsProvider` - the general and admin could be set up with any keys (`foo`, `bar` etc) and then checked to see that they are gone after project deletion.

I guess that would leave references to old settings objects around in `SharedPreferencesSettingsProvider` but they're just in-mem and I don't think would cause any problems?

What do you think?

lognaturel (dismissed):
Great! I think the most compelling reason to do this ends up being not to leave user credentials lying around.

lognaturel:
Sigh, hadn't noticed @seadowg's change request. Simpler and more easily testable is certainly nice. I agree it wouldn't be an issue to temporarily leave settings in memory. But I don't feel strongly about making the change.

grzesiek2010:
> What do you think?

Yeah it makes the fix a bit smaller so here you go.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDeleterTest.kt
```diff
@@ -22,6 +28,7 @@ import org.odk.collect.shared.TempFiles
 import org.odk.collect.testshared.BooleanChangeLock
 import java.io.File
 
+@RunWith(AndroidJUnit4::class)
```
seadowg: I don't think you need the Android test runner here right?

seadowg (approved):
Looks great! One comment about a line I don't think we need, but nothing blocking this getting merged for me.

PR #4723, opened 2021-07-22
Add analytics for projects

### LINKED ISSUE #4438: Analytics for projects

### LINKED ISSUE #4605: Update analytics behaviour for projects

seadowg:
Closes #4605
Closes #4438

I added tests for analytics initialization, but everything else is untested (as we generally do for analytics).

Comments inline. I did think about using the new [App Startup](https://developer.android.com/topic/libraries/app-startup#disable-individual) library for handling initialization, but it looked like it would take a little work to be able to get set up for it. Expect a "between release cycle" PR for that...

Should just make it possible to have analytics in release versions (good for someone to check) and add the events we want.

### DIFF
--- a/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
+++ b/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
@@ -14,4 +14,17 @@ interface Analytics {
     fun logNonFatal(message: String)
     fun setAnalyticsCollectionEnabled(isAnalyticsEnabled: Boolean)
     fun setUserProperty(name: String, value: String)
+
+    companion object {
+
+        private var instance: Analytics = NoopAnalytics()
+
+        fun setInstance(analytics: Analytics) {
+            this.instance = analytics
+        }
+
+        fun logEvent(event: String) {
+            instance.logEvent(event)
+        }
+    }
 }
--- a/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
+++ b/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
@@ -23,7 +23,8 @@ interface Analytics {
             this.instance = analytics
         }
 
-        fun logEvent(event: String) {
+        @JvmStatic
+        fun log(event: String) {
             instance.logEvent(event)
         }
     }
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -198,4 +198,11 @@ public class AnalyticsEvents {
      * Tracks how often users delete projects
      */
     public static final String DELETE_PROJECT = "DeleteProject";
+
+    /**
+     * These events track how often users change project display settings
+     **/
+    public static final String CHANGE_PROJECT_NAME = "ChangeProjectName";
+    public static final String CHANGE_PROJECT_ICON = "ChangeProjectIcon";
+    public static final String CHANGE_PROJECT_COLOR = "ChangeProjectColor";
 }
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -151,6 +151,11 @@ public class AnalyticsEvents {
     public static final String IMPORT_AND_ENCRYPT_INSTANCE = "ImportAndEncryptInstance";
 
     /**
+     * Tracks how often forms are added using disk sync rather than from a server
+     */
+    public static final String IMPORT_FORM = "ImportForm";
+
+    /**
      * Tracks responses from OpenMapKit to the OSMWidget
      */
     public static final String OPEN_MAP_KIT_RESPONSE = "OpenMapKitResponse";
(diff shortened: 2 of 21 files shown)


analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
```diff
@@ -14,4 +14,18 @@ interface Analytics {
...
+    companion object {
+
+        private var instance: Analytics = NoopAnalytics()
+
+        fun setInstance(analytics: Analytics) {
+            this.instance = analytics
+        }
+
+        @JvmStatic
+        fun log(event: String) {
```
seadowg, line 27: This is potentially a controversial change. Because we never test analytics (mainly because we often delete calls soon after adding them, and they are never critical) and because we never depend on its behaviour, it seems OK to me to have a static method for logging events. This of course goes against everything I would usually say, but I'll point you back to the last sentence in terms of justification for this rule break 😆.

For me, the advantage of this is that we don't have to rework constructor/method signatures when we add remove analytics calls from a component.

collect_app/src/main/java/org/odk/collect/android/application/initialization/AnalyticsInitializer.kt
```diff
@@ -0,0 +1,24 @@
+package org.odk.collect.android.application.initialization
+
+import org.odk.collect.analytics.Analytics
+import org.odk.collect.android.preferences.keys.GeneralKeys
+import org.odk.collect.android.preferences.source.SettingsProvider
+import org.odk.collect.android.version.VersionInformation
+
+class AnalyticsInitializer(
```
seadowg, line 8: As mentioned in the PR description, I've been eyeing up the App Startup library. A lot of app frameworks (across different platforms) use this multi object initialization style so moving to it felt pretty natural, allows for us to unit test individual chunks of init logic and should fit in well with App Startup when we do get there.

analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
lognaturel: Absolutely, I think this is pragmatic given we’ve decide we will very rarely if ever test analytics.

collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
```diff
@@ -176,12 +176,21 @@ private AnalyticsEvents() {
     public static final String TRY_DEMO = "TryDemo";
 
     /**
-     * Tracks how often projects are created using QR codes
-     */
+     * Tracks how often projects are created using QR codes. he total number of
+     * created projects can be calculated from the combined count of
+     * {@link AnalyticsEvents#MANUAL_CREATE_PROJECT} and {@link AnalyticsEvents#QR_CREATE_PROJECT}.
```
lognaturel: Trying a demo and configuring with Google Drive also create projects so maybe this comment is a little misleading? Creation from qr code image also isn't being tracked so we wouldn't be able to get a complete project count. I don't feel that we need to keep tracking that path since we've been tracking it for a while and we know it's important for those who need it but for most folks it's not necessary.

Does total project count seem meaningful to you?

collect_app/src/main/java/org/odk/collect/android/projects/ManualProjectCreatorDialog.kt
```diff
@@ -81,6 +81,7 @@ class ManualProjectCreatorDialog : MaterialFullScreenDialogFragment(), Duplicate
                 confirmationArgs.putString(MATCHING_PROJECT, uuid)
                 DialogUtils.showIfNotShowing(DuplicateProjectConfirmationDialog::class.java, confirmationArgs, childFragmentManager)
             } ?: run {
+                Analytics.logEvent(AnalyticsEvents.GOOGLE_ACCOUNT_PROJECT)
```
lognaturel: This only gets the case where there is no existing matching project. I think we also need to track when there is an existing matching project and the user confirms creating a new one. In fact, I think we should track both cases.

lognaturel (commented):
I've left a couple comments inline that can be addressed after merge. One thing I'd like to do before merge is rename the events so they use "Projects" as prefix. Events are displayed in alphabetical order by default so that will make it easier to get a sense of what's going on at a glance. I'll probably do that tomorrow.

analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
lognaturel: I want to say it more strongly -- I think this is a good decision because it makes adding and removing analytics very lightweight.

collect_app/src/main/java/org/odk/collect/android/projects/ManualProjectCreatorDialog.kt
lognaturel: When I said "both cases", both confirming and switching projects after a duplicate project configuration attempt. At the time I was imagining we could do this separately for Google Drive and other manual configuration but that gets pretty unwieldy. So instead I tracked duplicate attempts and resolution by switching across all config methods.

lognaturel (approved):
When you get a chance, @seadowg, please review my updates and comments.

My guess is that most of the new events will need to get some kind of project identifier hash before we can start to get meaning out of them but we can see what we learn with these and add hashes later as needed.

Issue #4728, opened 2021-07-23
ACTION_VIEW for instances does the same thing as ACTION_EDIT


lognaturel:
Currently, whether an instance should be viewed or edited is determined not by the action but by an intent extra `formMode` which can have values `editSaved` and `viewSaved`. Not supplying the extra falls back to `editSaved`.

I think we should start by making the `ACTION_VIEW` action actually view submissions and use that internally instead of the extras.

Eventually we should deprecate the extra.

I also wonder whether it really makes sense for Collect to use implicit intents internally. It's a bit weird that launching a form for edit/view now goes through `FormUriActivity`. I don't think it's harmful, just unexpected.

PR #4729, opened 2021-07-23
Pass intent extras on to support submission edit and view

### LINKED ISSUE #4722: User is able to edit form when editing is disabled if user opens MapActivity view
Collect v2021.5

#### Problem description
User is able to edit form when editing is disabled if user opens MapActivity view

1. `Edit Saved Form` is unselected from Admin Settings
2. User opens new MapActivity
3. User sees marker for each form instance
4. User clicks on some marker
5. User is moved to hierarchy view
6. User IS ABLE to open question and EDIT it

---

1. `Moving backward` is unselected from the Admin Settings
2. User opens new MapActivity
3. User sees marker for each form instance
4. User clicks on some marker
5. User is moved to hierarchy view
6. User IS ABLE to open question and EDIT it

#### Expected behavior
User is not able to open question and edit it

lognaturel:
Closes #4722 

Added a step to the view sent form test to verify that the hierarchy view is in view-only mode. Manually tried:
- Disallow backwards navigation, save an instance as not finalized, verify it can't be edited from the instance map
- Save an instance as finalized, send it, view it, verify it can't be edited

This maintains the same strategy for expressing view vs. edit submissions. The use of an extra here is fundamentally strange, I think. I've filed that under https://github.com/getodk/collect/issues/4728 for later consideration because there's no need to take on additional risk now.

This change is small and non-invasive. I don't think there's any regression risk. However, seeing this issue makes me think there could be other regressions related to external apps calling into Collect.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/SendFinalizedFormTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/SendFinalizedFormTest.java
@@ -70,7 +70,8 @@ public class SendFinalizedFormTest {
 
                 .clickViewSentForm(1)
                 .clickOnForm("One Question")
-                .assertText("123");
+                .assertText("123")
+                .assertText(R.string.exit);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


grzesiek2010 (approved):

mmarciniak90:
**Tested with success**

Verified on Android versions: 8.1, 10.0

Verified cases:
* Edit Saved Form option unselected -> I didn't find the option to edit saved form, finalized form
* Moving backward option unselected -> I didn't find the option to edit saved form, finalized form
* re-select both options - editing was possible
* sent submission is not editable even when settings allow editing

PR #4731, opened 2021-07-26
Fixed displaying and hiding preferences and categories

### LINKED ISSUE #4714: When hide some user settings, headers are still visible
Collect v2021x, Android vx.x.x, device used...

#### Problem description
When testing regression I have noticed that when hide form management settings or collect anonymous usage data headers connected to that options are still visible when open Form management and User and device identity.
ex. I hid all form management settings except show guidance for questions:
| Collect v2021 | Collect 1.30.1 |
| ----- | ------|
|Screenshot_2021-07-21-13-37-32-359_org odk collect android | Screenshot (Jul 21, 2021 1_41_43 PM) |
| Screenshot_2021-07-21-13-52-31-709_org odk collect android | Screenshot (Jul 21, 2021 1_53_19 PM) |

1. Hide in User Settings form management settings and collect anonymous usage data 
2. Open Form Management and User and device identity

#### Expected behavior
Unnecessary headers should be hidden. 

#### Other information 

grzesiek2010:
Closes #4714 

I tested settings manually and added new automated tests.

The issue was small ad we could fix it in a smaller pr but I thought it's a good time to do it in a better way and add some additional improvements/refactoring.

The biggest change is that I completely replaced the old `DisabledPreferencesRemover` with `PreferenceVisibilityHandler`. Why?:
- The original version of `DisabledPreferencesRemover` would remove preferences, then we improved it so that preferences are not removed but hidden (so that the user experience is better) and now we had to update categories to work with that change so I had to edit this class either way.
- that class didn't handle hiding/showing preferences from the first (settings) screen so we had separate code for that in `ProjectPreferencesFragment`
- additionally that class was obsolete and contained some code that we probably should have removed long time ago. That's why I decided to reimplement it completly.

I also sorted out our keys `ProjectKeys`, `ProtectedProjectKeys`, `MetaKeys`, there is still something to do in that area but I didn't want to make this pr even bigger.

The change might affect all settings so we need to test them carefully. The good news is that we wanted to do that either way so its a good time.

No.

No.


collect_app/src/main/java/org/odk/collect/android/preferences/PreferenceVisibilityHandler.kt
```diff
@@ -0,0 +1,144 @@
...
+package org.odk.collect.android.preferences
+
+import androidx.preference.PreferenceGroup
+import androidx.preference.PreferenceScreen
+import org.odk.collect.android.preferences.keys.ProjectKeys
+import org.odk.collect.android.preferences.keys.ProtectedProjectKeys
+import org.odk.collect.android.preferences.source.SettingsProvider
+import org.odk.collect.android.version.VersionInformation
+
+class PreferenceVisibilityHandler(
```
grzesiek2010, line 25: It might seem like we don't have test for this class but in fact everything is tested in fragment tests like: `FormManagementPreferencesFragmentTest`

collect_app/src/main/java/org/odk/collect/android/preferences/PreferenceVisibilityHandler.kt
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright 2017 Nafundi
```
grzesiek2010, line 2: Probably I should remove it? This class look nothing like its old version.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferenceVisibilityHandler.kt
```diff
@@ -0,0 +1,144 @@
...
+                ProjectKeys.KEY_GUIDANCE_HINT -> preference.isVisible = state == ProjectPreferencesViewModel.State.UNLOCKED || isOptionEnabled(ProtectedProjectKeys.KEY_GUIDANCE_HINT)
+                ProjectKeys.KEY_EXTERNAL_APP_RECORDING -> preference.isVisible = state == ProjectPreferencesViewModel.State.UNLOCKED || isOptionEnabled(ProtectedProjectKeys.KEY_EXTERNAL_APP_RECORDING)
+                ProjectKeys.KEY_INSTANCE_SYNC -> preference.isVisible = state == ProjectPreferencesViewModel.State.UNLOCKED || isOptionEnabled(ProtectedProjectKeys.KEY_INSTANCE_FORM_SYNC)
+
+                ProjectKeys.KEY_FORM_METADATA -> preference.isVisible = state == ProjectPreferencesViewModel.State.UNLOCKED || isOptionEnabled(ProtectedProjectKeys.KEY_CHANGE_FORM_METADATA)
+                ProjectKeys.KEY_ANALYTICS -> preference.isVisible = state == ProjectPreferencesViewModel.State.UNLOCKED || isOptionEnabled(ProtectedProjectKeys.KEY_ANALYTICS)
+
+                "experimental" -> preference.isVisible = !versionInformation.isRelease
```
grzesiek2010, line 102: I know all those should be constants placed somewhere but our codebase is full of places where we use const then strings like here. I would fix this but I didn't want to make this pr even bigger, so maybe later.

collect_app/src/main/java/org/odk/collect/android/preferences/DisabledPreferencesRemover.kt
```diff
@@ -31,11 +31,7 @@ class DisabledPreferencesRemover(
         hideEmptyCategories(pf.preferenceScreen)
     }
 
-    /**
-     * Removes any preferences from the category that are excluded by the admin settings.
-     *
-     * @param keyPairs one or more AdminAndGeneralKeys objects.
-     */
+    // Hides preferences that are excluded by the admin settings
```
lognaturel: Is there a reason you don't use KDoc?  The nice thing about using that structure is that it makes code navigation easier by providing quick definitions for symbols.

lognaturel (commented):
I've unfortunately only been able to take a pretty quick look. I don't see any blockers but this does move a lot around so feels pretty risky.

Unlock protected settings, rotate, "Unlock protected settings" reappears.

I think we should consider squashing this one on merge. The small commits are great for review so thank you! But I think they'd make blames and tracking meaningful changes harder.

collect_app/src/main/java/org/odk/collect/android/preferences/DisabledPreferencesRemover.kt
grzesiek2010: I just thought that in this case I needed a simple and short comment nothing more.

grzesiek2010:
> Unlock protected settings, rotate, "Unlock protected settings" reappears.

Thanks, it's fixed.

mmarciniak90:
**Tested with success**

Verified cases:
* exploratory testing
* hide user settings -> headers are hidden too (4714)
* 'Unlock protected settings' area and rotation
* unlocked protected settings from the middle of the form if another option was changed - [case](https://github.com/getodk/collect/pull/4687#issuecomment-881498482)
* uncheck/check option in User settings like server, maps which hides whole area and rotate - [case](https://github.com/getodk/collect/pull/4687#issuecomment-880092398)
* rotation
* light/dark theme

lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/preferences/DisabledPreferencesRemover.kt
lognaturel: I think it's always preferable to use syntax that the IDE can use when commenting on some symbol that can be used somewhere else (even if it's private) but not important enough to change here.

collect_app/src/main/java/org/odk/collect/android/preferences/PreferenceVisibilityHandler.kt
lognaturel: In this case it doesn't particularly matter since Nafundi can grant copyright to ODK. Maybe it would have been preferable to update it to 2021 ODK but we're not consistent with these headers at all anyway.

PR #4734, opened 2021-07-27
Remove cache folder instead of migrating

### LINKED ISSUE #4733: When form comes from v1.30 cache, selects from internal instances are broken in v2021.2
Collect v2021.2 betas

#### Problem description
[Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/29c4a0ad7245a5bb01fc22ccf7c46d8a?time=last-seven-days&sessionEventKey=60FCA878004E0001551C15B0B2B3317D_1567092885019574065) 

1. Open any form that has a select from internal secondary instance such as [this one](https://docs.google.com/spreadsheets/d/1wHRc5fDUZc06EyEPaGPP90uG-QgpLkuDkG6XIH5Hft0/edit#gid=0) in v1.30
2. Upgrade to master or a v2021.2 beta
3. Open the form and go to the select question
4. See dialog with error `XPath evaluation: Instance xyz not found`, select question can't be opened

#### Expected behavior
When upgrading from v1.30, select questions continue to work

#### Other information 
- External secondary instances seem to work as expected
- [This](https://github.com/getodk/javarosa/compare/v3.1.0...master) is the JavaRosa diff. I searched for `readExternal`/`writeExternal` thinking there might have been a change to serialization but there was none. I searched for `instanceName` but only usage has changed, not how it's written.
- [This](https://github.com/getodk/collect/commits/e4699b7b2af0150c3f71a8ab0c28866b0c3d08da/collect_app/src/main/java/org/odk/collect/android/utilities/FormDefCache.java) is the diff for `FormDefCache`. Nothing interesting.

lognaturel:
Closes #4733

Modified `ExistingProjectMigratorTest` to reflect new intended behavior.

Deleted everything in `/sdcard/Android/data/org.odk.collect.android/files`. Filled [this simple select form](https://docs.google.com/spreadsheets/d/1wHRc5fDUZc06EyEPaGPP90uG-QgpLkuDkG6XIH5Hft0/edit#gid=0) in v1.30.1. Installed from this branch. Verified that the newly-migrated cache folder is empty. Note that this can be difficult to reproduce because meta settings stick around in ways I don't quite understand. I ended up having to change some manually.

I considered making changes to the way internal secondary instances are cached instead. I think we should possibly revisit that at some point but now doesn't feel like a good time. Blowing away the cache seems reasonable when we're making such a big change to the app.

The intention is that folks who have cached files with selects in v1.30 or earlier will be able to use their forms without error after upgrading. A side effect of this change is that form load may be slower the first time each form is launched in v2021.2. I don't think this is significant enough to be a problem.

Any form with an internal secondary instance. See issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
@@ -46,7 +46,6 @@ class ExistingProjectMigrator(
             File(rootDir, "instances"),
             File(rootDir, "metadata"),
             File(rootDir, "layers"),
-            File(rootDir, ".cache"),
             File(rootDir, "settings")
         ).forEach {
             try {
@@ -56,6 +55,8 @@ class ExistingProjectMigrator(
                 // Original dir doesn't exist - no  need to copy
             }
         }
+
+        FileUtils.deleteDirectory(File(rootDir, ".cache"))
 
         val adminSharedPrefs = context.getSharedPreferences("admin_prefs", Context.MODE_PRIVATE)
         settingsProvider.getGeneralSettings(project.uuid).saveAll(generalSharedPrefs.all)
--- a/collect_app/src/test/java/org/odk/collect/android/projects/ExistingProjectMigratorTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/projects/ExistingProjectMigratorTest.kt
@@ -52,7 +52,6 @@ class ExistingProjectMigratorTest {
             File(rootDir, "instances"),
             File(rootDir, "metadata"),
             File(rootDir, "layers"),
-            File(rootDir, ".cache"),
             File(rootDir, "settings")
         )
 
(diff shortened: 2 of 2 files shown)


grzesiek2010 (approved):

lognaturel:
It might be easier to verify using [the published v1.30.1](https://github.com/getodk/collect/releases/tag/v1.30.1) and then 
[ODK-Collect-v2021.2-beta.5-37-g11cff09e1.apk.zip](https://github.com/getodk/collect/files/6886445/ODK-Collect-v2021.2-beta.5-37-g11cff09e1.apk.zip)

Note: I haven't made any changes just now, I just pushed the branch to upstream to make it easier for me to build the signed APK. I'm not sure why CI is running again.

mmarciniak90:
**Tested with success**

Verified cases:
* Central server
* Aggregate server
* Google drive 
* updated with finalized submission - after the update, I was able to edit and send it
* update with saved submission - after the update, I was able to edit it, finalize and send it

PR #4745, opened 2021-07-28
Fixed navigating away during QR code camera permission request

### LINKED ISSUE #4741: App can crash if user navigates away during QR code camera permission request
Collect v2021.2

#### Problem description

This is caused by the permission listener being called from a `Fragment` that might not exist anymore if the user navigates away and returns to the app. It's likely that this problem exists in more places than just the QR code scanner.

1. Enable "Don't Keep Activities" in Android's developer settings
2. Load up the app and click "Configure with QR code"
3. Navigate away and back to the app
4. Grant the camera permission

The app will crash.

#### Expected behavior

The app shouldn't crash and the QR code scanner should start working when the permission is granted. 

The easiest way to fix this is to wrap the listener logic in a check for `isAdded()` in the fragment. It would be good to check that there aren't other parts of the app that this explodes in as well.

#### Other information 

Crashlytics report: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/2607b09cab75f549f23db58cc5e2ccb4?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=6100DFE5035A00013F5508F69F6D6234_1568284941612917881

An example stack trace:

```
Caused by java.lang.IllegalStateException: Fragment QrCodeProjectCreatorDialog{e3de41b} (5918c3b4-96da-410e-b677-dacfc4fc02ca) not attached to an activity.
       at androidx.fragment.app.Fragment.requireActivity(Fragment.java:928)
       at org.odk.collect.android.projects.QrCodeProjectCreatorDialog.startScanning(QrCodeProjectCreatorDialog.java:204)
       at org.odk.collect.android.projects.QrCodeProjectCreatorDialog.access$startScanning(QrCodeProjectCreatorDialog.java:41)
       at org.odk.collect.android.projects.QrCodeProjectCreatorDialog$onCreateView$3.granted(QrCodeProjectCreatorDialog.java:129)
       at org.odk.collect.android.permissions.PermissionsProvider$2.granted(PermissionsProvider.java:80)
       at org.odk.collect.android.permissions.PermissionsProvider$8.onPermissionGranted(PermissionsProvider.java:190)
       at com.karumi.dexter.MultiplePermissionsListenerToPermissionListenerAdapter.onPermissionsChecked(MultiplePermissionsListenerToPermissionListenerAdapter.java:35)
       at com.karumi.dexter.MultiplePermissionListenerThreadDecorator$1.run(MultiplePermissionListenerThreadDecorator.java:8)
       at com.karumi.dexter.MainThread.execute(MainThread.java:6)
       at com.karumi.dexter.MultiplePermissionListenerThreadDecorator.onPermissionsChecked(MultiplePermissionListenerThreadDecorator.java:7)
       at com.karumi.dexter.DexterInstance.onPermissionsChecked(DexterInstance.java:56)
       at com.karumi.dexter.DexterInstance.updatePermissionsAsGranted(DexterInstance.java:26)
       at com.karumi.dexter.DexterInstance.onPermissionRequestGranted(DexterInstance.java)
       at com.karumi.dexter.Dexter.onPermissionsRequested(Dexter.java:4)
       at com.karumi.dexter.DexterActivity.onRequestPermissionsResult(DexterActivity.java:51)
       at android.app.Activity.dispatchRequestPermissionsResult(Activity.java:7994)
       at android.app.Activity.dispatchActivityResult(Activity.java:7816)
       at android.app.ActivityThread.deliverResults(ActivityThread.java:4700)
       at android.app.ActivityThread.handleSendResult(ActivityThread.java:4749)
       at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:49)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1991)
       at android.os.Looper.loop(Looper.java:216)
       at android.app.ActivityThread.main(ActivityThread.java:7258)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
```

grzesiek2010:
Closes #4741 

I reproduced the issue and verified that the fix does the job.

It's the easiest solution I can't think of anything better.

It's a safe fix so we can focus on verifying that the issue no longer exist.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
@@ -126,7 +126,9 @@ class QrCodeProjectCreatorDialog :
             requireActivity(),
             object : PermissionListener {
                 override fun granted() {
-                    startScanning(savedInstanceState)
+                    if (isAdded) {
+                        startScanning(savedInstanceState)
+                    }
                 }
 
                 override fun denied() {
--- a/collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
@@ -126,6 +126,7 @@ class QrCodeProjectCreatorDialog :
             requireActivity(),
             object : PermissionListener {
                 override fun granted() {
+                    // Do not call from a fragment that does not exist anymore https://github.com/getodk/collect/issues/4741
                     if (isAdded) {
                         startScanning(savedInstanceState)
                     }


grzesiek2010:
I took a quick look at reconfiguring with QR code (available in settings) expecting that there the behavior might be the same. The thing is that it works in a different way because we are asked for granting camera permission first and if we deny we are moved back to settings. I think this behavior should be changed and match what we have in adding projects with QR codes - even if we deny we are still in that fragment and we can for example load setting from a file. Do you agree @seadowg @lognaturel if so we can file a separate issue to change it keeping in mind that cases like this one should be handled.

seadowg:
@grzesiek2010 yeah, that sounds right. I'm actually going to create an issue around deduplicating the QR scanning code,  so I'll add a note there about this.

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
```diff
@@ -126,7 +126,9 @@ class QrCodeProjectCreatorDialog :
             requireActivity(),
             object : PermissionListener {
                 override fun granted() {
-                    startScanning(savedInstanceState)
+                    if (isAdded) {
```
seadowg, line 130: I'm not sure if there is a good way to add a test for this (it would be good to experiment if it can be simulated with `FragmentScenario`) but if not I think we should at least have comment on why we're doing this.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/projects/QrCodeProjectCreatorDialog.kt
grzesiek2010: I added a comment since I don't think there is a good way to test it. We would need to UIAutomator to accept that permission because espresso can't access components from outside of our application’s package (like those permission dialogs).

seadowg: Yeah fair enough!

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android versions: 8.1, 9.0, 10.0, 11.0
There is no permission manipulation option on Android 5.1 but I have also tested it to see if there is regression.

Verified cases:
* described case in issue executed from the first launch screen
* described case in issue executed from project dialog -> 'Add project' 
* similar case with Contact permission when Google Drive project was created

PR #4746, opened 2021-07-28
Throw exception on app launch if storage is not available

seadowg:
This will hopefully replace crashes we see in a couple of places (when trying to access external storage or creating directories, for instance) by causing the app to fail fast rather than later on in initialization. The goal is to identify if this is the root cause of other storage related crashes we're seeing (and to reduce noise in Crashlytics). We know the app won't function without storage, so I don't see any risks to adding this.

Ran all tests.

We could try and show a toast or something to the user, but I think for the moment it's best to identify if this indeed the problem and then if so, we can look at some UI in this situation if it seems like a priority.

I guess there might be cases where the app would work without access to storage, but I'm not sure I understand how that could be - we need storage access to hit our forms/instances dbs for example.


collect_app/src/test/java/org/odk/collect/android/application/RobolectricApplication.java
```diff
@@ -18,6 +20,9 @@
 
     @Override
     public void onCreate() {
+        // Make sure storage is accessible
+        ShadowEnvironment.setExternalStorageState(MEDIA_MOUNTED);
```
seadowg, line 24: We don't actually need this for the final implementation, but it should avoid problems in the future with objects that check this.

lognaturel (approved):

PR #4748, opened 2021-07-28
Catch exceptions creating project name file

### LINKED ISSUE #4747: Don't crash if creating project name file fails
Collect v2021.2

#### Problem description

We're seeing crashes when the app creates the project name file in the project directory sometimes.

No idea!

#### Expected behavior

We should just catch an error here and log a timber exception.

#### Other information 

Crashlytics 1: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/6cb3be6fd3c7416b7e26e3e323a94e69?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=610164F300AF00015A16A4C28B43EB15_1568426655704680846

Crashlytics 2: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/6cadfda014bd31a6beb82a9a4061f9b0?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=6100EC6602BF00012801ACBF9F2799DD_1568293742726811014

lognaturel:
Closes #4747

Ensured the app compiles and runs. We can't reproduce this so it's the best we can do.

The goal for now is to not block on creating this file and to learn more about why there might be an exception. @seadowg and I briefly considered escaping the project name in some way but we don't know enough about what's causing the issue to know how to do so. Something like URL-encoding would distort the name to the point of being unusable in many cases. It doesn't seem like a good option given that all the devices we've tried can write any filename we can think of.

All this does is wrap non-essential behavior in try/catch. It should just mean that users who have experienced crashes no longer do and should present no regression risk.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
@@ -18,6 +18,8 @@ import android.text.Html;
 
 import androidx.annotation.NonNull;
 
+import com.google.common.base.CharMatcher;
+
 import java.util.Iterator;
 import java.util.regex.MatchResult;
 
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/StringUtils.java
```diff
@@ -190,5 +192,13 @@ public static String removeEnd(String str, String remove) {
 
         return str;
     }
+
+    public static String getFilenameError(String filename) {
+        String possiblyRestricted = "?:\"*|/\\<>\u0000";
```
grzesiek2010, line 197: You don't wan't to log the whole name for privacy reasons or what?

seadowg: Yeah, @lognaturel and I chatted last night, and we wanted to avoid that.

grzesiek2010 (approved):

PR #4754, opened 2021-07-29
Stop crashes when cache dir can't be accessed during savepoint load

seadowg:
Should address this crash: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/4108b44429d720d85eb77fdc94c076a5?time=last-seven-days&versions=v2021.2.0%20(4242)&sessionEventKey=6102AD3103B3000140D7979B2770F328_1568785630859318267

Ran the app and tests with hacked code that never creates the cache directory to check that form loading now works without it. Amazingly, every one of our instrumentation tests passes with no cache directory now.

This doesn't fix the problem, but does make it so the app can survive whatever is happening and continues to log that something is wrong.

Should just fix the problem. It could be that other problems will come up later in a user's flow now the app doesn't explode at this stage, but I don't think we need to QA this change.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -645,42 +645,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
                  * Savepoints for forms that were explicitly saved will be recovered when that
                  * explicitly saved instance is edited via edit-saved-form.
                  */
-                final String filePrefix = formPath.substring(
-                        formPath.lastIndexOf('/') + 1,
-                        formPath.lastIndexOf('.'))
-                        + "_";
-                final String fileSuffix = ".xml.save";
-                File cacheDir = new File(storagePathProvider.getOdkDirPath(StorageSubdirectory.CACHE));
-                File[] files = cacheDir.listFiles(pathname -> {
-                    String name = pathname.getName();
-                    return name.startsWith(filePrefix)
-                            && name.endsWith(fileSuffix);
-                });
-
-                /**
-                 * See if any of these savepoints are for a filled-in form that has never
-                 * been explicitly saved by the user.
-                 */
-                for (File candidate : files) {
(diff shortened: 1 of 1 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -693,6 +658,51 @@ private void loadFromIntent(Intent intent) {
...
+                        && instanceDir.isDirectory()
+                        && !instanceFile.exists()) {
+                    // yes! -- use this savepoint file
+                    return instanceFile
+                            .getAbsolutePath();
+                }
+            }
+
+        } else {
+            Timber.e("Couldn't access cache directory when looking for save points!");
```
lognaturel, line 700: This will just mean we get a crash when trying to save a savepoint, right? Should we preemptively address that?

seadowg: Oh I can check. All our tests actually pass in the scenario where the cache directory is never created, so it might actually be fine - I was thinking it was tested, but it might not be.

seadowg: I tested savepoints without a cache directory and nothing exploded. If you look at `SavePointTask` and the call to `writeFile` you'll see that we create the directory if needed, which must be why. That's nice, as it's closer to the "lazy" creation of directories I'd like to see us move towards anyway.

lognaturel: Oh, how fabulous! Thanks for the extra check.

lognaturel (approved):

mmarciniak90:
**Tested with success**

Verified on Android 5.1, 8.1, 9.0, 10.0, 11.0

Verified cases:
- removed .cache directory in a project - crash does not occur but a consequence of that is expected data lost
- removed file in .cache directory
- removed .cache directory with unsaved changes for previously saved form

PR #4757, opened 2021-07-30
Fixed enable auto-send schedules submission of completed forms #3980

### LINKED ISSUE #3980: Enabling Auto send should submit forms on device
When the user enabled "Auto send" in Form Management settings should schedule submission of the forms currently on the device. Right now forms would not auto submit until another instance was filled out. 

**Acceptance**

- [ ] Given I've filled out (and finalized) a form
And I've still not sent that form to the server
When I go to Form Management settings (in General settings)
And I enable Auto send
And my connection is correct for the auto send type I enabled
Then my form is sent to the server

TomRCummings:
Closes #3980 

Tested manually with device and added both Roboelectric and Espresso tests.

The desired behavior results from a change in the Auto-send form management preference, so it seems appropriate to handle that behavior with `onSettingChange` listener function in `FormManagementPreferencesFragment`.

With this change, any time the Auto-send preference is toggled to something other than "off" a new `submitSchedule` call is made to the `InstanceSubmitScheduler`. I don't think this carries any risk with it, but I am not totally sure. Is it necessary to call `cancelSubmit` if a submission has been scheduled but not yet completed and the user toggles Auto-send to something different?

No specific form required.

No documentation changes required.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
@@ -6,7 +6,6 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.junit.runner.RunWith;
-import org.odk.collect.android.R;
 import org.odk.collect.android.support.CollectTestRule;
 import org.odk.collect.android.support.NotificationDrawerRule;
 import org.odk.collect.android.support.TestDependencies;
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
@@ -6,6 +6,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.junit.runner.RunWith;
+import org.odk.collect.android.R;
 import org.odk.collect.android.support.CollectTestRule;
 import org.odk.collect.android.support.NotificationDrawerRule;
 import org.odk.collect.android.support.TestDependencies;
@@ -61,6 +62,27 @@ public class AutoSendTest {
         mainMenuPage
                 .clickViewSentForm(1)
                 .assertText("One Question Autosend");
+
+        notificationDrawerRule.open()
+                .assertAndDismissNotification("ODK Collect", "ODK auto-send results", "Success");
+    }
+
+    @Test
+    public void whenFormIsFinalized_enablingAutoSend_schedulesSubmitFinalizedForm() {
+        MainMenuPage mainMenuPage = rule.startAtMainMenu()
+                .setServer(testDependencies.server.getURL())
+                .copyForm("one-question.xml")
+                .startBlankForm("One Question")
+                .inputText("31")
+                .swipeToEndScreen()
+                .clickSaveAndExit()
+                .enableAutoSend();
+
+        testDependencies.scheduler.runDeferredTasks();
+
+        mainMenuPage
+                .clickViewSentForm(1)
+                .assertText("One Question");
 
         notificationDrawerRule.open()
                 .assertAndDismissNotification("ODK Collect", "ODK auto-send results", "Success");
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
@@ -65,25 +65,4 @@ public class AutoSendTest {
         notificationDrawerRule.open()
                 .assertAndDismissNotification("ODK Collect", "ODK auto-send results", "Success");
     }
-
-    @Test
-    public void whenFormIsFinalized_enablingAutoSend_schedulesSubmitFinalizedForm() {
-        MainMenuPage mainMenuPage = rule.startAtMainMenu()
-                .setServer(testDependencies.server.getURL())
-                .copyForm("one-question.xml")
-                .startBlankForm("One Question")
-                .inputText("31")
-                .swipeToEndScreen()
-                .clickSaveAndExit()
-                .enableAutoSend();
-
-        testDependencies.scheduler.runDeferredTasks();
-
-        mainMenuPage
-                .clickViewSentForm(1)
-                .assertText("One Question");
-
-        notificationDrawerRule.open()
-                .assertAndDismissNotification("ODK Collect", "ODK auto-send results", "Success");
-    }
 }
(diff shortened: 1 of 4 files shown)


lognaturel:
Thanks so much for taking this on, @TomRCummings! I took a quick look and it looks ready for review to me. Is there a reason it's still a draft?

TomRCummings:
Excellent, I'll open it for review! I opened it as a draft out of an excess of caution.

seadowg:
@TomRCummings apologies on the delay in getting this reviewed. We're just finishing off our 2021.2 release (which was a bigger one than usual). I'm hoping we can start looking at PRs again next week.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormManagementPage.java
```diff
@@ -19,4 +19,9 @@ public FormManagementPage assertOnPage() {
         clickOnString(R.string.form_update_frequency_title);
         return new ListPreferenceDialog<>(R.string.form_update_frequency_title, this).assertOnPage();
     }
+
+    public ListPreferenceDialog<FormManagementPage> clickAutoSend() {
```
seadowg, line 23: This is unused. Were you planning to use it in `MainMenuPage#enableAutoSend`?

collect_app/src/test/java/org/odk/collect/android/preferences/screens/FormManagementPreferencesFragmentTest.kt
```diff
@@ -487,4 +499,15 @@ class FormManagementPreferencesFragmentTest {
...
         }
     }
+
+    @Test
+    fun `When Auto send preference is enabled, finalized forms should be scheduled for submission`() {
+        generalSettings.save(ProjectKeys.KEY_AUTOSEND, "off")
+
+        val scenario = FragmentScenario.launch(FormManagementPreferencesFragment::class.java)
+        scenario.onFragment { fragment: FormManagementPreferencesFragment ->
+            generalSettings.save(ProjectKeys.KEY_AUTOSEND, "wifi")
```
seadowg: Instead of changing the value in the underlying `Settings` (where settings are stored) we probably want to try simulating changing the Android Preference (the UI component) value. That could be like:

```kotlin
fragment.findPreference<ListPreference>(ProjectKeys.KEY_AUTOSEND)!!.value = "wifi"
```

That will just check that the UI is hooked up correctly as well as the listeners on settings.

collect_app/src/test/java/org/odk/collect/android/preferences/screens/FormManagementPreferencesFragmentTest.kt
```diff
@@ -487,4 +499,15 @@ class FormManagementPreferencesFragmentTest {
             assertThat(fragment.findPreference<PreferenceCategory>("form_import")!!.isVisible, `is`(true))
         }
     }
+
+    @Test
+    fun `When Auto send preference is enabled, finalized forms should be scheduled for submission`() {
```
seadowg, line 505: I think in addition to this we should have a test that checks that when I switch auto send `off` it doesn't schedule submissions. This prevents us from forgetting to have the `if` wrapped around the submission scheduling in the implementation.

seadowg (changes requested):
This is looking great! I think we should tweak the approach to testing slightly.

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
```diff
@@ -65,4 +65,25 @@ public void whenFormHasAutoSend_fillingAndFinalizingForm_sendsFormAndNotifiesUse
         notificationDrawerRule.open()
                 .assertAndDismissNotification("ODK Collect", "ODK auto-send results", "Success");
     }
+
+    @Test
+    public void whenFormIsFinalized_enablingAutoSend_schedulesSubmitFinalizedForm() {
```
seadowg: I think we have doubled up test coverage here between this and the test in `FormManagementPreferencesFragmentTest`. More than happy to discuss different approaches (here or on Slack), as obviously there are no right answers here, but I'd suggest removing this test and sticking with the coverage at the Fragment level for this feature.

seadowg: As a side note, I would often start with a test like this for a feature and then if I discover I can write a more focused test that covers all the same bases at a lower level later, I'd get rid of my high level one. I've found I've been doing that more in Collect than I would have in other apps (where I might just have kept both tests) as the app has so many features that the sheer number of tests is hard to keep under control.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormManagementPage.java
TomRCummings: I was yes, but ended up not. Do you think it is still worthwhile to keep to use for testing in the future? Or get rid of it since we got rid of the `MainMenuPage#enableAutoSend` test?

collect_app/src/androidTest/java/org/odk/collect/android/feature/instancemanagement/AutoSendTest.java
TomRCummings: This makes sense, especially alongside changing the other test to change the Auto send preference using the Android preference component.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormManagementPage.java
seadowg: Good question. I guess it might come in handy some day! Given it's test code not app code it's not going to contribute to the binary size or anything so nice to have around.

seadowg (approved):
Looks ace! Thanks for the working on the suggested changes.

seadowg:
> Is it necessary to call cancelSubmit if a submission has been scheduled but not yet completed and the user toggles Auto-send to something different?

I think it's fine not to. Here's my thinking: there is a case where the user has finished filling in forms while offline, and then they switch off autosend, come back online and the autosend task runs (because it was scheduled before we disabled autosend) and sends forms. I think it's hard to say whether the enumerator would expect that or not, but it feels worse to me to not send forms if they expect them to send than the other way around. Does that make sense?

TomRCummings:
@seadowg Since writing that, I've had a closer look at `AutoSendTaskSpec#getTask`. In the case you outline, if the user has their general autosend preference set to "off", even if a submission was scheduled while the preference was set to something different, when the user comes back online the form will not send since `AutoSendTaskSpec#networkTypeMatchesAutoSendSetting` checks the current value of autosend, not the value of the preference when the submission was scheduled. I could be totally wrong about that!

seadowg:
@TomRCummings no you're totally right! I always forget that the check gets made again. It's all a little confusing, as really that check should happen in `WorkManager` (we can tell it to only run under certain network conditions). That code is a lot older than `WorkManager` though. I think we're all good as there won't be a behaviour change here when disabling auto send (just when enabling it).  

Thanks for working on this! Let me know (or post on Slack) if you'd like more issues to work on.

kkrawczyk123:
**Tested with success!**

Tested scenarios:
- wifi, cellular, both
- different servers: Aggregate, Central, Ona, Kobo, Google drive
- autosend in form management
- autosend defined in form
- internet connection turned off
- internet connection restored
- airplane mode
- autosend when app working in background

mmarciniak90:
Tested with success also on Android versions 8.1, 9.0, 10.0

grzesiek2010:
Thanks @TomRCummings

PR #4758, opened 2021-07-30
Catch exceptions that might ocurr during updating widgets

### LINKED ISSUE #4750: Catch all errors when recomputing values in a field-list and show a helpful dialog
[This form](https://docs.google.com/spreadsheets/d/1nzMwwHbFuKU9LEiSwCWDi-2_7T_71KFSnihOFI_YZsY/edit#gid=0) has a select minimal with a value that gets used in a calculate for another select that was previously filtered to have no choices. It causes [this crash](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/15e3697cc76c108d13ec3a7303005186?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=61017F8E023900017486E6A412D3A914_1568454006980881959). Because the second select's choices have never been rendered, choices are not attached to the selection.

Somewhat related, [this crash](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/d31d1c268cdbe4b661aa40cced4ca763?time=last-seven-days&versions=v2021.2.0%20(4242)&sessionEventKey=6101100B015A000142AB98915C2DF016_1568442290527831492) can be reproduced by having a calculate on a select that produces a value not related to the choices. e.g. I put "foobar" in the `calculate` for a select that did not include "foobar" in its choices.

Also related to https://github.com/getodk/collect/issues/4170.

People think they can use calculates as dynamic defaults. That will not work and they should not do that. In a way, it's not so bad that it crashes so catastrophically because at least they're not getting bad values which they would if the calculate succeeded. This can be fixed by adding a `trigger`. For example, in the sample form at top, add `${select}` in the trigger column. That is how [dynamic defaults are supposed to be expressed](https://docs.getodk.org/form-logic/#dynamic-defaults-from-form-data).

Could we perhaps catch all of these and show a better error message suggesting there might be a misuse of calculates?

grzesiek2010:
Closes #4750 

As we discussed in the issue there is no easy way to detect various form design issues. Since the whole field-list has been a source of many issues catching all of them and displaying a message seems to be a bit better than just crashing the app.

It's a safe catch that shouldn't affect anything so we can focus on testing the sample form attached to the issue.

No.

### DIFF
--- a/collect_app/src/androidTest/assets/forms/form_design_error.xml
+++ b/collect_app/src/androidTest/assets/forms/form_design_error.xml
@@ -1,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <h:head>
+        <h:title>form_design_error</h:title>
+        <model odk:xforms-version="1.0.0">
+            <instance>
+                <data id="form_design_error">
+                    <group>
+                        <select />
+                        <select2 />
+                    </group>
+                    <meta>
+                        <instanceID />
+                    </meta>
+                </data>
+            </instance>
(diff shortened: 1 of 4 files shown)


lognaturel (approved):

mmarciniak90:
**Tested with success**

Dialog with error is visible and form is closed. The user is moved to the main screen.
Screenshot_20210802-171449

PR #4762, opened 2021-07-30
Show repeat in hierarchy view even if first instance has no relevant content

### LINKED ISSUE #4570: Repeat is not shown in hierarchy view if first instance has no relevant content
Collect v1.30.1

#### Problem description
Repeat is not shown in hierarchy view if first instance has no relevant content

See [this form](https://docs.google.com/spreadsheets/d/1gs1vBCotSUsZ7A3YJdtfs-KOrCcbhyFUOWMPO-lm4p0/edit#gid=0). Select 'beans', see that the repeat is shown in the hierarchy view. Select some other vegetables but not 'beans' and see that the repeat is not shown.

#### Expected behavior
The repeat should be shown even if its first instance has no relevant content. This is related to another issue I know we've discussed where if there are no instances of a repeat yet, it doesn't show in the hierarchy view. It might make sense to fix both and always show a repeat if there is one in the `body`.

dimwight:
Closes #4570
~~Blocked on #4570 discussion~~

Tested manually and with new test added to `FormHierarchyTest` (`empty_first_repeat` is a placeholder name, help needed for a suitable final one).

Trivial fix to ensure that
- repeats appear in hierarchy even if first is empty
- relevant repeats are nonetheless listed correctly 

Shows no regression when tested before and after with `testDebug` and `connectedDebugAndroidTest`.

The issue contains example forms.

No, behaviour is now as expected.


dimwight:
Once again **test_app** seems not to like me for no obvious reason.

dimwight:
Definitely awaiting review, does this reopen it? (Seems not)

seadowg:
@dimwight I've reopened and marked ready for review. When something is ready, you can just hit the "Ready for review" button above the test/quality checks, so we know you're not still working on it.

dimwight:
>  "Ready for review" button above the test/quality checks

Doesn't seem to appear for me, maybe because of failed check which I can't see how to resolve. 

Oops, seem to have closed it again. :(

seadowg:
Blocking this on making sure we fully understand the issue (I don't).

dimwight:
>  fully understand the issue 

Would it help if I spelled out in detail what I think is going on? (I certainly had some difficulty getting my head round it.)

seadowg:
> Would it help if I spelled out in detail what I think is going on? (I certainly had some difficulty getting my head round it.)

Yes that'd be great! Happy to dive in, but if you've already got a good idea of what's happening, it'd be great to hear.

dimwight:
Hi @seadowg, an overview as promised of the issue and my proposed resolution. 

In `FormHierarchyActivity`, when `refreshView` hits `EVENT_REPEAT` it is ready to add a `HierarchyElement` for one of two cases.

- _A_ (the default) if the repeat is the first in its group, display the group details; or
- _B_ (set by `shouldShowRepeatGroupPicker()`) display the repeat in a picker view.

Before proceeding either way, `refreshView` checks if the repeat is empty. If it is,
- in case _A_, the current logic assumes that if the first repeat in a group is empty, any subsequent repeats will be as well so the whole group should be skipped. 
- in case _B_ it should clearly skip the empty repeat which has no content to display or edit. 

Issue #4570 arises from the corner case of _A_ where repeats derive from a multiple select. Even if the first repeat is empty, some of those following will not be - the group should definitely be displayed. 

For this PR the existing logic is refined such that the first repeat is never skipped even when empty, so its group is always displayed. The unlikely case that all repeats in a group are empty simply results in a harmless (and maybe expressive) empty picker.

dimwight:
I've also updated the PR as follows.
- In `FormHierarchyActivity`, `refreshView` has been lightly reworked to make the logic clearer.
- In `FormHierarchyTest`, `empty_first_repeat` (just a placeholder for a more descriptive name) has been extended to check inside the group picker.
- In _empty_first_repeat.xml_ the choice list has been truncated to simplify debugging.

seadowg:
> in case A, the current logic assumes that if the first repeat in a group is empty, any subsequent repeats will be as well so the whole group should be skipped.

What do you mean by "empty" here? No relevant questions?

> Issue #4570 arises from the corner case of A where repeats derive from a multiple select. Even if the first repeat is empty, some of those following will not be - the group should definitely be displayed.

This is the part I'm having trouble understanding. Why does the example form in #4570 (and in this PR) cause that issue? I'm probably missing something obvious, but I couldn't see why selecting 'beans' would result in different behaviour from selecting other vegetables. Is there maybe a simpler/smaller form that could illustrate the problem?

dimwight:
Hi @seadowg, I'm sorry you're finding this issue hard to disentangle, it's trivial on the face of it but quite subtle. 

With a bit of luck you'll find my latest commit of some help. In essence it recapitulates my own investigations and proposed fix, by adding to `FormHierarchyActivity` a temporary `enum Scenarios4570` and associated class constant `SCENARIO_4570`. 

The four members of `Scenarios4570` encapsulate the possible permutations of two flags. 
#### Flag `testWithBeans`
(Where 'beans' is shorthand for 'first repeat with content because first item in select multiple was checked; so that `FormEntryModel` regards its `FormIndex`, other things being equal, as relevant')

This is tested in `FormHierarchyTest` in the method now renamed to `showsRepeatsWhenFirstIsEmpty` (hitherto with the placeholder name 'empty_first_repeat'). It defines which of two forms to open: _empty_first_repeat.xml_ to test the issue as defined in #4570 , or _empty_first_repeat_beans.xml_ (somehat confusingly named) for scenarios expecting a non-empty first repeat. The only difference between the two is at L10 between `<veg_type>2 3 5</veg_type>` and `<veg_type>1 2 3 5</veg_type>`. 

Based on the state of `testWithBeans` the test then adjusts its expectations of what `FormHierarchyTest` should display given the form chosen. 

Incidentally, I use the term 'empty' in preference to 'non-relevant' because it describes the repeat so neatly ie in XML `<veg_repeat/>` as opposed to a 'non-empty' repeat such as:

```
    <veg_repeat>
        <veg_group>
            <veg_name>Beans</veg_name>
           <veg_often />
        </veg_group>
    </veg_repeat>
```

#### Flag `restoreOldLogic`
This is tested in `FormHierarchyActivity` at L624, and when set negates the effect of my proposed fix.
 
By setting the `values()` index for `SCENARIO_4570` you can test all permutations of the two flags, with the following results:

0. `WITH_BEANS_OLD_LOGIC` will obviously pass as its first repeat will be non-empty
0. `WITHOUT_BEANS_OLD_LOGIC` won't pass because of its empty first repeat - the issue as raised by @lognaturel  in #4570
0. `WITH_BEANS_NEW_LOGIC` will pass as before
0. `WITHOUT_BEANS_NEW_LOGIC` will also pass despite having an empty first repeat

You can observe all this happening by stepping through the relevant code. In `FormHierarchyActivity` you will note the importance of testing `shouldShowRepeatGroupPicker` to ensure that repeat items appear in a picker. 

Once again rather a long explanation, I hope it helps.

dimwight:
And to explicitly answer your other questions as well:

> ...why selecting 'beans' would result in different behaviour from selecting other vegetables. 

_Beans_ is special because it's the _first_ item in the _veg_type_ choices; if it's not selected the first repeat is empty. And it's the first repeat that triggers the Frequencies repeat header in the hierarchy view: under the existing logic it doesn't if empty and therefore non-relevant (hence the issue), with my fix it always does.

Finally

>Is there maybe a simpler/smaller form that could illustrate the problem?

I do think this is the minimal form to demonstrate the issue.

seadowg (changes requested):
As per our discussion on Slack, next steps are:

1. Remove the `SCENARIO_4570` debug code
1. Use the simpler form (posted in the issue) for the test rather than the veg form

dimwight:
>     1. Remove the `SCENARIO_4570` debug code 
Done in 'Removed debug code'
>     2. Use the simpler form ... for the test rather than the veg form
Done in 'Updated test with simpler form'. 

Also checked that the test fails appropriately when in `FormHierarchyActivity` at L609ff the code is tweaked to
```
boolean asOldCode = true;
if (isEmpty && asOldCode ||
        // For #4570
        (!isFirst || forPicker)) {
    break;
}
```

dimwight:
And then **test_app** failed code that passed when previously committed!

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -596,20 +602,24 @@ private void refreshView(boolean isGoingUp) {
                         break;
                     }
                     case FormEntryController.EVENT_REPEAT: {
-                        if (!formController.isGroupRelevant()) {
+                        FormEntryCaption fc = formController.getCaptionPrompt();
+                        boolean isFirst = fc.getMultiplicity() == 0;
+                        boolean forPicker = shouldShowRepeatGroupPicker();
+                        boolean isEmpty = !formController.isGroupRelevant();
+                        if (isEmpty &&
```
seadowg: I'm actually wondering if this could be `isEmpty && forPicker` instead? That also seems to work. I think we now only want to avoid the next piece of code in the case where we're avoiding rendering an "empty" group in the picker. I wouldn't be surprised if I've missed a case that you'd thought about, and if that is true, let's discuss how we get a test in to enforce this implementation.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -596,20 +602,24 @@ private void refreshView(boolean isGoingUp) {
...
+                                // For #4570
```
seadowg: I think if we have a comment here, we'd want to stick it above the `if` so it's not interrupting the logic. Generally we wouldn't bother just linking to an issue as we'll hopefully get that through git history anyway!

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormHierarchyTest.java
```diff
@@ -30,6 +30,7 @@
             .around(new CopyFormRule("formHierarchy2.xml", null))
             .around(new CopyFormRule("formHierarchy3.xml", null))
             .around(new CopyFormRule("repeat_group_new.xml", null))
+            .around(new CopyFormRule("Empty First Repeat.xml", null))
```
seadowg: Just a note to myself to rework this test file to remove `CopyFormRule` before merging this. I won't make you slog through that @dimwight.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -596,20 +602,24 @@ private void refreshView(boolean isGoingUp) {
                         break;
                     }
                     case FormEntryController.EVENT_REPEAT: {
-                        if (!formController.isGroupRelevant()) {
+                        FormEntryCaption fc = formController.getCaptionPrompt();
+                        boolean isFirst = fc.getMultiplicity() == 0;
```
seadowg: Definitely more readable to pull these all up to the top and name them. Thanks!

seadowg (changes requested):
Thanks again for wrapping your head around this and then explaining it to me. There are a couple of tweaks I'd like to see, and I've also thought about a slightly different implementation that'd be good to discuss.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
dimwight: You're absolutely right, `isFirst` is unnecessary - it's tested further down so that non-empty repeats after the first are skipped anyway. I felt my updated logic was a bit fishy but couldn't spot exactly how.

To check this I've temporarily reinstated my original form and test (which initially caused difficulties - see below). I've simplified the code to match your suggestion, and both tests still pass. Tweak it further at L608 to restore the current logic, and both fail. 

The problem with the reinstatement seemed to be that the two test forms had different names but the same ID, breaking the whole test class in random ways. Adding __A_ to the ID of my form certainly fixed everything definitively.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
dimwight: I actually put it there specifically to emphasise the updated logic, but happy to move it or indeed leave it out. Though I do find comments like this helpful when reading code, as quick hints that there's been an update and why.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormHierarchyTest.java
dimwight: Probably a good thing as my code here is pure copy and paste, I've no idea as to how it works.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
dimwight: Of course two of them are now unnecessary!

seadowg:
@dimwight is this ready for another review? You can request that with the 🔃 icon next to my name in "Reviewers".

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormHierarchyTest.java
```diff
@@ -170,4 +172,27 @@ public void deletingLastGroupAndAddingOneShouldNotBreakHierarchy() {
...
+                .clickOnAdd(new FormEntryPage("Empty First Repeat"))
+                .answerQuestion("Question in repeat", "Not empty!")
+                .clickGoToArrow()
+                .clickGoUpIcon()
+                .clickGoUpIcon()
+                .assertText("Repeat", "Repeatable Group");
+    }
+
+    @Test
+    public void empty_first_repeat() {
```
seadowg: I don't think we need both tests now. As far as I can see `showRepeatsPickerWhenFirstRepeatIsEmpty()` enforces the correct code on its own.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -596,20 +602,24 @@ private void refreshView(boolean isGoingUp) {
                         break;
                     }
                     case FormEntryController.EVENT_REPEAT: {
-                        if (!formController.isGroupRelevant()) {
+                        boolean forPicker = shouldShowRepeatGroupPicker();
+                        // For #4570
+                        if (!formController.isGroupRelevant()
+                                && forPicker // Comment out this line to restore old logic
```
seadowg: Instead of referencing the issue and having this comment about deleting this line I think a comment that explains this behaviour might be more useful to someone looking at this code for the first time. Something like "Don't show repeat group in picker view if it's not relevant".

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormHierarchyTest.java
dimwight: Definitely, I only put the earlier one back as a check.

collect_app/src/main/java/org/odk/collect/android/activities/FormHierarchyActivity.java
dimwight: The idea of commenting out the line clearly had to go, it 
was just a development check. I've tweaked your suggestion to reflect the original motivation.

seadowg (approved):
I'm going to do a little work to rewrite a test file here in a newer style, but this is ready for QA I think! I think testing repeats in general, especially with relevance would be good in addition to testing the issue at hand.

dimwight:
Excellent, thanks @seadowg for your help!

kkrawczyk123:
**Tested with success!**
Tested on Androids: 5.1 and 11.

Verified cases:
- confirmed that the original issue has been fixed (both forms attached in issue)
- regression checks on other forms with repeats: add, remove group, jump to hierarchy, nested repeats, edit group, screen rotation

mmarciniak90:
Successfully tested on Android versions: 8.1, 9.0, 10.0

lognaturel:
Thank you for taking the time to work through all that, @dimwight and @seadowg! That is a very nice improvement.

PR #4769, opened 2021-08-02
Content Provider analytics

### LINKED ISSUE #4576: Add analytics to Content Providers
Blocked by #4462 

We want to be able to find out if any apps are using our Content Providers for forms or instances.

## Acceptance

- [x] Given there is a call to either `insert`, `update` or `delete` on the `FormsProvider`
When I navigate to the analytics dashboard
Then I see there has been an event named `"FormsProvider#<method>`

- [x] Given there is a call to either `insert`, `update` or `delete` on the `InstanceProvider`
When I navigate to the analytics dashboard
Then I see there has been an event named `"InsanceProvider#<method>`

- [x] Given there is a call to `FormsProvider#query` from outside Collect
When I navigate to the analytics dashboard
Then I see there has been an event named `"FormsProvider#query`

- [x] Given there is a call to `InstanceProvider#query` from outside Collect
When I navigate to the analytics dashboard
Then I see there has been an event named `"InstanceProvider#query`

## Note

The URIs we use will be different for internal and external apps as internal URIs will need to include a project in some way whereas external ones will not (they use old URIs and the "first" project is assumed to be being referred to). This will let us differentiate between internal (from `CursorLoader`s etc) and external calls.

Every event should include server hash as well (using `Analytics#logServerEvent`).

seadowg:
Closes #4576

I also reworked the analytics interface a little so it doesn't include anything around servers and forms. Logic around that go moved to `AnalyticsUtils`.

Ran tests.

Only thing of note here is that I needed a way to differentiate between internal and external query calls. To do this, I made `CursorLoaderFactory` add a query param to the URI it sends so the content provider can identify calls that it's responsible for.

Shouldn't be any risk here really, as it's just adding analytics. 

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
@@ -25,6 +25,7 @@ import androidx.annotation.NonNull;
 import org.jetbrains.annotations.NotNull;
 import org.odk.collect.android.analytics.AnalyticsEvents;
 import org.odk.collect.android.analytics.AnalyticsUtils;
+import org.odk.collect.android.dao.CursorLoaderFactory;
 import org.odk.collect.android.database.forms.DatabaseFormsRepository;
 import org.odk.collect.android.formmanagement.FormDeleter;
 import org.odk.collect.android.injection.DaggerUtils;
@@ -113,6 +114,11 @@ public class FormsProvider extends ContentProvider {
         deferDaggerInit();
 
         String projectId = getProjectId(uri);
+
+        // We only want to log external calls to the content provider
+        if (uri.getQueryParameter(CursorLoaderFactory.INTERNAL_QUERY_PARAM) != null) {
+            logServerEvent(projectId, AnalyticsEvents.FORMS_PROVIDER_QUERY);
+        }
 
         Cursor cursor;
         switch (URI_MATCHER.match(uri)) {
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
@@ -23,10 +23,14 @@ import android.net.Uri;
 import androidx.annotation.NonNull;
 
 import org.jetbrains.annotations.NotNull;
+import org.odk.collect.analytics.Analytics;
+import org.odk.collect.android.analytics.AnalyticsEvents;
+import org.odk.collect.android.analytics.AnalyticsUtils;
 import org.odk.collect.android.database.forms.DatabaseFormsRepository;
 import org.odk.collect.android.formmanagement.FormDeleter;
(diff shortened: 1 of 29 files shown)


collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
```diff
@@ -108,6 +115,11 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
 
         String projectId = getProjectId(uri);
 
+        // We only want to log external calls to the content provider
+        if (uri.getQueryParameter(CursorLoaderFactory.INTERNAL_QUERY_PARAM) != null) {
```
grzesiek2010: If you want to log external calls it should be == null right?

collect_app/src/main/java/org/odk/collect/android/external/InstanceProvider.java
```diff
@@ -82,6 +89,11 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
 
         String projectId = getProjectId(uri);
 
+        // We only want to log external calls to the content provider
+        if (uri.getQueryParameter(CursorLoaderFactory.INTERNAL_QUERY_PARAM) != null) {
```
grzesiek2010: The same as above, should be == null.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/external/FormsProvider.java
seadowg: 😔 yes. That's a silly mistake!

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -174,36 +172,26 @@ public void openHierarchy() {
...
+        AnalyticsUtils.logFormEvent(event);
     }
 
     public NonNullLiveData<Boolean> hasBackgroundRecording() {
         return hasBackgroundRecording;
     }
 
-    private String getFormIdentifierHash() {
-        if (formController != null) {
-            return formController.getCurrentFormIdentifierHash();
```
grzesiek2010, line 186: I think this method in `FormController` is no longer used.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Ah yeah good spot

grzesiek2010 (approved):
I think we should use the same way of logging events across the app because now sometimes it's `Analytics.log` (static methods) sometimes we use objects generated by dagger, but it's a thing for another pr.

PR #4770, opened 2021-08-02
Recreate fast external itemsets

### LINKED ISSUE #4767: "Fast external itemsets" don't work after upgrade
Collect v2021.2

#### Problem description
https://forum.getodk.org/t/problem-on-external-choices-after-application-update/34546

1. Download a form with a "fast external itemset" in a prior version
2. Upgrade to v2021.2

#### Expected behavior
Choices are visible.

#### Other information 
This is some kind of path issue. If we'd caught this before release, the answer might have been to remove and rebuild like we did with the form cache. But now that the release has been out for a few days, I think we should consider something like try rebuilding the table if the path is not found.

grzesiek2010:
Closes #4767 

I reproduced the issue and confirmed the fix works as expected.

As I said in the issue:

> I think the easiest and safest solution for now would be bumping the database version so that `onUpgrade()` is called which clears the database so that csv files are loaded again https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/fastexternalitemset/ItemsetDbAdapter.java#L57

> a bit better long term solution would be to check if the table we try to read from exist during form loading. If not the files should be loaded again. This would be better but also it will require more changes and testing so I would rather do that later and for now just add that small fix to help our users ASAP.

We should test the case described in the issue.

I used:
[fasteExternalItemsets.zip](https://github.com/getodk/collect/files/6915893/fasteExternalItemsets.zip)

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/fastexternalitemset/ItemsetDbAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fastexternalitemset/ItemsetDbAdapter.java
@@ -25,7 +25,7 @@ public class ItemsetDbAdapter implements Closeable {
 
     public static final String DATABASE_NAME = "itemsets.db";
     private static final String DATABASE_TABLE = "itemset_";
-    private static final int DATABASE_VERSION = 2;
+    private static final int DATABASE_VERSION = 3;
 
     private static final String ITEMSET_TABLE = "itemsets";
     public static final String KEY_ITEMSET_HASH = "hash";


lognaturel (approved):
Works great for me. It does mean a little more processing on first open but that’s a small price to pay. And it means those already in a bad state will be able to recover.

mmarciniak90:
**Tested with success**

Verified on Android 5.1, 8.1, 9.0, 10.0, 11.0

Verified cases:
* save filled forms on v1.30.1 -> upgrade to version with a fix
* save filled forms on v1.30.1 -> upgrade to 2021.2.1 -> upgrade to version with a fix
* verified form from forum and issue
* saved submissions have all answers
* new submissions work as expected
* loading takes a bit longer

Issue #4756, opened 2021-07-29
Crash trying to connect to db when current project doesn't exist


lognaturel:
#### Software and hardware versions 
Collect v2021.2

#### Problem description
If the project id saved in meta settings corresponds to a project that was deleted from the filesystem, Collect will crash on open. This is likely not a very common operation but given that we now see clear evidence that people are pushing both forms and instances through the filesystem, it wouldn't be surprising if they're also deleting projects that way.

I believe this explains a subset of [the database crashes we see in Crashlytics](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/07cccfa2bbf3b4e0e4fed4f0a9fcbd76?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=61031FAD03BA00010D413B71613F3C76_1568911492518499770).

#### Steps to reproduce the problem
1. Create a project from Collect
2. `adb shell` and delete the project
3. launch Collect

#### Expected behavior
No crash.

lognaturel:
I thought we could fix this by updating the current project based on the filesystem at app launch but I don't think that would be sufficient. It seems that what's actually crashing are the background tasks that keep running: [main menu instance count updater](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/07cccfa2bbf3b4e0e4fed4f0a9fcbd76?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=61031FA300B400010BF23B71613F3C76_1568911450225484665), and I've seen it from match exactly as well though no Crashlytics report that I can see.

I briefly discussed this with @seadowg and he thought this wasn't important to address because users shouldn't expect that they can manipulate the filesystem and get good behavior. I agree it's not a super common use case but on the other hand Collect has had that behavior for over a decade so it's an abrupt change. @seadowg, I now believe that 100% of [this Crashlytics issue](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/07cccfa2bbf3b4e0e4fed4f0a9fcbd76?versions=v2021.2.0%20(4242)&time=last-seven-days&sessionEventKey=61031FAD03BA00010D413B71613F3C76_1568911492518499770) are caused by this situation. That explains why it happens so many times for each user -- I see it over and over again locally as well.

Is there anything lightweight we can do to rescue users from this state?

[...]

PR #4772, opened 2021-08-02
Detect why database can't be accessed on crash

seadowg:
Should provide more info for #4756.

This adds code to the crash site that checks and then throws a more specific exception in the cases where:

* The `projects` dir has been deleted (probably meaning our extrenal files dir has been cleared)
* The current project's dir has been deleted
* The current project's dir has been cleared out (ignores if `.cache` is still there)
* The current project's metadata dir has been deleted

Checked that the cases throw the right exception manually.

I thought this would provide a way to get info on this crash without changing any behaviour which hopefully means we can get it out faster. I also think having a clearer idea of what the users are doing will help us approach a fix with more context - right now we'd probably want to add code to cover all these cases which adds more complexity, tests and risk than we might need to commit to.

This only adds code to a crash site, so shouldn't add any risk. No need for QA here.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
@@ -9,15 +9,20 @@ import androidx.lifecycle.ViewModel;
 import androidx.lifecycle.ViewModelProvider;
 
 import org.jetbrains.annotations.NotNull;
+import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.configure.SettingsUtils;
 import org.odk.collect.android.formmanagement.InstancesAppState;
+import org.odk.collect.android.instancemanagement.InstanceDiskSynchronizer;
 import org.odk.collect.android.preferences.FormUpdateMode;
 import org.odk.collect.android.preferences.keys.ProtectedProjectKeys;
 import org.odk.collect.android.preferences.source.SettingsProvider;
-import org.odk.collect.android.instancemanagement.InstanceDiskSynchronizer;
+import org.odk.collect.android.storage.StoragePathProvider;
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
```diff
@@ -118,13 +123,52 @@ private String appendToCommitDescription(String commitDescription, String part)
     public void refreshInstances() {
         scheduler.immediate(() -> {
             new InstanceDiskSynchronizer(settingsProvider).doInBackground();
-            instancesAppState.update();
+
+            try {
+                instancesAppState.update();
+            } catch (android.database.sqlite.SQLiteCantOpenDatabaseException e) {
```
grzesiek2010: You should add import for SQLiteCantOpenDatabaseException.

collect_app/src/main/java/org/odk/collect/android/activities/viewmodels/MainMenuViewModel.java
```diff
@@ -118,13 +123,52 @@ private String appendToCommitDescription(String commitDescription, String part)
...
         }, ignored -> {
         });
 
     }
 
+    /**
+     * Work out what has happened to make the database go missing and then throw the appropriate
+     * exception. Here to determine the cause of https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/07cccfa2bbf3b4e0e4fed4f0a9fcbd76
+     */
+    private void determineWhyDatabaseCannotBeAccessed(android.database.sqlite.SQLiteCantOpenDatabaseException e) {
```
grzesiek2010: As above it should be imported.

grzesiek2010 (changes requested):

seadowg:
Closing this as it shouldn't be needed after #4775.

PR #4777, opened 2021-08-05
Create directories on demand rather than up front

seadowg:
This changes the way we manage directories in Collect so that instead of needing to create them upfront, we just create them whenever we need them.

In the process of doing this I also cut down the `StoragePathProvider`'s API to make it a little cleaner. This is something that had been tempting for a while - mainly since we pulled the helpers responsible for calculating absolute and relative paths out.

New tests to check `StoragePathProvider` changes, ran existing tests, played with app.

We've seen a lot of crashes recently that have to do with users being able to delete directories that we need. Creating them on the fly seems like a great way to counter this as it means we don't have to worry about things getting deleted or having to create things up front.

This could cause problems for some users as directories won't always be there when they look at the disk anymore. My assumption is that that won't matter for the kind of user that would be copying files in with ADB but I could totally be wrong about that!

It would be good to do a general pass of Collect (managing forms, form entry etc) as this change does touch a pretty core part of our code.


grzesiek2010 (approved):

seadowg:
@grzesiek2010 do you think users might get confused that `forms`/`instances` might not exist for their project yet? We could just add a note to https://docs.getodk.org/collect-adb/ maybe?

grzesiek2010:
> @grzesiek2010 do you think users might get confused that forms/instances might not exist for their project yet? We could just add a note to https://docs.getodk.org/collect-adb/ maybe?

It might be a small problem for people who just install the app and immediately try to push forms via adb when the directory is not created yet but it does not worry me. I think they will figure it out. But yeah we can have a short note in the docs.

mmarciniak90:
@seadowg, is it expected that `metadata` and `instances` directories are created immediately when a project is created?
I also noticed that resetting all available options causes that all directories are created. What do you think about this behavior?

seadowg:
@mmarciniak90 they now get created where anything in the app references the directories, so I guess metadata and instances get created because we do the instance disk sync when the user views the main menu. Resetting creating the directories is weird, but it does make sense since we delete the contents of each directory (so create the directories if they don't exist when looking up where they are).

mmarciniak90:
**Tested with success**

Verified cases:
- `metadata` and `instances` directories are created when a project is created
- `forms` directory is created when user opens `Fill blank form`
- `forms` directory is created when user opens `Get Blank Form`
- `forms` directory is created when `Match exactly` OR `Previously downloaded` is turned on 
- `forms` directory is created when user opens the Blank Form tab in `Delete saved forms`
- `layers` directory is created when user opens `Maps` settings
- `layers` directory is created when user opens Map view (from Fill blank form view, from filled form)
- `settings` directory is created when user opens `Reconfigure with QR code`
- reset of all available options causes that all directories are created 
- deleted project
- form shortcut

kkrawczyk123:
Verified with success also on Androids 5.1 and 11.

PR #4781, opened 2021-08-10
Use Intent.ACTION_GET_CONTENT instead of Intent.ACTION_PICK to pick an image with settings

### LINKED ISSUE #4776: picking an image with settings might be spoiled
Collect all versions

#### Problem description
I discovered that on my emulator I can't pick a file with settings.
After a quick investigation the problem seems to steam from the fact that we use `ACTION_PICK` building an intent to trigger this action. Looks like `ACTION_GET_CONTENT` should be used instead https://developer.android.com/reference/android/content/Intent?authuser=1#ACTION_GET_CONTENT.

| ACTION_PICK  | ACTION_GET_CONTENT |
| Screenshot_1628005061  | Screenshot_1628005103  |

grzesiek2010:
Closes #4776

I tested picking images with settings.

According to the docs https://developer.android.com/reference/android/content/Intent#ACTION_GET_CONTENT we should use 
`Intent.ACTION_GET_CONTENT` instead of `Intent.ACTION_PICK`. I also confirmed that the behavior is better if we use `Intent.ACTION_PICK`.
We also discussed the issue on slack https://getodk.slack.com/archives/C35ENHA64/p1628005284054700

We should test picking images with settings, not importing just picking.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/configure/qr/QRCodeMenuDelegate.java
+++ b/collect_app/src/main/java/org/odk/collect/android/configure/qr/QRCodeMenuDelegate.java
@@ -55,7 +55,7 @@ public class QRCodeMenuDelegate implements MenuDelegate {
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case R.id.menu_item_scan_sd_card:
-                Intent photoPickerIntent = new Intent(Intent.ACTION_PICK);
+                Intent photoPickerIntent = new Intent(Intent.ACTION_GET_CONTENT);
                 photoPickerIntent.setType("image/*");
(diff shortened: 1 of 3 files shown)


seadowg (approved):
Looks good!

mmarciniak90:
**Tested with success**

Verified cases:
* new options to pick QR code: Images, Recent, Downloads, Drive, Photos, TotalCommander
* navigated deep into folders
* loaded QR code on first launch screen
* loaded QR code for new project
* loaded QR code to reconfigure current project
* regression check of reverted QR code
* regression check of invalid QR code and random image
* the same QR code loaded two times

kkrawczyk123:
Tested with success also on Androids 5.1 and 11.

PR #4782, opened 2021-08-10
Update Gradle plugin

seadowg:
This just upgrades the Android Gradle plugin to allow us to use Android Studio Arctic Fox.

### DIFF
--- a/analytics/build.gradle
+++ b/analytics/build.gradle
@@ -11,8 +11,6 @@ android {
     defaultConfig {
         minSdkVersion rootProject.androidMinSdk
         targetSdkVersion rootProject.androidTargetSdk
-        versionCode 1
-        versionName "1.0"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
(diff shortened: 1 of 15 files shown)


analytics/build.gradle
```diff
@@ -11,8 +11,6 @@ android {
     defaultConfig {
         minSdkVersion rootProject.androidMinSdk
         targetSdkVersion rootProject.androidTargetSdk
-        versionCode 1
```
seadowg, line 14: These were automatically remove by the migration, as I guess they aren't needed for libraries any longer (which makes sense).

config/lint.xml
```diff
@@ -59,4 +59,10 @@
     <issue id="AlwaysShowAction" severity="ignore" />
 
     <issue id="InvalidFragmentVersionForActivityResult" severity="ignore" />
-</lint>
\ No newline at end of file
+
+    <!-- This incorrectly identifies places where String.format is used on the string: https://github.com/JakeWharton/timber/issues/434 -->
+    <issue id="TimberArgCount" severity="ignore" />
```
seadowg, line 64: This and the lint below both failed after the upgrade, although both seem to be false positives. I've filed a bug for one and referenced an existing one for the other.

grzesiek2010 (approved):

PR #4788, opened 2021-08-12
Implemented new way of managing dependencies and updated them

### LINKED ISSUE #4691: Investigate the best way to manage dependency versions across modules
Right now we have a bunch of dependencies shared among modules and have been using version constants to keep them in sync where we can. Is there a better way to do this? 

If not, we should sweep the modules we have so far and try and pull out as many constants as we can.

grzesiek2010:
Closes #4691

I ran automated tests.

We considered options described in https://proandroiddev.com/how-to-manage-dependencies-in-multi-module-project-84620afbb415 and decided that the one with buildSrc module is the best one.

It doesn't require testing. As always we assume that dependency updates will be tested during our every day work.

No.

No.


config/pmd-ruleset.xml
```diff
@@ -17,6 +17,7 @@
         <exclude name="AvoidReassigningParameters" />
         <exclude name="AvoidStringBufferField" />
         <exclude name="ForLoopCanBeForeach" />
+        <exclude name="GuardLogStatement" />
```
grzesiek2010, line 20: I will review those new pmd rules and enable later. The same with checkstyle.

analytics/build.gradle
```diff
@@ -32,9 +35,8 @@ android {
 }
 
 dependencies {
-
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation "androidx.core:core-ktx:${rootProject.coreKtxVersion}"
-    implementation 'com.google.firebase:firebase-analytics:18.0.1'
-    implementation 'com.google.firebase:firebase-crashlytics:17.3.0'
+    implementation Dependencies.kotlin_stdlib
```
seadowg, line 38: Wow this is so much nicer!

seadowg, line 38: Wow this is so much nicer!

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -0,0 +1,99 @@
...
+    const val kotlin_stdlib_jdk7 = "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${Versions.kotlin}"
+    const val json = "org.json:json:20210307"
+    const val gson = "com.google.code.gson:gson:2.8.7"
+    const val firebase_analytics = "com.google.firebase:firebase-analytics:19.0.0"
+    const val firebase_crashlytics = "com.google.firebase:firebase-crashlytics:18.2.1"
+    const val findbugs_annotations = "com.google.code.findbugs:annotations:3.0.1"
+    const val findbugs_jsr305 = "com.google.code.findbugs:jsr305:3.0.2"
+    const val fastlane_screengrab = "tools.fastlane:screengrab:2.1.0"
+    const val leakcanary = "com.squareup.leakcanary:leakcanary-android:2.7"
+    const val timber = "com.jakewharton.timber:timber:${Versions.timber}"
```
seadowg: I don't think Timber needs its own `Version` constant as it's only used once? Also, Timber was just updated to fix a bug we ran into (https://github.com/JakeWharton/timber/issues/434#event-5155819740). Lets update here and remove the ignore in `lint.xml` on `TimberArgCount`.

config/pmd-ruleset.xml
seadowg: Can we create an issue for that?

seadowg (changes requested):
This is so nice! One small change to make while we're here, I think.

config/pmd-ruleset.xml
grzesiek2010: https://github.com/getodk/collect/issues/4794

buildSrc/src/main/java/dependencies/Dependencies.kt
grzesiek2010: Done.

seadowg (approved):

PR #4791, opened 2021-08-15
Allow complex emojis in icon

### LINKED ISSUE #4790: Support complex Emojis as icon in Collect
Collect v2021.2.3

#### Problem description
The multiple project feature is brilliant. Currently, the Icon can handle one character which is great for emojis. But it would be good to support complex emojis like the Argentina Flag (:argentina:) which combines 2 unicode characters.
Just try to use comple Emoji like  the Argentina Flag (:argentina:) 
#### Expected behavior
Allow only one character or onlu one Emoji but not check the size of the content
#### Other information 
To fix this I propose to use emoji-java (https://github.com/vdurmont/emoji-java) which is a well-developed Java library for working with Emojis. The idea of having an icon is brilliant and my approach would:
For changing the icon in Collect
- Check the text and separate the text from the emojis with emoji-java
- If the text input contains only text then limit the number of characters to 1
- If the text input contains only emojis hen limit the number of emojis to 1 using emoji-java
- If the text input contains a character and the user selects an emoji then replace the character with the emoji.
When reading a QR code with an icon:
- Check the icon part of the QR and separate the text from the emojis with emoji-java
- If the icon contains only text then use the first character uppercased
- if the icon contains only emojis then use the first emoji using emoji-java
- if the icon contains text and emojis then use the first emoji using emoji-java

ghost:
Closes #4790

I build the APK locally using Android Studio Bumblebee | 2021.1.1 Canary 8 and I test it on an Android Pixel 4a. Updated tests and ran ./gradlew testDebug
Working with Emojis is complicated. There are combinations of Emojis to create one. Using length() is not an approach. And instead of writing custom complex code to control Emojis it is better to use a Emoji-Java (https://github.com/vdurmont/emoji-java)
Users will be able to use complex Emojis. For example (family: man, woman, girl, boy) (1F468 200D 1F469 200D 1F467 200D 1F466) 👨‍👩‍👧‍👦 . No regression risks because we only allowing 1 character or 1 complex emoji
No. This change applies to the interface. I tested with different cases:
For changing the icon in Collect
- If the text input contains only text then limit the number of characters to 1
- If the text input contains only emojis hen limit the number of emojis to 1 using emoji-java
- If the text input contains a character and the user selects an emoji then replace the character with the emoji.
When reading a QR code with an icon:
- If the icon contains only text then use the first character uppercased
- If the icon contains only emojis then use the first emoji using emoji-java
- If the icon contains text and emojis then use the first emoji using emoji-java
I don't think so because icon is not currently documented in https://docs.getodk.org/

### DIFF
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -47,7 +47,9 @@ android {
 
 dependencies {
     //Handle Emojis in project settings
-    implementation "com.vdurmont:emoji-java:5.1.1"
+    implementation ("com.vdurmont:emoji-java:5.1.1") {
+        exclude group: 'org.json', module: 'json'
+    }
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     implementation "androidx.core:core-ktx:${rootProject.coreKtxVersion}"
     implementation "androidx.lifecycle:lifecycle-livedata-ktx:2.3.0"
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -46,6 +46,8 @@ android {
 }
 
 dependencies {
+    //Handle Emojis in project settings
+    implementation "com.vdurmont:emoji-java:5.1.1"
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     implementation "androidx.core:core-ktx:${rootProject.coreKtxVersion}"
     implementation "androidx.lifecycle:lifecycle-livedata-ktx:2.3.0"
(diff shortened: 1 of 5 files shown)


seadowg (approved):
There's a couple of tweaks I'd like to make here, but the approach looks good, and we'd like to include this in our next hotfix, so no need to hold this up. I'm going to make some quick changes to have this point at our hotfix branch, but then will mark this ready for QA.

Thanks for taking care of this @qlands!

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1, 7.1, 10 and 11.

Verified scenarios:
- add icon from devices keyboard
- edit icon
- cancel changes 
- try to add more than 1 icon
- try to input text when icon added
- try to add icon when text already added
- copy and paste icon
- scan qr code when icon has been used 
- import qr code from file

PR #4793, opened 2021-08-16
Trigger reloading property manager every time a form is loaded

### LINKED ISSUE #4792: Username isn't updated correctly when switching projects
Collect v2021.2

#### Problem description
When switching between projects, the username preload doesn't get updated. I haven't tried it but I assume that the same problem exists with phone number and email.

1. Create a form that accesses the username such as [this one](https://docs.google.com/spreadsheets/d/15Tp1CauiRuTgudfZ6OClGamM5q9ezV2G-wg1AoF8PQQ/edit#gid=0)
2. Create multiple Collect projects that access this form with and without usernames (e.g. from App Users `Foo` and `Username test` at https://test.getodk.cloud/#/projects/149/app-users)
3. Switch between projects and see that the username of whichever project was created last gets used
4. Close the form, switch to a project that has a username other than the one that was being used, kill Collect, reopen it, fill the form and see that the expected username is now used.

#### Expected behavior
Switching projects should reset username, phone number and email preloads.

#### Other information 
This is likely a matter of reloading some things from `PropertyManager` on project switch. It would be good to think through anything else that might not be reset.

grzesiek2010:
Closes #4792

I tested the fix manually.

It's the easiest issue. In the future we should consider loading those properties dynamically.

It should just fix the issue. Nothing else should be affected.

[metadata.xlsx](https://github.com/getodk/collect/files/6999022/metadata.xlsx)

No.


seadowg (dismissed):

seadowg (changes requested):
This should target the `v2021.2.x` branch

kkrawczyk123:
**Tested with success!**

Tested cases:
- username
- phone number
- email address
- username set in server settings
- username set in both places 
- create project from the same project in Central and other users
- create different server projects

mmarciniak90:
Verified also on Android versions: 8.1, 9.0, 10.0

Issue #4059, opened 2020-09-01
First nested repeat not created automatically after first outer repeat


seadowg:
Originally from https://forum.getodk.org/t/nested-repeatable-groups-only-allowing-child-repeatable-groups-for-first-record-in-odkcollect/29599

#### Software and hardware versions
 
Collect v1.27.3

#### Problem description

It seems for some reason that nested repeats are not displayed in the hierarchy view (as a group) until the user has created one of the nested repeats. This problem does not occur for the first outer repeat group where it seems the first nested repeat is automatically created (which id default behaviour and expected). This problem (that nested repeats aren't created automatically) seems to be the root of the problem which seems to contradict the default behaviour (https://xlsform.org/en/#representing-zero-repeats).

#### Steps to reproduce the problem

Using this form:

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <h:head>
        <h:title>One Question Nested Repeat</h:title>
        <model>
            <instance>
                <data id="one_question_repeat">
                    <person>
                        <age/>
                        <child>
                            <name/>
                        </child>
                    </person>
                </data>
            </instance>
            <bind nodeset="age" type="int"/>
            <bind nodeset="name" type="string"/>
        </model>
    </h:head>
    <h:body>
        <group ref="/data/person">
            <label>Person</label>
            <repeat nodeset="/data/person">
                <input ref="/data/person/age">
                    <label>What is their age?</label>
                </input>
                <group ref="/data/person/child">
                    <label>Child</label>
                    <repeat nodeset="/data/person/child">
                        <input ref="/data/person/child/name">
                            <label>What is their name?</label>
                        </input>
                    </repeat>
                </group>
            </repeat>
        </group>
    </h:body>
</h:html>
```

1. Open the form
2. Add a new person with the plus button (➕) or by swiping through.
3. Swipe to the next question

You are prompted to add a new "Child".

#### Expected behavior

The first "Child" (for every "Person") should already be created and shouldn't prompt the user.

#### Other information 

Using the example form this Espresso test will fail:

```java
new MainMenuPage(rule)
                .startBlankForm("One Question Nested Repeat")
[...]

PR #4712, opened 2021-07-21
Resolves #4500 autocomplete minimal…

dimwight:
Resolves #4500

This documents a revision of #4673 now closed.

#### What has been done to verify that this works as intended?
Tested with new `SelectOneResetTest`.

#### Why is this the best possible solution? Were any other approaches considered?
See design notes below.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
Improvement to behaviour from user perspective. 
No apparent regression risks. 

#### Do we need any specific form for testing your changes? If so, please attach one.
_[selectOneReset.zip](https://github.com/getodk/collect/files/6854731/selectOneReset.zip)_

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/getodk/docs/issues/new) and include the link below.
Perhaps form design advice could specify the benefits and limitations of the improved behaviour?

#### Before submitting this PR, please make sure you have:
- [x ] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x ] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/CONTRIBUTING.md#ui-components-style-guidelines)

-----
# Design notes
A lot of stuff here, but both [the issue](https://github.com/getodk/collect/issues/4500) and the proposed resolution have turned out to be quite complex. 

## The issue
A shout out to @aurdipas for a [form](https://github.com/getodk/collect/files/6246387/Entrevista_Nacimientos_Etnicos_filtro.xlsx) that demonstrates clearly the undesired behaviour of selects with fast external itemsets. 
- Most obviously, answers for cascades using fast external itemsets don't reset (or at best not immediately) when preceding members are updated, in constrast to their behaviour with internal itemsets. 

- This behaviour is most glaringly apparent with minimal appearances, especially in field lists - plain select widgets do reset when they can’t match an existing value in their current choices.

Desirable if less obvious behaviours also implied by the form are that

- Members should reset even when unrelated fields are interpolated in the cascade.
- Members hidden by an interpolated select when non-relevant should be reset once unhidden.

[...]

Issue #645, opened 2017-03-16
Remove deprecated methods from TimeWidget


rajatthakur:
The time widget class contains deprecated methods which need to be replaced.

PR #4797, opened 2021-08-17
Proposal for #4059 First nested repeat not created…

dimwight:
Proposal for resolving #4059 

This is very much a draft PR since it's on the wrong repository. I tried forking Javarosa but can't build it due to Checkstyle's dislike of Windows line endings. 

As a workaround I cloned to Collect the classes that seem to need attention, then committed my proposed changes on top of them. This approach has the merit of being testable within Collect more or less as described in #4059. 

## The issue
Logging the state of the form instance shows that the crucial code is indeed in Javarosa. The initial instance tree is created with its nested repeat(s), but subsequent repeats are not. 

Stepping through creation of a repeat shows that in the template tree acquired in `FormDef.createNewRepeat` the root has the right descendants.
However when called via `FormInstance.copyNode`, `TreeElement.deepCopy` fails to copy these because their multiplicity is `INDEX_TEMPLATE` ie -2; even if it did copy them their multiplicities would stop them being indexable. 

## The fix
The obvious solution is to clone the template tree and set multiplicities to `DEFAULT_MULTIPLICITY` ie 0: this will be correct for nested repeats and the appropriate value for the repeat root is set in `copyNode`.

Both `copyNode` and `deepCopy` have multiple call sites, so modifying either will most likely (certainly?) result in regression.

However in `createNewRepeat` we can pass to `copyNode` a clone of the template with adjusted multiplicities, created using an analogue in `TreeElement` of `deepCopy`. (The subsequent copying in `copyNode` will then be harmlessly redundant.)

This small change has the desired effect and should be safe from regression.

## The code
As noted above
- the first commit clones the two Javarosa classes - incidentally generating some tens of PMD complaints 
- the second makes the proposed changes
- the third comprises the test class and form

## Testing
`NestedRepeatTest` loads the form specified in
_[NestedRepeats.xlsx](https://github.com/getodk/collect/files/7000892/NestedRepeats.xlsx)_, incidentally demonstrating that the fix works for at least one further level of nesting.

### DIFF
--- a/collect_app/src/androidTest/assets/forms/NestedRepeats.xml
+++ b/collect_app/src/androidTest/assets/forms/NestedRepeats.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0"?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa"
+    xmlns:odk="http://www.opendatakit.org/xforms" xmlns="http://www.w3.org/2002/xforms">
+    <h:head>
+        <h:title>NestedRepeats</h:title>
+        <model odk:xforms-version="1.0.0">
+            <instance>
+                <data id="NestedRepeats" version="2021-08-15 B">
+                    <person jr:template="">
+                        <age />
+                        <child jr:template="">
+                            <name />
+                            <pet jr:template="">
+                                <colour />
+                            </pet>
+                        </child>
+                    </person>
+                    <person>
+                        <age />
+                        <child>
+                            <name />
+                            <pet>
+                                <colour />
+                            </pet>
+                        </child>
+                    </person>
+                    <meta>
+                        <instanceID />
+                    </meta>
+                </data>
+            </instance>
+            <bind nodeset="/data/person/age" type="string" />
+            <bind nodeset="/data/person/child/name" type="string" />
+            <bind nodeset="/data/person/child/pet/colour" type="string" />
+            <bind nodeset="/data/meta/instanceID" readonly="true()" type="string"
+                jr:preload="uid" />
+        </model>
+    </h:head>
+    <h:body>
+        <group ref="/data/person">
+            <label>Person</label>
+            <repeat nodeset="/data/person">
+                <input ref="/data/person/age">
+                    <label>age?</label>
+                </input>
+                <group ref="/data/person/child">
+                    <label>Child</label>
+                    <repeat nodeset="/data/person/child">
+                        <input ref="/data/person/child/name">
+                            <label>name?</label>
+                        </input>
+                        <group ref="/data/person/child/pet">
+                            <label>Pet</label>
+                            <repeat nodeset="/data/person/child/pet">
+                                <input ref="/data/person/child/pet/colour">
+                                    <label>colour?</label>
+                                </input>
+                            </repeat>
+                        </group>
+                    </repeat>
+                </group>
+            </repeat>
+        </group>
+    </h:body>
+</h:html>
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/NestedRepeatTest.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/NestedRepeatTest.java
@@ -1,0 +1,45 @@
+package org.odk.collect.android.feature.formentry;
+
+import android.Manifest;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.rule.GrantPermissionRule;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.runner.RunWith;
+import org.odk.collect.android.support.CollectTestRule;
+import org.odk.collect.android.support.CopyFormRule;
+import org.odk.collect.android.support.ResetStateRule;
+import org.odk.collect.android.support.pages.FormEntryPage;
+import org.odk.collect.android.support.pages.MainMenuPage;
+
+@RunWith(AndroidJUnit4.class)
+public class NestedRepeatTest {
+
+    private final CollectTestRule rule = new CollectTestRule();
+
+    @Rule
+    public RuleChain copyFormChain = RuleChain
+            .outerRule(GrantPermissionRule.grant(Manifest.permission.READ_PHONE_STATE))
+            .around(new ResetStateRule())
+            .around(new CopyFormRule("NestedRepeats.xml"))
+            .around(rule);
+
+    @Test
+    public void nestedRepeatsCreatedWithOuterRepeat() {
+        new MainMenuPage()
+                .startBlankForm("NestedRepeats")
+                .assertText("Person > 1")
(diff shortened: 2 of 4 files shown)


dimwight:
Incidentally, being well aware there's a lot going on with Collect at the moment, I'm not expecting review of my offerings anytime soon. 

When someone does have time to look at them, can I suggest in the following order? 
- #4762 is childishly simple
- #4797 (this) is essentially simple but complicated by the repository issue
- #4712 is a bit of a magnum opus, especially the testing

seadowg (changes requested):
It looks like quality checks are failing and need fixed before we look at this. Running `./gradlew checkCode` (or `gradlew.bat checkCode`) will show the report locally.

dimwight:
All checks* seemed ok at my end, but I'll run them again and update as required. 

\* Except of course for the thirty or so PMD fails in the cloned Javarosa code (and another thirty from circleci Checkstyle) . Now updated these - with some misgivings as not my code. 

Which reminds me that of course there's no way I'd want these commits merged.

seadowg:
Marking this as a draft seeing as it's a set of changes for JavaRosa rather than Collect.

seadowg:
@dimwight what errors are you seeing building JavaRosa on your machine? It'd be good to make sure JavaRosa development can happen on Windows as well as *nix.

dimwight:
As mentioned above, 
>  I tried forking Javarosa but can't build it due to Checkstyle's dislike of Windows line endings.

Unlike in Collect's separate `checkCode` task, Checkstyle is run as part of the JavaRosa  `build` task, so on Windows the whole task fails. Is there any way to remove or disable the Checkstyle element (preferably within AS)?

seadowg:
@dimwight I'm wondering if there is a way we could get CheckStyle running on Windows rather than avoiding. What exact error/failure are you seeing?

dimwight:
> get CheckStyle running on Windows 

There's no problem running it, it just fails every time on Windows line endings. `Do not use Windows line endings [RegexpMultiline]`
 In Collect this happens at the end of `checkCode` so it doesn't get in the way. In Javarosa on the other hand it stops `build` in its tracks.

seadowg:
@dimwight is that on a clean check out or only once you've added/changed code? I'm wondering if there is an editor setting that can sort this out.

dimwight:
>  a clean check out or only once you've added/changed code? 

To be exact, 
1. I follow instructions at https://github.com/getodk/javarosa,
to fork and import project to Intellij.
2. I run `gradle build`.
3. The build fails with these messages:

>Execution failed for task ':checkstyleJmh'. Checkstyle rule violations were found. See the report at: file:///...jmh.html Checkstyle files with violations: 18 Checkstyle violations by severity: [error:1984]

All the errors are `Do not use Windows line endings` which is what I get when Checkstyle is run in Collect `checkCode`

To my eye the most straightforward solution would be to omit Checkstyle from the Javarosa `build`, is this not possible?

dimwight:
Further update. Woke up this morning and decided to try the brute force solution of fixing by hand each of the eighteen files complained of by `checkstyleJmh`, which then indeed passed. 

However `checkstyleMain` then failed another 298 files, also for Windows line endings. 

Is there anyone on the Javarosa team I could raise this with, who's an expert on the build files?

seadowg:
@dimwight my best guess here is that git is checking out files with Windows line endings. 

To make your Git config more Unix friendly (probably a good idea for working with any open source project) you can set `core.autocrlf` to `input` to have Git check files out as-is and commit with Unix line endings (more [here](https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings)). Then so that IntelliJ doesn't add windows line endings that will break builds locally, you can switch the Line Separator to "Unix and macOS" (more [here](https://www.jetbrains.com/help/idea/configuring-line-endings-and-line-separators.html)). If that works, we can add that to the JavaRosa (and possibly Collect's) docs. 

Thanks for soldiering through the horrible Unix/Windows incompatibilities!

dimwight:
>  you can set `core.autocrlf` to `input` to have Git check files out as-is and commit with Unix line endings 

Sounds promising, even if it means I have to bite the bullet on the Git CLI.

seadowg:
@dimwight if you're using SourceTree, you can hit "Settings" (top right) then "Advanced" and then "Edit Config File..." to change the Git config being used. That way you can stay away from the CLI I think!

dimwight:
Thanks @seadowg for these suggestions. They were very helpful in that they put me on the right track, though the actual anwer came out different. 

On looking `autocrlf` up in the Git book I found at https://git-scm.com/docs/git-config the following: 

> _core.autocrlf_...Setting this variable to "true" is the same as setting the text attribute to "auto" on all files and core.eol to "crlf"... 
to have CRLF line endings in your working directory... 
can be set to _input_, in which case no output conversion is performed.

So this makes no promises about working files which are what Checkstyle was failing on.  But on the same page is the following:
>_core.eol_...Sets the line ending type to use in the working directory for files that are marked as text... ignored if core.autocrlf is set to true or input

This seemed much more promising and indeed does the trick.

So now I was ready to close this PR with thanks for your assistance, except that as a next stage I wanted to test my updated Javarosa code in Collect.

Unfortunately [these instructions](https://github.com/getodk/collect#debugging-javarosa) tripped me up at step 2 with the issue described [here](https://stackoverflow.com/questions/63227624/import-source-code-module-in-android-studio-is-not-working). In the absence of any error message it does seem likely that I need to prepare my Javarosa folder somehow before trying to import into Collect. 

Any thoughts?

dimwight:
I had a thought - switch to the instructions for building a jar and referencing it in Collect. I built my jar, but didn't succeed in setting the path.
1. To start with I deliberately broke the Collect build by opening _Dependencies.kt_ and changing the value of `javarosa` to the name of my new jar.
1. In _collect_app/build.gradle_, found the `dependencies` section.
1. Took a guess that I needed to update the `implementation(Dependencies.javarosa)` block. 
1. Pasted in the suggested code, deliberately not changing the dummy path and hoping for just a complaint about the bad path.
1. Instead got `Could not find method compile()`. [Apparently](https://stackoverflow.com/questions/20700053/how-to-add-local-jar-file-dependency-to-build-gradle-file/20700183#20700183) 
 "Gradle updates have deprecated `compile` ...in favor of `implementation` ..." which suggests the instructions are a bit dated.
1. Unfortunately Gradle documentation is even more opaque than Git and doesn't even seem to be searchable.

seadowg:
> "Gradle updates have deprecated compile ...in favor of implementation ..." which suggests the instructions are a bit dated.

Looks like our instructions are out of date, and you should use `implementation` instead of `compile` as the error suggested.

seadowg:
Closing this. If you have run into any more problems, then please post in #javarosa-code or #collect-code on Slack.

dimwight:
Hi @seadowg, you'll be interested to know I finally worked out my own solution to this, which works quite elegantly in the latest AS. Can document this if anyone is interested. 

Since #4059 was yours originally, do take a look at the revised version of this PR now on JR as [#645](https://github.com/getodk/javarosa/pull/645)

PR #4801, opened 2021-08-18
Improved handling swiping and displaying contextual menu

### LINKED ISSUE #4508: Remove response/group menu appears if user swipes on the question title with more pressing force
Collect v1.30.1 (older versions were not verified)

#### Problem description
Remove response/group menu appears if user swipes on the question title with more pressing force 

1. User swipes on line with question title
2. User needs to press harder during the swipe
3. Remove response/group menu appears but user is moved to a different question.

ezgif com-video-to-gif
ezgif com-video-to-gif(1)

#### Expected behavior
This behavior may be confusing and painful if user accidentally removes data

grzesiek2010:
Closes #4508

I tested the implemented changes manually and ran automated tests.

There is no one simple way to solve the issue as I explained in https://github.com/getodk/collect/issues/4508#issuecomment-901002950 so this seems to be the only solutiom.

Verifying that the issue no longer exist should be enough. Nothing else should be affected,

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1315,7 +1315,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
     }
 
     private boolean moveScreen(Direction direction) {
-        odkView.cancelPendingInputEvents();
+        currentView.cancelPendingInputEvents();
         closeContextMenu();
         FormController formController = getFormController();
         if (formController == null) {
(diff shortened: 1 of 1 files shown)


seadowg (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1 and 11.

Tested scenarios:
- try to reproduce the original issue
- swipe forward/backward
- try original scenario with using buttons 
- remove response 
- remove group 
- both
- try with different forms

mmarciniak90:
Tested also on Android 8.1, 9.0, 10.0

PR #4802, opened 2021-08-18
Allow complex emojis in icon [master]

### LINKED ISSUE #4790: Support complex Emojis as icon in Collect

seadowg:
Closes #4790
~~Blocked by #4791~~

I've made some changes to the original implementation to pull out some common logic between the `OneSignTextWatcher` and `ProjectDetailsCreator` to a new `firstCharacterOrEmoji` function (which also know has the majority of the test coverage for the logic). I also cleaned up our `StringUtils` so we now have a `StringUtils` in `shared` and an `HtmlUtils` in `collect_app`.

Worth QAing again - the feature is well tested at this point, but playing around with different emojis and text would be a good idea as it's hard to know that there won't some weird edge case that could break something.

### DIFF
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -53,7 +53,9 @@ dependencies {
     implementation Dependencies.androidx_lifecycle_livedata_ktx
     implementation Dependencies.android_material
     implementation Dependencies.androidx_fragment_ktx
-    implementation "com.vdurmont:emoji-java:5.1.1"
+    implementation ("com.vdurmont:emoji-java:5.1.1") {
+        exclude group: 'org.json', module: 'json'
+    }
 
     testImplementation project(':testshared')
     testImplementation Dependencies.junit
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     implementation Dependencies.androidx_lifecycle_livedata_ktx
     implementation Dependencies.android_material
     implementation Dependencies.androidx_fragment_ktx
+    implementation "com.vdurmont:emoji-java:5.1.1"
 
     testImplementation project(':testshared')
     testImplementation Dependencies.junit
--- a/androidshared/build.gradle
+++ b/androidshared/build.gradle
@@ -48,12 +48,12 @@ android {
 
 dependencies {
     implementation project(':strings')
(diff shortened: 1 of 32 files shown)


androidshared/build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
     implementation Dependencies.androidx_lifecycle_livedata_ktx
     implementation Dependencies.android_material
     implementation Dependencies.androidx_fragment_ktx
+    implementation project(path: ':shared')
```
grzesiek2010: Let's keep it clean so all project implementation are together at the beginning, that means this should be after `implementation project(':strings')`

shared/build.gradle
```diff
@@ -14,6 +14,7 @@ java {
 
 dependencies {
     implementation Dependencies.kotlin_stdlib
+    implementation ("com.vdurmont:emoji-java:5.1.1")
```
grzesiek2010: Please add it to `Dependencies` class and use like all the other dependencies.

grzesiek2010 (changes requested):

grzesiek2010 (approved):

PR #4803, opened 2021-08-18
Use MaterialComponents as theme

### LINKED ISSUE #4465: Scroll bar is pretty thick and has a hint of purple
Collect v1.30.1

#### Problem description
Scroll bar is pretty thick and has a hint of purple

Look at the all widgets form and scroll around.

#### Expected behavior
Scroll bar is a neutral gray color and is of a standard size (e.g., like Gmail's). I can be convinced that maybe the scroll bar needs to be a little thicker? But we should conform to Material design on the width. Either way, the purple has got to go.

#### Other information 

### LINKED ISSUE #4572: Switch to MaterialComponents theme

seadowg:
Closes #4572
Closes #4465

Played with the app on device, ran tests.

Comments inline.

Changes quite a lot visually, so looking out for things that could be a problem for the user or might just be wrong would be good. Also, good to think about if we should be using an alternative color palette to the one I'm used here.


seadowg:
@yanokwa @grzesiek2010 @kkrawczyk123 @mmarciniak90 I think now is a good time to get some feedback. The app now has the Central/ODK blue and magenta as its primary and secondary color, and I've taken as many other random colors out as I can, so hopefully the look and feel is now a bit cleaner/consistent. There are some things I feel we might want to change but I'm going to stay silent so we can see if others notice them. There also might be some genuine problems where colors need to be changed that I've missed, so keep an eye out!

Some things not to look at yet:

* Map views - I'm going to review the color palette used here as it's mostly non theme colors
* Dark theme - Will look bad as I've not set up the dark mode colors yet
* Any of the code - There are still clean-up tasks to take care of so don't bother reviewing code

androidshared/src/main/res/values/toolbars.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <style name="Widget.AndroidShared.Toolbar.Primary.CloseIcon" parent="Widget.MaterialComponents.Toolbar.Primary">
+    <style name="Widget.AndroidShared.Toolbar.Primary.CloseIcon" parent="Widget.MaterialComponents.Toolbar.Surface">
```
seadowg: We still want to keep a "grey" status bar rather than a primary one (which would be bright blue)

collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
```diff
@@ -50,6 +54,7 @@
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         themeUtils = new ThemeUtils(this);
+        AppCompatDelegate.setDefaultNightMode(themeUtils.isDarkTheme() ? MODE_NIGHT_YES : MODE_NIGHT_NO);
```
seadowg: This sets the theme to light or dark mode based on the setting. We could also have this use the automatic mode, but that should probably be another issue/feature.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -315,18 +314,8 @@ protected void restoreFromInstanceState(Bundle state) {
...
-            case Instance.STATUS_COMPLETE:
-                return enlarged ? R.drawable.ic_room_deep_purple_48dp : R.drawable.ic_room_deep_purple_24dp;
-            case Instance.STATUS_SUBMITTED:
-                return enlarged ? R.drawable.ic_room_green_48dp : R.drawable.ic_room_green_24dp;
-            case Instance.STATUS_SUBMISSION_FAILED:
-                return enlarged ? R.drawable.ic_room_red_48dp : R.drawable.ic_room_red_24dp;
-        }
-        return R.drawable.ic_map_point;
+    private static int getDrawableIdForStatus(boolean enlarged) {
+        return enlarged ? R.drawable.ic_room_primary_48dp : R.drawable.ic_room_primary_24dp;
```
seadowg: Could definitely be a controversial change: my thinking here is the same as with the form status icons in the form lists.

collect_app/src/main/java/org/odk/collect/android/projects/ProjectSettingsDialog.kt
```diff
@@ -82,7 +82,7 @@ class ProjectSettingsDialog : DialogFragment() {
             dismiss()
         }
 
-        return AlertDialog.Builder(requireContext())
+        return MaterialAlertDialogBuilder(requireContext())
```
seadowg, line 85: We should really be using this everywhere, but didn't want to go down that route here.

collect_app/src/main/res/layout/main_menu.xml
```diff
@@ -44,6 +43,8 @@
                 <org.odk.collect.android.views.MultiClickSafeButton
                     android:id="@+id/enter_data"
                     style="@style/Widget.Collect.Button.FullWidth"
+                    android:backgroundTint="?colorSurface"
```
seadowg: These buttons should really be `?colorPrimary` but they also shouldn't really be that giant. Down the line we should redesign the main menu interface, but for the moment I think sticking with the big grey buttons is the best idea.

collect_app/src/main/res/drawable/form_state_saved_circle.xml
```diff
@@ -4,14 +4,11 @@
         android:viewportWidth="120"
         android:viewportHeight="120">
 
-    <path
```
seadowg, line 7: Removing the colors from the form status icons could definitely be controversial. I've got a few thoughts on this:

* We view forms by their status ("Fill Blank Form", "Edit Saved Form" etc) so the icons are mostly going to be the same. The only exceptions are failed submissions (which I've kept the colouring on) and Edit Saved Form containing finalized and non-finalized forms (which we're planning to get rid of).
* The color is only fuzzily linked to the state and will have different implications for different people.
* In my opinion, the color actually distracts from the icon itself, which I think conveys a little more about the state.
* Generally, I think the app looks cleaner without the coloured icons.

grzesiek2010:
I played with your changes a bit and I noticed some things:

1. Menu items look like they were disabled (they are faded out):


2. In some dialogs positive/negative buttons are not visible:



3. The back arrow (in toolbar) is hardly visible:


4. Shouldn't we adjust other colors as well like the green color of `OpenMapKit` button or the selection color in selecting from csv images?:



5. The shadow above the navigation bottom bar does not look good now:


6. If Text fields in `Add new project ` are blue why there are not blue in `FormEntryActivity` and in settings?



7. I think you should remove `Experimental` option from settings


8. What about teeming our `DrawView`?

mmarciniak90:
I also noticed some other things:
1. Bearing and Geopoint dialogs opened on Dark Mode don't have visible content.

|||
|---|---|

2. Line for Likert widget is broken

|||
|---|---|

3. I used fresh installation, I set dark mode and started filling All widgets form **from the end to the beginning**
Widgets were not readable.

seadowg:
@grzesiek2010 @mmarciniak90 good finds. Most things here I'll just fix, but a few things below I thought were worth discussing.

> Menu items look like they were disabled (they are faded out)

Yeah by default the Material `DayNight` theme makes these controls medium emphasis. The code that does this in the library:

```xml
<style name="ThemeOverlay.MaterialComponents.Toolbar.Surface" parent="">
    <item name="colorControlNormal">@color/material_on_surface_emphasis_medium</item>
    <item name="actionMenuTextColor">@color/material_on_surface_emphasis_medium</item>
</style>
```

I was interested to see if people noticed/didn't like it. I'm not sure I have a strong preference on customizing it but I think if they looked disabled to you we should.

> If Text fields in Add new project are blue why there are not blue in FormEntryActivity and in settings?

They're different components. The Add Project screen is using Material's `TextInputEditText` which highlights using `?colorPrimary` and the forms/settings are using plain old `EditText` views which use `controlControlHighlight` (which by default is set to `colorSecondary` for Material themes). We could change `controlControlHighlight` to `colorPrimary` but this would also change the color used for radio buttons and checkboxes etc. Personally I'm leaning towards just letting that be at the moment and looking at changing our text fields to the material ones gradually. How does that sound?

> I think you should remove Experimental option from settings

It would be nice to keep it there for adding other feature flags in the future. I'll look into if there's an easy way to hide the option if it's prefs screen is "empty".

> Shouldn't we adjust other colors as well like the green color of OpenMapKit button or the selection color in selecting from csv images?

The OSM widget should be an easy change but the image select widget seems like it deals with highlights in Javascript (and won't have access to theme colors) so I think I'll leave that for the moment.

As an aside I've fixed the geo and bearing widget dialogs but these use old (and deprecated) `ProgressDialog`s so instead of using our specific theme colors I've just had them revert to default material light and dark. We could do more work to fix this but it seems like a better idea to just replace them with non-deprecated components that will automatically be styled correctly. I could do that as part of this PR but it doesn't feel like a big enough deal given it's only a few dialogs. Thoughts?

grzesiek2010:
> Personally I'm leaning towards just letting that be at the moment and looking at changing our text fields to the material ones gradually. How does that sound?

That sounds good. I kind of do not like the fact that we use different colors for text fields and buttons (two main components in our widgets). But if we fix it later that's fine.

> It would be nice to keep it there for adding other feature flags in the future. I'll look into if there's an easy way to hide the option if it's prefs screen is "empty".

ah ok but either way there was something wrong with it because clicking on it crashed the app.

> The OSM widget should be an easy change but the image select widget seems like it deals with highlights in Javascript (and won't have access to theme colors) so I think I'll leave that for the moment.

ok let's fix OSM at least.

seadowg:
> I used fresh installation, I set dark mode and started filling All widgets form from the end to the beginning
Widgets were not readable.

@mmarciniak90 I've seen this a couple of times, but I'm having a really hard time getting solid repro steps. Do you have a way of making this happen reliably? It's really weird!

mmarciniak90:
> Do you have a way of making this happen reliably?

@seadowg checked these steps:

1. Install app!
2. Select Demo project
3. Set dark theme
4. Download `All widgets` form
5. Start this form
6. Jump to the end
7. Swipe back until the `Select multiple minimal` widget

I know it's not common behavior to fill a form from the end but it's reproducible. I checked it on 3 devices. 
It works only **once** on a fresh install.

seadowg:
@mmarciniak90 that's great, thanks. It looks like it only happens once per app launch, rather than only on fresh install. I can make it happen again if I kill the app and then launch it again. Again: this is weird!

seadowg:
@mmarciniak90 @grzesiek2010 how do we feel about the likert widget without lines?



The `LikertWidget` layout code could do with a lot of love, and I think we could probably implement it in a much nicer way with `ConstraitLayout`.  I could get into some hackery to try and get the lines to work again correctly, but not sure if that's really worth spending a lot of time on?

lognaturel:
Great to see this in action!

To reiterate some things I said when we last spoke about this -- I understand why you're doing this from a code standpoint. My ask is to really try to minimize the user-facing impact. It doesn't seem like much but colors and sizes help with feeling oriented in the app and changing them can be surprisingly disruptive. I think overall the conversation here is going in that direction. 👍

How about making color secondary the same as color primary? I don't think the magenta color adds anything and it feels surprising to me every time I see it. Keeping just one accent color for now feels like the most consistent way to go.

> Remove status color coding from form map

I've seen map-based instructions that say "revisit the blue households" (non-finalized) so this is definitely disruptive. Is it necessary? It doesn't seem like a meaningful simplification. If you really think it's important to try removing the differentiation, then I think it should be announced as a breaking change and a beta should come out as early as possible for feedback. I have a pretty strong preference for reverting this change.

I liked the different colors on the different list screens but feel less strongly about those since, as @seadowg notes, we don't currently combine instances of different statuses except for on the edit screen.

The lines in the likert widget are what makes it look like a likert scale. I think those should be restored sooner than later.

In the form-filling context, I find the filled buttons and the color primary screen navigation bar to be pretty distracting. It's hard to focus on anything else! I liked the navigation being visually tied to the keyboard. I don't have a great suggestion for the form filling buttons. Maybe outlines? It's probably not a huge deal but did want to flag that it bothers me. Same thing about the buttons at the bottom of the form/instance list screens.

seadowg:
> My ask is to really try to minimize the user-facing impact.

Agreed. To be very clear about my approach here: I just wanted to try the app with everything as close to "default" material as possible and see how we felt about it. I don't want to bring any customizations with us without discussing them first. Definitely happy to roll back to anything that we feel we need to leave in.

> I've seen map-based instructions that say "revisit the blue households" (non-finalized) so this is definitely disruptive.

Ah, that's a great example. Definitely feels like we should keep the colors in this case. I'd like to change them to color-blind safe palette (assuming they're not already) but I think that would be pretty disruptive if people are already training on the existing ones, so we can punt on that.

> How about making color secondary the same as color primary?

Yeah I like that idea. I don't really like the magenta either (although it'll be sad to lose it - it's been good to us).

> I liked the different colors on the different list screens but feel less strongly about those since, as @seadowg notes, we don't currently combine instances of different statuses except for on the edit screen.

To me, it looks a lot cleaner without the colors. I'm in support of leaving them out.

> The lines in the likert widget are what makes it look like a likert scale. I think those should be restored sooner than later.

We can definitely have them in there, but it's going to be more work (we'd really need to make custom components for each item). Maybe we need a separate discussion about this, but I'm not sure if I understand why it has to be presented that way. Survey Monkey displays in a [completely different manner](https://www.surveymonkey.com/mp/customer-satisfaction-survey-template/) for instance.

> In the form-filling context, I find the filled buttons and the color primary screen navigation bar to be pretty distracting. It's hard to focus on anything else! I liked the navigation being visually tied to the keyboard. I don't have a great suggestion for the form filling buttons. Maybe outlines? It's probably not a huge deal but did want to flag that it bothers me. Same thing about the buttons at the bottom of the form/instance list screens.

Yeah I agree with most of this actually - not sure what you mean by the "form filling buttons" though. I think we should just make the nav bar `colorSurface`. The form/instance lists are using a bit of a weird/non-material pattern so I think we also just make those `colorSurface`?

lognaturel:
> it's been good to us

We can still keep the experimental theme. 😉

> I'd like to change them to color-blind safe palette

That's definitely a good thing to look into. I think green for sent, red for failed, and blue for editable have to stay but the shades could be different. I don't think the finalized color is as important.

> I'm in support of leaving them out.

Fair enough.

> I'm not sure if I understand why it has to be presented that way.

Maybe that's fair. I'm used to always seeing lines in academic contexts. To me they do make it feel more like a continuum but it's true that they don't really add a lot of meaning. And it's true that in things like customer satisfaction contexts there isn't a standard. Certainly, this shouldn't block the PR and can be considered separately. 

> not sure what you mean by the "form filling buttons" though

Sorry, I just mean buttons such as the one for image selection that show up in a form filling context. I think your proposals there sound good!

seadowg:
@mmarciniak90 I've been playing around with the weird bug and it looks like it happens when view the select image map widget. This form reproduces the problem once per app launch:

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms">
    <h:head>
        <h:title>One Question</h:title>
        <model>
            <instance>
                <data id="one_question" orx:version="1">
                    <select_one_image_map/>
                </data>
            </instance>
            <bind nodeset="select_one_image_map" type="select1"/>
        </model>
    </h:head>
    <h:body>
        <select1 appearance="image-map" ref="/data/select_one_image_map">
            <label>Select image</label>
            <hint>select_one with type image-map appearance and image body.svg</hint>
            <item>
                <label>Head</label>
                <value>head</value>
            </item>
            <item>
                <label>Neck</label>
                <value>neck</value>
            </item>
            <item>
                <label>Lungs</label>
                <value>lungs</value>
            </item>
            <item>
                <label>Left proximal arm</label>
                <value>left_proximal_arm</value>
            </item>
            <item>
                <label>Left distal arm</label>
                <value>left_distal_arm</value>
            </item>
            <item>
                <label>Right proximal arm</label>
                <value>right_proximal_arm</value>
            </item>
            <item>
                <label>Right distal arm</label>
                <value>right_distal_arm</value>
            </item>
            <item>
                <label>Left proximal leg</label>
                <value>left_proximal_leg</value>
            </item>
            <item>
                <label>Left distal leg</label>
                <value>left_distal_leg</value>
            </item>
            <item>
                <label>Right proximal leg</label>
                <value>right_proximal_leg</value>
            </item>
            <item>
                <label>Right distal leg</label>
                <value>right_distal_leg</value>
            </item>
        </select1>
    </h:body>
</h:html>
```

It looks like the problem is something to do with the first loading of a `WebView` in the app. If you switch to another app and then back to Collect the text colors corrects itself. Next thing I'll do is try this on a few different versions of Android (as the `WebView` changes with API level) and see if it only happens on certain ones.

seadowg:
Ok so remaining things here:

- [x] Make secondary actions that `colorSurface` (nav bar, nav actions and selection controls on form lists)
- [x] Use blue for primary and secondary colors
- [x] Restore color coding for instances on form map
- [x] Fix select image map bug

seadowg:
Ok this is ready for another look @mmarciniak90 @grzesiek2010!

grzesiek2010:
It looks great! I found some issues like:

1. Project settings dialog is gray what doesn't look good compared to other dialogs:

|  |  |  |
| Screenshot_1630416639  | Screenshot_1630416634  | Screenshot_1630416733  |

2. DatePicker with calendar mode seems to be broken:


3. Our Full screen dialogs are grey 

|  |  |  |
| Screenshot_1630417513  | Screenshot_1630417539 | Screenshot_1630417563  |

4. Answers displayed in image select widgets are grayed (other widgets have black answers)

|  |  |  |
| Screenshot_1630417788  | Screenshot_1630417769 | Screenshot_1630417734  |

5. I'm not sure if I like how the ran widget looks like... additionally options are not highlighted in any way during changing the order. Maybe the options should be white with a blue border?



6. Numbers in range widgets are grayed out. It looks fine when there is no option but once there is something selected it should probably be black like all the other answers.

|  |  | 
| Screenshot_1630417702 | Screenshot_1630417706 | 


I just tested the light theme so probably the same applies for the dark one too.

grzesiek2010 (changes requested):
Some issues to fix as listed in my comment.

grzesiek2010:
I'll just add some other things I noticed reviewing the changes:

1. Some dialogs use green progress indicators.

|  |  |  |
| Screenshot_1630534781 | Screenshot_1630534794 | Screenshot_1630534809 |

2. I think colors  that we use for geotrace/geoshape should not be changed, red was well visible.

|  Before | After | 
| Screenshot_1630535173  | Screenshot_1630535349 |

3. I wonder if our contextual menu should not be white. It would look similar to the clipboard menu then.

|  |  | 
| Screenshot_1630535653  | Screenshot_1630535664 |

seadowg:
> Project settings dialog is gray what doesn't look good compared to other dialogs:
> Our Full screen dialogs are grey

These dialogs are using our `colorSurface` and are actually [correct for Material](https://www.material.io/components/dialogs/android#full-screen-dialog). I'll try and make it so that our normal alert dialogs use the same color for their backgorund/container.

> Some dialogs use green progress indicators.

I'm actually going to propose we stick with this. The dialogs with this problem are `ProgressDialog`s and I don't feel like it's worth the time/effort to fix these when really we should just replace them. I've deprecated them (via the `DayNightProgressDialog`) and there are only 8 in the app. What do you think?

grzesiek2010:
> I'll try and make it so that our normal alert dialogs use the same color for their backgorund/container.

so you want to make all the dialogs grey? Whether it's correct for Material design or not I like the wait version more.

seadowg:
> so you want to make all the dialogs grey? Whether it's correct for Material design or not I like the wait version more.

I think in this case, it feels right to me to defer to Material. The dialogs in material are [described as a "surface"](https://material.io/components/dialogs#anatomy) and so it makes sense that the `MaterialAlertDialogBuilder` is defaulting to using `colorSurface` rather than the background color here. The non-material `AlertDialog`s are displaying with a white background as they are using the default Material `colorSurface` rather than ours. I think maybe what we can do is experiment with changing our `colorSurface` back to the default (I'm not sure why it's grey). We might need to pull out a special grey color for the main menu buttons etc but that's fine as they wouldn't usually be using `colorSurface` anyway.

seadowg:
@grzesiek2010 ok that's pretty much all the issues you found (other than the rank widget which I'm still working on) fixed. The big change you'll see is that the app bar is now white instead of grey (as I've changed the surface color to white). Let me know how that looks!

grzesiek2010:
It looks fantastic!

1. I think we should bring back the shadow that we had above the navigation bar because now it's not clear where the border is. It was too thick so would be good to improve it.



2. I noticed that the dialog in BearingWidget has no longer radios corners (I didn't check all dialogs we use)

| Before  | After |
| Screenshot_1630584886 | Screenshot_1630585022  |

3. In dark theme the toolbar is has the same color in full screen dialogs:



4. The stars in RatingWidget in dark mode are barely visible



5. In dark mode dialogs have different colors and also some of them no longer have radios corners (like in light mode)

|  |  |  |
| Screenshot_1630585478  | Screenshot_1630585495  | Screenshot_1630585525  |

seadowg:
@grzesiek2010 I'll look at fixes for those as well. The geo widgets are using the old `ProgressDialog` (which uses the no longer maintained `android.app.AlertDialog`) so, as I said before, I think it's best to skip trying to fix those and just look at replacing them in the future.

kkrawczyk123:
@seadowg there are my findings:

- duplicated project dialog doesn't have buttons:
Screenshot (Sep 2, 2021 10_25_02 AM)

- text on search bar on detele saved forms/blank forms in light mode is white:
deleteform
It works properly in dark mode.

kkrawczyk123:
I know you have mentioned that dark mode is not perfect yet so just want to make sure you will take a look at finalize icon  in dark mode:

| | |
| ---- | ---- |
| Screenshot (Sep 2, 2021 3_10_48 PM) | Screenshot (Sep 2, 2021 3_10_02 PM) |

seadowg:
@kkrawczyk123 I wasn't able to reproduce the icon problem you were seeing. 

That's everything else found (other than the dialog problems that I think we should punt on) fixed, hopefully!

kkrawczyk123:
@seadowg it is visible only on my device with Android 11 - sorry for not being specific before!

seadowg:
@kkrawczyk123 can't reproduce on Android 11 emulator. What device is it? Are you able to reproduce from a fresh install?

kkrawczyk123:
@seadowg it's Google Pixel 3a. I am able to reproduce on fresh install by following steps:
1. Create project, download some forms, fill submissions
2. Kill app
3. Change theme 
4. Open Fill Blank Form/Edit Blank form - icons are not visible

It happens when change from dark to light and the opposite as well. However, it happens often but it's not happening in every try.

seadowg:
@kkrawczyk123 is it always the same icon?

kkrawczyk123:
@seadowg  All icons - I can't reproduce for finalized submissions only :( maybe it's not even the same issue?
| | |
| --- | --- |
| Screenshot (Sep 3, 2021 9_18_22 AM) | Screenshot (Sep 3, 2021 11_04_37 AM) |

seadowg:
@kkrawczyk123 no worries, it makes more sense if it's all the icons. If switch to another app and then back, does it correct itself? I'll try and see if I can reproduce as well.

kkrawczyk123:
@seadowg not always actually but it happens. Sometimes it corrects after just changing the view to view sent form/send the finalized form and get back.

seadowg:
@kkrawczyk123 I've made a change that might fix the issue. I was only able to reproduce once before though, so I might not have cracked it.

mmarciniak90:
@seadowg I also noticed these places for improvement

1. Icon on missing permission dialog on dark theme

|Screenshot_20210903-133209_ODK Collect|Screenshot_20210903-133220_ODK Collect|Screenshot_20210903-133229_ODK Collect|
|---|----|---|

2. Dialog in `Get Blank Form` for Google Drive server

kkrawczyk123:
@seadowg it looks like this fix helped for this issue - I wasn't able to reproduce it not even once and tried plenty of times!

seadowg:
@kkrawczyk123 ace! I'll have a look at those other issues you found. I'm pretty sure the "Reading files" dialog is the old progress dialog that we just need to replace and won't fix here.

seadowg:
@kkrawczyk123 ok those icons should be fixed now!

mmarciniak90:
One more issue:

Sort by on all views

seadowg:
> Sort by on all views

@mmarciniak90 I'm probably just missing it, but what's wrong there?

mmarciniak90:
I expected darker background 😄

seadowg:
@mmarciniak90 ah no, that's the right color I think. Bottom sheets like that use the "surface" color (see [here](https://www.material.io/components/sheets-bottom#theming)) rather than the "background" color. In dark theme the surface is a dark grey and the background is black, whereas in the light theme they are both white.

yanokwa:
The current build seems to default to Dark theme for the first launch screen. Then after you hit "Try a demo", then delete that demo project, it goes back to the White theme for the first launch screen. 

And in the process of deletion, the Add project icon seems to be white the first time you use it. But it goes back to being black the second time. I think @kkrawczyk123 reported something similar...
Screen Shot 2021-09-03 at 11 35 13

seadowg:
@yanokwa are you using dark theme on your device? It looks like we're (accidentally) doing automatic dark/light detection until you're into a project at the moment - we override the system theme with whatever the project specifies. I'll hardcode light for the onboarding stuff at the moment, and then we can come back and have the autodetection work for the whole app later.

seadowg:
@kkrawczyk123 @mmarciniak90 I'm going to mark this ready for testing now. I know you've already done a bunch of investigation, so it might not require much more of a pass. Given a lot of people being out at the moment I don't think this will get a code review this week, but I don't think anything in there is going to be controversial enough to block QA.

kkrawczyk123:
@seadowg There is one more thing that is bothering me - Range rating widget in dark mode. I barely see stars that I can select. 
see: 
| this PR | store version |
| ---- | ---- |
| Screenshot_20210906-134355 | Screenshot_20210906-134808 |

seadowg:
@kkrawczyk123 I had thought that was an existing problem, but it does look like the background color change has caused problems there. I'll have a look at what we can do there.

mmarciniak90:
@seadowg I noticed a change in the Rank widget. Previously, the highlight was only visible when this option was selected. Now, the highlight does not disappear. Is it intentional?

|ezgif com-video-to-gif|new
|---|---|

seadowg:
@kkrawczyk123 @mmarciniak90 both those issues should be fixed now

kkrawczyk123:
@seadowg I've just encountered another weird behavior only on Android 11 after changing dark mode to light mode. Icons and jump arrow in hierarchy view are not visible: 
ezgif com-gif-maker (1)

seadowg:
@kkrawczyk123 I'm thinking that (and the related bugs) are rooted in a core Android library ("AppCompat") that we're using. I've been tweaking our approach throughout, but not sure if there is much left I can do. I'm going to attempt to upgrade it. This would usually be pretty simple, but the Android team has introduced some things that interfere with other core libraries (lint is a problem for us). If that doesn't get rid of these flaky icons, we might need to look at approaching this slightly differently. 

I think the crux of what we're running into here is that Android's dark/light mode forcing just doesn't work very well, as most apps probably just use the system theme and these problems don't get noticed. The good news is that we can also introduce a "System" theme option - I was just hoping to not have to do that yet.

*sigh*

seadowg:
OK it looks like we're running into bugs in `appcompat` here: https://issuetracker.google.com/issues/195570263. It seems like the Google team thinks it's a caching issue. That seems to make sense, I'm having some success reproducing the icon problem with the following steps:

1. Open app
2. Go to Fill Blank Form
3. Return to Main Menu
4. Change to Dark theme
5. Open Fill Blank Form again

I'm going to try using the alpha version of appcompat as suggested in the issue.

seadowg:
Experimenting with the alpha versions of appcompat didn't help. The above repro steps still lead to the same icon problem. We could opt to not use the automatic dark/light switching, but it seems like follow system colors is most definitely the [direction of the wind](https://material.io/blog/announcing-material-you) on Android at the moment and the key problem we're seeing only occurs briefly when switching between themes, which is not exactly something we expect to be part of someone's everyday workflow. 

I think the best option here to avoid these kinds of bugs (that seem to be core Android library problems) is to introduce a new "System default" theme option (and make it the default). I've spiked this out, and it seems switching the theme at the system level does not result in the same icon problems (probably because it handles drawable caches etc more gracefully/with less prejudice). This makes it far less likely that users would run into any of the dark/light flakiness (as they'd be less likely to be customizing the theme).

seadowg:
@kkrawczyk123 @mmarciniak90 the app will now use dark/light themes based on the system theme. You can still override this in User Interface settings, but all projects without a set theme will just use the "System default" one. 

There is still some occasional flakiness when switching between light and dark themes (in-app, not at the system level as far as I can see) but I'm not sure if that feels like a blocker to me given users will only need to use the in-app light/dark themes if they don't want the app's theme to match their system one (and we're only seeing the issue on Android 11). Interested in thoughts from @yanokwa on this also.

mmarciniak90:
@seadowg, following the device theme, gives us an opportunity to test the change of the device theme during some action on Collect.

Use case 1:
1. Google Drive is set as a server
2. Open Get Blank Forms
3. `Reading files` dialog is visible
4. Change device theme from expanding quick settings -> **Collect crashes**

Use case 2:
1. Click on `Set admin password` to open dialog
2. Change device theme from expanding quick settings when dialog is visible -> **Collect crashes**

I can imagine that this behavior is unlikely but it shows that these places are not secured to the same level as others.

Screenshot_20210913-125304_ODK Collect

Use case 3:
1. Change theme when some dialog with calendar/date/year/time is opened.
2. Unreadable dialog is visible or dialog is visible in an incorrect theme.

Surprising is that this problem does not occur for Ethiopian, Coptic, Islamic, etc. appearances.

collect_app/src/main/java/org/odk/collect/android/activities/DrawActivity.java
```diff
@@ -147,6 +147,13 @@ public void onClick(View view) {
             }
         });
 
+        cardViewClear.setOnClickListener(this::clear);
+        fabClear.setOnClickListener(this::clear);
+        cardViewSaveAndClose.setOnClickListener(this::close);
+        fabSaveAndClose.setOnClickListener(this::close);
+        cardViewSetColor.setOnClickListener(this::setColor);
+        fabSetColor.setOnClickListener(this::setColor);
```
grzesiek2010, line 155: All those methods can now be private if they are not used from a layout file.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -369,6 +371,12 @@ public void allowSwiping(boolean doSwipe) {
      */
     @Override
     public void onCreate(Bundle savedInstanceState) {
+        // Workaround for https://issuetracker.google.com/issues/37124582. Some widgets trigger
+        // this issue by including WebViews
+        if (Build.VERSION.SDK_INT >= 24) {
```
grzesiek2010, line 376: ahh nasty... does it need to be in an activity or maybe we can move it to the widget that uses WebView?

collect_app/src/main/java/org/odk/collect/android/geo/MapFragment.java
```diff
@@ -59,6 +59,8 @@
     String CENTER = "center";
     String BOTTOM = "bottom";
 
+    String LINE_COLOR = "#ffff0000";
```
grzesiek2010: Why this color is not in colors.xml?

collect_app/src/main/res/drawable/ic_map_point.xml
```diff
@@ -3,6 +3,6 @@
         android:height="36dp"
         android:viewportWidth="36"
         android:viewportHeight="36">
-    <path android:fillColor="@color/mapPointCircle" android:pathData="M18,13c-2.77,0 -5,2.24 -5,5s2.24,5 5,5s5,-2.24 5,-5s-2.24,-5 -5,-5z"/>
-    <path android:fillColor="@color/mapPointFill" android:pathData="M18,21.33c-1.84,0 -3.33,-1.5 -3.33,-3.33s1.5,-3.33 3.33,-3.33s3.33,1.5 3.33,3.33s-1.5,3.33 -3.33,3.33z"/>
+    <path android:fillColor="#ffff0000" android:pathData="M18,13c-2.77,0 -5,2.24 -5,5s2.24,5 5,5s5,-2.24 5,-5s-2.24,-5 -5,-5z"/>
```
grzesiek2010: It's the same color like we use to draw lines so shouldn't it be specified in colors.xml?

collect_app/src/main/res/values/shape.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="ShapeAppearance.Collect.MediumComponent" parent="ShapeAppearance.MaterialComponents.MediumComponent">
+        <item name="cornerFamily">rounded</item>
```
grzesiek2010: According https://material.io/develop/android/theming/shape `rounded` is the default value so you shouldn't need to override it here.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -98,51 +91,35 @@
...
-            android:textColor="?colorOnSurface"
-            android:layout_gravity="end"
-            app:icon="@drawable/ic_baseline_qr_code_scanner_24"
-            app:iconTintMode="multiply" />
-    </org.odk.collect.android.views.multiclicksafe.MultiClickSafeFrameLayout>
+        app:layout_constraintEnd_toStartOf="@id/about_button"
+        app:layout_constraintHorizontal_chainStyle="spread"
+        app:layout_constraintStart_toStartOf="parent" />
 
-    <org.odk.collect.android.views.multiclicksafe.MultiClickSafeFrameLayout
```
grzesiek2010, line 126: That MultiClickSafeFrameLayout was used to make the whole area clickable, now it's not but I think it should be in this case where the button itself is not very big.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
```diff
@@ -11,13 +11,6 @@
         android:orientation="vertical"
         app:layout_constraintGuide_begin="@dimen/margin_standard" />
 
-    <androidx.constraintlayout.widget.Guideline
```
grzesiek2010, line 14: The changes you implemented changed the way the buttons on the bottom are displayed. Before they were aligned to the center:
Screenshot_1631609189
what looked more like:
241812187_446116526693840_2718891950956638957_n

and now it's:
Screenshot_1631608904

(the difference might not be well visible in such a small screen but still.

collect_app/src/main/res/drawable/ic_room_blue_24dp.xml
```diff
@@ -4,6 +4,6 @@
         android:viewportWidth="24"
         android:viewportHeight="24">
     <path
-        android:fillColor="@color/blue_500"
+        android:fillColor="#2196F3"
```
grzesiek2010: Why this color (and some other below) is not declared in colors.xml?

grzesiek2010 (changes requested):

grzesiek2010:
I also noticed that elements in Ranking Widget are transparent during moving them.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
seadowg: The recommendation is to have this before `super.onCreate` so needs to be at an Activity level 😭

collect_app/src/main/java/org/odk/collect/android/geo/MapFragment.java
seadowg: I'll pull the map colors out to their own file (as they're not theme related).

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
seadowg: This looked better to me, but this is all pretty subjective. Right now, each button has equal space to their closest container edge, they are centered as a group. If you feel strongly about it, I'm happy to change it so that they are anchored to the center instead, like before.

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
seadowg: The click area felt fine to me, but we can bump up the padding on the buttons if it's feeling too small?

collect_app/src/main/res/drawable/ic_room_blue_24dp.xml
seadowg: I'll make a specific file for these colors as well.

seadowg:
@grzesiek2010 I've fixed/replied to the issues you found. I'll take a look at the crashes @mmarciniak90 found, but I think you're good to take another look.

kkrawczyk123:
@seadowg I can still see some icons issues when changing theme on Android 11 but this time I am not able to provide any valuable reproduction steps :( Good thing is that changing theme in the middle of filling form is not very popular.
| | |
| --- | --- |
|Screenshot (Sep 15, 2021 12_35_16 PM) | Screenshot (Sep 15, 2021 12_34_52 PM) |

seadowg:
@kkrawczyk123 is that changing theme from Collect's settings or changing at the system level?

kkrawczyk123:
@seadowg Collect settings.

seadowg:
@kkrawczyk123 right yeah that's expected on Android 11 as forcing dark/light mode seems to be buggy. I agree this is unlikely to be a problem people will run into though.

seadowg:
@mmarciniak90 I've fixed all the issues you found apart from the date picker. The fact that we don't see the problem with the Ethiopian, Coptic, Islamic, etc pickers is related to #4806. The more basic appearances actually use some out of date components that we probably need to get rid of instead of spending time working around. I think if you feel that we can live with this at the moment we can just check it's fixed whenever we do #4806. What do you think?

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: ah ok

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
grzesiek2010: I liked the first version more but that's not the most important thing here. The thing is that we shouldn't rather implement changes like this one - difficult to spot and not directly related to the subject of this pr. And also centering itself is not that important, more important was to have the whole area clickable what you also changed and now you have to click directly on a button to trigger any action.

grzesiek2010:
The problem with transparent elements in the RankWidget is fixed but in dark mode the dialog looks like (no borders):
Screenshot_1631782327

collect_app/src/main/res/layout/project_settings_dialog_layout.xml
seadowg: That's fair. I should really have called this out. I ended up making this change as I ran into some weird click propagation problems in the commit before (where I had to move the IDs around) and I was pretty sure I could simplify/flatten the view, but you're right that I didn't keep some behaviour intact. I'll see what I can do about increasing the click area for the buttons. 

It's also worth pointing out that it's best to avoid wrapping buttons in another clickable view, as it could cause problems for accessibility (unless we do extra work to make sure content descriptions are set up correctly). Basically, if the tests can't access the button with text, then screen readers will be confused as well. We might just need a custom view for this kind of thing. Again, this is all things I should have called out as part of the PR description/comments, so sorry about that!

seadowg:
> The problem with transparent elements in the RankWidget is fixed but in dark mode the dialog looks like (no borders):

I'd basically never seen this widget until working on this, and now it's public enemy number 1 😄. I think this is because that color (`surfaceButtonColor`) relies on the elevation in the buttons it's being used on, usually. We could add elevation to them to fix this. Alternatively, we could make the items `colorPrimary` and just remove the border that gets added while they're being dragged. What do you think @grzesiek2010?

grzesiek2010:
> I think this is because that color (surfaceButtonColor) relies on the elevation in the buttons it's being used on, usually. We could add elevation to them to fix this. Alternatively, we could make the items colorPrimary and just remove the border that gets added while they're being dragged. What do you think @grzesiek2010?

Doesn't that mean there is something wrong if `surfaceButtonColor` is in dark mode the same as color of dialogs (and we have that bug) but in light mode it's different?

seadowg:
> Doesn't that mean there is something wrong if surfaceButtonColor is in dark mode the same as color of dialogs

It ended up being the same because it's usually only being used in places where the view is also elevated and elevation works differently in dark mode (see the "Elevation" section [here](https://material.io/design/color/dark-theme.html#properties)). I'll experiment with just lightening up the `surfaceButtonColor` and also adding elevation to the rank widget items and see what feels right.

seadowg:
@grzesiek2010 ok those issues are fixed now. The tap area on the buttons is smaller (like we discussed on Slack) but seems good to me. Let me know what you think!

grzesiek2010:
> Let me know what you think!

I think you addressed all the things I mentioned so I'm going to accept this pr and let the QA team spend some more time on it.

grzesiek2010 (approved):

seadowg:
@grzesiek2010 just pushed one little commit. I noticed the buttons on the main menu were being cut off slightly by the padding on the surrounding view (existing problem) so I switched this for margins on the buttons.

mmarciniak90:
@seadowg one more thing, when I change the theme in the list of forms on Google Drive, the theme does not change and the view becomes unreadable when you go to some folder.

vidma-recorder-edited-20092021-1

What was the final decision about default theme text? Do we want to follow the same structure as in language?

seadowg:
> one more thing, when I change the theme in the list of forms on Google Drive, the theme does not change and the view becomes unreadable when you go to some folder.

Oh weird. I'll have a look at that.

> What was the final decision about default theme text? Do we want to follow the same structure as in language?

Good point. I'll make it "Use device theme".

seadowg:
@mmarciniak90 I fixed the theme switching problem on the Google Drive screen. This reintroduces a crash that you found before (changing themes while the "Reading files..." progress dialog is up), but I realized it's an existing issue ([here](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/f56123cbd648ba6878cb4e1fde9c4fc7?time=last-ninety-days&sessionEventKey=614599EA03AD00015FEF11C48CB15DDA_1587624632017986579)). We can do some extra work to fix it if we see a significant amount of people run into it, but I don't think we expect it.

mmarciniak90:
**Tested with success**

**Verified cases:**
- all icons, especially:
   - form icons on all views
   - save/hierarchy/menu icons during filling form
   - remove/add group icons
   - up icon in form
   - icons in missing permission dialogs
- all dialogs, especially:
   - dialog with progress
   - radios corners
   - duplicate project
- all widgets, especially:
   - likert widget without lines
   - rank widget
   - range widget
   - rating widget
   - internal audio recording 
- all buttons, especially:
   - navigation arrows
   - Go to start/end buttons
- search bar
- hints
- all settings
- protected settings
- imported QR code with defined theme 
- Maps
- notifications
- match exactly
- following device theme
- "Use device theme" text
- offline layers
- RTL language

**Known issues:**
- crash when device theme is changed on Google Drive forms list
- unreadable calendar when device theme is changed on opened date dialog
- icons issues when changing theme on Android 11

kkrawczyk123:
Verified also on Androids: 5.1, 7.1 and 11.

lognaturel:
Folks, I’m really not liking the text style for buttons. I’ve been trying to convince myself this doesn’t matter but it continues to bother me so I need to say something at least. 

The button text is very heavy and has exaggerated character spacing. I know Material makes significant use of increased character spacing as a form of emphasis but I think it’s not really used in practice, and particularly not by Google apps or websites. That spacing makes text harder to read for me and in languages where the landing screen button text is already long, it makes it more likely to wrap. 

The weight also makes that landing screen harder to process to me because every button looks like it’s screaming for attention.

Happy to hand that feedback off and let y’all decide what to do with it and/or to play with styles and make an alternate proposal.

seadowg:
@lognaturel happy to chat about this. Looking at it again, we actually customized the button letter spacing to be larger (bumped from 0.333333... to 0.04) in some earlier iterations of our "materialization" so it could be that that is the culprit here. Maybe you could experiment with removing that customization (in `typography.xml` at `TextAppearance.Collect.Button`) or I'm otherwise I'm happy to find a time with can tweak it on a call together.

lognaturel:
Interesting! Let me try to play with it today or tomorrow. Will contact you about a quick Weds call if we can’t finalize before then.

lognaturel:
Forgive me, I was wrong about character spacing. Setting it to 0.04 was great. I sometimes have issues with "apply changes" where resources get messed up and my physical device was showing the non-customized spacing. So it's not as bad as I thought.

Would y'all be open to simply `<style name="TextAppearance.Collect.Button" parent="TextAppearance.Collect.Subtitle1" />`? That would make it look very similar to 2021.2 and prior.

| 2021.2 	| subtitle1 landing 	| subtitle1 question 	|
|--------	|-------------------	|--------------------	|
|   |  subtitle1-landing |  subtitle1-question  	|

seadowg:
@lognaturel ah nice! I'm glad you spotted this as I hadn't clocked (or remembered) about the customization until you brought the spacing up.

> Would y'all be open to simply <style name="TextAppearance.Collect.Button" parent="TextAppearance.Collect.Subtitle1" />? That would make it look very similar to 2021.2 and prior.

I don't think we should link two typography styles like this. If we want `Button` to look like `Subtitle1` let's work out what bits we want and customize them on the `Button` style. That said, did you try it as just `Button` without the customizations? My preference is definitely to defer as much of these decisions to Material as possible, so I think we want a really good argument why the standard `Button` doesn't work.

lognaturel:
> did you try it as just Button without the customizations?

Yes, that's what I was seeing on my device and I think it's unacceptable because of the text spacing. `0.04` looks ok. Even after that customization, the text is still too bold and large to me.

> preference is definitely to defer as much of these decisions to Material as possible

I think where it looks ok and the semantics of what we use match what Material has in mind, that's fine. But I think our first priority should be to make things look ok. In this case, it's a screen people spend a lot of time on and it doesn't look great in general so I'd prefer it not look worse! The buttons we have there also don't match anything in Material so it's already a custom concept.

> I don't think we should link two typography styles like this.

Yes, right. I mean using that look.

Relatedly, I had issues with button styling when working in Figma. The various Material templates I started with were not consistent with button text styling. They're definitely not consistent with this Material default. See e.g. https://www.figma.com/file/3UJK55PPAvNBdYC7pjbQ6r/%E2%9D%96-Material-Theming-%2B--Android-11-(Community)?node-id=166%3A0 where button text is clearly smaller than the app bar title (as I think it should be).

lognaturel:
My button issues have been addressed at https://github.com/getodk/collect/pull/4853!

PR #4818, opened 2021-08-30
Fixed detecting duplicated projects

### LINKED ISSUE #4817: Duplicate project dialog displayed when user has drive project created and try to create demo project through qr code
Collect v2021.2.4 Android all, device used...

#### Problem description
I have noticed unexpected behavior when scanning project qr codes. When I created Google Drive project and tried to add demo project as a new one by scanning qr code from another device duplicated project dialog is displaying but it shouldn't.  
ezgif com-gif-maker

1. Create drive type of project from the first screen
2. On other device create demo project
3. Select add project on 1st device
4. Scan qr code from second device

#### Expected behavior
Duplicated project dialog should not be visible as drive and demo project are totally different projects. 

#### Other information 
- issue does not occur in the opposite scenario when demo project already exists and drive project is scanned one

grzesiek2010:
Closes #4817

The problem took place when we tried to add a Demo project already having a Google Drive project.
- if you have a Google Drive project, server url in preferences is set to an empty string
- if you scan a qr code for a Demo project that qr code does not contain url (because default values are not included during generating qr codes) and also an an empty string is returned

then those urls were compared (and treated as equal of course) and the fact that we had two different types of projects was ignored what caused detecting a duplicated project.

The problem was difficult to test automatically because in  `SettingsConnectionMatcherTest` we use `InMemSettings` which is a replacement for `SharedPreferences` but `InMemSettings` does not return the same defaults as our `SharedPreferences` and then everything seemed to fork fine (for example we had null instead of an empty string and in our code it was treated as two different values). I had to save those defaults first: https://github.com/getodk/collect/pull/4818/files#diff-19ba577600f3b2f53559efdece854ef897e0c0eb92058fecc6f68867cf7cd693R116
In a perfect world we should return the same defaults for production code and tests but maybe let's figure it out in a separate pr? @seadowg what do you think?

It's a safe fix and we can just test the scenario described in the issue to confirm the bug is fixed.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/projects/SettingsConnectionMatcher.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/SettingsConnectionMatcher.kt
@@ -30,12 +30,12 @@ class SettingsConnectionMatcher(
                 val projectUsername = projectSettings.getString(ProjectKeys.KEY_USERNAME)
                 val projectGoogleAccount = projectSettings.getString(ProjectKeys.KEY_SELECTED_GOOGLE_ACCOUNT)
 
-                if (jsonProtocol.equals(projectProtocol) && jsonProtocol.equals(ProjectKeys.PROTOCOL_GOOGLE_SHEETS)) {
-                    if (jsonGoogleAccount.equals(projectGoogleAccount)) {
-                        return it.uuid
-                    }
-                } else {
-                    if (jsonUrl.equals(projectUrl) && jsonUsername.equals(projectUsername)) {
+                if (jsonProtocol.equals(projectProtocol)) {
+                    if (jsonProtocol.equals(ProjectKeys.PROTOCOL_GOOGLE_SHEETS)) {
+                        if (jsonGoogleAccount.equals(projectGoogleAccount)) {
+                            return it.uuid
+                        }
+                    } else if (jsonUrl.equals(projectUrl) && jsonUsername.equals(projectUsername)) {
                         return it.uuid
                     }
                 }
--- a/collect_app/src/main/java/org/odk/collect/android/projects/SettingsConnectionMatcher.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/SettingsConnectionMatcher.kt
@@ -16,9 +16,9 @@ class SettingsConnectionMatcher(
         try {
             val jsonObject = JSONObject(settingsJson)
             val jsonSettings = jsonObject.getJSONObject(AppConfigurationKeys.GENERAL)
-            val jsonProtocol = try { jsonSettings.get(ProjectKeys.KEY_PROTOCOL) } catch (e: JSONException) { ProjectKeys.PROTOCOL_SERVER }
+            val jsonProtocol = try { jsonSettings.get(ProjectKeys.KEY_PROTOCOL) } catch (e: JSONException) { ProjectKeys.defaults[ProjectKeys.KEY_PROTOCOL]!! }
 
-            val jsonUrl = try { jsonSettings.get(ProjectKeys.KEY_SERVER_URL) } catch (e: JSONException) { "" }
+            val jsonUrl = try { jsonSettings.get(ProjectKeys.KEY_SERVER_URL) } catch (e: JSONException) { ProjectKeys.defaults[ProjectKeys.KEY_SERVER_URL]!! }
             val jsonUsername = try { jsonSettings.get(ProjectKeys.KEY_USERNAME) } catch (e: JSONException) { "" }
 
             val jsonGoogleAccount = try { jsonSettings.get(ProjectKeys.KEY_SELECTED_GOOGLE_ACCOUNT) } catch (e: JSONException) { "" }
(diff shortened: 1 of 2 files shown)


grzesiek2010:
@kkrawczyk123 @mmarciniak90 I also spotted another bug which gave us the possibility to add multiple demo projects. Both issues should be fixed.

seadowg:
Ach yeah it's annoying that the `SettingsConnectionMatcher` needs to know about defaults seeing as the entire concept is kind of "hidden" due to that being a detail of `SharedPreferencesSettings`. We could have `InMemSettings` return the same defaults, but I'm a little wary of everything having to reference the same constant. I'm thinking there are a couple of different ways we could go to make things simpler:

* Have our `Settings#getX()` methods take defaults as a second arg. This would mean that `SettingsConnectionMatcher` could just not specify defaults when access server URL/protocol and it wouldn't need to know about our `ProjectKeys.defaults`.
* Add a `isDefault` method to `Settings` to expose whether a key's value is a default or not. This would allow `SettingsConnectionMatcher` to match missing values in the json with default values.

I think we could look at one of those for this PR, but also happy to just go with what we've got for now. I think I prefer the second one. What do you think?

seadowg (commented):
Ach yeah it's annoying that the `SettingsConnectionMatcher` needs to know about defaults seeing as the entire concept is kind of "hidden" due to that being a detail of `SharedPreferencesSettings`. We could have `InMemSettings` return the same defaults, but I'm a little wary of everything having to reference the same constant. I'm thinking there are a couple of different ways we could go to make things simpler:

* Have our `Settings#getX()` methods take defaults as a second arg. This would mean that `SettingsConnectionMatcher` could just not specify defaults when access server URL/protocol and it wouldn't need to know about our `ProjectKeys.defaults`.
* Add a `isDefault` method to `Settings` to expose whether a key's value is a default or not. This would allow `SettingsConnectionMatcher` to match missing values in the json with default values.

I think we could look at one of those for this PR, but also happy to just go with what we've got for now. I think I prefer the second one. What do you think?

grzesiek2010:
> Add a isDefault method to Settings to expose whether a key's value is a default or not.

to achieve this we would need to have a method like that for every single type (String, Boolean Int etc) because you can't create one `isDefault()` method and pass a key there as a parameter since you wouldn't know what method you need to call on shared preferences (`getString()`, `getBoolean()` etc). 

am i right?

Even if we did that am not sure if it would work... comparing that values that come from json with settings that we have saved will always be tricky.
Let's assume we have a default project already saved then the value returned for server url will be: `https://demo.getodk.org` for shared preferences but might be an empty value for `InMemSettings` (if it doesn't use the same defaults). At the same time in json we can have nothing for server url (normal case) but I can also imagine cases where users that create qr codes on their own include that `https://demo.getodk.org` either way. In all the permutations the result of comparing them should be that they are equal.

I wonder if it wouldn't be better to change the whole process of detecting duplicated projects and maybe first save a new project without checking if it's duplicated and then compare all existing projects to detect if the new one is a duplication. If so remove it (if that's what a user wants). Thanks to that we would just compare values from our settings (not from json) what would narrow down that number of permutations). What do you think?

seadowg:
>  to achieve this we would need to have a method like that for every single type (String, Boolean Int etc) because you can't create one isDefault() method and pass a key there as a parameter since you wouldn't know what method you need to call on shared preferences (getString(), getBoolean() etc).

I think it'd be possible to grab the value saved as an `Object` then compare it to the default to see if they're equivalent? It all still feels pretty awkward.

> I wonder if it wouldn't be better to change the whole process of detecting duplicated projects and maybe first save a new project without checking if it's duplicated and then compare all existing projects to detect if the new one is a duplication. If so remove it (if that's what a user wants). Thanks to that we would just compare values from our settings (not from json) what would narrow down that number of permutations). What do you think?

Yeah, the whole way our defaults work is kind of ugly for this connection matcher stuff. I do like that we don't have to save defaults for pretty much everything else! Maybe saving a project is the right thing to do here, although it still doesn't feel quite right. I guess thinking along the same lines, we could alternatively convert the existing project settings to JSON and then use that for the comparison?

grzesiek2010:
> I guess thinking along the same lines, we could alternatively convert the existing project settings to JSON and then use that for the comparison?

still not a perfect solution. According to our implementation we do not include settings that are equal to defaults so if you have server url `https://demo.getodk.org` it won't be included in json that we generate but as I said before I can imagine a case where users that generate qr codes on their own include such defaults. In such a case comparison would return false.

seadowg:
@grzesiek2010 yup. Seems like this is a really awkward one. I think your current solution is probably the best way to go for now. I might make a few tweaks as I know you're out this week.

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- Google Drive project existed -> added Demo project
- Central project existed -> added Demo project
- Demo project existed -> tried to add Demo project  -> Duplicate project dialog appeared
- Central project existed -> tried to add the same Central project -> Duplicate project dialog appeared
- Google Drive project existed -> tried to add Google Drive project -> Duplicate project dialog appeared
- removed Demo project -> added Demo project

kkrawczyk123:

PR #4821, opened 2021-09-03
Fixed attaching gifs in ImageWidget

### LINKED ISSUE #4819: Unable to add gif as picture to image widgets
Collect v2021.2.3 current master, Android v all device used...

#### Problem description
I have noticed that there is no way to add gif as picture in image widgets. I am able to add the same file on store version so it looks like a regression issue.

| Current master: | Store version |
| --- | --- |
| Screenshot_20210902-132643 | Screenshot_20210902-133117 |

1. Open any image widget
2. Try to add gif as image
3. Except picture attached to widget you will see information that selected file is not a valid picture

#### Expected behavior
User should be able to upload gif into image widget/annotate widget. 

#### Other information 
file that I have used for testing: 
giftumb500

grzesiek2010:
Closes #4819 

I added automated tests and verified the fix manually.

The issue occurred after upgrading `androidx.exifinterface:exifinterface` I don't think that point release brought anything that we would need to apply on our side so I guess there must be a bug in that newest release. I downgraded that dependency and we can check it again after next release and then if it still doesn't work we can spend more time on it.

Generally the problem is that we set exif attributes no matter what image format is used. Gifs are not supported but it should not cause spoiling those files as well. At least it didn't in the previous release of that dependency. I also didn't want to make to many changes to avoid conflicts with https://github.com/getodk/collect/pull/4664 so yeah it's better to check it again later and for now just fix the problem in the easiest possible way.

It's a safe fix since I just downgraded one dependency to a version that we used before so we just need to focus on the described scenario.

Any form with `ImageWidget`.

No.


seadowg (approved):

mmarciniak90:
**Tested with success**

Verified cases:
- added gif to Image widget
- added gif to Annotate widget
- added gif from Google Drive
- added gif from internal storage: Images, Recent, Downloads 
- removed gif
- edited to another gif
- edited to another image format
- added other image formats - regression check

kkrawczyk123:

PR #4823, opened 2021-09-07
Use targetSdkVersion for Robolectric

### LINKED ISSUE #4236: Use `targetSdkVersion` for Robolectric
Right now we're specifying SDK 28 for Robolectric to run against (in each module's `robolectric.properties` file). Long term we should really have Robolectric running against our `targetSdkVersion`.

This requires us to install newer (9+) JDKs as part of dev environment set up this will be different for different platforms, but we should work out the simplest way to have Android Studio and the command line share a JDK and document it in the README.

There is a Google Issue open about Android Studio to Java 9+ that might help us here: https://issuetracker.google.com/issues/139093669

seadowg:
Closes #4236 

Ran tests.

Didn't need to do much here other than removing the config we had locking Robolectric to API 28. We were blocked on this because we were using JVM 8 before, but the new Android Studio upgrade has shifted us up to using JVM 11 by default. I was worried this change might result in some problems with Robolectric but everything seems to be working fine.

Just test config changes, so nothing requiring QA here!


download-robolectric-deps.sh
```diff
@@ -2,11 +2,11 @@ set -e
 
 wget -nc https://repo1.maven.org/maven2/org/robolectric/android-all/6.0.1_r3-robolectric-r1/android-all-6.0.1_r3-robolectric-r1.jar -P robolectric-deps
 wget -nc https://repo1.maven.org/maven2/org/robolectric/android-all/7.0.0_r1-robolectric-r1/android-all-7.0.0_r1-robolectric-r1.jar -P robolectric-deps
-wget -nc https://repo1.maven.org/maven2/org/robolectric/android-all/9-robolectric-4913185-2/android-all-9-robolectric-4913185-2.jar -P robolectric-deps
 wget -nc https://repo1.maven.org/maven2/org/robolectric/android-all/10-robolectric-5803371/android-all-10-robolectric-5803371.jar -P robolectric-deps
```
seadowg, line 5: We still need 10 and 11 as our app targets API 29 rather than the latest (30). There is another issue for that.

grzesiek2010 (approved):

Issue #4489, opened 2021-03-24
Augment "Form Download Failed" message with source of error


lognaturel:
Related to https://github.com/getodk/collect/issues/4345. We do swallow too much. 

Quite a few of these errors are because of misbehaving/incompatible servers and that is a nice courtesy to provide more info about.

A recent example we've seen is a CSV attachment with missing column names. It's reasonable for a server not to check this and it's currently unnecessarily difficult to troubleshoot. (related to https://github.com/getodk/javarosa/issues/444)

Mockup: https://www.figma.com/proto/XaN5etTC2Br1AyVJ1IqNi0/Errors?page-id=0%3A1&node-id=2%3A93&viewport=241%2C48%2C0.62&scaling=scale-down&starting-point-node-id=2%3A2

## Acceptance

- [x] Given I've setup Collect with a server with multiple forms
When I download the forms via "Get Blank Forms"
And all downloads succeed
Then I see a message that downloads succeeded

- [x] Given I've setup Collect with a server with multiple forms
When I download the forms via "Get Blank Forms"
And one or more downloads fail
Then I see a message that tells me how many downloads failed

- [x] Given I've setup Collect with a server with multiple forms
When I download the forms via "Get Blank Forms"
And one or more downloads fail
And I click "Show details"
Then I see a list of the forms (and their ID/version) that failed to download
And each form has a message corresponding to the error that occurred

## Notes

Right now the `FormDownloader` just throws `FormDownloadException` with no real message. We should move to including a `FormSourceException` in the `FormDownloadException` or create a similar mapping of different subtypes for different scenarios so we get more specific error messages when a download fails.

We're very likely going to want to use the "Errors" screen here in other cases (like match exactly syncs and form uploads) so it probably needs to be an `Activity` (rather than a full screen dialog or other fragment) and it would make sense for it to take a list of exceptions (with title/subtitles) in via its `Intent`.

We originally looked at making the error cards expand to show the stack trace, but we're not sure if we need that. We'll see if QA can find any cases that might move that back into scope.

seadowg:
We need to discuss exactly how we solve this. One idea has been to include an expandable section in the error dialog that shows the exception message.

seadowg:
It would be good to close #1703 as part of this.

seadowg:
[...]

PR #4824, opened 2021-09-20
Return an exception with a meaningful message for all errors that might occur during downloading forms

grzesiek2010:
Work towards #4489 

I added some automated test and tested the changes manually hardcoding some exceptions.

The problem was that the errors that might occur during downloading forms were turned into generic `FormDownloadException` so even if `FormSourceException`[ was caught with its message it was turned into `FormDownloadException` with no message](https://github.com/getodk/collect/pull/4824/files#diff-edbeccead389ffa8df15f518180a4aefdfa0f7fc236abd975b86dc3920d41c3cL116).
I changed that and also added some new errors like: https://github.com/getodk/collect/pull/4824/files#diff-4237726580b2c992e7730f2d4c0e0d9c5899d97c7c263bb8f905906f11ba6189R15 so that every exception has its own message.

I decided also  to remove `FormDownloadException`  and add those new exceptions in `FormSourceException` becasue to me it makes more sense so have one hierarchy of exceptions related to downloading forms.

It doesn't require testing, let's test it together with UI changes.

No.

No.


forms/src/main/java/org/odk/collect/forms/FormSourceException.kt
```diff
@@ -0,0 +1,20 @@
...
+    class FetchError : FormSourceException()
+    class SecurityError(val serverUrl: String) : FormSourceException()
+    class ServerError(val statusCode: Int, val serverUrl: String) : FormSourceException()
+    class ParseError(val serverUrl: String) : FormSourceException()
+
+    // Aggregate 0.9 and prior used a custom API before the OpenRosa standard was in place. Aggregate continued
+    // to provide this response to HTTP requests so some custom servers tried to implement it.
+    class ServerNotOpenRosaError : FormSourceException()
+
+    class DownloadingInterruptedException : FormSourceException()
```
seadowg: Looking at this now, it does feel a little strange to me that `FormSourceException` includes the download specific errors. This way these exceptions are declared in `forms` but never come out of any components defined in there. Do you think we could move back to a `FormDownloadException` hierarchy for the new errors and just have `FormDownloader` throw both? I still think having one mapper makes sense, so it could have a `getMessage` for both super types.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
```diff
@@ -180,7 +180,7 @@ public void whenFormListMissingHash_throwsError() throws Exception {
         try {
             downloader.downloadForm(serverFormDetails, null, null);
             fail("Expected exception because of missing form hash");
-        } catch (FormDownloadException e) {
+        } catch (FormSourceException e) {
```
seadowg: We could make this more specific now and just catch a `FormWithNoHashException`. Same goes for the other `catch (FormSourceException e)` tests below.

forms/src/main/java/org/odk/collect/forms/FormSourceException.kt
seadowg: Naming wise, I don't think these need the `...Exception` postfix as they are all inner classes, i.e. they'll be addressed as `FormSourceException.X`.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormDownloaderTest.java
grzesiek2010: Fixed.

forms/src/main/java/org/odk/collect/forms/FormSourceException.kt
grzesiek2010: > it does feel a little strange to me that FormSourceException includes the download specific errors

hmm now as I think about it maybe you are right. I'm not strongly convinced but let's do that.

> Naming wise, I don't think these need the ...Exception postfix as they are all inner classes, i.e. they'll be addressed as FormSourceException.X.

Fixed.

seadowg (approved):
This will definitely add a lot more detail to the download errors. Let's make sure to review the error text when we have the new UI for displaying these.

Going to wait on merging this until #4803 is merged so we can include that PR in a beta without this.

PR #4826, opened 2021-09-21
Resolves #4500 autocomplete minimal…

### LINKED ISSUE #4500: autocomplete minimal not working as expected with select_one_external

dimwight:
Resolves #4500 

This documents a revision of #4712 now closed (like its predecessor).

Tested with new `SelectOneResetTest`.

See 'Design notes' below.

More logical behaviour from user perspective. No apparent regression risks. 

_[selectOneReset.xlsx](https://github.com/getodk/collect/files/7202034/selectOneReset.xlsx)_

Perhaps form design advice could specify the benefits and limitations of the improved behaviour?

# Design notes
A lot of stuff here, as both [the issue](https://github.com/getodk/collect/issues/4500) and the proposed resolution have turned out to be quite complex - though the resolution has been progressively simplified in revisions of this PR. 

## The issue
A shout out to @aurdipas for a [form](https://github.com/getodk/collect/files/6246387/Entrevista_Nacimientos_Etnicos_filtro.xlsx) that demonstrates clearly the undesired behaviour of selects with fast external itemsets. 
- Most obviously, answers for cascades using fast external itemsets don't reset (or at best not immediately) when preceding members are updated, in constrast to their behaviour with internal itemsets. 

- This behaviour is most glaringly apparent with minimal appearances, especially in field lists - plain select widgets do reset when they can’t match an existing value in their current choices.

Desirable if less obvious behaviours also implied by the form are that

- Members should reset even when unrelated fields are interpolated in the cascade.
- Members hidden by an interpolated select when non-relevant should be reset once unhidden.

And all this should work inside groups and especially field lists; the hierarchy views should be reset immediately upon each change.

Not only may the current behaviour confuse the user, it permits saving of incoherent data.

So a resolution to this issue must focus on resetting the answers for a complete cascade, rather than just ensuring that widgets reset when rendered. 

#### Preparatory changes 
_Unchanged in this revision_

In `SelectOneWidget`, `clearFollowingItemsetWidgets` sort of resets succeding cascade members, but hitherto there’s been no equivalent in `SelectOneMinimalWidget`. 

So to start with I recycled `clearFollowingItemsetWidgets`:
- Pulled it up to `ItemsWidget` 
- Called it from `SelectOneMinimalWidget` by analogy with `SelectOneWidget`

#### Widget sequences (one per view)
_Updated in this revision_

`SelectOneWidgetUtils` now has `checkFastExternalCascade`, called from a stub `clearFollowingItemsetWidgets` and itself calling the private `doCascadeCheck`.

This last method extends the algorithm from the current `clearFollowingItemsetWidgets` by
- detecting fast external cascade members from their query strings
- quitting at a non-question or on encountering a new cascade
- skipping interpolated questions - including unrelated selects as in the issue form

Like the code it derives from, it has the helpful side-effect of resetting questions that are about to be hidden by an interpolated select.

#### Field lists
_Updated in this revision_

The new `checkFastExternalCascade` aborts silently when the cascade is in a field list as in the issue form; in the current `FormEntryActivity`, `updateFieldListQuestions` resets only the next cascade member (possibly as a side effect?). 

In this improvement `FormEntryActivity` calls `checkFastExternalCascadeInFieldList`, also in `SelectOneWidgetUtils` and also (as of this revision) calling `doCascadeCheck`; of which the side-effect in a field list is that hidden questions are reset when _unhidden_.

## Testing
_Updated for this revision_

The issue appears resolved as outlined above when changes are tested manually against the issue form (when [slightly modified ](https://github.com/getodk/collect/files/7202062/Entrevista_Nacimientos_Etnicos_filtro%2B.xlsx) with no cascade splitting between groups or nested field list).

For formal testing there are a number of use cases to consider, representing combinations of:
- 2 types of itemset: internal (as a baseline and to validate the testing code) and fast external
- 4 appearances: plain, minimal, minimal autocomplete and (not used in practice?) plain autocomplete
- 2 contexts: widget sequences and field lists

### Test form
The [form](https://github.com/getodk/collect/files/7202034/selectOneReset.xlsx) derives from _[selectOneExternal.xslx]( https://github.com/getodk/collect/files/5630004/selectOneExternal.zip)_ from #4258 - thanks @mmarciniak90. It covers
- four cascade members - seems the maximum in practice
- interpolated fields - actually relevance selects before last members as in the issue form
- use cases broken up into
  - sections covering each combination of itemset and appearance 
  - each section split into blocks for each context

The workbook includes these supporting sheets:

- The _variants_ sheet defines section numbers in terms of the two itemset types and the two essentially different appearances.
- The _asserts_ sheet records results of testing both before and after the update.
- The _master_ sheet defines two functionally identical context blocks:
  - _A_ - widget sequence (one per view)
  - _B_ - field-list
- It includes two alternative **type** and four **appearance** columns, and prefills **defaults** ready for testing of reset behaviour.
- These are propagated in the _survey_ sheet into sections testing the various combinations of appearances and itemsets.
- For this draft PR the form implements just the four core variants:
  - 0 and 1 - internal itemsets with plain and minimal appearances
  - 2 and 3 - fast external itemsets

### Test class
`SelectOneResetTest` borrows extensively from `FieldListUpdateTest`. In each block it makes changes and asserts expected outcomes as set out in the workbook _asserts_ sheet. 

- The `SectionVariant` `enum` combines an `itemset` based on the **type** of its section with an `appearance` from the **appearance**.  

- The `Block` `enum` provides the appropriate strings for each `SectionVariant`.

The single test method iterates through the variants, thus visiting all sections of the prefilled form and running logically equivalent tests on the context blocks comprising each section. 

Tests are identified in the code as detailed in the workbook _asserts_ sheet. The test code makes allowance for the failure of standard selects to reset on hide/unhide (asserts of type `(A|B)1(e|h)`).

The code makes use of new convenience methods in
`FormEntryPage` that wrap useful formulations found in `FieldListUpdateTest`. 

`SelectOneResetTest` can be validated by running against the current codebase and checking that it fails where expected. Its progress can be monitored by filtering the log on 'SelectOne' which also picks up the traces from `SelectOneWidgetUtils`.


grzesiek2010:
@dimwight 
My apologies for the delayed review!

I would like to take a step back first and discuss that whole problem...

I've never liked the fact that we needed to deal with clearing cascading selects in ODK Collect because:
- the current solution on master is pretty dummy because it assumes cascading selects is a set of questions placed one by one so if we had something between those selects (another question) rest of the levels wouldn't be cleared. On the other hand to much might be cleared if we placed selects that belong to another cascading select just after our first cascading select...
- we have had a few problems maintaining that code or using it properly so far as I remember
- it makes dealing with selects more complex

and furthermore `Cascading Selects` is an abstraction. There is no `cascading select` appearance or group it is just a set of selects that use `choice-filter` that we name `Cascading Select` so trying to detect a set of question as cascading select will always be error prone.

Taking all of what I described above into account... how about we stop thinking
_ohh this group of questions is cascading selects we need to handle clearing properly._

We can handle that clearing on a form level using [triggers](https://docs.getodk.org/form-logic/#triggering-calculations-on-value-change). 
If we recommended that solution we could remove our faulty code.
Plus it would be even better because it will work with all appearances and with mixed cascading selects (not only those where questions are placed one by one like in the sample in the documentation)

Here is a sample form I have built: [CascadingSelectWithExternalItemsets.xlsx](https://github.com/getodk/collect/files/7743824/CascadingSelectWithExternalItemsets.xlsx)

What do you think about it @lognaturel @seadowg

dimwight:
I must reluctantly admit @grzesiek2010 that this is an intriguing approach, even if it threatens to render my labours redundant.



Perhaps you could provide a further example that reworks my [revised issue form](https://github.com/getodk/collect/files/7202062/Entrevista_Nacimientos_Etnicos_filtro%2B.xlsx) so as to resolve the issues raised by @aurdipas?

grzesiek2010:
> Perhaps you could provide a further example that reworks my revised issue form so as to resolve the issues raised by @aurdipas?

Yes, this should work well:
[Entrevista_Nacimientos_Etnicos_filtro+.xlsx](https://github.com/getodk/collect/files/7801333/Entrevista_Nacimientos_Etnicos_filtro%2B.xlsx)

I've highlighted my changes in yellow.

dimwight:
> Yes, this should work well: [Entrevista_Nacimientos_Etnicos_filtro+.xlsx](https://github.com/getodk/collect/files/7801333/Entrevista_Nacimientos_Etnicos_filtro%2B.xlsx)
> 
> I've highlighted my changes in yellow.

On a quick check this does indeed work and even better than my approach - when hidden questions are revealed they only reset if invalid. 

And the coding seems elegantly simple - just copy from the choice_filter to the trigger column.

Just to double-check I'll try updating my test form and class.

dimwight:
> Just to double-check I'll try updating my test form…

Indeed this all seems to work fine, matching the behaviour with internal itemsets.

Since my day job is technical writing I'd be happy to update the documentation if someone can suggest where.

grzesiek2010:
@dimwight 
thanks for testing my solution. Let's wait for @lognaturel to see what she thinks.

seadowg:
Apologies that we haven't been able to look at this @dimwight. I'm going to close this for the moment, but feel to reopen if you want to keep working on it.

Issue #4768, opened 2021-08-01
Provide a path for external gps sensors to report accuracy


lognaturel:
As of Collect v1.30, the accuracy of points coming from the mock location provider is always submitted as 0. This was intended to identify points coming from location mocking apps. It turns out that external location sensors provide location through such apps. As far as I know, there’s no way to differentiate between the two but that would be a good thing to confirm.
https://forum.getodk.org/t/show-accuracy-for-gps-mock-location-for-external-sensors/34455/2

The discussion on exactly what to do should happen on the forum. This issue is a placeholder so we make sure it’s addressed.

seadowg:
Ok to solve this we're going to introduce an XLSForm parameter for geo questions (`geopoint`, `geotrace` and `geoshape`):

| type | name | label | parameters |
|------------ |------------ |------------ |-----------------------| 
| geopoint | location | Location | allow-mock-accuracy=true |

When the `allow-mock-accuracy` parameter is `true`, the client should allow “mock” locations. In Collect, this will mean that locations from Android’s mock providers will use the reported accuracy. The default for `allow-mock-accuracy` will be `false` so this behaviour will be “opt-in” and by default, “mock” locations will still have an accuracy of 0. This parameter will be represented in an XForm as a bind attribute `odk:allow-mock-accuracy` for the relevant data types (which will require changes to pyxform).

yanokwa:
The alternative I had suggested to @seadowg is that we revert the behavior from the previous release and use whatever accuracy the mock provider reports. Then we add a parameter of `allow-mock-providers` and default that to `true.` And if a manager sets it to `false`, we prevent anyone from using mock providers entirely.

@lognaturel's pushback about my alternative is that: 
> most people don’t want mock providers, don’t want data from them, and will never know that they are being used. That was why the issue was filed in the first place. Setting the accuracy to 0 means there’s always a marker of something going wrong for the common case.

@seadowg was also concerned:
> Let me think on it. I’m also not sure about our UI support for something like that. I think right now if you were banning mock locations you would end up with an infinite spinner on the geopoint dialog.

I discussed this with @lognaturel and I'm OK proceeding. To summarize...

With this implementation, if an enumerator uses a mock provider, we set accuracy to 0 so managers know there might be fake data. If they want to change this behavior, they can use `allow-mock-accuracy=true`. If they want to prevent data collectors from using mock providers, they can use a constraint of `selected-at(.,3) > 0`

PR #4827, opened 2021-09-21
Move ToastUtils to androidshared

seadowg:
This is prep for #4768 (need to move some files that depend on `ToastUtils` between modules)

Ran tests

I did consider just deprecating `ToastUtils` and removing it from the files that I needed to move, but after discussing with @grzesiek2010 we decided to hold on to it. The main reason for this is that we believe the ability to hide the last toast when displaying a new one is helping keep our UI tests stable.

Shouldn't be any change in behaviour as it's just moving code around. Don't think we need QA on this.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
@@ -42,8 +42,6 @@ import org.odk.collect.android.formmanagement.FormDownloader;
 import org.odk.collect.android.formmanagement.FormSourceExceptionMapper;
 import org.odk.collect.android.formmanagement.ServerFormDetails;
 import org.odk.collect.android.formmanagement.ServerFormsDetailsFetcher;
-import org.odk.collect.android.views.DayNightProgressDialog;
-import org.odk.collect.forms.FormSourceException;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.listeners.DownloadFormsTaskListener;
 import org.odk.collect.android.listeners.FormListDownloaderListener;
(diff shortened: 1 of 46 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
```diff
@@ -270,7 +270,7 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
      */
     private void downloadFormList() {
         if (!connectivityProvider.isDeviceOnline()) {
-            ToastUtils.showShortToast(R.string.no_connection);
+            ToastUtils.showShortToast(Collect.getInstance(), R.string.no_connection);
```
grzesiek2010: Why not `this`? You do that in many places so I'm not going to comment every single one. It doesn't have to be application context right? So why do you still using `Collect.getInstance()` when we wanted to get rid of it?

You use it to pass as a parameter calling `makeToast` but the documentation says:

` The context to use. Usually your android.app.Application or android.app.Activity object.`

so it doesn't have to be application context.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
seadowg: Ah I used refactoring tools to pull that parameter out. I went through and changed a few of the `Collect.getInstance()` calls that called pulled up to whatever local contexts there was (like in `GeoPolyActivity for instance), but I ~~probably missed a few~~ missed a lot. I'll go through it again.

grzesiek2010 (approved):

PR #4829, opened 2021-09-23
Add allow-mock-accuracy for default geopoint widget

seadowg:
Work towards #4768
~~Blocked by #4827~~
~~Blocked by https://github.com/XLSForm/pyxform/pull/560~~

This adds support for the new `odk:allow-mock-accuracy` bind attribute to `geopoint` questions. It doesn't include support when the `map` or `placement-map` appearance are in use. Those (and added support in `geoshape` and `geotrace`) will be covered in a follow-up PR and we can use this to validate the pyxform implementation and also to review the general structure of how `allow-mock-accuracy` works within the codebase.

As part of this work, I pulled out the `GeoPointActivity` (and some helpers) to a new `geo` module. It seemed like there was a clear seam between the form entry code and the geo location code. Pulling this out lets us keep that seam in place. I'll do the same for the other geo Activity objects when I get to them. I could have included these components in `location` but it felt to me like there was a nice (almost backend/frontend) split between "getting the device's current location in code" and "UI for allowing the user to grab location(s)".

I would recommend reviewing this commit by commit, as I've tried to stage the refactor (moving the geo code) into many commits and have included justification for some changes in commit messages.

Lots of new tests and verified using a handwritten XForm that accuracy was being retained when `allow-mock-accuracy` is `true`.

Comments inline and in commit messages.

Should just add support for the new attribute. I did a bunch of refactoring to make this easier for myself so it'd be good to check the geopoint widget (with no appearance) here.

To test the new feature I've created [a form](https://github.com/getodk/collect/files/7285696/external_gevice_geopoint.zip) (I've included the XSLX and the XForm) with the merged changes in XLSForm/pyxform#560 that has three `geopoint` questions: one with the new attribute set to true, one with it set to false and one without the attribute.


collect_app/src/main/AndroidManifest.xml
```diff
@@ -133,7 +133,10 @@ the specific language governing permissions and limitations under the License.
         <activity android:name=".preferences.screens.ProjectPreferencesActivity" />
         <activity android:name=".activities.FormHierarchyActivity" />
         <activity android:name=".activities.ViewOnlyFormHierarchyActivity" />
-        <activity android:name=".activities.GeoPointActivity" />
+        <activity
+            android:name="org.odk.collect.geo.GeoPointActivity"
+            android:theme="@style/Theme.Collect"
+            tools:replace="android:theme"/>
```
seadowg, line 140: This is a pretty nifty bit of support for Android libraries that lets us override manifest values and specify priority - here `collect_app` is saying it overrides whatever the original `GeoPointActivity` theme is.

collect_app/src/main/java/org/odk/collect/android/utilities/ReferenceLayerUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import android.os.Bundle
+import org.odk.collect.android.geo.MapFragment
+import org.odk.collect.shared.PathUtils.getAbsoluteFilePath
+import java.io.File
+
+object ReferenceLayerUtils {
+
+    @JvmStatic
+    fun getReferenceLayerFile(config: Bundle, layersPath: String?): File? {
```
seadowg, line 11: This method wouldn't make sense in `geo` as it feels like it is more "collect specific".

location/src/main/java/org/odk/collect/location/tracker/LocationTracker.kt
```diff
@@ -16,7 +16,8 @@ interface LocationTracker {
     /**
      * Starts tracking the device's location at a fixed interval.
      */
-    fun start()
+    fun start(retainMockAccuracy: Boolean)
+    fun start() = start(false)
```
seadowg, line 20: No @JvmOverloads for interfaces 😭

geo/src/main/res/layout/geopoint_dialog.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
seadowg, line 2: This layout is mostly stolen from `progress_dialog.xml`. As we've talked about revising this interface (and I wanted to move the layout to another module) I thought it made sense to just create a new layout for this.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -65,7 +65,7 @@ public static String getGeoPolyAnswerToDisplay(String answer) {
 
     static double getAccuracyThreshold(QuestionDef questionDef) {
         // Determine the accuracy threshold to use.
-        String acc = questionDef.getAdditionalAttribute(null, ActivityGeoDataRequester.ACCURACY_THRESHOLD);
+        String acc = questionDef.getAdditionalAttribute(null, "accuracyThreshold");
```
seadowg, line 68: This string only happened to be the same before. Could definitely pull it out as a constant, but we're only using the attribute in non-test code here.

geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
```diff
@@ -94,15 +88,6 @@ protected void onCreate(Bundle savedInstanceState) {
             numberOfAvailableSatellites = savedInstanceState.getInt(NUMBER_OF_AVAILABLE_SATELLITES);
         }
 
-        Intent intent = getIntent();
-
-        targetAccuracy = DEFAULT_LOCATION_ACCURACY;
```
seadowg, line 99: This use of the default was redundant, as we pass the default if no accuracy has been specified through the extra. It felt like the accuracy default was  a `collect_app` concept rather than a `geo` concept, so now the Activity just defaults to the max `Double` value meaning there is effectively no threshold.

geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
```diff
@@ -245,7 +231,7 @@ private void returnLocation() {
         if (location != null) {
             Intent i = new Intent();
 
-            i.putExtra(FormEntryActivity.ANSWER_KEY, getResultStringForLocation(location));
+            i.putExtra("value", getResultStringForLocation(location));
```
seadowg, line 234: This is a little strange. We can't reference `FormEntryActivity.ANSWER_KEY` from `geo` so need to just use an inline value. IMO, the `GeoPointActivity` should really be defining the intent it responds with, but this would buck the established pattern that `"value"` is used as the key for all these kinds of Activity objects.

geo/src/test/java/org/odk/collect/geo/GeoPointActivityTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    LocationClient locationClient = mock(LocationClient.class);
+
+    @Before
+    public void setUp() throws Exception {
+        LocationClientProvider.setTestClient(locationClient);
+    }
+
+    @Test
+    public void testLocationClientLifecycle() {
```
seadowg, line 40: I'd like to rewrite this test, but it felt like moving these files around was enough of a time sink. Another day...

seadowg:
Ach looks like I broke custom language support for the geopoint dialog. Will move this back to a draft.

seadowg:
@kkrawczyk123 @mmarciniak90 marking this ready for QA now. Good to test this out if we can even though we're waiting on a review.

collect_app/src/main/java/org/odk/collect/android/activities/GeoPointActivity.java
```diff
@@ -92,10 +91,9 @@ protected void onCreate(Bundle savedInstanceState) {
 
         Intent intent = getIntent();
 
-        targetAccuracy = DEFAULT_LOCATION_ACCURACY;
         if (intent != null && intent.getExtras() != null) {
             if (intent.hasExtra(ACCURACY_THRESHOLD)) {
-                targetAccuracy = intent.getDoubleExtra(ACCURACY_THRESHOLD, DEFAULT_LOCATION_ACCURACY);
+                targetAccuracy = intent.getDoubleExtra(ACCURACY_THRESHOLD, Double.MAX_VALUE);
```
lognaturel: ~Doesn't this change the behavior? Previously by default if 5m accuracy were reached then the value would be automatically committed. Now it looks to me like any value will be taken immediately unless a threshold is specified.~ I got confused by your comment about this but now I see that https://github.com/getodk/collect/pull/4829/files#diff-88c9ec98209d666bd2d8e9d8c3bb2f0523c6019fac83a786699f7f85124e9a5bR69 was already in place and ensures that from within Collect a threshold is always passed.

lognaturel (commented):
Overall this looks great with some nice bonus cleanup. ~I haven't yet had a chance to try it but I suspect that forms without an explicit target accuracy now have different behavior.~

lognaturel (approved):

kkrawczyk123:
@seadowg  We probably need some advices about how to test this one. I tried to use mobile phones ad external GPS devices but probably they are not enough or I have done something wrong, I have problem with connecting devices:
Screenshot_20211007-122232


The same toast displayed when tried on PR and store version.
How to test it without external GPS device? Should we do general regression checks on Geo widgets?

seadowg:
@kkrawczyk123 sorry I meant to include that in the PR description but forgot. Unless you have an actual external GPS device (which I'm assuming you don't) I'd recommend just testing with an app that mocks your location as it uses the same mechanism under the hood (an app is being set up as a mock provider in the developer settings). I can test with an external device (that a user sent me) using an app like Bluetooth GPS. 

I think the main thing to check here is really that we haven't broken any existing functionality as the new feature isn't that much of a risk - moving things around to enable it to all work was however.

kkrawczyk123:
**Tested with success!**

Verified cases:
regression checks on:
- Geo widgets with/without map
- GeoTrace, GeoShape widgets
- display submissions on map
- submissions with different statuses
- Google maps
- OSM
- Mapbox: when testing regression issue has been detected and reported in 4851
- offline layers
- light mode, dark mode
- Audit location in background
- different background location forms 
- form attached in PR
- screen rotation
- minimize app
- Geo widgets in read-only 
- fake GPS app

mmarciniak90:
Verified also on Android versions: 8.1, 9.0, 10.0

PR #4830, opened 2021-09-23
Improved showing errors that might occur during downloading forms

### LINKED ISSUE #4489: Augment "Form Download Failed" message with source of error

grzesiek2010:
Closes #4489

I added automated tests and tested the change manually.

This is what we discussed in the issue. It looks like:
Screenshot_1632403941

Please notice that compared to the mockups attached to the issue errors are not expandable. We discussed it with @seadowg and we agreed that we can implement it in the future IF there are errors that require adding stacktraces, for now every error seems to have a meaningful message. If you are able to reproduce errors that are cryptic @mmarciniak90 @kkrawczyk123 please let us know.

@seadowg please remember reviewing those new messages. And another thing... we discussed moving those new classes to a new module called `errors` but I'm not totally convinced... those are form downloading errors so I'm not sure if mixing them in one module with other errors is a good idea so I just put them in once package for now.

We need to test the result we get after downloading forms manually. Only this part of the app might be affected.

No.

No.


collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+import android.content.Context
+import android.os.Bundle
+import androidx.appcompat.app.AlertDialog
+import androidx.fragment.app.DialogFragment
+import org.odk.collect.android.R
+import org.odk.collect.android.formmanagement.ServerFormDetails
+
+class FormDownloadResultDialog : DialogFragment() {
+    companion object {
+        const val RESULT_KEY = "RESULT"
```
seadowg: Generally in Android these are `<kind of key>_<key name>` so this should really be ARG_RESULT. I think sticking the companion at the bottom is more idiomatic from what I've seen as well.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private lateinit var result: HashMap<ServerFormDetails, String>
+    private lateinit var listener: FormDownloadResultDialogListener
```
seadowg: I think the listener should be a `?` type rather than `lateinit` here. We always expect `result` to be assigned, so it makes sense as `lateinit` but `listener`  should be optional, as we only assign it conditionally (if the context implements the interface we want).

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    interface FormDownloadResultDialogListener {
```
seadowg: Let's stick this at the bottom as well.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        if (context is FormDownloadResultDialogListener) {
+            listener = context
+        }
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        isCancelable = false
```
seadowg: This can just be cancellable right? I don't think there is a problem with the user dismissing it by clicking outside.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        result = arguments?.getSerializable(RESULT_KEY) as HashMap<ServerFormDetails, String>
+
+        val builder = AlertDialog.Builder(requireContext())
```
seadowg: This should be a `MaterialAlertDialogBuilder`

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDetails.kt
```diff
@@ -0,0 +1,33 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.odk.collect.android.formmanagement
+
+import org.odk.collect.forms.ManifestFile
+import java.io.Serializable
+
+data class ServerFormDetails(
+    val formName: String?,
```
seadowg, line 20: I think we need to review this class and the others here (`ManifestFile` and `MediaFile`) to make sure we have the right types for each field. I'm guessing a bunch of them can be `String` instead of `String?`. Makes sense we end up with that because in Java they're all technically optional.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -56,4 +57,9 @@ class FormDownloadResultDialog : DialogFragment() {
 
     private fun countFailures() =
         result.values.count { it != getString(R.string.success) }
+
+    @TestOnly
```
seadowg: I think it's fine to just make the `listener` field `public` here instead of having a `@TestOnly` setter. It'd make sense that a parent might want to change the listener a dialog uses.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialogTest.kt
```diff
@@ -0,0 +1,203 @@
...
+        private val DUMMY_SERVER_FORM_DETAILS_1 = ServerFormDetails("1", "", "", "", "", false, false, ManifestFile("", emptyList()))
+        private val DUMMY_SERVER_FORM_DETAILS_2 = ServerFormDetails("2", "", "", "", "", false, false, ManifestFile("", emptyList()))
+        private const val DUMMY_ERROR_MSG = "error"
+        private val SUCCESS_MSG = ApplicationProvider.getApplicationContext<Collect>().getString(R.string.success)
+    }
+
+    val listener = mock<FormDownloadResultDialog.FormDownloadResultDialogListener>()
+
+    @Test
+    fun `The dialog should have no title`() {
```
seadowg: I don't know if we need this test. I think if we use Callum Stott's Lazy Programmer Principle™️ (i.e. people only write code they need to) we can assume that the dialog won't have a title unless there is a test enforcing that it does.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.fragment.app.DialogFragment
+import org.odk.collect.android.R
+import org.odk.collect.android.formmanagement.ServerFormDetails
+
+class FormDownloadResultDialog : DialogFragment() {
+    companion object {
+        const val RESULT_KEY = "RESULT"
+    }
+
+    private lateinit var result: HashMap<ServerFormDetails, String>
```
seadowg: It really feels like the results should be passed as a `List<Pair<..>>` instead of a `Map`. What do you think?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormsDownloadResultPage.kt
```diff
@@ -0,0 +1,32 @@
...
+        return this
+    }
+
+    fun assertMessage(message: String): FormsDownloadResultPage {
+        assertText(message)
+        return this
+    }
+
+    fun clickOK(): MainMenuPage {
+        clickOnId(android.R.id.button1)
```
seadowg: We have a helper that does this (same thing under the hood) you can switch to here: `clickOnOnDialog`

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList
+
+class FormsDownloadResultDialog : DialogFragment() {
+    companion object {
+        const val FAILURES = "FAILURES"
+        const val NUMBER_OF_ALL_FORMS = "NUMBER_OF_ALL_FORMS"
+    }
+
+    private lateinit var failures: ArrayList<FormsDownloadErrorItem>
+    private var numberOfAllForms = 0
```
seadowg, line 16: I know it would be weird in Java but I think because we expect this, and it would be an error not to pass it, we should make this a `lateinit` (like `failures`).

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    private lateinit var listener: FormDownloadResultDialogListener
```
seadowg: I think this can be `?` rather than `lateinit` as we don't **need** it for the dialog to operate correctly, and we only assign it conditionally. Listeners of end up nullable/optional so it also feels idiomatic. It'd also mean in tests you'd only need to set a listener when you wanted to.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        if (context is FormDownloadResultDialogListener) {
+            listener = context
+        }
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        isCancelable = false
```
seadowg: I'm not sure if we need to block the user from dismissing this with back/clicking outside. What do you think?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        failures = arguments?.getSerializable(FAILURES) as ArrayList<FormsDownloadErrorItem>
+        numberOfAllForms = arguments?.getInt(NUMBER_OF_ALL_FORMS)!!
+
+        val builder = AlertDialog.Builder(requireContext())
```
seadowg: This should be `MaterialAlertDialogBuilder`.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+    private fun getMessage(): String {
+        return if (failures.isEmpty()) {
+            getString(R.string.all_downloads_succeeded)
+        } else {
+            getString(R.string.some_downloads_failed, failures.size.toString(), numberOfAllForms.toString())
+        }
+    }
+
+    @TestOnly
+    fun setListener(listener: FormDownloadResultDialogListener) {
```
seadowg: I think you could just make `listener` public here, as it'd make sense that a parent might want to set the listener to something else. That way you just make it part of the objects API rather than needing a `@TestOnly` setter.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.appcompat.app.AlertDialog
+import androidx.fragment.app.DialogFragment
+import org.jetbrains.annotations.TestOnly
+import org.odk.collect.android.R
+import org.odk.collect.android.formmanagement.downloaderror.FormsDownloadErrorActivity
+import org.odk.collect.android.formmanagement.downloaderror.FormsDownloadErrorItem
+import java.util.ArrayList
+
+class FormsDownloadResultDialog : DialogFragment() {
+    companion object {
```
seadowg: I'm more of a fan of interfaces and inner types going at the bottom of the class. What do you think?

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -0,0 +1,74 @@
...
+        const val FAILURES = "FAILURES"
```
seadowg: Generally in the Android SDK key names are structure like `<type of key>_<key name>` so this should be `ARG_FAILURES`. Same goes for the other arg key.

collect_app/src/main/res/layout/activity_forms_download_error.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
seadowg: Should this be a `ConstraintLayout`? I'm still not the biggest fan of them, but it does seem like we're meant to be using them everywhere.

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_margin="@dimen/margin_small"
+    app:cardElevation="2dp"
+    xmlns:tool="http://schemas.android.com/tools">
+
+    <LinearLayout
```
seadowg: Again, this should probably be a `ConstraintLayout` I think? Is the idea that it's ok to use the other layouts when there's no nesting?

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        android:padding="16dp">
```
seadowg: Use `@dimens/margin_standard` here

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_margin="@dimen/margin_small"
+    app:cardElevation="2dp"
```
seadowg: Generally we've using `4dp` elevation. Do you think we should pull out a constant, or does it not really need to be consistent? I'll need to reread the Material Theming stuff.

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:textAppearance="?attr/textAppearanceHeadline6"
+            tool:text="Form 1"/>
+
+        <TextView
+            android:id="@+id/form_details"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="8dp"
```
seadowg: Again, this should just use a `@dimens`

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
...
+            android:textAppearance="?attr/textAppearanceBody2"
+            android:textColor="?android:attr/textColorSecondary"
+            tool:text="ID: form-1 Version: 202122"/>
+
+        <TextView
+            android:id="@+id/error_message"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="16dp"
```
seadowg: Use `@dimens`

collect_app/src/main/res/layout/form_download_error.xml
```diff
@@ -0,0 +1,41 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        android:padding="16dp">
+
+        <TextView
+            android:id="@+id/form_name"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:textAppearance="?attr/textAppearanceHeadline6"
```
seadowg: I feel like I want to tweak the text sizing in the card here, but we can do that when we review the error messages before merging. Good to play around with some different error message and form name lengths.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialogTest.kt
```diff
@@ -0,0 +1,206 @@
...
+import org.odk.collect.testshared.RobolectricHelpers
+import org.robolectric.Shadows
+
+@RunWith(AndroidJUnit4::class)
+class FormsDownloadResultDialogTest {
+    private val downloadErrorItem = FormsDownloadErrorItem("", "", "", "")
+    val listener = mock<FormsDownloadResultDialog.FormDownloadResultDialogListener>()
+
+    @Test
+    fun `The dialog should have no title`() {
```
seadowg: I don't think we need this test. If we apply Callum Stott's Lazy Programmer Principle™️ (i.e. programmers won't write code they don't have to) then there isn't going to be a title unless we have a test that enforces there being one.

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import android.os.Bundle
+import androidx.appcompat.widget.Toolbar
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import org.odk.collect.android.R
+import org.odk.collect.android.activities.CollectAbstractActivity
+import org.odk.collect.android.utilities.TranslationHandler
+
+class FormsDownloadErrorActivity : CollectAbstractActivity() {
```
seadowg: I think we can drop the `FormsDownload` prefix from a few things here as this errors screen doesn't need to be specific to form downloading:

* `FormDownloadsErrorActivity` -> `ErrorListActivity`
* `FormsDownloadErrorAdapter` -> `ErrorAdapter`
* `FormDownloadErrorItem` -> `ErrorItem`

The supporting layouts would then change as well. One change would be that the logic of creating the `"ID: blah Version: blah"` string would live outside the adapter, but I think that's nice personally. In that way we separate all knowledge of forms out of this code which makes it resusable and also means it won't be effected by any changes we make to forms.

If that was the structure in place - a more generic "errors" UI - do you think it would convince you that this could live in a separate module?

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+    companion object {
+        const val FAILURES = "FAILURES"
```
seadowg: I think this should be `EXTRA_FAILURES`.

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_forms_download_error)
+        initToolbar(TranslationHandler.getString(this, R.string.errors))
```
seadowg: In Kotlin you can just do `context.getLocalizedString` instead of needing to use the `TranslationHandler`. I think there's a couple of other places you could use this.

seadowg (changes requested):
Mostly looking great! It's really nice seeing this working. Most of the comments are just small tweaks I can take care of if you don't have time, but it would be good to make sure we discuss the [naming of the new classes (and potential separate module)](https://github.com/getodk/collect/pull/4830#discussion_r715498850) and the [dialog behaviour](https://github.com/getodk/collect/pull/4830#discussion_r715464614).

I'll come back to reviewing strings before merging and loop someone else in to help me.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: I'm indifferent, I implemented it like that because the old dialog was not cancellable too. If you want it to be cancellable I'm fine with it.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialogTest.kt
grzesiek2010: ok, is it a part of a book or online course somewhere? :smile:  I have two weeks off ahead of me so I could read something :smile:

collect_app/src/main/res/layout/form_download_error.xml
grzesiek2010: The default value was 1dp and I found it not enough. 2dp looks better but I can try 4 too. Where is that specified?

collect_app/src/main/res/layout/form_download_error.xml
grzesiek2010: I just copied this layout from https://material.io/components/cards/android#card

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialogTest.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/res/layout/form_download_error.xml
grzesiek2010: Fixed.

collect_app/src/main/res/layout/form_download_error.xml
grzesiek2010: Fixed.

collect_app/src/main/res/layout/form_download_error.xml
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: It has been changed and it uses a list of `FormsDownloadErrorItem`

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: It's not possible to use `lateinit` with primitive types.

collect_app/src/test/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialogTest.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
grzesiek2010: Fixed.

collect_app/src/main/res/layout/activity_forms_download_error.xml
grzesiek2010: There is a problem with including our toolbar (which uses `<merge>`) in constraint layouts. I think it should be addressed in a separate issue what I will do. https://stackoverflow.com/questions/46447621/using-include-with-merge-in-constraintlayout

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormDownloadResultDialog.kt
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormsDownloadResultPage.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/formmanagement/downloaderror/FormsDownloadErrorActivity.kt
grzesiek2010: Ok I agree with that. I implemented those naming changes and even tried to move the classes to a separate module. The problem is that it requires a lot of changes. One of those classes is an activity (ErrorListActivity) so we will need to move things like our toolbar, maybe `CollectAbstractActivity`, `ThemeUtils` to `androidshared` in order to use it. I feel like I don't have enough time today and it would be even better to do that in a separate pr later because it will be a pretty big pr.

seadowg: Ah fair enough! I think we can probably move the Activity without moving `CollectAbstractActivity` and `ThemeUtils` but it would probably involve pulling a couple of helpers out and it'd be good for us to chat about where things like the toolbar live as you say. If I get a moment next week (once this is merged) I'll do a quick sketch of how we could structure it, and then we can chat about it when you get back 👍

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
seadowg: Huh. Well there you go then.

collect_app/src/main/res/layout/form_download_error.xml
seadowg: Cool. I'm not overly bothered by it. This makes more sense as a `LinearLayout` to me anyway.

collect_app/src/main/res/layout/form_download_error.xml
seadowg: I don't think we have it specified anywhere, it just seems we've been using 4dp a lot of places. I can play with it when we review error messages and maybe tweak text sizes.

seadowg (approved):
I'm really happy with this! I'll do a review of the error message text itself and maybe tweak some of them. As discussed, we might also make some quick design tweaks while doing that, but then I'll hand this over to QA.

strings/src/main/res/values/strings.xml
```diff
@@ -628,19 +628,45 @@
...
+
+    <!-- Error message displayed when a disk error occurs during form download -->
+    <string name="form_save_disk_error">Disk error occurred while downloading this form.</string>
 
-    <string name="form_with_no_hash_error">The form has no hash.</string>
-    <string name="form_parsing_error">The form can not be parsed.</string>
-    <string name="form_save_disk_error">Disk exception occurred during saving the form.</string>
-    <string name="form_with_invalid_submission_error">The form has invalid submission.</string>
+    <!-- Error message displayed when a form being downloaded specifies an invalid submission URL -->
+    <string name="form_with_invalid_submission_error">This form has an invalid submission URL.</string>
```
seadowg: @yanokwa it'd be great if you could review these error messages for me. They will all have `report_to_project_lead` appended to the end of them (see `FormDownloadExceptionMapper` and `FormSourceExceptionMapper`) to try and push users to complain through their project/programme rather than leaving a negative review.

yanokwa: I think we should add a bit of a hint that the project lead can use to track down the problem. Is it a form problem? A server problem?

```
The server did not provide a form hash.
This form has an invalid submission URL.
A device error occurred during download.
```

`form_parsing_error` is a little trickier. Why can't the form be parsed? Is it because the file is corrupt due to a network error? Is it because of poor form design? I like processed over parsed, but I don't feel strongly about it. 

We could also be a little less passive, but I'd rather not because I think that'd be a style change across all the error messages.
```
Your form has an invalid submission URL
Your server did not provide a form has.
```

seadowg: Yeah I like the idea of being more specific about **where** the error occurred. I've made changes based on that feedback. I'm not sure about the slightly more aggressive "your" so I've not made that change.

yanokwa (approved):
The wording changes look great to me. That's the only thing I reviewed.

lognaturel:
I expected the raw exception text to also be surfaced. Will that be a follow-on? I lost track of the design but I thought that’s what would show up as the details. Things like form design exceptions are impossible to troubleshoot without it. Maybe it could just be “This form can’t be parsed. Details: %s” This might be tracked somewhere else.

seadowg:
> I expected the raw exception text to also be surfaced. Will that be a follow-on? I lost track of the design but I thought that’s what would show up as the details. Things like form design exceptions are impossible to troubleshoot without it. Maybe it could just be “This form can’t be parsed. Details: %s” This might be tracked somewhere else.

Yeah, we thought the new UI and more granular messages was enough of a positive step for one issue/PR and that we could add the exception details "expand thing" (the wee arrows in the [original design](https://www.figma.com/proto/XaN5etTC2Br1AyVJ1IqNi0/Errors?page-id=0%3A1&node-id=2%3A93&viewport=241%2C48%2C0.62&scaling=scale-down&starting-point-node-id=2%3A2)) later if we felt we needed it. My current thinking is that we might want to add the details in a patch release or in the next release (given our capacity at the moment).

lognaturel:
Fantastic, thanks!!

kkrawczyk123:
@seadowg I am not saying that testing has been finished for this one already but so far I have noticed one difference in dialog message for automatic form update and manual download for form update. After manual download new dialog is visible but for automatic form update still old one displays, see screenshots:
| automatic form update | manual download of form update: |
| ------ | ------|
| Screenshot_20211004-151204 | Screenshot_20211004-151817 |

seadowg:
@kkrawczyk123 is that when clicking on the auto update notification? I'd forgotten that shows results at all! I'll write up an issue to cover this new error UI being added there as well (and in match exactly).

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1 and 11

Tested scenarios:
- download form update manually: form version not specified on the download results dialog
- download all forms: general dialog for all forms without a specified list of forms
- download all forms with at least one fail: show details button visible - able to preview list of errors 
- more than 1 error with downloads: able to show details, scroll list of errors 
- lose internet connection in the middle of downloading forms: dialog with show details, ale to display list of failed forms
- use ok button
- display list of errors in light and dark mode 
- automatic version update: success - form version specified on dialog so it differs from manual download results dialog. 
- invalid Central's/Aggregate url: no changes visible
- not existing project number: no changes visible
- download forms with success from ona, kobo
- download forms with failures from ona, kobo
- screen rotation and minimizing when dialog displayed and on the error list 
- match exactly different cases: 15 min update, change form status on Central to closing/closed, update form version on Central, scan draft testing, publish draft wait for notification, revoke access to form, revoke app user's access: no regressions detected
- regression checks on downloading form from Google drive

mmarciniak90:
Also successfully tested on Android versions: 8.1, 9.0, 10.0

PR #4831, opened 2021-09-23
Fix dependency failures when building release

yanokwa:
I can confirm that this change, plus increasing the [heap size](https://github.com/getodk/collect#changing-jvm-heap-size) gets `./gradlew assembleOdkCollectRelease` to work on my machine.


grzesiek2010 (approved):

yanokwa:
Switched base to master per @lognaturel's suggestion. Will cherry pick commit and apply to branch.

I'm not too worried about the instrumented tests failure. Seems to be spurious.

Issue #656, opened 2017-03-16
Run lint/unit tests in CI on release builds


yanokwa:
Currently, we run lint and unit tests against debug builds. This is because developers are generally generating debug builds. On the CI server, we currently test against debug, but it might be a better idea to test against release.

lognaturel:
@grzesiek2010, @seadowg this seems worth doing? I think the advantage is that the build would go through all the same transformation as for release. I believe this issue was filed after we had some build issues that only showed up for release.

seadowg:
I don't see any harm in doing it! I haven't seen us run into any of these kinds of issues personally but if there's no blocker to doing it I say we go for it.

yanokwa:
It might be time to take this on. It's no fun to catch issues when doing a release.

seadowg:
I've put up a [PR](#4835) that adds a release build to CI. That's enough to catch the errors we've run into on the last few releases.

seadowg:
It feels like the above change (and moving QA to a minified build) has helped us a lot with any cracks we were seeing that this might have fixed. Closing this.

PR #4835, opened 2021-09-27
Add self signed release build to CI

seadowg:
Work towards #656

CI checks and ran `selfSignedRelease` to check it's still good locally. Verified that `selfSignedRelease` failed with the same problem that the beta build has recently.

We could use one of the other release builds but that would mean including keys/keystore in the CI config.

No need for QA on this.


collect_app/build.gradle
```diff
@@ -97,9 +97,8 @@ android {
             }
         }
 
-        // This uses the default debug keystore that the Android tooling creates https://stackoverflow.com/questions/16622528/android-studio-debug-keystore
         selfSignedRelease {
-            storeFile new File(System.getProperty("user.home") + '/.android/debug.keystore')
+            storeFile file('../debug.keystore')
```
seadowg, line 101: We were originally using the Android Studio `debug.keystore` for convenience, but this requires a full installation of Android Studio (rather than just having the Android SDK tools installed). Generating our own one works in basically the same way. I've set the expiry to 1000 years, so we won't have to do it again for a while...

.circleci/config.yml
```diff
@@ -143,6 +143,18 @@ jobs:
...
+    <<: *android_config
+    steps:
+      - attach_workspace:
+          at: ~/work
+      - restore_cache:
+          key: deps-{{ checksum "deps.txt" }}
+
+      - run:
+          name: Assemble self signed release build
+          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 assembleSelSignedRelease
```
yanokwa: Should probably be `assembleSelfSignedRelease`. Note the f in Self.

seadowg: Wait. How does this work?

.circleci/config.yml
```diff
@@ -183,6 +195,9 @@ workflows:
   workflow:
     jobs:
       - compile
+      - build_release:
```
yanokwa: I'd expect this farther down next to build_instrumented to match the ordering in top. Feel free to ignore.

seadowg: Yeah good point. Will change.

yanokwa (approved):
Confirmed that this does generate an APK that I can see on CircleCI.

PR #4837, opened 2021-09-28
Bump target version to API 30

### LINKED ISSUE #4533: [By Nov 1, 2021] Update targetSdkversion to 30
Must have for any release that will go out after Nov 1, 2021.

I have done a quick read through https://developer.android.com/about/versions/11/behavior-changes-11 and I believe the changes needed are:
- [x] Remove all legacy storage access
- ~~Change `READ_PHONE_STATE` to only API 29 and under, [add `READ_PHONE_NUMBERS`](https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers)~~
- [x] Look into [APK Scheme Signature v2](https://developer.android.com/about/versions/11/behavior-changes-11#minimum-signature-scheme). I believe we need to change the release process.
- ~~Consider any action to take given that [permissions won't be requested after being repeatedly denied](https://developer.android.com/about/versions/11/privacy/permissions#dialog-visibility).~~

seadowg:
Closes #4533

Very little had to change here, as we've been dealing with the largest issue (scoped storage) over the last year or so already. 

One change I made is to remove our custom toasts that show on constraint violations. Some of the methods we needed to create these [have been made no-ops](https://developer.android.com/about/versions/11/behavior-changes-11#text-toast-api-changes). These will now just show normal toasts.

Ran tests and had a quick play with the app. Also double-checked that nothing in the [docs](https://developer.android.com/about/versions/11/behavior-changes-11) looked like it would cause us problems. To address the big privacy changes:

* Scoped storage enforcement: Storage migration took care of that.
* Permissions auto-reset: Not really a problem for permissions to reset every few **months** of no use.
* Background location access: We moved our primary background location access to being foreground (with a foreground service) in the last release.
* Package visibility: Don't think this affects us.

Nothing to discuss here really.

I think a quick QA pass on just the critical app features would be good, but I'd timebox it to 30 mins or so as really the best test for this will be getting it out to beta. It's not a change we can choose not to include anyway so getting it out there is probably best.

### DIFF
--- a/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
+++ b/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
@@ -47,6 +47,7 @@ object ToastUtils {
     }
 
     @JvmStatic
+    @Deprecated("Toast position cannot be customized on API 30 and above. A dialog is shown instead for this API levels.")
     fun showShortToastInMiddle(activity: Activity, message: String) {
         showToastInMiddle(activity, message)
     }
--- a/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
+++ b/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
@@ -2,6 +2,9 @@ package org.odk.collect.androidshared.utils
 
 import android.app.Application
 import android.content.Context
+import android.view.Gravity
+import android.view.ViewGroup
+import android.widget.TextView
 import android.widget.Toast
 import org.odk.collect.strings.getLocalizedString
 
@@ -39,6 +42,11 @@ object ToastUtils {
         )
     }
 
+    @JvmStatic
+    fun showShortToastInMiddle(context: Context, message: String) {
+        showToastInMiddle(context.applicationContext as Application, message)
+    }
+
     private fun showToast(
         context: Application,
         message: String,
@@ -46,6 +54,25 @@ object ToastUtils {
     ) {
         hideLastToast()
         lastToast = Toast.makeText(context, message, duration)
+        lastToast.show()
+    }
+
+    private fun showToastInMiddle(
+        context: Application,
+        message: String,
+        duration: Int = Toast.LENGTH_SHORT
+    ) {
+        hideLastToast()
+        lastToast = Toast.makeText(context, message, duration)
+        try {
+            val group = lastToast.view as ViewGroup?
+            val messageTextView = group!!.getChildAt(0) as TextView
+            messageTextView.textSize = 21f
+            messageTextView.gravity = Gravity.CENTER
+        } catch (ignored: Exception) {
+            // ignored
+        }
+        lastToast.setGravity(Gravity.CENTER, 0, 0)
         lastToast.show()
     }
 
--- a/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
+++ b/androidshared/src/main/java/org/odk/collect/androidshared/utils/ToastUtils.kt
@@ -1,11 +1,15 @@
 package org.odk.collect.androidshared.utils
 
+import android.app.Activity
 import android.app.Application
 import android.content.Context
+import android.os.Build
 import android.view.Gravity
 import android.view.ViewGroup
 import android.widget.TextView
 import android.widget.Toast
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+import org.odk.collect.androidshared.R
 import org.odk.collect.strings.getLocalizedString
 
 /**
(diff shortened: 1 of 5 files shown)


yanokwa:
Is the Toast change related to the target API?

seadowg:
@yanokwa yeah it's in the docs I linked above. Here's a link to the specific section: https://developer.android.com/about/versions/11/behavior-changes-11#text-toast-api-changes.

yanokwa:
I don't love how the toasts are done. The constraint messages are very easy to miss now and I worry it will be disruptiv.

Can we do something different? Maybe we can flash the field (I think we do that in field lists) and add the warning text below? Bootstrap does something similar. See https://getbootstrap.com/docs/5.0/forms/validation/.  I don't want this to block this PR, but it does feel like it should go into this release.

lognaturel:
@seadowg does changing the target sdk version mean all OS versions ignore the toast styling? That is my understanding but would be good to confirm. I’m annoyed I didn’t anticipate that, sorry. 

If somehow not, consider leaving the toast customizations for this release. The Android 11 behavior has already changed anyway. 

If so, I think the question becomes whether it’s worth expanding the scope of this release to change the way toast text is displayed (probably not? There’s probably still time for the target sdk change to be bumped to the next release). I agree with @yanokwa that the toasts for in-form issues as they are in this PR can’t go out to all Android versions.

seadowg:
> @seadowg does changing the target sdk version mean all OS versions ignore the toast styling? That is my understanding but would be good to confirm. I’m annoyed I didn’t anticipate that, sorry.

Great question. I'll investigate. My assumption would be that it changes it for all OS levels, but let's see. 

> If somehow not, consider leaving the toast customizations for this release. The Android 11 behavior has already changed anyway.

Agreed. If we only see the problem on Android 11 after the target change, then we can make this change in v2021.4 instead.

> If so, I think the question becomes whether it’s worth expanding the scope of this release to change the way toast text is displayed (probably not? There’s probably still time for the target sdk change to be bumped to the next release). I agree with @yanokwa that the toasts for in-form issues as they are in this PR can’t go out to all Android versions.

I think we're close enough to Nov 1st that we need to make this change in the next release. If we release mid-late Oct without the target API change, we'd be risking not being able to push out hot fixes for that release (is my understanding anyway).

> Can we do something different? Maybe we can flash the field (I think we do that in field lists) and add the warning text below? Bootstrap does something similar. See https://getbootstrap.com/docs/5.0/forms/validation/. I don't want this to block this PR, but it does feel like it should go into this release.

Yeah, Material has a similar style (see "Adding errors to a text field" [here](https://material.io/components/text-fields/android#using-text-fields)). It's a non-trivial change, but would also be a lot nicer than even the toasts in the middle of the screen. If we do have to make the change in this release, it might be that just using a simple dialog would be the way to go. We can think more about that once I've done some investigation.

seadowg:
Ok so here's my findings:

* Running app with target API 30 on Android 10 (API 29): Constraint toasts still show in the middle (and large)
* Running app with target API 30 on Android 11 (API 30): Constraint toasts show as normal toasts (at the bottom and small)
* Running app with target API 29 on Android 11 (API 30): Constraint toasts still show in the middle (and large)

So it looks like the change won't affect old OS levels (I guess that makes sense given Toasts are still part of the OS level SDK) and that the API 30 SDK is only changing to the new no-op behaviour for apps that actually target API 30.

We actually have more Android 11 (API 30) users than I would expect - 59k in the last 28 day window according to [Firebase](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/overview/~2F%3Ft%3D1632997490878&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..dataFilters%253DoperatingSystemWithVersion%252CAndroid%25252011%252CEQ%252CDIMENSION%252C%252Ctrue&cs%3Dapp.m.dashboard.overview&g%3D1). Given that, I think we're still not able to push this out as-is, even if it only effects the Andorid 11 users. 

To me, it looks like we changing from a toast to a dialog on Android 11 would be fairly simple, so I'm going to make that change, so we can all review and see how we feel about it as a temporary solution. I think eventually we just want to use the Material error pattern (field highlighting) but that feels like it would be a big hold up for this next release.

seadowg:
@yanokwa @lognaturel ok I've made it so we show a dialog on Android 11 and above. This feels like a nice compromise to me and should hold us over until we revise the constraint violation UI. Let me know what you think!

yanokwa:
The dialog isn't great, but I agree it's a nice compromise for now. 

Looks kinda weird though with the super small font. Is that a standard dialog?
Screen Shot 2021-10-05 at 04 22 29

seadowg:
> Looks kinda weird though with the super small font. Is that a standard dialog?

@yanokwa yeah that's just a standard dialog. I do agree the default sizing in Material (`?textAppearanceBody2`) seems a little small (especially when the dialog doesn't have a title). I don't think we want to use a dialog long term though, so I'd vote we go with this instead of spending any time customizing the defaults here.

lognaturel (approved):
The dialog addition is a great little temporary fix. Would be good to merge this and start exercising it ASAP, I think.

PR #4840, opened 2021-09-29
Correct testshared dependencies

seadowg:
This is clean up after the build issues we had.

`testshared` should only ever really be built during debug builds (as it will only ever be a `testImplementation` dep) but the Android plugin's assemble tasks (when used on the app module) builds every module in the project annoyingly. This means that `debugImplementation` deps will be missing for release builds of test only dependencies.

### DIFF
--- a/testshared/build.gradle
+++ b/testshared/build.gradle
@@ -42,11 +42,7 @@ dependencies {
     implementation Dependencies.androidx_appcompat
     implementation Dependencies.robolectric
     implementation Dependencies.android_material
-    implementation Dependencies.androidx_fragment_testing
-
-    testImplementation Dependencies.androidx_test_ext_junit
-
-    debugImplementation(Dependencies.androidx_fragment_testing) {
+    implementation(Dependencies.androidx_fragment_testing) {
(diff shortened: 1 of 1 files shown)


testshared/build.gradle
```diff
@@ -42,11 +42,7 @@ dependencies {
     implementation Dependencies.androidx_appcompat
     implementation Dependencies.robolectric
     implementation Dependencies.android_material
-    implementation Dependencies.androidx_fragment_testing
-
-    testImplementation Dependencies.androidx_test_ext_junit
```
seadowg, line 47: This was just unused.

lognaturel:
> builds every module in the project 

Does building every module with their dependencies affect apk size? It has been creeping up.

seadowg:
> Does building every module with their dependencies affect build size? The apk size has been creeping up.

@lognaturel As far as I know it's just running `assemble` for every module but not actually including it in the APK. We should double-check that probably. Nothing I've done here would affect that (as test modules are getting built regardless of this change).

lognaturel:
Right, just want to make sure that check isn’t forgotten.

lognaturel (approved):

seadowg:
@lognaturel I'm going to double-check later today or tomorrow morning.

PR #4842, opened 2021-10-01
Add shared reference layer directory

### LINKED ISSUE #4784: Add shared map layers directory
From https://forum.getodk.org/t/use-reference-layers-across-several-collect-projects/34601.

Right now, if users want to add map layers, they can add them to the project level `layers` directory. This means if a layer is shared between multiple projects, it will need to be copied onto the device multiple times, which could end up wasting a lot of storage. We should add a shared `layers` directory at the top level that all projects can access.

## Acceptance

- [x] Given I've setup a Project in Collect
And I've copied a layer to the Project's `layers` directory (`/projects/{projectId}/layers`)
And I've copied a different layer to the shared `layers` directory (`/layers`)
When I navigate to map settings for my Project
And I click "Layer data file"
Then I see both layers are available

- [x] Given I've setup a Project in Collect
And I've copied a different layer to the shared `layers` directory (`/layers`)
When I navigate to map settings for my Project
And I click "Layer data file"
And I select the layer in the shared `layers` directory
And I open a map in a form
Then the layer loads correctly

### LINKED ISSUE #4862: Collect crashes when user tries to open `Layer data file` dialog when the previously selected layer is removed
Collect master, v2021.2.0

#### Problem description
Collect crashes when user tries to open `Layer data file` dialog when the previously selected layer is removed

1. User sets offline layer in `Settings/Maps`
2. The offline layer in use is no longer available - eg. it has been removed.
3. User is still on `Settings/Maps `view and sees layer name
4. User clicks on the `Layer data file` to open dialog -> Collect crashes

ezgif com-gif-maker

#### Expected behavior
Crash does not occur.
Dialog is opened and the removed layer is no longer visible

#### Other information 
Discovered while testing #4842 
Crash does not occur when the dialog is opened.

seadowg:
Closes #4784 
Closes #4862
~~Blocked by #4829~~

Played around with tiles in the new shared layer and project layer directories, new tests.

The general approach here was to break the out the reference layer "loading" into a repository so that the UI components (preference and map fragments) didn't need to care about where the layer files actually were. This made it pretty easy to then make it so that layers could be loaded from multiple directories. There was some weirdness around what to do with matching layer file names/paths in both the shared and project layer dirs, but I've left some comments about that.

Should just add the new feature! Good to test that the existing offline maps/reference layer features are working. Good to keep in mind that:

* Reference layers can live within subdirectories (and so on and so on) of either the project or shared layers directory
* If a reference layer in the shared dir and the project dir have the same name/path then only the project dir one should be available to the user


collect_app/src/main/java/org/odk/collect/android/preferences/PreferenceScreenMenuDelegate.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun onCreateOptionsMenu(menuInflater: MenuInflater, menu: Menu) {
+        menuInflater.inflate(R.menu.preferences_screen, menu)
+    }
+
+    override fun onPrepareOptionsMenu(menu: Menu) {
+        // No-op
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        externalWebPageHelper.openWebPageInCustomTab(activity, helpUrl)
```
seadowg: I didn't use the functionality in `ExternalWebPageHelper` that allows it to integrate with the Chrome Custom Tabs service. We use this elsewhere (the About screen and the URL Widget) - the advantage is that it can "warm up" the browser before the user actually needs to use it. Personally I'm not sure the extra work to bind/unbind a service is really worth it to optimize for an action that only gets performed occasionally - this is definitely something to consider in the places we're already using it as well.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ReferenceLayerPreferenceDialog.java
```diff
@@ -1,21 +1,30 @@
...
 
 public class ReferenceLayerPreferenceDialog extends ListPreferenceDialogFragmentCompat implements DialogInterface.OnClickListener {
 
+    /**
+     * Views should not be stored statically like this. The relationship on how the list is setup
+     * here should be inverted - {@link ReferenceLayerPreferenceDialog} should be asking
+     * {@link CaptionedListPreference} for items  rather than having them pushed through this static
+     * field.
+     */
+    @Deprecated
```
seadowg, line 26: I'd like to have cleaned this up, but I really didn't want to get bogged down in more untangling (I'm doing enough there with all the geopoint stuff). A job for another time.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ReferenceLayerPreferenceDialog.java
```diff
@@ -38,8 +47,10 @@ protected void onBindDialogView(View view) {
         if (getPreference() instanceof CaptionedListPreference) {
             preference = (CaptionedListPreference) getPreference();
         }
+
+        addHelpFooter(view);
```
seadowg, line 51: I preferred to have this added here rather than change `CaptionedListPreference` as it's specific to reference layers. We might eventually end up with a `PreferenceWithHelp` or something if we want this other places.

grzesiek2010 (approved):

mmarciniak90:
@seadowg, I'm not sure about this assumption:

>If a reference layer in the shared dir and the project dir have the same name/path then only the project dir one should be available to the user

Why did we decide to show only layer from project dir when we have a full path? If I put it in a subdirectory, it's available. I can imagine a case where 2 different layers have the same name, then the user cannot access one of them.

seadowg:
> Why did we decide to show only layer from project dir when we have a full path? If I put it in a subdirectory, it's available. I can imagine a case where 2 different layers have the same name, then the user cannot access one of them.

@mmarciniak90 fair question! @lognaturel and I discussed this and decided to go with that. I should have been clearer in the PR description that although that's a current limitation (due to a quirk of how the reference files are identified under the hood), we could do extra work later to allow users to have layers with matching names in the project and shared dir if it comes up as a need.  We didn't know of/couldn't think of a genuine workflow that would involve having matching files in both directories, especially since the shared directory is a new feature. Does that seem ok?

kkrawczyk123:
@seadowg I have noticed crash when followed scenario:
1. Add layer to layer folder (doesn't matter if it's project's folder or layer folder for all projects)
2. Open any GeoWidget
3. Select added layer on map view
4. Open layer dialog again
5. Minimize app
6. Delete selected layer from folder 
7. Go back to Collect
8. Click on just deleted layer 

see gif:
ezgif com-gif-maker (3)

This is a regression issue as that scenario works well on master. I am not attaching stacktrace as it was huge one.

seadowg:
@kkrawczyk123 ah nice that's a sneaky one. Will look into it.

seadowg:
@kkrawczyk123 ok that should be fixed now. I also fixed #4862 while I was in there as I spotted it playing around (before I saw the issue) and assumed my changes had caused it.

mmarciniak90:
**Tested with success**

Verified cases:
* all types of projects: Central, Aggregate, Drive, Ona, Kobo
* layers in a shared folder
* layers in each projects
* layers in subdirectories
* layers sorted alphabetically even if sources are mixed
* path under layer 
* select layers in settings and through map view (widget, map with submissions)
* get help with reference layers
* Google, Mapbox, OSM maps
* GeoPoint, GeoTrace, GeoShape
* all input types
* use with and without layer
* delete layers manually
* add few layers
* fixed reported crashes
* scroll on the Layer data file dialog
* rotation
* minimization
* light/dark mode

kkrawczyk123:

PR #4843, opened 2021-10-04
Tweak new theme

### LINKED ISSUE #4832: Issues with newly updated themes
Current master branch

#### Problem description

1. Empty area around the clipboard menu (only in preferences, in form filling everything seems to be fine)

|  |  |
| Screenshot_1632428898  | Screenshot_1632428969  |

2. Disabled buttons in form entry looks good (like disabled) but other buttons not really if they are in the same state

| Get selected button is disabled but it's hardly visible | This looks good |
| Screenshot_1632429051  | Screenshot_1632429066 |

~~3. I noticed that after switching themes sometimes some elements are not undated immediately like the `x` icon~~ EDIT (from @seadowg): This is an existing issue we noticed during the QA pass and decided not to fix. I'll pull it out as a [separate issue](https://github.com/getodk/collect/issues/4836) and mark it as such.

seadowg:
Closes #4832

As well as fixing the problems in the issue this makes the back/next button bar more visible. This should really use a `BottomAppBar` which would give us the correct color based on elevation but that's a switch we don't need to dive into right now.

Quick play with the app

The only real discussion here is around the nav bar, and I've covered that above.

Should just fix the issues! I think this can go in without QA as it's just some small fixes.


grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**

Tested cases:
- default/light/dark mode
- server url 
- username
- odk password
- fallback submission url
- project name
- project icon
- hex in project color
- form metadata/username
- form metadata/phone number
- form metadata/email address
- admin password/show admin password
- all widgets
- ex type widgets
- back and next button 
- get blank form/ get selected button
- delete saved form

mmarciniak90:
Tested with success on Android versions: 8.1, 9.0, 10.0

Issue #4170, opened 2020-10-15
Crash when editable field with calculate is used in relevance expression


grzesiek2010:
#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/85177cf7b388af8b6ec086cd77c7cc05?time=last-seven-days&versions=v1.28.2%20(3943)&sessionEventKey=5F88045E02020001727F5A1D764EC4D0_1462203208677627982

```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=2; index=2
       at org.odk.collect.android.activities.FormEntryActivity.saveAnswersForCurrentScreen(FormEntryActivity.java:1129)
       at org.odk.collect.android.activities.FormEntryActivity.updateFieldListQuestions(FormEntryActivity.java:2714)
       at org.odk.collect.android.activities.FormEntryActivity.access$1900(FormEntryActivity.java:200)
       at org.odk.collect.android.activities.FormEntryActivity$10.run(FormEntryActivity.java:2681)
       at android.app.Activity.runOnUiThread(Activity.java:6972)
       at org.odk.collect.android.activities.FormEntryActivity.widgetValueChanged(FormEntryActivity.java:2678)
       at org.odk.collect.android.formentry.ODKView.widgetValueChanged(ODKView.java:621)
       at org.odk.collect.android.widgets.QuestionWidget.widgetValueChanged(QuestionWidget.java:499)
       at org.odk.collect.android.widgets.items.SelectOneWidget.onItemClicked(SelectOneWidget.java:103)
       at org.odk.collect.android.formentry.questions.AudioVideoImageTextLabel.lambda$setTextView$0(AudioVideoImageTextLabel.java:118)
       at org.odk.collect.android.formentry.questions.-$$Lambda$AudioVideoImageTextLabel$MqCYJ2yJoQ6DnhH04sxqp-kc3wc.onClick(-.java:2)
```

#### Steps to reproduce the problem
[crashOnFieldList.xlsx](https://github.com/getodk/collect/files/5384239/crashOnFieldList.xlsx)
![ezgif com-video-to-gif (7)](https://user-images.githubusercontent.com/3276264/96113740-e6a77a80-0ee4-11eb-8954-827076d07a84.gif)

lognaturel:
[...]

PR #4845, opened 2021-10-04
Update JR to 3.3.0 snapshot and update test for re-calculation in field list

lognaturel:
Update JavaRosa to v3.3.0-SNAPSHOT. https://github.com/getodk/javarosa/pull/646 and https://github.com/getodk/javarosa/pull/644 have been QAed.

I've updated the test that no longer fails with the form from #4170. It's an extension of the problem that the old form displayed -- there's still a calculate being used as a default but then there's also relevance in a field list added to the mix. I'll cherry pick that commit when we have a real PR updating JR. I wanted to give @seadowg a chance to review in the meantime.

Side note, we could likely change the order of relevance update and field list re-rendering so that there wouldn't be an exception in that case. However, similar to the other problems from https://github.com/getodk/collect/issues/4750, the form is nonsense so it's good that there's a crash. As we make more improvements to field lists, it might go away and then this test can be removed.

<details><summary>Initial description for QA</summary>
NOT FOR MERGING

This is so that @getodk/testers can verify https://github.com/getodk/javarosa/pull/646 and https://github.com/getodk/javarosa/pull/644. As a side effect, the cases described at https://github.com/getodk/collect/issues/4750 should no longer show errors.

~Related to the last point, there is now a failing connected test. This is discussed at https://github.com/getodk/javarosa/pull/646#issuecomment-933337070.~
</details>

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/CatchFormDesignExceptionsTest.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/CatchFormDesignExceptionsTest.kt
@@ -23,9 +23,14 @@ class CatchFormDesignExceptionsTest {
         rule.startAtMainMenu()
             .copyForm("form_design_error.xml")
             .clickFillBlankForm()
-            .clickOnForm("form_design_error")
-            .openSelectMinimalDialog()
-            .clickOnText("A")
+            .clickOnForm("Relevance and calculate loop")
+            .answerQuestion(1, "B")
+            .scrollToAndAssertText("third")
+            .answerQuestion(2, "C")
+            // Answering C above triggers a recomputation round which resets fullName to name.
+            // They're then equal which makes the third question non-relevant. Trying to change the
+            // value of a non-relevant node throws an exception.
+            .answerQuestion(2, "D")
             .assertText(R.string.update_widgets_error)
             .clickOKOnDialog()
             .assertOnPage(MainMenuPage())
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormEntryPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormEntryPage.java
@@ -1,15 +1,5 @@
 package org.odk.collect.android.support.pages;
 
-import androidx.appcompat.widget.AppCompatImageButton;
-import androidx.test.espresso.Espresso;
-
-import org.hamcrest.Matchers;
-import org.odk.collect.android.R;
-import org.odk.collect.android.support.ActivityHelpers;
-import org.odk.collect.android.utilities.FlingRegister;
-
-import java.util.concurrent.Callable;
-
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.action.ViewActions.click;
 import static androidx.test.espresso.action.ViewActions.longClick;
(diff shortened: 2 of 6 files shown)


lognaturel:
@seadowg please verify that my proposed sequencing is ok with you (prioritizing QA over updating the test).

collect_app/build.gradle
```diff
@@ -286,11 +286,14 @@ dependencies {
...
     implementation Dependencies.commons_io
     implementation Dependencies.opencsv
-    implementation(Dependencies.javarosa) {
-        exclude group: 'joda-time'
-        exclude group: 'org.slf4j'
-        exclude group: 'org.hamcrest', module: 'hamcrest-all'
-    }
+    implementation 'net.sf.kxml:kxml2:2.3.0'
+    implementation 'org.apache.commons:commons-csv:1.4'
+    implementation files('javarosa-3.2.0.jar')
```
lognaturel: This is still 3.2.0 because it's off the unmerged https://github.com/getodk/javarosa/pull/646 but it's the release candidate for 3.3, basically.

seadowg:
@lognaturel yeah seems good! I'll add "needs testing" to this. I'll also just make it a draft so it's even clearer it's not for merging.

lognaturel:
I thought the build failure might have been something intermittent but was wrong. I assume it has to do with including the jar.

seadowg:
@lognaturel definitely could be. @mmarciniak90 @kkrawczyk123 this is still good to test even though the build is failing. We'll sort all that out when we've merged the JavaRosa PR.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/CatchFormDesignExceptionsTest.kt
```diff
@@ -23,9 +23,14 @@ class CatchFormDesignExceptionsTest {
         rule.startAtMainMenu()
             .copyForm("form_design_error.xml")
             .clickFillBlankForm()
-            .clickOnForm("form_design_error")
-            .openSelectMinimalDialog()
-            .clickOnText("A")
+            .clickOnForm("Relevance and calculate loop")
+            .answerQuestion(1, "B")
+            .scrollToAndAssertText("third")
```
lognaturel, line 28: This feels awkward but I couldn't find a better way to make sure that the third question was displayed on a short screen. Scrolling to the question text is insufficient because the edittext needs to be on-screen.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormEntryPage.java
```diff
@@ -171,11 +171,6 @@ public FormEntryPage clickRankingButton() {
         return this;
     }
 
-    public FormEntryPage putTextOnIndex(int index, String text) {
```
lognaturel, line 174: I moved and renamed this because it took me a long time to find it.

mmarciniak90:
**Tested with success**

Verified cases:
- filled and edited forms
- jumped to questions via hierarchy view on new and saved form
- added and removed groups
- no crash in `crashOnFieldList` form but the error is still visible
- all questions can be filled in `minimal-select-field-list` form and error does not appear
- regression check on all mentioned forms
- compared the behavior with v2021.2.0
- light/dark mode

Verified forms from [project 340](https://test.getodk.cloud/#/projects/340)
* cascade
* crashOnFieldList
* dynamiChoices
* External 52k split
* minimal-select-field-list
* Nigeria Wards Internal
* Relevance and calculate loop
* select-one-file

kkrawczyk123:

seadowg:
@lognaturel I think we can do a snapshot and then upgrade Collect to that now as part of this PR.

lognaturel:
Thanks, @kkrawczyk123, @mmarciniak90!

@seadowg did you review the test change? I’ll update the first commit in the next 24hrs to use the snapshot but you could look at the test if you haven’t already.

seadowg:
@lognaturel ah sorry I thought you hadn't got round to that because of the failures but looks like those are just mem issues. Will set myself as a reviewer. Circle CI won't let me rerun annoyingly as it's your PR!

seadowg (approved):

PR #4848, opened 2021-10-05
Add user property to track adoption of Match Exactly

### LINKED ISSUE #4779: Measure how popular each form management mode is
We lost this during the release of 2021.2 as we decided we didn't want to spend time adapting the user property work with multiple projects. We should find a way to measure this again for the next release.

seadowg:
Closes #4779

I've decided to just track if any project on the device is **not** using Match Exactly instead of trying to adapt user properties to be one user per project. 

~~As part of this I've also moved over to using Jetpack's [App Startup](https://developer.android.com/topic/libraries/app-startup) library for app initialization. We'd talked about doing this during the last release but didn't want to hold up what we were working on at the time.~~

Ran tests. I don't have access to add user properties to the debug Firebase project, but we can just verify this in beta.

Comments inline.

Shouldn't be any change in behaviour. Application initialization code has changed, but I don't think there's any specific QA to be done here.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
+++ b/collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
@@ -585,8 +585,8 @@ public class AppDependencyModule {
     }
 
     @Provides
-    public ApplicationInitializer providesApplicationInitializer(Application context, UserAgentProvider userAgentProvider, PropertyManager propertyManager, Analytics analytics, LaunchState launchState, AppUpgrader appUpgrader, AnalyticsInitializer analyticsInitializer, ProjectsRepository projectsRepository) {
-        return new ApplicationInitializer(context, userAgentProvider, propertyManager, analytics, launchState, appUpgrader, analyticsInitializer, projectsRepository);
+    public ApplicationInitializer providesApplicationInitializer(Application context, UserAgentProvider userAgentProvider, PropertyManager propertyManager, Analytics analytics, LaunchState launchState, AppUpgrader appUpgrader, AnalyticsInitializer analyticsInitializer, ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
+        return new ApplicationInitializer(context, userAgentProvider, propertyManager, analytics, launchState, appUpgrader, analyticsInitializer, projectsRepository, settingsProvider);
     }
 
     @Provides
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
@@ -20,6 +20,7 @@ import org.odk.collect.android.application.initialization.upgrade.AppUpgrader;
 import org.odk.collect.android.geo.MapboxUtils;
 import org.odk.collect.android.logic.PropertyManager;
 import org.odk.collect.android.logic.actions.setgeopoint.CollectSetGeopointActionHandler;
+import org.odk.collect.android.preferences.source.SettingsProvider;
 import org.odk.collect.android.utilities.LaunchState;
 import org.odk.collect.projects.ProjectsRepository;
 import org.odk.collect.utilities.UserAgentProvider;
@@ -38,12 +39,14 @@ public class ApplicationInitializer {
     private final AppUpgrader appUpgrader;
     private final AnalyticsInitializer analyticsInitializer;
     private final ProjectsRepository projectsRepository;
+    private final SettingsProvider settingsProvider;
 
     public ApplicationInitializer(Application context, UserAgentProvider userAgentProvider,
                                   PropertyManager propertyManager, Analytics analytics,
                                   LaunchState launchState, AppUpgrader appUpgrader,
                                   AnalyticsInitializer analyticsInitializer,
-                                  ProjectsRepository projectsRepository) {
+                                  ProjectsRepository projectsRepository,
+                                  SettingsProvider settingsProvider) {
         this.context = context;
         this.userAgentProvider = userAgentProvider;
         this.propertyManager = propertyManager;
@@ -52,6 +55,7 @@ public class ApplicationInitializer {
         this.appUpgrader = appUpgrader;
         this.analyticsInitializer = analyticsInitializer;
         this.projectsRepository = projectsRepository;
+        this.settingsProvider = settingsProvider;
     }
 
     public void initialize() {
@@ -75,7 +79,7 @@ public class ApplicationInitializer {
         initializeMapFrameworks();
         initializeJavaRosa();
         analyticsInitializer.initialize();
-        new UserPropertiesInitializer(analytics, projectsRepository).initialize();
+        new UserPropertiesInitializer(analytics, projectsRepository, settingsProvider, context).initialize();
     }
 
     private void initializeLocale() {
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
@@ -22,8 +22,13 @@ class UserPropertiesInitializer(
             "UsingLegacyFormUpdate",
             projects.any {
                 val settings = settingsProvider.getGeneralSettings(it.uuid)
+                val serverUrl = settings.getString(ProjectKeys.KEY_SERVER_URL)
                 val formUpdateMode = settings.getString(ProjectKeys.KEY_FORM_UPDATE_MODE)
-                formUpdateMode != FormUpdateMode.MATCH_EXACTLY.getValue(context)
+
+                val notUsingDefaultServer = serverUrl != ProjectKeys.defaults[ProjectKeys.KEY_SERVER_URL]
+                val notUsingMatchExactly = formUpdateMode != FormUpdateMode.MATCH_EXACTLY.getValue(context)
+
+                notUsingDefaultServer && notUsingMatchExactly
             }.toString()
         )
     }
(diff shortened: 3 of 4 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/CollectHelpers.java
```diff
@@ -1,42 +1,19 @@
...
 import org.odk.collect.android.injection.config.AppDependencyModule;
 import org.odk.collect.android.injection.config.DaggerAppDependencyComponent;
-import org.odk.collect.android.javarosawrapper.FormController;
 
 public final class CollectHelpers {
 
     private CollectHelpers() {
     }
 
-    public static FormController waitForFormController() throws InterruptedException {
```
seadowg: Not using these any longer!

collect_app/src/main/AndroidManifest.xml
```diff
@@ -166,6 +166,23 @@ the specific language governing permissions and limitations under the License.
...
+            android:authorities="${applicationId}.workmanager-init"
+            android:exported="false"
+            android:enabled="false"/>
+
+        <provider
+            android:name="androidx.startup.InitializationProvider"
+            android:authorities="${applicationId}.androidx-startup"
+            android:exported="false"
+            tools:node="merge">
+            <meta-data  android:name="org.odk.collect.android.application.initialization.UserPropertiesInitializer"
```
seadowg: We only ever need to specify `Initializer`s explicitly if no others depend on them.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -83,6 +82,11 @@
     private AudioRecorderDependencyComponent audioRecorderDependencyComponent;
     private ProjectsDependencyComponent projectsDependencyComponent;
 
+    public Collect() {
+        super();
+        singleton = this;
```
seadowg: This needs to move here for anything that might need access to it before `onCreate`

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -290,7 +291,12 @@ public FileProvider providesFileProvider(Context context) {
 
     @Provides
     public WorkManager providesWorkManager(Context context) {
-        return WorkManager.getInstance(context);
+        try {
+            return WorkManager.getInstance(context);
+        } catch (IllegalStateException e) {
+            WorkManager.initialize(context, new Configuration.Builder().build());
```
seadowg: We just call `WorkManager.initialize` lazily now.

collect_app/src/test/java/org/odk/collect/android/application/RobolectricApplication.java
```diff
@@ -49,6 +41,9 @@ public void onCreate() {
         // We don't want any clicks to be blocked
         MultiClickGuard.test = true;
 
+        // Run Jetpack startup initializers
+        Robolectric.setupContentProvider(InitializationProvider.class);
```
seadowg: Robolectric doesn't automatically boot up providers

collect_app/build.gradle
```diff
@@ -361,6 +361,8 @@ dependencies {
     compileOnly Dependencies.findbugs_annotations
     compileOnly Dependencies.findbugs_jsr305
 
+    implementation("androidx.startup:startup-runtime:1.1.0")
```
lognaturel: For consistency, this feels like it should be in `Dependencies` and grouped with other androix-prefixed dependencies.

collect_app/build.gradle
```diff
@@ -361,6 +361,8 @@ dependencies {
     compileOnly Dependencies.findbugs_annotations
     compileOnly Dependencies.findbugs_jsr305
 
+    implementation Dependencies.androidx_startup
```
lognaturel: This should go with other dependencies that have an `androidx` prefix. I think it's not just for aesthetics -- there's value to knowing all the androidx "stuff" we're carrying around at a glance. I'd also tend to order `Dependencies` the same way but that's probably just aesthetic.

lognaturel (changes requested):
Change request: I think we should only count projects where the server URL is not https://demo.getodk.org/. I suspect a number of people are just leaving the demo project and that could possibly really skew the results.

General comments/questions: Unfortunately, I'm having some trouble convincing myself that using the app startup library is a win. The approach here seems ok but it is adding another dependency and another layer so I do want to make sure the benefits have been considered.

I've read through the [App Startup](https://developer.android.com/topic/libraries/app-startup) docs and worked on understanding its purpose. My understanding is that its biggest value is in replacing [`ContentProvider`-based initialization for libraries that need `context`](https://andretietz.medium.com/auto-initialize-your-android-library-2349daf06920) or to avoid having an `Application` subclass. I'm not seeing either of these applying to Collect. It also feels like more initialization could happen as part of dependency creation/injection (or maybe this is part of moving away from Dagger one day?).

Wish I'd thought through this a little earlier so we had a chance to quickly discuss! I can hand off to @grzesiek2010 here to do a final review. It's likely I'm just overlooking a benefit and/or not seeing how App State enables future improvements. If @seadowg and @grzesiek2010 see value to introducing it then I approve once the small change described above is implemented.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
lognaturel: This makes a lot of sense and is what I mean by letting DI do more initialization. Am I correctly understanding that this wasn't possible before because WorkManager used ContentProvider initialization and now because it uses App Startup we can explicitly disable its automatic initialization? That seems like a very compelling use of the library by `WorkManager`.

seadowg: We can explicitly disable it with or without using App Startup ourselves, but it's an example of how we're technically already opted in to this system with some Android libraries.

seadowg:
> Change request: I think we should only count projects where the server URL is not https://demo.getodk.org/. I suspect a number of people are just leaving the demo project and that could possibly really skew the results.

Ah good point. I had completely missed the fact that the demo projects are using manual downloads. Will change.

> General comments/questions: Unfortunately, I'm having some trouble convincing myself that using the app startup library is a win. The approach here seems ok but it is adding another dependency and another layer so I do want to make sure the benefits have been considered.

For me it comes down to the fact that I was slowly building our own framework for initialization so that we could have a "list" of initialization steps (that are testable independently). The main benefit in my mind is having a documented, non-custom version of that. The dependency system in App Startup will also let us test ordering in a way that we'd have to roll ourselves. 

This also explicitly moves our initialization away from `Application#onCreate` which will hopefully decrease the noise inside our application object and gives us control of what actually kicks off initialization (we can't control when `Application#onCreate` is called during tests for instance). We did have that with `ApplicationInitializer` but, again, it's a pretty messy custom component.

I think if there's a real discomfort with the move, then I can just use the changes to `UserPropertiesInitializer` here instead (and we can properly discuss App Startup another time).

>  It also feels like more initialization could happen as part of dependency creation/injection (or maybe this is part of moving away from Dagger one day?).

Generally I'd like to keep as much crucial stuff out of the Dagger module declarations as it's easy for details to get hidden in there. I think we still want to be doing things as "lazily" as possible however and on that we should be looking for opportunities to take initialization code and move it elsewhere (like we just did with storage initialization). I'd love to move away from Dagger one day though!

collect_app/src/main/AndroidManifest.xml
```diff
@@ -170,6 +170,23 @@ the specific language governing permissions and limitations under the License.
             android:name="com.google.android.maps"
             android:required="false" />
 
+        <!-- disable WorkManager initializer and let startup take care of that -->
```
grzesiek2010: Shouldn't we add this to as a meta like in: https://developer.android.com/topic/libraries/app-startup#disable-individual?
something like:
```
<meta-data android:name="androidx.work.impl.WorkManagerInitializer"
              tools:node="remove" />
```

collect_app/src/main/java/org/odk/collect/android/application/initialization/AnalyticsInitializer.kt
```diff
@@ -1,15 +1,34 @@
...
+    lateinit var analytics: Analytics
+
+    @Inject
+    lateinit var versionInformation: VersionInformation
+
+    @Inject
+    lateinit var settingsProvider: SettingsProvider
+
+    constructor(
+        analytics: Analytics,
```
grzesiek2010: Would it be possible to have default values here that take it's implementations from DaggerUtils.getComponent? so that we don't have both injects and this constructor but only the constructor? Something like in StoragePathProvider where we have:

```
class StoragePathProvider @JvmOverloads constructor(
    private val currentProjectProvider: CurrentProjectProvider = DaggerUtils.getComponent(Collect.getInstance())
        .currentProjectProvider(),
    val odkRootDirPath: String = Collect.getInstance().getExternalFilesDir(null)!!.absolutePath
) {
```

grzesiek2010 (changes requested):
The implementation is fine itself but I don't understand why a pr that should solve one issue (https://github.com/getodk/collect/issues/4779) in 90% is about something completly else.
Maybe you both discussed it when I was off but still I feel like it should be divided into two prs and the one about using App Startup discussed.

as @lognaturel said:

> Unfortunately, I'm having some trouble convincing myself that using the app startup library is a win. The approach here seems ok but it is adding another dependency and another layer so I do want to make sure the benefits have been considered.

I also have concerns about it. Nice implementation but much more complex for understanding for someone who hears of that library for the first time. 
Apart from the structure (whether is better or not) does it bring any improvement regarding performance in our case?

seadowg:
I feel like the response to using App Startup is pretty negative here so, instead of spending more time on that, I'm going to just pull it out, and we can discuss it separately.

lognaturel (approved):
Thank you and sorry!

lognaturel:
Really the rejection of App Startup can be seen as a compliment. What y’all have come up with doesn’t seem that messy to me (maybe I’m missing messiness) and the thing I like most about it is that it’s explicit and straightforward. Will be interesting to discuss briefly.

grzesiek2010 (approved):

seadowg:
I've added `UsingLegacyFormUpdate` as a custom dimension in Firebase.

PR #4856, opened 2021-10-13
Create file with project name sanitizing reserved chars

### LINKED ISSUE #4766: Empty file with project name inside project folder is not recreated
Collect v2021.2

#### Problem description
Empty file with project name inside project folder is not recreated

Use case 1 - special characters
1. User opens Settings -> Project display
2. User changes `Project name`
3. User puts a few special characters
4. Empty file with project name inside project folder is not created

Examples causing issue:

|Screenshot_20210731-093440_ODK Collect|Screenshot_20210731-093103|
|---|---|

Use case 2 - removed directories
1. User creates project(s)
2. User removes `files` or `projects` or separate project directory
3. User restarts Collect, directories are recreated
4. Empty file with project name inside project folder is not recreated

#### Expected behavior
Empty file with project name inside project folder is recreated

grzesiek2010:
Closes #4766

I added automated tests and also verified the fix manually.

The only ting of note is that I moved creating file with project name to `StoragePathProvider` so that we make sure it's created again even after removing a project dir manually.

We need to test creating/updating project name files. Nothing else should be affected.

No.

No.


collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
```diff
@@ -3,25 +3,34 @@ package org.odk.collect.android.storage
...
+import org.odk.collect.android.utilities.FileUtils
+import org.odk.collect.projects.ProjectsRepository
+import timber.log.Timber
 import java.io.File
 
-class StoragePathProvider @JvmOverloads constructor(
-    private val currentProjectProvider: CurrentProjectProvider = DaggerUtils.getComponent(Collect.getInstance())
-        .currentProjectProvider(),
+class StoragePathProvider(
+    private val currentProjectProvider: CurrentProjectProvider = DaggerUtils.getComponent(Collect.getInstance()).currentProjectProvider(),
```
seadowg, line 13: Is it possible to get rid of the default values here or do we still have too many places using the zero args constructor?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/ProjectDisplayPreferencesFragment.kt
```diff
@@ -149,8 +151,9 @@ class ProjectDisplayPreferencesFragment :
                     )
                 }
 
+                val sanitizedNewProjectName = StringUtils.sanitizeFileName(newValue.toString())
```
seadowg: I'm wondering if there's a way we could just have the file creation logic in one place. That way we could also avoid having a UI level test for the feature.

One idea would be to always check for a file with the project name in `getProjectRootDirPath` and create it (while deleting anything files in the root) if it's missing. That would definitely mean some extra I/O work when grabbing the path but I thinking that we should probably be doing this async anyway (just with `Scheduler`) as we don't need the file to be there atomically. What do you think about that kind of approach?

shared/src/main/java/org/odk/collect/shared/strings/StringUtils.kt
```diff
@@ -94,4 +94,8 @@ object StringUtils {
             str.substring(0, str.length - remove.length)
         } else str
     }
+
+    // https://stackoverflow.com/questions/2679699/what-characters-allowed-in-file-names-on-android
+    @JvmStatic
+    fun sanitizeFileName(fileName: String) = fileName.replace("[\"*/:<>?\\\\|]".toRegex(), "_")
```
seadowg: I'm thinking this might make more sense in `PathUtils`? It could be something like `PathUtils.getPathSafeFileName`.

collect_app/src/main/java/org/odk/collect/android/storage/StoragePathProvider.kt
grzesiek2010: We still use it in many places.

shared/src/main/java/org/odk/collect/shared/strings/StringUtils.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/screens/ProjectDisplayPreferencesFragment.kt
grzesiek2010: > while deleting anything files in the root

you mean deleting every file? Don't you find it risky and generally bad? In the future we might need to store some other files in project dirs not only other nested dirs?

We could generate that file every time `getProjectRootDirPath()` is touched but then we would need for example to safe those files with a suffix like `.name` so that later we can remove all files with such a suffix before creating a new one...

Maybe it's a thing for a separate pr.

seadowg: > you mean deleting every file? Don't you find it risky and generally bad? In the future we might need to store some other files in project dirs not only other nested dirs?

Right now, we don't store any other files in the project dir root, so we could safely remove any of them, but I take your point it's a little excessive maybe, and I guess it might be a detail that's missed in the future if we do store other files there.

> We could generate that file every time getProjectRootDirPath() is touched but then we would need for example to safe those files with a suffix like .name so that later we can remove all files with such a suffix before creating a new one...

Yeah, I like that! Another alternative could be `.project`? That would also make it easier to script accessing these directories (you'd just grab the name from the `.name` file). I think you're correct that that's a change we could make later though. I can write up an issue and for the moment we can put up with needing the extra UI test.

seadowg (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1, 7.1 and 11

Tested cases:
- create project with special charts in project name
- remove project directory manually
- delete project
- scan qr code of project with special charts in project name 
- different charts combination
- exact chart combinations as described in issue

mmarciniak90:
Tested with success on Android versions: 8.1, 9.0, 10.0

seadowg:
@grzesiek2010 ach looks like we have a conflict here

PR #4857, opened 2021-10-13
Add `allow-mock-accuracy` for non-default `geopoint` and `geoshape`/`geotrace`

### LINKED ISSUE #4768: Provide a path for external gps sensors to report accuracy

seadowg:
Closes #4768 
~~Blocked by #4842~~

Every kind of geo/location widget should now support `allow-mock-accuracy`. Again, the refactor here  (moving the geo code to `geo`) is nice and gradual so going through commit by commit is a good idea.

New tests.

This takes the same approach as the last PR so not a lot to discuss here. This time most of the work is focused on moving `GeoPointMapActivity` and then adding support for the `allow-mock-accuracy` param in the same way we did for `GeoPointActivity`.

It's feeling to me like there's probably a `maps` module to be extracted at some point that contains the `MapFragment`, `MapPoint` and the providers/implementations (or possibly multiple modules for different implementations). That feels like an extra step we could take next time we're playing around in this area however.

Good to check the map appearances of `geopoint` with the different map providers - Mapbox should require special attention because it uses a different code path for grabbing the device's location. `geoshape` and `geotrace` need checked as well.

[The form](https://github.com/getodk/collect/files/7285696/external_gevice_geopoint.zip) used in the previous PR is a good starting point. Testing the specific support in this PR can be done by adding the map appearances to the form.


collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -112,8 +114,15 @@
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        previousState = savedInstanceState;
+
         DaggerUtils.getComponent(this).inject(this);
 
+        if (!permissionsProvider.areLocationPermissionsGranted()) {
```
grzesiek2010, line 121: Before this method was in `BaseGeoMapActivity` extended by this class but also by `GeoPointMapActivity` and `GeoPolyActivity`. Are you sure it's no longer needed in those two last classes? What if a user minimizes the app, disables location permissions and reopens the app?

geo/src/main/java/org/odk/collect/geo/GeoPointMapActivity.java
```diff
@@ -70,8 +68,16 @@
     public static final String LOCATION_STATUS_VISIBILITY_KEY = "location_status_visibility";
     public static final String LOCATION_INFO_VISIBILITY_KEY = "location_info_visibility";
 
+    public static final String EXTRA_LOCATION = "gp";
```
grzesiek2010, line 73: Why you didn't move this one to `geo/Constants.kt` too like `EXTRA_READ_ONLY` and  `EXTRA_DRAGGABLE_ONLY`?

geo/src/main/java/org/odk/collect/geo/GeoPointMapActivity.java
```diff
@@ -177,7 +185,7 @@ public void returnLocation() {
         }
 
         if (result != null) {
-            setResult(RESULT_OK, new Intent().putExtra(FormEntryActivity.ANSWER_KEY, result));
+            setResult(RESULT_OK, new Intent().putExtra("value", result));
```
grzesiek2010: Now we don't use the same constant anymore what might be error prone. It would be good to move that const to a shared module if now we need to use it from different modules.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -137,9 +137,15 @@ the specific language governing permissions and limitations under the License.
         <activity
             android:name="org.odk.collect.geo.GeoPointActivity"
             android:theme="@style/Theme.Collect"
-            tools:replace="android:theme"/>
-        <activity android:name=".activities.GeoPointMapActivity" />
-        <activity android:name=".activities.GeoPolyActivity" />
+            tools:replace="android:theme" />
```
grzesiek2010, line 140: What is it for?

geo/src/test/java/org/odk/collect/geo/GeoPolyActivityTest.java
```diff
@@ -0,0 +1,119 @@
...
+            activity.updateRecordingMode(R.id.automatic_mode);
+            activity.startInput();
+            verify(locationTracker).start(true);
+        });
+
+        intent.putExtra(Constants.EXTRA_RETAIN_MOCK_ACCURACY, false);
+        ActivityScenario.<GeoPolyActivity>launch(intent).onActivity(activity -> {
+            activity.updateRecordingMode(R.id.automatic_mode);
+            activity.startInput();
+            verify(locationTracker).start(true);
```
grzesiek2010: Why it's `true` when you passed false to `EXTRA_RETAIN_MOCK_ACCURACY`?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
seadowg: You're right about the minimize/deny permission case though as I'd forgotten that that kills the app and then reloads the same backstack (rather than starting from fresh). Good catch - we missed this when reviewing #4829. Right now if you perform those steps on a default `geopoint` question (`GeoPointActivity`) you get a crash caused by a `SecurityException`. 

Really we should only handle the permission at the Activity level instead of in `ActivityGeoDataRequester`. I'll make that change in this PR.

geo/src/main/java/org/odk/collect/geo/GeoPointMapActivity.java
seadowg: This one is only relevant to `GeoPointMapActivity`. I tried to keep it so `Constants` was only defining the shared ones.

geo/src/main/java/org/odk/collect/geo/GeoPointMapActivity.java
seadowg: Yeah I had a comment about this on the last PR. We also assume external apps use `"value"` so it's fairly well established. I'm not sure exactly where to define it given that it's not even specific to the Collect app. Maybe we could have a helper for Activities that return data to `FormEntryActivity` that external apps could copy? Kind of like without content provider API objects...

collect_app/src/main/AndroidManifest.xml
seadowg: It's need so that the build tools now which `android:theme` to use. Otherwise, there is a conflict between the `geo` and the `collect_app` manifests.

geo/src/test/java/org/odk/collect/geo/GeoPolyActivityTest.java
seadowg: WAT. That's a mistake. Not sure why the test passes though as the implementation looks right. Will have to look at that.

seadowg: Ah it's just that `start(true)` was called before. I'll add a `Mockito.reset` to make this test safer. This is a good example of where mocking frameworks can be dangerous in comparison to simple stateful fakes!

geo/src/main/java/org/odk/collect/geo/GeoActivityUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import android.Manifest
+import android.app.Activity
+import org.odk.collect.androidshared.system.PermissionsChecker
+import org.odk.collect.androidshared.ui.ToastUtils
+
+internal object GeoActivityUtils {
+
+    @JvmStatic
+    fun requireLocationPermissions(activity: Activity) {
```
seadowg, line 11: I don't like that we check this in the `Activity` and in `ActivityGeoDataRequester`. However, I don't want to get into untangling the interaction with `WaitingForDataRegistry` here to make sure that it's safe to just check this in the `Activity`.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -220,6 +220,10 @@
     private static final boolean EVALUATE_CONSTRAINTS = true;
     public static final boolean DO_NOT_EVALUATE_CONSTRAINTS = false;
 
+    /**
+     * Should use {@link org.odk.collect.android.externaldata.ExternalAppsUtils} instead
+     */
+    @Deprecated
```
seadowg, line 226: I didn't want to add the work/risk of replacing current usages of `ANSWER_KEY` in this PR but also wanted to make it clear we have a better option now.

grzesiek2010 (approved):

kkrawczyk123:
@seadowg I can see a problem with the display of Google Maps when using all Geo widgets (point, trace, shape) on all of my devices. It loads properly for display points on map. Issue does not occur for Mapbox and OSM. See gif:
ezgif com-gif-maker (7)

The issue does not occur on master.

seadowg:
@kkrawczyk123 I'll look into that

seadowg:
@kkrawczyk123 fixed! 

@grzesiek2010 it looks like the `MapProvider` needs to be a singleton. I've noted this on the class now to hopefully avoid the same mistake being made again, but you might want to take a look and see if there's anything you think we need to address there.

kkrawczyk123:
@seadowg I have tried that form https://test.getodk.cloud/#/projects/308/forms/external_device_geopoint/versions with Fake GPS app and my understanding is that fake location should be used only in the first question where mock accuracy=true but I can see that it is used also in the second question where mock accuracy=false. Is it an issue or did I misunderstand something?

seadowg:
@kkrawczyk123 the fake location will still be used, but its accuracy should be zero when `allow-mock-accuracy` is `false` (or absent). Is that not what you're seeing? The idea is just to remove accuracy from fake locations but allow people that have to use fake locations to be able to keep it if they trust it.

kkrawczyk123:
@seadowg I can see accuracy 0 for both when allow-mock-accuracy is true and false:
|    |    | 
| --- | ---- |
| Screenshot (Oct 22, 2021 9_13_45 AM) | Screenshot (Oct 22, 2021 9_13_54 AM) |

I assumed that there should be a difference but maybe I am wrong.

seadowg:
@kkrawczyk123 is this for all the geo widgets or just one of them? I'll check this out my side as well as it sounds like something it wrong.

seadowg:
@kkrawczyk123 I'm seeing the correct behaviour using mock the [GPS Connector](https://play.google.com/store/apps/details?id=de.pilablu.gpsconnector&hl=en_US&gl=US) app in simulation mode for geopoint (and maps appearances) and geotrace - I get points with 0 accuracy when `allow-mock-accuracy` is `false` and non zero when it's `true`. 

Are you potentially testing with manual placement (moving the map to somewhere and then placing a point)? This always results in zero accuracy as the location doesn't come from gps.

kkrawczyk123:
@seadowg I agree that accuracy difference is visible between both questions for geopoint with map view, geotrace and geoshape widgets but my question was about geopoint without map view - exact form that was attached in PR.

seadowg:
@kkrawczyk123 yeah I'm using that as well and still getting accuracy in the "Location (allows mock)" question. Is that happening every time? And it doesn't happen for the other geo widgets? What app are you using to fake the location? I'm wondering if the app is maybe giving you zero accuracy - I just got a `0.01m` out of the simulation in the one I was using.

kkrawczyk123:
@seadowg I can see accuracy 0 for both questions. It is happening every time for me. @mmarciniak90 Are you seeing the same thing on your devices? I am using the same app "Fake GPS" for that widget and others and there is a difference as I mentioned above it works for widgets with map view. I have set accuracy in the app to be different than 0 - I have 1 on one device and 20 on another and the same results - accuracy 0 for questions with and without allow-mock accuracy when filling a form.

seadowg:
@kkrawczyk123 could link me to the exact app?

kkrawczyk123:
sure, https://play.google.com/store/apps/details?id=com.lexa.fakegps&hl=en&gl=US

seadowg:
@kkrawczyk123 I get accuracy with that app. Changing its "Accuracy" setting is reflected in Collect as well (although never exactly for some reason, but that could be the app).

mmarciniak90:
> Are you seeing the same thing on your devices?

Unfortunately, I have the same result. I also use the same app to generate fake GPS.

kkrawczyk123:
@seadowg That's weird. What app have you been using before? We could try with it

seadowg:
Ok turns out I'd got myself confused about the release state of the `allow-mock-accuracy` changes in pyxform. It's out on https://getodk.org/xlsform/ but not released for Central yet. My bad! I've shared a form created with the newest pyxform with QA for them to test with.

kkrawczyk123:
**Finally tested with success!**

Verified cases:
- form from https://test.getodk.cloud/#/projects/306/forms/external_device_geopoint
- Fake GPS app
- GPS Connector app

regression checks on:
- Geo widgets with/without map
- GeoTrace, GeoShape widgets
- display submissions on map
- submissions with different statuses
- Google maps
- OSM
- Mapbox
- offline layers
- light mode, dark mode
- Audit location in the background
- screen rotation
- minimize app

mmarciniak90:
Tested with success on Android 8.1, 9.0, 10.0

PR #4859, opened 2021-10-13
Fix persian joda time dependency declaration

seadowg:
Allows the build to pull the PersianJodaTime dep again.


lognaturel (approved):
It compiles.

PR #4860, opened 2021-10-14
Fixed error prone registerToClearAnswerOnLongPress() method for string widget

### LINKED ISSUE #4834: Integer questions do not play video
Collect 2012.2.4

#### Problem description
Reported at https://forum.getodk.org/t/integer-questions-do-not-play-videos/35175. On integer question types, the video do not play. The video button is displayed but the video is not opened

#### Other information 
[Example Problem.zip](https://github.com/getodk/collect/files/7232052/Example.Problem.zip)

grzesiek2010:
Closes #4834 

I tested the fix manually and ran automated tests.

The issue affects all string widgets (text, integer, decimal). It is because in `StringWidget#registerToClearAnswerOnLongPress()` we override ids and then we can't react to click events in [AudioVideoImageTextLabel](https://github.com/getodk/collect/pull/4860/files#diff-22969420ab3442906012d0bb852a6b38a6ed9987ce6ebefa324a852265cfa1b0L219) by using those static ids.

The previous solution (draining all sub views and setting their ids) was error prone. We can just navigate through the first level of sub vies (only containers) and register those whole views except `answer_container` which contains the EditText to support the clipboard menu.

It should just fix the issue but please test removing answers (especially in field-list groups) to make sure nothing is spoiled there.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
@@ -105,21 +105,17 @@ public class StringWidget extends QuestionWidget {
     }
 
     /**
-     * Registers all subviews except for the EditText to clear on long press. This makes it possible
-     * to long-press to paste or perform other text editing functions.
+     * Registers all subviews except for the answer_container (which contains the EditText) to clear on long press.
+     * This makes it possible to long-press to paste or perform other text editing functions.
      */
     @Override
     protected void registerToClearAnswerOnLongPress(FormEntryActivity activity, ViewGroup viewGroup) {
-        for (int i = 0; i < viewGroup.getChildCount(); i++) {
-            View child = viewGroup.getChildAt(i);
-            if (child.getId() == R.id.help_layout) {
-                child.setId(getId());
-                activity.registerForContextMenu(child);
-            } else if (child instanceof ViewGroup) {
-                registerToClearAnswerOnLongPress(activity, (ViewGroup) child);
-            } else if (!(child instanceof EditText)) {
-                child.setId(getId());
-                activity.registerForContextMenu(child);
+        ViewGroup view = findViewById(R.id.question_widget_container);
+        for (int i = 0; i < view.getChildCount(); i++) {
+            View childView = view.getChildAt(i);
+            if (childView.getId() != R.id.answer_container) {
+                childView.setId(getId());
+                activity.registerForContextMenu(childView);
             }
         }
     }


seadowg:
> Another solution would be to avoid draining audio_video_image_text_label layout (and overriding ids) in StringWidget#registerToClearAnswerOnLongPress() but then long press wouldn't be supported.

It does feel to me like it would be best to try and clean up what's happening in `StringWidget#registerToClearAnswerOnLongPress()` as it seems like it could definitely cause other bugs in the future. This issue has existed for a while, so I think we're ok to take a little longer on it. 

I don't understand why we have to mess with view IDs in `registerToClearAnswerOnLongPress()`. Could we not explicitly register the answer view and question label separately (instead of just registering the whole view) to avoid this kind of problem? Then we could just not register the answer view in our string widgets that use an `EditText`. Maybe I'm missing some complexities here though!

seadowg (changes requested):
> Another solution would be to avoid draining audio_video_image_text_label layout (and overriding ids) in StringWidget#registerToClearAnswerOnLongPress() but then long press wouldn't be supported.

It does feel to me like it would be best to try and clean up what's happening in `StringWidget#registerToClearAnswerOnLongPress()` as it seems like it could definitely cause other bugs in the future. This issue has existed for a while, so I think we're ok to take a little longer on it. 

I don't understand why we have to mess with view IDs in `registerToClearAnswerOnLongPress()`. Could we not explicitly register the answer view and question label separately (instead of just registering the whole view) to avoid this kind of problem? Then we could just not register the answer view in our string widgets that use an `EditText`. Maybe I'm missing some complexities here though!

grzesiek2010:
> Could we not explicitly register the answer view and question label separately (instead of just registering the whole view) to avoid this kind of problem?

clipboard menu won't work in that case (in EditText). **EDIT: I think your idea was close to mine that I described below (I misunderstood something)**

> I don't understand why we have to mess with view IDs in registerToClearAnswerOnLongPress()

It's because we need to have different ids otherwise if we have field-list group removing answers might not work properly.

You know what... a simpler and less error prone solution would be to `registerForContextMenu` only the `AudioVideoImageTextLabel` so that we support both: our contextual menu and clipboard menu and avoid draining the whole layout.
It would look like this: https://github.com/getodk/collect/commit/cb4ea6cae8742aefad2a02fbc021cec32db241f1
 
So in case of String widgets we would register just the `AudioVideoImageTextLabel` an not answer view to support clipboard menu bot for all the other widgets everything will work in the same way as before.

what do you think?

seadowg:
>  EDIT: I think your idea was close to mine that I described below (I misunderstood something)

Yeah it looks like we're on the same page - register the views explicitly and just don't do it for the answer view. I was thinking that we probably also need to register the `help_text` part of the `question_widget` layout as well right? Also, I think we want to just register the whole widget for the `Ex...` widgets right? In that case there isn't an edit text to avoid.

grzesiek2010:
> Yeah it looks like we're on the same page - register the views explicitly and just don't do it for the answer view. I was thinking that we probably also need to register the help_text part of the question_widget layout as well right?

Right, my current implementation handles it. I navigate through the first level of sub vies and register them (except for the `answer_container`) and I don't drain that sub views like before what was error prone.

> Also, I think we want to just register the whole widget for the Ex... widgets right? In that case there isn't an edit text to avoid.

There is an edit text which is displayed if there is no activity to handle such an external widget.

seadowg (changes requested):
This is looking so much better/simpler than the existing implementation! 

I think we should cement this in with a test that checks that all the subviews (other than the answer container) have had their ID changed to the widget ID and have been registered with the Activity - I'm guessing this could be done by overriding `registerForContextMenu` on the test Activity and then recording the views it's passed.

I'll make this ready for QA as I don't think the test will disrupt them.

mmarciniak90:
**Tested with success**

Verified cases:
- play video in form from issue
- text, integer, decimal
- field-list groups
- removing answers in simple view
- removing answers in field-list groups
- read-only widgets, with and without answers 
- rotation, minimization 
- light/dark mode

kkrawczyk123:
Tested on Androids 5.1, 7.1 and 11.
When testing on Android 11 I have noticed an issue with playing video attachments in other forms as well - reported in #4876

seadowg:
I'm going to merge this now so it can be in for regression testing. Are you able to follow up with a testing I mentioned @grzesiek2010?

seadowg (approved):

grzesiek2010:
> Are you able to follow up with a testing I mentioned

yeah I'll try to write some valuable tests for this.

PR #4861, opened 2021-10-14
v2021.3 analytics

### LINKED ISSUE #4847: Add analytics for v2021.3
Things we should measure from 2021.3 are:

- [x] How many devices select light/dark theme (over system default)
- [x] How many forms include `allow-mock-accuracy=true`
~~- [ ] How many devices include a layer from the new shared directory~~

EDIT: I don't think we actually need to measure adoption of the new shared `layers` directory. We know a few projects have asked for it, and we seem pretty confident that use cases that would require it exist currently. I think that's something we could measure if we ever wanted to get rid of it. `allow-mock-accuracy` feels worth measuring however as we don't have a good sense of how popular external devices really are, and it will be interesting to track. This will mean we can be more informed if we get requests to tweak/expand external device support.

seadowg:
Closes #4847

Definitely best to go through this commit by commit as removing a constructor param changed a bunch of test files.

Ran tests to check nothing had broken. No tests for analytics.

Comments inline.

Shouldn't have any effect on behaviour. Can skip QA.

### DIFF
--- a/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
+++ b/analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
@@ -17,6 +17,7 @@ interface Analytics {
     companion object {
 
         private var instance: Analytics = NoopAnalytics()
+        private val params = mutableMapOf<String, String>()
 
         fun setInstance(analytics: Analytics) {
             this.instance = analytics
@@ -28,8 +29,24 @@ interface Analytics {
         }
 
         @JvmStatic
+        fun log(event: String, key: String) {
+            val paramValue = params[key]
+
+            if (paramValue != null) {
+                log(event, key, paramValue)
+            } else {
+                log(event)
+            }
+        }
+
+        @JvmStatic
         fun log(event: String, key: String, value: String) {
             instance.logEventWithParam(event, key, value)
+        }
+
+        @JvmStatic
+        fun setParam(key: String, value: String) {
+            params[key] = value
         }
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/questions/QuestionDetails.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/questions/QuestionDetails.java
@@ -10,25 +10,19 @@ import org.odk.collect.android.widgets.QuestionWidget;
 public class QuestionDetails {
 
     private final FormEntryPrompt prompt;
-    private final String formAnalyticsID;
     private final boolean isReadOnly;
 
-    public QuestionDetails(FormEntryPrompt prompt, String formAnalyticsID) {
-        this(prompt, formAnalyticsID, false);
+    public QuestionDetails(FormEntryPrompt prompt) {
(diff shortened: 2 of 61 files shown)


analytics/src/main/java/org/odk/collect/analytics/Analytics.kt
```diff
@@ -27,9 +28,25 @@ interface Analytics {
...
+            }
+        }
+
         @JvmStatic
         fun log(event: String, key: String, value: String) {
             instance.logEventWithParam(event, key, value)
         }
+
+        @JvmStatic
+        fun setParam(key: String, value: String) {
```
seadowg, line 48: I'm basically emulating `Timber` with the style of this static interface. Why is it ok for this state to be static? In my mind, these kinds of "ancillary" features like logging and analytics can use static references and state for the following reasons:

1. They are never essential to the parts of the code that call them
2. Given the above reason, interactions with them should never need to be tested
3. No app behaviour is in any way reliant on the static state being held
4. We're only storing a reference to data (a `String`) rather than a reference to a shared component (like `FormController` or a UI component)

And also, storing the form hash statically like this will allow us to stop passing analytics specific information down to components which hopefully reduces the noise added by analytics.

collect_app/src/main/java/org/odk/collect/android/formentry/questions/QuestionDetails.java
```diff
@@ -10,27 +10,21 @@
...
         this.prompt = prompt;
-        this.formAnalyticsID = formAnalyticsID;
         this.isReadOnly = readOnlyOverride || prompt.isReadOnly();
     }
 
     public FormEntryPrompt getPrompt() {
         return prompt;
     }
 
-    public String getFormAnalyticsID() {
```
seadowg, line 30: No longer need this as the form hash will be stored statically.

lognaturel (approved):
I like this.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -99,6 +103,15 @@ static String truncateDouble(String s) {
         return "";
     }
 
+    public static void logAllowMockAccuracy(FormEntryPrompt prompt) {
```
grzesiek2010, line 106: Wouldn't it be easier to log this event in `ActivityGeoDataRequester`? This is the only place where you check if the option is enabled and then it would be in one place instead of in every geo widget.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
seadowg: Yeah I thought about that. So in that case the even would only fire if the button is tapped, so we might not see it for optional questions? It's not a big deal but felt like having it fire when the widget is displayed made more sense.

PR #4865, opened 2021-10-15
Improvements to CI workflow

seadowg:
This drops the heap size we use on CI in an attempt to make our workflow more stable (avoiding OOM errors) and also cleans up some duplicated/unused CI config. I've also added parallelism to our `test_modules` which cuts its run time from around 5 mins (becoming the longest job in that part of the workflow) to around 2 mins.

Ran CI as part of this PR.

Comments inline.

No changes in behaviour. Can skip QA.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,7 +168,7 @@ jobs:
 
       - run:
           name: Assemble test build
-          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 assembleDebug assembleDebugAndroidTest
+          command: ./gradlew -PdisablePreDex --no-daemon --max-workers=2 assembleDebugAndroidTest
       - run:
           name: Authorize gcloud
           command: |
@@ -218,9 +218,7 @@ workflows:
               ignore: master
       - test_instrumented:
           requires:
-            - check_quality
-            - test_modules
-            - test_app
+            - compile
           filters:
             branches:
               only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,7 @@ jobs:
 
   test_modules:
     <<: *android_config
+    parallelism: 4
     steps:
       - attach_workspace:
           at: ~/work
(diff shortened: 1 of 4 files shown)


.circleci/test_modules.txt
```diff
@@ -0,0 +1,10 @@
+shared:test
+androidshared:testDebug
+formstest:test
+async:testDebug
+strings:testDebug
+audioclips:testDebug
+audiorecorder:testDebug
+projects:testDebug
+location:testDebug
+geo:testDebug
```
seadowg, line 10: In future when extracting a module, we can just add it here rather than having to change `config.yaml`.

.circleci/config.yml
```diff
@@ -218,9 +200,7 @@ workflows:
               ignore: master
       - test_instrumented:
           requires:
-            - check_quality
-            - test_modules
-            - test_app
+            - compile
```
seadowg, line 203: There's no reason to wait for everything else to kick this off. Running instrumentation tests in parallel to our other jobs should cut the full run time of our workflow.

.circleci/test_modules.txt
lognaturel: The implications of not adding it here are just going to be that tests don't run in CI, right? Seems like the kind of thing that's likely to be forgotten. Should it at least be in the README?

seadowg: You needed to add a specific step to the `test_module` before so if anything it's simpler than it was. For me, the Circle CI setup feels like it's self documenting to some degree. I was thinking of adding something around adding/extracting new modules though (probably to the code style guideline) and that could include notes on CI?

lognaturel: Yes, simpler, but feels less visible? Planning on adding a section on modules in style guidelines sounds great.

lognaturel (approved):

lognaturel:
@seadowg letting you do merges so you can pick order/timing

Issue #4198, opened 2020-11-02
Display sub-meter accuracies using cm


lognaturel:
#### Software and hardware versions 
Collect v1.28 and prior

#### Problem description
It's now somewhat practical to get sub-meter location accuracy readings and it's something @IvanGayton and others have been actively working on. Currently sub-meter accuracies are displayed using decimals. Since https://github.com/getodk/collect/issues/3336, two decimal points are included.

#### Steps to reproduce the problem
1. Get high-precision GPS equipment connected to a device
1. Get a sub-meter accuracy reading

This is going to be hard to reproduce because it generally will require specialized equipment. I think we can make the change just with tests and without having to reproduce.

#### Expected behavior
Accuracy readings under a meter should be displayed in centimeters. I don't think there should be any decimals.

https://twitter.com/SmartGeospatial/status/1322609245968281600

lognaturel:
@ivangayton How important is it to you/your users to see sub-meter accuracy in centimeters? That is, is seeing an accuracy of `0.01m` problematic?

ivangayton:
> @ivangayton How important is it to you/your users to see sub-meter accuracy in centimeters? That is, is seeing an accuracy of `0.01m` problematic?

ivangayton:
Ah, sorry @lognaturel I missed the notification here!

It's not important; decimal meters is totally fine. The kind of users who, like me, will sometimes be worried about centimeter-level accuracy will be absolutely fine with some decimal places. 

People using high-end surveying gear tend to use centimeters to describe accuracy, but it's definitely not a formal convention (and some people use millimeters anyway, so even the informal convention isn't really standardized).

On the very remote chance that someone might be using civil engineering grade GPS gear ($100,000 units that can achieve millimeter accuracy) with ODK, it could make sense to allow up to three decimal places for readings in meters, but I can't really imagine a circumstance where I or anyone I work with would need more than two decimal places.

PR #4866, opened 2021-10-16
Address #4198 Display sub-meter accuracies…

dimwight:
Addresses #4198

See 'Simulation' and 'Testing' below.

See 'The issue' and 'The update' below.

None?

_[geowidgets.xlsx](
https://github.com/getodk/collect/files/735789/geowidgets.xlsx
)_ used for development testing, see 'Simulation' below.

No.

### The issue
#4198 is essentially about display formatting, so the natural starting point in attacking it was to identify strings that display GPS accuracy. They seem to be the following:
- `gps_result` used by `getGeoPointAnswerToDisplay` in `GeoWidgetUtils`, called by `GeoPointActivity` and by `GeoPointMapActivity`
- `location_status_accuracy` etc used by `updateUi` in `GeoPolyActivity`
- `location_accuracy` used by `getAccuracyMessage` in `GeoPointActivity`, and by 
`formatLocationStatus` in `GeoPointMapActivity`

### The update
Each format string now needs two variants, for formatting accuracy as _cm_ or _m_. So code using these strings has to look at the accuracy value so as to
- decide which format to use, and
- where the answer is _cm_, scale the value to match ie multiply by 100.

This is coded independently for `gps_result` and `location_status_accuracy` etc, with the complication that `getGeoPointAnswerToDisplay` which uses `gps_result` has to extract the accuracy value from its `answer` parameter. 

To avoid duplicate code for `location_accuracy`, `getAccuracyMessage` and `formatLocationStatus` now get their formatted string from `getAccuracyUnitString` in `GeoUtils`. 

### Simulation
In addition to the formal testing described below, there was a need to simulate cm accuracy so as to confirm coverage of all occurrences in the UI. 

So for now at least `SIMULATED_ACCURACY` has been added to `GeoUtils`, for use when `simulateAccuracy` is set, and intercepts accuracy values to enable manual confirmation that _cm_ formatting is applied when appropriate (including a very dubious mod to `FormEntryPromptUtils`). 

The form defined in _[geowidgets.xlsx](
https://github.com/getodk/collect/files/735789/geowidgets.xlsx
)_ can be used to view the simulation.

### Testing
The new code is tested by three new methods added to existing classes, all called `locationAccuracyIsFormattedInAppropriateUnit` since they perform essentially the same task by iterating through the `TEST_ACCURACIES` added to `GeoUtils`. 

The variants in `GeoWidgetUtilsTest` and `GeoUtilsTest` check results for `gps_result` and `location_accuracy` as returned by their respective methods. In `GeoPolyActivityTest` the test is of the state of `locationStatus` following each fake fix. 

The new implementation of `getGeoPointAnswerToDisplay` necessitates an update to `getAnswerToDisplay…returnsAnswer` in  `GeoWidgetUtilsTest`, which in fact only tests currently for accuracies below 1 m.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/GeoPointMapActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/GeoPointMapActivity.java
@@ -32,14 +32,12 @@ import androidx.annotation.VisibleForTesting;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.geo.MapFragment;
-import org.odk.collect.geo.MapPoint;
 import org.odk.collect.android.geo.MapProvider;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.preferences.screens.MapsPreferencesFragment;
-import org.odk.collect.geo.GeoUtils;
 import org.odk.collect.androidshared.utils.ToastUtils;
-
-import java.text.DecimalFormat;
+import org.odk.collect.geo.GeoUtils;
+import org.odk.collect.geo.MapPoint;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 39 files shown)


dimwight:
Having finally got this to pass all tests I have serious concerns about the  localisation effort it implies, which I hadn't previously considered.  

For a cosmetic change that will be meaningful to very few users in at least the medium term, it's hardly fair to ask potentially twenty-five contributors for help in tidying it up.

So I'm closing this PR while I work out how at least to reduce this burden.

PR #4867, opened 2021-10-18
Fixed visibility of the location crosshair in mapbox

### LINKED ISSUE #4858: The location crosshair is not visible on Mapbox
Collect master

#### Problem description
The location crosshair is not visible on Mapbox

1. User uses Mapbox 
2. User opens Geowidget with map appearance
3. User sees a blue circle but the crosshair is not visible

#### Expected behavior
The user's location is displayed using a crosshair at the location

#### Other information 
Noticed during testing #4852

grzesiek2010:
Closes #4858

I tested the fix manually.

This seems to be a bug in mapbox which I reported here https://github.com/mapbox/mapbox-gl-native-android/issues/716 so this seems to be the only solution for now.

Please just verify that the issue is fixed. Nothing else should be affected.

No.

No.

### DIFF
--- a/collect_app/src/main/res/drawable/ic_crosshairs.xml
+++ b/collect_app/src/main/res/drawable/ic_crosshairs.xml
@@ -3,7 +3,9 @@
         android:height="48dp"
         android:viewportWidth="48"
         android:viewportHeight="48">
+
+    <!-- Do not use theme attributes to define the color https://github.com/mapbox/mapbox-gl-native-android/issues/716 -->
     <path
-        android:fillColor="?colorPrimary"
+        android:fillColor="@color/colorPrimary"
         android:pathData="M18,23v2h-18v-2h18z m12,0h18v2h-18v-2z m-7,-5v-18h2v18h-2z m0,12h2v18h-2v-18z" />
 </vector>


collect_app/src/main/res/drawable/ic_crosshairs.xml
```diff
@@ -3,7 +3,9 @@
         android:height="48dp"
         android:viewportWidth="48"
         android:viewportHeight="48">
+
+    <!-- Do not use theme attributes to define the color https://github.com/mapbox/mapbox-gl-native-android/issues/716 -->
```
seadowg, line 7: Wow that's annoying!

seadowg (approved):

kkrawczyk123:
@grzesiek2010 I have noticed that crosshair disappears when changing offline layers, see gif:
ezgif com-gif-maker (6)

A similar issue is visible when an offline layer is selected and minimize app. When go back to app crosshair is not visible. Issue does not occur when minimize app when offline layer is not selected. It was hard to find Collect version where crosshair was visible. I tried 1.26 where both cases that I am describing works well. Could you please take a look?

grzesiek2010:
This is a different bug, please report a separate issue for it.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 7.1 and 11

Tested cases:
- GeoPoint, GeoTrace, GeoShape
- Mapbox, Google maps and OSM
- all input methods
- offline layer
- screen rotation
- minimize app
- map view with displayed points 
- issue described in the comment above reported in #4873

mmarciniak90:
Tested with success on Android 8.1, 9.0, 10.0

PR #4868, opened 2021-10-18
Extract error UI to new module

seadowg:
As discussed in #4830. This would have been pretty awkward at the time, but the changes made when extracting the `geo` module (moving most of `CollectAbstractActivity` out to shared helpers) made this pretty simple.

I've also added some notes to our docs about adding new modules seeing as that's becoming a more and more common thing for us to be doing.

There isn't a lot of code moved out to the new module here, but I think it's just enough to be worth it. I think the main advantage of a separate module here is enforcing the dependency direction - the errors UI code should never depend on "Collect specific" code. I do wish that there wasn't so much boilerplate around the gradle files for every module but we're doing a good job of making those as easy as possible to create with changes like our new `buildSrc` dependency constants.

Just moving code around and not a lot seems risky here. Can skip QA.


androidshared/src/main/res/layout/app_bar_layout.xml
```diff
@@ -0,0 +1,16 @@
...
+    android:id="@+id/appBarLayout"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:elevation="4dp">
+
+    <com.google.android.material.appbar.MaterialToolbar
+        android:id="@+id/toolbar"
+        style="@style/Widget.AndroidShared.Toolbar.Surface.HighEmphasisControls"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
```
grzesiek2010: Shouldn't this be `?attr/actionBarSize` like in https://material.io/components/app-bars-top/android#regular-top-app-bar?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/BasePreferencesFragment.kt
```diff
@@ -43,9 +44,12 @@ abstract class BasePreferencesFragment : PreferenceFragmentCompat(), OnSettingCh
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        if (activity is CollectAbstractActivity) {
-            (activity as CollectAbstractActivity).initToolbar(preferenceScreen.title)
+        requireActivity().let {
+            if (it is AppCompatActivity) {
```
grzesiek2010: Do we need this check? I think we use `AppCompatActivity` pretty much everywhere now.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -7,12 +7,11 @@ import android.os.Bundle
...
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import org.odk.collect.android.R
-import org.odk.collect.android.errors.ErrorActivity
-import org.odk.collect.android.errors.ErrorItem
+import org.odk.collect.errors.ErrorActivity
 import java.util.ArrayList
 
 class FormsDownloadResultDialog : DialogFragment() {
-    private lateinit var failures: ArrayList<ErrorItem>
+    private lateinit var failures: ArrayList<org.odk.collect.errors.ErrorItem>
```
grzesiek2010: You can add import for the full path and here use only `ErrorItem`. The same in many other places.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -118,7 +118,9 @@ the specific language governing permissions and limitations under the License.
             android:name=".activities.FillBlankFormActivity"
             android:launchMode="singleTop" />
         <activity android:name=".activities.FormDownloadListActivity" />
-        <activity android:name="org.odk.collect.android.errors.ErrorActivity" />
+        <activity
```
grzesiek2010, line 123: Do we need this here? It's declared in errors module.

grzesiek2010 (changes requested):

collect_app/src/main/AndroidManifest.xml
seadowg: Yes so we can override the theme.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
seadowg: Oops! This looks like something that got added while moving the class.

collect_app/src/main/AndroidManifest.xml
grzesiek2010: I see so wouldn't it be better to make it possible to access the theme in other modules so that we can set it in `errors` module only?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/BasePreferencesFragment.kt
seadowg: We'd end up with a **technically** unsafe cast in the code without it as `requireActivity()` returns `FragmentActivity` rather than `AppCompatActivity` but as you say there's probably nowhere that we'd actually see a problem. I don't have strong feelings either way when it comes to the safe code vs exploding if we're using a non `AppCompatActivity`. What do you think?

androidshared/src/main/res/layout/app_bar_layout.xml
seadowg: Yup. Good catch.

collect_app/src/main/AndroidManifest.xml
seadowg: Personally I quite like being able to avoid using a shared component/interface here and just setting the theme in the manifest. It keeps the library modules simpler. An alternative would be doing something like `LocalizedApplication` for themes but, again, I think I prefer what we have?

collect_app/src/main/java/org/odk/collect/android/preferences/screens/BasePreferencesFragment.kt
grzesiek2010: I don't have strong feelings but I would like to avoid checks like that. It seems now like it's just one place but it can be anywhere too. Maybe that new `AppBarUtils` should take `Activity` and there check if it's `AppCompatActivity`?

seadowg: > Maybe that new AppBarUtils should take Activity and there check if it's AppCompatActivity

I like that! Keeps the code safe but also means we don't need to check everywhere. I'll make that change.

collect_app/src/main/AndroidManifest.xml
grzesiek2010: again no strong feelings so I could accept but now I'm thinking... If that ErrorActivity no longer extends CollectAbstractActivity (responsible for updating themes), what happens if we have `Use device theme` selected in settings and for example we have that activity opened in light mode and then we set to dark mode and reopen the app? Will the theme be updated properly?

seadowg: Great question. We do call `setTheme` in `CollectAbstractActivity` still, but the actual switching between light, dark and system happens in `MainMenuActivity#onCreate` (when returning from settings, switching projects or opening the app). 

The call in `CollectAbstractActivity` could actually be replaced with adding `theme` attributes to all our Activity declarations in the manifest (or potentially the [`application` `theme` attribute](https://developer.android.com/guide/topics/manifest/application-element#theme) which we're not using for some reason).

grzesiek2010: Ok so as I understand it's acceptable that in order to update the theme we need to go though `MainMenuActivity`? I'm asking because since we had this in `CollectAbstractActivity` which we used as a base activity for other activities I think we wanted to  have that theme updating everywhere.

If it's the expected behavior I think we should have a pr that cleans that up (removes that code from `CollectAbstractActivity`) because it's confusing - of course it might be a separate pull request.

collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/FormsDownloadResultDialog.kt
```diff
@@ -36,8 +36,8 @@ class FormsDownloadResultDialog : DialogFragment() {
 
         if (failures.isNotEmpty()) {
             builder.setNegativeButton(getString(R.string.show_details)) { _, _ ->
-                val intent = Intent(context, ErrorActivity::class.java).apply {
-                    putExtra(ErrorActivity.EXTRA_ERRORS, failures)
+                val intent = Intent(context, org.odk.collect.errors.ErrorActivity::class.java).apply {
+                    putExtra(org.odk.collect.errors.ErrorActivity.EXTRA_ERRORS, failures)
```
grzesiek2010: You can remove `org.odk.collect.errors` here too.

grzesiek2010 (changes requested):

collect_app/src/main/AndroidManifest.xml
seadowg: >  I'm asking because since we had this in CollectAbstractActivity which we used as a base activity for other activities I think we wanted to have that theme updating everywhere.

We're no longer updating the actual Android theme being used any longer - we just change the night mode setting that appcompat uses (using `AppCompatDelegate.setDefaultNightMode`) and we only need to do this when UI "theme" setting has changed. The night mode then dictates which resources get loaded (`-night` or not).

> If it's the expected behavior I think we should have a pr that cleans that up (removes that code from CollectAbstractActivity) because it's confusing - of course it might be a separate pull request.

I think you're right - it's confusing in the current state. I can do that clean up here as it's a refactor PR anyway and it'd be nice to not have to file an issue etc!

seadowg: Done!

collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
```diff
@@ -47,7 +45,6 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         DaggerUtils.getComponent(this).inject(this);
 
         themeUtils = new ThemeUtils(this);
```
grzesiek2010: This seems to be unused now so you can remove it.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -78,7 +78,8 @@ the specific language governing permissions and limitations under the License.
         android:largeHeap="true"
         android:networkSecurityConfig="@xml/network_security_config"
         android:supportsRtl="true"
-        android:usesCleartextTraffic="true">
+        android:usesCleartextTraffic="true"
+        android:theme="@style/Theme.Collect">
```
grzesiek2010, line 82: If we have it here we probably do not need it in geo activities?:
```
        <activity
            android:name="org.odk.collect.geo.GeoPointActivity"
            android:theme="@style/Theme.Collect"
            tools:replace="android:theme" />
```

grzesiek2010 (changes requested):

collect_app/src/main/AndroidManifest.xml
seadowg: The geo activities have a theme (`Theme.MaterialComponents`) set on them in their module manifest which allows them to function in isolation there for tests, so we override that in `collect_app`. For library activities we're basically overriding their "default" theme (hence the `tools:replace`). We don't currently set a default theme for `ErrorActivity` so we don't need to declare it in the `collect_app` manifest.

collect_app/src/main/java/org/odk/collect/android/activities/CollectAbstractActivity.java
seadowg: Huh good catch! I thought it was still being used.

grzesiek2010 (approved):

grzesiek2010:
Normally I would merge this pr without testing but we went the extra mile here what might affect setting themes so it would be good to test this a bit to make sure that everything is fine.

seadowg:
> Normally I would merge this pr without testing but we went the extra mile here what might affect setting themes so it would be good to test this a bit to make sure that everything is fine.

Agreed! For QA: it'd be good to have a brief play with the app (no area in particular) in the system, light and dark themes.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 9.0

Verified cases:
- system default theme: change mode in device's settings/use app
- light mode
- dark mode
- import settings scan QR code
- import settings from the file
- reset settings
- play around with other user interface settings: language, font size, navigation 
- add multiple projects with other theme settings

srujner:
Tested with success!

Verified cases:

- System default theme
- Light and dark mode
- Import settings from QR code and from file
- Reset and delete settings
- Exploratory testing through project and general settings

PR #4869, opened 2021-10-18
Make sure to build APK on same machine as test APK

seadowg:
This should allow instrumentation tests to run again.


lognaturel (approved):
Annoying!

PR #4875, opened 2021-10-20
Fixed disappearing location icon

### LINKED ISSUE #4873: Crosshair disappears when change offline layer and minimize app when use Mapbox
Collect v1.x.x, Android v all, device used...

#### Problem description
When testing #4867 I have noticed another issue connected to crosshair when using Mapbox with offline layers.

Scenario 1:
1. Add at least 1 offline layer
2. Set Mapbox in settings
3. Open map view (crosshair visible)
4. Select offline layer - crosshair is no longer visible
ezgif com-gif-maker (6)

Scenario 2:
1. Set Mapbox and offline layer in settings
2. Open Map view (crosshair visible)
3. Minimize app
4. Go back to app - crosshair is no longer visible

#### Expected behavior
Crosshair should not disappear when minimize app or select offline layers

#### Other information 
Both issues work correctly on v1.26

grzesiek2010:
Closes #4873 

I tested the fix manually.

We need to make sure that the location icon is above all the other layers. Since we have a placeholder layer (always on top) it was relatively easy (I just added `.layerAbove(PLACEHOLDER_LAYER_ID)`).
Unfortunately it started throwing exceptions because of how we implemented changing layers. When we change layers we set the whole map style what removes all layers and then we add all of them again: https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java#L268
before we do that mapbox internally makes some calls on the location component (which after adding `.layerAbove(PLACEHOLDER_LAYER_ID)` is tied to the placeholder layer) before we add that layer again (so it doesn't exist at that moment). That's why I had to add [resetLocationComponent()](https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4873?expand=1#diff-c8e8d852506b504077b378b6fc12b4c48a1737094bade407b4a443f8c8187c80R293) to temporarily remove that relation.
I don't know why our implementation works in that way instead of for example just removing the existing layer and adding a new one but I didn't want to touch it here.

It should just fix the issue so please verify it does the job.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
@@ -5,6 +5,7 @@ import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineColor;
 import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineOpacity;
 import static com.mapbox.mapboxsdk.style.layers.PropertyFactory.lineWidth;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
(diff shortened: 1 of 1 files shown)


seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android versions: 8.1, 10.0

Verified cases:
* set offline layer on map view -> crosshair is visible
* minimize when the offline layer is applied and crosshair is visible
* Mapbox, Google Maps, OSM
* Geopoint with a map view, Geoshape, Geotrace
* offline layer form shared directory
* offline layer from project directory
* different offline layers

kkrawczyk123:
Verified also on Androids 5.1, 9.0 and 11.

PR #4877, opened 2021-10-21
Test register for context menu

grzesiek2010:
I just reviewed implemented changes.

This is what we discussed in https://github.com/getodk/collect/pulls?q=is%3Apr+is%3Aclosed

It doesn't require testing.

No.

No.


collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -135,8 +135,8 @@ public QuestionWidget(Context context, QuestionDetails questionDetails) {
             addAnswerView(answerView);
         }
 
-        if (context instanceof FormEntryActivity && !questionDetails.isReadOnly()) {
-            registerToClearAnswerOnLongPress((FormEntryActivity) context, this);
+        if (context instanceof Activity && !questionDetails.isReadOnly()) {
```
seadowg, line 138: Good change. Anything we can do to severe the dependency from Widgets on `FormEntryActivity` is a step forward!

collect_app/src/test/java/org/odk/collect/android/support/TestScreenContextActivity.kt
```diff
@@ -0,0 +1,25 @@
+package org.odk.collect.android.support
+
+import android.view.View
+import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.LifecycleOwner
+import org.odk.collect.android.utilities.ScreenContext
+import java.util.ArrayList
+
+class TestScreenContextActivity : FragmentActivity(), ScreenContext {
```
seadowg: We can probably just call this `WidgetTestActivity` now it has a couple of different features for testing widgets with.

collect_app/src/test/java/org/odk/collect/android/support/TestScreenContextActivity.kt
```diff
@@ -0,0 +1,25 @@
...
+    @JvmField
+    val viewsRegisterForContextMenu: MutableList<View> = ArrayList()
```
seadowg: It's the same thing, but you could just use `mutableListO<View>()` here.

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
```diff
@@ -109,11 +109,12 @@ public void cancelLongPress() {
      * This makes it possible to long-press to paste or perform other text editing functions.
      */
     @Override
-    protected void registerToClearAnswerOnLongPress(FormEntryActivity activity, ViewGroup viewGroup) {
+    protected void registerToClearAnswerOnLongPress(Activity activity, ViewGroup viewGroup) {
         ViewGroup view = findViewById(R.id.question_widget_container);
         for (int i = 0; i < view.getChildCount(); i++) {
             View childView = view.getChildAt(i);
             if (childView.getId() != R.id.answer_container) {
+                childView.setTag(childView.getId());
```
seadowg, line 117: Do you think we should really be using the tag instead of ID as a way to propagate the widget ID? Overriding the ID feels pretty dodgy. It feels like a warning sign that you had to add the original ID to the tag here.

Maybe that's not a change we want to make in this PR as it would need a lot of QA or more testing, but I think it'd be good to discuss quickly here.

seadowg (changes requested):
Looks great! Just a couple of wee tweaks and one thing to discuss before moving on.

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
grzesiek2010: Our:
```
    @Override
    public boolean onContextItemSelected(MenuItem item) {
        if (item.getItemId() == DELETE_REPEAT) {
            DialogUtils.showIfNotShowing(DeleteRepeatDialogFragment.class, getSupportFragmentManager());
        } else {
            ODKView odkView = getCurrentViewIfODKView();
            if (odkView != null) {
                for (QuestionWidget qw : odkView.getWidgets()) {
                    if (item.getItemId() == qw.getId()) {
                        createClearDialog(qw);
                        break;
                    }
                }
            }
        }

        return super.onContextItemSelected(item);
    }
```

is based on ids but I agree that changing them (like we do now) is error prone. We could swap it, set tags for every view we register for context menu and use tags in `onContextItemSelected()`. Is that what you meant?

collect_app/src/test/java/org/odk/collect/android/support/TestScreenContextActivity.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/support/TestScreenContextActivity.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/widgets/StringWidget.java
seadowg: Yeah exactly: use the tags as the identifier rather than view ID. Unless you disagree I think that's something we should do another time though? Just good to chat about it so that we've got it in mind for next time we touch that area.

seadowg (approved):

seadowg:
Whoops. Must have been on autopilot when I added "needs testing" to this. Sorry if you already looked at it @kkrawczyk123 @mmarciniak90.

PR #4878, opened 2021-10-26
Fixed starting activities

### LINKED ISSUE #4876: Unable to play video attached to form on Android 11
Collect v play store/master Android v 11 device used...

#### Problem description
When testing #4860 I have noticed that playing video does not work on Android 11 device:
Screenshot (Oct 20, 2021 3_37_09 PM)

1. Open any form with video attachment/ 1st question of attached form
2. Try to play video

#### Expected behavior
User should be able to play video

#### Other information 
- example of form: https://github.com/getodk/collect/files/7232052/Example.Problem.zip
- works well on v2021.2 beta 0

grzesiek2010:
Closes #4876 

I have tested staring some external apps.

I had to reimplement the way we check if an activity we want to star is available. Before we used our [ActivityAvailability](https://github.com/getodk/collect/pull/4878/files#diff-214f3bcddf9d06f60193bd716bc979fa1fe2e92c273b8c7ea36da9684934dd42L8) which used `queryIntentActivities()`. According to https://developer.android.com/about/versions/11/privacy/package-visibility?authuser=2 that method will return an empty array on android 11 if we don't declare <queries> in our AndroidManifest.xml like https://stackoverflow.com/a/63246789

We could do that but that would open a lot of room for error since we can easily forget something so we agreed that a better option would be to get rid of it and just try to start activities and catch exceptions that are thrown if they are not available.

Please test:
- taking pictures
- recording videos
- recording audio
- attaching arbitrary files
- opening media files in Collect (recorded using our widgets but also media files attached to questions)
- starting external widgets
- choosing files (fore example qr codes for configuring the app)

plus please test the behavior when an external app is not available (there should be a toast in most cases).

It's difficult to list everything so basically everything that is related to staring external apps might be affected.

No.

No.


androidshared/build.gradle
```diff
@@ -55,12 +55,15 @@ dependencies {
     implementation Dependencies.android_material
     implementation Dependencies.androidx_fragment_ktx
     implementation Dependencies.timber
+    implementation Dependencies.dagger
```
seadowg: Not sure that I understand why we're adding Dagger `androidshared`. I'm guessing you added for something you ended up not needing?

androidshared/src/main/java/org/odk/collect/androidshared/utils/IntentLauncherImpl.kt
```diff
@@ -0,0 +1,59 @@
+package org.odk.collect.androidshared.utils
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import androidx.activity.result.ActivityResultLauncher
+
+object IntentLauncherImpl : IntentLauncher {
```
seadowg: Huh. I'd never thought about this pattern of having an `object` implement an interface for stateless implementations. As far as I can see it's a nice idea! Really useful for pulling out related helper functions as an interface that don't need really need a class. I guess we'd need to be careful about not adding any state to an implementation like this (as that would be static).

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.verifyNoMoreInteractions
+import org.mockito.kotlin.whenever
+import java.lang.Exception
+
+@RunWith(AndroidJUnit4::class)
```
seadowg: It looks like you won't need this as the all the Android objects are being mocked.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+class IntentLauncherImplTest {
+    lateinit var context: Context
```
seadowg: You can call `mock` at test construction you can actually just have `val` for all these fields with inline assignments, and then you don't need `setup` at all.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    fun `startActivity with given intent should be called on the context when calling IntentLauncher#launch(context, intent, onError)`() {
+        intentLauncher.launch(context, intent, onError)
+        verify(context).startActivity(intent)
+        verifyNoMoreInteractions(onError)
+    }
+
+    @Test
+    fun `onError should not be called if no exception occurs when calling IntentLauncher#launch(context, intent, onError)`() {
```
seadowg: The assertions in this test double up the one above. I think you can either keep both and remove `verifyNoMoreInteractions(onError)` in the first test or just have the one test for the "positive" case (I think I prefer the former at a "unit" level like this).

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    fun `startActivity with given intent should be called on the context when calling IntentLauncher#launchForResult(context, intent, requestCode, onError)`() {
+        intentLauncher.launchForResult(activity, intent, 1, onError)
+        verify(activity).startActivityForResult(intent, 1)
+        verifyNoMoreInteractions(onError)
+    }
+
+    @Test
+    fun `onError should not be called if no exception occurs when calling IntentLauncher#launchForResult(context, intent, requestCode, onError)`() {
```
seadowg: Same doubling up here as far as I can see!

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    fun `onError should be called if any error occurs when calling IntentLauncher#launchForResult(context, intent, requestCode, onError)`() {
+        whenever(activity.startActivityForResult(intent, 1)).then {
+            throw Error()
+        }
+        intentLauncher.launchForResult(activity, intent, 1, onError)
+        verify(onError).invoke()
+    }
+
+    @Test
+    fun `startActivity with given intent should be called on the context when calling IntentLauncher#launchForResult(resultLauncher, intent, onError)`() {
```
seadowg: Description is a little off here. Should probably mention activity result launcher instead of `startActivity`

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    fun `onError should be called if any error occurs when calling IntentLauncher#launch(context, intent, onError)`() {
+        whenever(context.startActivity(intent)).then {
+            throw Error()
+        }
+        intentLauncher.launch(context, intent, onError)
+        verify(onError).invoke()
+    }
+
+    @Test
+    fun `startActivity with given intent should be called on the context when calling IntentLauncher#launchForResult(context, intent, requestCode, onError)`() {
```
seadowg: "`startActivityForResult` with given..."

androidshared/src/main/java/org/odk/collect/androidshared/utils/IntentLauncherImpl.kt
```diff
@@ -0,0 +1,59 @@
+package org.odk.collect.androidshared.utils
```
seadowg: This could go in the `system` package I reckon? Also, I think the file should just be named after the interface in these cases where you have a default implementation as well.

testshared/src/main/java/org/odk/collect/testshared/FakeIntentLauncher.kt
```diff
@@ -0,0 +1,50 @@
+package org.odk.collect.testshared
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import androidx.activity.result.ActivityResultLauncher
+import org.odk.collect.androidshared.utils.IntentLauncher
+
+class DummyIntentLauncher : IntentLauncher {
```
seadowg: Looks like this isn't being used?

seadowg (changes requested):
This is looking really good! I still need to go through the new widget intent launchers, but I've got a few tweaks that we could make that I'll just share now so that I'm not holding you up!

seadowg (changes requested):
I think we can maybe revise the naming for the external widget intent launchers to follow the existing external data interfaces we have like `GeoDataRequester`. For instance `ExFileWidgetIntentLauncher` could be `FileRequester`. 

I also think we could simplify the interfaces a little for these by giving the implementations constructors (and making them `class` not `object`). Right now we have to pass dependencies like `IntentLauncher` into the widgets (increasing reliance on the class hierarchy here) whereas really they could be detached through these interfaces. For example, `ExStringWidgetIntentLauncher#launch` could just take a `FormEntryPrompt` and the `onError` callback (and I think the request code) while the other dependencies could be passed in at construction in `WidgetFactory`. 

Thoughts?

androidshared/build.gradle
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/test/java/org/odk/collect/androidshared/utils/IntentLauncherImplTest.kt
grzesiek2010: Fixed.

androidshared/src/main/java/org/odk/collect/androidshared/utils/IntentLauncherImpl.kt
grzesiek2010: Fixed.

testshared/src/main/java/org/odk/collect/testshared/FakeIntentLauncher.kt
grzesiek2010: Fixed.

grzesiek2010:
> I think we can maybe revise the naming for the external widget intent launchers to follow the existing external data interfaces we have like GeoDataRequester. For instance ExFileWidgetIntentLauncher could be FileRequester.

> I also think we could simplify the interfaces a little for these by giving the implementations constructors (and making them class not object). Right now we have to pass dependencies like IntentLauncher into the widgets (increasing reliance on the class hierarchy here) whereas really they could be detached through these interfaces. For example, ExStringWidgetIntentLauncher#launch could just take a FormEntryPrompt and the onError callback (and I think the request code) while the other dependencies could be passed in at construction in WidgetFactory.

I agree with that and it's now implemented.

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -106,6 +107,9 @@
     @Inject
     SettingsProvider settingsProvider;
 
+    @Inject
+    IntentLauncher intentLauncher;
```
seadowg: Don't think this is being used anywhere - can just be deleted. As far as I can see, we're passing an `IntentLauncher` in at the constructor everywhere we need it.

seadowg (approved):
One tweak but this is great!

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
grzesiek2010: Right, fixed.

mmarciniak90:
@grzesiek2010, I checked external widgets on a separate page without CollectAnswerProvider installed and I see unexpected message in toast.

Form to test: [external_widgets](https://test.getodk.cloud/#/projects/270/forms/external_widgets) or `All widgets`

|incorrect toast|expected toast|
|---|---|
|Screenshot_20211102-113758_ODK Collect|Screenshot_20211102-113835_ODK Collect|

grzesiek2010:
@mmarciniak90 The problem should be solved now.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 9.0 and 11

Tested cases:
- video widget: attach, record, play
- image widget: attach, take a picture
- selfie widgets
- audio widget
- draw widget
- signature widget
- annotate widget
- barcode widget
- file widget
- URL widget 
- use external widgets with app 
- toasts when there is no external app have been fixed
- numerical widget with video attachment
- play audio and video attachments in form 
- Collect counter app
- Collect Tester app
- edited project via imported QR code file
- added new project via imported QR code file

mmarciniak90:
Tested with success on Android versions: 8.1, 10.0

Issue #4718, opened 2021-07-22
Improve geopoint dialog


lognaturel:
The geopoint dialog currently does not make it easy for users to decide when to accept the point.

- [x] Consolidate suggestions from [forum post](https://forum.getodk.org/t/collect-improve-geopoint-dialog-so-it-better-highlights-acceptable-accuracy/34246) + [#3324](https://github.com/getodk/collect/issues/3324) and share a mockup

seadowg:
As part of this, it'd also be good to move away from using `ProgressDialog` in the geopoint dialog implementation (deprecated in #4803).

seadowg:
Keeping notes on research/design [here](https://github.com/seadowg/collect-design/blob/master/research/geopoint.md).

seadowg:
Ok some design sketches I've been working on. The most basic is using the current text based style but rejigging what we show:

<img width="306" alt="Screenshot 2021-11-03 at 14 32 45" src="https://user-images.githubusercontent.com/556280/140080299-26dcd51f-3f77-4f08-a0f8-947fb26f923b.png">

The most significant change here is that we don't show satellites or provider unless there is a problem with them - we'd show a warning about satellites if there are less than 4 and a warning about providers if it's not "fused". These warnings could include instructions on how to improve the situation. For me this feels like an improvement on its own and I think we could maybe make it a first step/PR regardless of where we go.

The next concept is using progress bars:

<img width="297" alt="Screenshot 2021-11-03 at 14 34 44" src="https://user-images.githubusercontent.com/556280/140080655-f7f9d6a5-5433-4db4-9e29-e3de92185b3b.png">

<img width="359" alt="Screenshot 2021-11-03 at 14 35 05" src="https://user-images.githubusercontent.com/556280/140080724-827448ad-960f-4691-9d1c-7b95db490e2b.png">

The idea is here is to have a progress bar that will fill when the target accuracy is reached (and will be proportionally fill from the min of the lowest in the session and some chosen arbitrary accuracy).

I've also been playing with a more "animated" version based on how maps present this suggested by @yanokwa:

<img width="300" alt="Screenshot 2021-11-03 at 14 38 04" src="https://user-images.githubusercontent.com/556280/140081275-ac2daa93-d36d-4819-bcb6-a0e241a20f94.png">

Here the outer circle gets smaller until it target accuracy is achieved. Then the circles line up and we get a nice big tick:

[...]

PR #4880, opened 2021-10-27
Add form event when `accuracyThreshold` used

seadowg:
We're looking at reworking the geopoint UI (#4718) and having a better sense of how popular `accuracyThreshold` is would be very useful.

Nothing other than double-checking the code.

Like with the `allow-mock-accuracy` analytics I've chosen to fire the event in the widget rather than the `ActivityGeoDataRequester`. This prevents us from potentially missing out events from optional questions.

No behaviour change! Just added analytics so this can go in without QA.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
+++ b/collect_app/src/main/java/org/odk/collect/android/analytics/AnalyticsEvents.java
@@ -235,4 +235,9 @@ public final class AnalyticsEvents {
      * Tracks how many forms include the `allow-mock-accuracy` parameter in `geo` questions
      */
     public static final String ALLOW_MOCK_ACCURACY = "AllowMockAccuracy";
+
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
My instinct would have been to log the value used. I'm not sure it's necessary but it seems it might give us insights into whether we should change the default, for example. 5m used to be an aggressive target. Now devices can get lower values easily because of hardware and software improvements.

I'll let you consider that but am fine with this as-is, too. Let's definitely get into the current release.

seadowg:
> I'll let you consider that but am fine with this as-is, too. Let's definitely get into the current release.

We've been writing our analytics helpers with the assumption that we only send the form or the server with the event so adding this would require a bit of rework. Let's leave it for now and come back to it if we really feel like we need it. My intention for the moment was just to get a sense of whether people are using this or not.

PR #4883, opened 2021-10-28
Updated dependencies for v2022.1

grzesiek2010:
I've run automated tests.

It doesn't require testing.

No.

No.


.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*.{kt,kts}]
+disabled_rules=import-ordering
```
grzesiek2010: I disabled checking import ordering because it has always been kind of annoying especially that in Android Studio we don't have any easy way to reorder imports so that ktlint is satisfied. And now with new updates they changed something and I was asked to change the order of menu imports despite the fact everything was fine before.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormMetadataTest.kt
```diff
@@ -25,12 +25,10 @@ class FormMetadataTest {
     fun startGeopoint_withCachedFormDefinition_doesNotCauseError() {
         rule.startAtMainMenu()
             .copyForm("start-geopoint.xml")
-
```
grzesiek2010: They just decided we shouldn't use blank lines like that https://github.com/pinterest/ktlint/issues/1248
I could disable the rule but it would also mean disabling checking blank lines in all the other cases (that I like).

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -97,7 +97,7 @@ object Dependencies {
     const val hamcrest = "org.hamcrest:hamcrest:2.2"
     const val powermock_module_junit4 = "org.powermock:powermock-module-junit4:2.0.9"
     const val powermock_api_mockito2 = "org.powermock:powermock-api-mockito2:2.0.9"
-    const val robolectric = "org.robolectric:robolectric:4.6.1"
```
grzesiek2010: I had to downgrade robolectric for now because of this issue https://github.com/robolectric/robolectric/issues/6593 there is a workaround there but I don't like it so I prefer waiting for a better version of this dependency.

build.gradle
```diff
@@ -15,7 +15,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:7.0.3'
-        classpath 'com.google.gms:google-services:4.3.10'
```
grzesiek2010: The updated broke lint throwing hundreds `Unused resource` errors so I downgraded for now.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormMetadataTest.kt
seadowg: I feel pretty strongly about keeping those blank lines. In a lot of case it might not make sense, but I do think it helps readability in our UI tests when related groups are broken up by a blank line like that. Can we put ignore the rule and revert related changes?

.editorconfig
seadowg: Is there no way to fix using "Import layout" in Android Studio > Preferences > Editor > Code Style > Kotlin? It would be nice to keep ordering enforced, so we don't end up causing accidental changes in files using "optimize imports".

build.gradle
seadowg: Let's pull out a separate issue for this.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormMetadataTest.kt
grzesiek2010: No because the rule is `no-consecutive-blank-lines` so as I said if you disable it you disable checking all the other cases where having blank lines is obviously a bad thing.

build.gradle
grzesiek2010: I usually don't file new issues for every dependency that I need to skip during upgrading. Usually if there is something wrong next time it works fine. If during next code freeze the same issue occurs then I will do that.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormMetadataTest.kt
seadowg: Sorry what I mean is that I'd be happy to just disable that rule. I feel like the blank lines are useful enough in some places to make up for other places they might appear and not look quite right.

grzesiek2010: Are you sure? I'm going to repeat: if we do that you can add multiple blank lines in many different ways in the codebase and it won't be detected. I didn't want to do that because in all the other cases detecting blank lines worked well.

.editorconfig
grzesiek2010: Ok I downgraded ktlint because they might have something broken regarding the ordering of imports in the newest version if what we have on master is no longer accepted. I even tried different settings like:
```
ij_kotlin_imports_layout=* # alphabetical with capital letters before lower case letters (e.g. Z before a), no blank lines
ij_kotlin_imports_layout=*,java.**,javax.**,kotlin.**,^ # default IntelliJ IDEA style, same as alphabetical, but with "java", "javax", "kotlin" and alias imports in the end of the imports list
```
to no avail so let's check it after next release.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/FormMetadataTest.kt
seadowg: Ah, you mean it's not just for chains? Sorry you're right I'm misunderstanding and that it's not worth disabling.

grzesiek2010: > it's not just for chains?

yeah it's a broader rule.

You know what? now as I downgraded ktlint because of that problem with imports (see above) I can bring those lines back too. So maybe let's do that for now and complain about that in their repo (and ask if we can disable this particular check somehow if they are unwavering). What do you think?

seadowg: Yeah, agreed! It does seem a little broken as a line followed by a comment shouldn't count as "consecutive" blank lines. I can post a reply (seeing as I seem to care so much 😆).

grzesiek2010: Ok I reverted that commit.

seadowg (approved):
Good to merge once we unfreeze!

grzesiek2010:
@seadowg 

I've added (or actually removed :)) some more stuff:
- `ButterKnife` since we should use view binding instead
- `findbugs` since it hasn't been used for a long time either way and I think we don't need to bring it back. It checks only java code but now most of the new code is in Kotlin.

seadowg (approved):

PR #4887, opened 2021-10-29
Fixed comparing media files in order to detect updates

### LINKED ISSUE #4882: User sees a message about update after successfully auto-update when `Previously downloaded forms only` option is set
Collect v2021.3 beta 2, v2021.2.4

#### Problem description
User sees a message about update after successfully auto-update when `Previously downloaded forms only` option is set

1. Central/Aggregate server is set
2. `Previously downloaded forms only` and `Automatic download` is set
3. User has downloaded form with media - e.g. All widgets
4. New version of this form with new media is added to the server
5. User sees a notification that form was downloaded successfully, user can fill new version
6. User opens `Get Blank Forms` and sees message suggesting that there is a new version

#### Expected behavior
Message in `Get Blank Forms` should not be visible

grzesiek2010:
Closes #4882

I have reproduced the issue and confirmed that this fix works well. I Have also added automated tests.

The problem was that if we had multiple forms with the same id but different versions during comparing media files we would get media files from the first one from the list which in our case usually would be the oldest one and not the one with the version and id we are looking for https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4882?expand=1#diff-d241c237e50a73d5233d52f1d174193cc3692fe2aa977b654ac5e02ff39f13d7L83

Please just verify that the issue no longer exists and the message about outdated forms is really displayed when it should be. Nothing else should be affected.

Any form with media files.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -31,6 +31,7 @@ import org.odk.collect.shared.strings.Md5;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import javax.annotation.Nullable;
 
@@ -163,10 +164,7 @@ public class ServerFormsDetailsFetcher {
 
     private Form getFormByVersion(List<Form> forms, @Nullable String expectedVersion) {
         for (Form form : forms) {
-            if (expectedVersion == null && form.getVersion() == null) {
-                return form;
-            }
-            if (expectedVersion != null && form.getVersion() != null && expectedVersion.equals(form.getVersion())) {
+            if (Objects.equals(form.getVersion(), expectedVersion)) {
                 return form;
             }
         }
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -32,6 +32,8 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.annotation.Nullable;
+
 import timber.log.Timber;
 
 public class ServerFormsDetailsFetcher {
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
```diff
@@ -158,4 +160,16 @@ private static boolean isMediaFileAlreadyDownloaded(List<File> localMediaFiles,
     private String getMd5HashWithoutPrefix(String hash) {
         return hash == null || hash.isEmpty() ? null : hash.substring("md5:".length());
     }
+
+    private Form getFormByVersion(List<Form> forms, @Nullable String expectedVersion) {
+        for (Form form : forms) {
+            if (expectedVersion == null && form.getVersion() == null) {
```
lognaturel: You could use `Objects.equals(form.getVersion(), expectedVersion)`, I believe.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
```diff
@@ -131,13 +131,48 @@ public void whenAFormExists_andHasUpdatedMediaFileOnServer_isUpdated() throws Ex
...
+
+        formsRepository.save(new Form.Builder()
+                .formId("form-2")
+                .version("2")
+                .md5Hash("form-2-hash")
+                .formFilePath(FormUtils.createXFormFile("form-2", "2").getAbsolutePath())
+                .formMediaPath(mediaDir2.getAbsolutePath())
+                .build());
+
+        File oldMediaFile2 = TempFiles.createTempFile(mediaDir2, "blah", ".txt");
```
seadowg: Could you just reference the `mediaFile` field here to make it really explicit we're using the same file?

seadowg (changes requested):
One small tweak and I think @lognaturel's suggestion is good as well.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
grzesiek2010: Right, fixed.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
grzesiek2010: Ok I improved it a bit.

seadowg (approved):

mmarciniak90:
**Tested with success**

Verified on Android 8.1 and 10.0

Verified cases:
* Central server with `Previously downloaded forms only` and `Automatic download`
* Aggregate server with `Previously downloaded forms only` and `Automatic download`
* auto-download new version with new media
* auto-download new version of form without media
* Central server and match exactly
* Aggregate server and match exactly

kkrawczyk123:
Verified also on Androids: 5.1, 9.0 and 11.

PR #4891, opened 2021-11-01
Remove mention of admin settings

lognaturel:
The project config tip text is very visible so I think it's worth changing. I had hoped to change it at the beginning of the release cycle but forgot. I can go through translations before I download them and update them where I can. I think it will be a clear change in most languages.

The other string is not common so I think it's fine to change without translations. At the beginning of the next release cycle, we should replace the few strings that still use "General settings" (#4892).


seadowg (approved):
Great catch! Agreed about changing the general ones as part of the net cycle.

PR #4894, opened 2021-11-04
Update translations for v2021.3

lognaturel:
Routine translation update

Verified that only language files have changed


yanokwa (approved):
Made sure only XML files are being changed and no file had a crazy number of additions or deletions.

seadowg (approved):

PR #4897, opened 2021-11-05
Fixes #4480 Text added in angle brackets …

### LINKED ISSUE #4480: Text added in < > is not visible in the hierarchy view
Collect 1.30 (older versions were not verified)

#### Problem description
Text added in < > is not visible in the hierarchy view

1. User starts form with String widget, e.g. `All widgets` form
2. User adds value in < >
3. User opens the hierarchy view -> **added value is not visible**
4. Correct text is visible on servers: Central, Aggregate, Google Sheet

ezgif com-video-to-gif

#### Expected behavior
The whole value is visible on the hierarchy list

dimwight:
Resolves #4480

Tested manually using [text4480.xlsx](https://github.com/getodk/collect/files/7486160/text4480.xlsx)

As described in [the docs](https://docs.getodk.org/form-styling/?highlight=markdown#inline-html), Collect supports inline HTML in label text. 

So in `HierarchyListAdapter`, `onBindViewHolder` currently treats _any_ `secondaryText` of `HierarchyElement` as HTML, including what in fact is answer text. 

Thus in the test form answers, text within <> 
- if arbitrary as in _test1_ is discarded 
- if defining valid markup as in _test2_ is rendered as such

The fix is easy - when the `Type` of the element is `QUESTION`, render `secondaryText` as is. 

Bug as described in issue has been removed. Only possible regression would be in tests that depend on rendering of answers. 

Developed and tested with [text4480.xlsx](https://github.com/getodk/collect/files/7486160/text4480.xlsx) which defines defaults that reproduce the issue. 

No?

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
@@ -16,13 +16,15 @@
 
 package org.odk.collect.android.adapters;
 
-import androidx.annotation.NonNull;
-import androidx.recyclerview.widget.RecyclerView;
+import android.graphics.drawable.Drawable;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.logic.HierarchyElement;
@@ -48,17 +50,24 @@ public class HierarchyListAdapter extends RecyclerView.Adapter<HierarchyListAdap
 
     @Override
     public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
-        holder.bind(hierarchyElements.get(position), listener);
-        if (hierarchyElements.get(position).getIcon() != null) {
+        HierarchyElement element = hierarchyElements.get(position);
+        holder.bind(element, listener);
+        Drawable icon = element.getIcon();
+        if (icon != null) {
             holder.icon.setVisibility(View.VISIBLE);
-            holder.icon.setImageDrawable(hierarchyElements.get(position).getIcon());
+            holder.icon.setImageDrawable(icon);
         } else {
             holder.icon.setVisibility(View.GONE);
         }
-        holder.primaryText.setText(HtmlUtils.textToHtml(hierarchyElements.get(position).getPrimaryText()));
-        if (hierarchyElements.get(position).getSecondaryText() != null && !hierarchyElements.get(position).getSecondaryText().isEmpty()) {
+        holder.primaryText.setText(HtmlUtils.textToHtml(element.getPrimaryText()));
+        String secondaryText = element.getSecondaryText();
+        if (secondaryText != null && !secondaryText.isEmpty()) {
             holder.secondaryText.setVisibility(View.VISIBLE);
-            holder.secondaryText.setText(HtmlUtils.textToHtml(hierarchyElements.get(position).getSecondaryText()));
+            //For #4480
+            holder.secondaryText.setText(
+                    element.getType() == HierarchyElement.Type.QUESTION
+                            ? secondaryText
+                            : HtmlUtils.textToHtml(secondaryText));
         } else {
             holder.secondaryText.setVisibility(View.GONE);
         }
--- a/collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
+++ b/collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
@@ -16,19 +16,12 @@
 
 package org.odk.collect.android.adapters;
 
-import android.graphics.drawable.Drawable;
-import android.view.LayoutInflater;
-import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
```diff
@@ -48,17 +50,24 @@ public HierarchyListAdapter(List<HierarchyElement> listElements, OnElementClickL
...
-        holder.primaryText.setText(HtmlUtils.textToHtml(hierarchyElements.get(position).getPrimaryText()));
-        if (hierarchyElements.get(position).getSecondaryText() != null && !hierarchyElements.get(position).getSecondaryText().isEmpty()) {
+        holder.primaryText.setText(HtmlUtils.textToHtml(element.getPrimaryText()));
+        String secondaryText = element.getSecondaryText();
+        if (secondaryText != null && !secondaryText.isEmpty()) {
             holder.secondaryText.setVisibility(View.VISIBLE);
-            holder.secondaryText.setText(HtmlUtils.textToHtml(hierarchyElements.get(position).getSecondaryText()));
+            //For #4480
+            holder.secondaryText.setText(
+                    element.getType() == HierarchyElement.Type.QUESTION
```
seadowg: Nice catch! I'm wondering if this can be even simpler however: is there ever a case where we would set HTML as `secondaryText`? It looks like in all cases other than `QUESTION` it gets set to a static label for the type of element (like "Repeat" or "Group").

dimwight: Indeed, do please simplify further if you're certain it's safe.

seadowg: I've changed it so it never uses HTML rendering for the secondary text. Would be good for @grzesiek2010 or @lognaturel to review the PR now as I've touched it too much to be unbiased I reckon!

dimwight: Now such a neat fix that it scarcely needs the test?

seadowg: You could argue that. I'd want a test for even a small change like this to enforce that the behaviour we want stays there. In a big ol' code base like this it's also advantageous to have a test that documents that something is intentional and not just there because of duplication. I had to double-check we don't want HTML rendering there, but now it's explicit. You'll see that writing the test also pushed me to extract a separate view class which, in my opinion, is an improvement to the overall structure here - it keeps the adapter purely boilerplate.

I guess an interesting point is that, if I was writing this view class from scratch, I would probably have written a simpler test that just checks that the string is displayed and wouldn't have bothered with a case for HTML as I'd have assumed "no one is going to bother adding the HTML rendering if it's not needed".

grzesiek2010 (approved):

collect_app/src/main/java/org/odk/collect/android/adapters/HierarchyListAdapter.java
grzesiek2010: Looks good I added one commit too, why not :smile:  (to improve the new test).

kkrawczyk123:
@grzesiek2010 @dimwight @seadowg It looks like the issue has been fixed but I have noticed a regression issue in navigating through the hierarchy view. Initially, I was thinking about performance but after a little investigation it turned out that jumping to questions changed from tapping the whole question line (blue frame on screenshot) to only text area (red frame on screenshot):
Screenshot_2021-11-22-09-26-31-961_org odk collect android (1)
I compared behavior with store version and master where I can still use the whole line (blue area).

seadowg:
@dimwight feel free to take a look but if you don't have time, let me know, and I can investigate.

dimwight:
Looks interesting but should it be raised as a new issue?

grzesiek2010:
> Looks interesting but should it be raised as a new issue?

If it's an issue caused by this pr it can't be treated as something separate.

dimwight:
That was actually my point, isn't it a separate regression identified during testing of this PR?

grzesiek2010:
No I'm pretty sure @kkrawczyk123 would have filed a separate issue then, but we can ask for confirmation @kkrawczyk123

kkrawczyk123:
I have mentioned in my previous comment that this behavior is not visible on the current master and play store version so it looks like it has been unintentionally introduced in this pr.

dimwight:
So sorry @kkrawczyk123, I didn't read your comment carefully enough.

dimwight:
Hopefully this fixes the regression found by @kkrawczyk123.

mmarciniak90:
**Tested with success on Android 10**

Verified cases:
- scenario form issue on _All widgets_ form and _text4480_ form
- display in hierarchy different type of answers
- click on group name and group area in hierarchy
- text widget
- ex string widget
- form with last-saved
- encrypted form with last-saved
- audit 
- tracking changes reason
- light and dark mode
- screen rotation
- minimize app
- extinguish screen 
- view groups in hierarchy

srujner:
**Tested with success on Android 11**

Verified cases:

- click on group name and group area in hierarchy
- scenario form issue on All widgets form
- display in hierarchy different type of answers
- view groups in hierarchy
- text widget
- ex string widget
- form with last-saved
- encrypted form with last-saved
- tracking changes reason
- light and dark mode
- screen rotation
- minimize app
- extinguish screen

PR #4898, opened 2021-11-05
Fixed detecting form updates

### LINKED ISSUE #4895: Attempt to invoke virtual method 'java.lang.String org.odk.collect.forms.Form.getFormMediaPath()' on a null object reference
Collect v2021.3.0

#### Problem description
https://console.firebase.google.com/u/2/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/be730c9d272352bd5b8685ab0d197899?time=last-seven-days&versions=v2021.3.0%20(4316)&sessionEventKey=6184E33F02B600015883C970B2D53C78_1605436904284051748

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.odk.collect.forms.Form.getFormMediaPath()' on a null object reference
       at org.odk.collect.android.utilities.FormUtils.getMediaFiles(FormUtils.java:20)
       at org.odk.collect.android.formmanagement.ServerFormsDetailsFetcher.areNewerMediaFilesAvailable(ServerFormsDetailsFetcher.java:130)
       at org.odk.collect.android.formmanagement.ServerFormsDetailsFetcher.fetchFormDetails(ServerFormsDetailsFetcher.java:86)
       at android.os.AsyncTask$3.call(AsyncTask.java:378)
       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:289)
       at java.lang.Thread.run(Thread.java:929)
```

1. Upload a simple form with media files to Central and download it in Collect
2. Update the form bumping its version and download in Collect
3. Fill one form and save it
4. Delete the newer version of the form
5. Try to open the list of forms to download

The app doesn't crash but loads and loads and in logs I get the same NPE

If you have a form but without media files and you do the same, then there is no crash or infinitive loading
but the message that an update is available is not visible what is also a bug.

I used this form: [formTest.zip](https://github.com/getodk/collect/files/7485591/formTest.zip)

grzesiek2010:
Closes #4895

I added automated tests and verified the fix manually.

The bug was in this line https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4895?expand=1#diff-d241c237e50a73d5233d52f1d174193cc3692fe2aa977b654ac5e02ff39f13d7L126

we took all forms from database by hash but we should skip those soft deleted.

So when we had two versions of the same form and the newest one was soft deleted (deleted but still present in our database because there are instances of that form) `getOneByMd5Hash` would return that deleted form and it would be treated as a form we already have.

This should just fix the issue and have no side effects so please focus on the reproduction steps. If you have some other ideas (and time) it would be of course good to play with form updates to make sure we don't have more hidden bugs. 

Any form with media files like the one I attached to the issue.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -123,7 +123,8 @@ public class ServerFormsDetailsFetcher {
         }
 
         String hash = getMd5HashWithoutPrefix(formListItem.getHashWithPrefix());
-        return formsRepository.getOneByMd5Hash(hash) == null;
+        Form form = formsRepository.getOneByMd5Hash(hash);
+        return form == null || form.isDeleted();
     }
 
     private boolean areNewerMediaFilesAvailable(Form existingForm, List<MediaFile> newMediaFiles) {
--- a/collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
+++ b/collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
@@ -144,8 +144,8 @@ public class ServerFormsDetailsFetcherTest {
                 .formMediaPath(mediaDir1.getAbsolutePath())
                 .build());
 
-        File oldMediaFile1 = TempFiles.createTempFile(mediaDir1, mediaFile.getFilename());
-        writeToFile(oldMediaFile1, "old blah");
+        File mediaFile1 = TempFiles.createTempFile(mediaDir1, mediaFile.getFilename());
+        writeToFile(mediaFile1, "old blah");
 
         File mediaDir2 = TempFiles.createTempDir();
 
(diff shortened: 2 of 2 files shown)


forms/src/main/java/org/odk/collect/forms/FormsRepository.java
```diff
@@ -19,6 +19,9 @@
     @Nullable
     Form getOneByMd5Hash(@NotNull String hash);
 
+    @Nullable
+    Form getOneNotDeletedByMd5Hash(@NotNull String hash);
```
seadowg: I think we could just check the `Form#isDeleted` on the returned value of `getOneByMd5Hash` instead of introducing this function. That would mean we aren't increasing the surface area of the `FormsRepository` (and the number of tests required). What do you think?

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.kt
```diff
@@ -0,0 +1,64 @@
+package org.odk.collect.android.utilities
+
+import org.javarosa.core.reference.ReferenceManager
+import org.javarosa.core.reference.RootTranslator
+import org.odk.collect.android.logic.FileReferenceFactory
+import org.odk.collect.android.storage.StoragePathProvider
+import org.odk.collect.forms.Form
+import java.io.File
+
+object FormUtils {
```
seadowg: Is there a reason we needed to do this in this PR? Maybe we can pull it out of the branch if not just so there's no extra risk added. If there's an advantage for the fix I'm missing I'm happy for it to stay in though!

seadowg (changes requested):

forms/src/main/java/org/odk/collect/forms/FormsRepository.java
grzesiek2010: Yeah we can do that like that too.

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.kt
grzesiek2010: No it's not crucial so I removed that change.

seadowg (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 9 and 11

Tested cases:
- scenario from issue
- finalize/save form
- update form version for few times
- download the same form more than once
- automatic form update, previously downloaded form
- match exactly server 
- Central, Aggregate 
- form update regression checks
- try with different forms with media 
- do not hide all form versions
- fill different versions of forms 
- send different versions to Central

mmarciniak90:
Tested with success on Android 8.1 and 10.0.

Issue #4896, opened 2021-11-05
Prevent swipes before form has loaded properly


seadowg:
It seems like from this [crash](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/30a3178c00619de299e65f60078556f8?time=last-seven-days&sessionEventKey=6184F4B302540001171121663967053E_1605465151459213822) that it's possible to swipe before the form has loaded (and there's no view on screen). This causes a `NullPointerException` as `currentView`  (in `FormEntryActivity`) hasn't been assigned yet.

It's hard to reproduce this, but it should be easy enough to protect against, and it does seem like a nasty instability in our swipe code.

## Notes

I think fixing this is a nice opportunity to encapsulate our swipe code (scattered currently between `SwipeHandler` and various methods in `FormEntryActivity`) into a dedicated view, separating swiping from the actual view creation and form index manipulation entirely. In that world it'd be easy to just contain `currentView` in there and ignore swipes or next/back clicks if the view hasn't been populated. In my mind, a possibility would be something like this:

```kotlin
class SwipeView {

    // For refreshing, setting first view
    fun setView(view: View)

    // For explicit moves forward and backwards like when prompting for a repeat.
    // This might get rid of the need for `FormIndexAnimationHandler`
    fun setView(view: View, direction: Direction)

    // Called by nav bar buttons. These would then trigger the appropriate listener method
    fun next()
    fun back()
   
    // The `SwipeView` can use the listener to "ask" for the next/previous view. 
    // `FormEntryActivity` could implement the methods and carry out the form index
    // manipulation, create the view and then return it in each.
    fun setListener(listener: Listener) 

    interface Listener {

        fun onSwipeForward(): View

        fun onSwipeBackward(): View
    }
}
```

We've chatted about moving to a `ViewPager` before, but that process seems to be so hard at this point that I think doubling down on a custom swipe system but extracting it as a reusable component (probably in its own module) would at least give us some separation and could untangle some of `FormEntryActivity`.

PR #4901, opened 2021-11-08
Guard against NPE when cancelling input events

seadowg:
Related to #4896

This should prevent the crashes we're seeing, but leaves in what we **think** is the root cause - swiping is possible when there is no "current" view in form entry.

No way to reproduce yet so just ran tests.

It would be nice to go with something like what's described in #4896, but I think we should prevent the crash people are seeing now and then follow up with a more general fix.

Should hopefully just fix the crash. Good to give form entry a quick whirl to check, especially form loading and trying to swipe while the form is loading.


grzesiek2010 (approved):

grzesiek2010:
@mmarciniak90 @kkrawczyk123 I merged this pr because I think reproduction would be very difficult but when you have some time you can take a look at it. Maybe you will be able to do that.

PR #4903, opened 2021-11-08
Run UI tests against API 29

### LINKED ISSUE #4874: Use API 30 for instrumentation tests
We should be using our target API with Firebase, so we catch problems on the newest (and often strictest) OS.

## Notes

I did a quick recce on this and found there are a bunch of failures (in the 20s) after moving to an API 30 device. It looks like they are all related to these problems:

- [x] Toast assertions don't work (https://github.com/android/android-test/issues/803)
- [x]  Middle toasts show as dialogs and need to be closed before moving on (`FormValidationTest`)
- [x]  Not permitted to create temp files (`IntentGroupTest`)
- [x] UIAutomator notification assertions are potentially flakey?

seadowg:
Work towards #4874

I ended up running into some hard to explain flakiness running against API 30 and decided to step back to 29 to not have all my time sucked into this. Definitely more work to do here but this is a step in the right direction so worth merging in.

This has been pretty demoralizing as it seems like API 29 and 30 introduce bugs into Espresso, and it doesn't seem like the Android testing team is doing much about it. Hopefully we can keep building workarounds, but it's worrying to see the stability of the framework itself start to decrease. 

Ran test lab a bunch of times.

Comments inline. A lot of the work I needed to do is described in the issue.


collect_app/src/androidTest/java/org/odk/collect/android/support/DisableDeviceAnimationsRule.kt
```diff
@@ -0,0 +1,29 @@
+package org.odk.collect.android.support
+
+import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import androidx.test.uiautomator.UiDevice
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class DisableDeviceAnimationsRule : TestRule {
```
seadowg, line 9: I'm **fairly** sure that animations are disabled on Firebase, but I saw some failures that were hard to explain otherwise. This also means we don't need to remember about switching animations off locally, so I think it's a net positive.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.java
```diff
@@ -29,11 +33,18 @@ public Statement apply(Statement base, Description description) {
             @Override
             public void evaluate() throws Throwable {
                 ActivityScenario.launch(getActivityIntent());
+                formEntryPage = new FormEntryPage(formName);
+                formEntryPage.assertOnPage();
```
seadowg, line 37: Make sure the form has loaded, and we're in the right place before the tests start making assertions, performing actions.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.java
```diff
@@ -153,14 +153,11 @@ public T assertTextDoesNotExist(String... text) {
     }
 
     public T checkIsToastWithMessageDisplayed(String message) {
-        try {
-            onView(withText(message)).inRoot(new ToastMatcher()).check(matches(isDisplayed()));
-        } catch (RuntimeException e) {
-            // The exception we get out of this is really misleading so cleaning it up here
+        Espresso.onIdle();
```
seadowg, line 156: We need to make sure we wait for any background work to finish before checking the toasts.

androidshared/src/main/java/org/odk/collect/androidshared/ui/ToastUtils.kt
```diff
@@ -17,6 +17,10 @@ import org.odk.collect.strings.localization.getLocalizedString
  */
 object ToastUtils {
 
+    @JvmStatic
+    var recordToasts = false
```
seadowg, line 21: I looked at pulling out an interface that we can inject a recording fake of instead of using static state like this, but it felt complicated by the middle toast functionality. I think we can revise this when we look at reworking constraints for Android 11.

lognaturel (approved):
Annoying stuff to deal with! I don't see anything controversial or anything @grzesiek2010 particularly needs to take notice of so I'll go ahead and merge to avoid conflicts piling up.

Issue #4888, opened 2021-10-30
Two different forms with the same ID and version are on device


yanokwa:
#### Software and hardware versions 
Collect v2021.2.4, Central v1.3.x

#### Problem description
> An error has occurred You have downloaded two different forms with the same ID and version. You may have loaded them at different times or from different sources. You must delete one.

#### Steps to reproduce the problem
Unclear

#### Expected behavior
It should be impossible to have different forms with the same ID.

#### Other information 
Apparently, already deleting and and re-downloaded Collect didn't fix it. Here's what fixed it...
Main screen > Project icon > Settings > Project management > Reset > Blank forms.

lognaturel:
This can happen if using the same Collect project for multiple drafts or a draft and published version. Was the person who experienced this involved with any drafts?

Are you absolutely sure they were using the latest Collect version the whole time? That they hadn’t connected to another server or pushed a form manually?

yanokwa:
They were not using any server side drafts. They deleted and re-downloaded the app about a week ago, but had only downloaded it for the first time at most two weeks before that. They did not push a form manually.

yanokwa:
This has happened again with a customer who is on v2021.2.4 (maybe recently upgraded?) and I'm pretty sure hasn't had drafts for months. I know that isn't the best repro, but I wanted to report it anyway.

lognaturel:
@seadowg @grzesiek2010 I want to make sure you've seen that impossible things are happening. This is starting to feel more serious but I have no idea where to even start troubleshooting. Maybe it's a good opportunity to think about what I've previously mentioned and use the form hash to link instances to form defs? That does feel like a pretty big change, though.

seadowg:
> They did not push a form manually.

I'd missed this update. My assumption would have been that that's how they'd got into this situation. Could we get details on how their form management is set up? Are they using Match Exactly/Previously Downloaded etc?

yanokwa:
Everyone is on the latest version of Central with managed QR codes.

seadowg:
Great that at least eliminates a few other async things. I've pretty much forgotten why this **should be** impossible, so it'd be good to go through again with a fresh mind. I'll see if I can come up with any theories and if not I'll pass it over to @grzesiek2010 (unless he immediately jumps in with exactly why it's happening).

grzesiek2010:
@seadowg do you want me to take a look at it?

seadowg:
@grzesiek2010 that'd be great I haven't had a chance yet.

[...]

PR #4906, opened 2021-11-10
Increase form update resilience and track analytics for badly behaved servers that don't provide form hashes

### LINKED ISSUE #4895: Attempt to invoke virtual method 'java.lang.String org.odk.collect.forms.Form.getFormMediaPath()' on a null object reference

grzesiek2010:
Closes #4895

I added automated tests and reviewed implemented changes.

The problem seems to be on a server side so this just makes the code more resistant to wrong data provided by servers.

This does not require testing since reproducing the bug is difficult.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -16,6 +16,8 @@
 
 package org.odk.collect.android.formmanagement;
 
+import org.odk.collect.analytics.Analytics;
+import org.odk.collect.android.analytics.AnalyticsEvents;
 import org.odk.collect.android.openrosa.OpenRosaFormSource;
 import org.odk.collect.forms.Form;
 import org.odk.collect.forms.FormListItem;
@@ -75,7 +77,7 @@ public class ServerFormsDetailsFetcher {
             boolean isNewerFormVersionAvailable = false;
 
             if (!isHashValid(listItem.getHashWithPrefix())) {
-                // TODO maybe log this in analytics?
+                Analytics.log(AnalyticsEvents.NULL_OR_EMPTY_FORM_HASH);
             } else if (thisFormAlreadyDownloaded) {
                 Form form = getFormByHash(listItem.getHashWithPrefix());
                 if (form == null || form.isDeleted()) {
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -77,13 +77,14 @@ public class ServerFormsDetailsFetcher {
             if (!isHashValid(listItem.getHashWithPrefix())) {
                 // TODO maybe log this in analytics?
             } else if (thisFormAlreadyDownloaded) {
-                if (isNewerFormVersionAvailable(listItem.getHashWithPrefix())) {
+                Form form = getFormByHash(listItem.getHashWithPrefix());
+                if (form == null || form.isDeleted()) {
                     isNewerFormVersionAvailable = true;
                 } else if (manifestFile != null) {
                     List<MediaFile> newMediaFiles = manifestFile.getMediaFiles();
 
                     if (newMediaFiles != null && !newMediaFiles.isEmpty()) {
-                        isNewerFormVersionAvailable = areNewerMediaFilesAvailable(getFormByHash(listItem.getHashWithPrefix()), newMediaFiles);
+                        isNewerFormVersionAvailable = areNewerMediaFilesAvailable(form, newMediaFiles);
                     }
                 }
             }
@@ -115,11 +116,6 @@ public class ServerFormsDetailsFetcher {
             Timber.w(formSourceException);
             return null;
         }
-    }
-
-    private boolean isNewerFormVersionAvailable(String hashWithPrefix) {
-        Form form = getFormByHash(hashWithPrefix);
-        return form == null || form.isDeleted();
     }
 
     private boolean areNewerMediaFilesAvailable(Form existingForm, List<MediaFile> newMediaFiles) {
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
@@ -83,7 +83,7 @@ public class ServerFormsDetailsFetcher {
                 } else if (manifestFile != null) {
                     List<MediaFile> newMediaFiles = manifestFile.getMediaFiles();
 
-                    if (newMediaFiles != null && !newMediaFiles.isEmpty()) {
+                    if (!newMediaFiles.isEmpty()) {
                         isNewerFormVersionAvailable = areNewerMediaFilesAvailable(form, newMediaFiles);
                     }
                 }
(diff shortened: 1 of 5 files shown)


grzesiek2010:
@seadowg I'm creating this draft because as you know I will be off till Monday so you might need to take over if I'm not able to finish today.

I was able to come up with another scenario for this crash which is:
1. You have a form with media files downloaded
2. There is an updated version of the form on the server (updated form version not only media files)
3. When you try do download form details and check if it's a newer version its hash is null or empty. 

The first commit is a test that covers that scenario.

Maybe there are even more scenarios... I need to spend more time on that code.

grzesiek2010:
> its hash is null or empty

generally any wrong hash provided by a server might cause the same issue so: null, empty or just wrong

If my theory from https://github.com/getodk/collect/issues/4888#issuecomment-964070762 is right it might be related...

seadowg:
> I was able to come up with another scenario for this crash

Right. We basically don't trust the version identifier on the server, so we compare the hashes and then the media file hashes. The new code we added for the patch (grabbing the correct form to compare media files) relies on the version though. Maybe we should be using the hash there instead as well? 

Either way I guess we just need to be careful we're not passing around a `null` form. I'd be happy with a Kotlin conversion if you fancy it? Should make catching these crashes easier - I think our current code will potentially fail to compile in Kotlin.

grzesiek2010:
> The new code we added for the patch (grabbing the correct form to compare media files) relies on the version though. Maybe we should be using the hash there instead as well?

I agree that it would be better to relay on one parameter to reduce the risk and I implemented it. Generally yeah there are two options either server provides wrong hashes or form version. The second option is no longer possible because now I use hashes everywhere. For the case with invalid hashes we can use analytics. https://github.com/getodk/collect/pull/4906/files#diff-d241c237e50a73d5233d52f1d174193cc3692fe2aa977b654ac5e02ff39f13d7R78 
What do you think?

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
```diff
@@ -159,16 +146,12 @@ private static boolean isMediaFileAlreadyDownloaded(List<File> localMediaFiles,
...
 
-    private Form getFormByVersion(List<Form> forms, @Nullable String expectedVersion) {
-        for (Form form : forms) {
-            if (Objects.equals(form.getVersion(), expectedVersion)) {
-                return form;
-            }
-        }
-        return null;
+    private boolean isHashValid(String hash) {
+        return hash != null && hash.startsWith("md5:");
```
seadowg, line 157: I think we need a test for the invalid (doesn't start with `md5:` case)

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
```diff
@@ -167,6 +172,22 @@ public void whenAFormExists_andItsNewerVersionWithUpdatedMediaFilesHasBeenAlread
         assertThat(form.isNotOnDevice(), is(false));
     }
 
+    @Test
+    public void whenAFormExists_andItsNewerVersionIsAvailableButHasNullHash_isNotNewOrUpdated() throws Exception {
```
seadowg, line 193: I'm just trying to think through this behaviour and if there are any problems in it being the case. Before this situation would just crash right?

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
```diff
@@ -16,6 +18,7 @@ private FormUtils() {
 
     }
 
+    @NonNull
     public static List<File> getMediaFiles(Form form) {
```
seadowg: We should probably add a `@NonNull` to the param as well.

seadowg:
> The second option is no longer possible because now I use hashes everywhere. For the case with invalid hashes we can use analytics. https://github.com/getodk/collect/pull/4906/files#diff-d241c237e50a73d5233d52f1d174193cc3692fe2aa977b654ac5e02ff39f13d7R78
What do you think?

Yeah as far I can see the behaviour is the same for `null` and invalid hashes after this change - they just show as not being updated. The only difference is that we don't go on to crash during the media file checks. I think analytics here would be good as from the number of crashes, it seems like there are a few people with servers/workflows that cause this.

collect_app/src/test/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcherTest.java
grzesiek2010: Yes this test was failing with the same stacktrace like we can see in firebase reports.

collect_app/src/main/java/org/odk/collect/android/utilities/FormUtils.java
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.java
grzesiek2010: Like this https://github.com/getodk/collect/pull/4906/commits/ec965920fe9b15e78dee2867513851ec17f87d96?

seadowg (approved):
I think this is looking good! I did a quick Kotlin conversion to check there were no obvious nullable calls left over, and it looks fine. I think it'd be good to do some Kotlin conversion and take a look at restructuring the tests a little (now there's more of them), but that's a follow-up I can take care of.

Before merging:
* @kkrawczyk123 if you have a chance before the end of the day could you look at doing a quick pass on this (form downloading, checking for updates etc). As @grzesiek2010 says, there's not much to reproduce without a broken server, but I'd still feel better having this put through its paces with a working server given how many edge cases keep coming up. I'd say a PR like this only needs testing on one Android version.
* @grzesiek2010 are you able to quickly rework the history so we have a rebase rather than a merge commit? If you don't have time I can take care of that.

grzesiek2010:
> @grzesiek2010 are you able to quickly rework the history so we have a rebase rather than a merge commit? If you don't have time I can take care of that.

Done.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 9.0.

Tested cases:
- form update: Aggregate, Central, Google Drive
- check for form update periodically
- automatic download
- forms with version and media
- match exactly 
- fill different form versions
- send different form versions to Central
- tried scenario from issue

PR #4911, opened 2021-11-11
Convert `ServerFormsDetailsFetcher` to Kotlin

seadowg:
~~Blocked by #4906~~

We've seen a couple of recent crashes that we'd probably have caught with Kotlin's null safety/strictness, so I wanted to make sure we got this code out of Java. I also restructured the tests a little as they've grown in size and were feeling a little awkward.

Ran tests.

We've done a lot of QA in this area recently, and it seems like our test coverage is pretty well lined up with our regression checks as we've mostly been catching things in production once they hit users using custom servers. I think it'd be good for the reviewer to think of cases we might not have covered. This would be a good opportunity to add tests at the unit level or in `BadServerTest`.


forms/src/main/java/org/odk/collect/forms/FormListItem.kt
```diff
@@ -0,0 +1,10 @@
+package org.odk.collect.forms
+
+data class FormListItem(
+    val downloadURL: String,
+    val formID: String,
+    val version: String,
+    val hashWithPrefix: String?,
```
seadowg: It'd be good to think about if any of the other fields (here and in the other new data classes) should `String?` here. If so, we missing test coverage of whatever case that would be from.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.kt
```diff
@@ -0,0 +1,146 @@
...
+                    getManifestFile(formSource, it)
+                } else {
+                    null
+                }
+            }
+
+            val forms = formsRepository.getAllNotDeletedByFormId(listItem.formID)
+            val thisFormAlreadyDownloaded = forms.isNotEmpty()
+            val isNewerFormVersionAvailable = listItem.hashWithPrefix.let {
+                if (it == null || !it.startsWith("md5:")) {
```
grzesiek2010: I would move checking for that prefix `md5:` to [OpenRosaResponseParserImpl](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.java#L106) so that we deal with it only there. If that prefix does not exist we should return null like in case of empty value. What do you think?

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.kt
```diff
@@ -0,0 +1,146 @@
...
+        (formSource as OpenRosaFormSource).updateWebCredentialsUtils(webCredentialsUtils)
+    }
+
+    @Throws(FormSourceException::class)
+    open fun fetchFormDetails(): List<ServerFormDetails> {
+        diskFormsSynchronizer.synchronize()
+
+        val formList = formSource.fetchFormList()
+        return formList.map { listItem ->
+            val manifestFile = listItem.manifestURL.let {
```
grzesiek2010: I think you can simplify this:
```kotlin
val manifestFile = listItem.manifestURL?.let {
    getManifestFile(formSource, it)
}
```

seadowg: Yes! I remember writing this and being like "I swear there's a cleaner way to do this I'm forgetting". `?.let` is a great trick - basically `map` from a `Maybe` monad.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.kt
seadowg: That's an interesting idea. I'll have a play around with it.

grzesiek2010: I would even go the extra mile and remove that prefix there so that in `FormListItem` we have it without `md5:` and we don't need to call `substring` every time we use the hash.

grzesiek2010: ...or pass it to `FormListItem` with the prefix but have a custom getter which would ignore it.

seadowg: Yeah I was also thinking about a new type (like a `Hash`) that enforces the prefix but means we don't have to deal with it everywhere. I'll play around with a few of these ideas, see what I like, and then we can review!

grzesiek2010: If you decide to implement changes in `OpenRosaResponseParserImpl` can convert it to Kotlin as well,

collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
```diff
@@ -42,4 +46,21 @@ public void whenHashNotIncludedInFormList_showError() {
...
     }
+
+    @Test
+    /*
+     A server that doesn't return hashes correctly for media files would fool Collect into thinking
+     there was a new one each time. This happens because we would compare the existing file's
+     (computed) hash with the server one assuming there is a prefix and end up comparing against
+     a substring.
+    */
+    public void whenMediaFileHasMissingPrefix_showsAsUpdated() {
```
seadowg, line 57: This is probably not the behaviour we want, but it is the behaviour we have. I wrote this test against `master` to double-check. I'll pull out another issue, so we can discuss how to improve this.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormsDetailsFetcher.kt
seadowg: Ok done!

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
...
+                    when (child.name) {
+                        "filename" -> {
+                            filename = XFormParser.getXMLText(child, true)
+                            if (filename != null && filename.length == 0) {
+                                filename = null
+                            }
+                        }
+                        "hash" -> {
+                            hash = XFormParser.getXMLText(child, true)
+                            if (hash != null && hash.length == 0) {
```
grzesiek2010: Don't you need to check if it starts with `md5:` here too?

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
...
+                    val child = mediaFileElement.getElement(j)
+                    if (!isXformsManifestNamespacedElement(child)) {
+                        // someone else's extension?
+                        continue
+                    }
+
+                    when (child.name) {
+                        "filename" -> {
+                            filename = XFormParser.getXMLText(child, true)
+                            if (filename != null && filename.length == 0) {
```
grzesiek2010: You can replace all those `length == 0` with `isEmpty()`

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
...
+                        if (manifestUrl != null && manifestUrl.length == 0) {
+                            manifestUrl = null
+                        }
+                    }
+                    "hash" -> {
+                        hash = XFormParser.getXMLText(child, true)
+                        if (hash != null && (hash.length == 0 || !hash.startsWith("md5:"))) {
+                            hash = null
+                        } else {
+                            hash = hash.substring("md5:".length)
```
grzesiek2010: Since we use this in a few places maybe it's better to have a const for that `md5:`

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
grzesiek2010: Ah you want to improve it later https://github.com/getodk/collect/pull/4911/files#diff-97ea04c120d7632ab55eb3218d1ec61bf0c45056bc43f9ca6c6573feac4250cbR57?

collect_app/src/test/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImplTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.junit.Test
+import org.kxml2.io.KXmlParser
+import org.kxml2.kdom.Document
+import org.xmlpull.v1.XmlPullParser
+import java.io.StringReader
+import java.lang.StringBuilder
+
+class OpenRosaResponseParserImplTest {
```
grzesiek2010, line 12: I think we should have more test here (it might be a separate issue). It looks like we haven't had `OpenRosaResponseParserImpl` covered with tests and you just added one for your case.

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
+package org.odk.collect.android.openrosa
```
grzesiek2010, line 1: What do you think about moving it to its own module?

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
...
+                }
+            }
+
+            if (formId == null || downloadUrl == null || formName == null) {
+                formList.clear()
+                return null
+            }
+
+            val formListItem =
+                FormListItem(downloadUrl, formId, version!!, hash, formName, manifestUrl)
```
grzesiek2010: `version` might be null but it's not nullable in FormListItem.

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
```diff
@@ -0,0 +1,220 @@
...
+                    }
+                }
+            }
+
+            if (formId == null || downloadUrl == null || formName == null) {
+                formList.clear()
+                return null
+            }
+
+            val formListItem =
```
grzesiek2010: Does that make sense to construct this object ad all? You can just add it directly to the list:
`            formList.add(FormListItem(downloadUrl, formId, version!!, hash, formName, manifestUrl))
`

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
seadowg: Yeah exactly!

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
seadowg: Yeah I'd like to also have extracted an `Md5` data type, but I think I've gone deep enough. I'll do as you suggest and pull out the constant.

collect_app/src/test/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImplTest.kt
seadowg: Yeah I think we pulled it out for testing the integration between it and `OpenRosaFormSource` so it never got its own tests (this was all originally untested code). This would be a nice object to backfill though, so I'll create an issue as you suggest.

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
seadowg: Definitely! I've looked at this a few times before, but I'm not sure if we've got all the ties to `collect_app` severed yet. I think I'd rather leave that for another time rather than adding to this PR.

grzesiek2010: Yes yes not in this pr.

collect_app/src/test/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImplTest.kt
seadowg: https://github.com/getodk/collect/issues/4919

collect_app/src/androidTest/java/org/odk/collect/android/feature/smoke/BadServerTest.java
seadowg: https://github.com/getodk/collect/issues/4919

collect_app/src/main/java/org/odk/collect/android/openrosa/OpenRosaResponseParserImpl.kt
grzesiek2010: I'm not sure if putting that constant there is a good idea. It's like our spec that we add that prefix (`md5:`) and not `md5` or `md5-` for example right? So it should be a part of this package and a separate module (in the future) as we agreed on a few minutes later.

seadowg: Ah I was thinking eventually we'd make `Md5` a class and have a `parsePrefixString` method on it or something. I think you're right that I'm jumping ahead there. Right now the prefix is only used in the parser so I'll move the constant there.

seadowg: Ok done!

grzesiek2010 (approved):

PR #4912, opened 2021-11-16
Add "Open Settings" button to permission denied dialogs

### LINKED ISSUE #4839: Add "Open Settings" button to permission denied dialogs
Permissions can only be requested twice on Android 11+. After denying twice, the user will just see our permission denied dialogs and won't be offered a way to grant the permission again (other than being told to "try again" which won't work). We should reword these dialogs and add an "Open Settings" button to them that lets the user navigate quickly to the systems App Details page for ODK.

## Acceptance

- [ ] Given I've not granted any permissions yet
When I navigate to a screen that needs permission (like "Form Metadata" settings)
And I deny the permission
Then I see a dialog explaining why Collect needs that permission
And the dialog explains that I can grant the permission in settings

- [ ] Given I've not granted any permissions yet
When I navigate to a screen that needs permission (like "Form Metadata" settings)
And I deny the permission
And I click "Open Settings" on the dialog
Then I end up on the ODK Collect app info/details page in system settings
And I can click into "Permissions" and enable the permission needed

grzesiek2010:
Closes #4839 

I've tested the change manually and added automated tests.

I updated the dialog according to the issue and additional went the extra mile and moved classes responsible for dealing with permissions to their own module.

Please verify that the updated dialog works as expected and also test granting/denying permissions since this part of the app might be affected.

No.

No.


seadowg:
@grzesiek2010 looks like there's some conflicts to fix here before review

grzesiek2010:
> @grzesiek2010 looks like there's some conflicts to fix here before review

solved.

collect_app/src/main/java/org/odk/collect/android/permissions/PermissionsProvider.java
```diff
@@ -224,15 +225,14 @@ public void onPermissionRationaleShouldBeShown(List<PermissionRequest> permissio
     }
 
     protected void showAdditionalExplanation(Activity activity, int title, int message, int drawable, @NonNull PermissionListener action) {
-        AlertDialog alertDialog = new AlertDialog.Builder(activity)
-                .setTitle(title)
-                .setMessage(message)
-                .setPositiveButton(android.R.string.ok, (dialogInterface, i) -> action.denied())
-                .setCancelable(false)
```
seadowg, line 231: Looks like we lose the cancel blocking here, but I'm not sure that matters. Can you think of any reason we'd want these dialogs to not be cancellable?

collect_app/src/test/java/org/odk/collect/android/permissions/PermissionDeniedDialogTest.kt
```diff
@@ -0,0 +1,94 @@
...
+        scenario.onFragment {
+            Intents.init()
+
+            assertThat(it.dialog!!.isShowing, `is`(true))
+            (it.dialog!! as AlertDialog).getButton((AlertDialog.BUTTON_NEUTRAL)).performClick()
+
+            Intents.intended(IntentMatchers.hasAction(Settings.ACTION_APPLICATION_DETAILS_SETTINGS))
+            Intents.intended(IntentMatchers.hasData(Uri.fromParts("package", ApplicationProvider.getApplicationContext<Application>().packageName, null)))
+
+            Intents.release()
```
seadowg: We can use `RecordedIntentsRule` here I think.

collect_app/src/test/java/org/odk/collect/android/utilities/PermissionsTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.manifest.AndroidManifest
+import org.robolectric.res.Fs
+
+/**
+ * Test for checking permissions in [AndroidManifest]
+ */
+@RunWith(AndroidJUnit4::class)
+class PermissionsTest {
```
seadowg: I'm not sure if I see the need for this test. It feels like we're just testing config right?

permissions/build.gradle
```diff
@@ -43,15 +43,19 @@ android {
 
 dependencies {
     implementation project(':strings')
+    implementation project(':androidshared')
 
     implementation Dependencies.androidx_core_ktx
     implementation Dependencies.androidx_fragment
     implementation Dependencies.android_material
+    implementation Dependencies.karumi_dexter
```
seadowg, line 51: We can probably remove this dependency from the `collect_app` `build.gradle` now.

seadowg (changes requested):
Ace! Just a couple of notes but liking this a lot. Great move to extract the permissions components to their own module.

collect_app/src/main/java/org/odk/collect/android/permissions/PermissionsProvider.java
grzesiek2010: No I think there is no reason to make it not cancellable.

collect_app/src/test/java/org/odk/collect/android/permissions/PermissionDeniedDialogTest.kt
grzesiek2010: I had to move `RecordedIntentsRule` to `testshared` but ok now it is possible.

collect_app/src/test/java/org/odk/collect/android/utilities/PermissionsTest.kt
grzesiek2010: I also don't think we need this so ok let's remove it.

permissions/build.gradle
grzesiek2010: It was used in one test but now I got rid of it and it is really unused.

grzesiek2010: It was used in one test but now I got rid of it and it is really unused.

seadowg (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 8.1 and 9.0

Verified cases:
- regression check on Android 5.1
- all permission dialogs: camera, microphone, contacts, phone, location 
- deny/allow
- deny for few times 
- use open settings button
- enable/disable permission in settings 
- screen rotation, extinguish screen when app opened
- light/dark mode

srujner:
**Tested with success!**
Verified on Androids: 10 and 11.0

Verified cases:
- light and dark mode
- All permission dialogs
- Open settings button
- Enable and disable permission in settings
- Screen rotation, extinguish screen when app opened

PR #4914, opened 2021-11-17
Use MaterialAlertDialogBuilder instead of AlertDialog.Builder across the app

grzesiek2010:
Closes #4915 

#### What has been done to verify that this works as intended?
I reviewed implemented changes and tested some dialogs used in our app.

#### Why is this the best possible solution? Were any other approaches considered?
The first thing I changed was using `MaterialAlertDialogBuilder` instead of `AlertDialog.Builder`. Unfortunately our settings which are based on `PreferenceFragmentCompat` use `AlertDialog.Builder` under the hood so I also had to improve styling of `AlertDialog.Builder` to make sure dialogs created using that builder look the same like those created using `MaterialAlertDialogBuilder`.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
This change is related to the way our dialogs are styled so please review them to make sure they look the same.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
Review dialogs used across the app and make sure 

#### Do we need any specific form for testing your changes? If so, please attach one.
No.

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/getodk/docs/issues/new) and include the link below.
No.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#ui-components-style-guidelines)

PR #4917, opened 2021-11-19
Fixed: EDIT action for forms will crash if there are no projects

### LINKED ISSUE #4743: EDIT action for forms will crash if there are no projects
Collect v2021.2

#### Problem description

Trying to open a form from a URI or shortcut when there are no projects causes a crash.

1. Create a shortcut for a form
2. Delete all the projects in Collect
3. Click on shortcut

#### Expected behavior

Collect should a warning dialog and then close when the dialog is dismissed.

#### Other information 

Crashlytics report: https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/7abd65245bbd49ab8e4cd8acb06ad287?time=last-seven-days&versions=v2021.2.0%20(4242)&sessionEventKey=61012C22032C00010321849F7FC5E726_1568363353904663998

Stacktrace:

```
Caused by java.util.NoSuchElementException: List is empty.
       at kotlin.collections.CollectionsKt___CollectionsKt.first(CollectionsKt___CollectionsKt.java:212)
       at kotlin.collections.CollectionsKt.first(CollectionsKt.java)
       at org.odk.collect.android.external.FormUriActivity.onCreate(FormUriActivity.java:28)
       at android.app.Activity.performCreate(Activity.java:7895)
       at android.app.Activity.performCreate(Activity.java:7883)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1306)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3310)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3491)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2055)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7649)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:503)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

grzesiek2010:
Closes #4743 

It's just a fix so this seems to be the only solution.

It should just fix the issue and do not affect anything else so please focus on the scenario described in the issue.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/FormEditActionTest.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/feature/external/FormEditActionTest.kt
@@ -1,20 +1,16 @@
 package org.odk.collect.android.feature.external
 
 import android.content.Intent
-import android.net.Uri
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.RuleChain
 import org.junit.runner.RunWith
-import org.odk.collect.android.R
 import org.odk.collect.android.external.FormsContract
 import org.odk.collect.android.support.CollectTestRule
 import org.odk.collect.android.support.ContentProviderUtils
 import org.odk.collect.android.support.TestRuleChain
-import org.odk.collect.android.support.pages.AppClosedPage
 import org.odk.collect.android.support.pages.FormEntryPage
-import org.odk.collect.android.support.pages.OkDialog
 
 @RunWith(AndroidJUnit4::class)
 class FormEditActionTest {
@@ -35,62 +31,5 @@ class FormEditActionTest {
 
         val intent = Intent(Intent.ACTION_EDIT).also { it.data = uri }
         rule.launch(intent, FormEntryPage("One Question"))
-    }
-
-    @Test
-    fun whenFormIsNotCurrentProject_showsWarningAndExits() {
-        rule.startAtMainMenu()
-            .copyAndSyncForm("one-question.xml")
-            .addAndSwitchToProject("https://example.com")
-
-        val formId = ContentProviderUtils.getFormDatabaseId("DEMO", "one_question")
-        val uri = FormsContract.getUri("DEMO", formId)
-
-        val intent = Intent(Intent.ACTION_EDIT).also { it.data = uri }
-        rule.launch(intent, OkDialog())
-            .assertText(R.string.wrong_project_selected_for_form)
-            .clickOK(AppClosedPage())
-    }
-
-    @Test
-    fun whenUriDoesNotHaveProjectId_andCurrentProjectIsFirstOne_opensForm() {
-        rule.startAtMainMenu()
-            .copyAndSyncForm("one-question.xml")
-            .addAndSwitchToProject("https://example.com")
-            .openProjectSettingsDialog()
-            .selectProject("Demo project")
-
-        val formId = ContentProviderUtils.getFormDatabaseId("DEMO", "one_question")
-        val uri = FormsContract.getUri("DEMO", formId)
-        val uriWithoutProjectId = Uri.Builder()
-            .scheme(uri.scheme)
-            .authority(uri.authority)
-            .path(uri.path)
-            .query(null)
-            .build()
-
-        val intent = Intent(Intent.ACTION_EDIT).also { it.data = uriWithoutProjectId }
-        rule.launch(intent, FormEntryPage("One Question"))
-    }
-
-    @Test
-    fun whenUriDoesNotHaveProjectId_andCurrentProjectIsNotFirstOne_showsWarningAndExits() {
-        rule.startAtMainMenu()
-            .copyAndSyncForm("one-question.xml")
-            .addAndSwitchToProject("https://example.com")
-
-        val formId = ContentProviderUtils.getFormDatabaseId("DEMO", "one_question")
-        val uri = FormsContract.getUri("DEMO", formId)
-        val uriWithoutProjectId = Uri.Builder()
-            .scheme(uri.scheme)
-            .authority(uri.authority)
-            .path(uri.path)
-            .query(null)
-            .build()
-
-        val intent = Intent(Intent.ACTION_EDIT).also { it.data = uriWithoutProjectId }
-        rule.launch(intent, OkDialog())
-            .assertText(R.string.wrong_project_selected_for_form)
-            .clickOK(AppClosedPage())
     }
 }
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
@@ -26,26 +26,37 @@ class FormUriActivity : Activity() {
         super.onCreate(savedInstanceState)
         DaggerUtils.getComponent(this).inject(this)
 
-        val firstProject = projectsRepository.getAll().first()
-        val uri = intent.data
-        val uriProjectId = uri?.getQueryParameter("projectId")
-        val projectId = uriProjectId ?: firstProject.uuid
-
-        logAnalytics(uriProjectId)
-
-        if (projectId == currentProjectProvider.getCurrentProject().uuid) {
-            startActivity(
-                Intent(this, FormEntryActivity::class.java).also {
-                    it.data = uri
-                    intent.extras?.let { sourceExtras -> it.putExtras(sourceExtras) }
-                }
-            )
-        } else {
+        val projects = projectsRepository.getAll()
+        if (projects.isEmpty()) {
             AlertDialog.Builder(this)
-                .setMessage(R.string.wrong_project_selected_for_form)
+                .setMessage(R.string.no_projects_detected)
                 .setPositiveButton(R.string.ok) { _, _ -> finish() }
                 .create()
                 .show()
+        } else {
+            val firstProject = projects.first()
+            val uri = intent.data
+            val uriProjectId = uri?.getQueryParameter("projectId")
+            val projectId = uriProjectId ?: firstProject.uuid
+
+            logAnalytics(uriProjectId)
+
+            if (projectId == currentProjectProvider.getCurrentProject().uuid) {
+                startActivity(
+                    Intent(this, FormEntryActivity::class.java).also {
+                        it.data = uri
+                        intent.extras?.let { sourceExtras -> it.putExtras(sourceExtras) }
+                    }
+                )
+            } else {
+                AlertDialog.Builder(this)
+                    .setMessage(R.string.wrong_project_selected_for_form)
+                    .setPositiveButton(R.string.ok) { _, _ ->
+                        finish()
+                    }
+                    .create()
+                    .show()
+            }
         }
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
@@ -29,7 +29,7 @@ class FormUriActivity : Activity() {
         val projects = projectsRepository.getAll()
         if (projects.isEmpty()) {
             MaterialAlertDialogBuilder(this)
-                .setMessage(R.string.no_projects_detected)
+                .setMessage(R.string.app_not_configured)
                 .setPositiveButton(R.string.ok) { _, _ -> finish() }
                 .create()
                 .show()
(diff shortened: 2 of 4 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/feature/external/FormEditActionTest.kt
```diff
@@ -36,61 +32,4 @@ class FormEditActionTest {
         val intent = Intent(Intent.ACTION_EDIT).also { it.data = uri }
         rule.launch(intent, FormEntryPage("One Question"))
     }
-
-    @Test
-    fun whenFormIsNotCurrentProject_showsWarningAndExits() {
```
seadowg, line 41: Just wanted to talk through the test restructuring here before reviewing. Is the idea that `FormEditActionTest` checks the "happy path" or simplest case (and most importantly that the intent filter is set up correctly) and that the `FormUriActivityTest` checks the more detailed cases (missing project etc)? It looks like a sort of "feature and component test" breakdown but just wanted to check we were on the same page!

grzesiek2010: `FormEditActionTest` covered the same cases like the new `FormUriActivityTest`. I think `FormEditActionTest` was created as a UI test that requires android emulator because we weren't able to test implicit intents using activity scenario https://github.com/android/android-test/issues/496 that issue will be solved soon.

I kept that one test in `FormEditActionTest` for two reasons:
1. What I mentioned above - currently in `FormUriActivityTest` we only test explicit intents so it's nice to keep one that odes that in an implicit way.
2. It's nice to have one UI test for that scenario (simple scenario) but all the other cases should be tested using activity scenario.

seadowg: Right that all makes sense!

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+                projectsRepository: ProjectsRepository?
+            ): CurrentProjectProvider {
+                return currentProjectProvider
+            }
+        })
+    }
+
+    @Test
+    fun `When there are no projects then display alert dialog`() {
+        projectsRepository.deleteAll()
```
seadowg: Personally I'd find these tests easier to read if the projects were added at the start of the test rather than in `setup`. Then, the state of the world (the "given") can be seen next to the actions and assertions (the "when" and "then").

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        ActivityScenario.launch<FormUriActivity>(getIntent())
+
+        Intents.intended(hasComponent(FormEntryActivity::class.java.name))
+
+        Intents.release()
+    }
+
+    @Test
+    fun `When everything is fine form filling is started with given data`() {
```
seadowg: Is this not the same as `When there is project id specified in uri and it matches current project id then start form filling`?

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.odk.collect.android.preferences.source.SettingsProvider
+import org.odk.collect.android.projects.CurrentProjectProvider
+import org.odk.collect.android.support.CollectHelpers
+import org.odk.collect.projects.InMemProjectsRepository
+import org.odk.collect.projects.Project
+import org.odk.collect.projects.ProjectsRepository
+import org.odk.collect.shared.strings.UUIDGenerator
+
+@RunWith(AndroidJUnit4::class)
+class FormUriActivityTest {
```
seadowg, line 43: I think you could use `RecordedIntentsRule` to avoid all the setup and teardown steps in this test.

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        Intents.release()
+    }
+
+    @Test
+    fun `When there is no project id specified in uri and first available project id does not match current project id then display alert dialog`() {
+        whenever(currentProjectProvider.getCurrentProject()).thenReturn(secondProject)
+
+        val scenario = ActivityScenario.launch<FormUriActivity>(getIntent())
+
+        onView(withText(R.string.wrong_project_selected_for_form)).inRoot(isDialog()).check(matches(isDisplayed()))
```
seadowg: I did not know that we could use matchers on dialogs in Robolectric. That's definitely useful.

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        assertThat(scenario.result.resultCode, `is`(Activity.RESULT_CANCELED))
+    }
+
+    @Test
+    fun `When there is project id specified in uri and it matches current project id then start form filling`() {
+        Intents.init()
+
+        ActivityScenario.launch<FormUriActivity>(getIntent(firstProject.uuid))
+
+        Intents.intended(hasComponent(FormEntryActivity::class.java.name))
```
seadowg, line 104: We should probably check the intent details whenever we launch `FormEntryActivity` (like you do in `When everything is fine form filling is started with given data`).

strings/src/main/res/values/strings.xml
```diff
@@ -1111,6 +1111,8 @@
 
     <!-- Error shown when a user tries to open a form that's in a different project -->
     <string name="wrong_project_selected_for_form">To open this form, you must first open ODK Collect and switch to the project that contains it.</string>
+    <!-- Error shown when a user tries to open a form when there are no projects -->
+    <string name="no_projects_detected">No projects detected.</string>
```
seadowg: This is a weird one because it means the shortcut/URI will never work right? If there are no projects, they'd need to add one and the form would end up with a different ID (probably).

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
```diff
@@ -28,26 +28,37 @@ class FormUriActivity : Activity() {
...
+
+            if (projectId == currentProjectProvider.getCurrentProject().uuid) {
+                startActivity(
+                    Intent(this, FormEntryActivity::class.java).also {
+                        it.data = uri
+                        intent.extras?.let { sourceExtras -> it.putExtras(sourceExtras) }
+                    }
+                )
+            } else {
+                AlertDialog.Builder(this)
```
seadowg: Dialogs should use `MaterialAlertDialogBuilder` as per #4914.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
grzesiek2010: Kind of but this one is focused on testings passed data and extras. [EDIT]: Done

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
grzesiek2010: Every time check data and extras?

seadowg: Yeah I think so! Just to make sure we're definitely sending the right stuff through in each case.

strings/src/main/res/values/strings.xml
grzesiek2010: Yes it would work in case of demo projects but other projects will have unique ids.

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
grzesiek2010: Ok, done.

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/external/FormUriActivityTest.kt
grzesiek2010: Done.

strings/src/main/res/values/strings.xml
seadowg: Ok how about: 

```
"ODK Collect has not been configured. Try to open ODK Collect and configure it.\n\nIf you tapped on a shortcut, you might need to recreate it after configuring ODK Collect."
```

grzesiek2010: I agree it's better, done.

strings/src/main/res/values/strings.xml
```diff
@@ -1111,6 +1111,8 @@
 
     <!-- Error shown when a user tries to open a form that's in a different project -->
     <string name="wrong_project_selected_for_form">To open this form, you must first open ODK Collect and switch to the project that contains it.</string>
+    <!-- Error shown when a user tries to open a form when there are no projects -->
+    <string name="app_not_configured">ODK Collect has not been configured. Try to open ODK Collect and configure it.\n\nIf you tapped on a shortcut, you might need to recreate it after configuring ODK Collect.</string>
```
seadowg, line 1115: @lognaturel would appreciate a sanity check on this copywriting.

seadowg (approved):

mmarciniak90:
@grzesiek2010 I noticed one unexpected case:

1. create DEMO project and download forms
2. add shortcuts to one of the form, e.g. _All widgets_
3. remove project
4. click on shortcuts and there is no crash 
5. open Collect and re-create DEMO project 
6. try to download forms -> connecting to server never ends

grzesiek2010:
Reproducing your steps I get `SQLiteReadOnlyDatabaseException`:

```
```

I'm pretty sure it's not my fault because I'm able to reproduce the same exception on master too:
1. Crete Demo project
2. Download one form
3. Delete the project
4. Create Demo project again
5. Try to download forms.

Could you confirm? If I'm right we should create a separate issue.

strings/src/main/res/values/strings.xml
lognaturel: I think it’s ok! This only happens when there are 0 projects, right? Not also when a deleted project is referenced? If it also includes the case where a deleted project is referenced thrn I’d want to edit it a bit.

seadowg: Yeah, you're right: at the moment this only shows for the 0 projects case.

seadowg:
@mmarciniak90 @kkrawczyk123 lets confirm whether the problem here is a regression or not so we can move on with testing for this

kkrawczyk123:
@seadowg It's a little weird, I am able to reproduce the issue on both store version and master but with slightly different results. On master, I see infinite load on connecting to server dialog but on store version at the same step I see a crash. It's not connected to changes from this PR.

grzesiek2010:
Yes I can confirm that the result is different as @kkrawczyk123 said but logs are the same so it's not something introduced by this issue.
Let's continue testing and I will report it in a separate issue.

kkrawczyk123:
**Tested with success!**

Verified cases:
- use form shortcut
- scenario form issue
- information dialog visible 
- use shortcut after downloading form again 
- fill form

I have noticed that I am not able to create shortcut on Xiaomi devices (I have tried on 2 separate devices and also older versions of Collect). I am not sure if I should report it and if we can even do something with that on Collect side. @grzesiek2010 @seadowg what do you think?

seadowg:
> I am not sure if I should report it and if we can even do something with that on Collect side. @grzesiek2010 @seadowg what do you think?

@kkrawczyk123 I'd report that as a new issue, and we can investigate. There might be something we're doing that they don't support.

srujner:
**Tested with success!**

I also noticed not being able to create shortcut on Xiaomi

PR #4918, opened 2021-11-23
Make landing screen layout more robust to different button text lengths

### LINKED ISSUE #4693: Make landing screen configuration button constraints more robust
The constraint between the landing screen QR code and manual configuration buttons currently assumes that the manual configuration label text will be longer. This constraint should be updated so that it doesn't matter which button is wider.

grzesiek2010:
Closes #4693 

I tested the fix manually hardcodding strings with different length.

I think it's best that both buttons take the whole space horizonatlly no matter how long their strings are:

| Original strings | First string shorter | Second string shorter |
| Screenshot_1638187850 | Screenshot_1638187513 | Screenshot_1638187532  |

It's a pretty safe change so maybe just test the `First Launch` screen in two different languages ( to verify different string lengths).

No.

No.

### DIFF
--- a/collect_app/src/main/res/layout/first_launch_layout.xml
+++ b/collect_app/src/main/res/layout/first_launch_layout.xml
@@ -93,49 +93,46 @@
             android:id="@+id/app_name"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
+            android:layout_marginTop="@dimen/margin_extra_large"
             android:layout_marginBottom="@dimen/margin_extra_small"
             android:textAppearance="?textAppearanceBody1"
             android:textColor="@color/color_on_surface_medium_emphasis"
-            android:layout_marginTop="@dimen/margin_extra_large"
-            app:layout_constraintBottom_toTopOf="@+id/configure_later"
+            app:layout_constraintBottom_toTopOf="@id/dont_have_server"
             app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintHorizontal_bias="0.5"
-            app:layout_constraintVertical_bias="1"
             app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toBottomOf="@+id/center"
+            app:layout_constraintTop_toBottomOf="@id/center"
+            app:layout_constraintVertical_bias="1"
             tools:text="ODK Collect v2022.3" />
 
-        <androidx.constraintlayout.widget.ConstraintLayout
-            android:id="@+id/configure_later"
+        <TextView
+            android:id="@+id/dont_have_server"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_marginBottom="@dimen/margin_standard"
-            android:background="?attr/selectableItemBackgroundBorderless"
+            android:layout_marginBottom="@dimen/margin"
+            android:text="@string/dont_have_project"
+            android:textAppearance="?textAppearanceBody2"
             app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toStartOf="@id/try_collect"
+            app:layout_constraintHorizontal_chainStyle="packed"
+            app:layout_constraintStart_toStartOf="parent" />
+
+        <TextView
+            android:id="@+id/try_collect"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="@dimen/margin_extra_small"
+            android:text="@string/try_demo"
+            android:textAppearance="@style/TextAppearance.Collect.Link"
+            app:layout_constraintBottom_toBottomOf="@+id/dont_have_server"
             app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintHorizontal_bias="0.5"
-            app:layout_constraintVertical_bias="1"
-            app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toBottomOf="@+id/center">
+            app:layout_constraintStart_toEndOf="@+id/dont_have_server"
+            app:layout_constraintTop_toTopOf="@+id/dont_have_server" />
 
-            <TextView
-                android:id="@+id/dont_have_server"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:text="@string/dont_have_project"
-                android:textAppearance="?textAppearanceBody2"
-                app:layout_constraintBottom_toBottomOf="parent"
-                app:layout_constraintStart_toStartOf="parent" />
-
-            <TextView
-                android:id="@+id/try_collect"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:layout_marginStart="@dimen/margin_extra_small"
-                android:text="@string/try_demo"
-                android:textAppearance="@style/TextAppearance.Collect.Link"
-                app:layout_constraintStart_toEndOf="@+id/dont_have_server"
-                app:layout_constraintTop_toTopOf="@+id/dont_have_server" />
-        </androidx.constraintlayout.widget.ConstraintLayout>
+        <androidx.constraintlayout.widget.Group
+            android:id="@+id/configureLater"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:visibility="visible"
+            app:constraint_referenced_ids="dont_have_server,try_collect" />
     </androidx.constraintlayout.widget.ConstraintLayout>
 </ScrollView>
--- a/collect_app/src/main/res/layout/first_launch_layout.xml
+++ b/collect_app/src/main/res/layout/first_launch_layout.xml
@@ -43,48 +43,39 @@
                 android:lines="2"
                 android:text="@string/tagline"
                 android:textAppearance="?textAppearanceHeadline4"
-                app:layout_constraintBottom_toTopOf="@+id/configuration_buttons"
-                app:layout_constraintStart_toStartOf="@+id/configuration_buttons"
+                app:layout_constraintStart_toStartOf="parent"
                 app:layout_constraintTop_toTopOf="parent" />
 
-            <androidx.constraintlayout.widget.ConstraintLayout
-                android:id="@+id/configuration_buttons"
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/configure_via_qr_button"
+                style="@style/Widget.Collect.Button.OutlinedButton"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_marginTop="@dimen/margin"
+                android:backgroundTint="?colorPrimary"
+                android:text="@string/configure_with_qr_code"
+                android:textAppearance="?attr/textAppearanceBody1"
+                android:textColor="@android:color/white"
+                android:layout_marginTop="@dimen/margin_standard"
+                app:icon="@drawable/ic_baseline_qr_code_scanner_24"
+                app:iconTint="@android:color/white"
+                app:layout_constraintTop_toBottomOf="@id/tagline"
+                app:layout_constraintBottom_toTopOf="@+id/configure_manually_button"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent" />
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/configure_manually_button"
+                style="@style/Widget.Collect.Button.OutlinedButton"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="@dimen/margin_small"
+                android:text="@string/configure_manually"
+                android:textAppearance="?attr/textAppearanceBody1"
+                app:icon="@drawable/ic_outline_edit_24"
+                app:iconTint="?colorOnSurface"
                 app:layout_constraintEnd_toEndOf="parent"
                 app:layout_constraintStart_toStartOf="parent"
-                app:layout_constraintTop_toBottomOf="@id/tagline">
-
-                <com.google.android.material.button.MaterialButton
-                    android:id="@+id/configure_via_qr_button"
-                    style="@style/Widget.Collect.Button.OutlinedButton"
-                    android:layout_width="match_parent"
-                    android:layout_height="wrap_content"
-                    android:backgroundTint="?colorPrimary"
-                    android:text="@string/configure_with_qr_code"
-                    android:textAppearance="?attr/textAppearanceBody1"
-                    android:textColor="@android:color/white"
-                    app:icon="@drawable/ic_baseline_qr_code_scanner_24"
-                    app:iconTint="@android:color/white"
-                    app:layout_constraintBottom_toTopOf="@+id/configure_manually_button"
-                    app:layout_constraintEnd_toEndOf="parent"
-                    app:layout_constraintStart_toStartOf="parent" />
-
-                <com.google.android.material.button.MaterialButton
-                    android:id="@+id/configure_manually_button"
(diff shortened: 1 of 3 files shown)


collect_app/src/main/res/layout/first_launch_layout.xml
```diff
@@ -48,7 +49,7 @@
 
             <androidx.constraintlayout.widget.ConstraintLayout
```
seadowg: What do you think about flattening this out so we don't nested constraint layouts? I'm not seeing a reason to nest them here and doing so has a (probably very small) performance cost and, for me, makes the layout preview harder to understand.

seadowg (changes requested):

lognaturel:
Unless I’ve misunderstood, I don’t think this is what we want. Can you please post a screen capture on a tablet? Also consider landscape on a phone form factor.

seadowg:
>  Can you please post a screen capture on a tablet? Also consider landscape on a phone form factor.

@lognaturel great point. You can see the problem in the layout inspector (using Nexus 10 here):

Screenshot 2021-11-26 at 08 53 31

I think we might want to use the `0dp` ("match constraint") width on the buttons here and then use `constraintWidth_max` to set a max width. The title can then anchor to the buttons.

grzesiek2010:
> Unless I’ve misunderstood, I don’t think this is what we want.

so what's the expected behavior? because it's not clear from the issue. I'm attaching more screenshots
Landscape:
Screenshot_1637916554

Tablet:
Screenshot_1637916895
:

grzesiek2010:
Do you maybe want to use `wrap_content`?
Screenshot_1637917037

Screenshot_1637916973

lognaturel:
The ideal outcome would be that both buttons are the width needed to accommodate the widest text. I think that should be possible but I haven’t had a chance to come back and see what tools we could use for that. The constraints might need to be defined programmatically but I think it would be possible with the layout editor.

seadowg:
> The ideal outcome would be that both buttons are the width needed to accommodate the widest text.

This is actually pretty complex because the views end up referencing each other's width in a circle. I found a good SO about how to solve it here, but it's a bit of a mess: https://stackoverflow.com/questions/59035011/constraint-layout-two-views-with-largest-width.

Just to restate my original suggestion:

> I think we might want to use the 0dp ("match constraint") width on the buttons here and then use constraintWidth_max to set a max width. The title can then anchor to the buttons.

grzesiek2010:
I updated the solution suing [Barrier](https://developer.android.com/reference/androidx/constraintlayout/widget/Barrier) according to this answer: https://stackoverflow.com/a/63630534/5479029
I think it looks good (screenshots have been updated so you can take a look at the description).

seadowg:
@grzesiek2010 @lognaturel using the `Barrier` does indeed seem to work but that version makes it so that the buttons are left aligned rather than centred. I couldn't get my head around a way to do it with `Barrier` and keep everything centred. The best way I can come up with to think about this and get it working is to go back to using max width. To be clear, by that I mean just letting the buttons grow as large as they can and keep them limited by margins (on smaller devices) and max width (on larger devices). This means the buttons won't ever get smaller than the size available to them, but I don't think that's a problem here - we always want these buttons to be pretty big. I don't think keeping the buttons the same size as each other is actually what we need. Here's an example of a flattened (which for me is easier to think about) version of the layout that uses `chainStyle`, `constraintWidth_max` and margins to get this working:

```xml
<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:fillViewport="true">

    <androidx.constraintlayout.widget.ConstraintLayout
        android:background="?attr/colorSurface">

        <ImageView
            android:id="@+id/logo"
            android:layout_width="60dp"
            android:layout_marginTop="@dimen/margin_standard"
            android:adjustViewBounds="true"
            android:contentDescription="@string/collect_app_name"
            android:src="@drawable/odk_logo"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toTopOf="parent" />

        <TextView
            android:id="@+id/tagline"
            android:lines="2"
            android:text="@string/tagline"
            android:textAppearance="?textAppearanceHeadline4"
            app:layout_constraintBottom_toTopOf="@id/configure_via_qr_button"
            app:layout_constraintStart_toStartOf="@id/configure_via_qr_button"
            app:layout_constraintTop_toTopOf="parent"
            app:layout_constraintVertical_chainStyle="packed" />

        <com.google.android.material.button.MaterialButton
            android:id="@+id/configure_via_qr_button"
            style="@style/Widget.Collect.Button.OutlinedButton"
            android:layout_width="0dp"
            android:layout_marginHorizontal="54dp"
            android:layout_marginTop="@dimen/margin_standard"
            android:backgroundTint="?colorPrimary"
            android:text="@string/configure_with_qr_code"
            android:textAppearance="?attr/textAppearanceBody1"
            android:textColor="@android:color/white"
            app:icon="@drawable/ic_baseline_qr_code_scanner_24"
            app:iconTint="@android:color/white"
            app:layout_constraintBottom_toTopOf="@+id/configure_manually_button"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toBottomOf="@id/tagline"
            app:layout_constraintWidth_max="400dp"
            tools:text="Blah" />

        <com.google.android.material.button.MaterialButton
            android:id="@+id/configure_manually_button"
            style="@style/Widget.Collect.Button.OutlinedButton"
            android:layout_width="0dp"
            android:layout_marginHorizontal="54dp"
            android:layout_marginTop="@dimen/margin_small"
            android:text="@string/configure_manually"
            android:textAppearance="?attr/textAppearanceBody1"
            app:icon="@drawable/ic_outline_edit_24"
            app:iconTint="?colorOnSurface"
            app:layout_constraintBottom_toBottomOf="parent"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toBottomOf="@id/configure_via_qr_button"
            app:layout_constraintWidth_max="400dp"
            tools:text="Oh wow this text is long wow how long it is goodness gracious me" />

        <TextView
            android:id="@+id/app_name"
            android:layout_marginTop="@dimen/margin_extra_large"
            android:layout_marginBottom="@dimen/margin_extra_small"
            android:textAppearance="?textAppearanceBody1"
            android:textColor="@color/color_on_surface_medium_emphasis"
            app:layout_constraintBottom_toTopOf="@id/dont_have_server"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toBottomOf="@id/configure_manually_button"
            app:layout_constraintVertical_bias="1"
            tools:text="ODK Collect v2022.3" />

        <TextView
            android:id="@+id/dont_have_server"
            android:layout_marginBottom="@dimen/margin"
            android:text="@string/dont_have_project"
            android:textAppearance="?textAppearanceBody2"
            app:layout_constraintBottom_toBottomOf="parent"
            app:layout_constraintEnd_toStartOf="@id/try_collect"
            app:layout_constraintHorizontal_chainStyle="packed"
            app:layout_constraintStart_toStartOf="parent" />

        <TextView
            android:id="@+id/try_collect"
            android:layout_marginStart="@dimen/margin_extra_small"
            android:text="@string/try_demo"
            android:textAppearance="@style/TextAppearance.Collect.Link"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toEndOf="@+id/dont_have_server"
            app:layout_constraintTop_toTopOf="@+id/dont_have_server" />
    </androidx.constraintlayout.widget.ConstraintLayout>
</ScrollView>
```

This looks like this on a Pixel 2:

Screenshot 2021-11-30 at 12 10 07

And like this on a Nexus 10:

Screenshot 2021-11-30 at 12 10 57

If the text for one of the buttons is really short we get this:

Screenshot 2021-11-30 at 12 11 42

And if it's really long we get this:

Screenshot 2021-11-30 at 12 12 57

Thoughts?

grzesiek2010:
It can be centered I would just need to set this to wrap_content https://github.com/getodk/collect/pull/4918/files#diff-4bdd0c9940f9e48488525c6896117ea1308978c0df801859eed4c4d807717ad2R29 I just didn't know it has to be centered because I usually work on smaller devices. Doesn't it work for you?

seadowg:
> It can be centered I would just need to set this to wrap_content https://github.com/getodk/collect/pull/4918/files#diff-4bdd0c9940f9e48488525c6896117ea1308978c0df801859eed4c4d807717ad2R29

Ah yeah that works! I'd still prefer a single flatted `ConstraintLayout` if I'm being honest but if this gets the job done then I guess that's good enough.

grzesiek2010:
> Ah yeah that works!

ok so I have updated the pr with that fix.

seadowg (dismissed):
Actually it looks like this layout doesn't work when the tagline is longer than the buttons (as it is in French):

Screenshot 2021-12-01 at 09 44 30

Again, I think using max width is a simpler way to solve this but maybe that's just me!

grzesiek2010:
The issue was about buttons not the tagline so I didn't adjust it in any way. What's the expected behavior then? Is it bad that it is longer than buttons?

seadowg:
> The issue was about buttons not the tagline so I didn't adjust it in any way. What's the expected behavior then? Is it bad that it is longer than buttons?

Sorry I wasn't clear. The problem is that the buttons move off center if the tagline is longer.

grzesiek2010:
> Sorry I wasn't clear. The problem is that the buttons move off center if the tagline is longer.

It's because the whole container is centered. I'm pretty sure it's the same behavior we have on master, am I wrong? I need to check...

seadowg:
> It's because the whole container is centered.

Yeah we can avoid that by using a single, flattened out `ConstraintLayout` (from my earlier example):

Screenshot 2021-12-01 at 10 07 42

grzesiek2010:
Ok I have nothing against your solution if it works well but the issue was not about it and I don't know if what you expect (centering buttons not the whole container) is the desired behavior. We didn't discuss it.

seadowg:
> I don't know if what you expect (centering buttons not the whole container) is the desired behavior. We didn't discuss it.

Yeah fair enough it's not well-defined. I guess my assumption is that the buttons should always be centered (and of matching size) and that the tagline should be left aligned to the start of the buttons. Does that sound right @lognaturel? I know it's not part of the issue directly I think we should remove the nesting in the layout as it seems like it's not ideal and goes against the reasoning behind using a `ConstraintLayout` (SO on the subject [here](https://stackoverflow.com/questions/53813792/nested-constraint-layout-is-it-a-bad-practice-or-not) which references the Android Dev Blog as well). Personally I think it's also complicating the implementation here, but I understand if that's just me!

Also, if this is just not a big deal™️ then feel free to override me and accept this @lognaturel. I'm aware I'm being a stickler here.

lognaturel:
Thanks for sticking with this! What I had in mind was the outcome this pr now has (I believe. I’ve also explicitly answered one of your qs below). Now I remember that I  did expect barrier would be involved.

> the buttons should always be centered (and of matching size) and that the tagline should be left aligned to the start of the buttons.

Agreed. 

That said, I think matching the tagline width is an additional nice improvement I hadn’t considered. And I’m always for keeping things simpler to reason about. You two should decide together — I’m happy either way.

seadowg:
> That said, I think matching the tagline width is an additional nice improvement 

I do think we should fix the layout so we don't lose the button centering in French (or any other language with a long tagline) - `master` doesn't have that problem as far as I can see. It's not a big deal as no one's really going to be looking at the app in two different languages and get annoyed, but it does mean our layout doesn't behave as intended.

grzesiek2010:
So which option is the expected one?:
- both buttons should be centered (wrapped in a container which will be centered) then the tagline placed above.
- only the first button should be centered the the second button should be placed below and the tagline above.

seadowg:
> only the first button should be centered the the second button should be placed below and the tagline above.

Kind of this. Doesn't need to be implemented exactly that way, but I'm guessing you mean that the tagline and the two buttons should be vertically centered as a group. That was my reasoning behind using `chainStyle="packed"` in the example - it packs the constrained items in to the space (and then the first/last items are constrained to the parent). In addition:

* Both buttons should be horizontally centred and of equal width
* The tagline should be anchored to the first button's start/left and be start/left aligned
* The logo and the footer (app name and try a demo) shouldn't overlap the tagline/buttons on shorter screens (scrolling should kick in)

We might also want to limit the width on the tagline, but that's an existing problem on master so we can look at that another time.

grzesiek2010:
Do you agree @lognaturel?

lognaturel:
Yes, I agree with everything @seadowg wrote in https://github.com/getodk/collect/pull/4918#issuecomment-984537570 except 

> We might also want to limit the width on the tagline, but that's an existing problem on master so we can look at that another time.

We explicitly tell translators to break the tagline into two lines in a way that makes sense in their language. I think that's sufficient. We don't give guidance on number of characters so there might be some screen sizes where things aren't quite right but narrow devices are pretty rare these days so unless anyone complains I think what we're doing is ok.

grzesiek2010:
So I think my solution pretty much does everything we need. I tried your solution @seadowg (the code you have attached) and the result is:
Screenshot_1639139485

as I understand this is not what we expected and buttons with shorter text should not have that much padding right? Both buttons should take less space horizontally and be centered.

seadowg:
> as I understand this is not what we expected and buttons with shorter text should not have that much padding right? Both buttons should take less space horizontally and be centered.

That was the only way I could find to get this to work. In reality that button text is always pretty long. Also, I think when short it still looks good for them to fill the space on a narrower device, but stay within a max width for wider ones. 

The problems I have with the current implementation are:

1. We sometimes end up with left aligned buttons (if the tagline text is long enough):

    Screenshot 2021-12-15 at 10 13 59

2: I'm not happy that we have nested `ConstraintLayout`s going on here. Again, that's [not ideal](https://stackoverflow.com/questions/53813792/nested-constraint-layout-is-it-a-bad-practice-or-not).

I suppose the alignment on the buttons isn't a big deal as it actually looks fine and people aren't going to be switching languages on that screen but to me, it indicates that the layout doesn't actually behave the way we want it to. This is a new issue introduced by this PR as well, and not something that existed before.

However, I've mentioned all this a few times now so if both of you are happy with the current implementation then lets just merge. This isn't a big enough issue to get stuck up on the details. I'll dismiss my review so I'm not blocking anything!

grzesiek2010:
> That was the only way I could find to get this to work. In reality that button text is always pretty long. Also, I think when short it still looks good for them to fill the space on a narrower device, but stay within a max width for wider ones.

I understand but the goal of this issue was to make those buttons smarter not keep the same width like in your case (no matter if we have short text or long the size of the buttons is the same):
| Short text  | Long text |
| Screenshot_1639755147 | Screenshot_1639755094 |


> I'm not happy that we have nested ConstraintLayouts going on here. Again, that's not ideal.

I know  that and I have even experienced how severe problems nesting multiple constraint layouts might cause. But in our case it's just one nested layout (I removed the one that was on the bottom of the screen to make you at least a bit happy :)) so it won't cause visible problems. 

So I will stick to my solution unless you both decide it's wrong.

lognaturel:
Looking at this closer, I'm really pleased with it! We could get away with something simpler as suggested by @seadowg but I think this behavior is spot on and the implementation is not that complex. Thanks for sticking with it and bouncing ideas back and forth to get there.

lognaturel (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 10.

Verified cases:
- English
- Spanish
- German
- RTL language 
- portrait/landscape

srujner:
Tested with success!
Verified also on Android: 11

PR #4920, opened 2021-11-25
Attempt to solve CI memory issues

seadowg:
I ran these changes a few times and didn't see any 137 crashes (out of memory). Hopefully this is an improvement?

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -197,9 +197,6 @@ android {
         unitTests {
             includeAndroidResources = true
             returnDefaultValues = true
-            all {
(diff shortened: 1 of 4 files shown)


collect_app/build.gradle
```diff
@@ -197,9 +197,6 @@ android {
         unitTests {
             includeAndroidResources = true
             returnDefaultValues = true
-            all {
-                maxHeapSize = robolectricHeapSize
```
seadowg, line 201: This looks like it might not be relevant and actually be causing problems with newer versions of Robolectric.

grzesiek2010 (approved):

PR #4924, opened 2021-12-02
Use JSON Schema to validate settings

### LINKED ISSUE #4716: Write up schema for Collect settings
We eventually want Central to be able to generate a QR code for Collect that contains settings other than just the server URL (customized by the Central user). Central will probably need a reference of what settings Collect supports to help it create the UI and also prevent the two from getting out of sync.

One possibility for this would be to use [JSON Schema](https://json-schema.org/) to define what settings Collect supports. Collect could also use this for validation when ingesting settings from a QR code (although that can be a separate issue).

The schema should probably live in this repo in the `docs` dir.

seadowg:
Closes #4716 

I ended up using the schema for validation within Collect as well as just writing up as it was pretty simple to get that working. When reviewing, it would be good to double-check the types and `enum` values and think about if there is any validation we should add/remove.

We talked back and forth about testing here, and that inspired me to pull out a `settings` module that manages ODK app settings. This ended up being a (mostly good) rabbit hole as it seems there's a lot of beneficial restructuring we can do in this area. However, I've stopped short of doing everything I'd have like to do as this change set has been off `master` for a while, I've already had to deal with a couple of nasty rebases, and frankly I'm a little burnt out on this part of the code. 

Here's some things I'd have liked to do (interested to see if anyone thinks these should be done before merging):

* Create a separate module for `Settings` (including migration framework and in mem implementation) with another android library module for shared prefs implementation
* Meta settings should be separated back out to `collect_app` - importing shouldn't touch meta and meta should have it's own migration code
* `ProjectDetailsCreator` should move to `projects` and should not have interface
* Rework `SettingsImporter` to deal with a single `Settings` and have `ODKAppSettingsMigrator` call it twice
* Make `projects` and `settings` modules pure Java (rather than Android)
* Make `ColorPickerDialog`/`ColorPickerViewModel` use dynamic colors (so color list isn't duplicated)

There are a lot of files touched here, but most of them are just imports. I'll add comments inline to make sure the more important things are highlighted.

Ran tests.

Even without the need to provide Central with a spec, this seems like a better solution that what we had before. It creates a single source of truth for our settings structure, and it also means that the validation code is mostly out of our hands.

Should have no effect whatsoever. Good to check adding projects with QR codes to check if there are any mistakes in the spec - the tests cover a few different configs but not everything.

I do think it'd be good to host the schema at the URL specified in `$id` in the spec. We should also replace the spec in the docs with this new schema. We should plan to work on that once this has merged but for now there's no rush as the schema is still only being used internally (within Collect).


collect_app/src/main/java/org/odk/collect/android/preferences/keys/ProjectKeys.kt
```diff
@@ -19,9 +19,6 @@ object ProjectKeys {
     const val KEY_SELECTED_GOOGLE_ACCOUNT = "selected_google_account"
     const val KEY_GOOGLE_SHEETS_URL = "google_sheets_url"
 
-    // project_display.xml
-    const val KEY_PROJECT_DISPLAY = "project_display"
```
seadowg, line 23: This is a navigation item in prefs, not an actual setting.

collect_app/src/test/java/org/odk/collect/android/configure/JsonSchemaSettingsValidatorTest.kt
```diff
@@ -0,0 +1,68 @@
+package org.odk.collect.android.configure
+
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.junit.Test
+
+class JsonSchemaSettingsValidatorTest {
```
grzesiek2010: Don't you think it would make sense to test our real settings too with their schema?

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+          "enum": [
+            "1",
+            "2",
+            "3",
+            "4"
+          ]
+        },
+        "mapbox_map_style": {
+          "type": "string",
+          "enum": [
```
grzesiek2010: I think here you should have 6 values and you forgot `mapbox://styles/mapbox/streets-v11`?

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+            "off",
+            "wifi_only",
+            "cellular_only",
+            "wifi_and_cellular"
+          ]
+        },
+        "delete_send": {
+          "type": "boolean"
+        },
+        "default_complete": {
```
grzesiek2010: Should be `default_completed` I guess.

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+          "type": "boolean"
+        },
+        "constraint_behaviour": {
```
grzesiek2010: Should be `constraint_behavior`

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+            "on_finalize"
+          ]
+        },
+        "high_resolution": {
+          "type": "boolean"
+        },
+        "image_size": {
+          "type": "string",
+          "enum": [
+            "original",
```
grzesiek2010: This should be `original_image_size` I guess.

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+        "form_update_mode": {
+          "type": "boolean"
+        },
+        "periodic_form_updates_check": {
+          "type": "boolean"
+        },
+        "automatic_update": {
+          "type": "boolean"
+        },
+        "hide_old_form_version": {
```
grzesiek2010: Should be `hide_old_form_versions`

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,339 @@
...
+        "change_autosend": {
+          "type": "boolean"
+        },
+        "delete_after_send": {
+          "type": "boolean"
+        },
+        "default_to_finalized": {
+          "type": "boolean"
+        },
+        "change_constraint_behaviour": {
```
grzesiek2010: Should be `change_constraint_behavior`

collect_app/src/main/res/raw/settings_schema.json
seadowg: Hmm it looks like this is wrong in our docs ([here](https://docs.getodk.org/collect-import-export/)) and the admin and general keys don't match in our code. Will have a quick look into that.

collect_app/src/main/res/raw/settings_schema.json
```diff
@@ -0,0 +1,340 @@
...
+        "save_mid": {
+          "type": "boolean"
+        },
+        "save_as": {
+          "type": "boolean"
+        },
+        "mark_as_finalized": {
+          "type": "boolean"
+        }
+      }
```
grzesiek2010: What about `allow_other_ways_of_editing_form` It is not a real setting in our preferences but it's a result of other settings so probably if we do not allow to include it it might lead to some strange states?

grzesiek2010: but including it also might lead to an incorrect state if  the values doesn't match related prefs. So it shouldn't be here but we should make sure we handle this case during importing settings, I'm afraid we don't...

collect_app/src/test/java/org/odk/collect/android/configure/JsonSchemaSettingsValidatorTest.kt
seadowg: This is a good thing to discuss. The schema validation with real settings gets tested (as part of the full stack) through a few different places already - `SettingsImporterRegressionTest`, `ConfigureWithQRCodeTest` for example. I think if we were starting from scratch it would be nice to have been adding a test for each setting as we went (checking it accepts and rejects the right things) but backfilling this now would be a massive job. 

The weird thing that comes up in my head whenever I think about adding tests for the schema itself is that once we're sharing this with Central the schema itself becomes the source of truth. In that world, our job is just to integrate with it (which the tests in `JsonSchemaSettingsValidatorTest` ensure to some degree). Maybe we adding a case to `SettingsImporterRegressionTest` to check the "all settings" case (testing almost the full stack of the import) or maybe we want to make sure the `enum` settings all work correctly seeing as they're a bit more complex? Do you have an idea of what kind of testing you'd like to see?

collect_app/src/main/res/raw/settings_schema.json
seadowg: I'm actually not familiar with that setting, but I can look into it. So right now could someone include that in JSON and end up in a weird state?

grzesiek2010: That option is set to false if you uncheck `Moving backwards` (in Protected -> Access Control -> From Entry Settings) and click Yes (it has an additional dialog). Then if you check `Moving backwards` it is set to true.

If it's false `Jump to` and `Save` options in form entry are disabled.

So I think if someone loads settings where `Moving backwards` is checked (true) but `allow_other_ways_of_editing_form` is false it leads to a state that normally should not be possible to take place.

seadowg: Interesting. That's hopefully something we can guard against with this schema. Are you able to create an issue for that? It sounds like it's an existing issue unless I'm misunderstanding something.

grzesiek2010: Ok I will report a separate issue for that.

collect_app/src/test/java/org/odk/collect/android/configure/JsonSchemaSettingsValidatorTest.kt
grzesiek2010: The thing is that those test we already have verify just a narrow set of prefs. I don't know... Here you are just testing an external library with some simple data I don't think it's important if you took that library it should have it's own tests so I think that it would be more important to test our scheme if it works correctly instead of the library (of course if you do that with our scheme you will test both).

seadowg: > Here you are just testing an external library with some simple data

Yeah `JsonSchemaValidatorTest` is really just checking our implementation integrates with the library correctly - it doesn't have anything to do with our actual schema. I think that's good to have, but I do agree that we might want more. Do you think we should just add a smoke test that checks importing a JSON with all settings? I think it might also be nice to backfill tests for validating the different `enum` values, but that would be better as a separate issue I think.

grzesiek2010: Ok so I guess `SettingsImporterTest` could be improved... We should verify cases like:
- Generally invalid jsons should be ignored
- General Settings should be required
- Admin settings should be required
- Unknown keys should be ignored 
- If enums used and values other than listed in enums exist, should be ignored

some of those cases are already covered.

> Do you think we should just add a smoke test that checks importing a JSON with all settings?

Yes I think it would make sense to do that to make sure our scheme is not broken but also that importing all the keys works properly.

grzesiek2010: Now as I look at `SettingsImporter` a lot of things happen there. For example we first load all prefs from json and then after doing that we remove those unknown. This is weird. It would be better to sanitize our json first. We can remove objects from json right so we could go through it and remove unknown elements first so that later we can import everything and don't think about some potential issues.

grzesiek2010: @seadowg what do you think about those tests? are you going to add anything?

seadowg: Yup sorry will be back to this soon!

collect_app/src/main/java/org/odk/collect/android/preferences/Defaults.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import com.google.android.gms.maps.GoogleMap
+import com.mapbox.mapboxsdk.maps.Style
+import org.odk.collect.android.R
+import org.odk.collect.android.application.Collect
+import org.odk.collect.android.utilities.QuestionFontSizeUtils
+import org.odk.collect.settings.keys.ProjectKeys
+import org.odk.collect.settings.keys.ProtectedProjectKeys
+
+object Defaults {
```
seadowg, line 11: Separating out defaults made it easier to move things over to the new module in steps and also felt cleaner.

settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
```diff
@@ -0,0 +1,33 @@
+package org.odk.collect.settings
+
+import org.odk.collect.projects.Project
+import org.odk.collect.projects.ProjectsRepository
+import org.odk.collect.settings.importing.ProjectDetailsCreatorImpl
+import org.odk.collect.settings.importing.SettingsChangeHandler
+import org.odk.collect.settings.importing.SettingsImporter
+import org.odk.collect.settings.validation.JsonSchemaSettingsValidator
+
+class ODKAppSettingsImporter(
```
seadowg, line 10: Once I'd moved everything over, I realized that the constructor for `SettingsImporter` needed items that could now be internal to `settings`. Instead of spending time reworking that I've just wrapped `SettingsImporter` in a new, more convenient API. As noted in the PR description, there is more work we could do here.

settings/src/test/java/org/odk/collect/settings/ODKAppSettingsImporterTest.kt
```diff
@@ -0,0 +1,44 @@
+package org.odk.collect.settings
+
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.junit.Test
+import org.odk.collect.projects.InMemProjectsRepository
+import org.odk.collect.projects.Project
+import org.odk.collect.settings.support.SettingsUtils.assertSettingsEmpty
+
+class ODKAppSettingsImporterTest {
```
seadowg, line 10: Testing at this level is nice for checking specifics of our schema etc while also exercising the whole import "framework". I looked at adding tests for the enums like @grzesiek2010 suggested, but I backed out as it was going to be a lot of busy work. We don't currently have coverage for those, so I don't think it makes sense to spend any longer here adding that in.

As in other comments and the description, the coverage here and in `SettingsImporterTest` could be moved around with a rework of that relationship.

seadowg:
@grzesiek2010 I've updated the PR description so worth checking out as part of the review.

settings/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+
+    kotlinOptions {
+        jvmTarget = '1.8'
+    }
+
+    packagingOptions {
+        // These library licenses will be referenced in-app
```
grzesiek2010, line 39: It looks like we don't need this.

collect_app/build.gradle
```diff
@@ -225,6 +231,7 @@ configurations.all {
         cacheChangingModulesFor(0, 'seconds')
 
         force 'org.objenesis:objenesis:2.6'
+        force 'joda-time:joda-time:2.10.12'
```
grzesiek2010: I think we don't need this.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
```diff
@@ -87,20 +90,6 @@ class ProjectDetailsCreatorTest {
 
     @Test
     fun `If project color is included in project details but invalid should be generated based on project name`() {
-        assertThat(projectDetailsCreator.createProjectFromDetails(name = "Project X", icon = "#cc").color, `is`("#9f50b0"))
-    }
-
-    @Test
-    fun `Generated project color should be the same for identical project names`() {
```
grzesiek2010, line 93: Don't we need these tests anymore?

settings/src/test/java/org/odk/collect/settings/support/SettingsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+            settings.save((pairs[i] as String?)!!, pairs[i + 1])
+            i += 2
+        }
+    }
+
+    @JvmStatic
+    fun assertSettings(settings: Settings, vararg pairs: Any?) {
+        var i = 0
+        while (i + 1 < pairs.size) {
+            Assert.assertEquals(pairs[i + 1], settings.getAll()[pairs[i]])
```
grzesiek2010: We usually import `Assert.assertEquals` in our tests and do not repeat `Assert. `

settings/src/test/java/org/odk/collect/settings/support/SettingsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Assert
+import org.odk.collect.shared.settings.Settings
+
+object SettingsUtils {
+
+    @JvmStatic
+    fun initSettings(settings: Settings, vararg pairs: Any?) {
+        settings.clear()
+        var i = 0
+        while (i + 1 < pairs.size) {
```
grzesiek2010: It could be simplified:
```
        for (i in pairs.indices step 2) {
            settings.save((pairs[i] as String?)!!, pairs[i + 1])
        }
```

settings/src/test/java/org/odk/collect/settings/support/SettingsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+            settings.save((pairs[i] as String?)!!, pairs[i + 1])
+            i += 2
+        }
+    }
+
+    @JvmStatic
+    fun assertSettings(settings: Settings, vararg pairs: Any?) {
+        var i = 0
+        while (i + 1 < pairs.size) {
```
grzesiek2010: As above.

grzesiek2010 (changes requested):

settings/src/test/java/org/odk/collect/settings/support/SettingsUtils.kt
seadowg: Right. This was moved over but definitely a good opportunity to get switch over. Will do that.

settings/src/test/java/org/odk/collect/settings/support/SettingsUtils.kt
seadowg: Magic! I didn't know about the `step` extension.

collect_app/src/test/java/org/odk/collect/android/projects/ProjectDetailsCreatorTest.kt
seadowg: Yes! Good catch. I wanted to replace this with tests for the helper I pulled out (`itemFromHashOf`) but forgot. I'll add that in.

settings/build.gradle
seadowg: Actually we do need them for building instrumented APKs (you'll see CI is failing). Annoying.

shared/src/test/java/org/odk/collect/shared/collections/CollectionExtensionsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.junit.Test
+import org.odk.collect.shared.collections.CollectionExtensions.itemFromHashOf
+
+class CollectionExtensionsTest {
+
+    @Test
+    fun `itemFromHashOf returns item using hashcode as index`() {
```
seadowg, line 11: These tests check the hash based access works as expected and there are tests in `ProjectDetailsCreatorImplTest` that check that the colors used come from the passed colors list.

shared/src/test/java/org/odk/collect/shared/collections/CollectionExtensionsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        assertThat(listOf("0", "1").itemFromHashOf(HashCode(5)), equalTo("1"))
+    }
+
+    @Test
+    fun `itemFromHashOf works with negative hashcode`() {
+        assertThat(listOf("0", "1").itemFromHashOf(HashCode(-1)), equalTo("1"))
+    }
+}
+
+private data class HashCode(private val value: Int) {
```
grzesiek2010, line 28: Why do you need this class it seems like test work without it?

seadowg: It's being used in all the above tests. Do you mean we could write the test without it? My goal here was to use a class with a settable `hashcode` to make it clearer what's going on (rather than reverse engineering the implementation of an existing class.

grzesiek2010: I mean you implemented this class to make sure that for 0 hashcode returns 0 for 1 retruns 1 etc
but if you pass 0, 1, etc directly to `itemFromHashOf` the result will be the same because that's how hashcode works for integers. So if you remove this wrapper class your tests still pass.

seadowg: Ah right I didn't actually realize that `hashcode()` works like that for `Int`. Although, it makes a lot of sense! Thinking about it though, a problem  with using `Int` as the test value is that the test would pass with an incorrect implementation:

```kotlin
abs(any) % this.size
```

grzesiek2010: I can't change `abs(any.hashCode()) % this.size` to` abs(any) % this.size` it causes `None of the following functions can be called with the arguments supplied. `
Did you change anything else?

seadowg: Oh, sorry that will be because you can't call `abs` with `Any`! The bad implementation that would pass the test would be:

```kotlin
abs(any as Int) % this.size
```

grzesiek2010: Ok I see.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1 and 8.1.

Verified cases:
- add project using QR code
- different types of servers: Central, Ona, Kobo, Drive
- update settings in project using QR code
- reset settings 
- import settings from file
- save settings to file
- generate qr code with and without admin and server passwords
- add project duplicate 
- scan invalid qr code settings/add project
- scan inverted qr code settings/add project
- use barcode widget
- use front camera to scan barcode 

When testing I have noticed an issue that was also visible on the store version so reported in #5027

srujner:
**Tested with success!**  
Verified on Androids 10 and 11.

I was able to reproduce the issue mentioned in #5027 on both Androids

PR #4926, opened 2021-12-04
Display sub-meter accuracies using cm

dimwight:
Addresses #4198 (revises #4866)

See 'Simulation' and 'Testing' below.

See 'The issue' and 'The update' below.

None?

_[geowidgets.xlsx](
https://github.com/getodk/collect/files/7653693/geowidgets.xlsx
)_ used for development testing, see 'Simulation' below.

No.

### The issue
Several strings are used to display GPS accuracy:
- `gps_result` by `getGeoPointAnswerToDisplay` in `GeoWidgetUtils`
- `location_status_accuracy` etc by `updateUi` in `GeoPolyActivity`
- `location_accuracy` by `getAccuracyMessage` in `GeoPointActivity`, and by 
`formatLocationStatus` in `GeoPointMapActivity` 

### The update
Each format string now needs two variants, for formatting accuracy as _cm_ or _m_. So code using these strings has to look at the accuracy value so as to
- decide which format to use, and
- where the answer is _cm_, scale the value to match ie multiply by 100.

This is coded independently for `gps_result` and `location_status_accuracy` etc, with the complication that `getGeoPointAnswerToDisplay` which uses `gps_result` has to extract the accuracy value from its `answer` parameter. 

To avoid duplicating code for `location_accuracy`, `getAccuracyMessage` and `formatLocationStatus` now get their formatted string from `getAccuracyUnitString` in `GeoUtils`. 

Although `truncateDouble` is no longer called by `getAccuracyMessage` its logic is still required in a couple of places. So it has been made static and moved to `GeoUtils` as `truncateDoubleValue`, to avoid confusion with `truncateDouble` in `GeoWidgetUtils`. 

#### The strings
The [previous version](https://github.com/getodk/collect/pull/4866) of this PR replaced each existing string with a pair suffixed `_m` and `_cm`. This however implied a significant burden of localisation for a cosmetic change that will be visible to very few users in at least the medium term. 

For this revision the existing strings are retained for _m_ formatting, with new strings only for _cm_. The new strings are commented as `<!-- Low-priority for localisation -->`, given that the few users who encounter them will almost certainly understand the default. 

### Simulation
In addition to the formal testing described below, there was a need to simulate cm accuracy so as to confirm coverage of all occurrences in the UI. 

So for now at least `SIMULATED_ACCURACY` has been added to `GeoUtils`, for use when `simulateAccuracy` is set. It replaces real accuracy values to enable manual confirmation that _cm_ formatting is applied when appropriate (including a pretty specious mod to `FormEntryPromptUtils`). 

The form defined in _[geowidgets.xlsx](
https://github.com/getodk/collect/files/7653693/geowidgets.xlsx
)_ can be used to test with suitable values for `SIMULATED_ACCURACY`.

### Testing
The new code is tested by three new methods added to (mostly) existing classes, all named `locationAccuracyIsFormattedInAppropriateUnit` since they perform essentially the same task by iterating through the `TEST_ACCURACIES` added to `GeoUtils`. 

The variants in `GeoWidgetUtilsTest` and `GeoUtilsTest` check results for `gps_result` and `location_accuracy` as returned by their respective methods. In `GeoPolyActivityTestOld` the test is of the state of `location_status_accuracy` following a fake fix. 

The rationale for `GeoPolyActivityTestOld` is that the work for this PR was largely completed before the recent improvements to geolocation code. The updated `GeoPolyActivityTest` doesn't seem to lend itself to the test defined within the framework of the old class, so for the time being this has been partially reinstated as `…Old`.

Two further tests have been updated:

- The new implementation of `getGeoPointAnswerToDisplay` necessitates an update in  `GeoWidgetUtilsTest`to `getAnswerToDisplay…returnsAnswer`, which in fact only tests currently for accuracies below 1 m.

- `GeoPointMapActivityTest` has been updated for now to switch off simulation.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
@@ -1,7 +1,5 @@
 package org.odk.collect.android.widgets.utilities;
 
-import static org.odk.collect.geo.GeoUtils.SIMULATED_ACCURACY;
-import static org.odk.collect.geo.GeoUtils.simulateAccuracy;
 import static org.odk.collect.shared.strings.StringUtils.removeEnd;
 
 import android.content.Context;
@@ -29,10 +27,6 @@ public final class GeoWidgetUtils {
     public static String getGeoPointAnswerToDisplay(Context context, String answer) {
         try {
             if (answer != null && !answer.isEmpty()) {
-                if (simulateAccuracy) {
-                    answer = answer.replaceAll("\\b\\d+\\.\\d$",
-                            String.valueOf(SIMULATED_ACCURACY));
-                }
                 String[] parts = answer.split(" ");
                 //Guard against bad answer
                 if (parts.length < 4) {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
@@ -1,5 +1,7 @@
 package org.odk.collect.android.widgets.utilities;
 
+import static org.odk.collect.geo.GeoUtils.SIMULATED_ACCURACY;
+import static org.odk.collect.geo.GeoUtils.simulateAccuracy;
 import static org.odk.collect.shared.strings.StringUtils.removeEnd;
 
 import android.content.Context;
@@ -27,6 +29,10 @@ public final class GeoWidgetUtils {
     public static String getGeoPointAnswerToDisplay(Context context, String answer) {
         try {
             if (answer != null && !answer.isEmpty()) {
+                if (simulateAccuracy) {
+                    answer = answer.replaceAll("\\b\\d+\\.\\d$",
+                            String.valueOf(SIMULATED_ACCURACY));
+                }
                 String[] parts = answer.split(" ");
                 //Guard against bad answer
                 if (parts.length < 4) {
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
@@ -23,16 +23,22 @@ public final class GeoWidgetUtils {
 
     }
 
+    //Cm accuracy #4198
     public static String getGeoPointAnswerToDisplay(Context context, String answer) {
         try {
             if (answer != null && !answer.isEmpty()) {
                 String[] parts = answer.split(" ");
-                return context.getString(
-                        R.string.gps_result,
+                //Guard against bad answer
+                if (parts.length < 4) {
+                    return "";
+                }
+                double accuracy = Double.parseDouble(parts[3]);
+                boolean useCm = accuracy < 1;
+                return context.getString(useCm ? R.string.gps_result_cm : R.string.gps_result,
                         convertCoordinatesIntoDegreeFormat(context, Double.parseDouble(parts[0]), "lat"),
                         convertCoordinatesIntoDegreeFormat(context, Double.parseDouble(parts[1]), "lon"),
                         truncateDouble(parts[2]),
-                        truncateDouble(parts[3])
+                        useCm ? accuracy * 100 : truncateDouble(parts[3])
                 );
             }
         } catch (NumberFormatException e) {
--- a/collect_app/src/test/java/org/odk/collect/android/location/GeoPolyActivityTestOld.java
+++ b/collect_app/src/test/java/org/odk/collect/android/location/GeoPolyActivityTestOld.java
@@ -40,7 +40,7 @@ import org.robolectric.android.controller.ActivityController;
 import org.robolectric.shadows.ShadowApplication;
 
 @RunWith(AndroidJUnit4.class)
-public class GeoPolyActivityTest_Old {
+public class GeoPolyActivityTestOld {
 
     static {
         GeoUtils.simulateAccuracy = false;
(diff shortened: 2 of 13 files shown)


seadowg (changes requested):
Thanks for thinking about simulation here with `SIMULATED_ACCURACY`! I think for the moment I'd actually find it easier to review without that in there (just another thing to keep track of in my head). Could you take it out so I can review without it? I'm happier switching hacking the underlying values myself to see what it looks like.

dimwight:
Removed as requested - though in fact I dreamt it up because I couldn't see how to tweak the underlying value directly.

seadowg:
> though in fact I dreamt it up because I couldn't see how to tweak the underlying value directly

Fair enough! For future reference, pretty much all location are run through `LocationUtils.sanitizeAccuracy` so you could hack the code there to change the accuracy.

dimwight:
Aha! Indeed the neatest solution, had I but known.

strings/src/main/res/values/strings.xml
```diff
@@ -547,7 +551,11 @@
     </plurals>
     <string name="location_status_searching">Searching for location, please wait&#8230;</string>
     <string name="location_status_accuracy">Location accuracy: %.2f m</string>
+    <!-- Low-priority for localisation -->
```
seadowg, line 554: Is this helpful for translators @lognaturel? It's not something we've done before.

collect_app/src/test/java/org/odk/collect/android/location/GeoPolyActivityTestOld.java
```diff
@@ -0,0 +1,79 @@
...
+import org.odk.collect.android.support.CollectHelpers;
+import org.odk.collect.geo.GeoPolyActivity;
+import org.odk.collect.geo.GeoUtils;
+import org.odk.collect.location.LocationClientProvider;
+import org.robolectric.Robolectric;
+import org.robolectric.android.controller.ActivityController;
+import org.robolectric.shadows.ShadowApplication;
+
+@RunWith(AndroidJUnit4.class)
+public class GeoPolyActivityTestOld {
```
seadowg, line 43: I would avoid adding this test file as I think you have the important logic covered in `GeoUtilsTest`.

geo/src/test/java/org/odk/collect/geo/GeoUtilsTest.java
```diff
@@ -52,12 +59,27 @@ public void test_formatLocationResultString() {
...
-        assertEquals("network", GeoUtils.capitalizeGps(locationProvider));
+        assertEquals("network", capitalizeGps(locationProvider));
 
         String nullLocationProvider = null;
-        assertNull(GeoUtils.capitalizeGps(nullLocationProvider));
+        assertNull(capitalizeGps(nullLocationProvider));
+    }
+
+    @Test
+    //Cm accuracy #4198
```
seadowg, line 72: I think it's best to avoid comments that reference issues like this. Better to do this in the git commit description when adding the code. Let's remove the other `//Cm accuracy #4198` comments as well.

geo/src/test/java/org/odk/collect/geo/GeoUtilsTest.java
```diff
@@ -52,12 +59,27 @@ public void test_formatLocationResultString() {
...
+    public void locationAccuracyIsFormattedInAppropriateUnit() {
+        final Context context = ApplicationProvider.getApplicationContext();
+        for (double accuracy : TEST_ACCURACIES) {
```
seadowg, line 75: I think instead of using a "table test" style here it might be simpler to have separate tests for metres and centimetres cases (`getAccuracyUnitString_returnsAccuracyInMeters` and `getAccuracyUnitString_whenAccuracyBelowOne_returnsAccuracyInCentimetres`). I think you only need to test one value for each unless I'm missing some edge cases.

Also, as a general rule we'd want to keep the test fixtures (`TEST_ACCURACIES` here) in test code rather than our actual app code. I think I'd prefer something like that to just live in the test class it's used in.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -23,16 +23,22 @@ private GeoWidgetUtils() {
...
 
+    //Cm accuracy #4198
     public static String getGeoPointAnswerToDisplay(Context context, String answer) {
         try {
             if (answer != null && !answer.isEmpty()) {
                 String[] parts = answer.split(" ");
-                return context.getString(
-                        R.string.gps_result,
+                //Guard against bad answer
+                if (parts.length < 4) {
```
seadowg, line 32: I don't think we expect this at all and probably want a non-silent failure in the case that it did happen. We can take this check out for the moment.

geo/src/test/java/org/odk/collect/geo/GeoUtilsTest.java
seadowg: Same goes for `GeoWidgetUtilsTest`

seadowg (changes requested):
I've left a few comments on some tweaks I think we should make but on a grander scale, I'm wondering if we could restructure the strings a little so that this only needs to be tested and implemented in one place. How about a `formatAccuracy` helper (could be on `GeoUtils` potentially) that takes an accuracy and spits out `x m` or `x cm`? This could be shared between all the different places that we display accuracy and the strings would just format in the already formatted accuracy instead of just the number (also reducing the total number of strings). We'd need more translation effort yes, but I think we're generally happy with that if it gives us simpler code.

strings/src/main/res/values/strings.xml
lognaturel: I don’t think so. Active languages have many translators and an additional string is not a big deal!

dimwight:
> …How about a `formatAccuracy` helper…that takes an accuracy and spits out `x m` or `x cm`? 

I can't see how this would work for RTL locales as here. 
arabic

seadowg:
@dimwight could the helper not use a translated string to handle that? I'm just trying to get away from having duplicated formatting logic for the accuracy value.

seadowg:
Given recent changes, I think support cm accuracies can now be achieved by simply updating the implementation of `GeoUtils#formatAccuracy`. Does that seem right @dimwight?

seadowg:
I'm going to close this for the moment but feel free to reopen if you want to come back to it.

Issue #4927, opened 2021-12-07
Demo project is duplicated after updating app


kkrawczyk123:
#### Software and hardware versions 
Collect v2021.3 Android all device used...

#### Problem description
When testing Collect's update I have noticed that demo project is duplicated when it's used before updating app.
![ezgif com-gif-maker (14)](https://user-images.githubusercontent.com/33342488/145028829-7828144b-7755-4f24-ab56-27551797efa1.gif)

#### Steps to reproduce the problem
1. Install Collect ex. v2021.3.1
2. Use demo project
3. Update Collect to v2021.3.2
4. See that you have 2 demo projects

#### Expected behavior
Projects should not be duplicated when updating the app. 

#### Other information 
I have tried also app update when using other project - demo project is also created but it's not a duplicate then.

seadowg:
I think I can see what's going on here. Basically if you install a version of Collect with Projects (v2021.2+) and then upgrade, Collect will attempt to create a project out of the "existing one" (one that would exist in pre v2021.2) and because there won't be one, we end up creating a "Demo project" (this is to do with some small details in `ProjectDetailsCreator`). 

I'm going to confirm that I'm definitely right about all this, but it looks like we probably need to make our upgrade code smarter to avoid this - we not only want to just run the existing project migration once, but we also don't want to run it when upgrading from versions that don't need to do it.

seadowg:
Ok yeah it really looks like the above is all true. Annoyingly I don't think there is a way for us to work out which project has been added by this problem (correct if I'm wrong here @grzesiek2010), so it's probably best we announce this on the forum and/or on social media when we have a fix.

I'll start working on a hot fix now.

PR #4928, opened 2021-12-07
Mark "run once" upgrades as run on fresh install

seadowg:
Work towards #4927 (we'll need to do another PR to fix on `master`)

This can be a little hard to get your head around. Basically, we have a way of saying upgrades should only run once, but we never considered the fact that we also didn't want those upgrades to ever run if they exist during a fresh install. This changes our upgrade process so that we mark "run once" upgrades (currently just the existing project migration) as done when doing a fresh install.

New tests.

I also considered cleaning up the extra project created due to the bug but couldn't find a way to identify it safely.

Should just fix the bug! Good to check upgrades (from different versions) and fresh installs. As @grzesiek2010 pointed out, installing a previous version and then this version from source will not simulate an upgrade as the version code used will be the same (it's based on the number of commits on the `master` branch). To simulate an upgrade:

1. Check out the last released version (`v2021.3.2`)
2. Set `versionCode` in `collect_app/build.gradle` to `1`
3. Install on device (make sure to uninstall first)
4. Check out the PR branch again
5. Set `versionCode` in `collect_app/build.gradle` to `2`
6. Install on device

This simulates a "real" upgrade where the version code would change as well as the code.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/ExistingProjectMigrator.kt
@@ -34,6 +34,11 @@ class ExistingProjectMigrator(
     }
 
     override fun run() {
+        if (settingsProvider.getMetaSettings().contains(MetaKeys.LAST_LAUNCHED)) {
+            // We're upgrading from a version with Projects so shouldn't be running this
+            return
+        }
+
         val generalSharedPrefs = PreferenceManager.getDefaultSharedPreferences(context)
 
         val newProject = projectDetailsCreator.createProjectFromDetails(
--- a/collect_app/src/test/java/org/odk/collect/android/projects/ExistingProjectMigratorTest.kt
+++ b/collect_app/src/test/java/org/odk/collect/android/projects/ExistingProjectMigratorTest.kt
@@ -7,8 +7,11 @@ import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.matcher.ViewMatchers.assertThat
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.hamcrest.Matchers.`is`
+import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.odk.collect.android.application.initialization.upgrade.AppUpgrader
+import org.odk.collect.android.application.initialization.upgrade.Upgrade
 import org.odk.collect.android.injection.DaggerUtils
 import org.odk.collect.android.preferences.keys.MetaKeys
 import org.odk.collect.android.preferences.keys.ProjectKeys
@@ -27,6 +30,12 @@ class ExistingProjectMigratorTest {
     private val settingsProvider = component.settingsProvider()
     private val currentProjectProvider = component.currentProjectProvider()
     private val rootDir = storagePathProvider.odkRootDirPath
+
+    @Before
+    fun setup() {
+        // Should not run when upgrading from version with LAST_LAUNCHED
+        settingsProvider.getMetaSettings().remove(MetaKeys.LAST_LAUNCHED)
+    }
 
     @Test
     fun `creates existing project with details based on its url`() {
@@ -166,6 +175,26 @@ class ExistingProjectMigratorTest {
     @Test
     fun `has key`() {
         assertThat(existingProjectMigrator.key(), `is`(MetaKeys.EXISTING_PROJECT_IMPORTED))
+    }
+
+    /**
+     * Versions of the app before v2021.3.3 did not include a call to [AppUpgrader.install] and so
+     * this [Upgrade] would get incorrectly run on upgrades from versions after
+     * [AppUpgrader] and Projects (along with LAST_LAUNCHED) were added but before the call
+     * to [AppUpgrader.install] was introduced.
+     */
+    @Test
+    fun `does not run if previous app version already had projects`() {
+        PreferenceManager
+            .getDefaultSharedPreferences(context)
+            .edit()
+            .putString(ProjectKeys.KEY_SERVER_URL, "https://my-server.com")
+            .apply()
+
+        settingsProvider.getMetaSettings().save(MetaKeys.LAST_LAUNCHED, 3115)
+
+        existingProjectMigrator.run()
+        assertThat(projectsRepository.getAll().size, `is`(0))
     }
 
     private fun getProjectDirPaths(projectId: String): Array<String> {


lognaturel:
I will forever be shocked at what users don’t report. Glad you found it, @kkrawczyk123! Looks good to me so I’ll mark as needing testing for @getodk/testers to start right away. Would still be good for you to take a carefully look, @grzesiek2010. 🙏

yanokwa:
I just got a report about this from an ODK Cloud user a few hours ago. So they do report things, but @kkrawczyk123 is a lot faster and more reliable!

grzesiek2010:
I wanted to dig deeper to fully understand what is happening but I can't seem to reproduce the issue. I tried on two emulators and one real device those steps:
1. Install v2021.3.1
2. Select Demo project
3. Install v2021.3.2
4. And I have just on project

@seadowg did you have to do anything else? I talked to @kkrawczyk123 but she claims that those steps are enough in her case.

srujner:
@grzesiek2010  I was able to reproduce the issue using the steps provided by @kkrawczyk123 on Android 10 and 11.

seadowg:
@grzesiek2010 those are the steps I followed. An extra subtlety is that you need to do a fresh install of v2021.3.1 - although you should be able to reproduce with any version from v2021.2 onwards really. What's happening in terms of the repro steps is:

1. Install v2021.3.1 - app starts up and does not run `AppUpgrader.upgrade` as we're doing a fresh install
2. Select Demo project - demo project is setup with
3. Install v2021.3.2 - app starts up and runs `AppUpgrader.upgrade` which ends up running `ExistingProjectMigrator` as it was never run before

We end up with a project called "Demo project" because the legacy general and admin settings are empty so `ProjectDetailsCreator` gives us that project name due to us using the default URL (see `ProjectDetailsCreator#getProjectNameFromConnectionIdentifier`).

grzesiek2010:
Ok I was finally able to reproduce the issue...

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
```diff
@@ -69,6 +69,8 @@ public void initialize() {
     private void performUpgradeIfNeeded() {
         if (launchState.isUpgradedFirstLaunch()) {
             appUpgrader.upgrade();
+        } else {
+            appUpgrader.install();
```
grzesiek2010: Doesn't it mean that if you have v2021.3 and install v2021.4 (assuming we are going to release it with this fix) it will still go and perform `appUpgrader.upgrade()` first and run those migrations?

grzesiek2010: I'm pretty sure I'm right...

seadowg: Errr maybe? Just going to try and write this out to get it straight in my head:

* Fresh install of 2021.3.2 (or any version with projects) and upgrade to 2022.1: `upgrade()` runs without ever running `install()` so they get an extra project 👎
* Fresh install of 2021.3.3 (next hot fix) and upgrade to 2022.1: `upgrade()` runs but after `install()` on fresh install 👍
* Upgrade to 2021.3.3 from 2021.3.2 with previous upgrade from 2021.3.1 (or any version with projects): `upgrade()` runs but after `ExistingProjectMIgrator` has already run on previous upgrades 👍

So yeah I think you're correct that this hasn't quite fixed the problem. If we'd included `install()` in our original release with `ExistingProjectMigrator` everything would work but as you point out, we need to account for the existing bad state. 

I think to correct the problem we need to run `install()` before `upgrade`() when upgrading from a version before 2021.3.3 (our next hot fix) and after 1.30.1 (all versions after this have projects and `ExistingProjectMigrator`). Does that sound right? We'll need to work out version codes for these as well.

grzesiek2010: Can't you just base your fix on `LAST_LAUNCHED` value that we keep in meta settings? If that value represents a version that already had projects the migration should be skipped.

seadowg: Yeah sorry I should have worded that differently, calling `install()` is a bit over the top. How about: "stop `ExistingProjectMigrator` from running when upgrading from a version after 1.30.1 (using version code to determine the last version)". Or actually, even simpler: "stop `ExistingProjectMigrator` running when upgrading from a version without `LAST_LAUNCHED`" as `LAST_LAUNCHED` was introduced as part of 2021.2.

grzesiek2010: If `LAST_LAUNCHED`  was released in 2021.2 it won't be a perfect fix because still upgrading from 2021.0 and 2021.1 will be broken but that doesn't seem to be a big deal. Alternatively you can think of another solution like for example:
`If projects repository is not empty that means we do upgrade from a version with projects and can skip migrations.`

grzesiek2010:
Another thing: I'm pretty sure the QA team is not able to verify this fix correctly because the flow requires installing an apk with a newer version on an existing one. If we build apks on our own no matter from which commit we start the version will be always the same because it's calculated in gradle:
`LEGACY_BUILD_NUMBER_OFFSET + getMasterCommitCount()`
where `LEGACY_BUILD_NUMBER_OFFSET` is a static number 1067 and `getMasterCommitCount()` it's a git command
`git rev-list --first-parent --count master` which will return the same number no matter on what branch you are because it takes master and your master branch is always in the same state.
That means we always end up with the same version number and upgrading won't take place at all.
To reproduce the issue you used released apks and there  those versions differ of course but we can't install our builds on release builds.

So to be able to verify the fix correctly the QA team would need an additional commit with bumped version number. Am I right?

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.java
seadowg: There's no 2021.0 or .1. 1.30 was the last version before projects and 2021.2 first version with them. We **want** upgrading from 1.30 (and before) to run `ExistingProjectMigrator` and there is already the upgrade `key` concept that will take care of making sure that the upgrade doesn't run again after that, so I think we're fine on older versions.

grzesiek2010: > There's no 2021.0

Right my mistake then ok it will be fine.

seadowg:
> Another thing: I'm pretty sure the QA team is not able to verify this fix correctly because the flow requires installing an apk with a newer version on an existing one.

@grzesiek2010 great point. I'm taking `needs testing` off this right now anyway as we're pretty sure the fix isn't good enough (cc/ @sruner and @kkrawczyk123). You're right that to simulate a true upgrade we're going to need to change `versionCode` in `collect_app`/`build.gradle`. I'll work on improving this fix based on my conversation with @grzesiek2010 and also post a workflow for simulating the upgrade.

seadowg:
Ok that's hopefully in a better place now. I've also included steps to simulate an upgrade in the PR description. Will wait for approval to mark this ready for testing.

collect_app/src/main/java/org/odk/collect/android/application/initialization/upgrade/AppUpgrader.kt
```diff
@@ -16,6 +16,16 @@ class AppUpgrader(private val metaSettings: Settings, private val upgrades: List
             }
         }
     }
+
+    fun install() {
```
grzesiek2010: I'm not sure if I understand the purpose of keeping it... if you keep that and run in v2021.3.3 that is going to set flags for all migrators to true (do not perform in the future). Maybe not in the first run when you upgrade from an older version but then if you kill the app and open again it will happen.

Let's say we have a user who uses v2021.3.3 so all those flags are set to true on his device.
Then in v2022.1 you decide to rename one of our preferences (`ExistingSettingsMigrator` is responsible for that).
Our user upgrades form v2021.3.3 to v2022.1 but his `ExistingSettingsMigrator` is not trigger because the flag of this migrator was set to true.

Am i missing something?

grzesiek2010: The thing is that we have two types of migrators:
- final migartors
- editable migrators

`ExistingProjectMigrator` is final if it was performed once we shouldn't need it anymore so setting it's flag to true is fine.
`ExistingSettingsMigrator` I would say it's not final we might need it every time we decide to update some prefs
`FormUpdatesUpgrade` here I'm not sure I would need to dig depper.

so you can't disable all of them unless I'm missing something.

seadowg: `Upgrades` can only be skipped if they have a non `null` `key()` (you can see in `AppUpgrader` or its tests). This lets us run `ExistingSettingsMigrator` and `FormUpdatesUpgrade` every upgrade and only run `ExistingProjectMigrator` once. Calling `install()` will not cause the first two to be skipped as they return `null` from their `key()` implementations so can't be marked as already having been run. We currently don't **need** the call to `install()` because we've put a check in `ExistingProjectMigrator` but it lets us add another "run once" (or final as you say) upgrade. Basically, the upgrade logic wasn't correct before but now is in my opinion.

grzesiek2010: Ok now I see but the way it woks (with those keys null or nonnull) is confusing...
So this will be run only for `ExistingProjectMigrator` so why is it important if we do the check inside `ExistingProjectMigrator` either way?

seadowg: Ok for this PR we can just take out the `install()` call as I think it's confusing things. I do want this to go in `master` but we can discuss that separately as improvements to the way upgrades work rather than just a fix for this specific issue.

grzesiek2010 (approved):

lognaturel:
Thanks for really thinking through it, @grzesiek2010!

srujner:
**Tested with success!**
Verified on Androids: 10, 11

Verified cases:
- Update with only demo project created
- Update with multiple projects created
- Update without demo project 
- Update without any projects added
- Changed settings before updating the app
- Adding a project with qr code and then updating the app
- Update from .apk and Android studio

kkrawczyk123:
Verified also on Androids: 5.1, 8.1 and 9.

PR #4929, opened 2021-12-07
Notification improvements

### LINKED ISSUE #4787: Improve Match Exactly notification experience
Match Exactly notifications currently have two problems we've identified:

1. Clicking a failure notification will open the "Fill Blank Form" screen for the current project, which might not be the project that experienced the error
2. Failures happen pretty frequently - just under 20% syncs will result in an failure notification ([here](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/events/~2Freport~2FMatchExactlySyncCompleted%3Ft%3D1628690834457&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_r..eventId%253DMatchExactlySyncCompleted&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1))

To alleviate these, we should make the following changes:

- [x] The notification should specify the project that failed to sync (clicking it should still just take the user to "Fill Blank Form" for now) in the subtext (fixed in https://github.com/getodk/collect/pull/4929)
- [x] Instead of showing a notification immediately, a failure (from any request, including form downloads) should result in a limited number of retries which is limited to a certain number of retries (most likely using `WorkManager`'s [Retry and Backoff](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/define-work#retry_and_backoff_policy) and `WorkerParameters#getRunAttemptCount`) followed by a failure notification if the sync still hasn't succeeded
- [x] Update analytics calls to only measure failures that the user sees

We're not doing a lot to fix problem **1** here, but my thinking is we try and fix problem **2** as much as possible to make **1** less of a common experience. If we're right in that thinking, we should see the proportion of **visible** failures drop.

This can definitely be solved as multiple PRs.

### LINKED ISSUE #4841: Form update notification from not-used project displays while using another project
Collect v store version, master Android vx.x.x, device used...

#### Problem description
When testing Central I have noticed that I am obtaining form update notifications from another project than one that I am using. 

1. Scan draft testing qr code from Central as project1
2. Publish form on Central
3. Scan other qr code from Central as project2, this time legacy qr code so an automatic form update is disabled
4. Stay on 2nd project 
5. Wait ~ 15min
I am getting a form update failed notification from another project than the one that I am using.

#### Expected behavior
User should only see notifications from the currently used project.

### LINKED ISSUE #4844: Auto update should have the same success/failure flow as Get Blank Form
Currently, clicking the notification shown after an auto update (Previously Downloaded only with auto downloads enabled) will just show the old error UI (a dialog with all the different errors inline).

Instead, we should have the notification link to our new error flow:

* The notification text should change - the title should be "Form download succeeded"/"Form download failed" and the text should use the same style as the Get Blank Form flow ("All downloads succeeded!"/"2 of 3 forms failed to download")
* Clicking the notification should open Fill Blank Form if the downloads succeed
* Clicking the notification should show the new errors list screen if some downloads fail
* Notifications should include project name as subtext

### LINKED ISSUE #4884: Revise submit notifications
Auto and background manual submits should have the new error flow as part of their notifications.

Success notification:
* Should have [subtext](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setSubText(java.lang.CharSequence)) with project name
* Title text should be "All uploads succeeded!"
* Clicking takes user to Main Menu

Failure notification:
* Should have [subtext](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setSubText(java.lang.CharSequence)) with project name
* Title text should be "<failures> of <total uploads> uploads failed!
* Clicking takes user to Send Finalized Form
* Additional action "Show details" takes user to errors screen with errors

grzesiek2010:
Closes #4844 Fully fixed by commits 1-13
Closes #4787 Partially fixed (only the first checkbox) by commits 14-17
Closes #4841 Fully fixed by commits 18-21
Closes #4884 Partially fixed by commits 19-32 It's a part of [Revise submit notifications](https://github.com/getodk/collect/issues/4884) so for now I just added project name as a subtext. I also updated the displayed titles to match what we have for `auto download` (so they differ a bit comparing with what we have in the issue). The last thing of note is that for now I decided to display the failure message in the notification because current implementation of auto submit does return just one message like in case of match exactly so there is nothing we could display in our new `ErrorActivity`. This will be improved later.

I tested the improvements manually and added automated tests.

This is what we we have discussed earlier. Our notification should be consistent and better support projects that we have introduced this year.

| Task | Success | Failure |
| Updates check |   | N/A |
| Match exactly | N/A |   |
|  Auto update  |  |  |
| Auto submit |   |  |

Please verify six types of notifications that we support. Nothing else should be affected.

No.

No.


seadowg (commented):
Really nice to be cleaning this all up in one go! Some high level notes for discussion after having a look through:

* I like the notification builder concept. Nice to separate the display logic from the interactions with `NotificationManager` :+1: 
* I think the builders could just be constructed inside `NotificationManager` rather than having to be passed in. :heavy_check_mark: 
* Could `FormsDownloadResultInterpreter` just be a series of helper functions/methods on an `object`? :heavy_check_mark: 
* I'm not sure about the `androidTest` `...NotificationTest` clases. It feels like the flows (clicking on the notification etc) should be (and in some cases maybe is) tested in the `PreviouslyDownloadedOnlyTest` and other `formmanagement` tests and the view details could be tested in unit tests for the new notification builder objects. That way the behaviour could be tested without knowledge of the `Notifier` etc. :heavy_check_mark:

grzesiek2010:
> I think the builders could just be constructed inside NotificationManager rather than having to be passed in.

Because those builders are used only there and also tests only via that manager? Yeah my brain forces me to pass everything via constructors but in this case I could agree.

> Could FormsDownloadResultInterpreter just be a series of helper functions/methods on an object?

yeah, done.

> I'm not sure about the androidTest ...NotificationTest clases. It feels like the flows (clicking on the notification etc) should be (and in some cases maybe is) tested in the PreviouslyDownloadedOnlyTest and other formmanagement tests and the view details could be tested in unit tests for the new notification builder objects.

Yes we already have some tests that verify notifications (as a part of a bigger flow) but not everything is tested there. For example on master we have a crash when a user clicks on a notification that uses `NotificationActivity` and there is no test for that. So I wanted to make sure every notification has its own small test that fully verify its content + opening an activity after clicking on it.

seadowg:
> Because those builders are used only there and also tests only via that manager? 

Yeah exactly! Right now, the builder feels like nice clean-up for the implementation of the `NotifiationManager` but doesn't really need to be pulled out as a separate dependency.

>  Yes we already have some tests that verify notifications (as a part of a bigger flow) but not everything is tested there. For example on master we have a crash when a user clicks on a notification that uses NotificationActivity

I'd still lean towards testing that though our feature tests. For me the advantage of that would be that the test doesn't need to know about the `NotificationManager` etc and is just testing the notification as part of the user's flow.  The notification tests feel like they end up with a slightly strange set of test "boundaries" as they start with a calling a method (not a user interaction) but are checking the flow of clicking on a notification (a user interaction). For instance, instead of writing `FormUpdatesDownloadedNotificationTest`, I think we could add the cases covered there to `PreviouslyDownloadedOnlyTest` (same goes for the other notification types, and they're relevant features). As I said before, I don't think we need to test all the details of the displayed notification at that level as we could write unit tests for the new builder classes. Does that make sense?

seadowg:
@grzesiek2010 is this ready for another review?

grzesiek2010:
> is this ready for another review?

the last point from you comments (about tests) is not done yet I will try to finish it today but it it is the only thing we can even mark it as `needs testing` I guess.

grzesiek2010:
@seadowg ok now I fixed tests too improving the existed tests and removed those you didn't like.

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        deferredTasks = testDependencies.scheduler.deferredTasks
+
+        assertThat(deferredTasks.size, `is`(1))
+        assertThat(deferredTasks[0].tag, `is`(previouslyDownloadedTag))
+        assertThat(deferredTasks[0].repeatPeriod, `is`(1000L * 60 * 60))
+    }
+
+    @Test
+    fun whenPreviouslyDownloadedOnlyEnabled_checkingAutoDownload_downloadsUpdatedForms_andDisplaysNotification() {
```
seadowg: The auto download tests were here already but looking at it now it feels like they might be better places in `PreviouslyDownloadedOnlyTest`. I'm not sure though. What do you think?

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyComponent.java
```diff
@@ -322,4 +326,6 @@
     Scheduler scheduler();
 
     LocationClient locationClient();
+
+    Notifier notifier();
```
seadowg: I don't think this is being used anymore

seadowg (approved):
Looks great. Couple of potential tweaks inline but nothing that blocks this.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyComponent.java
grzesiek2010: Right, fixed.

collect_app/src/androidTest/java/org/odk/collect/android/feature/settings/FormManagementSettingsTest.kt
grzesiek2010: Yeah I think it makes sense, done.

kkrawczyk123:
**Tested with success!**

Verified cases:
- autosend from settings success message
- autosend failure message
- autosend defined in form
- Central/Google Drive
- check for form update: redirect to get blank form 
- automatic download: redirect to fill blank form 
- automatic download failure
- match exactly failure: redirect to fill blank form
- crashes not visible after taping on notifications
- select check for update in 1 project and open another project: notification is visible, redirect to get blank form in the current project where there is nothing to update/download is visible - I am not a fan of this solution and I think it can be confusing. Maybe we should reconsider that in the future? @lognaturel @grzesiek2010 @seadowg

seadowg:
> select check for update in 1 project and open another project: notification is visible, redirect to get blank form in the current project where there is nothing to update/download is visible - I am not a fan of this solution and I think it can be confusing. Maybe we should reconsider that in the future?

Yeah, it's definitely a problem with all the notifications. We don't currently auto switch between projects for any notifications or shortcuts, although I think we're getting to a place where we want to start looking at that as we now have almost a third of users using multiple projects on device (looking at our `ProjectsCount` user property [here](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/overview/~2F%3Ft%3D1642074647152&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..comparisons%253D%255B%257B%2522name%2522%253A%2522All%2520Users%2522%252C%2522filters%2522%253A%255B%257B%2522isCaseSensitive%2522%253Atrue%252C%2522expression%2522%253A%25220%2522%252C%2522fieldName%2522%253A%2522audience%2522%257D%255D%257D%252C%257B%2522name%2522%253A%2522ProjectsCount%2520includes%25201%2522%252C%2522filters%2522%253A%255B%257B%2522fieldName%2522%253A%2522userProperty%253A2%2522%252C%2522expressionList%2522%253A%255B%25221%2522%255D%252C%2522isCaseSensitive%2522%253Atrue%257D%255D%257D%252C%257B%2522name%2522%253A%2522ProjectsCount%2520includes%25200%2522%252C%2522filters%2522%253A%255B%257B%2522fieldName%2522%253A%2522userProperty%253A2%2522%252C%2522expressionList%2522%253A%255B%25220%2522%255D%252C%2522isCaseSensitive%2522%253Atrue%257D%255D%257D%255D%2526_u.dateOption%253Dlast90Days%2526_u.comparisonOption%253Ddisabled&cs%3Dapp.m.dashboard.overview&g%3D1)*\**). I'll add an issue (#4964) to cover this that we can discuss later.

*\* As a quick side note I think we should change our `ProjectsCount` user property to just be `UsingMultipleProjects` and only set it to `true` or `false`. Personally, that's what I want to know (the uptake of projects), and it's a little awkward to get to from the way we have it now. @lognaturel @grzesiek2010 what do you think?*

kkrawczyk123:
Actually, I have noticed one small issue. When open autosend success notification Main menu is opened twice. The issue was not visible before as success dialog was displayed in front of the main menu. 

reproduction steps:
1. set autosend in settings
2. fill submission
3. tap on notification
4. tap on devices back button

see gif:
ezgif com-gif-maker (16)

@grzesiek2010 Can you take a look at that? Or should I open a new issue?

grzesiek2010:
> Can you take a look at that? Or should I open a new issue?

I will look into it. I think it won't be a problem to fix it so if I'm right I'll do that but if not it shouldn't block us we can create a separate issue too.

grzesiek2010:
> * As a quick side note I think we should change our ProjectsCount user property to just be UsingMultipleProjects and only set it to true or false. Personally, that's what I want to know (the uptake of projects), and it's a little awkward to get to from the way we have it now. @lognaturel @grzesiek2010 what do you think?

I don't have any strong preferences so whatever works for you.

> Actually, I have noticed one small issue. When open autosend success notification Main menu is opened twice. The issue was not visible before as success dialog was displayed in front of the main menu.

It should be fixed. You might need to reinstall the app to make it work properly.

srujner:
**Tested with success!**

@grzesiek2010 Also the issue @kkrawczyk123 mentioned before is no longer reproducing after the fix

PR #4930, opened 2021-12-08
Extra analytics for `geopoint`

seadowg:
This adds analytics to track users behaviour when saving a location using the geopoint dialog. It tracks these outcomes:

* The user saves a geopoint within 2 seconds
* The user saves a geopoint  after 2 seconds
* The geopoint is saved automatically because the accuracy threshold was matched

Both manual cases include whether the accuracy is `acceptable`, `poor` or `unacceptable`.

### DIFF
--- a/geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
+++ b/geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
@@ -207,16 +207,7 @@ public class GeoPointActivity extends LocalizedActivity implements LocationListe
                 (dialog, which) -> {
                     switch (which) {
                         case DialogInterface.BUTTON_POSITIVE:
-                            if (System.currentTimeMillis() - startTime < 2000) {
-                                Analytics.log(AnalyticsEvents.SAVE_POINT_MANUAL_IMMEDIATE);
-                            } else if (location.getAccuracy() > 100) {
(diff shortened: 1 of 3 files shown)


geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
```diff
@@ -205,6 +207,16 @@ private void setupLocationDialog() {
                 (dialog, which) -> {
                     switch (which) {
                         case DialogInterface.BUTTON_POSITIVE:
+                            if (System.currentTimeMillis() - startTime < 2000) {
+                                Analytics.log(AnalyticsEvents.SAVE_POINT_MANUAL_IMMEDIATE);
+                            } else if (location.getAccuracy() > 100) {
```
seadowg: We could separate this from the `if-else` check so we could, for example, fire a `SAVE_POINT_MANUAL_IMMEDIATE` and a `SAVE_POINT_MANUAL_UNACCEPTABLE` event, but I think the two-second cutoff makes it unlikely that the user will even have had time to check the accuracy. I think I like that here `SAVE_POINT_MANUAL_IMMEDIATE` is for when the user doesn't realize they should wait and `SAVE_POINT_MANUAL_UNACCEPTABLE` etc is for when they are accepting a bad value.

lognaturel (approved):
Thanks!

Issue #4931, opened 2021-12-08
Manual downloads shows as failing when language not set to English


seadowg:
#### Software and hardware versions 
Collect v2021.3.2

#### Problem description
Downloads via "Get Blank Form" will always show as failures if the language is not set to English. The error list will show each form download as having the status "Success" however.

#### Steps to reproduce the problem

1. Set language to anything German (or any language we have translations for)
2. Navigate to Get Blank Form
3. Download one or more forms

We see the "x of y downloads failed!" message in German and can click through to the error list to see each form download was a "Success".

#### Expected behavior

We should see the "All downloads succeeded!" message when all downloads succeed.

PR #4932, opened 2021-12-08
Use localized strings in `DownloadFormsTask`

seadowg:
Work towards #4931 (we'll need to do another PR to fix on `master`)

Verified manually. I considered writing a test but decided not to - I think instead of cherry picking this onto `master` we should look at a more thorough fix that removes the use of a translated string in our download results logic (switch to a boolean or a nullable exception would be an improvement). We might want to introduce a smoke or regression test that checks the basic app flows in a second language, but we can consider that when looking at a proper fix.

Should just fix the bug. Good to check form downloads English and at least one other language. It'd also be good to add a case to our set of manual regression checks to carry out at least form download, entry and send in a non-English language so we have this covered in future.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
@@ -14,6 +14,8 @@
 
 package org.odk.collect.android.tasks;
 
+import static org.odk.collect.strings.localization.LocalizedApplicationKt.getLocalizedString;
+
 import android.os.AsyncTask;
 
 import org.odk.collect.android.R;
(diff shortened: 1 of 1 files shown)


grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
verified on Androids: 5.1, 8.1 and 9.0.

Tested cases:
- download forms in different languages: German, Spanish, French, RTL, English
- download forms from different servers: Central, Aggregate, Google Drive, Kobo, Ona
- send submissions to different servers
- download all forms
- error on failed download 

srujner:
**Tested with success!**

Verified also on Androids: 10, 11

PR #4934, opened 2021-12-09
Stop running `ExistingProjectMigrator` incorrectly

### LINKED ISSUE #4927: Demo project is duplicated after updating app

seadowg:
Closes #4927

This brings over the fix made in #4928 to `master` and also restructures the upgrade code to prevent similar issues in the future and to isolate it from the rest of the app (in a new module).

This also includes a bunch of changes to our CI JVM configuration as the other changes in this PR for some reason aggravated some OOM problems we'd seen before.

New tests.

Comments inline.

Should just fix the issue. Good to check upgrades as it's hard to write tests that cover a full app upgrade.


collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModelTest.kt
```diff
@@ -8,21 +8,18 @@ import org.mockito.Mockito.`when`
...
 import org.odk.collect.android.preferences.keys.ProjectKeys
 import org.odk.collect.android.preferences.source.SharedPreferencesSettings
-import org.odk.collect.android.utilities.LaunchState
 import org.odk.collect.projects.Project
 import org.odk.collect.projects.ProjectsRepository
 
 class SplashScreenViewModelTest {
 
     private lateinit var generalSettings: SharedPreferencesSettings
-    private lateinit var launchState: LaunchState
```
seadowg, line 18: Not sure why this was here. Probably in use at some point.

upgrade/src/main/java/org/odk/collect/upgrade/AppUpgrader.kt
```diff
@@ -0,0 +1,50 @@
...
+
+                if (key == null) {
+                    it.run()
+                } else if (!settings.getBoolean(key)) {
+                    it.run()
+                    settings.save(key, true)
+                }
+            }
+        } else {
+            upgrades.forEach {
```
seadowg, line 44: Doing this protects us from running into the problems we ran into with `ExistingProjectMigrator` with any other "run once" upgrades we add in the future. We don't need it at the moment, but I wanted to include the functionality now while we're thinking about it as it'll be easy to forget about later.

upgrade/src/main/java/org/odk/collect/upgrade/Upgrade.kt
```diff
@@ -0,0 +1,13 @@
+package org.odk.collect.upgrade
+
+interface Upgrade {
+
+    /**
+     * If this returns a non-`null` value, the [Upgrade] will only be run once (the next app
+     * upgrade). [Upgrade] implementations that return `null` from [Upgrade.key] will be run on
+     * every app upgrade.
+     */
+    fun key(): String?
```
seadowg, line 10: @grzesiek2010 hopefully this makes more sense now? If not, let's try and work out something we both like!

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -91,7 +91,7 @@ object Dependencies {
     const val okhttp3_mockwebserver = "com.squareup.okhttp3:mockwebserver:${Versions.okhttp3}"
     const val kotlinx_coroutines_test = "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.2"
     const val hamcrest = "org.hamcrest:hamcrest:2.2"
-    const val robolectric = "org.robolectric:robolectric:4.5.1"
-    const val robolectric_shadows_multidex = "org.robolectric:shadows-multidex:4.5.1"
+    const val robolectric = "org.robolectric:robolectric:4.7.3"
```
seadowg: This fixes some memory leaks caused by Robolectric

lognaturel (approved):
I'll admit that I'm still warming up to everything becoming a micro-framework! The balance between isolation and not getting lost in abstraction/many layers is tricky. I mostly focused on the tests to get confidence that the layers do what they're supposed to. I read through the commits and the code. All looks good to me.

I think it would be good for @grzesiek2010 to take a look as well, at least around the question asked to him. @grzesiek2010 please mark as ready for QA when you're ready.

lognaturel:
One idea would be to QA this on `master` so we can verify all the fiddly upgrade path stuff carefully once as part of regression testing rather than twice?

seadowg:
> One idea would be to QA this on master so we can verify all the fiddly upgrade path stuff carefully once as part of regression testing rather than twice?

That sounds good to me. What do you think @kkrawczyk123?

collect_app/src/test/java/org/odk/collect/android/activities/viewmodels/SplashScreenViewModelTest.kt
grzesiek2010: probably

grzesiek2010:
> I think it would be good for @grzesiek2010 to take a look as well, at least around the question asked to him. @grzesiek2010 please mark as ready for QA when you're ready.

I'm on it, will be done today/tomorrow morning.

seadowg:
I've made this a draft while I pull out the CI tweaks to a different PR

collect_app/src/main/java/org/odk/collect/android/application/initialization/upgrade/BeforeProjectsInstallDetector.kt
```diff
@@ -0,0 +1,22 @@
+package org.odk.collect.android.application.initialization.upgrade
+
+import android.content.Context
+import androidx.preference.PreferenceManager
+import org.odk.collect.android.utilities.FileUtils
+import org.odk.collect.upgrade.InstallDetector
+import java.io.File
+
+class BeforeProjectsInstallDetector(private val context: Context) : InstallDetector {
```
grzesiek2010, line 16: I think it would be good to have a comment that explains what that `BeforeProjectsInstall` means. Now we know it but in the future who knows... you can explain it a bit and include version name that was the first one with projects.

upgrade/src/main/java/org/odk/collect/upgrade/LaunchState.kt
```diff
@@ -0,0 +1,32 @@
...
+interface LaunchState {
+    fun isUpgradedFirstLaunch(): Boolean
+    fun appLaunched()
+}
+
+class VersionCodeLaunchState(
+    private val key: String,
+    private val settings: Settings,
+    private val currentVersion: Int,
+    private val installDetector: InstallDetector? = null
```
grzesiek2010: Why is this nullable?

upgrade/src/main/java/org/odk/collect/upgrade/LaunchState.kt
```diff
@@ -0,0 +1,32 @@
...
+            return installDetector?.installDetected() ?: false
+        }
+    }
+
+    override fun appLaunched() {
+        settings.save(key, currentVersion)
+    }
+}
+
+interface InstallDetector {
```
grzesiek2010, line 30: Why did you put this interface here? I think we usually would do that in the same class with its implementation if it has only one implementation.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -584,17 +580,33 @@ public ExistingSettingsMigrator providesExistingSettingsMigrator(ProjectsReposit
     }
 
     @Provides
-    public AppUpgrader providesAppUpgrader(SettingsProvider settingsProvider, ExistingProjectMigrator existingProjectMigrator, FormUpdatesUpgrade formUpdatesUpgrade, ExistingSettingsMigrator existingSettingsMigrator) {
-        return new AppUpgrader(settingsProvider.getMetaSettings(), asList(
+    public AppUpgrader providesAppUpgrader(Context context, SettingsProvider settingsProvider, ExistingProjectMigrator existingProjectMigrator, FormUpdatesUpgrade formUpdatesUpgrade, ExistingSettingsMigrator existingSettingsMigrator) {
```
grzesiek2010: Is this used anywhere?

upgrade/src/main/java/org/odk/collect/upgrade/AppUpgrader.kt
```diff
@@ -0,0 +1,50 @@
...
+) {
+    constructor(
+        key: String,
+        settings: Settings,
+        versionCode: Int,
+        installDetector: InstallDetector,
+        upgrades: List<Upgrade>
+    ) : this(
+        settings,
+        VersionCodeLaunchState(key, settings, versionCode, installDetector),
```
grzesiek2010, line 19: Wouldn't it be better to pass it as a parameter then you wouldn't need this secondary constructor at all.

upgrade/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+    }
+}
+
+dependencies {
+    implementation Dependencies.androidx_core_ktx
+    implementation project(path: ':shared')
```
grzesiek2010, line 47: You can skip `path: ` we use both versions it would be good to choose one and use across the app.

grzesiek2010: You can skip path: we use both versions it would be good to choose one and use across the app.

grzesiek2010 (changes requested):
Generally it looks good, a few comments from me and conflicts to solve.

upgrade/src/main/java/org/odk/collect/upgrade/LaunchState.kt
seadowg: The `upgrade` module owns the interface here - `LaunchState` and `BeforeProjectsInstallDetector` both depend on `InstallDetector`. In Java, I'd probably have given this its own file but I kind of like it just sitting beside the place that needs it in Kotlin?

upgrade/src/main/java/org/odk/collect/upgrade/AppUpgrader.kt
seadowg: The idea here is that we can keep `LaunchState` and `VersionCodeLaunchState` internal to `upgrade` - dependent modules don't need to know anything about them and can just pass in an `InstallDetector` etc. I intended to make `LaunchState` etc `internal` to make use of this but completely forgot. I've taken care of that now, so it's hopefully clear what the benefit of the secondary "convenience" constructor is.

upgrade/src/main/java/org/odk/collect/upgrade/LaunchState.kt
grzesiek2010: But if `BeforeProjectsInstallDetector` was in `upgrade` module you would probably put that interface there since it's the only implementation?
Is there any reason why `BeforeProjectsInstallDetector` is not there (in `upgrade` module) it seems like it is not coupled with any other modules so it could be I guess?

upgrade/src/main/java/org/odk/collect/upgrade/AppUpgrader.kt
grzesiek2010: I see that makes sense.

upgrade/src/main/java/org/odk/collect/upgrade/AppUpgrader.kt
grzesiek2010: Now I understand it especially that you added this comment but it would be good to have a short comment in that class as well because it is really confusing and I think without it  we might want to remove this code next time we touch this class because at first glance it seem s like fomenting we do not need.

upgrade/src/main/java/org/odk/collect/upgrade/LaunchState.kt
seadowg: It could be moved to `upgrade` given that there are no logical dependencies between `BeforeProjectsInstallDetector` and anything else in `collect_app`, but for me `BeforeProjectsInstallDetector` is "Collect specific" code whereas `upgrade` could work with any Android app(theoretically). I'd see it as a domain split. In other words, by exposing the `InstallDetector` interface `upgrade` is saying to `collect_app`: "I need one of these passed to me - you choose how it works".

grzesiek2010: ok if it's a matter of taste and not something important the one who implemented it (you in this case) can decide so let's leave what you like more.

upgrade/src/main/java/org/odk/collect/upgrade/Upgrade.kt
grzesiek2010: Yes now it's a bit clearer. I'm not sure if it's the best option but probably it would be difficult to implement something simpler but with backwards compatibility at the same time. Let's leave it maybe next time you or me touch that class will be able to improve it.

grzesiek2010 (approved):
I would add one comment what I described here https://github.com/getodk/collect/pull/4934#discussion_r782125422
but since it's just a comment I'm accepting this pr.

PR #4935, opened 2021-12-13
Detect and log potential duplicate project as user property

seadowg:

Created the problem on device and then checked the user property was set to true.

I think the heuristic I've used (demo project name without demo ID and all default settings) should be good enough to detect projects created accidentally. It would also detect a project created manually named "Demo project" where no settings have been changed, but I think we can assume that they'll likely only be a handful of false positives there. Definitely interested to see if anyone can think of any edge cases other than that though!

Using a user property for this lets us quickly see how many devices are experiencing the issue (and compare to the total user base).

Shouldn't add any risk as it just changes analytics.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
@@ -4,6 +4,8 @@ import android.content.Context
 import org.odk.collect.analytics.Analytics
 import org.odk.collect.android.preferences.FormUpdateMode
 import org.odk.collect.android.preferences.keys.ProjectKeys
+import org.odk.collect.android.preferences.keys.ProjectKeys.KEY_SERVER_URL
+import org.odk.collect.android.preferences.keys.ProtectedProjectKeys
 import org.odk.collect.android.preferences.source.SettingsProvider
 import org.odk.collect.projects.Project
 import org.odk.collect.projects.ProjectsRepository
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
This catches both the case where a demo project had been created so now there are two and where a different project was previously created so now there’s one demo project and it was not user-requested? Or just the duplicate case?

seadowg:
> This catches both the case where a demo project had been created so now there are two and where a different project was previously created so now there’s one demo project and it was not user-requested? Or just the duplicate case?

It actually doesn't really care whether the project is a duplicate but just that it looks like it's been accidentally created. I realized as I was writing the code that the key indicator was a project called "Demo project" that doesn't have the `DEMO` ID (which it would have if it was created by the actual "Try Collect" button). So it's both cases I think.

lognaturel:
> So it's both cases I think.

Great. I was pretty sure but not 100%. Maybe we can ask @kkrawczyk123 to independently verify by trying both cases and seeing the project ids (just using ls from adb shell) so that we’re confident we know what the analytics mean?

lognaturel:
We’ll release because this is the best we can do but @kkrawczyk123 please proceed with the check to be totally confident of what we’re tracking. The check will be between real releases, not using this PR. I think the three cases are:
- install a pre-project version, don’t do anything with it, upgrade right away to a version with projects. There should be a demo project there. Confirm its id is not DEMO (as discovered from doing ls in the projects folder).
- install a version with projects. Create the demo project. Upgrade to a point release after that. There should be two demo projects. Confirm one had ID DEMO and the other does not. 
- install a version with projects. Create a project other than demo. Upgrade to a point release after that. Confirm the demo project has an id that’s not DEMO


Please edit if I got any of this wrong, @seadowg!

kkrawczyk123:
@lognaturel @seadowg Do I understand correctly that I should test it once the newest beta including this PR will be released?

seadowg:
@kkrawczyk123 No you'll need to test using the previously released versions (v2021.3.1 and v2021.3.2 would do it). The new version will have the problem fixed so you won't be able to reproduce it.

kkrawczyk123:
@seadowg ok, thanks - I don't know what I was thinking xd

My results:
- installed pre-project version: 1.30.1, updated with v2021.3.2: demo project is created; project id is not DEMO. 
- installed project version: v2021.2.4, added demo project, updated with v2021.3.2: 2 demo projects are created, one with project id DEMO and another one with different id
- installed project version: v2021.2.4, added other project than demo, updated with v2021.3.2: demo project created with id not demo. 

Is there anything more that you want me to check?

lognaturel:
Thanks, @kkrawczyk123, that’s it. I’m glad the behavior is as @seadowg had figured. Neat trick using the project ID that way. This info means I’d feel comfortable cleaning up the extra projects but I can certainly understand that may still feel risky. Let’s see how common it is.

PR #4936, opened 2021-12-14
Stop using localized strings in `DownloadFormsTask`

### LINKED ISSUE #4931: Manual downloads shows as failing when language not set to English

seadowg:
Closes #4931
~~Blocked by #4929~~

This makes the same changes as #4932 but also removes the source of the problem by taking localized strings out of `DownloadFormsTask`.

Existing tests checking download flows.

I didn't want to add the risk of a refactor to the original hotfix, but I feel pretty good about this change after making it. Using the localized success/failure message in our logic was a little strange but it also opened us up to bugs like the one we experienced (a translated string getting compared to a non translated one). I think using a nullable exception to track success/failure is much clearer **and** safer.

Good to just check the issue is fixed!

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
+++ b/collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
@@ -14,6 +14,8 @@
 
 package org.odk.collect.android.tasks;
 
+import static org.odk.collect.strings.localization.LocalizedApplicationKt.getLocalizedString;
+
 import android.os.AsyncTask;
 
 import org.odk.collect.android.R;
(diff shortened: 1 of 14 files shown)


collect_app/src/main/java/org/odk/collect/android/tasks/DownloadFormsTask.java
```diff
@@ -62,7 +60,7 @@ public DownloadFormsTask(FormDownloader formDownloader) {
                 publishProgress(serverFormDetails.getFormName(), currentFormNumber, totalForms);
 
                 formDownloader.downloadForm(serverFormDetails, count -> {
-                    String message = Collect.getInstance().getString(R.string.form_download_progress,
+                    String message = getLocalizedString(Collect.getInstance(), R.string.form_download_progress,
```
seadowg, line 63: This is still a localized string but it is actually displayed rather than being used as part of logic.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 10.0.

Tested cases:
- download forms in different languages: German, Spanish, French, English
- download forms from different servers: Central, Google Drive, Kobo, Ona
- send submissions to different servers
- download all forms
- error on failed download

PR #4938, opened 2021-12-17
Prevent crashes when clicking Save Geopoint early

seadowg:
Fixes the crash reported in [Firebase](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/0c4681222cfeb34c2e29805ae831dc55?time=last-seven-days&type=crash&versions=v2021.3.3%20(4332)&sessionEventKey=61BC7099023000016A8366CC23160BDB_1621074324837477558).

We won't need a `master` version of this change as we are reimplanting the `GeoPointActivity` (in #4718). I'll add a note to that issue to block saving the geopoint early.

New test to catch the case we missed in review and checked manually.

Comments inline.

Should hopefully just fix the crash. I think it'd be a good idea to run QA over this quickly (just the default geopoint widget).

### DIFF
--- a/geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
+++ b/geo/src/main/java/org/odk/collect/geo/GeoPointActivity.java
@@ -53,7 +53,7 @@ import java.util.TimerTask;
 import timber.log.Timber;
 
 public class GeoPointActivity extends LocalizedActivity implements LocationListener,
-        LocationClient.LocationClientListener, GpsStatus.Listener {
+        LocationClient.LocationClientListener, GpsStatus.Listener, DialogInterface.OnClickListener {
 
     public static final String EXTRA_ACCURACY_THRESHOLD = "accuracyThreshold";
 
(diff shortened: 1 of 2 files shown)


lognaturel (approved):

lognaturel:
I almost want another analytics state for this but not enough to slow down getting a fix. We don’t necessarily know that this is premature. It’s possible someone has been waiting for a while for some reason and still doesn’t have a point. I have seen that before. But we can easily address it by blocking the save button as @seadowg said.

seadowg:
> I have seen that before. But we can easily address it by blocking the save button as @seadowg said.

Exactly. I thought about an extra analytics call, but I think we'll be getting rid of the case either way so won't be able to measure a change.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1, 8.1 and 10 (as Android 9 updated overnight)

Verified cases:
- I was able to reproduce issue on store version (I was able to tap save location button before location loaded after disabling GPS, restart device, enable GPS again)
- performed the same scenario on PR apk: crash not visible
- geopoint without map, with map
- tried also with geotrace and geoshape

PR #4942, opened 2021-12-21
New geopoint design 📍

### LINKED ISSUE #4718: Improve geopoint dialog

seadowg:
Closes #4718 

I've made the new geopoint design an experimental feature for now, so it can be enabled in Settings > Experimental. This is so we can still switch easily between them for comparison. It also might be that we want to release it as an "opt in" beta to get a more concentrated set of feedback.

New tests and verified manually in the wild.

Comments inline.

Should just add the new design! Worth testing the new geopoint experience and the other geo widgets as some changes were made that could affect any of them.

Any form with the default geo widget.


lognaturel:
This looks great! The pulsing is nice. Have you found a way to provide an accuracy to the emulator? Or have you been sending in values in code to experience changing accuracy? I’m unfortunately without a physical device at the moment. Does the pulsing coincide with accuracy reading updates in normal usage?

I think this is ready for a beta for folks to try.

seadowg:
> Have you found a way to provide an accuracy to the emulator? Or have you been sending in values in code to experience changing accuracy?

I've been hacking `LocationUtils.sanitizeAccuracy` to try out different values.

> Does the pulsing coincide with accuracy reading updates in normal usage?

Right not it pulses in a 4s loop (2s fade in/2s fade out) so doesn't correspond to the loop. The UI gets a new value from the `LocationTracker` in 1s intervals. I'm going to give this a play out in the wild (once it stops raining here 🌧🏴󠁧󠁢󠁳󠁣󠁴󠁿) to see how this feels, but I'll probably end up tweaking the animation to trigger on an update.

> I think this is ready for a beta for folks to try.

Ace! I'll add a todo list to the PR description of things I need to tweak before and get this ready for review!

lognaturel:
First, no pressure to look at this on any particular time frame. 🎄

I have good news and hopefully not too annoying news! The good news is that I do have access to a physical device. The possibly frustrating news is that I don’t love the current state as much as I’d hoped. @yanokwa and I looked at it and we agreed it’d be good to do one more round before a beta. You’re right that there’s nothing like trying the real deal. I’m looking forward to finding out whether you agree with my observations and have any more to add once you play with it more yourself.

The pulse looks nice to me when there’s a single, infrequently changing accuracy as I saw on the emulator. But my experience with the real device is as I’d remembered — the accuracy changes at each new reading. That makes the pulsing feel like visual noise.

I got less of a sense of progress from the changing messages than I had hoped. I’ve typically been starting with an accuracy under 100m so I don’t get to see many states. A typical experience has been that I launch the dialog with an accuracy around 20m. Then it kind of hovers around that and takes a bit of time to get down to 5m. Ideally I’d feel like some good things have already happened and if I just wait a little longer it will be even better. I think that’s the general feeling folks advocating for a status visualizer would like to get across. 

So I think I’ve come back full circle to liking a linear progress bar concept like you initially mentioned and similar to what Tino showed [here](https://forum.getodk.org/t/collect-improve-geopoint-dialog-so-it-better-highlights-acceptable-accuracy/34246/12?u=ln). What do you think of replacing the two lines of text under the accuracy with a single line of status text and a definite progress bar with 5 steps that each add 20% to the progress (and could move backwards):
* No location available: `Trying to get location. Please wait.` (this 20% would always be filled when the dialog opens)
* Worse than unacceptable threshold (default 100m): `Unacceptable accuracy. Please wait.`
* Between unacceptable threshold and 5m before capture threshold: `Poor accuracy. Please wait.`
* 5m before capture threshold to capture threshold: `Improving accuracy. Please wait.`
* If capture threshold is -1, last 20% fills, dialog stays open: `Location ready to save.`

We don’t currently have the last concept and that could certainly be added separately. Without it the most the bar would ever fill is 80% (the dialog would close at what would be 100%).

This is the same basic concept we have now with the addition of a visual sense of progress. I’ve also tried to tighten the language which ended up feeling more redundant and long-winded than I’d hoped. Maybe with the progress bar we wouldn’t even need “Please wait”. In particular I thought the “X from goal of Y” would give that feeling of progress but seeing it in action I agree with those who said it’d be too many numbers to look at. We could later layer on some suggestions like making sure the sky is visible.

Looking forward to comparing notes soon!

seadowg:
> The pulse looks nice to me when there’s a single, infrequently changing accuracy as I saw on the emulator. But my experience with the real device is as I’d remembered — the accuracy changes at each new reading. That makes the pulsing feel like visual noise.

Did you try with the latest code? I changed the update interval from 1s to 5s. I had **exactly** the same reaction but found it much easier to deal with at 5s. I'm wondering if the 1s interval has been causing problems for people in the old one as it would be pretty easy to hit "Save" and end up with the "next" point rather than the one you thought you were saving.

> I got less of a sense of progress from the changing messages than I had hoped. I’ve typically been starting with an accuracy under 100m so I don’t get to see many states. A typical experience has been that I launch the dialog with an accuracy around 20m. Then it kind of hovers around that and takes a bit of time to get down to 5m.

Same! I went out in the car yesterday (it's still and will forever be raining here) and tried this out in a bunch of different places and was seeing the same thing.

> So I think I’ve come back full circle to liking a linear progress bar concept like you initially mentioned and similar to what Tino showed here.

Yeah, my head is going to the same place after trying it out. It does seem that we're seeing the "progress" with the GPS warming up that I initially assumed. I'm glad you're thinking along the same lines as I was unsure I'd be able to look at the idea again given the pushback it had. I realized when I was out and about that an obvious metaphor we missed for this is the signal strength indicators that phones (and lots of other things) have. I'm definitely up for trying out the suggested changes! I was wanting to try out a progress bar but needed to come up with a scheme for how it advances based on the distance from the threshold - yours looks good! It might be that people see different GPS behaviour from us, but I think we can explicitly ask beta testers to see if the progress jumps around too much.

Other than the things you've brought up I found a couple of other problems:

* The dialog seems to be dismissible by clicking outside - not sure what's going on there yet
* `LocationTracker` caches locations so if you get a point under the threshold, move to a different physical location and then hit "Change" in the widget you just get the same point again
* The 100m "unacceptable" threshold feels too high

I'll look at fixes for the first two, and then I'm hoping that the progress bar helps with the third one.

lognaturel:
Happy New Year geopoint enthusiasts! 🎉 😆 

Really glad to hear we're thinking along the same lines, @seadowg!

> signal strength indicators that phones (and lots of other things) have

Ah yes, totally! I'm liking that the progress bar feels pretty easy to lay out and link to progress text but if you have another simple visual in mind, please run with it!

> I think we can explicitly ask beta testers to see if the progress jumps around too much.

Will be interesting to hear what folks experience. I was worried about it initially but I'm less so. I've experienced a series of big jumps down or a single huge jump up if something changes in the conditions. I imagine hovering right around a threshold would be possible but not likely. I also think making the "unacceptable" cutoff point configurable would help address that issue.

> The 100m "unacceptable" threshold feels too high

Agreed for urban conditions. I don't think it will be for places that have to rely only on GPS. Definitely something to get feedback on so we set a good default. I can't remember whether I made this explicit or linked to a source but 100m is about a city block. It used to be the cutoff for coarse vs fine accuracy. I can't quickly confirm whether it still is. Again, making this configurable will be valuable.

By the way, we're going to have some issues with Android 12 around accuracy: https://developer.android.com/training/location/permissions

> Did you try with the latest code? I changed the update interval from 1s to 5s.

I still haven't had a chance to try this on a real device, unfortunately. My gut sense is that this is too long. Am I understanding correctly that if your first reading is at 5.1m then it immediately goes below 5m there'll be a 5s lag before that's reflected (and the point is captured in that case)? In many scenarios you'll get to a good point in less than 5s so having to wait such a long time doesn't feel right. The 1s update feels good to me. I  meant that the fade didn't add much on top of that. Saving the "next" point would rarely be an issue because typically accuracy does improve over time. You're right that there must occasionally be cases where it goes up and the user taps the save button just as it's updated. But I can't imagine those two conditions at the same time are common? Apologies if I misunderstood what you're getting at -- hopefully I can actually try it tomorrow.

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.odk.collect.analytics.Analytics.Companion.log
+import org.odk.collect.async.Scheduler
+import org.odk.collect.geo.analytics.AnalyticsEvents
+import org.odk.collect.location.Location
+import org.odk.collect.location.tracker.LocationTracker
+
+internal abstract class GeoPointViewModel : ViewModel() {
+    abstract val accuracyThreshold: Float
+
+    abstract val location: LiveData<Location?>
```
lognaturel: Feels like this should be `acceptedLocation`?

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+
+            if (trackerLocation.value == null) {
+                updateLocation()
+            }
+        },
+        1000L
+    )
+
+    private val trackerLocation = MutableLiveData<Location?>(null)
+    private val acceptedLocation = MutableLiveData<Location?>(null)
```
lognaturel: Why is this not `_acceptedLocation` with corresponding public property grouped with it?

geo/src/test/java/org/odk/collect/geo/LocationTrackerGeoPointViewModelTest.kt
```diff
@@ -0,0 +1,192 @@
...
+    @Test
+    fun `start() starts LocationTracker with with retain mock accuracy value when set`() {
+        val viewModel = LocationTrackerGeoPointViewModel(locationTracker, { 0 }, scheduler)
+        viewModel.start(retainMockAccuracy = true)
+
+        verify(locationTracker).start(true, 1000L)
+    }
+
+    @Test
+    fun `location is null when no location`() {
```
lognaturel: Related to my naming comment in the implementation -- I was initially puzzled by what this was testing for. It's really like `location is null when no location selected` or something. Just an additional nudge towards the tiny rename.

geo/src/test/java/org/odk/collect/geo/LocationTrackerGeoPointViewModelTest.kt
```diff
@@ -0,0 +1,192 @@
...
+        viewModel.start(accuracyThreshold = 1.0f)
+
+        val location = liveDataTester.activate(viewModel.location)
+        whenever(locationTracker.getCurrentLocation()).thenReturn(Location(0.0, 0.0, 0.0, 1.1f))
+        scheduler.runForeground()
+        assertThat(location.value, equalTo(null))
+    }
+
+    @Test
+    fun `location is tracker location when accuracy is equal to threshold value`() {
```
lognaturel: I think with the tests as they exist there could be a tragic bug that only accepts when the tracker location ends up exactly matching the threshold which is unlikely. Maybe add the less than case?

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    private val _timeElapsed = MutableLiveData<Long>(0)
+    override val timeElapsed = _timeElapsed
+
+    override fun start(retainMockAccuracy: Boolean, accuracyThreshold: Float?) {
+        if (accuracyThreshold != null) {
+            this.accuracyThreshold = accuracyThreshold
+        }
+
+        locationTracker.start(retainMockAccuracy, 1000L)
```
lognaturel, line 59: I'm confused about this request to set the update interval. Prior to this addition it was using the default values from the `LocationClient` implementations, right? Those would be 5s update and 2.5s fastest update? Then why were we seeing the value change every second?

strings/src/main/res/values/strings.xml
```diff
@@ -504,6 +504,16 @@
...
     <string name="get_trace">Start GeoTrace</string>
 
+    <string name="time_elapsed">Time elapsed: %1$s</string>
+    <string name="point_will_be_saved">Point will be saved at %1$s</string>
+    <string name="waiting_for_location">Trying to get location. Please wait.</string>
+    <string name="poor_accuracy">Poor accuracy. Please wait.</string>
+    <string name="unacceptable_accuracy">Unacceptable accuracy. Please wait.</string>
+    <string name="improving_accuracy">Improving accuracy. Please wait.</string>
+    <string name="accuracy_status_title">Finding Location</string>
+    <string name="empty_accuracy">--</string>
```
lognaturel: Maybe this should be untranslated?

strings/src/main/res/values/strings.xml
```diff
@@ -504,6 +504,16 @@
     <string name="get_shape">Start GeoShape</string>
     <string name="get_trace">Start GeoTrace</string>
 
+    <string name="time_elapsed">Time elapsed: %1$s</string>
+    <string name="point_will_be_saved">Point will be saved at %1$s</string>
+    <string name="waiting_for_location">Trying to get location. Please wait.</string>
+    <string name="poor_accuracy">Poor accuracy. Please wait.</string>
+    <string name="unacceptable_accuracy">Unacceptable accuracy. Please wait.</string>
+    <string name="improving_accuracy">Improving accuracy. Please wait.</string>
+    <string name="accuracy_status_title">Finding Location</string>
```
lognaturel: Maybe "getting location" is more clearly linked to the user's goals? Or maybe it'd be more helpful to give the user a sense of why they're in the loop. Something like "Optimizing location reading" though maybe that sounds too technical? Not sure action is needed here but a few ideas to chew over.

geo/src/main/res/layout/accuracy_status.xml
```diff
@@ -0,0 +1,61 @@
...
+        android:textColor="?colorOnPrimary"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/title" />
+
+    <TextView
+        android:id="@+id/text"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/margin_standard"
```
lognaturel: I went with margin_small here and margin_extra_small for the strength below and that got me to what makes sense to me.

lognaturel (commented):
I’ve reviewed the implementation and have no significant feedback, just a couple small questions inline. 

I’ve also made one small layout suggestion inline. 

The 5s update interval still feels problematic to me. I had a particularly frustrating experience where I was dancing around the threshold for 15s. I got something like 5.07m, 6m, 5.2m before the point was accepted with 4.67m. I’ve never had to wait 15s to get a point before in that location and I’m confident that with more frequent updates it would have captured more quickly. I don’t think these small adjustments in accuracy are meaningful and I think we want to optimize for quick capture where possible. We can go to beta like this and see what folks think. 

Relatedly, I think the fade isn’t helping giving a sense of progress as-is. As we discussed, I’m curious to see how it looks on the progress bar fill instead.

seadowg:
> The 5s update interval still feels problematic to me. I had a particularly frustrating experience where I was dancing around the threshold for 15s. I got something like 5.07m, 6m, 5.2m before the point was accepted with 4.67m. I’ve never had to wait 15s to get a point before in that location and I’m confident that with more frequent updates it would have captured more quickly.

Right this case of "missing" the more accurate point convinces me we should drop it back to 1s. My concern around user's getting the point "after" the one they thought they were saving is still something I think we should keep in mind, but feels like less of an issue.

> Relatedly, I think the fade isn’t helping giving a sense of progress as-is. As we discussed, I’m curious to see how it looks on the progress bar fill instead.

Yeah I'm not convinced by it either. I wasn't a fan of the alpha animation on the progress bar, but I didn't try just doing it on the "indicator" bar. I'll give that a try - it might be too complex to be worth it and in which case I'll bail on it.

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
seadowg: Yeah this made sense at the time I think, but I agree with you now!

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
seadowg: The `LocationTracker` just exposes the current value as a `Location` (not a `LiveData` or via a callback). Basically the app is asking the OS to update the location every second, but we're only updating the value we're displaying (via the repeat foreground loop in `GeoPointViewModel`) every 5s (with the logic of giving the user a chance to accept it). I'm going to switch the UI update back to 1s, so it won't really matter anymore. 

We coud rework `LocationTracker` to expose its `Location` as a reactive value or us a callback but originally we were integrating with geo trace where the update time was configurable and easier to keep the `LocationTracker` using the "fastest" update time needed and maintain a separate repeated check at the UI level. I think that's also simpler here personally.

strings/src/main/res/values/strings.xml
seadowg: I definitely like "Getting location" the most for now so will change to that.

androidshared/src/main/java/org/odk/collect/androidshared/ui/Animations.kt
```diff
@@ -0,0 +1,82 @@
...
+import android.animation.ValueAnimator
+import android.view.View
+import org.odk.collect.androidshared.ui.Animations.DISABLE_ANIMATIONS
+
+/**
+ * Helpers/extensions for running animations. These are "test safe" in that animations can be disabled
+ * using [DISABLE_ANIMATIONS] - this should be set to `true` in Robolectric tests to avoid
+ * infinite loops.
+ */
+object Animations {
```
seadowg, line 14: Not using this at the moment, but I think it'd most likely come in handy later (for any other animations we want) so don't want to get rid of it.

collect_app/src/main/java/org/odk/collect/android/preferences/keys/ProjectKeys.kt
```diff
@@ -66,6 +66,9 @@ object ProjectKeys {
     const val KEY_BACKGROUND_LOCATION = "background_location"
     const val KEY_BACKGROUND_RECORDING = "background_recording"
 
+    // experimental_preferences.xml
+    const val KEY_EXPERIMENTAL_GEOPOINT = "experimental_geopoint"
```
seadowg, line 70: We might want to delete the old geopoint stuff now and move to always using the new experience, but I'd be happy to do a beta with it as "opt in" to get more targeted feedback for the moment.

seadowg:
> I'll give that a try - it might be too complex to be worth it and in which case I'll bail on it.

Going to avoid this for now. We can add it later if it feels needed.

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
lognaturel: The thing I'm more puzzled by is why we were getting a 1s update even with what I understand as a request to cap updates at 2.5s previously! Makes me nervous that there's a bug somewhere.

lognaturel (approved):

lognaturel:
Looks great! A few good questions for beta. @seadowg I think we want to merge today before QA so we can tag a beta on `master`, right? I think that can be done now.

@getodk/testers we want to get user feedback as soon as possible because it's fairly isolated and has been dragging on through many holidays! So it would be good to get your verification off `master`. There have been some light refactors to how location is acquired and passed on to the various geo widgets so some light regression checking in those areas would be appreciated. If you have a device without Google Play Services, it would be good to verify it still gets points.

kkrawczyk123:
@seadowg  I have noticed that on devices with a lower screen, information about saving points and time elapsed is not visible in the landscape, see:
Screenshot (03_32PM, Jan 11, 2022)


other verified cases on Androids 5.1 and 8.1:
- collect location when new dialog not enabled
- enable new location dialog 
- light/dark mode
- change language/RTL language
- minimize app when dialog opened
- screen rotation
- save button
- cancel button
- save automatically 
- regression check on all location widgets

seadowg:
@kkrawczyk123 ok looks like we just need to make it scrollable. I'll look at fixing that.

srujner:
**Tested with Success!** 

Verified on Androids: 10 and 11:

- collect location when new dialog not enabled
- enable new location dialog
- light/dark mode
- change language/RTL language
- minimize app when dialog opened
- screen rotation
- save/cancel button
- save automatically
- regression check on all location widgets

PR #4944, opened 2021-12-29
Fix class reference

seadowg:
Just fixes a reference to a class that was breaking test code compilation. No need for QA checks.


lognaturel (approved):
Whoops!

PR #4945, opened 2021-12-29
Increase CI resource size

seadowg:
We've been seeing a lot of "Gradle build daemon disappeared unexpectedly" errors in CI. These seem to have replaced the "out of memory" errors we were seeing before - I think we've traded exceeding our memory allowance for not having enough memory to dependably compile our code.

Ran CI on this branch several times.

After looking at a bunch of support posts (including [this one](https://discuss.circleci.com/t/gradle-build-daemon-disappeared-unexpectedly-react-native/31003)), it seems like the best thing to do is up the size of our CI machine so that we can increase the heap size we allow the JVM to have. It will cost 1.5x credits but avoiding reruns might actually drop our overall costs.

Just changes CI config. No need for CI.

### DIFF
--- a/.circleci/gradle.properties
+++ b/.circleci/gradle.properties
@@ -1,5 +1,5 @@
-org.gradle.jvmargs=-Xmx3072M -XX:+UseParallelGC -Dkotlin.daemon.jvm.options\="-Xmx3072M"
-org.gradle.workers.max=2
+org.gradle.jvmargs=-Xmx5120M -XX:+UseParallelGC -Dkotlin.daemon.jvm.options\="-Xmx5120M"
+org.gradle.workers.max=3
 org.gradle.daemon=false
 org.gradle.parallel=true
 kotlin.compiler.execution.strategy=in-process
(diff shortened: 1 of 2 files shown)


.circleci/gradle.properties
```diff
@@ -1,5 +1,5 @@
-org.gradle.jvmargs=-Xmx3072M -XX:+UseParallelGC -Dkotlin.daemon.jvm.options\="-Xmx3072M"
-org.gradle.workers.max=2
+org.gradle.jvmargs=-Xmx5120M -XX:+UseParallelGC -Dkotlin.daemon.jvm.options\="-Xmx5120M"
```
yanokwa, line 1: I would expect 4608M to match the earlier ratio. Where did you get the 5120 from? I'm not blocking approval, just curious so we know if we need to change it again.

yanokwa (approved):

.circleci/gradle.properties
seadowg: We seemed to be getting along fine on avoiding the JVM processes exceeding the container memory with the 1G headroom, so I kept that consistent.

PR #4955, opened 2022-01-05
Fixed playing videos in android 12

### LINKED ISSUE #4954: Videos in forms do not play in Android 12
Collect v2021.3.2, Android 12, Pixel 3a

Use this [farmer-registration.zip](https://github.com/getodk/collect/files/7812482/farmer-registration.zip) with videos on an Android 12 device. Try to play one of the videos in the form. You will get a "No activity found to handle: View Video".

#### Other information 
This same form works on earlier versions of Android. I have Google Photos (traditionally what Android has used to play videos) installed.

grzesiek2010:
Closes #4954 

I tested the fix manually.

It uses the new `IntentLauncher` that we implemented recently to keep the code responsible for launching intents in one place to avoid bugs like this one.

This should just fix the issue but please take a quick look at other media files too especially on android 12 to make sure there are no other bugs in this area.

I used:
[mediaTest.zip](https://github.com/getodk/collect/files/7817997/mediaTest.zip)

No.


collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
```diff
@@ -41,25 +41,32 @@
...
+
 import timber.log.Timber;
 
 /**
  * Represents a label for a prompt/question or a select choice. The label can have media
  * attached to it as well as text (such as audio, video or an image).
  */
 public class AudioVideoImageTextLabel extends RelativeLayout implements View.OnClickListener {
+    @Inject
+    MediaUtils mediaUtils;
```
seadowg: I think we could avoid using Dagger here and just pass `MediaUtils` in with `setVideo`. We do state in [our code guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#dependency-injection) that it's "best to avoid Dagger and use normal Java constructors" if possible.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.kt
```diff
@@ -41,6 +41,13 @@ class MediaUtils {
     }
 
     fun openFile(context: Context, file: File, expectedMimeType: String?) {
+        if (!file.exists()) {
```
seadowg, line 44: Moving this to a helper seems like a great opportunity to add a test for the behaviour. Good time to create a `MediaUtilsTest`?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -106,6 +107,9 @@
     @Inject
     SettingsProvider settingsProvider;
 
+    @Inject
+    MediaUtils mediaUtils;
```
seadowg, line 112: Again, I think we could avoid Dagger here and just use the constructor as before.

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
```diff
@@ -134,6 +135,9 @@
     @Inject
     StringRequester stringRequester;
 
+    @Inject
+    MediaUtils mediaUtils;
```
seadowg: Again, I think we could avoid Dagger here and just use the constructor.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/formentry/ODKView.java
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
grzesiek2010: We agree not to change it because that would require passing that object through the entire hierarchy of widgets (something we need to look into since it's not the first problem like that).

collect_app/src/main/java/org/odk/collect/android/formentry/questions/AudioVideoImageTextLabel.java
grzesiek2010: Here the case is similar like with `QuestionWidget`... this class is used not only in questions (`QuestionWidget`) but also in choices `SelectMultipleListAdapter` and `SelectOneListAdapter` those adapters might be used in select widgets but also in dialogs (widgets with `minimal` appearance) so i would need to inject MediaUtils there (`SelectMultipleListAdapter` and `SelectOneListAdapter`) and pass through that hierarchy. Do you like such a solution more?

seadowg: Yeah, I think I'd still prefer to keep Dagger out of the view. `MediaUtils` could be passed to the adapter constructors in the widgets right (given they all have access to the `@Inject`ed field)?

grzesiek2010: Yeah if you have normal widgets that's true but choices might be also used in dialogs you you have `minimal` appearance and then it has to go through more classes but ok it's done.

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.kt
grzesiek2010: Yeah I will look into this. If this is the only thing let's mark the pr as `needs testing`.

seadowg:
@grzesiek2010 could you back off the `merge` and `rebase` this onto master instead?

grzesiek2010:
Theoretically yeah but rebasing onto master throws a lot of conflicts and I find it much riskier because during solving those conflicts I can make a mistake and spoil something. Fixing those conflicts here using Github was muuuch easier so I decided to do that.

Generally I agree that we should rebase onto master but in all cases? To me what I said about the risk make me think it's ok to solve conflicts in this way.

seadowg:
> Generally I agree that we should rebase onto master but in all cases? To me what I said about the risk make me think it's ok to solve conflicts in this way.

100% agree. If it was feeling dicey doing a rebase then fair enough - backing out to a merge makes sense.

kkrawczyk123:
**Tested with success!**

Verified cases:
- verified on emulator with Android 12
- regression check on Android 5.1, 8.1 and 10 
- add, create, play video and seflie and video widgets 
- form with input question and play video
- form with select question and play video

collect_app/src/main/java/org/odk/collect/android/utilities/MediaUtils.kt
grzesiek2010: Ok tests added.

srujner:
**Tested with success!**

Verified also on emulator with Android 12 and I did a regression check on Android 11

collect_app/src/test/java/org/odk/collect/android/utilities/MediaUtilsTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.odk.collect.android.BuildConfig
+import org.odk.collect.androidshared.system.IntentLauncher
+import org.robolectric.shadows.ShadowToast
+import java.io.File
+
+@RunWith(AndroidJUnit4::class)
+class MediaUtilsTest {
+
+    private val context = ApplicationProvider.getApplicationContext<Context>()
+    private lateinit var intentLauncher: IntentLauncher
```
seadowg, line 27: These can all be initialized up here and avoid the `lateinit`.

seadowg: These can all be initialized up here and avoid the `lateinit`.

collect_app/src/test/java/org/odk/collect/android/utilities/MediaUtilsTest.kt
```diff
@@ -0,0 +1,80 @@
...
+            Uri.parse("content://org.odk.collect.android.provider/external_files/Android/data/org.odk.collect.android/files/image.png")
+        )
+
+        val contextCaptor = argumentCaptor<Context>()
+        val intentCaptor = argumentCaptor<Intent>()
+        val runnableCaptor = argumentCaptor<() -> Unit>()
+
+        mediaUtils.openFile(context, file, "image/*")
+
+        verify(intentLauncher).launch(
```
seadowg, line 59: An alternative to this that might be more readable (that using agument captors) is to build an `Intent` for the expectation. I think `Intent.equals` works pretty well if I recall correctly.

seadowg (changes requested):

collect_app/src/test/java/org/odk/collect/android/utilities/MediaUtilsTest.kt
grzesiek2010: Done

collect_app/src/test/java/org/odk/collect/android/utilities/MediaUtilsTest.kt
grzesiek2010: The problem is that we have a function as a parameter in `launch`. I'm not interested in verifying it here but also I can't use `any()` since mixing `any()` and real object when you verify parameters is not possible. If I'm missing something about testing cases like this one please let me know.

seadowg: Oh you can do `any()` and object assertions together, but you have to use `eq()` for the object assertions rather than just passing the object as the parameter. The error message you get back from mixing `any()` and plain objects is really not helpful!

As an aside, if you find you need to use an argument captor, it can often be a sign that it'd be easier to make the mock manually by extending the class than using mockito.

grzesiek2010: Ok but intent does not override equals so I can't use eq() for that since it would always check referential equality.

seadowg: Ach, my bad I thought `Intent` had a nice `equals()`. Never mind!

seadowg (approved):

Issue #4956, opened 2022-01-06
Make sure to teardown `ActivityScenario`/`FragmentScenario` objects


seadowg:
These objects are both `Closable` and so should be treated as resources rather than as objects that will cleaned up properly by GC alone. We need to add `close()` calls in tests using the test scenarios to prevent memory leaks in our suite. In some places we can just use `ActivityScenarioRule` for this. We might need to create our own rule for other places where we create multiple scenarios.

Doing this might enable us to drop our test `maxHeapSize` (changed in #4394) back down to `512m`.

PR #4958, opened 2022-01-10
Fix test heap for CI

seadowg:
This fixes the test memory heap configuration so that we don't end up seeing OOMs on CI (or locally for bigger runs).


buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -91,7 +91,7 @@ object Dependencies {
     const val okhttp3_mockwebserver = "com.squareup.okhttp3:mockwebserver:${Versions.okhttp3}"
     const val kotlinx_coroutines_test = "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.2"
     const val hamcrest = "org.hamcrest:hamcrest:2.2"
-    const val robolectric = "org.robolectric:robolectric:4.5.1"
-    const val robolectric_shadows_multidex = "org.robolectric:shadows-multidex:4.5.1"
+    const val robolectric = "org.robolectric:robolectric:4.7.3"
```
seadowg, line 94: This helped solve some of the memory leaks, but we still have something else going on (probably at least #4956)

lognaturel (approved):

PR #4966, opened 2022-01-14
Avoid crashes in FormEntryActivity when webview is absent

### LINKED ISSUE #4965: android.util.AndroidRuntimeException: java.lang.reflect.InvocationTargetException
Collect v2021.3.4 – v2022.1-beta.0

#### Problem description
https://console.firebase.google.com/u/2/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/4b36e1690704196937b97c4705b56758?time=last-seven-days&versions=v2022.1-beta.0%20(4342)&sessionEventKey=61E086290088000127F4C824A8F75CB2_1631230155384501725

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.odk.collect.android/org.odk.collect.android.activities.FormEntryActivity}: android.util.AndroidRuntimeException: java.lang.reflect.InvocationTargetException
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3846)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4022)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2336)
       at android.os.Looper.loop(Looper.java:246)
       at android.app.ActivityThread.main(ActivityThread.java:8653)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
```

1. Disable Android Web View in your phone (it should be in the list of apps in Settings)
2. Try to open any form.

grzesiek2010:
Closes #4965 

I tested the fix manually.

In `FormEntryActivity` we create `WebView` to handle the issue described in:
- https://github.com/getodk/collect/compare/master...grzesiek2010:COLLECT-4965?expand=1#diff-150be8209a860d77a445a11267821bfba08a71b0b8c581bbb115641f42dc7919R380
- https://github.com/getodk/collect/pull/4803#issuecomment-906168621

some devices might not have  `WebView` (or have it disabled) and then the app would crash. The fix which might cause the crash handles a small bug:
- it occurs only in forms that use selects with `image-map` appearance
- even if the bug is not handled it occurs only if we open a form for the first time after [cold start](https://developer.android.com/topic/performance/vitals/launch-time) 

so it's a small bug but the crash it causes on devices without  `WebView` occurs every time no matter if a from contains that particular widget and no matter if we open a form for the first time or not.
So the issue should be handled. There is not much we can do we we can just catch the exception.

It should just fix the issue. Nothing else should be affected.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -383,7 +383,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             try {
                 new WebView(this);
             } catch (Exception | Error e) {
-                // ignored
+                // Don't crash if WebView not available
             }
         }
 
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -380,7 +380,11 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         // Workaround for https://issuetracker.google.com/issues/37124582. Some widgets trigger
         // this issue by including WebViews
         if (Build.VERSION.SDK_INT >= 24) {
-            new WebView(this);
+            try {
+                new WebView(this);
+            } catch (Exception | Error e) {
+                // ignored
+            }
         }
 
         super.onCreate(savedInstanceState);


grzesiek2010:
@seadowg I think this should go in v2022.1 if you agree please prioritize reviewing.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -380,7 +380,11 @@ public void onCreate(Bundle savedInstanceState) {
         // Workaround for https://issuetracker.google.com/issues/37124582. Some widgets trigger
         // this issue by including WebViews
         if (Build.VERSION.SDK_INT >= 24) {
-            new WebView(this);
+            try {
+                new WebView(this);
+            } catch (Exception | Error e) {
+                // ignored
```
seadowg: Good to have a comment here like "Don't crash if `WebView` not available"

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: Done.

seadowg (approved):
Waiting for CI to go green to merge

PR #4968, opened 2022-01-17
Make it easier to determine how many users have multiple projects

seadowg:
The current user property (`ProjectsCount`) makes it a little awkward to answer the question "how many devices have multiple (> 1) projects configured on them?". While we see projects being adopted, this is the question I've been asking in my head more than "how many projects do people typically have configured?".

Nothing.

We could keep both properties in, but we can just add `ProjectsCount` back in if we feel like we need it later.

No changes to user visible features here. Can skip QA on this.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/initialization/UserPropertiesInitializer.kt
@@ -18,7 +18,7 @@ class UserPropertiesInitializer(
     fun initialize() {
         val projects = projectsRepository.getAll()
 
-        analytics.setUserProperty("ProjectsCount", projects.size.toString())
+        analytics.setUserProperty("UsingMultipleProjects", (projects.size > 1).toString())
 
         analytics.setUserProperty(
             "UsingLegacyFormUpdate",


lognaturel:
I haven’t had much of a chance to look at analytics recently and it’s something I really need to get back to. I thought we’d be able to build an audience with an expression using a custom user property (so ProjectsCount > 1) but I haven’t been able to do it on mobile. Will need to look again on a computer. 

Here are questions I have that relate to the project count:
- Would there be value to tools to organize and/or group projects (would be the case if we’re seeing high numbers a lot)
- Are there any layout changes we should consider? We assumed a small number of projects in a few places

Somewhat related questions I’d like to be able to eventually answer:
- are projects all on the same server or multiple? This would feed into designing the flow for authing as a Central web user 
- do people have multiple central draft projects at once? Would feed into redesigning the drafts flow

seadowg:
> I haven’t had much of a chance to look at analytics recently and it’s something I really need to get back to. I thought we’d be able to build an audience with an expression using a custom user property (so ProjectsCount > 1) but I haven’t been able to do it on mobile. Will need to look again on a computer.

Ah if you can work that out let me know as that'd definitely be better. I had to pull out comparisons of everyone vs people with 1 or 0 projects to get the "multiple" value.

lognaturel:
> Ah if you can work that out let me know as that'd definitely be better. I had to pull out comparisons of everyone vs people with 1 or 0 projects to get the "multiple" value.

Where are you doing this? I’ve added an audience that I think does what you want. Let me know whether you see it.

seadowg:
@lognaturel ah that looks good! Thanks.

PR #4969, opened 2022-01-17
Fix JCenter dependencies

seadowg:
We're running into problems building the app on machines without cached dependencies as JCenter is down. JCenter is going to be "read only" moving forward but also seems to basically be "deprecated" in terms of support so we can no longer rely on it as a source for dependencies. This PR moves any dependencies we were fetching from JCenter over to Maven Central. 

The one exception is `mapbox-android-accounts` which only exists in JCenter currently. To resolve this, I've added a included a local maven repo that includes the mapbox dependency.

Long term, we probably want to think about having a plan for disaster scenarios where dependencies disappear from the internet. The app is **old** and has a broad set of dependencies that are hard to keep track of.


buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -22,7 +22,7 @@ object Dependencies {
     const val androidx_preference_ktx = "androidx.preference:preference-ktx:1.1.1"
     const val androidx_fragment_ktx = "androidx.fragment:fragment-ktx:${Versions.androidx_fragment}"
     const val android_material = "com.google.android.material:material:1.4.0"
-    const val android_flexbox = "com.google.android:flexbox:2.0.1"
+    const val android_flexbox = "com.google.android.flexbox:flexbox:3.0.0"
```
grzesiek2010, line 25: Is this necessary? If not I would rather keep the same version since we are about to release a new version and this pr is not for things like that.

seadowg: 3.0.0 is the only version available in the Google maven repo annoyingly: https://maven.google.com/web/index.html#com.google.android.flexbox:flexbox. It looked like the upgrade was mostly around build tooling and the move out of JCenter from the [release notes](https://github.com/google/flexbox-layout/releases/tag/3.0.0).

grzesiek2010: Ah ok

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -36,14 +36,14 @@ object Dependencies {
...
+    const val burgstaller_okhttp_digest = "io.github.rburgst:okhttp-digest:2.5"
     const val persian_joda_time = "com.github.mohamadian:persianjodatime:1.2"
     const val myanmar_calendar = "com.github.chanmratekoko:myanmar-calendar:1.0.6.RC3"
     const val bikram_sambat = "bikramsambat:bikram-sambat:1.1.0"
     const val danlew_android_joda = "net.danlew:android.joda:2.10.12.2"
-    const val rarepebble_colorpicker = "com.rarepebble:colorpicker:3.0.1"
+    const val rarepebble_colorpicker = "com.github.martin-stone:hsv-alpha-color-picker-android:3.0.1"
     const val commons_io = "commons-io:commons-io:2.5" // Commons 2.6+ introduce java.nio usage that we can't access until our minSdkVersion >= 26 (https://developer.android.com/reference/java/io/File#toPath())
-    const val opencsv = "net.sf.opencsv:opencsv:2.4"
+    const val opencsv = "net.sf.opencsv:opencsv:2.3"
```
seadowg, line 46: We should make sure to highlight the effected areas (just the CSV secondary instances) for regression testing. This dependency is old and unmaintained so needs to go anyway (I'll file a separate issue).

lognaturel: We should use Apache Commons CSV to match JavaRosa

seadowg: Yeah, I think we might even have that pulled in to our classpath anyway. To be clear, I'm not going to switch that over as part of this PR.

grzesiek2010 (approved):

PR #4971, opened 2022-01-17
Replace deprecated CopyFormRule

### LINKED ISSUE #4967: Replace deprecated CopyFormRule
#### Problem description
`CopyFormRule` is a [rule](https://ncorti.com/blog/junit-rules) we use in our tests to copy forms that we need to use in particular tests to app storage before starting a test. It is marked as deprecated and should be removed. We should use `copyForm()` form `Page` instead, like we do for example in `AndroidShortcutsTest`

dczerwinski:
Closes #4967 

Gradle task "checkTests" is running successfully.

`CopyFormRule` is deprecated so it should be removed.

For tests that use `CopyFormRule` and `FormActivityTestRule` simultaneously, additional changes must be made, because `FormActivityTestRule` requires that the form be copied before calling the `evaluate()` method from `FormActivityTestRule` otherwise the `getActivityIntent()` method will throw a `NullPointerException` due to missing form. I think the best solution would be to add the possibility to copy the form when calling the `evaluate()` method in the `CollectTestRule` class.

It should not affect users.

No.

No


seadowg:
@dczerwinski rebasing on top of `master` should get the build passing here.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/tasks/FormLoaderTaskTest.java
```diff
@@ -20,6 +21,7 @@
 public class FormLoaderTaskTest {
 
     private final StoragePathProvider storagePathProvider = new StoragePathProvider();
+    private final CollectTestRule rule = new CollectTestRule();
```
seadowg: This class is just testing a task object (but it runs the test on the device) so we don't need to launch an Activity or make assertions on the UI - no need for `CollectTestRule` here. 

Instead of using `copyForm` you could just use the helper under the hood directly (`AdbFormLoadingUtils.copyFormToDemoProject`). That could be called from a `RunnableRule` which you can add to the rule chain after `ResetStateRule` (using `around`).

collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
```diff
@@ -16,16 +16,35 @@
...
     private final boolean useDemoProject;
+    private String formFilename;
+    private boolean copyToDatabase;
+    private List<String> mediaFilePaths;
 
     public CollectTestRule() {
         this(true);
     }
 
+    public CollectTestRule(String formFilename, boolean copyToDatabase) {
```
grzesiek2010: Copying forms should not happen here. Almost everywhere you can use `copyForm` from `Page` but tests that start `FormEntryActivity` directly need to have their forms copied earlier. Those tests use `FormActivityTestRule` so I believe that rule should be responsible for that.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.java
```diff
@@ -426,6 +426,16 @@ public T copyForm(String formFilename) {
         return (T) this;
     }
 
+    public T copyForm(String formFilename, List<String> mediaFileNames, boolean copyToDatabase) {
```
grzesiek2010: Can you maybe convert this class to kotlin and use parameters with default values to avoid having like now 5 `copyForm` methods?

grzesiek2010 (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
dczerwinski: I moved copying forms to `FormActivityTestRule `.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.java
dczerwinski: Converted. I replace also few other methods, with methods with default parameters values.

seadowg (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.java
```diff
@@ -45,6 +58,12 @@ public FormEntryPage startInFormEntry() {
         return formEntryPage;
     }
 
+    private void initialCopyForm() {
+        ActivityScenario.launch(SplashScreenActivity.class);
+        FirstLaunchPage firstLaunchPage = new FirstLaunchPage().assertOnPage();
+        firstLaunchPage.clickTryCollect().copyForm(formFilename, mediaFilePaths, true);
```
grzesiek2010: The idea of this rule is to launch `FormEntryActivity` directly without starting from the staring screen. You should just copy files here using `AdbFormLoadingUtils`

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.java
```diff
@@ -10,28 +10,41 @@
...
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.external.FormsContract;
 import org.odk.collect.android.storage.StorageSubdirectory;
+import org.odk.collect.android.support.pages.FirstLaunchPage;
 import org.odk.collect.android.support.pages.FormEntryPage;
 import org.odk.collect.forms.Form;
 
+import java.util.List;
+
 public class FormActivityTestRule implements TestRule {
```
grzesiek2010, line 23: Please convert to kotlin as you are here.

grzesiek2010: Please convert to kotlin as you are here.

collect_app/src/androidTest/java/org/odk/collect/android/support/AdbFormLoadingUtils.java
```diff
@@ -58,10 +58,6 @@ public static void copyFormToStorage(String formFilename, List<String> mediaFile
         }
     }
 
-    public static void copyFormToStorage(String formFilename, String projectName) throws IOException {
```
grzesiek2010, line 61: I think you can convert this class to kotlin too.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.java
```diff
@@ -77,7 +77,7 @@ private void clearAppState(Application application) {
     private void setTestState() {
```
grzesiek2010: Please convert this class to kotlin.

dczerwinski: Converted.

collect_app/src/androidTest/java/org/odk/collect/android/support/AdbFormLoadingUtils.java
dczerwinski: Converted.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.java
dczerwinski: Converted.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.java
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/instrumented/tasks/FormLoaderTaskTest.java
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private fun resetDagger() {
+        CollectHelpers.overrideAppDependencyModule(appDependencyModule ?: AppDependencyModule())
+    }
+
+    private fun clearPrefs(context: Context) {
+        settingsProvider.clearAll()
+
+        // Delete legacy prefs in case older version of app was run on test device
```
grzesiek2010: Let's get rid of it, it's been a while since we introduced projects so it's not going to happen I think.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.kt
```diff
@@ -0,0 +1,409 @@
...
+ * operations can be chained without casting. For example `FooPage` would extend
+ * `Page<FooPage>` and calling `fooPage.closeSoftKeyboard()` would return
+ * a `FooPage`.
+ *
+ * @see [Page Objects](https://www.martinfowler.com/bliki/PageObject.html)
+ *
+ * @see [Fluent Interfaces](https://en.wikipedia.org/wiki/Fluent_interface)
+ */
+
+abstract class Page<T : Page<T>> {
```
grzesiek2010, line 78: Across the codebase we try to import static methods like:
- onView
- withText
- matches

etc so that instead of for example: `Espresso.onView` we just call `onView` what should make tests a bit more readable. Please take a look at other tests and clean this one.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.odk.collect.projects.Project.Companion.DEMO_PROJECT
+import java.io.IOException
+
+class FormActivityTestRule : TestRule {
+    private val formFilename: String
+    private val formName: String
+    private val mediaFilePaths: List<String>?
+    private var formEntryPage: FormEntryPage? = null
+
+    constructor(formFilename: String, formName: String) {
```
grzesiek2010: Now as it's kotlin can we use one primary constructor instead of those two?

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException
+
+class ResetStateRule @JvmOverloads constructor(private val appDependencyModule: AppDependencyModule? = null) : TestRule {
+
+    private val settingsProvider = getSettingsProvider()
+
+    override fun apply(base: Statement, description: Description): Statement = ResetStateStatement(base)
+
+    private inner class ResetStateStatement(private val base: Statement) : Statement() {
+        @Throws(Throwable::class)
```
grzesiek2010: You can remove it it's not valuable it it throws something there is noticing we can do about it.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    constructor(formFilename: String, formName: String, mediaFilePaths: List<String>?) {
+        this.formFilename = formFilename
+        this.formName = formName
+        this.mediaFilePaths = mediaFilePaths
+    }
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            @Throws(Throwable::class)
```
grzesiek2010: The same here to me it just clutters the code.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.kt
dczerwinski: Yes, done.

collect_app/src/androidTest/java/org/odk/collect/android/support/FormActivityTestRule.kt
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/Page.kt
dczerwinski: Done.

grzesiek2010 (approved):

grzesiek2010:
It looks good to me. @seadowg if you accept this pr I think you can merge it because it touches only tests and I confirmed they pass on firebase.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ContextMenuTest.java
```diff
@@ -4,24 +4,21 @@
...
     @Rule
     public RuleChain copyFormChain = TestRuleChain.chain()
-            .around(new CopyFormRule(STRING_WIDGETS_TEST_FORM, true));
+            .around(activityTestRule);
 
     @Test
     public void whenRemoveStringAnswer_ShouldAppropriateQuestionBeCleared() {
-        new FormEntryPage("fl")
+        activityTestRule.startInFormEntry()
                 .assertOnPage()
```
seadowg: You should be able to get rid of the `assertOnPage()` calls after `startInFormEntry()` calls throughout. You'll see in `FormActivityTestRule` that it does an `assertOnPage()` before the test (in `apply()`).

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/DeletingRepeatGroupsTest.java
```diff
@@ -143,13 +140,13 @@ public void requestingDeletionOfAllRepeatsInHierarchyStartingFromIndexThatWillNo
                 .deleteGroup()
                 .clickOnText("repeatGroup > 1")
                 .deleteGroup()
-                .pressBack(new FormEntryPage("repeatGroups"))
+                .pressBack(activityTestRule.startInFormEntry())
```
seadowg: We should probably avoid calling `startFormEntry()` and then returning the result to another helper like this (and in a few other places) - the method is setting up an initial state for the test, so we might end up adding some side effect to it (it could navigate through the app to get to form entry for instance). Here I think it'd be better to just create a new `FormEntryPage` object. That also makes it easier to see what screen we'll be on when reading down the test.

collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
```diff
@@ -37,7 +39,7 @@ public Statement apply(Statement base, Description description) {
             public void evaluate() throws Throwable {
                 ActivityScenario.launch(SplashScreenActivity.class);
 
-                if (CopyFormRule.projectCreated) {
+                if (projectCreated) {
```
seadowg: I think we can just get rid of `projectCreated` and this `if-else` here. As far as I can see we no longer set it to `true` now you've got rid of `CopyFormRule so we only need the logic from the `else` branch.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
```diff
@@ -0,0 +1,71 @@
...
+import java.io.File
+import java.io.IOException
+
+class ResetStateRule @JvmOverloads constructor(private val appDependencyModule: AppDependencyModule? = null) : TestRule {
+
+    private val settingsProvider = getSettingsProvider()
+
+    override fun apply(base: Statement, description: Description): Statement = ResetStateStatement(base)
+
+    private inner class ResetStateStatement(private val base: Statement) : Statement() {
```
seadowg: I'm guessing this is probably what Kotlin conversion gave you, but you can get rid of the `inner` here by moving the helper methods into `ResetStateStatement`. Then you can just move `ResetStateStatement` to the top level as a `private` class in the same file. Definitely an advantage of Kotlin!

seadowg (changes requested):
This is looking great! Loads of improvements to our tests here. I've left some comments on a few tweaks we should make while we're here.

One other thing that I think would be worth taking care of before merging is the raw uses of `RuleChain` - often seen alongside `FormActivityTestRule`. We should switch these over to `TestRuleChain.chain` so these tests also get our standard setup and won't need `ResetStateRule` explicitly added (you'll see that gets setup for us in `TestRuleChain.chain`).

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/DeletingRepeatGroupsTest.java
dczerwinski: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/CollectTestRule.java
dczerwinski: Yes I agree. I removed `projectCreated `.

collect_app/src/androidTest/java/org/odk/collect/android/support/ResetStateRule.kt
dczerwinski: Done.

collect_app/src/androidTest/java/org/odk/collect/android/feature/formentry/ContextMenuTest.java
dczerwinski: Done.

seadowg (approved):
Looking good! Liking the changes and I'm a green run on Test Lab. Will wait for code freeze to end before merging this.

PR #4975, opened 2022-01-18
Make unacceptable accuracy threshold configurable

### LINKED ISSUE #4961: Make unacceptable geopoint accuracy cutoff configurable
```
 category = if accuracy > unnacceptableAccuracyThreshold
    unacceptable
else if accuracy > accuracyThreshold + 5
    poor
else
    improving
```
<details>
  <summary>Brainstorming</summary>
Currently there is a accuracyThreshold body attribute: “For <input> with type geopoint, geotrace, or geoshape this sets the auto-accept threshold in meters for geopoint captures”. I’m not entirely sure how/if this is actually supported for traces and shapes. 

We want to also make the unacceptable accuracy cutoff configurable. I think by analogy to `accuracyThreshold` it should be a body attribute in camel case. Additionally, I think we should put it in the http://www.opendatakit.org/xforms namespace. For now it should only apply to manual geopoint.  

I propose `odk:worstAcceptableAccuracy`

Note: I think these are appropriate as body attributes rather than bind attributes because they're more like user experience concerns rather than data concerns. In the case of this new attribute, it can be entirely ignored by a client that doesn't have the new dialog with no data difference.

I think implementation can happen even if we’re not totally sure about the name since it would be easy enough to change. What users will really interact with will be XLSForm which we can use a different name for. 

If the attribute is set, its value is used to determine when to go past the unacceptable accuracy step in the geopoint dialog. If it’s not set, 100m is used. If it is set but is less than accuracyThreshold or accuracyThreshold + 5, then the unacceptable accuracy step on the progress bar is always filled in and the text never is displayed. 
<details>

seadowg:
Closes #4961

New tests.

Comments inline.

It'd be good to test the geo widgets but nothing else has really been touched.

Any form with geo widgets involved.

https://github.com/getodk/docs/issues/1422


collect_app/src/main/java/org/odk/collect/android/utilities/FormEntryPromptUtils.java
```diff
@@ -131,12 +131,18 @@ public static String markQuestionIfIsRequired(String questionText, boolean isReq
...
         List<TreeElement> attributes = prompt.getBindAttributes();
         Optional<TreeElement> attribute = attributes.stream().filter(attr ->
                 attr.getName().equals(attributeName)
         ).findAny();
 
         return attribute.map(TreeElement::getAttributeValue).orElse(null);
     }
+
+    @Nullable
+    public static String getAttribute(FormEntryPrompt formEntryPrompt, String attributeName) {
```
lognaturel: Maybe getBodyAttribute to be even more specific

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.kt
```diff
@@ -0,0 +1,164 @@
...
+import org.odk.collect.geo.Constants.EXTRA_READ_ONLY
+import org.odk.collect.geo.Constants.EXTRA_RETAIN_MOCK_ACCURACY
+import org.odk.collect.geo.GeoPointActivity
+import org.odk.collect.geo.GeoPointMapActivity
+import org.odk.collect.geo.GeoPolyActivity
+import org.odk.collect.permissions.PermissionListener
+import org.odk.collect.permissions.PermissionsProvider
+import java.lang.Boolean.parseBoolean
+
+class ActivityGeoDataRequester(
```
seadowg, line 21: I looked at converting the test as well but backed out when I didn't get a clean auto convert - don't want to hold this up or introduce more risk than we need to.

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -68,10 +70,11 @@ public static String getGeoPolyAnswerToDisplay(String answer) {
         return gp;
     }
 
-    static float getAccuracyThreshold(QuestionDef questionDef) {
+    @Nullable
+    static Float getFloatAttribute(QuestionDef questionDef, String attribute) {
         // Determine the accuracy threshold to use.
```
lognaturel: Remove comment

collect_app/src/main/java/org/odk/collect/android/widgets/utilities/GeoWidgetUtils.java
```diff
@@ -68,10 +70,11 @@ public static String getGeoPolyAnswerToDisplay(String answer) {
...
-        String acc = questionDef.getAdditionalAttribute(null, "accuracyThreshold");
```
lognaturel, line 73: `acc` doesn't make sense as var name anymore

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
```diff
@@ -19,12 +19,22 @@ internal abstract class GeoPointViewModel : ViewModel() {
...
 
-    abstract fun start(retainMockAccuracy: Boolean = false, accuracyThreshold: Float? = null)
+    abstract fun start(
+        retainMockAccuracy: Boolean = false,
+        accuracyThreshold: Float? = null,
+        unacceptableAccuracyThreshold: Float? = null
+    )
     abstract fun forceLocation()
+
+    enum class Category {
```
lognaturel: Artifact of another direction you considered?

geo/src/main/java/org/odk/collect/geo/GeoPointViewModel.kt
```diff
@@ -113,3 +141,12 @@ internal class LocationTrackerGeoPointViewModel(
 }
 
 interface GeoPointViewModelFactory : ViewModelProvider.Factory
+
+internal sealed class GeoPointAccuracy {
+
+    abstract val value: Float
+
+    data class Improving(override val value: Float) : GeoPointAccuracy()
```
lognaturel: This is a very interesting way to represent the way the accuracy is classified! The enum it looks like you considered would have been more like what I would have reached for but I see the advantages of this approach.

collect_app/src/main/java/org/odk/collect/android/utilities/FormEntryPromptUtils.java
```diff
@@ -141,7 +141,7 @@ public static String getBindAttribute(FormEntryPrompt prompt, String attributeNa
     }
 
     @Nullable
-    public static String getAttribute(FormEntryPrompt formEntryPrompt, String attributeName) {
+    public static String getBodyAttribute(FormEntryPrompt formEntryPrompt, String attributeName) {
         String acc = formEntryPrompt.getQuestion().getAdditionalAttribute(null, attributeName);
```
lognaturel: World's tiniest nit: `acc` doesn't make any sense anymore (it used to mean accuracy).

lognaturel (approved):

lognaturel:
Will let you decide if you want to fix the var name before merge.

Feels very low risk. Should we merge and do another beta so @getodk/testers can verify as part of regression testing?

seadowg:
> Should we merge and do another beta so @getodk/testers can verify as part of regression testing?

Sounds good. I haven't had a chance to validate this with an example form, yet so let's do that first if we're going to merge and release without QA. Let me know if you've already given it a try.

seadowg:
OK I've given it a try so going to merge now!

PR #4976, opened 2022-01-18
Remove InstancesDaoHelper

### LINKED ISSUE #4974: Get rid of deprecated InstancesDaoHelper
#### Problem description
`InstancesDaoHelper` is deprecated and should no longer be used. Instead we should use `InstancesRepository` as described in the comments in `InstancesDaoHelper`.

dczerwinski:
Closes #4974 

I tested the fix manually.

`InstancesDaoHelper ` is deprecated so it should be removed. I based my solution on comments in code.

It should not affect users.

No.

No.


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2637,4 +2636,33 @@ private boolean isQuestionRecalculated(FormEntryPrompt mutableQuestionBeforeSave
...
     }
+
+    /**
+     * Checks the database to determine if the current instance being edited has
+     * already been 'marked completed'. A form can be 'unmarked' complete and
+     * then resaved.
+     *
+     * @return true if form has been marked completed, false otherwise.
+     **/
+    private boolean isInstanceComplete(boolean end) {
```
grzesiek2010: This class is big enough so this should be added somewhere else. `FormEntryViewModel` maybe? @seadowg any preferences?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
dczerwinski: I moved it to `FormEntryViewModel `.

seadowg: This might not actually be a very simple refactor. I think if we're going to start moving this logic around, want this to not have `FormEntryActivity` to have to care about it. I think really `FormEntryActivity` should pass a boolean through to `FormSaveViewModel#saveForm` to denote whether the user has indicated the save should be finalized or not, but the check to see if the form should automatically be finalized or not should probably be hidden somewhere within the form save logic (maybe even in `InstancesRepository` itself). 

We also should make sure we have test coverage for the flows with different settings here which I don't think we have.  My preference here is that we take a step back and add some tests (probably to `FormFinalizingTest`) to cover flows form finalization when editing, different "completed by default" settings etc). 

Potentially this is one to leave for now?

grzesiek2010: Ok let's discuss this later and now I'm closing.

PR #4980, opened 2022-01-20
Fixed the 'Reset app' dialog

### LINKED ISSUE #4979: White progress dialog is visible when reset settings in dark mode
Collect v2022.1 Android vx.x.x, device used...

#### Problem description
I have noticed a small regression issue connected to the progress bar dialog displayed when reset settings in dark mode, it's white. See gif:
ezgif com-gif-maker (19)_2022

the same case from v2021.3:
ezgif com-gif-maker (19)2021

1. Set dark mode
2. Reset settings 

#### Expected behavior
Progress dialog should not be white when dark mode is selected.

#### Other information 

grzesiek2010:
Closes #4979

I tested the fix manually.

I replaced android progress dialog with our ProgressDialogFragment which works well with themes and is used in many other places.

It should just fix the issue and has no side effects.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.java
+++ b/collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.java
@@ -1,6 +1,5 @@
 package org.odk.collect.android.preferences.dialogs;
 
-import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.graphics.Color;
@@ -19,6 +18,7 @@ import org.odk.collect.android.activities.CollectAbstractActivity;
 import org.odk.collect.android.fragments.dialogs.ResetSettingsResultDialog;
 import org.odk.collect.android.injection.DaggerUtils;
 import org.odk.collect.android.utilities.ProjectResetter;
+import org.odk.collect.androidshared.ui.DialogFragmentUtils;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -35,7 +35,6 @@ public class ResetDialogPreferenceFragmentCompat extends PreferenceDialogFragmen
     @Inject
     ProjectResetter projectResetter;
 
-    private ProgressDialog progressDialog;
     private AppCompatCheckBox preferences;
     private AppCompatCheckBox instances;
     private AppCompatCheckBox forms;
@@ -125,7 +124,7 @@ public class ResetDialogPreferenceFragmentCompat extends PreferenceDialogFragmen
             new AsyncTask<Void, Void, List<Integer>>() {
                 @Override
                 protected void onPreExecute() {
-                    showProgressDialog();
+                    DialogFragmentUtils.showIfNotShowing(ResetProgressDialog.class, ((CollectAbstractActivity) context).getSupportFragmentManager());
                 }
 
                 @Override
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.java
```diff
@@ -143,10 +143,11 @@ protected void onPostExecute(List<Integer> failedResetActions) {
     }
 
     private void showProgressDialog() {
-        progressDialog = ProgressDialog.show(context,
-                        context.getString(R.string.please_wait),
-                        context.getString(R.string.reset_in_progress),
-                        true);
+        progressDialog = new ProgressDialogFragment();
```
seadowg: Keeping references to Fragments can be dangerous/flakey - if the host Activity is recreated, the Fragments will be as well, and we might end up with a reference to one of multiple fragments or no reference at all when we come to call `hideProgressDialog`.

We should switch to using `DialogFragmentUtils.showIfNotShowing`/`DialogFragmentUtils.dismiss` here for managing the dialog. That might mean pulling out a new subclass (see `FormLoadingDialogFragment`) or we could add a new overload to `DialogFragmentUtils` that lets us pass a tag in (as it currently always uses the class to calculate the tag).

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.java
grzesiek2010: It was my firs solution but I ended up with some test failing on Firebase that I wasn't able to reproduce locally. Surprisingly this solution doesn't cause problems. I will give it another try maybe I missed something obvious since yesterday I also felt under the weather after getting my booster dose.

grzesiek2010: Ok I was able to make it work so looks like I was affected by booster does indeed :smile:

collect_app/src/test/java/org/odk/collect/android/preferences/dialogs/ResetProgressDialogTest.kt
```diff
@@ -0,0 +1,50 @@
...
+            assertThat((it.dialog!! as AlertDialog).getButton((AlertDialog.BUTTON_POSITIVE)).visibility, `is`(View.GONE))
+
+            // Button neutral
+            assertThat((it.dialog!! as AlertDialog).getButton((AlertDialog.BUTTON_NEUTRAL)).visibility, `is`(View.GONE))
+
+            // Button negative
+            assertThat((it.dialog!! as AlertDialog).getButton((AlertDialog.BUTTON_NEGATIVE)).visibility, `is`(View.GONE))
+
+            // Title
+            val titleId: Int = context.resources.getIdentifier("alertTitle", "id", context.packageName)
```
seadowg: We could make it much easier to test `ProgressDialogFragment` implementations by simply adding getters for title and message - these being rendered properly is tested in `ProgressDialogFragmentTest`. Let's add them now. I think given that we're testing that this is a correctly configured `ProgressDialogFragment` we also don't need to test the buttons (as they wouldn't be there by default).

seadowg (changes requested):
Looks good. Just a tweak to testing so will pass this to QA.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 10.

Verified cases:
- reset in dark mode
- reset in light mode
- other progress dialogs: send, delete, download form

collect_app/src/test/java/org/odk/collect/android/preferences/dialogs/ResetProgressDialogTest.kt
grzesiek2010: You mean adding getters for `title` and `message` to `ProgressDialogFragmentTest` so that we can use it in all implementations of `ProgressDialogFragmentTest`?

srujner:
**Tested with success!**
Verified on Android 11.

seadowg (approved):

PR #4982, opened 2022-01-20
Stop using deprecated methods from RobolectricHelpers

### LINKED ISSUE #4981: Stop using deprecated methods from RobolectricHelpers
#### Problem description
In `RobolectricHelpers` we have a few deprecated methods which we should stop using. Every method has its own description so it should be clear what we want to use instead.

dczerwinski:
Closes #4981 

Gradle task "checkTests" is running successfully.

The methods I removed were deprecated and have been replaced by the methods mentioned in the comments.

It should not affect users.

No.

No.


testshared/src/main/java/org/odk/collect/testshared/RobolectricHelpers.java
```diff
@@ -99,38 +96,6 @@ public static void mountExternalStorage() {
         return null;
     }
```
grzesiek2010: I think it's a good opportunity to convert this class to kotlin.

dczerwinski: Converted.

testshared/src/main/java/org/odk/collect/testshared/RobolectricHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.robolectric.shadows.ShadowMediaPlayer
+import org.robolectric.shadows.ShadowMediaPlayer.MediaInfo
+import org.robolectric.shadows.util.DataSource
+
+object RobolectricHelpers {
+
+    var services: MutableMap<Class<*>, ServiceScenario<*>> = HashMap()
+
+    @JvmStatic
+    fun <T : FragmentActivity?> createThemedActivity(clazz: Class<T>?, theme: Int): T {
```
grzesiek2010: This should not be nullable (both).

testshared/src/main/java/org/odk/collect/testshared/RobolectricHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+        val activity = Robolectric.buildActivity(clazz)
+        activity.get()!!.setTheme(theme)
+        return activity.setup().get()
+    }
+
+    @JvmStatic
+    fun getCreatedFromResId(drawable: Drawable?): Int {
```
grzesiek2010: `drawable` should not be nullable.

testshared/src/main/java/org/odk/collect/testshared/RobolectricHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+        return Shadows.shadowOf(drawable).createdFromResId
+    }
+
+    @JvmStatic
+    @JvmOverloads
+    fun setupMediaPlayerDataSource(testFile: String?, duration: Int = 322450): DataSource {
```
grzesiek2010: `testFile` should not be nullable.

testshared/src/main/java/org/odk/collect/testshared/RobolectricHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+        ShadowMediaPlayer.addMediaInfo(dataSource, MediaInfo(duration, 0))
+        return dataSource
+    }
+
+    fun mountExternalStorage() {
+        ShadowEnvironment.setExternalStorageState(Environment.MEDIA_MOUNTED)
+    }
+
+    @JvmStatic
+    fun <T : ViewGroup?> populateRecyclerView(view: T): T {
```
grzesiek2010: This should not be nullable.

grzesiek2010 (approved):

seadowg (approved):

PR #4986, opened 2022-01-23
Fix location tracking notification

seadowg:
I noticed when testing out the new geopoint widget that the "Tracking location..." notification looks wrong:

Screenshot 2022-01-23 at 16 11 18

Turns out that it hadn't been configured properly - the notification icon was `null`. I've fixed it:

Screenshot 2022-01-23 at 17 53 11

Verified manually.

Comments inline.

Should just fix the notification. Not much risk to anything else here.


collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.kt
```diff
@@ -31,10 +28,7 @@ class ApplicationInitializer(
     private val userAgentProvider: UserAgentProvider,
     private val propertyManager: PropertyManager,
     private val analytics: Analytics,
-    private val upgradeInitializer: UpgradeInitializer,
-    private val analyticsInitializer: AnalyticsInitializer,
-    private val projectsRepository: ProjectsRepository,
-    private val settingsProvider: SettingsProvider
+    private val initializers: List<Initializer>
```
seadowg: We've talked before about not wanting to use the Jetpack App Initialization framework and just continue to develop our own custom approach. I wanted to move to using a list of initializers here rather than passing another explicit one in - this made it easier to add the new initializer and it'll make it easy to add others in the future.

collect_app/src/main/java/org/odk/collect/android/injection/config/AppDependencyModule.java
```diff
@@ -584,7 +587,17 @@ public UpgradeInitializer providesUpgradeInitializer(Context context, SettingsPr
 
     @Provides
     public ApplicationInitializer providesApplicationInitializer(Application context, UserAgentProvider userAgentProvider, PropertyManager propertyManager, Analytics analytics, UpgradeInitializer upgradeInitializer, AnalyticsInitializer analyticsInitializer, ProjectsRepository projectsRepository, SettingsProvider settingsProvider) {
-        return new ApplicationInitializer(context, userAgentProvider, propertyManager, analytics, upgradeInitializer, analyticsInitializer, projectsRepository, settingsProvider);
+        return new ApplicationInitializer(context, userAgentProvider, propertyManager, analytics, asList(
```
seadowg: We might want to introduce dependencies for `Initializer`s in the future but at the moment it's nice to be able to see the order they are executed in to make sure any interdependencies between them work out ok.

collect_app/src/main/java/org/odk/collect/android/utilities/IconUtils.java
```diff
@@ -31,10 +31,6 @@
     private IconUtils() {
     }
 
-    public static int getNotificationAppIcon() {
```
seadowg, line 34: This didn't feel worth hanging on to

collect_app/src/main/java/org/odk/collect/android/application/initialization/LocationInitializer.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.odk.collect.android.R
+import org.odk.collect.location.tracker.ForegroundServiceLocationTracker
+
+/**
+ * Initializes the [org.odk.collect.location] framework
+ */
+class LocationInitializer : Initializer {
+
+    override fun initialize() {
+        ForegroundServiceLocationTracker.notificationIcon = R.drawable.ic_notes_white
```
grzesiek2010: Can't we just use this icon statically in `ForegroundServiceLocationTracker` like we do in case of all the other notifications? Do we need to initialize it like that?

seadowg: I guess my thinking was that the icon is "Collect" specific and shouldn't really be in the `location` module

grzesiek2010: I see then that icon could be in `androidshared` or even we could have a separate module for drawables like we have for strings as it is possible to use the same drawables in different modules (rather in the future). You are a master of abstractions and I usually like the way you think but here it's too much for me I guess especially now as you want to add this to v2022.1 it should be a small fix without touching all those initializes (you can do that in a separate pr).

seadowg: > You are a master of abstractions and I usually like the way you think

🙏

> ...but here it's too much for me I guess especially now as you want to add this to v2022.1

Yeah I do see what you're saying. I'm trying to adhere to a self enforced rule that we should be able to pull the module out and use it in another application but needing to do initialization like this is a pretty clunky way of making that work. We can probably do something nicer with a shared interface for grabbing the "notification icon" from the `Application` but for now let's just move the icon to `location`. We're using a different icon for `AudioRecorderService` which I don't think makes sense - really all the app notifications should use a consistent icon. I can create an issue for harmonizing that and we can think about a more "elegant" way to configure the icon then.

grzesiek2010: Ok

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**

Verified cases:
- geopoint with no appearance 
- regression check on other notifications: autosend in setting/forms, form update download, autodownload

srujner:
**Tested with success!**

PR #4990, opened 2022-01-24
Prompt user to enable Location if not enabled

### LINKED ISSUE #4987: GeoPoint does nothing if Location is disabled at a system level

Collect v2022.1 Beta 2

#### Problem description

Trying to to use the new GeoPoint with Location disabled in Android just causes the dialog to wait indefinitely (a location is never fixed and there is no error message).

1. Make sure Collect has location permissions granted
2. Disable Location in Android by going to "Settings" > "Location" and untoggling "Use Location"
3. Open a form with a GeoPoint (without appearance) question and navigate to it
4. Click on "Start GeoPoint"

#### Expected behavior

Collect should show an error message and ask the user to enable location/gps. We already do this for other geo widgets in the same scenario (`geopoint` with `placement-map` appearance for instance).

v2021.3.4 behaves weirdly (in my opinion) in this case: clicking "Save GeoPoint" results in the user being jumped to "Location" settings with a Toast (that's easy to missing) saying "Sorry, Location providers are disabled!". I think we should get closer to the behaviour we have in `placement-map` and show a dialog allowing the user to move to settings if they want.

seadowg:
Closes #4987 

I've made is Location services are checked after Location permissions and the user should only get to the GeoPoint dialog if the permissions are granted **and** Location is enabled now.

Verified manually.

We discussed different places to put this check and decided that we should incorporate it in `PermissionsProvider` so we could have the case covered whenever location permissions are requested. I didn't want to add the risk of changing everything now (during regression testing) so I've introduced a new method that we can switch to later (I'll create an issue).

I also left out tests here as I think we need to rework how we test `PermissionsProvider` - it currently isn't easy to test permission requests without a spy. Let's cover that when we switch over to using this method everywhere.

Just changes the permissions flow for the default GeoPoint widget. It'd be good to try different combinations of permissions granted/denied and location services enabled/disabled. I think this problem caught us all by surprise so good for everyone to keep in mind that disabling "Location" in Android is different from denying location permissions.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
@@ -21,6 +21,8 @@ public class FormHierarchyPage extends Page<FormHierarchyPage> {
     @Override
     public FormHierarchyPage assertOnPage() {
         assertToolbarTitle(formName);
+        assertText(R.string.jump_to_beginning);
+        assertText(R.string.jump_to_end);
         return this;
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.kt
@@ -14,7 +14,6 @@ import org.odk.collect.geo.Constants.EXTRA_RETAIN_MOCK_ACCURACY
 import org.odk.collect.geo.GeoPointActivity
 import org.odk.collect.geo.GeoPointMapActivity
 import org.odk.collect.geo.GeoPolyActivity
-import org.odk.collect.location.LocationUtils
 import org.odk.collect.permissions.PermissionListener
 import org.odk.collect.permissions.PermissionsProvider
 import java.lang.Boolean.parseBoolean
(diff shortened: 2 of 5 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
```diff
@@ -21,6 +21,8 @@ public FormHierarchyPage(String formName) {
     @Override
     public FormHierarchyPage assertOnPage() {
         assertToolbarTitle(formName);
+        assertText(R.string.jump_to_beginning);
```
seadowg, line 24: Before, the assertion was the same as the one in `FormEntryPage` which looks like it caused a timing issue in a Test Lab run I did.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5.1, 8.1 and 10

Verified cases:
- displayed dialog
- light/dark mode
- cancel button
- Enable GPS button
- minimize app
- enable GPS in device
- able to save geo point

One thing that I am not a fan of is that dialog disappears when rotating the screen but it's more nice to have than a bug.

seadowg:
> One thing that I am not a fan of is that dialog disappears when rotating the screen but it's more nice to have than a bug.

It should pop up when they hit the button again though right? I think in general we've been trying to avoid worrying about dialogs staying on screen during rotation (unless they're blocking like progress dialogs) as it's pretty unlikely to happen in real life.

srujner:
**Tested with success!**
Verified all the cases on Android 11

PR #4993, opened 2022-01-25
Replace deprecated DayNightProgressDialog

### LINKED ISSUE #4988: Replace deprecated DayNightProgressDialog
#### Problem description
Replace deprecated DayNightProgressDialog. As the description says:

```
@deprecated {@link ProgressDialog} is deprecated and should be replaced by
 * {@link androidx.appcompat.app.AlertDialog} or
 * {@link org.odk.collect.android.fragments.dialogs.ProgressDialogFragment}
```

dczerwinski:
Closes #4988 

Tested manualy.

The class I removed was deprecated and have been replaced by the classes mentioned in the comments.

The appearance of the progressdialog has changed slightly. It now has rounded corners and a slightly smaller title. This is because ProgressDialogFragment uses MaterialAlertDialogBuilder. Nothing else should change.

No.

No. 

- [ ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#ui-components-style-guidelines)


collect_app/src/main/java/org/odk/collect/android/activities/BearingActivity.java
```diff
@@ -69,44 +68,36 @@ protected void onResume() {
...
-        bearingDialog.setIcon(android.R.drawable.ic_dialog_info);
-        bearingDialog.setTitle(getString(R.string.getting_bearing));
-        bearingDialog.setMessage(getString(R.string.please_wait_long));
-        bearingDialog.setButton(DialogInterface.BUTTON_POSITIVE,
-                getString(R.string.accept_bearing),
-                geopointButtonListener);
-        bearingDialog.setButton(DialogInterface.BUTTON_NEGATIVE,
-                getString(R.string.cancel_location),
-                geopointButtonListener);
+        ProgressDialogFragment progressDialogFragment = new ProgressDialogFragment();
```
grzesiek2010: I think it would be better to have a new implementation of ProgressDialogFragment for this case. Please take a look at what we did recently for ResetProgressDialog: https://github.com/getodk/collect/pull/4980/files

grzesiek2010: And other dialogs should also have their own implementation do you agree @seadowg?

seadowg: I'd actually just been thinking about this. I think in some cases it'd be easier to just create a `ProgressDialogFragment` inline and manage showing/hiding with a fragment tag. I reworked `DialogFragmentUtils` and the fragment implementation to make this possible [here](https://github.com/getodk/collect/pull/5002). Maybe we should rebase this PR onto that one?

grzesiek2010:
@dczerwinski please take a look a the changes introduced by @seadowg in https://github.com/getodk/collect/pull/5002 and follow the same way of creating progress dialogs.

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
```diff
@@ -52,22 +59,35 @@ public String getMessage() {
         return getArguments().getString(MESSAGE);
     }
 
+    public @DrawableRes int getIcon() {
```
grzesiek2010: Those icons do not look good I've been thinking about removing them for a while. @seadowg what do you think it looks like we display them in only two dialogs so maybe we should be consistent and remove it at all?

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
```diff
@@ -134,4 +171,12 @@ private void setArgument(String key, Serializable value) {
...
         return dialogView;
     }
+
+    public boolean isShowing() {
+        AlertDialog dialog = (AlertDialog) getDialog();
+        if (dialog != null) {
+            return dialog.isShowing();
+        }
+        return false;
+    }
```
grzesiek2010: Please convert this class to kotlin.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/SavedFormListProgressDialog.kt
```diff
@@ -0,0 +1,14 @@
+package org.odk.collect.android.preferences.dialogs
+
+import android.content.Context
+import org.odk.collect.android.R
+import org.odk.collect.material.MaterialProgressDialogFragment
+
+class SavedFormListProgressDialog : MaterialProgressDialogFragment() {
```
grzesiek2010: The idea behind https://github.com/getodk/collect/pull/5002 was to allow creating progress dialogs without subclasses but you ended up with a bunch of new ones. Please handle that according to https://github.com/getodk/collect/pull/5002

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/SavedFormListProgressDialog.kt
dczerwinski: Done.

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
dczerwinski: I'm not sure why, but when i convert this class to kotlin I get following error when trying to build app:

> Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class org.odk.collect.android.formentry.RefreshFormListDialogFragment, unresolved supertypes: MaterialProgressDialogFragment
    class org.odk.collect.android.formentry.FormLoadingDialogFragment, unresolved supertypes: MaterialProgressDialogFragment
    class org.odk.collect.android.formentry.saving.SaveFormProgressDialogFragment, unresolved supertypes: MaterialProgressDialogFragment
    class org.odk.collect.android.gdrive.GoogleSheetsUploaderProgressDialog, unresolved supertypes: MaterialProgressDialogFragment

dczerwinski: Done.

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
seadowg: What icons are you referring to? It won't let me navigate to the commented code as the changeset is out of date.

grzesiek2010: This Icon, I think it is used only in two dialogs:
Screenshot_1644256568

lognaturel: This is the (i) icon, right? I would really appreciate it going away. That was the intent as dialogs got reworked. It was a pattern in Android a really long time ago.

grzesiek2010: Exactly it's something we don't add to our new dialogs so it exists in 2-3 old ones.

collect_app/src/main/java/org/odk/collect/android/gdrive/GoogleSheetsUploaderProgressDialog.java
```diff
@@ -86,15 +82,12 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
-        dialog = new DayNightProgressDialog(getActivity());
-        dialog.setTitle(getString(R.string.uploading_data));
-        dialog.setMessage(getString(R.string.please_wait));
-        dialog.setCancelable(false);
-        dialog.setIndeterminate(true);
-        dialog.setProgressStyle(ProgressDialog.STYLE_SPINNER);
-        dialog.setButton(getString(R.string.cancel), (dialog1, which) -> onSendingFormsCanceled.onSendingFormsCanceled());
-        return dialog;
+        setTitle(getString(R.string.uploading_data));
+        setTitle(getString(R.string.uploading_data));
```
grzesiek2010, line 86: This is duplicate.

collect_app/src/main/java/org/odk/collect/android/activities/BearingActivity.java
```diff
@@ -58,55 +55,43 @@ protected void onPause() {
...
-        bearingDialog.setMessage(getString(R.string.please_wait_long));
-        bearingDialog.setButton(DialogInterface.BUTTON_POSITIVE,
-                getString(R.string.accept_bearing),
-                geopointButtonListener);
-        bearingDialog.setButton(DialogInterface.BUTTON_NEGATIVE,
-                getString(R.string.cancel_location),
-                geopointButtonListener);
+        MaterialProgressDialogFragment dialog = new MaterialProgressDialogFragment();
+        dialog.setCancelable(false);
+        dialog.setIcon(android.R.drawable.ic_dialog_info);
```
grzesiek2010: It would be nice if methods in MaterialProgressDialogFragment returned the dialog itself the instead of this we could have chaining like: 
```
dialog.setCancelable(false);
   .setIcon(android.R.drawable.ic_dialog_info);
   .setTitle(getString(R.string.getting_bearing));
   .setMessage(getString(R.string.please_wait_long));
   .setPositiveButton(getString(R.string.accept_bearing), geopointButtonListener);
   .setNegativeButton(getString(R.string.cancel_location), geopointButtonListener);
```

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/BearingActivity.java
seadowg: Maybe we should introduce a builder as it seems like we're moving towards that kind of pattern. It could look like:

```
new MaterialProgressDialogFragment.Builder()
   .setCancelable(false)
   .setIcon(android.R.drawable.ic_dialog_info)
   .setTitle(getString(R.string.getting_bearing))
   .setMessage(getString(R.string.please_wait_long))
   .setPositiveButton(getString(R.string.accept_bearing), geopointButtonListener)
   .setNegativeButton(getString(R.string.cancel_location), geopointButtonListener)
   .build();

grzesiek2010: @seadowg 
Something like [MaterialProgressDialogFragmentNew](https://github.com/getodk/collect/pull/4993/files#diff-1c77f14689d2d89daa2ff00faab89f24800f3253c3dd47745bc6da5fd82f4f03)? it's a solution that allows us to use the builder patter and also create dialogs using its constructor because we need this in dialogs that extend `MaterialProgressDialogFragmentNew`.
I Used that new `MaterialProgressDialogFragmentNew` in a few places to show how it looks like. What do you think?

seadowg: That won't work with rotatations/recreations - remember Fragments will be recreated using their no arg constructors so like Activity objects, storing state in fields/overriding constructors will cause problems. That's the reason for the use of args in `MaterialProgressDialogFragment`. You'll see that there are tests to make sure the state is carried between recreated instances in `MaterialProgressDialogFragmentTest`.

grzesiek2010: Right I knew I was missing something... actually I was a bit tricked because when I asked to create separate classes for those dialogs that previously used `DayNightProgressDialog` I was right. Most of those dialogs have listener that should be used with `POSITIVE`/`NEGATIVE` buttons but we can't support such listeners without subclassing `MaterialProgressDialogFragment` because it won't work with recreation.
In this case this whole pr contains a lot of bugs like that and I think we should just close it for now. I can get back to it later when I'm free.
So if we need to support listeners we need sub classes for all the other cases we can add a simple builder.

PR #4994, opened 2022-01-25
Remove translucence so previous point lat/long isn't visible

seadowg:
Does what it says on the tin! This really just needs someone else to verify it and then it can be merged.


lognaturel (approved):
Tried and looks good. It's funny, I really do like it better without the distracting background. regardless of screen size.

PR #4997, opened 2022-01-25
Remove unused json.org dependency

lognaturel:
Removes the misleading org.json dependency. It's shadowed by [Android's org.json](https://developer.android.com/reference/org/json/package-summary). I discovered this when I tried using org.json in JavaRosa. That worked but when I tried it in Collect there were errors about classes missing. [This Stackoverflow post](https://stackoverflow.com/questions/33803776/dont-use-androids-built-in-org-json) helped me understand what was going on.

Ran Collect and verified that things compile.

This makes the origin of `org.json` imports clearer.

There should be no impact. The fact that things compile gives me confidence that all is well. There should be no actual difference as far as I understand.

Importing/exporting settings, settings QR codes and vector offline tile support for Mapbox are potentially affected.

No.

No.

### DIFF
--- a/buildSrc/src/main/java/dependencies/Dependencies.kt
+++ b/buildSrc/src/main/java/dependencies/Dependencies.kt
@@ -64,7 +64,6 @@ object Dependencies {
     const val mp4parser_muxer = "org.mp4parser:muxer:1.9.41"
     const val kotlin_stdlib = "org.jetbrains.kotlin:kotlin-stdlib:${Versions.kotlin}"
     const val kotlin_stdlib_jdk7 = "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${Versions.kotlin}"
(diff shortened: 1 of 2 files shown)


lognaturel:
Hmm, the modules that CI ran tests for seems wrong? I ran tests locally and that seemed to hit all of them.

seadowg (approved):

lognaturel:
I was thinking this might not need QA? @seadowg were you thinking a bit of exploratory testing around the areas I mentioned? It's hard to imagine what would be different given that the APIs are identical and all tests pass. Seems it'd have to be a gnarly edge case or something.

seadowg:
@lognaturel agreed. I'd assumed from your comments in "How does this change affect users..." that you'd prefer testing, and I'm always happy to defer to the author on that one. I'll remove the label and you can merge when ready.

PR #4999, opened 2022-01-26
Updated dependencies for v2022.2

### LINKED ISSUE #4972: Remove `opencsv` dependency
We're using a very old version (2.3). We should either switch to a newer one or look at alternatives like Apache Commons CSV (which JavaRosa is using).

grzesiek2010:
Closes #4972

I've run automated tests.

It doesn't require testing.

No.

No.


grzesiek2010:
in https://github.com/getodk/collect/pull/4999/commits/a8ba7fa4085a709048f41be952152450e81a59f6 I addressed https://github.com/getodk/collect/issues/4972

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -11,39 +11,39 @@ object Dependencies {
     const val androidx_lifecycle_livedata_ktx = "androidx.lifecycle:lifecycle-livedata-ktx:2.3.1" // Upgrading will require compileSdk 31
     const val androidx_lifecycle_viewmodel = "androidx.lifecycle:lifecycle-viewmodel:2.3.1" // Upgrading will require compileSdk 31
     const val androidx_core_ktx = "androidx.core:core-ktx:1.6.0" // Upgrading will require compileSdk 31
-    const val androidx_browser = "androidx.browser:browser:1.3.0"
+    const val androidx_browser = "androidx.browser:browser:1.3.0" // Upgrading will require compileSdk 31
```
seadowg: Can we see if we can get away with this now? If it's simple I don't think there's a reason not to do it.

seadowg (changes requested):

buildSrc/src/main/java/dependencies/Dependencies.kt
grzesiek2010: You mean bumping compileSdk to 31 as a part of this pr?

seadowg: Yeah, or at least just investigating if that's something we can't do. Just looks like there are a lot of `androidx` dependencies that will fall behind otherwise.

grzesiek2010: Done.

config/lint.xml
```diff
@@ -62,4 +62,5 @@
 
     <!-- This flags IDs used via View Binding: https://issuetracker.google.com/issues/170658134 -->
     <issue id="UnusedIds" severity="ignore" />
+    <issue id="Range" severity="ignore" />
```
seadowg: Could you add a comment explaining this? I'm guessing it's to do with `Cursor` indices but good to make sure it's clear here for the future.

seadowg (changes requested):

config/lint.xml
grzesiek2010: > I'm guessing it's to do with Cursor indices. 

Yes. I added a link to a discussion on stackoverflow.

seadowg (approved):

PR #5002, opened 2022-01-27
Allow creating progress dialogs without subclassing

seadowg:
This reworks `ProgressDialogFragment` slightly, so it's now easier to create and show one without needing to create a subclass. This is useful in simpler cases where the displaying Activity/Fragment can just use a static tag to show/hide the dialog.

Extra test coverage for `ProgressDialogFragment` and ran existing tests.

Comments inline.

This can go in without QA as all the changes are pretty small - just moving things around mostly.

### DIFF
--- a/.circleci/gradle.properties
+++ b/.circleci/gradle.properties
@@ -1,5 +1,5 @@
-org.gradle.jvmargs=-Xmx5g -Dkotlin.compiler.execution.strategy="in-process"
+org.gradle.jvmargs=-Xmx2560m -Dkotlin.compiler.execution.strategy="in-process"
 org.gradle.daemon=false
-org.gradle.parallel=false
-org.gradle.workers.max=3
+org.gradle.parallel=true
+org.gradle.workers.max=2
 test.heap.max=1g
--- a/.circleci/gradle.properties
+++ b/.circleci/gradle.properties
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx4g -Dkotlin.compiler.execution.strategy="in-process"
+org.gradle.jvmargs=-Xmx5g -Dkotlin.compiler.execution.strategy="in-process"
 org.gradle.daemon=false
 org.gradle.parallel=true
 org.gradle.workers.max=3
--- a/.circleci/gradle.properties
+++ b/.circleci/gradle.properties
@@ -1,5 +1,5 @@
 org.gradle.jvmargs=-Xmx5g -Dkotlin.compiler.execution.strategy="in-process"
 org.gradle.daemon=false
-org.gradle.parallel=true
+org.gradle.parallel=false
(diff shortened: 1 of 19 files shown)


material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
```diff
@@ -52,29 +62,18 @@ public void setTitle(String title) {
...
-        try {
-            manager
-                    .beginTransaction()
-                    .add(this, tag)
-                    .commit();
-        } catch (IllegalStateException e) {
-            Timber.w(e);
-        }
+    public void setCancelable(boolean cancelable) {
+        setArgument(CANCELABLE, cancelable);
```
seadowg, line 75: Without this, the "cancellable" behaviour wouldn't have been persisted between recreations of the fragment.

.circleci/gradle.properties
```diff
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx4g -Dkotlin.compiler.execution.strategy="in-process"
+org.gradle.jvmargs=-Xmx5g -Dkotlin.compiler.execution.strategy="in-process"
```
seadowg, line 1: I noticed that our `build_instrumented` job seemed to be having a hard time completing, so I'm bumping up the memory to help it out.

.circleci/gradle.properties
```diff
@@ -1,5 +1,5 @@
-org.gradle.jvmargs=-Xmx4g -Dkotlin.compiler.execution.strategy="in-process"
+org.gradle.jvmargs=-Xmx2560m -Dkotlin.compiler.execution.strategy="in-process"
```
seadowg, line 1: We need to make sure that parallel builds don't overstep memory so dropping total limit to half the container memory with 2 max workers.

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
```diff
@@ -52,29 +62,18 @@ public void setTitle(String title) {
...
-        setArgument(message, MESSAGE);
+        setArgument(MESSAGE, message);
 
         AlertDialog dialog = (AlertDialog) getDialog();
         if (dialog != null) {
             setupView(dialog);
         }
     }
 
-    /*
```
grzesiek2010, line 63: Are you sure it's not needed anymore? I remember that long time ago it was something I was able to reproduce or at least I saw it in crash reports.

seadowg: Ah sorry I thought I left a comment about this : `DialogFragmentUtils#showIfNotShowing` has the same workaround, so it was just duplicated rather than not needed.

grzesiek2010 (approved):

PR #5003, opened 2022-01-28
Use coroutines instead of asynctask in ResetDialogPreferenceFragmentCompat

### LINKED ISSUE #5001: Use coroutines instead of asynctask in ResetDialogPreferenceFragmentCompat
#### Problem description
We have multiple old asynctasks across the project. As you know it's better to use coroutines and asynctask is now even deprecated.
`ResetDialogPreferenceFragmentCompat` contains a small asynctask and replacing it should be a good first step.

dczerwinski:
Closes #5001 

Tested manually.

Async tasks are deprected. I replaced them with Corutines.

Should not affect users.

No.

No.


collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+import org.odk.collect.android.fragments.dialogs.ResetSettingsResultDialog
+import org.odk.collect.android.injection.DaggerUtils
+import org.odk.collect.android.utilities.ProjectResetter
+import org.odk.collect.android.utilities.ProjectResetter.ResetAction.RESET_PREFERENCES
+import org.odk.collect.androidshared.ui.DialogFragmentUtils.dismissDialog
+import org.odk.collect.androidshared.ui.DialogFragmentUtils.showIfNotShowing
+import timber.log.Timber
+import javax.inject.Inject
+
+class ResetDialogPreferenceFragmentCompat :
```
grzesiek2010: I like converting classes that you are editing to kotlin but could you do that in a separate commit? It's better to have the fix in one commit so that you can clearly see what was changed.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done. I will remember to split it up into two commits in future.

collect_app/src/main/java/org/odk/collect/android/tasks/DeleteInstances.kt
```diff
@@ -0,0 +1,49 @@
+package org.odk.collect.android.tasks
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.odk.collect.android.instancemanagement.InstanceDeleter
+import org.odk.collect.android.listeners.DeleteInstancesListener
+import timber.log.Timber
+
+class DeleteInstances(
```
grzesiek2010: Please create a separate pr for this refactoring. This pr should be just about `ResetDialogPreferenceFragmentCompat`

dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+        if (cache.isChecked) {
+            resetActions.add(ProjectResetter.ResetAction.RESET_CACHE)
+        }
+        if (resetActions.isNotEmpty()) {
+            showIfNotShowing(
+                ResetProgressDialog::class.java,
+                (_context as CollectAbstractActivity).supportFragmentManager
+            )
+
+            scope.launch(Dispatchers.IO) {
```
grzesiek2010: We have a special module `async` for handling such background tasks. Please take a look at `CoroutineScheduler` and its `immediate` method. I believe that construction should be used here.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+    CompoundButton.OnCheckedChangeListener {
+
+    @JvmField
+    @Inject
+    var projectResetter: ProjectResetter? = null
+
+    private val job = Job()
+    private val scope = CoroutineScope(Dispatchers.Main + job)
+
+    private var _preferences: AppCompatCheckBox? = null
```
grzesiek2010: Ohh I don't like those getters... can we just make those fields nullable without getters and use `!!` when they are called?

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+                ProjectResetter.ResetAction.RESET_FORMS -> if (failedResetActions.contains(action)) {
+                    resultMessage.append(
+                        String.format(
+                            _context.getString(R.string.reset_blank_forms_result),
+                            _context.getString(R.string.error_occured)
+                        )
+                    )
+                } else {
+                    resultMessage.append(
+                        String.format(
```
grzesiek2010: You can use our `getLocalizedString` then a simple form like this:
`resultMessage.append(_context.getLocalizedString(R.string.reset_blank_forms_result, R.string.success))`
should work. Please refactor other cases in this method too.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+    private lateinit var _context: Context
+
+    override fun onAttach(context: Context) {
+        _context = context
+        super.onAttach(context)
+        DaggerUtils.getComponent(context).inject(this)
+    }
+
+    public override fun onBindDialogView(view: View) {
+        _preferences = view.findViewById(R.id.preferences)
```
grzesiek2010: Would setting listeners immediately work? Something like:
```
preferences = view.findViewById<AppCompatCheckBox?>(R.id.preferences).apply {
            setOnCheckedChangeListener(this@ResetDialogPreferenceFragmentCompat)
        }
```

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+import timber.log.Timber
+import javax.inject.Inject
+
+class ResetDialogPreferenceFragmentCompat :
+    PreferenceDialogFragmentCompat(),
+    CompoundButton.OnCheckedChangeListener {
+
+    @JvmField
+    @Inject
+    var projectResetter: ProjectResetter? = null
```
grzesiek2010: This should be `lateinit var`

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,284 @@
...
+import org.odk.collect.androidshared.ui.DialogFragmentUtils.dismissDialog
+import org.odk.collect.androidshared.ui.DialogFragmentUtils.showIfNotShowing
+import timber.log.Timber
+import javax.inject.Inject
+
+class ResetDialogPreferenceFragmentCompat :
+    PreferenceDialogFragmentCompat(),
+    CompoundButton.OnCheckedChangeListener {
+
+    @JvmField
```
grzesiek2010: I think it's redundant.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Yep. I forgot about this class. Corrected.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Yes. Deleted.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,270 @@
...
+    @Inject
+    lateinit var scheduler: Scheduler
+
+    private var preferences: AppCompatCheckBox? = null
+    private var instances: AppCompatCheckBox? = null
+    private var forms: AppCompatCheckBox? = null
+    private var layers: AppCompatCheckBox? = null
+    private var cache: AppCompatCheckBox? = null
+
+    private lateinit var _context: Context
```
grzesiek2010: I think it doesn't make sense to keep this field you can call `getContext()` / `requireContext()` whenever you need it.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,270 @@
...
+                )
+            )
+        }
+    }
+
+    private fun getPartiallyTransparentColor(color: Int): Int =
+        Color.argb(150, Color.red(color), Color.green(color), Color.blue(color))
+
+    companion object {
+        fun newInstance(key: String?): ResetDialogPreferenceFragmentCompat {
```
grzesiek2010: key should not be nullable I think.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
```diff
@@ -0,0 +1,270 @@
...
+                }
+            )
+        }
+    }
+
+    private fun handleResult(resetActions: List<Int>, failedResetActions: List<Int>) {
+        val resultMessage = StringBuilder()
+        for (action in resetActions) {
+            when (action) {
+                RESET_PREFERENCES -> if (failedResetActions.contains(action)) {
```
grzesiek2010: This builder is soo verbose. I think we could make it shorter by converting every case like (example for the first one):

```
                RESET_PREFERENCES -> resultMessage.append(
                    requireContext().getLocalizedString(
                        R.string.reset_settings_result,
                        if (failedResetActions.contains(action)) R.string.error_occured else R.string.success
                    )
                )
```

grzesiek2010 (changes requested):
It looks good I mentioned some final things we should fix.

grzesiek2010:
@seadowg 
please take a look at this pull request.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Yes, done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: Done.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: We need this because after selecting item to reset in the dialog, we show the progress dialog so that the current dialog is detached from `context `and the `getContext()` method or `requireContext()` will return null in the `handleResult()` method.

seadowg: I think we're duplicating the pain points of `AsyncTask` here - we've ended up needing to hold onto state so that the "result" of the async work (`onPostExecute` in `AsyncTask` and the `foreground` block in `Scheduler#immediate`) can access it. If `requireContext()` returns `null` it means that the underlying Activity (the `context` for the `Fragment`) has gone along with the UI, so there is no need for us to display anything. We definitely don't want to be holding our own references to a `context` instance for async work as this can cause nasty memory leaks (Activity objects can be huge as they hold the whole view hierarchy).

The pattern to adopt here (as mentioned in [`STATE.md`](https://github.com/getodk/collect/blob/master/docs/STATE.md)) is to use `Scheduler` (coroutines under the hood) and `LiveData` to avoid needing to hold on to `Context` or views and have the view code "react" (through an `observe` call) to the result (rather than being manipulated from a task/coroutines). 

An example of this would be the use of `BlankFormsListViewModel#syncWithServer` - this fires off work in the background with `Scheduler#immediate` and then sets the resulting data on a `LiveData`. The UI code uses `observe` to react to the returned `LiveData` and we're protected from the Activity being closed etc via the lifecycle awareness features in `observe`.

For `ResetDialogPreferenceFragmentCompat`, we probably want to pull out a method (or create a `ViewModel` to host one) that returns a `LiveData` with the reset and failed actions (set from `foreground`) and the fragment can just call `handleResult` (and hide the progress dialog) from an `observe` observation.

seadowg (changes requested):

seadowg:
Another note here: we want to avoid commit messages that require outside context (like "Apply new review suggestions"). We might need to look back at why a change was made and that wouldn't help us. Even if it's just based on a review suggestion the commit message should describe what actually changed (and hopefully why). It's totally OK to use multiline messages if needed.

collect_app/src/main/java/org/odk/collect/android/preferences/dialogs/ResetDialogPreferenceFragmentCompat.kt
dczerwinski: [ResetDialogPreferenceFragmentCompat.kt](https://github.com/getodk/collect/pull/5003/files/caee35315dd499b470d00e9cf1c52aa3246e703a#diff-be7a893a9d6744e30e4526689f3e99b0165cc2c8aecee98478345b92e22b7b0d) is dialog. When user clicks button "Reset" on this dialog, then we are showing progress dialog, running resetting in background and as result of resetting we are showing other dialog. All of this was written in the [ResetDialogPreferenceFragmentCompat.kt](https://github.com/getodk/collect/pull/5003/files/caee35315dd499b470d00e9cf1c52aa3246e703a#diff-be7a893a9d6744e30e4526689f3e99b0165cc2c8aecee98478345b92e22b7b0d)  class. Until now, we needed the `_context` variable because after pressing the "Reset" button, the dialog was dissmised, so that the `context` of dialog was null, and we used the previously saved `_context` to display the reset results. I think that was a very bad solution because like @seadowg mentioned there may be memory leak problems. Even using `LiveData` would not help because we wanted to use a `context` from view that no longer displays. 

I solved this problem by moving the responsibility of displaying the results to the fragment from which the above-mentioned dialog is shown. I also added `viewModel` to this fragment and used `LiveData` as recommended.

dczerwinski:
> Another note here: we want to avoid commit messages that require outside context (like "Apply new review suggestions"). We might need to look back at why a change was made and that wouldn't help us. Even if it's just based on a review suggestion the commit message should describe what actually changed (and hopefully why). It's totally OK to use multiline messages if needed.

Thanks for the clarification. I've already corrected the commit names.

seadowg:
Going to close this for the moment. I think we need to some extra thinking about hiding and showing progress and result dialogs and how that interacts with the view model, and it would be good to add tests around some of this before merging, but I don't think we'll be able to come back and focus on it anytime soon.

PR #5004, opened 2022-01-28
Reduced jpg quality to reduce file size

### LINKED ISSUE #4978: Size of images taken/chosen by Collect might be inflated
Collect all versions

#### Problem description
Photos taken/chosen by Collect might have bigger size than those taken by the camera app directly.
The issue has been originally reported on the forum: https://forum.getodk.org/t/inflated-image-size-quality-compression/36282

grzesiek2010:
Closes #4978 

I tested capturing/choosing images to make sure file sizes are not inflated.

It is just not recommended to use 100 quality when saving jpg files. It inflates the file size but there is no visible difference between for example the value I used 80 and 100.

Please verify image sizes captured/chosen via ODK Collect vs directly from the camera app. 
Currently (in v2022.1) the issue should be visible only when using `Maximum pixels of the long edge of the image` from `Settings` -> `Form management` (I was able to reproduce it for example using `Large` option). So you should be able to reproduce it only in this way.
This pull request should fix that bug so now no matter if you use original sizes or scale dwon images the sizes should be appropriate (not bigger than original and in case of scaling respectively smaller)

Some camera apps might still produce bigger files when called via ODK Collect (when scaling is not used) but there is nothing we can do with it if those app do not allow us to change quality in their settings (see the discussion in https://forum.getodk.org/t/inflated-image-size-quality-compression/36282/19)

No.

No.


collect_app/src/main/java/org/odk/collect/android/utilities/ImageFileUtils.kt
```diff
@@ -28,7 +28,7 @@ object ImageFileUtils {
             else CompressFormat.JPEG
         try {
             if (bitmap != null) {
-                FileOutputStream(path).use { out -> bitmap.compress(compressFormat, 100, out) }
+                FileOutputStream(path).use { out -> bitmap.compress(compressFormat, 80, out) }
```
yanokwa: How did you decide on 80? My quick search suggests that it's reasonable, but 90 might be safer.

https://photo.stackexchange.com/a/30283

grzesiek2010: There is no one recommended value I think. It depends on what you read. From what I have noticed testing the app 90 in some cases may still produce images bigger than original ones. Also I wasn't able to notice any visible difference (in quality) testing even lower values so 80 seemed to me like a good choice.

https://sirv.com/help/articles/jpeg-quality-comparison/ for example says:

```
90% JPEG quality gives a very high-quality image while gaining a significant reduction on the original 100% file size.
80% JPEG quality gives a greater file size reduction with almost no loss in quality.
75% JPEG quality and lower begins to show obvious differences in the image, which can reduce your website user experience.
```

https://www.lenspiration.com/2020/07/what-quality-setting-should-i-use-for-jpg-photos/ says:

> So, here’s my conclusion: Under “normal” circumstances, you can get by with a JPG quality setting of 80

I don't think we are abnormal an need more than 80.

grzesiek2010: Plus this will only affect images that are scaled down (using max-pixels on a form level or one of the setting we provide in ODK Collect) either way (otherwise we don't call this method as we just copy the original file to our storage). So if someone decides to scale down they really want smaller files.

yanokwa: Thanks for the additional detail. I'm OK with 80.

seadowg: This all sounds good. Could you pull the `80` value out as a private constant and add a comment with the notes on why it's that value?

seadowg (changes requested):

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/utilities/ImageFileUtils.kt
grzesiek2010: Ok done.

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1 and 8.1. 

Verified cases:
- default image size
- large image size 
- small image size
- device's camera image size
- compare results before and after fix: I have noticed a significant decrease in image size after fixing issue 
- image widget
- annotate widget

srujner:
**Tested with success!**
Verified also on Androids 10 and 11.

PR #5009, opened 2022-01-31
Use coroutines instead of asynctask in DownloadFormListTask

### LINKED ISSUE #5001: Use coroutines instead of asynctask in ResetDialogPreferenceFragmentCompat

dczerwinski:
Closes #5001 

Tested manually.

Async tasks are deprected. I replaced them with Corutines.

Should not affect users.

No.

No.


collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.kt
```diff
@@ -0,0 +1,708 @@
...
+ * and you'll have to hit 'refresh' where it will ask for credentials again. Technically a server
+ * could point at other servers requiring authentication to download the forms, but the current
+ * implementation in Collect doesn't allow for that. Mostly this is just because it's a pain in the
+ * butt to keep track of which forms we've downloaded and where we're needing to authenticate. I
+ * think we do something similar in the instanceuploader task/activity, so should change the
+ * implementation eventually.
+ *
+ * @author Carl Hartung (carlhartung@gmail.com)
+ */
+class FormDownloadListActivity :
```
grzesiek2010: Usually I would encourage you to convert classes that you touch to kotlin but in this case I feel uneasy. The class is big and I think it makes it risky. Please just focus on replacing asynctask with coroutines and revert this change.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.kt
dczerwinski: Done. I moved logic of task to viewModel and operate there with coroutines and livedata.

seadowg:
Closing this for the moment but we should come back to it as a reference when we're looking at the relevant code again.

PR #5010, opened 2022-02-01
Cache bitmaps used to display markers on maps

### LINKED ISSUE #5007: java.lang.RuntimeException: Could not allocate dup blob fd.
Collect v1.x.x, Android vx.x.x, device used...

#### Problem description
https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/bb3de6a367388d12a48d24f90bab97a6?versions=v2022.1.0%20(4359)&time=last-seven-days&types=crash&sessionEventKey=61F6AB76002B000177D2525D3AABD425_1637466206334167960

```
Fatal Exception: java.lang.RuntimeException: Unable to resume activity {org.odk.collect.android/org.odk.collect.android.activities.FormMapActivity}: java.lang.RuntimeException: Could not allocate dup blob fd.
       at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4559)
       at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:4591)
       at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:52)
       at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2151)
       at android.os.Looper.loop(Looper.java:250)
       at android.app.ActivityThread.main(ActivityThread.java:7844)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
```

```
Caused by java.lang.RuntimeException: Could not allocate dup blob fd.
       at android.graphics.Bitmap.nativeCreateFromParcel(Bitmap.java)
       at android.graphics.Bitmap.access$000(Bitmap.java:48)
       at android.graphics.Bitmap$1.createFromParcel(Bitmap.java:2132)
       at android.graphics.Bitmap$1.createFromParcel(Bitmap.java:2124)
       at ars.a(ars.java:2)
       at com.google.android.gms.maps.model.internal.a.bj(a.java:7)
       at arr.onTransact(arr.java:4)
       at android.os.Binder.transact(Binder.java:1052)
       at com.google.android.gms.internal.maps.zza.zzH(com.google.android.gms:play-services-maps@@18.0.0:2)
       at com.google.android.gms.internal.maps.zzg.zzg(com.google.android.gms:play-services-maps@@18.0.0:3)
       at com.google.android.gms.maps.model.BitmapDescriptorFactory.fromBitmap(com.google.android.gms:play-services-maps@@18.0.0:2)
```

grzesiek2010:
Closes #5007

I tested the fix manually to make sure there is no regression in adding markers in maps.

I wasn't able to reproduce the issue which seems to take place on some devices with limited memory when a lot of markers is displayed on a map. Thus I  decided to implement caching to avoid creating for example 100 new bitmaps if we have 100 markers that in fact use the same bitmap.

The solution is based on:
- https://developer.android.com/topic/performance/graphics/cache-bitmap#memory-cache
- https://proandroiddev.com/using-vector-drawables-as-google-map-markers-on-android-1eb69790fc61

Please just check if there is no regression in maps (Geotrace / Geoshape widget, FormsMapView) when we display markers (**only with GoogleMaps**).

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
@@ -49,7 +49,6 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.injection.DaggerUtils;
-import org.odk.collect.android.utilities.IconUtils;
 import org.odk.collect.android.utilities.MapFragmentReferenceLayerUtils;
 import org.odk.collect.android.utilities.ThemeUtils;
 import org.odk.collect.androidshared.ui.ToastUtils;
@@ -180,6 +179,11 @@ public class GoogleMapFragment extends SupportMapFragment implements
     @Override public void onStop() {
         mapProvider.onMapFragmentStop(this);
         super.onStop();
+    }
+
+    @Override public void onDestroy() {
+        MapsMarkerCache.clearCache();
+        super.onDestroy();
     }
 
     @Override public void applyConfig(Bundle config) {
@@ -635,8 +639,7 @@ public class GoogleMapFragment extends SupportMapFragment implements
     }
 
     private BitmapDescriptor getBitmapDescriptor(int drawableId) {
-        return BitmapDescriptorFactory.fromBitmap(
-            IconUtils.getBitmap(getActivity(), drawableId));
+        return BitmapDescriptorFactory.fromBitmap(MapsMarkerCache.getMarkerBitmap(drawableId, getContext()));
     }
 
     private void showGpsDisabledAlert() {
--- a/collect_app/src/main/java/org/odk/collect/android/geo/MapsMarkerCache.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/MapsMarkerCache.kt
@@ -1,0 +1,31 @@
+package org.odk.collect.android.geo
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.util.LruCache
+import androidx.annotation.DrawableRes
+import androidx.core.content.ContextCompat
+import androidx.core.graphics.drawable.toBitmap
+
+object MapsMarkerCache {
+    /**
+     * We use different markers in different features but it looks like we will never need to
+     * support more than 10 different types of markers at the same time.
+     */
+    private val cache = LruCache<Int, Bitmap>(10)
+
+    @JvmStatic
+    fun getMarkerBitmap(@DrawableRes drawable: Int, context: Context): Bitmap {
+        if (cache[drawable] == null) {
+            ContextCompat.getDrawable(context, drawable)?.toBitmap().also {
+                cache.put(drawable, it)
+            }
+        }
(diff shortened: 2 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -635,8 +640,7 @@ private float getIconAnchorValueY(@IconAnchor String iconAnchor) {
     }
 
     private BitmapDescriptor getBitmapDescriptor(int drawableId) {
-        return BitmapDescriptorFactory.fromBitmap(
-            IconUtils.getBitmap(getActivity(), drawableId));
+        return BitmapDescriptorFactory.fromBitmap(MapMarkerCache.getMarkerBitmap(drawableId, getContext()));
```
grzesiek2010, line 643: This is used only in `GoogleMapFragment` because other providers do not require bitmaps.

collect_app/src/main/java/org/odk/collect/android/geo/MapsMarkerCache.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.annotation.DrawableRes
+import androidx.core.content.ContextCompat
+import androidx.core.graphics.drawable.toBitmap
+
+object MapsMarkerCache {
+    /**
+     * We use different markers in different features but it looks like we will never need to
+     * support more than 10 different types of markers at the same time.
+     */
+    private val cache = LruCache<Int, Bitmap>(10)
```
seadowg, line 15: I guess we're thinking that we should always have enough memory for all 10 right? If we don't, I guess we won't be able to render the map (with all 10 types).

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/geo/MapsMarkerCache.kt
grzesiek2010: Yes 10 doesn't seem like a big number especially taking into account that our markers are not huge. In many cases we are not going to need all 10. For example in case of:
- Geotrace widget we use just one marker which is a red circle
- Geoshape widgets the same as above
- FormMapActivity it depends in how many states your form are saved we have 4 states and every state might have normal and big marker (the big one is displayed if you click on it) so it's 8 + 1 (for the default red circle marker which is displayed by default and then replaced by one of the states - this is something we can improve in the future I think) = 9

I don't know what would happen if we had such a poor device that can't handle up to 9 bitmaps, probably the same exception would be thrown.

kkrawczyk123:
**Tested with success!**

Verified cases:
- display submission on maps
- different statuses
- tried adding multiple points, up to 30 
- regression checks on geotrace and geoshape
- offline layers
- Google maps, OSM and Mapbox

srujner:
**Tested with Success!**

PR #5016, opened 2022-02-03
Get rid of TranslationHandler

### LINKED ISSUE #5013: Get rid of TranslationHandler
#### Problem description
`TranslationHandler` is just a wrapper that delegates the work directly to our extension function `getLocalizedString`. There is no reason not to use `getLocalizedString` directly.

dczerwinski:
Closes #5013 

Tested manually.

`TranslationHandler` is just a wrapper that delegates the work directly to our extension function `getLocalizedString`. There is no reason not to use `getLocalizedString` directly.

It should not affect users.

No.

No.


collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
```diff
@@ -685,7 +686,7 @@ public static String getDownloadResultMessage(Map<ServerFormDetails, String> res
         for (ServerFormDetails k : keys) {
             b.append(k.getFormName() + " ("
                     + ((k.getFormVersion() != null)
-                    ? (TranslationHandler.getString(Collect.getInstance(), R.string.version) + ": " + k.getFormVersion() + " ")
+                    ? (getLocalizedString(Collect.getInstance(), R.string.version) + ": " + k.getFormVersion() + " ")
```
grzesiek2010: This whole method seems to be unused so please remove it at all.

strings/src/main/java/org/odk/collect/strings/localization/LocalizedApplication.kt
```diff
@@ -9,7 +9,7 @@ interface LocalizedApplication {
     val locale: Locale
 }
 
-fun Context.getLocalizedString(stringId: Int, vararg formatArgs: Any): String {
+fun Context.getLocalizedString(stringId: Int, vararg formatArgs: Any?): String {
```
grzesiek2010: Are you sure you need this change?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormDownloadListActivity.java
dczerwinski: Done.

strings/src/main/java/org/odk/collect/strings/localization/LocalizedApplication.kt
dczerwinski: I set this first to nullable because that was the way this argument was set in some of the methods in the Page class, but as I checked now the getString method should not take null as formatArgs, so I changed it to non-nullable.

grzesiek2010 (approved):

PR #5017, opened 2022-02-03
Return instance URI for Form EDIT action

### LINKED ISSUE #5008: FormEntryActivity does not return instance uri after finishing a form
Collect v2021.3.4

#### Problem description
`FormEntryActivity` does not return instance uri after finishing a form. The problem seems to be caused by the fact that we hide `FormEntryActivity` behind `FormUriActivity` after merging https://github.com/getodk/collect/commit/0e927d7a8bdc3254690cf27b49083235bf95214e and `FormUriActivity` does not return the value returned by `FormEntryActivity` it is just ignored.

#### Other information 
The problem has been originally reported on the forum: https://forum.getodk.org/t/data-uri-is-null-in-onactivityresult/36548

seadowg:
Closes #5008

Updated test.

Comments inline.

Should just fix the problem. The VIEW/EDIT action for forms is the only thing that's had any changes.

### DIFF
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -288,7 +288,6 @@ the specific language governing permissions and limitations under the License.
         Do not use the activity directly as its name/package might change. -->
         <activity
             android:name=".external.FormUriActivity"
-            android:noHistory="true"
             tools:ignore="AppLinkUrlError">
 
             <intent-filter>
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
@@ -42,11 +42,13 @@ class FormUriActivity : Activity() {
             logAnalytics(uriProjectId)
 
             if (projectId == currentProjectProvider.getCurrentProject().uuid) {
-                startActivity(
+                startActivityForResult(
                     Intent(this, FormEntryActivity::class.java).also {
+                        it.action = intent.action
                         it.data = uri
                         intent.extras?.let { sourceExtras -> it.putExtras(sourceExtras) }
-                    }
+                    },
+                    1
                 )
             } else {
                 MaterialAlertDialogBuilder(this)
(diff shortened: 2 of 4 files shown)


collect_app/src/main/AndroidManifest.xml
```diff
@@ -288,7 +288,6 @@ the specific language governing permissions and limitations under the License.
         Do not use the activity directly as its name/package might change. -->
         <activity
             android:name=".external.FormUriActivity"
-            android:noHistory="true"
```
seadowg, line 291: As far as I can see we aren't using this any longer now we explicitly call `finish()` in all cases. `noHistory` doesn't work with activity results as the calling Activity is finished as soon as the user hits the next Activity.

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
```diff
@@ -14,14 +15,20 @@ import org.odk.collect.android.utilities.ApplicationConstants
...
+class FormUriActivity : ComponentActivity() {
 
     @Inject
     lateinit var currentProjectProvider: CurrentProjectProvider
 
     @Inject
     lateinit var projectsRepository: ProjectsRepository
 
+    private val openForm =
+        registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
```
seadowg, line 27: Seemed like a good opportunity to move to the new API for this.

grzesiek2010 (approved):

grzesiek2010:
I've added `needs testing` label but I don't know if it's something our QA team is able to verify, probably not. @kkrawczyk123 @srujner please at least take a look to make sure there is no regression when starting forms especially using external apps like our tester: https://github.com/grzesiek2010/collectTester

kkrawczyk123:
**Tested with success!**

Verified cases:
- download, fill and send forms form/to Central, Drive, Ona and Kobo 
- use CollectTester, download, fill form, send
- use CollectCounter

srujner:
**Tested with success!**
Verified on Androids 10 and 11:

- use CollectTester, download, fill form, send
- use CollectCounter
- download, fill and send forms

PR #5020, opened 2022-02-08
Use Rules to manage Activity/Fragment scenarios

### LINKED ISSUE #4956: Make sure to teardown `ActivityScenario`/`FragmentScenario` objects

seadowg:
Closes #4956 

AndroidX Test doesn't yet provide an easy way to manage `Activity`/`Fragment` scenario objects, so I added some test rules to give us an easy way of creating scenarios without having to worry about the clean-up. After doing that I wanted to clean up our "missing from AndroidX Test" code and have ended up with 3 modules:

* `androidtest` - Things missing from AndroidX Test that we (and most likely everyone else) needs. Can be used with local or instrumented tests.
* `fragmenttest` - Things missing from the AndroidX Test `fragment-testing` module. Mostly concerned with testing Dialog Fragments.
* `servicetest` - Provides a missing (IMO from AndroidX Test) "XScenario" for `Service` that wraps the Robolectric `ServiceController` (like how `ActivityScenario` wraps `ActivityController`). This only works with local tests as it looks like the fancy stuff that makes `ActivityScenario` work in both places uses APIs in Robolectric and instrumentation that would need added to.
* `testshared` - Helpers we use in Android tests but that other apps probably wouldn't need.

While moving stuff I had performed a quick inventory to check everything seemed relevant but I think a good follow up (which I'll make a personal todo) would be to make sure there are issues and feature requests for each of the things in `androidtest`, `fragmenttest` and `servicetest`. It might be that some of the things we've created would be viable as contributions to AndroidX Test itself.

Annoyingly we still can't run all the tests without at least a 1GB heap (double the default). My theory is that `ActivityController` now also needs a `close()` call to prevent memory leaks (I've left a comment on the commit that adds the method [here](https://github.com/robolectric/robolectric/commit/feaded1a56ba818ecc1f580f862f4f40c95b3627#commitcomment-66337000)). I've changed our Gradle setup so we only use the larger heap size in `collect_app` now which means running all unit tests will only spin up one 1GB process rather than one for each module in parallel.


collect_app/src/test/java/org/odk/collect/android/activities/FirstLaunchActivityTest.kt
```diff
@@ -25,13 +25,17 @@ import org.odk.collect.android.projects.QrCodeProjectCreatorDialog
 import org.odk.collect.android.support.CollectHelpers
 import org.odk.collect.android.version.VersionInformation
 import org.odk.collect.strings.localization.getLocalizedString
+import org.odk.collect.testshared.ActivityScenarioLauncherRule
 
 @RunWith(AndroidJUnit4::class)
 class FirstLaunchActivityTest {
 
+    @get:Rule
+    val launcherRule = ActivityScenarioLauncherRule()
```
grzesiek2010, line 34: Why here you need this custom rule but in the previous commit you could use `ActivityScenarioRule`. Can't you do the same here?

seadowg: It's needed for the last test: `ActivityScenarioRule` launches the Activity before the test but in `The app name with its version should be displayed` we want to launch the Activity after overriding some dependencies. This kind of situation, and when we launch multiple Activity objects, are what we require `ActivityScenarioLauncherRule` for. Personally, I think it's how `ActivityScenarioRule` should behave (I'm going to be giving this feedback to AndroidX Test).

grzesiek2010: Ok now I understand. Thanks.

audiorecorder/src/test/java/org/odk/collect/audiorecorder/recording/internal/AudioRecorderServiceTest.kt
```diff
@@ -303,7 +303,7 @@ class AudioRecorderServiceTest {
     private fun startService(intent: Intent): ServiceScenario<AudioRecorderService> {
         return serviceInstance.let { instance ->
             if (instance == null) {
-                ServiceScenario.launch(AudioRecorderService::class.java, intent).also {
+                org.odk.collect.servicetest.ServiceScenario.launch(AudioRecorderService::class.java, intent).also {
```
grzesiek2010: `org.odk.collect.servicetest` is redundant.

grzesiek2010 (changes requested):
Just one thing to fix + conflicts... and I'm not sure if I understand the benefit of having a separate module `servicetest`. Was that problem to have robolectric as dependency in `androidtest` module?

seadowg:
>  and I'm not sure if I understand the benefit of having a separate module servicetest. Was that problem to have robolectric as dependency in androidtest module?

Fair question. My thinking here was to make sure that `androidtest` contained things that could be considered "extensions" to AndroidX Test alone without a need for Robolectric. The code in `servicetest` could just live in `testshared` but I'd already gone to the work of extracting it (to move to `androidtest`) and it felt good to have it contained in one module so it's clearer to see what we've had to add to test services.

grzesiek2010 (approved):

PR #5022, opened 2022-02-09
Re-enable ImageView tests for views using Glide

### LINKED ISSUE #5021: Re-enable `ImageView` tests for views using Glide
We recently switched to Glide to fix some bugs and introduce support of SVGs but ended up being unable to test that images are correctly loaded into `ImageView`s (`DrawWidgetTest` is an example). We should rework our integration with Glide (I imagine we'll end up with some kind of "image loader" interface) so that we can test which images have been loaded into an image view and reenable the `@Ignore`d tests.

grzesiek2010:
Closes #5021 

I've run test and also tested loading images manually.

Implementing an interface that can be overridden in tests seem like the easiest solution so there is not much to discuss here.

Please test image widgets to make sure images are displayed properly. Other than that there should be no side effects.

No.

No.


glide/src/main/java/org/odk/collect/glide/ImageLoaderImpl.kt
```diff
@@ -12,25 +12,24 @@ import com.bumptech.glide.request.target.Target
 import org.odk.collect.glide.svg.SvgSoftwareLayerSetter
 import java.io.File
 
-object ImageLoader {
+class ImageLoaderImpl : ImageLoader {
```
seadowg: This could be named `GlideImageLoader` rather than needing the `Impl` suffix. I think we should probably also rename the module to `imageloader` (or something similar) as the beauty of moving this all to a module and hiding the implementation behind an interface is that dependent modules don't even need to know that Glide exists.

seadowg (changes requested):
Looking good! I think we should do a rename - I've left a comment inline about that.

glide/src/main/java/org/odk/collect/glide/ImageLoaderImpl.kt
grzesiek2010: Done.

seadowg (approved):

kkrawczyk123:
**Tested with success!**

Regression checks on:
- take a picture
- remove response
- preview images
- screen rotation
- add image from drive, devices internal storage
- image, draw, signature, annotate images, selfie
- external image widget
- image select widgets
- default image for annotate and signature widgets

srujner:
**Tested with success!**

Verified also on Androids: 10 and 11

PR #5024, opened 2022-02-10
In Match Exactly mode, retry form download requests before showing failure notification

### LINKED ISSUE #4787: Improve Match Exactly notification experience

grzesiek2010:
Closes #4787

Going through the commits form the first to the last one should be a good way to understand the changes.

I tested the app manually using `Background task inspector` in Android studio and also I added new automated tests.

This is a result of what we have discussed in https://github.com/getodk/collect/issues/4787.
I used `EXPONENTIAL` BackoffPolicy with 60s delay that means:
- if the first attempt fails we retry after 60s
- if it fails again we retry after 120s (after the second attempt)
- if it fails we retry after 240s (after the third attempt)

in total we have 4 attempts (the original one + 3 retries) and it takes 7 minutes

We need to test `matchexactly` since it is about this feature but also it would be good to test two other tasks that uses [Workmanager](https://developer.android.com/topic/libraries/architecture/workmanager) (`autodownload` and `autoupload`).

No.

No.


collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
```diff
@@ -46,10 +49,24 @@ class SyncFormsTaskSpecTest {
     }
 
     @Test
-    fun `calls synchronize with projectId`() {
+    fun `getTask calls synchronize with projectId`() {
```
seadowg: This is a little misleading. Maybe just `task calls synchronize with projectId` (and the same change for the other test)?

collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
```diff
@@ -46,10 +49,24 @@ class SyncFormsTaskSpecTest {
...
         val inputData = HashMap<String, String>()
         inputData[SyncFormsTaskSpec.DATA_PROJECT_ID] = "projectId"
         SyncFormsTaskSpec().getTask(ApplicationProvider.getApplicationContext(), inputData).get()
         verify(formsUpdater).matchFormsWithServer("projectId")
     }
+
+    @Test
+    fun `getTask returns proper result value`() {
```
seadowg: `...returns result from FormUpdater`

collect_app/src/test/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpecTest.kt
```diff
@@ -64,4 +66,9 @@ class AutoSendTaskSpecTest {
         AutoSendTaskSpec().getTask(ApplicationProvider.getApplicationContext(), inputData).get()
         verify(instanceAutoSender).autoSendInstances(projectId)
     }
+
+    @Test
+    fun `numberOfRetries should not be limited`() {
+        assertThat(AutoSendTaskSpec().numberOfRetries, `is`(-1))
```
seadowg: Could we make `numberOfRetries` a `Int?` instead of giving negative numbers a special meaning? Maybe the field name could be `maxRetries` instead also?

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    fun `when task returns true should work succeed`() {
+        whenever(spec.getTask(any(), any())).thenReturn(Supplier { true })
+
+        assertThat(workerAdapter.doWork(), `is`(ListenableWorker.Result.success()))
+    }
+
+    @Test
+    fun `when task returns false should work retry if numberOfRetries not specified`() {
```
seadowg: `when task returns false retries if numberOfRetries not specified` - same pattern for the tests below

async/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
seadowg: What's this?

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.hamcrest.MatcherAssert.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import java.util.function.Supplier
+
+@RunWith(AndroidJUnit4::class)
```
seadowg, line 23: Does this test need Robolectric? I'm guessing something explodes without it?

async/src/main/java/org/odk/collect/async/TaskSpec.kt
```diff
@@ -17,4 +17,8 @@ interface TaskSpec {
      * WorkManager framework
      */
     fun getWorkManagerAdapter(): Class<out WorkerAdapter>
+
+    companion object {
+        const val DATA_PROJECT_ID = "projectId"
```
seadowg: I think this should move back or live in a shared object in `collect_app`. `async` doesn't need to know about Projects, and we very well might add a task that doesn't care about them.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpec.kt
```diff
@@ -18,9 +18,12 @@ class SyncFormsTaskSpec : TaskSpec {
     override fun getTask(context: Context, inputData: Map<String, String>): Supplier<Boolean> {
         DaggerUtils.getComponent(context).inject(this)
         return Supplier {
-            formsUpdater.matchFormsWithServer(
-                inputData[TaskSpec.DATA_PROJECT_ID]!!
-            )
+            val projectId = inputData[TaskSpec.DATA_PROJECT_ID]
+            if (projectId != null) {
+                return@Supplier formsUpdater.matchFormsWithServer(projectId)
```
seadowg: You can just omit the `return` here so you don't need the whole `@Supplier` thing.

async/src/main/java/org/odk/collect/async/TaskSpec.kt
```diff
@@ -20,5 +20,6 @@ interface TaskSpec {
 
     companion object {
         const val DATA_PROJECT_ID = "projectId"
+        const val LAST_UNIQUE_ATTEMPT = "lastAttempt"
```
seadowg: This could potentially get passed as a param on `TaskSpec#getTask` rather than as part of the `inputData` leaving that to the extending classes. I'm not sure though. What do you think?

collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
```diff
@@ -50,19 +50,30 @@ class SyncFormsTaskSpecTest {
     }
 
     @Test
-    fun `getTask calls synchronize with projectId`() {
+    fun `when LAST_UNIQUE_ATTEMPT equals true getTask calls synchronize with proper data`() {
```
seadowg: Instead of `proper data` in these tests it could be `with/without notify` or `with notify true/false`

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
```diff
@@ -69,6 +75,63 @@ class WorkerAdapterTest {
...
+                mapOf(
+                    TaskSpec.DATA_PROJECT_ID to "123",
+                    TaskSpec.LAST_UNIQUE_ATTEMPT to "true"
+                )
+            )
+        )
+    }
+
+    @Test
+    fun `when numberOfRetries is specified and it is higher than runAttemptCount, should data contain LAST_UNIQUE_ATTEMPT equal false`() {
```
seadowg: Again these tests don't need `should` in them really. Could just be `..runAttemptCount, data contains...`

async/src/main/java/org/odk/collect/async/CoroutineScheduler.kt
```diff
@@ -49,6 +50,10 @@ open class CoroutineScheduler(private val foregroundContext: CoroutineContext, p
         throw UnsupportedOperationException()
     }
 
+    override fun networkDeferred(tag: String, spec: TaskSpec, repeatPeriod: Long, backoffPolicy: BackoffPolicy, backoffDelay: Long, inputData: Map<String, String>) {
```
seadowg: Maybe these should live on the `TaskSpec` rather than here? I remember having the same debate with myself about `repeatPeriod`...

seadowg: Either way I'm not sure we need them yet. Right now we could use the same backoff policy and backoff delay for any task that specifies a number of retries (as only match exactly does).

seadowg (changes requested):
Really liking the direction of this. There are a couple of tweaks and also some things about the general `async` API that would be good for us to discuss here that I've left comments for inline.

async/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
grzesiek2010: It enables mocking final classes. I use it in `WorkerAdapterTest` where I mock `WorkerParameters` which is a final class.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
grzesiek2010: Yes I create an instance of `WorkerAdapter` in this test.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
grzesiek2010: Fixed.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
grzesiek2010: Fixed.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
grzesiek2010: Fixed.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpecTest.kt
grzesiek2010: Fixed.

async/src/main/java/org/odk/collect/async/CoroutineScheduler.kt
grzesiek2010: But the existing `networkDeferred` is used by `autoupdate`. I didn't want to affect that task (it wouldn't have limited retries of course but it would change it's default `backoffPolicy` and default `backoffDelay`.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/SyncFormsTaskSpec.kt
grzesiek2010: Right, fixed.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpecTest.kt
grzesiek2010: I renamed `numberOfRetries` to `maxRetries` but I'm not sure if I understand the fist part correctly. You added this comment ins a commit. The final version doe not have that raw -1 but it's represented by a const `UNLIMITED_NUMBER_OF_RETRIES`. Is there anything else you would like to change here?

async/src/main/java/org/odk/collect/async/TaskSpec.kt
grzesiek2010: `TaskSpec#getTask` has `inputData: Map<String, String>` so using it was a natural choice to me. The same question you could ask about `projectId` - why not to pass it as a parameter?

collect_app/src/test/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpecTest.kt
seadowg: Sorry maybe that ended up being unclear. What I mean is using an optional type (`Int?`) instead of a non-optional and then having `null` denote "no retries". Then the check can be `if (spec.maxRetries == null...` and there'd be no need for a special constant. I know using `-1` is pretty common in Java, but I feel like having safe nullables in Kotlin gives us a clearer way of dealing with that kind of "optional" value.

async/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
seadowg: Urgh yeah I remember this : `WorkerParameters` is `final` and has a restricted constructor. I think it'd be good to have an experiment with the worker manager testing tools (https://developer.android.com/topic/libraries/architecture/workmanager/how-to/testing-worker-impl) to see if we can use that instead of introducing `final` mocking to the module.

async/src/main/java/org/odk/collect/async/CoroutineScheduler.kt
seadowg: I was just meaning on `TaskSpec` we could define the fields optionally (like I'm suggesting for `maxRetries`):

```kotlin
interface TaskSpec {
    val backoffPolicy: BackoffPolicy? = null
    val backoffDelay: BackoffDelay? = null
}
```

Then a task can override the value if it wants, and otherwise we just use the default.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpecTest.kt
grzesiek2010: Ah I see but making `maxRetries` nullable would make using it a bit more difficult even in kotlin for example in `WorkerAdapter` instead of:
```
this[TaskSpec.DATA_LAST_UNIQUE_EXECUTION] = (runAttemptCount >= spec.maxRetries).toString()
```
I would need to check if `maxRetries` is not null first (and it's not the only place like that).
Of course I could use this for example:
```
this[TaskSpec.DATA_LAST_UNIQUE_EXECUTION] = (spec.maxRetries?.let { runAttemptCount >= it }).toString()
```
but it makes the code less readable to me.

seadowg: Yeah, I mean I think this is getting into personal preference, but I prefer the latter to the former there. The first example relies on knowledge that a negative number is being used to denote "no retries" whereas I'd argue that the second example makes it clearer that there is no last unique execution if there is no `maxRetries`.  The first example also relies on a bit of a "trick": a negative number is always lower than the run attempt count which is concise, but I don't personally feel it's clearer.

The other problem (and probably the more important one) I have with using negative numbers, is that it brings in more risk of writing buggy code. For instance, this code works with any negative value, but it'd be easy to write code that relies on `-1` specifically and have a value like `-2` cause problems. 

I think if you feel strongly that the negative number is better, then let's keep it - it's just stylistic peeve of mine.

async/src/main/java/org/odk/collect/async/TaskSpec.kt
seadowg: Agreed it could go in either. `projectId` is different as not every `TaskSpec` would (or should) care about Projects (also mentioned [here](https://github.com/getodk/collect/pull/5024#discussion_r805332832)). I do think it'd be better as a parameter as it gives you the simplicity of knowing for sure that the value is present over it being in a map which introduces uncertainty (complexity). I guess which comparing these kinds of things (which feel pretty equal) I'd try to go with more certainty over less, even it's pretty minor.

async/src/main/java/org/odk/collect/async/CoroutineScheduler.kt
grzesiek2010: Ok, done.

async/src/main/java/org/odk/collect/async/TaskSpec.kt
grzesiek2010: Ok, done.

collect_app/src/test/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpecTest.kt
grzesiek2010: Ok we can do that like that, done.

async/src/main/java/org/odk/collect/async/TaskSpec.kt
grzesiek2010: Done.

async/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
grzesiek2010: Yeah we can use it. Done.

async/build.gradle
```diff
@@ -47,4 +47,8 @@ dependencies {
     implementation Dependencies.androidx_work_runtime
 
     testImplementation Dependencies.hamcrest
+    testImplementation Dependencies.robolectric
```
seadowg, line 50: This **should** work without Robolectric - the AndroidX Test stuff should be enough as you're not referencing Robolectric explicitly in the test. However, if I try and remove Robolectric, everything compiles, but I get the following error:

```java
Caused by: java.lang.IllegalStateException: No instrumentation registered! Must run under a registering instrumentation.
	at androidx.test.platform.app.InstrumentationRegistry.getInstrumentation(InstrumentationRegistry.java:45)
	at androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner.createRunnerParams(AndroidJUnit4ClassRunner.java:53)
	at androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner.<init>(AndroidJUnit4ClassRunner.java:48)
```

I'm going to quickly dig into what's happening here as it's confusing.

seadowg: Ok yeah it's just that the docs are all wrong or everything is broken (standard): https://issuetracker.google.com/issues/219885372. Robolectric needs to be included to get AndroidX Test to work with "local tests".

async/src/main/java/org/odk/collect/async/CoroutineAndWorkManagerScheduler.kt
```diff
@@ -42,13 +42,16 @@ class CoroutineAndWorkManagerScheduler(foregroundContext: CoroutineContext, back
...
-        val workRequest = PeriodicWorkRequest.Builder(worker, repeatPeriod, TimeUnit.MILLISECONDS)
+        val builder = PeriodicWorkRequest.Builder(worker, repeatPeriod, TimeUnit.MILLISECONDS)
             .addTag(tag)
             .setInputData(workManagerInputData)
             .setConstraints(constraints)
-            .build()
 
-        workManager.enqueueUniquePeriodicWork(tag, ExistingPeriodicWorkPolicy.REPLACE, workRequest)
+        if (spec.backoffPolicy != null && spec.backoffDelay != null) {
+            builder.setBackoffCriteria(spec.backoffPolicy!!, spec.backoffDelay!!, TimeUnit.MILLISECONDS)
```
seadowg: We can avoid the `!!` with a nested `?.let`:

```kotlin
spec.backoffPolicy?.let { backoffPolicy ->
    spec.backoffDelay?.let { backoffDelay ->
    
    }
}

```

...or by using passing both optionals to a `Pair` and then using `.let`  with destructuring:

```kotlin
Pair(spec.backoffPolicy, spec.backoffDelay).let { (policy, delay) ->
}
```

Another option is just to create a `Backoff` type (probably an inner type of `TaskSpec` that holds both so the whole thing is one optional and both values are non-optional.

async/src/main/java/org/odk/collect/async/Scheduler.kt
```diff
@@ -54,6 +55,27 @@ interface Scheduler {
...
+     *
+     * @param tag used to identify this task in future. Previously scheduled tasks using the same
+     * tag will be replaced
+     * @param spec defines the task to be run
+     * @param repeatPeriod the period between each run of the task
+     * @param backoffPolicy the BackoffPolicy to use when increasing backoff time
+     * @param backoffDelay the time to wait before retrying the work
+     * @param inputData a map of input data that can be accessed by the task
+     */
+    fun networkDeferred(
```
seadowg: I think this is unused now?

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        whenever(spec.getTask(any(), any(), any())).thenReturn(Supplier { false })
+        whenever(spec.maxRetries).thenReturn(3)
+
+        assertThat(worker.doWork(), `is`(ListenableWorker.Result.retry()))
+    }
+
+    @Test
+    fun `when task returns false, fails if maxRetries is specified and is equal to runAttemptCount`() {
+        whenever(spec.getTask(any(), any(), any())).thenReturn(Supplier { false })
+        whenever(spec.maxRetries).thenReturn(1)
```
seadowg: Not sure if I understand this. If `maxRetries` is `1` it should allow a single retry right? So 2 total runs?

async/src/main/java/org/odk/collect/async/WorkerAdapter.kt
```diff
@@ -12,12 +12,16 @@ abstract class WorkerAdapter(
 
     override fun doWork(): Result {
         val stringInputData = inputData.keyValueMap.mapValues { it.value.toString() }
-        val completed = spec.getTask(applicationContext, stringInputData).get()
+        val completed = spec.getTask(applicationContext, stringInputData, isLastUniqueExecution()).get()
 
         return if (completed) {
             Result.success()
-        } else {
+        } else if (spec.maxRetries == null || runAttemptCount < spec.maxRetries!!) {
```
seadowg: I think this should potentially be `runAttemptCount <= spec.maxRetries` (I've left a comment on the test about that). Also, this could avoid `!!` with `spec.maxRetries.let {}` around the `if`.

testshared/build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
     implementation Dependencies.androidx_recyclerview
     implementation Dependencies.kotlin_stdlib
     implementation Dependencies.androidx_core_ktx
+    implementation Dependencies.androidx_work_runtime
```
seadowg, line 43: I think this is only needed because of the unused `networkDeferred` overload.

seadowg (changes requested):
Thanks for indulging my nitpicking here. Probably being extra sensitive as we build out these micro frameworks for ourselves, but hopefully it'll pay off in the future.

Just a couple of tweaks I think we should look at, and I'm also confused about how many retries we end up with. Comments inline.

testshared/build.gradle
grzesiek2010: No, it is used in other places too.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
grzesiek2010: In tests attempts are counted starting from 1 but in normal life from 0 I don't know why...

async/src/main/java/org/odk/collect/async/WorkerAdapter.kt
grzesiek2010: First runAttemp is marked as 0 (in tests 1 as I said above) so if our maxRetries is 3 it's like:
0 < 3 - retry
1 < 3 - retry
2 < 3 - retry
3 < 3 - fail

in total we have 3 reties as we expected. If I used your syntax <= we would have one additional.

grzesiek2010: > Also, this could avoid !! with spec.maxRetries.let {} around the if.

Could you share the whole construction. I'm not sure if I can use anything like that here.

async/src/main/java/org/odk/collect/async/Scheduler.kt
grzesiek2010: Right, fixed.

async/src/main/java/org/odk/collect/async/CoroutineAndWorkManagerScheduler.kt
grzesiek2010: Done.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
seadowg: Ah looks like you can pass `runAttemptCount` to `TestWorkerBuilder` to make this clearer: https://developer.android.com/topic/libraries/architecture/workmanager/how-to/testing-worker-impl#testing-workers

async/src/main/java/org/odk/collect/async/WorkerAdapter.kt
seadowg:
```kotlin
return spec.maxRetries.let { maxRetries ->
    if (completed) {
        Result.success()
    } else if (maxRetries != null && runAttemptCount < maxRetries) {
        Result.retry()
    } else {
        Result.failure()
    }
}
```

Alternatively, you can just pull out the value:

```kotlin
val maxRetries = spec.maxRetries
return if (completed) {
    Result.success()
} else if (maxRetries == null || runAttemptCount < maxRetries) {
    Result.retry()
} else {
    Result.failure()
}
```

grzesiek2010: Right, I like the second option, done.

async/src/test/java/org/odk/collect/async/WorkerAdapterTest.kt
grzesiek2010: Done.

seadowg (approved):

kkrawczyk123:
**Tested with success!**

Verified cases:
- match exactly and the invalid url - wait for failure: ~6-7min
- match exactly and not existing project - wait for failure: ~6-7min
- relaunch match exactly manually - failure notification displayed immediately 
- wait 15min after failure notification 
- failure notification from not currently used project
- turn on match exactly in 2 projects in the same time 
- after 1st fail, update URL to valid and wait for an update - successful update
- turn off internet connection, notification after connection is restored.
- turn on airplane mode
- previously downloaded form notification
- previously download form automatic update
- regression check on autosend: settings and defined in form
- fail notification on autosend

srujner:
**Tested with success!**

PR #5025, opened 2022-02-11
Copy media files from old version rather than redownloading

seadowg:
We recently discovered that media files are redownloaded for new form versions even if they exist in the media directory for an older version. This optimizes that process so that existing media files (files with the same name and contents hash) are copied from old versions rather than performing a download.

New tests and existing tests.

Comments inline.

Hopefully just saves people a bunch of bandwidth. That said, making changes in this part of the code is always dangerous as there are a lot of edge cases, bad-yet-supported behaviour and bugs that have become features which we haven't discovered yet and as such, don't have tests. We'll want to QA form downloading and updating with and without media files and with our usual roster of different servers.

It's also important for reviewers to think of edge cases that we might have missed in tests or in the code itself.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
@@ -1,6 +1,5 @@
 package org.odk.collect.android.formmanagement
 
-import org.odk.collect.android.utilities.FileUtils.constructMediaPath
 import org.odk.collect.android.utilities.FileUtils.copyFile
 import org.odk.collect.android.utilities.FileUtils.interuptablyWriteFile
 import org.odk.collect.async.OngoingWorkListener
@@ -10,80 +9,51 @@ import org.odk.collect.forms.FormSourceException
 import org.odk.collect.forms.FormsRepository
 import org.odk.collect.forms.MediaFile
 import org.odk.collect.shared.strings.Md5.getMd5Hash
-import timber.log.Timber
 import java.io.File
 import java.io.IOException
 
-class FormMediaDownloader constructor(
-    private val formsDirPath: String,
+class FormMediaDownloader(
     private val formsRepository: FormsRepository,
     private val formSource: FormSource
 ) {
 
     @Throws(IOException::class, FormSourceException::class, InterruptedException::class)
     fun download(
-        tempMediaPath: String,
-        stateListener: OngoingWorkListener?,
+        formToDownload: ServerFormDetails,
         files: List<MediaFile>,
+        tempMediaPath: String,
         tempDir: File,
-        formFileName: String,
-        fd: ServerFormDetails
+        stateListener: OngoingWorkListener?
     ) {
         val tempMediaDir = File(tempMediaPath).also { it.mkdir() }
 
         for (i in files.indices) {
             stateListener?.progressUpdate(i + 1)
 
-            val (filename, hash, downloadUrl) = files[i]
-            val tempMediaFile = File(tempMediaDir, filename)
-            val finalMediaPath = constructMediaPath(
-                formsDirPath + File.separator + formFileName
-            )
-            val finalMediaFile = File(finalMediaPath, filename)
+            val mediaFile = files[i]
+            val tempMediaFile = File(tempMediaDir, mediaFile.filename)
 
-            if (!finalMediaFile.exists()) {
-                val allFormVersions = formsRepository.getAllByFormId(fd.formId)
-                val existingFileInOtherVersion = allFormVersions.map { form: Form ->
-                    File(
-                        form.formMediaPath,
-                        filename
-                    )
-                }.firstOrNull { file: File ->
-                    val currentFileHash = getMd5Hash(file)
-                    val downloadFileHash = validateHash(hash)
-                    file.exists() && currentFileHash.contentEquals(downloadFileHash)
-                }
-
-                if (existingFileInOtherVersion != null) {
-                    copyFile(existingFileInOtherVersion, tempMediaFile)
-                } else {
-                    val mediaFile = formSource.fetchMediaFile(downloadUrl)
-                    interuptablyWriteFile(mediaFile, tempMediaFile, tempDir, stateListener)
-                }
+            val existingFile = searchForExistingMediaFile(formToDownload, mediaFile)
+            if (existingFile != null) {
+                copyFile(existingFile, tempMediaFile)
             } else {
-                val currentFileHash = getMd5Hash(finalMediaFile)
-                val downloadFileHash = validateHash(hash)
-                if (currentFileHash != null && downloadFileHash != null && !currentFileHash.contentEquals(
-                        downloadFileHash
-                    )
-                ) {
-                    // if the hashes match, it's the same file otherwise replace it with the new one
-                    val mediaFile = formSource.fetchMediaFile(downloadUrl)
-                    interuptablyWriteFile(
-                        mediaFile,
-                        tempMediaFile,
-                        tempDir,
-                        stateListener
-                    )
-                } else {
-                    // exists, and the hash is the same
-                    // no need to download it again
-                    Timber.i(
-                        "Skipping media file fetch -- file hashes identical: %s",
-                        finalMediaFile.absolutePath
-                    )
-                }
+                val mediaFile = formSource.fetchMediaFile(mediaFile.downloadUrl)
+                interuptablyWriteFile(mediaFile, tempMediaFile, tempDir, stateListener)
             }
+        }
+    }
+
+    private fun searchForExistingMediaFile(
+        formToDownload: ServerFormDetails,
+        mediaFile: MediaFile
+    ): File? {
+        val allFormVersions = formsRepository.getAllByFormId(formToDownload.formId)
+        return allFormVersions.map { form: Form ->
+            File(form.formMediaPath, mediaFile.filename)
+        }.firstOrNull { file: File ->
+            val currentFileHash = getMd5Hash(file)
+            val downloadFileHash = validateHash(mediaFile.hash)
+            file.exists() && currentFileHash.contentEquals(downloadFileHash)
         }
     }
 
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
@@ -52,12 +52,8 @@ class FormMediaDownloader(
             File(form.formMediaPath, mediaFile.filename)
         }.firstOrNull { file: File ->
             val currentFileHash = getMd5Hash(file)
-            val downloadFileHash = validateHash(mediaFile.hash)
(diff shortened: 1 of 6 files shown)


collect_app/src/main/java/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -466,4 +467,101 @@ public static String getFilenameError(String filename) {
...
     }
+
+    /**
+     * Common routine to take a downloaded document save the contents in the file
+     * 'file'.
+     * <p>
+     * The file is saved into a temp folder and is moved to the final place if everything
+     * is okay, so that garbage is not left over on cancel.
+     */
+    public static void interuptablyWriteFile(InputStream inputStream, File destinationFile, File tempDir, OngoingWorkListener listener)
```
seadowg, line 478: I'm still not sure if we actually need to be this cautious - I think we could just clean up after ourselves on a cancel, but I'm still not feeling confident enough to rip the extra temp dir out.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.odk.collect.forms.FormSource
+import org.odk.collect.forms.FormSourceException
+import org.odk.collect.forms.FormsRepository
+import org.odk.collect.forms.MediaFile
+import org.odk.collect.shared.strings.Md5.getMd5Hash
+import timber.log.Timber
+import java.io.File
+import java.io.IOException
+
+class FormMediaDownloader constructor(
```
seadowg: We're getting to a place where it feels like the best strategy for `ServerFormDownloader` is going to be breaking it into several components that we can just chain together in a "pipeline". Looking at this file in isolation, it's much easier to understand media file downloading that when it's surrounded by the whole form download process's logic.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -10,83 +9,54 @@ import org.odk.collect.forms.FormSourceException
...
-                    copyFile(existingFileInOtherVersion, tempMediaFile)
-                } else {
-                    val mediaFile = formSource.fetchMediaFile(downloadUrl)
-                    interuptablyWriteFile(mediaFile, tempMediaFile, tempDir, stateListener)
-                }
+            val existingFile = searchForExistingMediaFile(formToDownload, mediaFile)
+            if (existingFile != null) {
+                copyFile(existingFile, tempMediaFile)
             } else {
-                val currentFileHash = getMd5Hash(finalMediaFile)
```
seadowg: I realized (although I could be wrong) that we don't need to consider the "file we're going to download is already there" and "a version of this form on device already has this file" as separate cases. The latter includes the former.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -0,0 +1,63 @@
...
+    fun download(
+        formToDownload: ServerFormDetails,
+        files: List<MediaFile>,
+        tempMediaPath: String,
+        tempDir: File,
+        stateListener: OngoingWorkListener?
+    ) {
+        val tempMediaDir = File(tempMediaPath).also { it.mkdir() }
+
+        for (i in files.indices) {
```
grzesiek2010: I would use `forEachIndexed` like
`files.forEachIndexed {index, mediaFile ->`

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val formSource: FormSource
+) {
+
+    @Throws(IOException::class, FormSourceException::class, InterruptedException::class)
+    fun download(
+        formToDownload: ServerFormDetails,
+        files: List<MediaFile>,
+        tempMediaPath: String,
+        tempDir: File,
+        stateListener: OngoingWorkListener?
```
grzesiek2010: I think we can make it not nullable?

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+        val tempMediaDir = File(tempMediaPath).also { it.mkdir() }
+
+        for (i in files.indices) {
+            stateListener?.progressUpdate(i + 1)
+
+            val mediaFile = files[i]
+            val tempMediaFile = File(tempMediaDir, mediaFile.filename)
+
+            val existingFile = searchForExistingMediaFile(formToDownload, mediaFile)
```
grzesiek2010: More kotlin version would be:
```
            searchForExistingMediaFile(formToDownload, mediaFile)?.let {
                copyFile(it, tempMediaFile)
            } ?: run {
                val mediaFile = formSource.fetchMediaFile(mediaFile.downloadUrl)
                interuptablyWriteFile(mediaFile, tempMediaFile, tempDir, stateListener)
            }
```
but it's a matter of taste.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
```diff
@@ -0,0 +1,63 @@
...
+            File(form.formMediaPath, mediaFile.filename)
+        }.firstOrNull { file: File ->
+            val currentFileHash = getMd5Hash(file)
+            val downloadFileHash = validateHash(mediaFile.hash)
+            file.exists() && currentFileHash.contentEquals(downloadFileHash)
+        }
+    }
+
+    private fun validateHash(hash: String?): String? {
+        return if (hash == null || hash.isEmpty()) null else hash
```
grzesiek2010: According to our implementation in `OpenRosaResponseParserImpl` media file hash can't be null or empty. If it is null or empty the whole file is skipped and not added to the list of media files. That means you should not need to validate it here.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -102,17 +98,18 @@ private void processOneForm(ServerFormDetails fd, FormDownloaderListener stateLi
 
             // download media files if there are any
             if (fd.getManifest() != null && !fd.getManifest().getMediaFiles().isEmpty()) {
-                downloadMediaFiles(tempMediaPath, stateListener, fd.getManifest().getMediaFiles(), tempDir, fileResult.file.getName());
+                FormMediaDownloader mediaDownloader = new FormMediaDownloader(formsRepository, formSource);
```
grzesiek2010, line 101: I think you can inject `FormMediaDownloader` via constructor instead of creating it here.

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
seadowg: I got this through the auto conversion and thought the `indices` thing was cool but yeah, I think `forEachIndexed` is probably easier to understand.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
seadowg: I think you're correct that it could be more concise. How about:

```
searchForExistingMediaFile(formToDownload, mediaFile).let {
    if (it != null) {
        copyFile(it, tempMediaFile)
    } else {
        val file = formSource.fetchMediaFile(mediaFile.downloadUrl)
        interuptablyWriteFile(file, tempMediaFile, tempDir, stateListener)
    }
}
```

I find multiline "elvis" (`?:`) statements hard to follow personally.

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
seadowg: Right now I'd opt not to as `FormMediaDownloder` doesn't have any unit tests (it's behaviour is driven out/verified by `ServerFormDownloader`) and the component using it doesn't have tests that swap it out for something else. In short, it's a "detail" of `ServerFormDownloader` still. It might be that we want to get to that point soon (`ServerFormDownloader` taking in "smaller" use case objects and just testing it does the right thing with them but didn't feel like that was a path I wanted to go down right now - we can probably shift tests over if we feel like it next time we touch media downloading. For the moment I just wanted to move the code around to remove noise from the other file. 

One thought I have is we might want to actually break the whole thing into a pipeline of different use cases, so it could be that `FormMediaDownloader` ends up as a step in that.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
seadowg: Hmmm good point. I don't know if we're ensuring that for media files when we parse everything. I'll look into that.

seadowg: Ok I dug into it and added a test to make it clearer what's happening here: https://github.com/getodk/collect/pull/5025/commits/bb5b5d00b98cb56c50177ce16fc916d42f666082. Basically you're right that we don't need to validate. The hash is never `null` as `MediaFile#hash` is of type `String` and if a media file has an empty hash the whole list will be returned as null which results in `OpenRosaFormSource` throwing an exception. It might be that we want to look at introducing some custom types like `NonEmptyString` to deal with this kind of thing, but I'm not sure if I want to do that now.

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormMediaDownloader.kt
grzesiek2010: That's ok.

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**

I haven't noticed any regressions, verified cases:
- download form with media from central
- download another version of form, add new media 
- replace media on central and download 
- delete form from central, upload new version without uploaded media - download form update on collect - the newest version of form doesn't have media folder 
- central draft testing 
- delete media folder and download form update - media are re-downloaded
- update only media without version update 
- automatic download of previously downloaded form 
- delete media folder and download the same form twice - media are re-downloaded
- a different types of media files 
- track differences in usage of mobile data transmission between download 1st (~50MB) and 2nd (~2MB) form version with quite big media files
- form with external CSV file, do manual change in CSV and download another form version
- rename media in the form and rename file - file downloaded from the server
- rename media file only in files folder - file not copied 
- form with external CSV file, manual change in CSV in files and download new version - media file is downloaded
- form with external CSV file, change in CSV, upload to server and download new version - new version of csv is downloaded
- form with default images

srujner:
**Tested with success!**

Issue #4947, opened 2021-12-30
Extract generic "select item from map" implementation from `FormMapActivity`


seadowg:
The goal here is to end up with reusable code which we can use to build the [GeoJSON features](https://github.com/getodk/collect/issues/4943) with. 

Some thoughts on what this could look:

* We could probably end up with a new Activity in the `geo` module
* The new Activity could take in a list of "selectable items" as an extra and return a single one of those items as a result
* The Activity will also need to take a title (form name) and action ("Edit Saved Form") and action icon in as an extra
* "Selectable items" will need to specify name, color and other details in some generic form

PR #5029, opened 2022-02-16
Extract navigation code from `FormMapActivity`

seadowg:
Work towards #4947 
~~Blocked by #5020~~

I wanted to do an early check in with the first steps in this refactor, so it's not a massive review or an overly long-running branch. The first steps I took here was to extract the navigation code (Activity starting essentially) from `FormMapActivity` so it could be configured from the calling Activity (in this case `FillBlankFormActivity`). 

I think the next step thing to focus on will be extracting the data loading from `FormMapActivity` so that form instances are passed in (as a generic data type that can be rendered on the map for selection) rather than passing in a form ID.

New and existing tests.

Comments inline and above.

No change in behaviour! Just moving things around. I think we should avoid QA on this and just merge one a reviewer is happy. We can do some QA on the selection map once we the issue finished off.


geo/src/main/java/org/odk/collect/geo/SelectionMapActivity.kt
```diff
@@ -0,0 +1,38 @@
+package org.odk.collect.geo
+
+import android.app.Activity
+import android.content.Intent
+import androidx.activity.result.contract.ActivityResultContract
+import org.odk.collect.strings.localization.LocalizedActivity
+
+abstract class SelectionMapActivity : LocalizedActivity() {
```
seadowg, line 8: I think this should eventually become non `abstract` and all the config needed to select an item from a list of things will be passed using an implementation of `SelectItemFromMap`. Not quite there yet though!

collect_app/src/main/java/org/odk/collect/android/formmanagement/FormNavigator.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.odk.collect.android.external.InstancesContract
+import org.odk.collect.android.projects.CurrentProjectProvider
+import org.odk.collect.android.utilities.ApplicationConstants.BundleKeys.FORM_MODE
+import org.odk.collect.android.utilities.ApplicationConstants.FormModes.VIEW_SENT
+import org.odk.collect.forms.instances.Instance
+import org.odk.collect.forms.instances.InstancesRepository
+import org.odk.collect.settings.SettingsProvider
+import org.odk.collect.settings.keys.ProtectedProjectKeys.KEY_EDIT_SAVED
+
+class FormNavigator(
```
seadowg, line 15: It does feel like there might be scope for us to introduce a more general `Navigator` component in the future. Android provides a navigation component, but it's only really aimed at "single Activity/multi Fragment" setups.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
```diff
@@ -66,18 +64,19 @@
 /**
  * Show a map with points representing saved instances of the selected form.
  */
-public class FormMapActivity extends CollectAbstractActivity {
+public class FormMapActivity extends SelectionMapActivity {
```
grzesiek2010, line 67: You converted some classes to kotlin maybe it's good time to do the same with this class too?

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+        if (formSyncTask != null) {
+            formSyncTask!!.setDiskSyncListener(this)
+            if (formSyncTask!!.status == AsyncTask.Status.FINISHED) {
+                syncComplete(formSyncTask!!.statusMessage)
+            }
+        }
+    }
+
+    override fun onPause() {
+        if (formSyncTask != null) {
```
grzesiek2010: Could be just: `formSyncTask?.setDiskSyncListener(null)`

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+                }
+
+                override fun denied() {}
+            }
+        )
+    }
+
+    override fun onResume() {
+        super.onResume()
+        if (formSyncTask != null) {
```
grzesiek2010: Could be:
```
        formSyncTask?.let {
            it.setDiskSyncListener(this)
            if (it.status == AsyncTask.Status.FINISHED) {
                syncComplete(it.statusMessage)
            }   
        }
```

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    private fun init() {
+        setupAdapter()
+
+        // DiskSyncTask checks the disk for any forms not already in the content provider
+        // that is, put here by dragging and dropping onto the SDCard
+        formSyncTask = lastCustomNonConfigurationInstance as FormSyncTask?
+        if (formSyncTask == null) {
+            Timber.i("Starting new disk sync task")
+            formSyncTask = FormSyncTask()
```
grzesiek2010: Could be:
```
            formSyncTask = FormSyncTask().apply {
                setDiskSyncListener(this@FillBlankFormActivity)
                execute()
            }
```

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+                this,
+                FormNavigator(
+                    currentProjectProvider,
+                    settingsProvider,
+                    instancesRepositoryProvider::get
+                )
+            )
+        )
+
+        val blankFormsListViewModel = ViewModelProvider(this, blankFormsListViewModelFactory).get(
```
grzesiek2010: Could be:
`val blankFormsListViewModel = ViewModelProvider(this, blankFormsListViewModelFactory)[BlankFormsListViewModel::class.java]`

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+                currentProjectProvider.getCurrentProject().uuid,
+                idFormsTable
+            )
+            val action = intent.action
+            if (Intent.ACTION_PICK == action) {
+                // caller is waiting on a picked form
+                setResult(RESULT_OK, Intent().setData(formUri))
+            } else {
+                // caller wants to view/edit a form, so launch formentryactivity
+                val intent = Intent(this, FormEntryActivity::class.java)
```
grzesiek2010: Could be:
```
                Intent(this, FormEntryActivity::class.java).apply {
                    action = Intent.ACTION_EDIT
                    data = formUri
                    putExtra(
                        ApplicationConstants.BundleKeys.FORM_MODE,
                        ApplicationConstants.FormModes.EDIT_SAVED
                    )
                    startActivity(this)   
                }
```

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+     */
+    override fun onItemClick(parent: AdapterView<*>?, view: View, position: Int, id: Long) {
+        if (allowClick(javaClass.name)) {
+            // get uri to form
+            val idFormsTable = listView.adapter.getItemId(position)
+            val formUri = FormsContract.getUri(
+                currentProjectProvider.getCurrentProject().uuid,
+                idFormsTable
+            )
+            val action = intent.action
```
grzesiek2010: There is no need to create this additional object. I would call 
`Intent.ACTION_PICK == intent.action`
below directly.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    companion object {
+        private const val FORM_CHOOSER_LIST_SORTING_ORDER = "formChooserListSortingOrder"
+    }
+}
+
+class SelectFormFromMap : SelectItemFromMap<Input>() {
+
+    override fun createIntent(context: Context, input: Input): Intent {
+        return Intent(context, FormMapActivity::class.java).also {
```
grzesiek2010, line 326: If you use apply instead of also you can call parameters/functions directly like (without `it.` or `intent.`):

```
putExtra(FormMapActivity.EXTRA_FORM_ID, formId)
            putExtra(
                SelectionMapActivity.EXTRA_NEW_ITEM,
                Intent(context, FormEntryActivity::class.java).apply {
                    action = Intent.ACTION_EDIT
                    data = FormsContract.getUri(projectId, formId)
                }
            )
```

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
seadowg: Good point. Best to leave it for the moment though as I'll be moving most of the code from `FormMapActivity` over to `SelectionMapActivity` and it will be simpler to deal with the conversion piece by piece I think.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
seadowg: Ah, yeah definitely better to inline this. Just for clarity though: `val action = intent.action` won't create a new object, it just creates a new reference to an existing object (the object returned by `intent.action`).

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
seadowg: I tend to prefer `also` over `apply` as it I'm not really a fan of the magic used to make it work (receiver added to a lambda signature). I feel like every time I see `apply` I have to go work out how it works under the hood 😆.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
grzesiek2010: Right it's not a new object of course but still it's better to inline.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.java
grzesiek2010: Ok if it's more convenient for you.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
grzesiek2010: I think it's ok not to understand how some methods that we use work under the hood. We believe they work well if they are provided for use. For the same reason we don't write test for functions from libraries - we believe they are tested.

seadowg: > For the same reason we don't write test for functions from libraries - we believe they are tested.

Agreed. If I can reword what I mean, I find the lack of `it` (or an explicit parameter) distracting for two reasons:

1. It seems unclear what the receiver is
2. If I (correctly) assume the receiver is the subject of `apply` call then it's unclear how it works to me (I basically forget that you can add a receiver type to function literals)

I definitely don't think it is bad form to use `apply`, but I personally prefer `also` or `let` as I find those more readable. Does that make sense?

grzesiek2010: Ok if you prefer `also` that's fine you will probably find more `apply` in my pull request but that is not a problem.

grzesiek2010 (approved):

PR #5030, opened 2022-02-16
Remember last used pen color

### LINKED ISSUE #5028: Default draw and annotate pen color to last used
Context [on the forum](https://forum.getodk.org/t/image-markup-ability-to-set-default-pen-colour/36279)

- [x] Given that I open a draw or annotate question in any form, in any project
    And I set the color to something other than black (color1)
    And I open another draw or annotate question in any form, in any project
    When I start drawing
    Then the line I produce is in color1

grzesiek2010:
Closes #5028

I tested the fix manually and also added some automated tests for the new code.

There is not much to discuss here. I used meta settings to store last used pen color and every time a user draws that color is set as a default one.

One potential thing we might need to consider is that the new draw package seems like a nice candidate for a separate module... what do you think @seadowg?

We should test [DrawWidget](https://docs.getodk.org/form-question-types/#draw-widget), [SignatureWIdget](https://docs.getodk.org/form-question-types/#signature-widget) and [AnnotateWidget](https://docs.getodk.org/form-question-types/#image-widget-with-annotation)

No.

No.


collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
```diff
@@ -0,0 +1,39 @@
...
+            assertThat(it.isCancelable, `is`(true))
+        }
+    }
+
+    @Test
+    fun `pen color in view model should be set after clicking ok`() {
+        val scenario = DialogFragmentTest.launchDialogFragment(PenColorPickerDialog::class.java)
+        scenario.onFragment {
+            val testViewModel = mock<PenColorPickerViewModel>()
+            it.model = testViewModel
```
seadowg: I think it'd be better to override this with Dagger as otherwise there's a risk that something gets setup using the real view model during `Fragment` creation.

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.mockito.kotlin.verify
+import org.odk.collect.android.R
+import org.odk.collect.fragmentstest.DialogFragmentTest
+
+@RunWith(AndroidJUnit4::class)
+class PenColorPickerDialogTest {
+
+    @Test
+    fun `dialog should be cancelable`() {
+        val scenario = DialogFragmentTest.launchDialogFragment(PenColorPickerDialog::class.java)
```
seadowg: This should be using `FragmentScenarioRule`/`FragmentLauncherRule` now so that the scenarios are cleaned up properly.

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerViewModelTest.kt
```diff
@@ -0,0 +1,42 @@
...
+        viewModel = PenColorPickerViewModel(metaSettings)
+    }
+
+    @Test
+    fun `default pen color should be black`() {
+        assertThat(viewModel.penColor.value, `is`(Color.BLACK))
+    }
+
+    @Test
+    fun `last used pen color should be saved in meta settings`() {
```
seadowg: This should really be `setPenColor sets penColor`

seadowg (changes requested):
This looks good. Just a couple of changes here which I've highlighted inline.

> One potential thing we might need to consider is that the new draw package seems like a nice candidate for a separate module

Yup that sounds like a nice move here!

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerViewModelTest.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/draw/PenColorPickerDialog.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.rarepebble.colorpicker.ColorPickerView
+import org.odk.collect.android.injection.DaggerUtils
+import org.odk.collect.androidshared.R
+import javax.inject.Inject
+
+class PenColorPickerDialog : DialogFragment() {
+    @Inject
+    lateinit var factory: PenColorPickerViewModel.Factory
+
+    private lateinit var model: PenColorPickerViewModel
```
seadowg: Just discovered that there is a less noisy way to set up `ViewModels` that I think you'll like:

```kotlin
@Inject
lateinit var factory: PenColorPickerViewModel.Factory
private val model: PenColorPickerViewModel by activityViewModels { factory }
```

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun `dialog should be cancelable`() {
+        val scenario = launcherRule.launchDialogFragment(PenColorPickerDialog::class.java)
+        scenario.onFragment {
+            assertThat(it.isCancelable, `is`(true))
+        }
+    }
+
+    @Test
+    fun `pen color in view model should be set after clicking ok`() {
+        val factory = PenColorPickerViewModel.Factory(InMemSettingsProvider().getMetaSettings())
```
seadowg: I think you should still use a test double for `PenColorPickerViewModel` here like you were doing before. You can create a fake (using extension of mockito) `PenColorPickerViewModel.Factory` that returns a mock `PenColorPickerViewModel`  and then you check that `setPenColor` is called when you click "OK". 

I think you get clearer tests for Activity/Fragments when you treat them as a "unit" test of the component itself and use doubles for the ViewModel (or any other) layer. Then we don't need to care how the ViewModel itself works.

fragmentstest/src/main/java/org/odk/collect/fragmentstest/DialogFragmentTest.kt
```diff
@@ -15,25 +13,6 @@ import org.hamcrest.Matcher
...
 object DialogFragmentTest {
 
-    /**
-     * Creates a FragmentScenario using a test theme that supports AppCompat and MaterialComponents
-     * attributes but also won't explode (like MaterialComponents themes do).
-     *
-     * This uses [FragmentScenario.launch] here instead of [FragmentScenario.launchInContainer] as
-     * per docs at https://developer.android.com/guide/fragments/test#dialog.
-     */
-    internal fun <F : Fragment> launchDialogFragment(
```
seadowg: I'd prefer to keep this where it is, so we have the `DialogFragment` in one place. Hopefully we can get rid of this method once the bugs in Robolectric/AndroidX are fixed.

seadowg (changes requested):

fragmentstest/src/main/java/org/odk/collect/fragmentstest/DialogFragmentTest.kt
grzesiek2010: ok

collect_app/src/test/java/org/odk/collect/android/draw/PenColorPickerDialogTest.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/draw/PenColorPickerDialog.kt
grzesiek2010: Done.

seadowg (approved):

seadowg:
> One potential thing we might need to consider is that the new draw package seems like a nice candidate for a separate module

 Are you thinking of doing a follow-up to this PR with that or just leaving it for another time?

grzesiek2010:
> Are you thinking of doing a follow-up to this PR with that or just leaving it for another time?

it would be better to have it in a separate pr because from what I have noticed a lot of classes would need to be moved to shared modules like `androidshared` for example.

kkrawczyk123:
**Tested with success!**

Verified cases:
- Draw widget
- Signature widget
- Annotate widget
- fill submission
- edit submission
- update an answer
- fill another submission
- set color and exit from widget, open it once again
- set color and ignore changes
- reset, save and close
- set color, do not save and kill app 
- ignore changes and fill another submission
- delete the submission and fill another 
- delete form and download again
- light, dark theme
- use draw widgets in few projects
- signature and annotate widgets with default images
- reset all settings

srujner:
**Tested with success!**

PR #5038, opened 2022-02-24
Create reusable `SelectionMapFragment`

seadowg:
Work towards #4947 

There's been a bunch of discussions about whether we use an Activity or a Fragment as the reusable component here, and we decided a Fragment made the most sense. There's still a bunch of tidying up and solidifying to do but this felt like a good point to get a review before I go any further.

Next steps:

1. Convert `FormMapActivity` to Kotlin
1. Distribute tests driving out form map between a (rewritten) `FormMapActivityTest` and new tests for `SelectionMapFragment` and `SelectionSummarySheet`
1. Remove `FormMapViewModel` (making sure the above covers all of its logic)
1. Convert `FormMapActivity` to Kotlin
1. Reevaluate need for `FormMapTest` (reworking the tests might make it redundant or at least allow us to delete some of the tests)
1. Investigate moving shared map behaviour from `SelectionMapFragment` into `MapFragment` implementations (maintaining map zoom on recreate for instance)
1. Investigate extracting `maps` module to house `MapFragment`, `MapPoint` and `MapFragmentFactory` and creating modules for each implementation (classic **D**epenednecy **I**nversion **P**rinciple setup)

Managed to keep existing tests working (with some tweaks to make the boundaries less sporadic) and added a couple of new ones where it there was no coverage to safely move code around. Verified manually as well.

Discussed on Slack.

Again, this shouldn't result in any changes to behaviour. I think we can just merge this without QA (and look at the whole feature again when we're finished).


collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -282,7 +283,7 @@ public ReferenceLayerSettingsNavigator providesReferenceLayerSettingsNavigator()
                         @Provides
                         @Override
                         public MapFragmentFactory providesMapFragmentFactory() {
-                            return applicationComponent.mapProvider();
+                            return applicationComponent.mapFragmentFactory();
```
seadowg, line 286: I think we should really be using the `MapFragmentFactory` interface wherever this in injected. `MapProvider` is just the implementation of that interface.

collect_app/src/main/java/org/odk/collect/android/utilities/IconUtils.java
```diff
@@ -52,17 +52,18 @@ public static Bitmap getBitmap(Context context, int drawableId) {
         return bitmap;
     }
 
-    public static Drawable getSubmissionSummaryStatusIcon(Context context, String instanceStatus) {
+    public static int getSubmissionSummaryStatusIcon(String instanceStatus) {
```
grzesiek2010: I think `IconUtils` is not a good place for this method so maybe we could move it?

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+import org.odk.collect.geo.databinding.SelectionMapLayoutBinding
+import org.odk.collect.geo.databinding.SelectionSummarySheetLayoutBinding
+import org.odk.collect.geo.maps.MapFragment
+import org.odk.collect.geo.maps.MapFragment.ReadyListener
+import org.odk.collect.geo.maps.MapFragmentFactory
+import org.odk.collect.geo.maps.MapPoint
+import org.odk.collect.permissions.PermissionsProvider
+import javax.inject.Inject
+
+class SelectionMapFragment : Fragment() {
```
grzesiek2010, line 31: Now as it is a fragment we should clear binding `onDestroyView()` https://developer.android.com/topic/libraries/view-binding#fragments

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+        }
+    }
+
+    companion object {
+        const val REQUEST_SELECT_ITEM = "select_item"
+        const val RESULT_SELECTED_ITEM = "selected_item"
+        const val RESULT_CREATE_NEW_ITEM = "create_new_item"
+
+        private val MAP_CENTER_KEY = "map_center"
+        private val MAP_ZOOM_KEY = "map_zoom"
```
grzesiek2010: Those two might be consts.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
+package org.odk.collect.geo
```
grzesiek2010, line 1: Now in this file we have a few classes  what makes it big are you going to split it?

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+        for (item in items) {
+            val point = MapPoint(item.latitude, item.longitude)
+            val featureId = map.addMarker(point, false, MapFragment.BOTTOM)
+
+            map.setMarkerIcon(
+                featureId,
+                if (featureId == selectionMapViewModel.getSelectedItemId()) item.largeIcon else item.smallIcon
+            )
+
+            itemsByFeatureId.put(featureId, item)
```
grzesiek2010: Could be `itemsByFeatureId[featureId] = item`

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+            binding.action.visibility = View.GONE
+        } else if (item.action != null) {
+            binding.action.text = item.action.text
+            binding.action.chipIcon = ContextCompat.getDrawable(context, item.action.icon)
+            binding.action.visibility = View.VISIBLE
+            binding.info.visibility = View.GONE
+        }
+    }
+
+    private fun onActionClick(view: View) {
```
grzesiek2010: view is unused.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+        private val MAP_ZOOM_KEY = "map_zoom"
+    }
+}
+
+class SelectionMapViewModel : ViewModel() {
+
+    private var mapTitle = MutableLiveData<String>()
+    private var mappableItems = MutableLiveData<List<MappableSelectItem>>(emptyList())
+    private var itemCount = MutableLiveData(0)
+    private var selectedItemId = -1
```
grzesiek2010: Can we make it a const? (that -1) or use null to represent this state I remember you saying recently that you prefer null over -1 like here

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    private fun update(items: List<MappableSelectItem>) {
+        if (map == null) {
+            return
+        }
+
+        updateFeatures(items)
+
+        if (!viewportInitialized && !points.isEmpty()) {
```
grzesiek2010: You can use `points.isNotEmpty()` instead of `!points.isEmpty()`

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    private fun onClick() {
+        if (summarySheetBehavior.getState() == BottomSheetBehavior.STATE_EXPANDED) {
+            summarySheetBehavior.setState(BottomSheetBehavior.STATE_HIDDEN)
+        }
+    }
+
+    private fun update(items: List<MappableSelectItem>) {
+        if (map == null) {
```
grzesiek2010: `map` is lateinint so it shouldn't be null?

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -0,0 +1,416 @@
...
+                summarySheetBehavior.setState(BottomSheetBehavior.STATE_EXPANDED)
+            }
+
+            selectionMapViewModel.setSelectedItemId(featureId)
+        }
+    }
+
+    private fun onClick() {
+        if (summarySheetBehavior.getState() == BottomSheetBehavior.STATE_EXPANDED) {
+            summarySheetBehavior.setState(BottomSheetBehavior.STATE_HIDDEN)
```
grzesiek2010: You can call just state. Please fix other `setSate` too.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -218,11 +197,15 @@ class FillBlankFormActivity :
...
         permissionsProvider.requestLocationPermissions(
             this,
-            object : PermissionListener {
+            object :
+                PermissionListener {
                 override fun granted() {
-                    selectFormFromMap.launch(
-                        Input(currentProjectProvider.getCurrentProject().uuid, id)
-                    )
+                    val intent =
```
grzesiek2010: This could be just:
```
Intent(this@FillBlankFormActivity, FormMapActivity::class.java).also {
                        it.putExtra(FormMapActivity.EXTRA_FORM_ID, id)
                        startActivity(it)
                    }
```

instead of:

```                   
 val intent =
                        Intent(this@FillBlankFormActivity, FormMapActivity::class.java).also {
                            it.putExtra(FormMapActivity.EXTRA_FORM_ID, id)
                        }

                    startActivity(intent)

```

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
```diff
@@ -218,11 +197,15 @@ class FillBlankFormActivity :
     private fun onMapButtonClick(id: Long) {
         permissionsProvider.requestLocationPermissions(
             this,
-            object : PermissionListener {
+            object :
+                PermissionListener {
```
grzesiek2010: We don't have any check for this but to me it's weird to have that in two lines I would rather have: 
`object : PermissionListener {`

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/utilities/IconUtils.java
seadowg: Yeah, good point. I'll work out somewhere else for it to live.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: Huh interesting. My thinking was that there would be no need for this as recreation would simply replace the reference. There's a suggestion in [this SO](https://stackoverflow.com/questions/65295104/android-view-binding-clear-binding-in-fragment-lifecycle) that one potential reason to `null` out the reference is to prevent leaks if the `Fragment` hangs around after the view is destroyed. However, it doesn't seem like that's a case anyone has a tangible example of (and [this linked article](https://www.techyourchance.com/android-fragment-lifecycle/) even refers to the practice as "premature optimization". But of course, the official docs do state that: 

> Fragments outlive their views. Make sure you clean up any references to the binding class instance in the fragment's [onDestroyView()](https://developer.android.com/reference/kotlin/androidx/fragment/app/Fragment#ondestroyview) method.

It honestly feels like View Binding wasn't though out for Fragments properly (like everything else), and they've had to add that in due to memory problems people have run into. It's not even in their [code samples](https://github.com/android/architecture-components-samples/blob/main/GithubBrowserSample/app/src/main/java/com/android/example/github/ui/user/UserFragment.kt) for instance.

After getting grumpy about all this and looking through Android's issue tracker I found [this discussion](https://issuetracker.google.com/issues/152333892) that links [this article](https://issuetracker.google.com/issues/152333892) that demonstrates using`<binding class>.bind(view)` in `onViewCreated` to avoid create binding fields. I'm going to file an issue around that being added to the View Binding docs (https://developer.android.com/topic/libraries/view-binding) and go with that to avoid needing an `onDestroyView`.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: Yeah it feels messy. How about we discuss the package structure as part of the next PR, so it's as closed to "finished" as possible? That way we hopefully avoid any churn on things moving around.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: Oh thanks yes! I'd meant to clean this up but it escaped me.

collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
seadowg: I went with inlining `Intent` instead. I think it's good to make sure `startActivity` is front and center as it's the code here's main feature.

grzesiek2010: I would use my syntax but ok it's a matter of taste.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
grzesiek2010: Ok

grzesiek2010 (approved):

Issue #4870, opened 2021-10-19
Collect crashes when Try a demo project after deleting all projects on Android 5.1


kkrawczyk123:
#### Software and hardware versions 
Collect v current master Android v 5.1, device used...

#### Problem description
When testing #4855 I have noticed crash on Android 5.1 device only when deleted all projects and then run demo project.

#### Steps to reproduce the problem
1. Create project
2. Delete project
3. Click on Try a demo

#### Expected behavior
Collect should not crash.

#### Other information 
stacktrace:
`2021-10-19 10:34:40.451 10957-10957/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 10957
    android.database.sqlite.SQLiteDiskIOException: disk I/O error (code 1802)
        at android.database.sqlite.SQLiteConnection.nativeExecuteForCursorWindow(Native Method)
        at android.database.sqlite.SQLiteConnection.executeForCursorWindow(SQLiteConnection.java:845)
        at android.database.sqlite.SQLiteSession.executeForCursorWindow(SQLiteSession.java:836)
        at android.database.sqlite.SQLiteQuery.fillWindow(SQLiteQuery.java:62)
        at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:144)
        at android.database.sqlite.SQLiteCursor.getCount(SQLiteCursor.java:133)
        at org.odk.collect.android.database.instances.DatabaseInstancesRepository.getCountByStatus(DatabaseInstancesRepository.java:109)
        at org.odk.collect.android.formmanagement.InstancesAppState.update(InstancesAppState.kt:36)
        at org.odk.collect.android.activities.viewmodels.MainMenuViewModel.lambda$refreshInstances$0$MainMenuViewModel(MainMenuViewModel.java:121)
        at org.odk.collect.android.activities.viewmodels.MainMenuViewModel$$ExternalSyntheticLambda1.get(Unknown Source)
        at org.odk.collect.async.CoroutineScheduler$immediate$1$result$1.invokeSuspend(CoroutineScheduler.kt:16)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
[...]

PR #5043, opened 2022-02-25
Make sure database files always exist when used

### LINKED ISSUE #5042: The app crashes after deleting and adding a project again
#### Problem description
The app crashes after deleting and adding a project again.

```
```

1. Crete Demo project
2. Download one form
3. Delete the project
4. Create Demo project again
5. Try to download forms.

#### Other information 
The issue is similar to https://github.com/getodk/collect/issues/4870 but it looks like not the same.

grzesiek2010:
Closes #5042 

I tested the app manually and also added automated tests.

We store `dbHepers` [in a map that lives as long as the application does](https://github.com/getodk/collect/pull/5043/files#diff-6f4c63e8a1ee58d804a57e12e4101c5c12e6df35c95a318ab164e0f7b7156353R50). Thanks to that we don't need to create new (duplicate) objects every time. The problem was that if database file was deleted it wasn't recreated if we accessed such an object saved in memory (because the object was taken from that map).
Now every time we access `dbHelper` I check if its file exist.  

This change should just fix the issue and have no side effects. Please verify briefly accessing our databases so basicly downloading and opening forms. Please also verify this issue: https://github.com/getodk/collect/issues/4870 because I think it might be also addressed by this pull request.

No.

No.


collect_app/src/test/java/org/odk/collect/android/database/DatabaseFilePresenceTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.odk.collect.android.database.forms.DatabaseFormsRepository
+import org.odk.collect.shared.TempFiles.createTempDir
+import java.io.File
+
+@RunWith(AndroidJUnit4::class)
+// https://github.com/getodk/collect/issues/5042
+class DatabaseFilePresenceTest {
```
seadowg: Could this not be a direct test of `DatabaseConnection` so we're not involving a repo?

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
```diff
@@ -26,29 +27,28 @@ open class DatabaseConnection(
...
-                migrator
-            )
+    private val dbHelper: SQLiteOpenHelper
+        get() {
+            val databasePath = path + File.separator + name
+            if (openHelpers.containsKey(databasePath) && !File(databasePath).exists()) {
+                openHelpers.remove(databasePath)
+            }
+
+            return openHelpers.getOrPut(databasePath) {
```
seadowg, line 39: Maybe we could just use `if-else` (either get one of put and return) here instead of `getOrPut` now? I think that'd be clearer.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
```diff
@@ -26,29 +27,28 @@ open class DatabaseConnection(
...
-                name,
-                null,
-                databaseVersion,
-                migrator
-            )
+    private val dbHelper: SQLiteOpenHelper
+        get() {
+            val databasePath = path + File.separator + name
+            if (openHelpers.containsKey(databasePath) && !File(databasePath).exists()) {
+                openHelpers.remove(databasePath)
```
seadowg: Do you think we need to call `close()` on this invalid "open" helper now? It feels like we must be leaking something otherwise.

seadowg (changes requested):
Wow. This was subtle. Great detective work 🕵️‍♂️. There are a couple of things to discuss/tweak that I've left comments about inline.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
grzesiek2010: `getOrPut` is better because it's null safe with if-else the syntax would look worse unless I'm missing something.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
grzesiek2010: Yeah I think you are right.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
seadowg: Ah you're right we'd always end up having to deal with the nullable result of `get`

collect_app/src/test/java/org/odk/collect/android/database/DatabaseFilePresenceTest.kt
grzesiek2010: I think you meant: **Could this ~~not~~ be** ?

seadowg: Yes. 

Sorry this is ridiculous (probably colloquial?) use of the English language which where I'm really asking: "**is there a reason this is not** a direct test of DatabaseConnection so we're not involving a repo". I guess saying "Could this not be X" is like saying "Why can't this be X". I've no idea where this comes from, and I'll try and avoid using it - as soon as you pointed it out I got very confused as to what I even meant!

grzesiek2010: Ah right... I was aware of such constructions in English but I didn't get enough sleep today to catch it probably :) 
Yes it can be DatabaseConnection test.

seadowg: haha yes it definitely does what English does best: be ambiguous. Probably one of those things that makes a lot more sense spoken (with emphasis and tone) than written. I'm also not sure if it's even valid, but native speakers are always the worst at sticking to the rules I guess.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseConnectionTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.odk.collect.android.database.forms.FormDatabaseMigrator
+import org.odk.collect.shared.TempFiles.createTempDir
+import java.io.File
+
+@RunWith(AndroidJUnit4::class)
+// https://github.com/getodk/collect/issues/5042
```
seadowg: We should move this down to the individual test so that it doesn't end up being out of context if we add a second test case.

seadowg (approved):
Looks good! One tweak to the test but this is ready for QA.

collect_app/src/test/java/org/odk/collect/android/database/DatabaseConnectionTest.kt
grzesiek2010: Done.

kkrawczyk123:
**Tested with success!**
Verified on Androids 5.1, 8.1 and 10

Verified cases:
- unfortunately, issue #4870 has not been fixed and it's still visible on Android 5.1
- demo project
- another Central project, Google Drive project
- download forms, fill, send, auto-send, match exactly
- add project manually, using qr code

srujner:
**Tested with success!**
Verified also on Android 11

PR #5046, opened 2022-03-02
Remove mention of bot from contributing guide

seadowg:
The ODK bot has been retired, so we shouldn't encourage people to use it any longer.


lognaturel (approved):
Thank you!!

PR #5047, opened 2022-03-03
Only trigger InstanceAutoSender if autosend is enabled in settings or the form requests autosending

grzesiek2010:
I've added automated tests.

In https://github.com/getodk/collect/pull/5024/files#diff-988ec72f11211c5e52159ad89b1a182ee276edd0482b79d4eb8e027b7b51e742 I made a mistake refactoring `AutoSendTaskSpec`. The problem is that `InstanceAutoSender` is always called even it it shouldn't when `autosend` is disabled in settings and there are no forms with `autosend` used on a forma level: https://github.com/getodk/collect/pull/5024/files#diff-988ec72f11211c5e52159ad89b1a182ee276edd0482b79d4eb8e027b7b51e742R76
It hasn't been spotted because:
- we didn't have tests for it
- the fact that we trigger that task doesn't mean anything will be sent, we have another check point which verifies if it should happen: https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.java#L91

So it looks like the bug is not severe but it's definitely a bug we should fix ASAP.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/autosend/FormLevelAutoSendChecker.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/autosend/FormLevelAutoSendChecker.kt
@@ -1,0 +1,17 @@
+package org.odk.collect.android.backgroundwork.autosend
+
+import org.odk.collect.android.utilities.FormsRepositoryProvider
+import org.odk.collect.forms.Form
+
+class FormLevelAutoSendChecker(private val formsRepositoryProvider: FormsRepositoryProvider) {
+
+    /**
+     * Returns true if at least one form currently on the device specifies that all of its filled
+     * forms should auto-send no matter the connection type.
+     */
+    fun isAutoSendEnabled(projectId: String): Boolean {
+        return formsRepositoryProvider.get(projectId).all.any { form: Form ->
+            form.autoSend.toBoolean()
+        }
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
@@ -17,6 +17,8 @@ import android.content.Context
 import android.os.Environment
 import androidx.work.BackoffPolicy
 import androidx.work.WorkerParameters
+import org.odk.collect.android.backgroundwork.autosend.FormLevelAutoSendChecker
+import org.odk.collect.android.backgroundwork.autosend.GeneralAutoSendChecker
 import org.odk.collect.android.injection.DaggerUtils
 import org.odk.collect.android.instancemanagement.InstanceAutoSender
 import org.odk.collect.android.network.NetworkStateProvider
@@ -26,8 +28,6 @@ import org.odk.collect.async.WorkerAdapter
 import org.odk.collect.settings.SettingsProvider
 import java.util.function.Supplier
 import javax.inject.Inject
-import org.odk.collect.android.backgroundwork.autosend.FormLevelAutoSendChecker
-import org.odk.collect.android.backgroundwork.autosend.GeneralAutoSendChecker
 
 class AutoSendTaskSpec : TaskSpec {
     @Inject
--- a/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
@@ -14,8 +14,6 @@
 package org.odk.collect.android.backgroundwork
 
 import android.content.Context
-import android.net.ConnectivityManager
-import android.net.NetworkInfo
 import android.os.Environment
 import androidx.work.BackoffPolicy
 import androidx.work.WorkerParameters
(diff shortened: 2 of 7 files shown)


collect_app/src/main/java/org/odk/collect/android/backgroundwork/autosend/FormLevelAutoSendChecker.kt
```diff
@@ -0,0 +1,17 @@
+package org.odk.collect.android.backgroundwork.autosend
+
+import org.odk.collect.android.utilities.FormsRepositoryProvider
+import org.odk.collect.forms.Form
+
+class FormLevelAutoSendChecker(private val formsRepositoryProvider: FormsRepositoryProvider) {
```
grzesiek2010, line 6: I added those two classes to make testing easier otherwise I would end up with plenty tests in `AutoSendTaskSpec` because I would need to verify all those variations where `autosend` is enabled/disabled in settings and enabled/disabled on a form level.

seadowg (changes requested):
We had a chat about the direction to take this on Slack, but I'll summarize my thoughts here as well, so it's easier to refer to: 

* I think that we should pull as much logic as possible out of the `AutoSendTaskSpec` so that all it's responsible for is integrating our logic/use case (here `InstanceAutoSender`) with the `Scheduler`. For me, this means that all it should do is call `InstanceAutoSender#autoSendInstances` and that should work out whether (and what) to send.
* `AutoSendTaskSpec` should be responsible for working out if the network conditions are correct to run the task or not as that feels like it has more to do with the "scheduling" of the job than the actual work of doing the job itself. We should be able to move that logic to `WorkManager` itself by using [network type `Constraint`](https://developer.android.com/reference/androidx/work/NetworkType) - I believe `NetworkType.UNMETERED` is equivalent to "Wifi only" for all intents and purposes.
* Given the above, I don't think we'd need separate `...Checker` use cases. That said, separating the "question" from the "work" itself is a nice strategy for simplifying code (often called "command query separation" I guess), so we might still want at least the `FormLevelAutoSendChecker` (the other should be replaced by `WorkManager` scheduling hopefully).
* One final thought (that we didn't discuss on Slack) is that we could change the naming (and intention) of `InstanceAutoSender` to mirror `FormsUpdater` (`InstanceSender`). I imagine once we start reworking manual send to happen in the background, it'd be useful to have one class that handles sending Instances (auto send or not). This class would also probably benefit from the `ProjectSandbox` helper that `FormsUpdater` uses.

grzesiek2010:
> I think that we should pull as much logic as possible out of the AutoSendTaskSpec so that all it's responsible for is integrating our logic/use case (here InstanceAutoSender) with the Scheduler. For me, this means that all it should do is call InstanceAutoSender#autoSendInstances and that should work out whether (and what) to send.

I agree with this. This is a different solution but it seems fine too.

> AutoSendTaskSpec should be responsible for working out if the network conditions are correct to run the task or not as that feels like it has more to do with the "scheduling" 

I think you missed **not**? Above you said it shouldn't be its responsibility...

> We should be able to move that logic to WorkManager itself by using [network type Constraint](https://developer.android.com/reference/androidx/work/NetworkType) - I believe NetworkType.UNMETERED is equivalent to "Wifi only" for all intents and purposes.

I don't think we can use it like that. If you have `auto-send` enabled on a form level the task should be performed no matter what option is selected in our settings. So we can't block starting the task specifying `NetworkType` that would be determined based on our settings.

seadowg:
> I think you missed not? Above you said it shouldn't be its responsibility...

Oooft yes sorry that's a bad word to miss there.

> I don't think we can use it like that. If you have auto-send enabled on a form level the task should be performed no matter what option is selected in our settings.

Argh I knew there was a reason we didn't move that when we were doing Match Exactly. @lognaturel we've probably talked about this before but what's the historical reasoning for having auto send at a form level? It adds complexity to our implementation and I can't personally fathom it making more sense that using the setting (given how commonly we see network constraints on projects).

lognaturel:
> what's the historical reasoning for having auto send at a form level?

I wasn't there when it was added so I can't say for sure. I think it was kind of a precursor to what we do now with configuration QR codes. Before those, the best way to ensure a certain usage pattern would have been through form definitions. Even now, settings could be changed so having the directive at the form level is the strongest directive. It might also make sense for forms to have different behaviors -- a small text-only form might be great to auto send but a form with many heavy image attachments may be better to deliberately send when e.g. connected to wifi.

That said, it's not documented [in ODK docs](https://docs.getodk.org/xlsform/#) or [in XLSForm docs](https://xlsform.org/en/#settings-worksheet) as far as I can tell. So maybe it's a good candidate for gathering some analytics about and likely putting on a deprecation path. The only place I find it is [in the spec](https://getodk.github.io/xforms-spec/#submission-attributes).

lognaturel:
> it's definitely a bug we should fix ASAP

There's no change in behavior, right? Is there likely to be a performance issue or anything user-visible that could be bad?

Maybe we want to shift this to a discussion on what the behavior should be and how we can test the desired behavior and then come back to any updates to the code that might be improvements.

grzesiek2010:
> So maybe it's a good candidate for gathering some analytics about

That makes sense I can add that.

> Maybe we want to shift this to a discussion on what the behavior should be and how we can test the desired behavior and then come back to any updates to the code that might be improvements.

I would prefer doing this now as I have fresh knowledge about that background tasks since I have worked with them recently a lot.

@seadowg is there anything you would like to add here? We can't use that network type constraints but other than that I think the direction is pretty clear.

seadowg:
@grzesiek2010 Sorry for the big delay getting back to this. Are you still feeling like you'd rather continue with this PR now rather than picking it up some time in the future? If so, I can take another look.

grzesiek2010:
> Are you still feeling like you'd rather continue with this PR now rather than picking it up some time in the future? If so, I can take another look.

No I'm not going to work on it this or next week but I would like to get back to it in May. I think you will have enough time in this case.

seadowg:
@grzesiek2010 great! I'll switch it to a draft, and you can ping me when you'd like me to look again.

grzesiek2010:
Closed in favor of https://github.com/getodk/collect/pull/5277

Issue #5050, opened 2022-03-04
Reduce app size


lognaturel:
#### Software and hardware versions 
After Collect v2022.1

#### Problem description
Commits added after 2022.1 was released have increased the app size by 2mb (to 14mb).

My hunch is that this is because many of the modules are not minified. The recent changes pulled out a geo module which imports a lot of dependencies.

seadowg:
> My hunch is that this is because many of the modules are not minified. The recent changes pulled out a geo module which imports a lot of dependencies.

Modules should not be minified (they should have `minifyEnabled false`). When you're building an APK you're compiling your app module (ours is `collect_app`) and then minification gets run on the output of that compilation ([example SO](https://stackoverflow.com/a/50819864/11400117)). It could be that we've broken that setup somehow or that our proguard rules are allowing something big to not get minified (moving things around could definitely exacerbate a wild card rule). Another possibility is that we've included some resources in the wrong place.

seadowg:
After a `git bisect` it looks like the offending commit is 538e7c68f9408306bf691da1530a69122e2ed9d6. It's likely that something from `json-schema-validator` is the problem.

seadowg:
Yup! Removing the `json-schema-validator` dependency but leaving the reset of the commit the same (other than commenting out the references to the library code) drops the APK size back down to 12.3MB.

seadowg:
We get a much smaller increase (to 12.8MB) using https://github.com/networknt/json-schema-validator which is less popular but actually seems to be more up to date. Both libraries use Jackson, but it might be that this one is easier to minimize for some reason. I'll run tests etc with the new library to fully confirm it can be a drop-in replacement.

lognaturel:
Aha, I clearly had some proguard misconceptions all around. Thanks for getting me untwisted! I am absolutely shocked that json schema validation could add 2mb to the app. Glad to hear there's likely a drop-in replacement.

Feeling very glad we decided to do an early beta. 👍

seadowg:
Should have been closed by #5053

PR #5053, opened 2022-03-04
Fix proguard settings

### LINKED ISSUE #5052: Crash when scanning a QR code on release build
Collect v2022.2-beta.0

#### Problem description
Scanning a QR code using a debug build works but with a release build a runtime exception is thrown: `class "com.github.fge.jsonschema.keyword.validator.common.AdditionalItemsValidator" has no appropriate constructor (keyword "additionalItems"; expected a constructor with a JsonNode as an argument)`

Scan any settings QR code with a release build.

#### Expected behavior
Scanning QR codes works with a release build.

#### Other information 
I addressed this for the beta by adding the following to the `collect_app` proguard rules:
`-keep class com.github.fge.jsonschema.keyword.validator.** { *; }`

It feels like this should be handled within the settings module, though.

Note: keeping `AdditionalItemsValidator` is not enough -- there's at least one other class that gets pruned.

seadowg:
Closes #5052, #5050

Ran app and tests.

Comments inline.

Main thing to check is importing settings (or creating a project) from a QR code as the underlying framework we use for validation has changed.

### DIFF
--- a/buildSrc/src/main/java/dependencies/Dependencies.kt
+++ b/buildSrc/src/main/java/dependencies/Dependencies.kt
@@ -73,7 +73,7 @@ object Dependencies {
     const val slf4j_api = "org.slf4j:slf4j-api:2.0.0-alpha0"
     const val slf4j_timber = "com.arcao:slf4j-timber:3.1@aar"
     const val emoji_java = "com.vdurmont:emoji-java:5.1.1"
-    const val json_schema_validator = "com.github.java-json-tools:json-schema-validator:2.2.14"
+    const val json_schema_validator = "com.networknt:json-schema-validator:1.0.66"
 
     // Test dependencies
     const val junit = "junit:junit:4.13.2"
--- a/settings/consumer-rules.pro
+++ b/settings/consumer-rules.pro
@@ -1,0 +1,1 @@
+-keep class com.github.fge.jsonschema.keyword.validator.** { *; }
--- a/settings/src/main/resources/settings_schema.json
+++ b/settings/src/main/resources/settings_schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
   "$id": "https://getodk.com/settings.schema.json",
   "title": "Settings",
   "type": "object",
--- a/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
+++ b/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
@@ -1,23 +1,20 @@
 package org.odk.collect.settings.validation
 
-import com.github.fge.jackson.JsonNodeReader
-import com.github.fge.jsonschema.main.JsonSchemaFactory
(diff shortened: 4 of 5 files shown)


buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -73,7 +73,7 @@ object Dependencies {
     const val slf4j_api = "org.slf4j:slf4j-api:2.0.0-alpha0"
     const val slf4j_timber = "com.arcao:slf4j-timber:3.1@aar"
     const val emoji_java = "com.vdurmont:emoji-java:5.1.1"
-    const val json_schema_validator = "com.github.java-json-tools:json-schema-validator:2.2.14"
+    const val json_schema_validator = "com.networknt:json-schema-validator:1.0.66"
```
seadowg, line 76: This framework is less "starred" on Github but actually seems to be more active and brings in a far less heavy set of dependencies (its README mentions this being intentional as well).

settings/consumer-rules.pro
```diff
@@ -0,0 +1 @@
+-keep class com.github.fge.jsonschema.keyword.validator.** { *; }
```
seadowg, line 1: This was the rule @lognaturel had needed in the `collect_app` proguard rules to release the beta.

settings/src/main/resources/settings_schema.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
```
seadowg, line 2: https://github.com/networknt/json-schema-validator/ does not currently support the newest JSON schema version (`2020-12`). I think we should fix this as a contribution (issue here: https://github.com/networknt/json-schema-validator/issues/427) but right now it's not essential that we have it as far as I can see.

lognaturel (approved):

kkrawczyk123:
**Tested with success!**

Verified cases:
- compared size of apk PR 21,8 and 25,9 on the maser
- create project scanning qr code
- different types of servers: Central, Aggregate, Drive, Kobo, Ona 
- different settings combination
- update settings within the project through qr code
- import settings from file
- scan inverted qr code
- scan invalid qr code 
- use front barcode scanner

srujner:
**Tested with success!**

lognaturel:
Is it possible this wasn't tried on a release build? I can't get it to work, I always get the "QR code does not contain valid settings" message. It works fine on a debug build. I'll take a look and file an issue if I can't figure it out.

settings/consumer-rules.pro
lognaturel: Whoops, this should now be for the new library, I think.

seadowg:
@lognaturel as far as I remember I was **only** trying it on a release build (`selfSignedRelease` specifically). It's all manual though, so it's possible I messed something up.

lognaturel:
Or maybe you first verified the proguard rule and then swapped out the library? Either way, I merged https://github.com/getodk/collect/pull/5066 myself and all is good.

@kkrawczyk123, @srujner I'm wondering whether you should just always use `selfSignedRelease`? That would have helped us catch some of these errors earlier. We can talk about it on Tues.

PR #5054, opened 2022-03-04
Add release checklist

lognaturel:
Document some things that I do when preparing a release that might not be obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Please note that the `master` branch reflects ongoing development and is not pro
 ## Release cycle
 Releases can be requested by any community member and generally happen every 2 months.
 
-Before release we perform a "code freeze" (we stop merging pull requests) and then carry out regression testing. If any problems are found, the release is blocked until we can merge fixes. Once the process is complete, [@yanokwa](https://github.com/yanokwa) pushes the releases to the Play Store. The code is "unfrozen" after a short grace period to make hot fixing easier.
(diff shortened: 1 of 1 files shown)


lognaturel:
There are starting to be lots of open PRs! @grzesiek2010, do you think you could give this and https://github.com/getodk/collect/pull/5067 a quick sanity check? I don't think either has anything controversial. 🙏

README.md
```diff
@@ -233,22 +233,32 @@ This is encountered when Robolectric has problems downloading the jars it needs
...
+  RELEASE_STORE_FILE=/path/to/collect.keystore
+  RELEASE_STORE_PASSWORD=secure-store-password
+  RELEASE_KEY_ALIAS=key-alias
+  RELEASE_KEY_PASSWORD=secure-alias-password
+  ```
+
+- a `google-services.json` file in the `collect_app/src/odkCollectRelease` folder. The contents of the file are similar to the contents of `collect_app/src/google-services.json`.
+
+### Release checklist:
+- tag the build by [adding a release](releases).
```
grzesiek2010: This link doesn't seem to work

grzesiek2010 (changes requested):

README.md
lognaturel: It's a relative link that would work once the PR is merged. But I guess in this case there's only one set of releases so I've changed it to an absolute link.

grzesiek2010: > It's a relative link that would work once the PR is merged.

But the one we already have in https://github.com/getodk/collect doesn't work as well. The absolute one seems to work well.

grzesiek2010 (approved):

README.md
lognaturel: Oh, well then! Thanks for the fix.

Issue #4540, opened 2021-04-30
Allow manually-triggered submission to happen in background


lognaturel:
There are two parts to this: refactoring for maintainability and reconsidering the submission user experience.

Currently automatic sending and manual sending use almost entirely disjoint code paths. This duplication is not good and should be addressed as much as possible. Implementation note: look at using a `ForegroundService` via `WorkManager`.

Currently manually-triggered submissions are done synchronously and block the UI. 

In some cases this is good and desirable. For example, data collectors may be asked to submit high-priority submissions first and verify that they are all received before moving on to others (and maybe risk running out of credits). 

In other cases, even if sending is manually triggered, it could be better for data collectors to be able to do other work in parallel. One hypothetical scenario is one where some submissions will have big attachments and others won't. Maybe we want data collectors to submit small submissions while in the field and wait for big ones to be sent from the office wifi. In the case of the small submissions, they'd like to select them once a day, queue them for sending once there's connectivity, and go about the rest of their data collection. They can't use auto-send because they don't want the big submissions to be queued. This sounds plausible but I'm not sure how likely it is. If it's a real case, maybe we should consider alternative solutions like auto sending submissions below a certain size or actually exposing submission size in the UI.

## Backlog
   
- [ ] #4881
- [x] #4884
- [ ] #4885
- [x] #4907
- [ ] #4908
- [ ] Remove experimental background submissions setting and dead code

seadowg:
Quick thoughts on this:

* This would be a great way to drive out getting rid of the `InstanceUploaderTask` hierarchy and to push us to refactor/clean up/test the upload paths.
* An initial pass might be to generalize our auto send path and use `WorkManager` for uploading in both manual and auto cases - we could still have the UI block for manual uploads
* We could introduce an experimental setting that switches uploads to using a foreground service and not blocking the user - this would let us iterate on the feature and allow people to test in beta without any imperfections blocking a release

seadowg:
> Get real examples of scenarios where data collectors should not use auto-send but have other Collect-based work to do while submissions happen.

[...]

PR #5055, opened 2022-03-07
Improved submit notifications

### LINKED ISSUE #4884: Revise submit notifications

grzesiek2010:
Closes #4884 

I've performed manual tests and also added some automated tests for the new code.

This makes handling errors that might occur during uploading forms consistent with handling errors that occur during downloading. Now in both cases we:
- produce a map that contains exceptions
- the map is passed to `NotificationManagerNotifier` (`onUpdatesDownloaded`/`onSubmission`)
- then it is passed to a proper notification builder: `FormUpdatesDownloadedNotificationBuilder`/`FormsSubmissionNotificationBuilder`
- then it is handled using result interpreters: `FormsDownloadResultInterpreter`/`FormsUploadResultInterpreter` 

the updated notification should look like:
|  |  |
| image | image |

what is consistent with the changes we have added for other notifications in https://github.com/getodk/collect/pull/4929 

As I said the new implementation is consistent with what we already have for downloading forms. The only **thing of note** is the way I handle general exceptions. By general exceptions I mean those that happen before trying to upload any form and they stop the whole process. In our code they are represented by `SubmitException`. So the previous implementation was that we displayed one message no matter if we have one of that general exceptions or an exception in one of x forms, everything was displayed in the notification. Now we don't want to display errors in the notification directly so I decided that the easiest way for us is to map those general exceptions into all forms: https://github.com/getodk/collect/pull/5055/files#diff-1c02297e141d84c421610681ff16faed7522fd347a33cb1dd81c6b8fbd6c79bcR60
That means now if one of those exceptions takes place in `ErrorActivity` the list of all forms will be displayed with that exception:
Screenshot_1646662305

Handling such general exception (if you want to do that in other way for example displaying in the notification directly) should be discussed in a separate pull request because it's something not related only to uploading but generally to all errors. For now I think it's not important and I think my solution is ok.

As always it's good to review the commits one by one.

Please test not only the output notification produced by `auto sending` but the whole functionality to make sure there is no regression. Please also notice that the notification might look different on different android versions, for example on some devices the icon in `SHOW DETAIL` will be displayed (make sure it looks good).

No.

No.


collect_app/src/main/java/org/odk/collect/android/upload/UploadException.kt
```diff
@@ -32,3 +33,19 @@ open class UploadException : Exception {
...
 }
+
+/**
+ * Thrown to indicate that the server an upload attempt was made to is requesting authentication.
+ * This may lead to a re-try attempt if the upload was triggered manually by the user (as opposed to
+ * auto-send).
+ */
+class UploadAuthRequestedException(
+    message: String,
+    /**
```
seadowg: This could go in the docs above the class as an `@param`

errors/src/main/java/org/odk/collect/errors/ErrorAdapter.kt
```diff
@@ -1,30 +1,30 @@
...
+import org.odk.collect.errors.databinding.ErrorItemBinding
 
 class ErrorAdapter(private val errors: List<ErrorItem>) : RecyclerView.Adapter<ErrorAdapter.ViewHolder>() {
 
-    class ViewHolder(view: View) : RecyclerView.ViewHolder(view) {
-        val title: TextView = view.findViewById(R.id.title)
-        val secondaryText: TextView = view.findViewById(R.id.secondary_text)
-        val supportingText: TextView = view.findViewById(R.id.supporting_text)
-    }
+    private lateinit var binding: ErrorItemBinding
```
seadowg: I don't think the binding should be shared between items like this. Will this not result in each item sharing the same view object?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/NotificationDrawer.kt
```diff
@@ -48,6 +48,21 @@ class NotificationDrawer {
         return this
     }
 
+    fun <D : Page<D>?> clickAction(
```
seadowg: I think it's probably a mistake that the generic ends up nullable here. Maybe a left over from an earlier conversion?

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.odk.collect.android.utilities.InstancesRepositoryProvider
+import org.odk.collect.forms.FormsRepository
+import org.odk.collect.forms.instances.Instance
+import org.odk.collect.forms.instances.InstancesRepository
+import org.odk.collect.permissions.PermissionsProvider
+import org.odk.collect.settings.SettingsProvider
+import org.odk.collect.settings.keys.ProjectKeys
+import org.odk.collect.shared.settings.Settings
+
+class InstanceAutoSender(
```
seadowg, line 20: A lot of the details here are untested (like filtering out instances not to send for example). Do you think we should pull out a specific issue for adding coverage to this and the `InstanceSubmitter`, `InstanceUploader` objects?

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
```diff
@@ -0,0 +1,97 @@
...
+    private val formsRepositoryProvider: FormsRepositoryProvider,
+    private val instancesRepositoryProvider: InstancesRepositoryProvider,
+    private val googleAccountsManager: GoogleAccountsManager,
+    private val googleApiProvider: GoogleApiProvider,
+    private val permissionsProvider: PermissionsProvider,
+    private val settingsProvider: SettingsProvider,
+    private val instancesAppState: InstancesAppState
+) {
+    fun autoSendInstances(projectId: String): Boolean {
+        val formsRepository = formsRepositoryProvider.get(projectId)
```
seadowg: We could look at pulling out the `ProjectSandbox` from `FormsUpdater` and using that here?

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
```diff
@@ -0,0 +1,97 @@
...
+                val toUpload = getInstancesToAutoSend(
+                    formsRepository,
+                    instancesRepository,
+                    generalSettings
+                )
+                try {
+                    val result: Map<Instance, FormUploadException?> = instanceSubmitter.submitInstances(toUpload)
+                    notifier.onSubmission(result, projectId)
+                } catch (e: SubmitException) {
+                    if (e.type == SubmitException.Type.GOOGLE_ACCOUNT_NOT_SET) {
```
seadowg: It looks like this doesn't handle the `NOTHING_TO_SUBMIT`? I'm guessing we don't do anything in that case (no notification) but either way I think making this a when on `SubmitException#type` and then explicitly "no-oping" in the `NOTHING_TO_SUBMIT` case would be clearer.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.kt
```diff
@@ -0,0 +1,130 @@
...
+    private val analytics: Analytics,
+    private val formsRepository: FormsRepository,
+    private val googleAccountsManager: GoogleAccountsManager,
+    private val googleApiProvider: GoogleApiProvider,
+    private val permissionsProvider: PermissionsProvider,
+    private val generalSettings: Settings
+) {
+
+    @Throws(SubmitException::class)
+    fun submitInstances(toUpload: List<Instance>): HashMap<Instance, FormUploadException?> {
```
seadowg: This can just return `Map` instead of `HashMap`

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.kt
```diff
@@ -0,0 +1,130 @@
...
+        if (toUpload.isEmpty()) {
+            throw SubmitException(SubmitException.Type.NOTHING_TO_SUBMIT)
+        }
+        val result = HashMap<Instance, FormUploadException?>()
```
seadowg: `mutableMapOf`

errors/src/test/java/org/odk/collect/errors/RobolectricApplication.kt
```diff
@@ -0,0 +1,11 @@
+package org.odk.collect.errors
+
+import android.app.Application
+import org.odk.collect.strings.localization.LocalizedApplication
+import java.util.Locale
+
+class RobolectricApplication : Application(), LocalizedApplication {
+
+    override val locale: Locale
+        get() = this.resources.configuration.locales[0]
```
seadowg: I see we did this in `geo` as well. Do you think we should make `LocalizedActivity` fallback to the default `Locale` instead of relying on the underlying application being a `LocalizedApplication`? It feels like that would match what we do in `getLocalizedString` - just fallback to using the default Locale rather than exploding.

errors/build.gradle
```diff
@@ -33,13 +33,32 @@ android {
...
-    implementation project(path: ':strings')
-    implementation project(path: ':androidshared')
+
+    testImplementation project(path: ':androidtest')
+    testImplementation project(path: ':testshared')
+
+    testImplementation Dependencies.androidx_test_ext_junit
+    testImplementation Dependencies.androidx_test_espresso_core
+    testImplementation Dependencies.androidx_test_espresso_contrib
+    testImplementation Dependencies.robolectric
```
seadowg: I don't think you need the `contrib` or the `robolectric` dependency here.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/upload/UploadException.kt
grzesiek2010: Fixed.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/NotificationDrawer.kt
grzesiek2010: Fixed.

errors/build.gradle
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.kt
grzesiek2010: Fixed.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceSubmitter.kt
grzesiek2010: Fixed.

errors/src/main/java/org/odk/collect/errors/ErrorAdapter.kt
grzesiek2010: Right, fixed.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
grzesiek2010: Fixed.

errors/src/test/java/org/odk/collect/errors/RobolectricApplication.kt
grzesiek2010: Yeah that makes sense, fixed.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
grzesiek2010: Yeah we can file a separate issue to investigate it.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
grzesiek2010: I was thinking about it but it would basically moving dependencies from one class to another. I can't construct the whole `ProjectSandbox` using dagger and inject to `AutoSendTaskSpec`  because I don't know what project (`projectId`) it should represent. That means I would need to construct it in `AutoSendTaskSpec` and to do that I would need to inject all the dependencies it needs there instead of to `InstanceAutoSender`. I don't think it makes a lot of sense.... maybe if it would make testing `InstanceAutoSender` easier but it's related to the question above.

seadowg: Oh I still think you'd want to use it in the same way as `FormsUpdater` rather than injecting it with Dagger - constructing it based on the `projectId`. I was just thinking of it as a way to clean up the noise from grabbing all the repositories etc for the Project. It's just an aesthetic change really so you don't have a big preamble before the logic.

collect_app/src/main/java/org/odk/collect/android/instancemanagement/InstanceAutoSender.kt
seadowg: I actually remembered that we already have this in a fashion - we'll want to add tests and think about structure again for all this for https://github.com/getodk/collect/issues/4540.

seadowg (changes requested):
[As in last comment](https://github.com/getodk/collect/pull/5055#discussion_r839556576)

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
```diff
@@ -44,6 +45,9 @@ class AutoSendTaskSpec : TaskSpec {
     @Inject
     lateinit var instanceAutoSender: InstanceAutoSender
 
+    @Inject
```
seadowg, line 48: I don't think there's any reason to inject a creator here instead of just creating the `ProjectSandbox` in `autoSendInstances` like in `FormsUpdater`. Right now, I'd still prefer the `projectId` being the argument to `autoSendInstances`. Maybe I'm missing something though. To be explicit I was thinking:

```kotlin
fun autoSendInstances(projectId: String): Boolean {
   val sandbox = getProjectSandbox(projectId)
   ...
```

seadowg (changes requested):
I think we got wires crossed around what I meant by using the `ProjectSandbox`. Comment inline.

collect_app/src/main/java/org/odk/collect/android/backgroundwork/AutoSendTaskSpec.kt
grzesiek2010: I know what you meant. It was just playing with other ideas and it wasn't even finished but to be honest I kind of liked this direction.
Without such a creator I would need to inject like 4 new objects into `AutoSendTaskSpec` to construct `ProjectSandbox` there so ok I would make `InstanceAutoSender` a bit smaller but I would clutter `AutoSendTaskSpec` instead.
Maybe having such a creator for `ProjectSandbox` would make sense?

seadowg: > It was just playing with other ideas and it wasn't even finished

Ah sorry I forget that notifications can just be commits and are not always review requests. My bad.

> Without such a creator I would need to inject like 4 new objects into AutoSendTaskSpec to construct 

Is that because `ProjectSandbox` requires dependencies that `AutoSendTaskSpec` wasn't already injecting? I guess you'd have to inject the `FormSourceProvider` and the `StoragePathProvider` so you could pass that through to the `InstanceAutoSender` and allow it to construct the `ProjectSandbox`. 

> Maybe having such a creator for ProjectSandbox would make sense?

You know the more I think about it the more I like it. It seems a little weird as it moves a bunch of dependency management away from Dagger, but I'm usually in favour of that. There are a lot of places now where we need to grab "project level" dependencies, and it is annoying to have to inject them all. The fact that `ProjectSandbox` does `lazy` construction as well means we're not constructing objects we don't need too...

How about we go with the pattern you're using here but renaming `ProjectSanbox` to `ProjectDependencyProvider` and `ProjectSanboxCreator` to `ProjectDependencyProviderFactory`. It's "hyper Java"™️ naming, but it seems clearer  to me that we're passing in a provider and that Dagger is responsible for giving us the factory for that provider. Let's also switch `FormsUpdater` to taking a `ProjectDependencyProvider` in its arguments rather than a `projectId` so we're consistent. I think we'll need interfaces for this eventually so that we can fake them easily, but we can skip that for the moment.

grzesiek2010: Ok I renamed the classes according to your advice and also used in `FormsUpdater`. The only difference is that I passed `ProjectDependencyProviderFactory` via constructor to `FormsUpdater` since `FormsUpdater` is created using dagger it's easier I think. Then in methods were we already passed projectId I just create `ProjectDependencyProvider` using that id.

seadowg (approved):

lognaturel:
@seadowg @grzesiek2010 Could this be QAed on master?

grzesiek2010:
Yes I think it would be ok.

kkrawczyk123:
Verified on Androids 5.1 and 8.1 :
- autosend defined in form
- autosend defined in settings
- failed autosend: show details visible
- able to open show details
- exit from error list
- able to scroll down error list ( As I don't like the way that errors are displayed in landscape opened new issue in: #5082) 
- tap on notification: send finalized form view displayed
- match exactly 
- failed match exactly
- form version update available 
- automatic download
- failed automatic download 

For form version update available and failed automatic download, I had some problems in obtaining them after 15 minutes. I am not sure if it was my connection issues or anything else. @grzesiek2010 Could you please take a look at that in case it's something more serious?

grzesiek2010:
> Could you please take a look at that in case it's something more serious?

sure

grzesiek2010:
> For form version update available and failed automatic download, I had some problems in obtaining them after 15 minutes. I am not sure if it was my connection issues or anything else.

I've tested those notifications and to me everything seems fine. Please keep in mind that the `forms update available` notification is displayed only once for a particular update. I mean if you update your form once and the notification for that change is displayed you won't see it again after 15minutes/1h/6h/24h (depending on your settings) unless you add another change to that form. It is because we do not want to keep displaying that notification for exactly the same change.

When it comes to the `forms updated failed notification` I'm curious how you try to reproduce that case. Simple messing with the url won't work because such exception at an early stage of the process are ignored. Something bad would need to happen during downloading a particular form what for you might be difficult to reproduce... maybe you have a broken form accepted by central that could cause exceptions like that? Maybe @lognaturel has something like that? If not we have an automated test for that in `PreviouslyDownloadedOnlyTest` and we should probably relay on it.

srujner:
**Tested with success** except for the things mentioned above by @kkrawczyk123 

lognaturel:
> Maybe @lognaturel has something like that?

I can't think of a straightforward way to reproduce. @kkrawczyk123 maybe something you can start doing a bit more is reading through some of the automated tests and seeing whether you can think of cases they don't cover. In this case you could take a quick look at `PreviouslyDownloadedOnlyTest` and confirm that it tests the case you had in mind.

PR #5057, opened 2022-03-08
Make sure database files always exist when used

### LINKED ISSUE #5042: The app crashes after deleting and adding a project again

grzesiek2010:
Reverts https://github.com/getodk/collect/pull/5045
Closes #5042 

This pr adds changes implemented (and then reverted) in #5043 back.
I just got rid of one commit: https://github.com/getodk/collect/pull/5043/commits/8402b65b831312c15a55191d1e0c54a05405c594 which included closing databases and was a cause of failing tests.

I don't think the solution without closing databases is very bad because it's basically what we have on master and it doesn't seem to cause any big problems (on master we have a cache where we store database helpers and we never remove anything from that cache so if we create 100 projects and then remove 99 of them we will still keep all database helper object with open databases).
Of course it would be good to properly handle closing databases but it looks like it would require a lot of investigation. I'm also not a fan of removing caching at all. I think creating new objects for every operation might be something that will only bring other bugs.


lognaturel:
I think this history is going to be tricky to follow. How about actually bringing back the individual commit(s) from the original PR?

grzesiek2010:
> I think this history is going to be tricky to follow. How about actually bringing back the individual commit(s) from the original PR?

right, done.

grzesiek2010:
There is something wrong with new tests. I'm not going to investigate it right away because if @seadowg is strongly against adding this back it wouldn't make sense.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
```diff
@@ -26,29 +27,28 @@ open class DatabaseConnection(
...
-                name,
-                null,
-                databaseVersion,
-                migrator
-            )
+    private val dbHelper: SQLiteOpenHelper
+        get() {
+            val databasePath = path + File.separator + name
+            if (openHelpers.containsKey(databasePath) && !File(databasePath).exists()) {
+                openHelpers.remove(databasePath)
```
seadowg: Let's add a comment here (probably referencing the original PR) around why we're not including a call to `close` here.

seadowg (dismissed):
I think we should go ahead with this. We will end up leaking `SQLiteOpenHelper` instances for projects that get deleted, but as we've all talked about, that's quite a rare occurrence, and we don't expect it to have any large consequences. @grzesiek2010 could you please file an issue around this so that we can have a think about how we might improve our connection pooling for later release.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
grzesiek2010: Ok I added a short comment and the issue https://github.com/getodk/collect/issues/5078

seadowg (approved):

grzesiek2010:
It was tested before so probably we can merge without testing right?

PR #5058, opened 2022-03-09
Improved handling errors during importing new projects

grzesiek2010:
I tested the fix manually.

I think the most common issues that might happen in real life could be caused by `JSONException` which we used to catch in `SettingsImporter#fromJSON` but ignore. I decided to remove that try/catch so that the exception will be propagated and caught in `ODKAppSettingsImporter` where we catch not only `JSONException` but all possible exceptions and errors because no matter what goes wrong a new project should not be added.

I think testing scaaning qr codes to make sure there is no regression would be enough.

No.

No.

### DIFF
--- a/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
+++ b/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
@@ -1,6 +1,5 @@
 package org.odk.collect.settings
 
-import org.json.JSONException
 import org.odk.collect.projects.Project
 import org.odk.collect.projects.ProjectsRepository
 import org.odk.collect.settings.importing.ProjectDetailsCreatorImpl
@@ -29,10 +28,6 @@ class ODKAppSettingsImporter(
     )
 
     fun fromJSON(json: String, project: Project.Saved): Boolean {
-        return try {
-            settingsImporter.fromJSON(json, project)
-        } catch (e: JSONException) {
-            false
-        }
+        return settingsImporter.fromJSON(json, project)
     }
 }
--- a/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
+++ b/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
@@ -1,5 +1,6 @@
 package org.odk.collect.settings
 
+import org.json.JSONException
 import org.odk.collect.projects.Project
 import org.odk.collect.projects.ProjectsRepository
 import org.odk.collect.settings.importing.ProjectDetailsCreatorImpl
@@ -30,7 +31,7 @@ class ODKAppSettingsImporter(
     fun fromJSON(json: String, project: Project.Saved): Boolean {
         return try {
             settingsImporter.fromJSON(json, project)
-        } catch (e: Throwable) {
+        } catch (e: JSONException) {
             false
         }
     }
--- a/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
+++ b/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
@@ -28,6 +28,10 @@ class ODKAppSettingsImporter(
     )
 
     fun fromJSON(json: String, project: Project.Saved): Boolean {
-        return settingsImporter.fromJSON(json, project)
+        return try {
+            settingsImporter.fromJSON(json, project)
+        } catch (e: Throwable) {
+            false
+        }
     }
 }
--- a/settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
+++ b/settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
@@ -4,27 +4,13 @@ import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers.equalTo
 import org.junit.Test
 
-private const val SCHEMA = """
-            {
-                "${'$'}schema": "https://json-schema.org/draft/2019-09/schema",
-                "${'$'}id": "https://example.com/example.schema.json",
-                "title": "Schema",
-                "type": "object",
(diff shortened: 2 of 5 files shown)


grzesiek2010:
I wanted to add automated tests of course to verify that if `SettingsImporter#fromJSON` throws any error then `ODKAppSettingsImporter#fromJSON` returns false. Unfortunately it's not easy because of the structure in `ODKAppSettingsImporter` where `SettingsImporter` is not passed via constructor but created. This was implemented like that to make some classes that belong to `settings` module internal and not available outside. However this make testing more complex so I would rather change it. Since it's something @seadowg implemented let's wait for his opinion.

settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
```diff
@@ -28,6 +28,10 @@ class ODKAppSettingsImporter(
     )
 
     fun fromJSON(json: String, project: Project.Saved): Boolean {
-        return settingsImporter.fromJSON(json, project)
+        return try {
+            settingsImporter.fromJSON(json, project)
+        } catch (e: Throwable) {
```
seadowg: I'm not sure if I agree with the issue (and it also seems more worried about camera code), but I think changing what we do with JSON errors makes sense, so I'll post in the issue and here. 

I think catching `JSONException` would be enough here as it's currently the only thing we'd expect that wouldn't be a code/build problem (like ProGuard problems) that would usually crash every time. And for testing that, I'd just add a test case with invalid JSON to `ODKAppSettingsImporterTest`.  We could alternatively keep the `catch` in `SettingsImporter` and just return `false` in that case. I'd be tempted to go with the former though as long term I suspect we'll  end up pulling `SettingsImporter` apart (discussed in https://github.com/getodk/collect/pull/4924).

seadowg (changes requested):

settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
grzesiek2010: > I think catching JSONException would be enough here as it's currently the only thing we'd expect that wouldn't be a code/build problem (like ProGuard problems) that would usually crash every time.

Done.

> And for testing that, I'd just add a test case with invalid JSON to ODKAppSettingsImporterTest

I've extended the fix a bit and added catching `JsonParseException` see: https://github.com/getodk/collect/pull/5058/commits/70c84a67e8b90ddd7ef70977ece97ba930b48fbb

I've added a test in `ODKAppSettingsImporterTest` too but actually it doesn't test [JSONException](https://github.com/getodk/collect/pull/5058/commits/ffbaddb9f5b0caac0899f0ada4fa15c73ee6d7b1)
but that `JsonParseException`. I can't come up with any json that would not be caught in JsonSchemaSettingsValidator but cause  `JSONException` later.

settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
```diff
@@ -28,6 +29,10 @@ class ODKAppSettingsImporter(
     )
 
     fun fromJSON(json: String, project: Project.Saved): Boolean {
-        return settingsImporter.fromJSON(json, project)
+        return try {
+            settingsImporter.fromJSON(json, project)
+        } catch (e: JSONException) {
+            false
```
seadowg: I'm not sure how this could actually happen - currently it's not covered by a test. I could definitely be missing a case, but as far as I can tell, the changes you've made to `JsonSchemaSettingsValidator` (rejecting invalid JSON) should prevent a `JSONException` from occurring later?

seadowg (changes requested):

settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt
grzesiek2010: I'm not able to fully remember my thinking since it's been a while but in my comment above I also said that:
> I can't come up with any json that would not be caught in JsonSchemaSettingsValidator but cause JSONException

so probably you are right and we can get rid of it at least for now.

settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
```diff
@@ -4,12 +4,7 @@ import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers.equalTo
 import org.junit.Test
 
-class JsonSchemaSettingsValidatorTest {
-
-    @Test
-    fun `returns true when json is valid based on schema`() {
-        val validator = JsonSchemaSettingsValidator {
-            """
+private const val SCHEMA = """
```
seadowg: Can we move this below the test class, so you don't have to scroll down to get to the tests themselves here?

settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
```diff
@@ -21,9 +16,15 @@ class JsonSchemaSettingsValidatorTest {
                     }
                 }
             }
-            """.byteInputStream()
-        }
+            """
 
+class JsonSchemaSettingsValidatorTest {
+    private val validator = JsonSchemaSettingsValidator {
```
seadowg: I think it would be easier to read if the validator was constructed in the test. Then you can click-navigate to the schema definition directly.

seadowg (approved):
Couple of minor tweaks to a test but looks good!

srujner:
**Tested with Success!**
Verified on Android 10 and 11

Verified Cases:
- Adding project by scanning QR code
- Adding duplicated projects
- Inverted QR code
- Invalid QR code
- Reconfiguring via QR code
- Exporting and importing settings
- Light and Dark mode
- Screen rotation

kkrawczyk123:
Verified also on Androids 5.1 and 8.1.

Issue #2574, opened 2018-09-25
Buttons in android permission dialog don't cause any action if an activity has been recreated after showing the dialog.


grzesiek2010:
#### Software and hardware versions 
Collect v1.17.0

#### Problem description
Buttons in android permission dialog don't cause any action if an activity has been recreated after showing the dialog.

#### Steps to reproduce the problem
It's one of many possible ways to reproduce the issue:
1. Go to device settings -> Apps -> ODK Collect -> Permission and make sure that Microphone is turned off
2. Open any form which contains the AudioWidget (eg AllWidgets form)
3. Click on the Record Sound button
4. Rotate your device
5. Click any button (Deny or Allow)

#### Expected behavior
If you click Deny another dialog with an explanation should be displayed otherwise (allow) an external app (audio recorder) should be started.

#### Other information 
It's caused because we pass activity https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/PermissionUtils.java#L160 and after activity recreation, there is no reference to a current activity . We could fix this issue using for example rxEvents but it would make the code complex and then I don know if using the library we use makes sense.

This problem doesn't exist using the approach from the doc https://developer.android.com/training/permissions/requesting

grzesiek2010:
any other approaches @jd-alexander ?

lognaturel:
Interesting. I think we’re probably not using the lib quite as it is intended. Have you maybe looked at some of the samples linked from https://github.com/Karumi/Dexter/issues/15?

jd-alexander:
> Interesting. I think we’re probably not using the lib quite as it is intended. Have you maybe looked at some of the samples linked from Karumi/Dexter#15?

The methods that were introduced there were depreciated and then removed. Will be investigating what was implemented in favour of them.

grzesiek2010:
> I think we’re probably not using the lib quite as it is intended

It was also my first idea but then I downloaded the sample app https://github.com/Karumi/Dexter/tree/master/sample and the behavior there is the same.

lognaturel:
@grzesiek2010 how about filing a new issue on the Dexter repo asking what the latest strategy may be and referring to issues at https://github.com/Karumi/Dexter/search?q=rotation&type=Issues

I’m not optimistic we’ll get a response since there hasn’t been activity in a while but good to at least have it out there. 

I’d say this is pretty low priority because it is recoverable and I doubt screen rotation is very common (maybe we should get analytics).

[...]

Issue #2571, opened 2018-09-24
Runtime permission dialogs behave badly on orientation changes in Android 6, 7


lognaturel:
#### Software and hardware versions 
Collect v1.17.0-beta.0, Android 6.0 and 7.0

#### Problem description
From @mmarciniak90 at https://github.com/opendatakit/collect/pull/2506#issuecomment-420361713 verifying the phone data permission check:

```
Case 1:

The user sees information that 'ODK Collect does not make calls...'
The user changes device orientation
The user sees the dialog with Deny / Allow buttons
Case 2:

The user sees the dialog with Deny / Allow buttons
The user changes device orientation eg. 4 times
The user clicks on Deny
The user sees an empty screen and needs to click 4 times to see the dialog with information that 'ODK Collect does not make calls...'
A similar situation is when the user clicks on Allow in point 3
Problems are not visible on Android 8.1
```

@grzesiek2010 suggests in https://github.com/opendatakit/collect/pull/2506#issuecomment-420290579 that this is a problem for all of the permission checks.

The best approach to take here needs discussion. Since this is not hugely disruptive, it's not very high priority.

#### Steps to reproduce the problem
1. Use an Android 6.0+ device
1. Make sure Collect does not have permissions to read phone state
1. Load a form like [preloads.xml.txt](https://github.com/opendatakit/collect/files/2318208/preloads.xml.txt) which reads phone data
1. See the permissions dialog, rotate the screen

#### Expected behavior
A single dialog is shown and does the right thing on orientation changes

grzesiek2010:
> @grzesiek2010 suggests in #2506 (comment) that this is a problem for all of the permission checks.

In fact, the problem I pointed out is different (but also related to changing device orientation). I created a separate issue which I think we should fix firs.

https://github.com/opendatakit/collect/issues/2574

seadowg:
@getodk/testers are you able to double-check if this is still a problem or not?

dbemke:
I'm able to reproduce those steps on **Android 8.1, 10, 11**
#### Steps to reproduce the problem
1. Go to the metadata form.
[metadata.xml.txt](https://github.com/getodk/collect/files/10765633/metadata.xml.txt)
2. When you see a permission asking to allow to make phone call, rotate the device.
3. Tap "deny" ( the number how many times you need to tap deny depends on the number of rotation done before).
4. Then there is a "Loading Form" state, click the device back button.
5. "Read phone state permission" appears, tap "ok".

[...]

PR #5059, opened 2022-03-10
Always handle disabled Location when checking location permissions

### LINKED ISSUE #4996: Always handle disabled Location when checking location permissions
The handles requesting/checking location permissions separately from disabled Location (at a system level). We've introduced a new helper that covers both (`PermissionsProvider#requestEnabledLocationPermissions`) and could now use that instead of `requestLocationPermissions` to cover both cases everywhere and remove duplicated code.

### Notes

`PermissionsProvider`'s `requestX` methods are mostly untested at the moment as we don't have a way to fake out the Dexter framework code. We should try and rework the tests as part of this so we don't need to use a `spy` anywhere and have coverage of the permission request code.

grzesiek2010:
Closes #4996 

There is not much to discuss here we just should always check if location is enabled as a part of location requests.

Recently we have added checking if location is enabled for location permission requests (see: https://github.com/getodk/collect/pull/4990) it was used only before starting a geopoint widget. However there are other places when we request location permissions:
- when a user tries to open the map of forms from the list of forms
- when a user uses geo widgets (geotrace and geoshape)
- when a user uses a form with background location

so we need to verify that always when we check location permissions we also check if location is enabled at all.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FillBlankFormActivity.kt
@@ -195,7 +195,7 @@ class FillBlankFormActivity :
     }
 
     private fun onMapButtonClick(id: Long) {
-        permissionsProvider.requestLocationPermissions(
+        permissionsProvider.requestEnabledLocationPermissions(
             this,
             object : PermissionListener {
                 override fun granted() {
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -2475,7 +2475,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
         displayUIFor(backgroundLocationViewModel.activityDisplayed());
 
         if (backgroundLocationViewModel.isBackgroundLocationPermissionsCheckNeeded()) {
-            permissionsProvider.requestLocationPermissions(this, new PermissionListener() {
+            permissionsProvider.requestEnabledLocationPermissions(this, new PermissionListener() {
                 @Override
                 public void granted() {
                     displayUIFor(backgroundLocationViewModel.locationPermissionsGranted());
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/utilities/ActivityGeoDataRequester.kt
@@ -89,7 +89,7 @@ class ActivityGeoDataRequester(
         answerText: String?,
         waitingForDataRegistry: WaitingForDataRegistry
     ) {
-        permissionsProvider.requestLocationPermissions(
+        permissionsProvider.requestEnabledLocationPermissions(
             activity,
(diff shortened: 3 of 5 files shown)


grzesiek2010:
In the description of the issue there was mentioned that we should add some refactoring as a part of the fix to make testing easier. I skipped this part because we should discuss the way we handle permissions (mostly requests) first. We have been using [Dexter](https://github.com/Karumi/Dexter) library to make the whole process easier however the library has been archived so I don't think it's a good idea to build new abstractions around that library. It had its issues like: 
- https://github.com/getodk/collect/issues/2571
- https://github.com/getodk/collect/issues/2574

and since it's not maintained anymore we might only expect other new bugs.

I like how we can handle requests using the new [Activity Result API](https://developer.android.com/training/basics/intents/result) here is a sample how it can be used https://gist.github.com/antoniolg/edac2587da74668776c4838666102079

The advantage of this solution is that the current activity is always used to handle permissions so we don't have that bug like in case of Dexter that after rotating a device with displayed permission dialog there was no action after accepting/denying.

The problem is that now it won't be possible to request permissions that easily in widgets for example. If they were fragments we could call `registerForActivityResult` there but they are not so this is something we would need to keep in `FormEntryActivity` and pass to widgets to let them launch the `ActivityResultLauncher`.
 
so yeah definitely this is something we should discuss first.

lognaturel (approved):

lognaturel:
Sounds good to have a separate big picture conversation about permissions. Put on the Tues agenda?

kkrawczyk123:
@grzesiek2010 I have noticed an issue when testing form with background location and disabled GPS in the device.
reproduction steps:
1. Start form
2. Tap cancel on displayed GPS dialog
3. Jump to hierarchy view 
4. Tap go to end button - see that duplicate of dialog displays on save form and exit view 
see video:
ezgif com-gif-maker (22)

I was able to reproduce the issue with a few different forms with background location on Androids 5.1 and 8.1 and also tested previous behavior on master where the issue is not visible.

grzesiek2010:
> I have noticed an issue when testing form with background location and disabled GPS in the device.

It should work well now.

kkrawczyk123:
**Testes with success!**

Verified cases:
- light/dark mode
- disable GPS in device
- open map from form list
- geopoint/geotrace/geoshape
- form with background location
- enable GPS using button on dialog 
- cancel button
- able to save points and use widgets after enabling GPS
- location permission turned on and off
- minimize app

srujner:
**Tested with success!**
Verified on Androids: 10 and 11. on the debug mode.

@grzesiek2010 @lognaturel @seadowg 
I noticed a strange behavior on this PR regarding SelfSignedRelease mode.
When SelfSignedRelease mode is chosen, the app is crashing after adding project manually or by QR code both on Android 10 and 11. There is no issue in debug mode. 
I also checked on the master branch for SelfSignedRelease and Debug mode and the issue did not occur so it may be related only to this PR.

srujner:
I was able to reproduce it on #5062 too

grzesiek2010:
Yeah probably it's not directly related to those prs. I will look into it. Thanks!

grzesiek2010:
@srujner I can confirm this pr was outdated and didn't contain one of the fixes we have merged recently. On master everything should be fine even with this pr.

PR #5061, opened 2022-03-14
Replace existing form with matching formid and version

### LINKED ISSUE #1693: Problem with two forms with the same id and form version
Collect v1.11.1

1. Download any form from Aggregate
2. Delete that form from Aggregate
3. Edit the form (without changing form id or form version) and upload again
4. Download the form again

#### Problem description
I can see two forms with the same id and version number on your device. We can't use both of in that case:
screenshot_2017-12-08-17-44-26

#### Expected behavior
An older form should be removed automatically.

any thoughts?

grzesiek2010:
Closes #1693 

I tested the app manually and added/update automated tests.

Another approach would be to delete the existing form first if we try to add a new one with the same id and version so that then we wouldn't need to update a form but just add it... but that would also remove all related files. I think the case with uploading new drafts to Central is that users implement small changes so redownloading all files might not be a good solution.

This should not allow users to download two forms with the same id and version. With Central it is possible to achieve using drafts and uploading a new form version.

No.

No.

- [ ] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#ui-components-style-guidelines)


grzesiek2010:
@lognaturel please take a look and let me know what you think about this approach.
Now after playing with this a bit more I see one problem with updating which is that after doing that we end up with a new form file and a new media dir and files that belonged to the old form are not removed. If you think it's a problem maybe we should remove the old form (along with its files) first and then insert the new one instead of updating?

lognaturel:
> we end up with a new form file and a new media dir

This is surprising to me. I would expect an in-place update leading to the form def + any existing, changed media to be replaced. There might be extra media that's still there and that's fine. Can you share the actual directory structure that's being produced?

grzesiek2010:
Let's say you have a draft form on Central named `TestForm` along with some media files. You download it and `TestForm.xml` is created and `TestForm-media` directory where all media files are downloaded. Then you update one of the questions in the form and/or some media files. After redownloading `TestForm_2.xml` and its `TestForm_2-media` will be created. If we just update the existing row in database the old `TestForm.xml` and `TestForm-media` will not be removed.

So now I think that removing the old forms first if they have the same id and version would be better. Do you see any danger in this approach?

grzesiek2010:
and maybe @seadowg has something to add here.

lognaturel:
Discussed briefly with @grzesiek2010. He's going to explore whether it's possible to add a condition when the numeric suffix is added: in the case of matching formid/version, replace in place.

grzesiek2010:
Ok I've played with it a bit more and I think the best option (short and safe) is to insert new forms like now even if they are duplicate but collect and then remove pre-existing forms with the same id and version at the end of the process of downloading a form: https://github.com/getodk/collect/pull/5063/commits/7ec62025d11d977916fc01779dc8bbbb1a6a861c
Thanks to the functionality that copies media files from other forms with the same id we won't need to download them again and at the same time we won't end up with abandoned media files. So it seems like a win-win solution

I was also able to find one possible scenario when users might end up with duplicate forms.... if forms are being synchronized with the server (match exactly) we first download an xml file and then save a new row in database. If between those two steps we trigger our `FormSyncTask` (by opening the list of forms), then it might detect that xml file and try to add it to database too. I've fixed the issue in my first commit: https://github.com/getodk/collect/pull/5063/commits/0db7931b9bec98fd4c4d2854e2c776257cdfb819

@lognaturel what do you think?

lognaturel:
> So it seems like a win-win solution 

Agreed! That does mean that on disk the newly-downloaded forms will have a numeric suffix (e.g. `_2`), right? I don't think that matters but wanted to point it out and think about it a bit more.

> then it might detect that xml file and try to add it to database too

Ah! That makes a lot of sense. 🙈 Concurrency is hard. It definitely feels to me like the issue comes from Collect so this is consistent with that.

grzesiek2010:
> forms will have a numeric suffix (e.g. _2), right?

correct but to me it doesn't seem like a problem.

lognaturel:
> doesn't seem like a problem

Yeah, I guess it really is an implementation detail. We always get the path from the DB anyway. Ok, seems good to proceed. 👍  Great to have an explanation for the weird behavior!

grzesiek2010:
Cool so I'm closing in favor of https://github.com/getodk/collect/pull/5063

Issue #4949, opened 2021-12-30
Add support for `map` and `quick` appearance in `select_one`/`select_one_from_file`


seadowg:
## Acceptance

- [x] Given I've loaded a form into Collect with a `select_one`/`select_one_from_file` with multiple choices with a point in their `geometry` column and `map`/`quick` appearances
When I open the form
And I navigate to the question
And I click "Select place"
Then I see all the points on a map at the correct coordinates
And the question label is shown at the top of the screen

- [x] Given I'm on the map after clicking "Select place"
When I click on a point
Then I return to the question with the selected choice as the answer

## Notes

An example form with a `geometry` column for choices is [here](https://docs.google.com/spreadsheets/d/191Taxl2BU1UmBPpU4SLXEiP2cnhVx7mSe8TJeKnEDAI/edit#gid=1988906120).

The [forum post](https://forum.getodk.org/t/ability-to-load-locations-into-the-geowidget-to-guide-data-collection/27406/12) mentions matching the form map colors for the map markers here, but I think it would be a good idea to experiment with using theme colors. For instance, we could use `?colorSurface` for unselected markers and `?colorSecondary` for selected ones (to match radio buttons etc).

lognaturel:
The necessary JavaRosa snapshot should now be available: https://oss.sonatype.org/content/repositories/snapshots/org/getodk/javarosa/3.4.0-SNAPSHOT/

PR #5062, opened 2022-03-15
Improve `SelectionMapFragment` integration

### LINKED ISSUE #4947: Extract generic "select item from map" implementation from `FormMapActivity`

seadowg:
Closes #4947 

There's still some clean up I'd like to do here but I've been in here long enough and think we can move these tasks until after doing GeoJSON:

1. Investigate moving shared map behaviour from `SelectionMapFragment` into `MapFragment` implementations (maintaining map zoom on recreate for instance)
1. Investigate extracting `maps` module to house `MapFragment`, `MapPoint` and `MapFragmentFactory` and creating modules for each implementation (classic **D**epenednecy **I**nversion **P**rinciple setup)
1. Organize `geo` package structure
1. Add loading state/progress dialog to form map - currently DB operations are all happening on the UI thread (existing problem)

I can create an issue for these if the reviewer agrees we can move on (and start on #4949).

I've moved a bunch of testing around and tried to keep coverage intact, however Android Studio's coverage tool is currently broken which has meant I've been flying manually on that so good to keep an eye out for any omissions. There are a few things I almost got bogged down in that weren't covered and that I'll ignore for the moment (bottom sheet behaviour for example) - I'd like to make this part of the follow up issue as well.

New tests and played around manually.

Comments inline.

Just effects form the form map feature and I think now would be a good time to do some QA on that before moving on.

Any forms that include geo questions (so they work with the form map).


geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -321,41 +340,23 @@ class SelectionMapFragment : Fragment() {
...
-    fun setMapTitle(title: String) {
-        this.mapTitle.value = title
-    }
+abstract class SelectionMapViewModel : ViewModel() {
+    abstract fun getMapTitle(): LiveData<String>
+    abstract fun getItemCount(): LiveData<Int>
+    abstract fun getMappableItems(): NonNullLiveData<List<MappableSelectItem>>
 }
 
 data class MappableSelectItem(
```
seadowg: We'll probably need to make the `status` field here a list of `IconifiedText` so we can display arbitrary columns from GeoJSON, but it feels like we should wait to get there before doing that.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+                    return FormMapViewModel(
+                        resources,
+                        formId,
+                        formsRepositoryProvider.get(),
+                        instancesRepositoryProvider.get(),
+                        settingsProvider
+                    ) as T
+                }
+            }
+        )[ARG_VIEW_MODEL_KEY, FormMapViewModel::class.java]
```
seadowg, line 69: There were a few different options that I went through on how best to communicate between the `SelectionMapFragment` and the host activity (the "client" of the framework component). In the end this feels like the best way - allow the client to create its own implementation of `SelectionMapViewModel` and have it pass a key to `SelectionMapFragment` which allows it to grab the ViewModel in a "sevice locator" inspired way.

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+        longitude: Double,
+        resources: Resources
+    ): MappableSelectItem {
+        val instanceLastStatusChangeDate = InstanceProvider.getDisplaySubtext(
+            resources,
+            instance.status,
+            Date(instance.lastStatusChangeDate)
+        )
+
+        val info = when {
```
seadowg: This code could definitely be improved - the relationship between `info` and `action` is pretty dicey right now. I think the energy for doing that can be reserved for another time however as I've mainly been trying to focus on the structure here and giving as a reusable pattern for implementing GeoJSON.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -57,6 +65,11 @@ class SelectionMapFragment : Fragment() {
 
     private var previousState: Bundle? = null
 
+    @VisibleForTesting
+    internal constructor(selectionMapViewModel: SelectionMapViewModel) : this() {
```
seadowg, line 77: Fragments shouldn't have constructors - they are created by the framework. The new `FragmentManager` system actually allows us to add constructors which could possibly give us an alternative to using Dagger. However, the most obvious use at the moment is to create simpler tests that allow us to override dependencies that aren't appropriate for Dagger (like here where we grab the ViewModel from the ViewModelProvider).

geo/src/test/java/org/odk/collect/geo/support/Fixtures.kt
```diff
@@ -0,0 +1,20 @@
+package org.odk.collect.geo.support
+
+import android.R
+import org.odk.collect.geo.MappableSelectItem
+
+object Fixtures {
+    fun mappableSelectItem(): MappableSelectItem {
```
seadowg: We use builders a lot in Java, but I like this pattern a lot more in Kotlin - provide a basic "fixture" object and allow tests to customize through `copy()`.

lognaturel:
I've taken a quick look and looks good to me. Seems fine to keep some cleanup for later. @grzesiek2010 you're more in tune with architecture these days so I requested your review. Would be amazing if this can start QA on Friday.

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+    override fun getMappableItems(): NonNullLiveData<List<MappableSelectItem>> {
+        return mappableItems
+    }
+
+    fun load() {
+        val instances = instancesRepository.getAllByFormId(form.formId)
+        val items: MutableList<MappableSelectItem> = ArrayList()
+
+        for (instance in instances) {
+            if (instance.geometry != null && instance.geometryType == "Point") {
```
grzesiek2010: We should have a cost for `Point`

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+                    Timber.w("Invalid JSON in instances table: %s", instance.geometry)
+                }
+            }
+        }
+
+        mappableItems.value = items
+        itemCount.value = instances.size
+    }
+
+    private fun convertItem(
```
grzesiek2010: `createItem` would be a better name there is no conversion here.

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+                getSubmissionSummaryStatusIcon(instance.status),
+                instanceLastStatusChangeDate
+            ),
+            info,
+            action
+        )
+    }
+
+    private fun getDrawableIdForStatus(status: String, enlarged: Boolean): Int {
+        when (status) {
```
grzesiek2010: You can simplify this:
```
        return when (status) {
            Instance.STATUS_INCOMPLETE -> return if (enlarged) R.drawable.ic_room_form_state_incomplete_48dp else R.drawable.ic_room_form_state_incomplete_24dp
            Instance.STATUS_COMPLETE -> return if (enlarged) R.drawable.ic_room_form_state_complete_48dp else R.drawable.ic_room_form_state_complete_24dp
            Instance.STATUS_SUBMITTED -> return if (enlarged) R.drawable.ic_room_form_state_submitted_48dp else R.drawable.ic_room_form_state_submitted_24dp
            Instance.STATUS_SUBMISSION_FAILED -> return if (enlarged) R.drawable.ic_room_form_state_submission_failed_48dp else R.drawable.ic_room_form_state_submission_failed_24dp
            else -> R.drawable.ic_map_point
        }
```

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    private fun getSubmissionSummaryStatusIcon(instanceStatus: String?): Int {
+        when (instanceStatus) {
+            Instance.STATUS_INCOMPLETE -> return R.drawable.form_state_saved
+            Instance.STATUS_COMPLETE -> return R.drawable.form_state_finalized
+            Instance.STATUS_SUBMITTED -> return R.drawable.form_state_submited
+            Instance.STATUS_SUBMISSION_FAILED -> return R.drawable.form_state_submission_failed
+        }
+
+        throw IllegalArgumentException()
```
grzesiek2010: Maybe it would be better not to throw an exception but return anything for example: `form_state_saved`?

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+
+        mappableItems.value = items
+        itemCount.value = instances.size
+    }
+
+    private fun convertItem(
+        instance: Instance,
+        latitude: Double,
+        longitude: Double,
+        resources: Resources
```
grzesiek2010: You don't have to pass this parameter here it's available because you pass it via constructor.

geo/src/test/java/org/odk/collect/geo/support/FakeMapFragment.java
```diff
@@ -11,15 +11,24 @@
 import org.odk.collect.geo.maps.MapFragment;
 import org.odk.collect.geo.maps.MapPoint;
 
+import java.util.ArrayList;
 import java.util.List;
 
+import kotlin.Pair;
+
 public class FakeMapFragment implements MapFragment {
```
grzesiek2010: It would be good to convert it to kotlin too.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+                Bundle().also {
+                    it.putString(SelectionMapFragment.ARG_VIEW_MODEL_KEY, ARG_VIEW_MODEL_KEY)
+                }
+            )
+            .commit()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        viewModel.load()
```
grzesiek2010, line 102: Do we need this here to load data every time the activity is resumed, in this case to make sure set of items is updated after saving a new form an returning to this activity? If so can't we do that using activity result api? Reloading data every time a user rotates a device seems strange.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+                        formsRepositoryProvider.get(),
+                        instancesRepositoryProvider.get(),
+                        settingsProvider
+                    ) as T
+                }
+            }
+        )[ARG_VIEW_MODEL_KEY, FormMapViewModel::class.java]
+
+        val formNavigator = FormNavigator(
+            currentProjectProvider,
```
grzesiek2010: I think we can pass just current project id instead of the whole `currentProjectProvider`.

grzesiek2010: Plus I would even inject it (`FormNavigator`) instead of create here.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.odk.collect.android.utilities.InstancesRepositoryProvider
+import org.odk.collect.geo.SelectionMapFragment
+import org.odk.collect.settings.SettingsProvider
+import org.odk.collect.strings.localization.LocalizedActivity
+import javax.inject.Inject
+
+/**
+ * Show a map with points representing saved instances of the selected form.
+ */
+class FormMapActivity : LocalizedActivity() {
```
grzesiek2010, line 34: You don't have any test for this activity now. Don't you think it's needed?
The class is small so it would be simple like:
- verifying that formNavigator is called (testing the fragment result API)
- verifying that viewmodel.load is called when needed

geo/src/main/java/org/odk/collect/geo/SelectionSummarySheet.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        binding.statusIcon.setImageDrawable(ContextCompat.getDrawable(context, item.status.icon))
+        binding.statusIcon.background = null
+        binding.statusText.text = item.status.text
+
+        if (item.info != null) {
+            binding.info.text = item.info
+            binding.info.visibility = View.VISIBLE
+            binding.action.visibility = View.GONE
+        } else if (item.action != null) {
```
grzesiek2010: Isn't it possible that both item and action are nit null?

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -199,13 +216,13 @@ class SelectionMapFragment : Fragment() {
 
         summarySheetBehavior.addBottomSheetCallback(object : BottomSheetCallback() {
             override fun onStateChanged(bottomSheet: View, newState: Int) {
-                val selectedSubmissionId = selectionMapViewModel.getSelectedItemId()
+                val selectedSubmissionId = selectionViewModel.getSelectedItemId()
                 if (newState == BottomSheetBehavior.STATE_HIDDEN && selectedSubmissionId != null) {
                     map.setMarkerIcon(
                         selectedSubmissionId,
                         itemsByFeatureId[selectedSubmissionId]!!.smallIcon
```
grzesiek2010: I have a crash here:
```
java.lang.NullPointerException
        at org.odk.collect.geo.SelectionMapFragment$setUpSummarySheet$1.onStateChanged(SelectionMapFragment.kt:223)
        at com.google.android.material.bottomsheet.BottomSheetBehavior.setStateInternal(BottomSheetBehavior.java:1288)
        at com.google.android.material.bottomsheet.BottomSheetBehavior$SettleRunnable.run(BottomSheetBehavior.java:1878)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:972)
        at android.view.Choreographer.doCallbacks(Choreographer.java:796)
```

it happens if you click at least twice on a marker 
Screenshot_1647592656
that is close to your current location:

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
```diff
@@ -321,41 +340,23 @@ class SelectionMapFragment : Fragment() {
     }
 }
 
-class SelectionMapViewModel : ViewModel() {
+internal class SelectionViewModel : ViewModel() {
```
grzesiek2010: I'm not sure if I understand why you need `SelectionViewModel` and `SelectionMapViewModel` can't the first be apart of `SelectionMapViewModel`?

geo/src/main/java/org/odk/collect/geo/SelectionSummarySheet.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.core.content.ContextCompat
+import org.odk.collect.geo.databinding.SelectionSummarySheetLayoutBinding
+
+internal class SelectionSummarySheet(context: Context, attrs: AttributeSet?) :
+    FrameLayout(context, attrs) {
+
+    constructor(context: Context) : this(context, null)
+
+    val binding =
+        SelectionSummarySheetLayoutBinding.inflate(LayoutInflater.from(context), this, true)
```
grzesiek2010, line 17: The layout (`selection_summary_sheet_layout`.xml) contains nested `LinearLayouts` maybe it would be better to refactor it using `ConstraintLayout`.

grzesiek2010 (changes requested):

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: I can't reproduce this. It does seem like there is a problem there as you shouldn't be seeing "0 shown on map" will investigate that as well.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
seadowg: We probably could use the result API. Reloading data in `onResume` is an existing part of the Form Map functionality that we could change, but I think we should prioritize moving on to the GeoJSON work rather than going down that avenue here. What do you think?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
seadowg: The code in `FormMapActivity` (including the features you mention here) are covered by `FormMapTest`. Personally, if we can avoid an Activity test I'll take the opportunity!

geo/src/main/java/org/odk/collect/geo/SelectionSummarySheet.kt
seadowg: Yup that's true, but the view doesn't support it. We never actually have that scenario though, so I'll clean up seeing as you're pointing it out as confusing.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: Initially I did, but it felt off to me as `SelectionMapViewModel` is really just an interface for implementers/clients (like `FormMapActivity`) to use whereas the state of the selection (which I have in `SelectionViewModel`) is an internal detail of `SelectionMapFragment`. Does that make sense?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
seadowg: I think I'd live to avoid injecting it unless we're swapping it out for a test - we'll also want an interface if we're doing that really. How about we pass the current project to the methods themselves? Maybe that makes more sense?

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
grzesiek2010: Did you try again to reproduce it or it's not solved?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
grzesiek2010: Ok we can skip it for now but please keep in mind that this is not the best solution so that maybe when you have time later you can fix it too.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
grzesiek2010: I see what you mean but without this explanation the structure is a bit confusing especially the naming `SelectionViewModel` vs `SelectionMapViewModel`. Maybe a comment in the code would help if you can't come up with better naming?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
grzesiek2010: so you prefer espresso tests to activityscenario? O just in this case you think test cases would be duplicate?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
grzesiek2010: > How about we pass the current project to the methods themselves?

I think I like the current version more. `projectId` will stay the same after creating the activity like `settingsProvider` and `instancesRepositoryProvider`. And it is used by both methods so I think it's ok to pass it via constructor. The thing is that we don't need the whole `currentProjectProvider` if in `FormNavigator` we use it just to take `projectId`.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: I tried reproducing but couldn't cause a crash. Could you give it another go and post your exact steps if it's still happening?

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
seadowg: The latter. I think the Espresso test is good, a nice level to test the feature over all, and it drives out the code in `FormMapActivity`. I don't think there's any need for a unit test for `FormMapActivity` right now.

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
seadowg: Fair. I'll rethink the naming and see if I can make it clearer what's going on!

seadowg:
@grzesiek2010 this should be ready for another look

geo/src/main/java/org/odk/collect/geo/SelectionMapFragment.kt
grzesiek2010: I still have that crash:

```
java.lang.NullPointerException
        at org.odk.collect.geo.SelectionMapFragment$setUpSummarySheet$1.onStateChanged(SelectionMapFragment.kt:237)
        at com.google.android.material.bottomsheet.BottomSheetBehavior.setStateInternal(BottomSheetBehavior.java:1288)
        at com.google.android.material.bottomsheet.BottomSheetBehavior$SettleRunnable.run(BottomSheetBehavior.java:1878)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:972)
        at android.view.Choreographer.doCallbacks(Choreographer.java:796)
        at android.view.Choreographer.doFrame(Choreographer.java:727)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:957)
        at android.os.Handler.handleCallback(Handler.java:938)
```

I'm not a tester so I will leave it for the QA team. Everything else seems to be fine so I would add `needs testing` label but I think you should check it more carefully because the QA team will catch it either way.

seadowg: > I still have that crash:

What steps are you going through to produce this? I've tried clicking on a marker twice, but it seems to just show and then hide the summary sheet as expected.

grzesiek2010: 1. Download the `All widgets` form
2. Record location using the `GeoWidget` an d save the form as finalized
3. Open the map of forms
4. Click on the marker once to open the summary sheet
5. Click on the marker again - **CRASH**

seadowg: Is this definitely with the latest code on this PR? Sorry I'm grasping here as it seems really strange that I can't reproduce - I'm going through those same steps (even using the same form).

grzesiek2010: Yes I'm sure. I use an emulator Pixel 4 API 30 Google maps I don't know if it has anything to do with the issue.

seadowg: Right got it! It looks like the crash does not occur with Mapbox (I was using it to avoid needing the Google Maps API key). That's surprising but I'll investigate and see if I can work out what's going on.

seadowg: Oh wow this is a fun one! The problem actually stems from the fact that `GoogleMapFragment` uses a marker under the hood to display the crosshairs. This means that we get passed an ID we don't recognize to the `onFeatureClicked` listener. We should protect from that in `SelectionMapFragment` (it makes sense to just ignore features that aren't mappable items) but I'll also have a look and see if we want to be filtering out those "feature clicks" in `GoogleMapFragment`.

seadowg:
@grzesiek2010 crash fixed!

grzesiek2010 (approved):

lognaturel:
@getodk/testers Please prioritize this PR if you can. We want to get a beta with this and the next step out tomorrow (3/25) during my day. If you don't get to it, it's not a big deal, we can merge and you can test on `master`.

kkrawczyk123:
**Tested with success!**
Verified on Androids: 5,1 and 8.1

Verified cases:
Regression checks on widgets: 
- geopoint 
- geotrace
- goeshape
- geowidgets with read-only - view points
- Google Maps, OSM and Mapbox
- background location
- display submissions on map
- offline layers

srujner:
**Tested with success!**

PR #5063, opened 2022-03-16
Replace existing form with matching formid and version

### LINKED ISSUE #1693: Problem with two forms with the same id and form version

grzesiek2010:
Closes #1693 

I've tested the fix manually and also added automated tests.

This is what we have discussed. As I said:

> I think the best option (short and safe) is to insert new forms like now even if they are duplicate but collect and then remove pre-existing forms with the same id and version at the end of the process of downloading a form
Thanks to the functionality that copies media files from other forms with the same id we won't need to download them again and at the same time we won't end up with abandoned media files. So it seems like a win-win solution

This should fix the problem with duplicate forms with the same id and version. We can reproduce it using central drafts. I aslo implemented changes in syncing forms from disc. The issue I solved will be difficult to reproduce:

> I was also able to find one possible scenario when users might end up with duplicate forms.... if forms are being synchronized with the server (match exactly) we first download an xml file and then save a new row in database. If between those two steps we trigger our FormSyncTask (by opening the list of forms), then it might detect that xml file and try to add it to database too.

so please focus on regression testing to make sure that functionality works as expected.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
@@ -20,6 +20,7 @@ import org.odk.collect.shared.strings.Validator;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -52,6 +53,7 @@ public class ServerFormDownloader implements FormDownloader {
     @Override
     public void downloadForm(ServerFormDetails form, @Nullable ProgressReporter progressReporter, @Nullable Supplier<Boolean> isCancelled) throws FormDownloadException {
         Form formOnDevice;
+        List<Form> preExistingFormsWithSameIdAndVersion = new ArrayList<>();
         try {
             formOnDevice = formsRepository.getOneByMd5Hash(validateHash(form.getHash()));
         } catch (IllegalArgumentException e) {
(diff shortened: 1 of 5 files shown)


lognaturel (approved):

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
```diff
@@ -63,8 +65,8 @@ public void downloadForm(ServerFormDetails form, @Nullable ProgressReporter prog
                 formsRepository.restore(formOnDevice.getDbId());
             }
         } else {
-            List<Form> allSameFormIdVersion = formsRepository.getAllByFormIdAndVersion(form.getFormId(), form.getFormVersion());
-            if (!allSameFormIdVersion.isEmpty() && !form.getDownloadUrl().contains("/draft.xml")) {
+            preExistingFormsWithSameIdAndVersion = formsRepository.getAllByFormIdAndVersion(form.getFormId(), form.getFormVersion());
+            if (!preExistingFormsWithSameIdAndVersion.isEmpty() && !form.getDownloadUrl().contains("/draft.xml")) {
                 analytics.logEventWithParam(DOWNLOAD_SAME_FORMID_VERSION_DIFFERENT_HASH, "form", AnalyticsUtils.getFormHash(form.getFormId(), form.getFormName()));
```
grzesiek2010, line 70: I think this log doesn't make much sense now. I mean it would be good to know if it happens for central users somehow (we can't rule out that there is a bug in central too) but here we don't have that knowledge. We log the event only if it's not central draft but it might very well be any other compatible server like ONA where replacing forms and adding new ones with the same `id` and `version` is perfectly fine and we shouldn't care. Should I just get rid of it or you want to try to determine if it's central? @lognaturel @seadowg

kkrawczyk123:
**Tested with success!**

Verified cases:
- Central draft testing: upload the same version with some change, upload exactly the same form, fill submissions, send, upload another version as draft, able to fill and send both versions
- aggregate: upload the same version with some change, upload exactly the same form, fill submissions, send, upload another version, able to fill and send both versions
- regression check on match exactly
- regression check and form update: blank form, previously downloaded form, automatic download 
- regression checks done for Central and Aggregate both

srujner:
**Tested with success!**

collect_app/src/main/java/org/odk/collect/android/formmanagement/ServerFormDownloader.java
lognaturel: I've been thinking about this and can't come up with a use for the analytics. So yes, please remove.

PR #5066, opened 2022-03-17
Set proguard rules for new json validation lib

lognaturel:
Addresses https://github.com/getodk/collect/pull/5053#issuecomment-1071260344, `master` release build doesn't work.

Whoever reviewed #5053 should be banned from doing reviews. (It was me)

Created a release build, scanned a QR code.

This is the approach we previously took but with the correct library.

No regression risk. Seriously.

Not form-specific. Scanning any config QR code should do.

Nope.


lognaturel:
I'm going to use my big hammer and merge now so that I can do a beta that tracks master.

PR #5069, opened 2022-03-22
Do not report RepeatsInFieldListException as error in firebase [FIX[

grzesiek2010:

Not long time ago we decided that we shouldn't report `RepeatsInFieldListException` in firebase but it's still there: https://console.firebase.google.com/u/1/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/86ab1d9829929f39b117e831d788f498?time=last-ninety-days&versions=v2022.2-beta.0%20(4390);v2022.2-beta.1%20(4393)&sessionEventKey=6236E7C5021000011F406D9C4A71846B_1655556457296479999

The problem was that in `AuditUtils` we converted `RepeatsInFieldListException` to `RuntimeException` which should be reported so as a result `RepeatsInFieldListException` was reported too. I decided to get rid of rethrowing exceptions in `AuditUtils` at all since it's not important in that functionality, if it happens we can ignore it and the same exception will be handled in other places in the app if a form contains repeats in a field list.

It doesn't need testing.

No.

No.


lognaturel:
Looks good. Did an extra sanity check with https://docs.google.com/spreadsheets/d/159jGIYoZ5-f1JQTAWyfUFqk9-3k4mmbaM14KgbKbCbA/edit#gid=0 to make sure the dialog still looks as expected.

lognaturel (approved):

Issue #4948, opened 2021-12-30
Add support for `select_one`/`select_one_from_file` with `map` appearance and without `quick` appearance


seadowg:
## Acceptance

- [x] Given I'm on the map after clicking "Select place"
When I click on a point
And I click "Select"
Then I return to the question with the selected point as the answer

- [x] Given I'm on the map after clicking "Select place"
When I click on a point with a `label` column
Then I see the `label`

- [x] Given I'm on the map after clicking "Select place"
When I click on a point with additional non `name`/`label` columns
Then I see the `"<column name>: <column value>"` for the point

lognaturel:
https://github.com/getodk/collect/pull/5073#discussion_r836654302 has discussion about localized forms and what to do in the case of no label.

Issue #4943, opened 2021-12-24
Support `map` appearance in `select`/`select_one_from_file`


seadowg:
As discussed on [the forum](https://forum.getodk.org/t/ability-to-load-locations-into-the-geowidget-to-guide-data-collection/27406/12) (along with design details), we want to add support for `.geojson` files in Select questions. This means adding a `map` appearance for `select_one`/`select_one_from_file` (it's opaque to Collect which "select one" is actually being used in the form) that allows the enumerator to select an item from a map. Each item in the map is positioned using data from the select choice's `geometery` column.

## Backlog
 
- [x] #4947
- [x] #4949
- [x] #4948 
- [x] Make sure button text size uses dynamic font size (`answerSize`)

## Cleanup

- [x] Investigate moving shared map behaviour from `SelectionMapFragment` into `MapFragment` implementations (maintaining map zoom on recreate for instance)
- [x] Remove individual implementations of zoom behaviour
- [x] Investigate extracting `maps` module to house `MapFragment`, `MapPoint` and `MapFragmentFactory` and creating modules for each implementation (classic **D**epenednecy **I**nversion **P**rinciple setup)
- [x] Organize `geo` package structure
- [x] Add loading state/progress dialog to maps - currently DB operations are all happening on the UI thread (existing problem)
- [x] Add tests to cover bottom sheet behaviour (currently untestable due to animations always running)

lognaturel:
> Exposing standard GeoJSON Feature attributes (type and coordinates) for select choices

In the forum writeup, I proposed exposing `geometry` in ODK format. That means the type isn't needed. A first split on ";" could be used to figure out how many points there are. If there are more than one, it's a shape if the first and last points are identical. Does that seem ok, @seadowg or do you see an issue with it?

Because we're using the `map` appearance to determine that a map should be drawn, there's nothing special about the source file being `geojson` and I think work on this can start right away. By the time JavaRosa does its parsing, we won't be able to tell what the source was anyway. Any select with `geometry`, `name` and `id` fields should work.

lognaturel:
> Any select with geometry, name and id fields should work.

I think it's even better than that -- only `geometry` is needed. The XForm configures which fields to use as `name` and `label`. It will be `id` and `title` by default for geojson as it is now `name` and `label` by default for other selects but [will soon be configurable](https://github.com/XLSForm/pyxform/issues/461).

[...]

PR #5073, opened 2022-03-25
Add select one from map widget

### LINKED ISSUE #4949: Add support for `map` and `quick` appearance in `select_one`/`select_one_from_file`

seadowg:
Closes #4949

This adds a new "Select one from map" widget that can be used by adding the `map` appearance to a `select_one`. As stated in the issue, this currently only works with the `quick` appearance behavior which will be used whether `quick` is present or not.

Comments inline.

Only the new feature should need to be tested really! There are some changes to the underlying components shared with form map so a brief check there would be useful as well.

You'll need a form with `select_one`/`select_one_from_file` with the `map` appearance and `geometry` column data for the choices.
- [internal select](https://docs.google.com/spreadsheets/d/191Taxl2BU1UmBPpU4SLXEiP2cnhVx7mSe8TJeKnEDAI/edit#gid=0)
- [internal select localized](https://docs.google.com/spreadsheets/d/1j3COXsYsgA4tX4d922ho19Qm0jn9Xr7NnjxaTc6kJV4/edit#gid=0)
- [this comment](https://github.com/getodk/collect/pull/5073#issuecomment-1079266728) for a form def with 1000 points

We should add documentation for the widget but probably want to wait until we're closer to release and the experience has stabilized.


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ references:
   android_config: &android_config
     working_directory: ~/work
     docker:
-      - image: circleci/android:api-30
+      - image: cimg/android:2022.03.1
```
seadowg, line 10: Made this update to Circle's new image as it was giving me warnings when CI ran.

collect_app/src/main/java/org/odk/collect/android/activities/FormMapActivity.kt
```diff
@@ -45,28 +47,30 @@ class FormMapActivity : LocalizedActivity() {
...
+                ) as T
+            }
+        }
+    }
 
     public override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
         DaggerUtils.getComponent(this).inject(this)
-        setContentView(R.layout.form_map_activity)
+        supportFragmentManager.fragmentFactory = FragmentFactoryBuilder()
```
seadowg, line 68: Using a `FragmentFactory` here seems much cleaner to me than the previous args/service locator approach I'd taken. I've used `FragmentFactory` in a few places, so I'll discuss that here. In general, I think it gives us a really nice solution for passing data to `Fragment` instances. Because the factory is used whenever a fragment is created (both initial `Activity` creation and configuration changes etc) we get the same kind of state retention as arguments but without the need for the objects we're passing to support serialization and be small enough to be written to disk using Android's 1MB buffer. We could look at using this approach as a replacement to Dagger as well (which I think I'd be in favour of) but I didn't want to introduce that change in tactic here. 

In terms of "is this the Android thing to do" I'm less clear there. The newer Fragment docs seem to [enthusiastically suggest providing custom `FragmentFactory` implementations](https://developer.android.com/guide/fragments/fragmentmanager#dependencies). In a contradiction of that however, the [Fragment class documentation](https://developer.android.com/reference/androidx/fragment/app/Fragment#Fragment()) suggests using args instead but without any real justification. 

I'll file an issue with the docs (https://issuetracker.google.com/issues/221872195) to try and get some clarity on all this but I feel pretty good about using `FragmentFactory` in the way I am here.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -179,6 +184,25 @@ public void logFormEvent(String event) {
...
 
+    public FormEntryPrompt getQuestionPrompt(FormIndex formIndex) {
+        return formController.getQuestionPrompt(formIndex);
+    }
+
+    public void setAnswerListener(@Nullable AnswerListener answerListener) {
+        this.answerListener = answerListener;
+    }
+
+    public void answerQuestion(FormIndex index, IAnswerData answer) {
```
seadowg, line 195: I wanted to avoid using the `WaitingForDataRegistry` for this as it felt too heavy weight. This solution feels much better to me - it currently wouldn't support the acivity result use cases but should work nicely for cases that are "internal" to our app. It's currently using a listener rather than a reactive value pattern (`LiveData`) as I didn't want to move to introduce the storing answer data in a `ViewModel` quite yet (I think eventually we'd want to get to that though).

collect_app/src/main/java/org/odk/collect/android/widgets/items/ItemsWidgetUtils.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.odk.collect.android.fastexternalitemset.ItemsetDao
+import org.odk.collect.android.fastexternalitemset.ItemsetDbAdapter
+import org.odk.collect.android.fastexternalitemset.XPathParseTool
+import org.odk.collect.android.widgets.QuestionWidget
+import java.io.FileNotFoundException
+
+object ItemsWidgetUtils {
+
+    @JvmStatic
+    fun loadItemsAndHandleErrors(widget: QuestionWidget, prompt: FormEntryPrompt): List<SelectChoice> {
```
seadowg, line 18: I extracted this before realizing I only needed to call `FormEntryPrompt#getSelectChoices` for the GeoJSON stuff.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapWidget.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    override fun setOnLongClickListener(l: OnLongClickListener?) {}
+
+    override fun setData(answer: Any?) {
+        updateAnswer(answer as SelectOneData)
+    }
+
+    private fun updateAnswer(answer: SelectOneData?) {
+        this.answer = answer
+        binding.answer.text = (answer?.value as? Selection)?.choice?.labelInnerText
```
seadowg: This might not be the correct thing to do for internationalization. @lognaturel @grzesiek2010 maybe you can steer me in a more correct direction?

collect_app/src/test/java/org/odk/collect/android/widgets/items/SelectOneFromMapWidgetTest.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.odk.collect.android.support.MockFormEntryPromptBuilder
+import org.odk.collect.android.support.WidgetTestActivity
+import org.odk.collect.android.widgets.support.QuestionWidgetHelpers.promptWithAnswer
+import org.odk.collect.testshared.RobolectricHelpers.getFragmentByClass
+import org.robolectric.Robolectric
+
+@RunWith(AndroidJUnit4::class)
+class SelectOneFromMapWidgetTest {
+
+    private val activity = Robolectric.setupActivity(WidgetTestActivity::class.java)
```
seadowg: I've got an ongoing discussion (https://github.com/robolectric/robolectric/issues/7041) with the Robolectric team about what we're doing about `setupActivity` deprecation. This is a case where we need it really.

lognaturel:
New install, no perms granted:
```
03-25 10:38:54.539 21608 21608 E AndroidRuntime: java.lang.SecurityException: uid 10323 does not have any of[android.permission.ACCESS_FINE_LOCATION, android.permission.ACCESS_COARSE_LOCATION].
...
org.odk.collect.android.geo.GoogleMapFragment.onClientStart(GoogleMapFragment.java:446)
```
Maybe somewhat related to https://github.com/getodk/collect/pull/5059? I know we had a conversation about whether location perms should be required for this but I can't remember where we landed or see a public conversation about it. I think we should definitely request the perms. I believe other location widgets close if the perms are not granted. I would tend to do the same for consistency but I can see an argument for this functionality still being useful even without the device location.

lognaturel:
Form with 1000 points! With bonus script to create your own! [geojson-1000.zip](https://github.com/getodk/collect/files/8352819/geojson-1000.zip)

On my moto G5 it takes about 15s to render on Mapbox. While it's rendering, the bottom bar says "0 shown on map" (fine) and the add an instance button is shown (shouldn't). One time I got a hard crash but I'm not sure why and couldn't find the trace when I looked at the log. Once it's rendered it works as expected.

osmdroid crashes with
```
03-25 10:51:07.182 23960 23960 E AndroidRuntime: java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]
03-25 10:51:07.182 23960 23960 E AndroidRuntime: 	at org.osmdroid.util.BoundingBox.set(BoundingBox.java:74)
03-25 10:51:07.182 23960 23960 E AndroidRuntime: 	at org.osmdroid.util.BoundingBox.<init>(BoundingBox.java:49)
03-25 10:51:07.182 23960 23960 E AndroidRuntime: 	at org.osmdroid.views.overlay.Marker.setPosition(Marker.java:192)
03-25 10:51:07.182 23960 23960 E AndroidRuntime: 	at org.odk.collect.android.geo.OsmDroidMapFragment.createMarker(OsmDroidMapFragment.java:537)
```
If I generate a set of points with lat between -85 and 85, it loads almost instantly in osm!! (side note, I'm kind of obsessed with how the point distribution shows that the Earth is spherical)

Once I select a point in OSM I get:
```
03-25 10:56:46.636   946   946 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke interface method 'void org.osmdroid.api.IMapController.stopAnimation(boolean)' on a null object reference
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.disableFollowLocation(MyLocationNewOverlay.java:408)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.odk.collect.android.geo.OsmDroidMapFragment.enableLocationUpdates(OsmDroidMapFragment.java:438)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.odk.collect.android.geo.OsmDroidMapFragment.onPause(OsmDroidMapFragment.java:155)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.Fragment.performPause(Fragment.java:3200)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.FragmentStateManager.pause(FragmentStateManager.java:630)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:293)
```

The map loads quickly in Google Maps. I can't quite understand what's going on, though. Zoom to bounding box doesn't work as I expect. I click on certain markers and nothing happens. I click on certain spots without markers and a place is selected.

It occurs to me that ideally when a location is selected, the zoom on re-open should be to that location.

seadowg:
@lognaturel could you file the OSM issues separately. It looks like both problems (the crash on leaving the screen and the issue with locations outside the bounds) exist in the form map (and probably any use of OSM) as well.

> The map loads quickly in Google Maps. I can't quite understand what's going on, though. Zoom to bounding box doesn't work as I expect. I click on certain markers and nothing happens. I click on certain spots without markers and a place is selected.

Is this only with a large number of items? I haven't seen problems like this with the small sets I've been playing with.

> It occurs to me that ideally when a location is selected, the zoom on re-open should be to that location.

I've added this to the [meta issue](https://github.com/getodk/collect/issues/4943)

lognaturel:
> file the OSM issues separately. 

Will do (+ the other things I said I would write up)

> Is this only with a large number of items? 


Yes, with the form I attached. Agreed it doesn’t happen with just a few items. I can look into this further and see if I can get any ideas of what’s going on. 

> I've added this to the [meta issue](https://github.com/getodk/collect/issues/4943)

🙏

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+                val latitude = geometry.split(" ")[0].toDouble()
+                val longitude = geometry.split(" ")[1].toDouble()
+
+                list + MappableSelectItem.WithInfo(
+                    index.toLong(),
+                    latitude,
+                    longitude,
+                    R.drawable.ic_map_marker_24dp,
+                    R.drawable.ic_map_marker_48dp,
+                    selectChoice.labelInnerText,
```
seadowg: @lognaturel @grzesiek2010 should this use `FormEntryPrompt#getSelectChoiceText` instead? Also, in #4948 we described displaying the `name` if `label` was absent/blank - it doesn't look like `getSelectChoiceText` does this, so I'm not sure why we're adding that flexibility here and not to normal selects?

collect_app/src/main/java/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -157,7 +158,7 @@ public QuestionWidget(Context context, QuestionDetails questionDetails) {
      * rendering the widget. It is also passed the size to be used for question text.
      */
     @SuppressWarnings("PMD.EmptyMethodInAbstractClassShouldBeAbstract")
-    protected View onCreateAnswerView(Context context, FormEntryPrompt prompt, int answerFontSize) {
+    protected View onCreateAnswerView(@NonNull Context context, @NonNull FormEntryPrompt prompt, @NonNull int answerFontSize) {
```
grzesiek2010: `answerFontSize` has primitive type which can not be null so that annotation is useless.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+    companion object {
+        const val ARG_FORM_INDEX = "form_index"
+    }
+}
+
+internal class SelectChoicesMapData(prompt: FormEntryPrompt) : SelectionMapData {
+
+    private val mapTitle = MutableLiveData(prompt.longText)
+    private val itemCount = MutableLiveData<Int>()
+    private val items = MutableNonNullLiveData(emptyList<MappableSelectItem>())
```
grzesiek2010, line 95: Do those fields need to be livedata? The values are set only in init block so probably they could be normal fields?

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+    }
+
+    override fun getMapTitle(): LiveData<String> {
+        return mapTitle
+    }
+
+    override fun getItemType(): String {
+        return "Choices"
```
grzesiek2010: This is a part of a displayed text right? And it won't be translated?

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapWidget.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override fun onCreateAnswerView(
+        context: Context,
+        prompt: FormEntryPrompt,
+        answerFontSize: Int
+    ): View {
+        binding = SelectOneFromMapWidgetAnswerBinding.inflate(LayoutInflater.from(context))
+
+        binding.button.setOnClickListener {
+            permissionsProvider.requestLocationPermissions(
```
grzesiek2010: You will need to rebase onto master and update this method name because we have merged: https://github.com/getodk/collect/pull/5059

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapWidget.kt
```diff
@@ -0,0 +1,77 @@
...
+                object : PermissionListener {
+                    override fun granted() {
+                        DialogFragmentUtils.showIfNotShowing(
+                            SelectOneFromMapDialogFragment::class.java,
+                            Bundle().also { it.putSerializable(ARG_FORM_INDEX, prompt.index) },
+                            (context as FragmentActivity).supportFragmentManager
+                        )
+                    }
+
+                    override fun denied() {
```
grzesiek2010: If a function like this is ignored you can just have `override fun denied() = Unit`

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
grzesiek2010: I don't know I vaguely remember something that there were cases where `attachChoice` which is a part of `getSelectChoiceText` was needed but now I can't say anything for sure.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
seadowg: Yes because they need to be returned as `LiveData` in the overriden getters. You'll see the form map case (`FormMapViewModels`) makes more use of the reactive values, and we'll probably want to expand that by adding progress dialogs/loading states to these views in the future.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
lognaturel: `labelInnerText` won't work for localized forms. [Here is an example](https://docs.google.com/spreadsheets/d/1j3COXsYsgA4tX4d922ho19Qm0jn9Xr7NnjxaTc6kJV4/edit?usp=sharing). With that form none of the points are showing up on the map. Are forms without labels maybe filtered out currently?

I had described the extra flexibility because the geojson spec only talks about an optional `id` and does not define any way of including a label. We've adopted the ["simple style spec"](https://github.com/mapbox/simplestyle-spec/tree/master/1.1.0) `title` standard but that's not widely used.

We could require the label initially and relax that later if we hear that it's important. If we do that, we might want to do it for all selects.

lognaturel:
> Agreed it doesn’t happen with just a few items. I can look into this further and see if I can get any ideas of what’s going on.

👀  I can't reproduce anymore on the new branch without the `quick` appearance so probably no special action needed. 🎉

lognaturel:
> 03-25 10:51:07.182 23960 23960 E AndroidRuntime: java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]

I don't think we should worry about this. It may mean that there are issues representing points exactly at the poles with osm. My fake GPS apps are having trouble giving me points that exercise this. I think if folks are at the poles we can recommend using one of the other mapping libraries.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
seadowg: That sounds good. Right now I'll fix this up and edit out the fallback from `label` to `id` in #4948.

grzesiek2010 (approved):

kkrawczyk123:
@seadowg I can see that Collect is crashing when using OSM maps on all of the attached forms on all of my Android devices, see:
ezgif com-gif-maker (24)

It works well for Google maps and for Mapbox but it's significantly slower than on Google maps.

Stacktrace
`2022-03-31 13:22:44.317 10355-10355/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 10355
    java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]
        at org.osmdroid.util.BoundingBox.set(BoundingBox.java:74)
        at org.osmdroid.util.BoundingBox.<init>(BoundingBox.java:49)
        at org.osmdroid.views.overlay.Marker.setPosition(Marker.java:192)
        at org.odk.collect.android.geo.OsmDroidMapFragment.createMarker(OsmDroidMapFragment.java:537)
        at org.odk.collect.android.geo.OsmDroidMapFragment.access$400(OsmDroidMapFragment.java:79)
        at org.odk.collect.android.geo.OsmDroidMapFragment$MarkerFeature.<init>(OsmDroidMapFragment.java:650)
        at org.odk.collect.android.geo.OsmDroidMapFragment.addMarker(OsmDroidMapFragment.java:294)
        at org.odk.collect.geo.SelectionMapFragment.updateFeatures(SelectionMapFragment.kt:334)
        at org.odk.collect.geo.SelectionMapFragment.updateItems(SelectionMapFragment.kt:302)
        at org.odk.collect.geo.SelectionMapFragment.initMap$lambda-12(SelectionMapFragment.kt:180)
        at org.odk.collect.geo.SelectionMapFragment.$r8$lambda$6Oa2wD_syldgUPYIwnc_MDA57rM(Unknown Source:0)
        at org.odk.collect.geo.SelectionMapFragment$$ExternalSyntheticLambda5.onChanged(Unknown Source:6)
        at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:146)
        at androidx.lifecycle.LiveData$ObserverWrapper.activeStateChanged(LiveData.java:468)
        at androidx.lifecycle.LiveData$LifecycleBoundObserver.onStateChanged(LiveData.java:425)
        at androidx.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:354)
        at androidx.lifecycle.LifecycleRegistry.addObserver(LifecycleRegistry.java:196)
        at androidx.lifecycle.LiveData.observe(LiveData.java:205)
        at org.odk.collect.geo.SelectionMapFragment.initMap(SelectionMapFragment.kt:179)
        at org.odk.collect.geo.SelectionMapFragment.onViewCreated$lambda-2(SelectionMapFragment.kt:113)
        at org.odk.collect.geo.SelectionMapFragment.$r8$lambda$bY6xKjn0l8SUiBtQtyAlJVWtwCI(Unknown Source:0)
        at org.odk.collect.geo.SelectionMapFragment$$ExternalSyntheticLambda11.onReady(Unknown Source:4)
        at org.odk.collect.android.geo.OsmDroidMapFragment.lambda$onCreateView$0(OsmDroidMapFragment.java:210)
        at org.odk.collect.android.geo.OsmDroidMapFragment.$r8$lambda$9klTDIvrR0tPZAErveXmPUllvH8(Unknown Source:0)
        at org.odk.collect.android.geo.OsmDroidMapFragment$$ExternalSyntheticLambda3.run(Unknown Source:2)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)`

seadowg:
@kkrawczyk123 yeah @lognaturel found the same crash and filed an issue [here](https://github.com/getodk/collect/issues/5075). Do all your forms include a "wrap around point" as described in the issue? I'm wondering if you're actually seeing this crash with OSM when selecting a point:

```
03-25 10:56:46.636   946   946 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke interface method 'void org.osmdroid.api.IMapController.stopAnimation(boolean)' on a null object reference
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.disableFollowLocation(MyLocationNewOverlay.java:408)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.odk.collect.android.geo.OsmDroidMapFragment.enableLocationUpdates(OsmDroidMapFragment.java:438)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at org.odk.collect.android.geo.OsmDroidMapFragment.onPause(OsmDroidMapFragment.java:155)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.Fragment.performPause(Fragment.java:3200)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.FragmentStateManager.pause(FragmentStateManager.java:630)
03-25 10:56:46.636   946   946 E AndroidRuntime: 	at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:293)
```

This is also filed at #5074

kkrawczyk123:
@seadowg I have used forms that you attached to PR. It is crashing even before selecting the point.

seadowg:
> I have used forms that you attached to PR. It is crashing even before selecting the point.

@kkrawczyk123 with all 3 forms? I just pulled down "internal select" and it loaded the map fine in OSM. It crashes when picking a location, but that's filed at https://github.com/getodk/collect/issues/5074.

kkrawczyk123:
@seadowg before selecting points just like on gif I have attached on geojson form, that one with 1000 points and on internal select after selecting point. Sorry for not being clear enough about that.

seadowg:
> before selecting points just like on gif I have attached on geojson form, that one with 1000 points and on internal select after selecting point. Sorry for not being clear enough about that.

Right that looks like it's #5075 (for the 1000 points) and #5074 (for the internal select).

kkrawczyk123:
As crashes are already reported separately - **Tested with success**

Verified cases:
- forms from issue
- edit the submission and send to Central 
-  Google Maps, Mapbox, OSM, use offline layer
- regression checks on geowidgets, background location 
- regression checks on external geo widgets 

srujner:
**Tested with success** 

PR #5076, opened 2022-03-28
Make json schema identifiers more specific

lognaturel:
I talked to @grzesiek2010 and @seadowg briefly about this and figured I'd make the change instead of filing an issue. Happy to entertain alternatives, though. The `title` is intended to stand more on its own. The idea with changing the URI is to have a story in case we have other kinds of settings we want to specify. None of this has been released yet so the change should have no impact.

### DIFF
--- a/settings/src/main/resources/settings_schema.json
+++ b/settings/src/main/resources/settings_schema.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema",
-  "$id": "https://getodk.com/settings.schema.json",
-  "title": "Settings",
+  "$id": "https://getodk.com/client-settings.schema.json",
+  "title": "ODK Client Settings",
   "type": "object",
   "properties": {
     "project": {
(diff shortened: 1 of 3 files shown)


settings/src/main/resources/settings_schema.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema",
-  "$id": "https://getodk.com/settings.schema.json",
-  "title": "Settings",
+  "$id": "https://getodk.com/client-settings.schema.json",
```
seadowg: Can we set up a forward or DNS for this so that it resolves to the file? I think changing the name of the file itself would be good as well.

seadowg (changes requested):

settings/src/main/resources/settings_schema.json
lognaturel: Yes

lognaturel:
@seadowg Resources that are going to be accessed through `R.` need to have filenames that are valid Java identifiers. I can't see any reason for a restriction on this filename. I have tried this on a real device as well.

seadowg:
> Resources that are going to be accessed through R. need to have filenames that are valid Java identifiers. I can't see any reason for a restriction on this filename. I have tried this on a real device as well.

Ah I think the file originally lived in `res` so would have been subject to those restrictions. Definitely nicer to have the name it has.

seadowg (approved):

PR #5077, opened 2022-03-30
Refactored list of forms

grzesiek2010:
I've played with the list of forms manually. I've also added some new automated tests. A lot of existing tests go through the list of forms too so the coverage seems to be pretty good.
 
It's one of the biggest prs I remember (in terms of the number of commits) so although they form a pretty logical structure it might be also fine to take a look at it as a whole.

Big changes worth mentioning:
- replaced deprecated loaders with viewmodel (see https://developer.android.com/topic/libraries/architecture/viewmodel?gclid=CjwKCAjwrqqSBhBbEiwAlQeqGr4p9fb84SptocOjoV8inSuM_3ls5iGfLLVIMBXiHoivCw357l_PrxoC318QAvD_BwE&gclsrc=aw.ds#loaders)
- reimplemented the list of blank forms using `RecyclerView`
- separated the new `BlankFormListActivity` from the old hierarchy of lists `AppListActivity -> FormListActivity -> FillBlankFormListActivity`. Such a hierarchy does not make sense anymore we have a lot of logic that is used only in this one activity (more than in the past) and in the future it might differ even more from other lists.
- manage scanning forms on storage using coroutines instead of AsyncTask (see: https://github.com/getodk/collect/pull/5077/files#diff-753aa7ab27aa86da15dd4c38ca810b17bc884db22cec003d19c2ad5451a5c523R130)
- the new implementation is a little bit simpler - I removed reopening the sorting list dialog or the search bar after rotating a device. It was implemented in the past where we used to handle things like that but then we realized it doesn't make sense and had that conversation that we should handle cases that are crucial (for example answers in a form) but not such elements that a user can easily reopen and it's not a problem.

**TODO**
Please verify that performance is not worse compared to the master branch (forms should not be loaded slower when the list is started or when a user sorts/filters them. You might need to use a long list of forms to test this scenario).

No.

No.


collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    fun syncWithServer(): LiveData<Boolean> {
+        logManualSyncWithServer()
+        val result = MutableLiveData<Boolean>()
+        scheduler.immediate(
+            { formsUpdater.matchFormsWithServer(projectId) },
+            { value: Boolean ->
+                loadFromDatabase()
```
grzesiek2010, line 171: In a perfect world after syncing with server or storage if our database is update the viewmodel should be informed about this fact (repository should return livedata) and reload the forms automatically. Unfortunately we do not support returning livedata from our repositories. We should change it in the future implementing it somehow for sqlie or switching to Room where it works out of the box.

collect_app/src/main/java/org/odk/collect/android/formlists/FormListRecyclerView.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.DividerItemDecoration
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import org.odk.collect.android.R
+
+class FormListRecyclerView(context: Context, attrs: AttributeSet?) : RecyclerView(context, attrs) {
```
grzesiek2010, line 11: This and `FormListSortingDialog` are classes that can be shared between form lists (maybe something else too but that will be clear once we refactor other lists). Other classes should belong to particular lists only (like the `blankformlist` package) to avoid spaghetti code since those lists have a lot logic used only in one of them.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
seadowg: Another approach might be to have a reactive layer in front of the repositories. We are already doing this for Instances with `InstancesAppState`. I've only started running through this PR now (and probably won't get through it all). We can discuss more later, but just wanted to throw that idea in quickly while it's in my head!

grzesiek2010: Yes but it doesn't have to be a part of this pr. We can even finish this one and refactor other lists and then at the end ad such and improvement for all of them at once.

collect_app/src/main/java/org/odk/collect/android/formlist/FormListAdapter.kt
```diff
@@ -28,7 +30,7 @@ class FormListAdapter : RecyclerView.Adapter<FormListAdapter.ViewHolder>() {
 
     override fun getItemCount() = filteredFormItemsList.size
 
-    class ViewHolder(private val binding: FormListItemBinding) : RecyclerView.ViewHolder(binding.root) {
+    inner class ViewHolder(private val binding: FormListItemBinding) : RecyclerView.ViewHolder(binding.root) {
```
seadowg: I'd prefer this to not be `inner` and just passed the listener so that there's no implicit state available to `ViewHolder`.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
```diff
@@ -81,7 +80,7 @@ class BlankFormListViewModelTest {
     private val context = ApplicationProvider.getApplicationContext<Application>()
     private val syncRepository: SyncStatusAppState = mock()
     private val formsUpdater: FormsUpdater = mock()
-    private val scheduler = FakeScheduler()
+    private val scheduler = TestScheduler()
```
seadowg: Would `FakeScheduler` not be better here? It makes it explicit in the tests that things are happening asynchronously and hopefully avoids the tests hiding problems that could come up.

collect_app/src/main/AndroidManifest.xml
```diff
@@ -111,7 +111,7 @@ the specific language governing permissions and limitations under the License.
             android:windowSoftInputMode="stateHidden" />
         <activity android:name=".activities.InstanceChooserList" />
         <activity
-            android:name=".activities.FillBlankFormActivity"
+            android:name=".formlists.blankformlist.BlankFormListActivity"
             android:launchMode="singleTop" />
```
seadowg, line 131: I don't seem to be able to get any tests to fail by removing `singleTop` here. Obviously that coverage would have been missing before as well, but do you have any idea why we need this to be `singleTop`?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListAdapter.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.odk.collect.android.R
+import org.odk.collect.android.databinding.BlankFormListItemBinding
+import org.odk.collect.androidshared.ui.multiclicksafe.MultiClickGuard
+import timber.log.Timber
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class FormListAdapter : RecyclerView.Adapter<FormListAdapter.ViewHolder>() {
+    private var formItems = emptyList<BlankFormListItem>()
+    lateinit var listener: OnFormItemClickListener
```
seadowg: Could the listener not just be passed in as a constructor param to avoid the `lateinit` here?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.odk.collect.androidshared.ui.DialogFragmentUtils
+import org.odk.collect.permissions.PermissionListener
+import org.odk.collect.permissions.PermissionsProvider
+import org.odk.collect.strings.localization.LocalizedActivity
+import javax.inject.Inject
+
+class BlankFormListActivity : LocalizedActivity(), OnFormItemClickListener {
+
+    @Inject
+    lateinit var viewModelFactory: BlankFormListViewModel.Factory
```
seadowg, line 30: I think you can just construct `BlankFormListViewModel.Factory` directly here as there are no tests that need to swap the factory out yet - nicer to avoid the extra Dagger setup if we don't need it.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListAdapter.kt
```diff
@@ -0,0 +1,70 @@
...
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import org.odk.collect.android.R
+import org.odk.collect.android.databinding.BlankFormListItemBinding
+import org.odk.collect.androidshared.ui.multiclicksafe.MultiClickGuard
+import timber.log.Timber
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class FormListAdapter : RecyclerView.Adapter<FormListAdapter.ViewHolder>() {
```
seadowg: I like the separation here - the ViewModel cares about the fact that we're dealing with blank forms, but the adapter could be used with any list of forms. Nice!

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
```diff
@@ -0,0 +1,144 @@
...
+                    )
+                }
+
+                override fun denied() = Unit
+            }
+        )
+    }
+
+    private fun initObservers() {
+        viewModel.backgroundTasksObserver.observe(this) {
```
seadowg: Could we just expose a `LiveData<Boolean>` here like `isLoading` instead of `backgroundTasksObserver`?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        viewModel.syncResult.observe(this) { result ->
+            if (!result.isConsumed()) {
+                SnackbarUtils.showShortSnackbar(findViewById(R.id.formList), result.value)
+            }
+        }
+
+        viewModel.formsToDisplay.observe(this) { forms ->
+            findViewById<RecyclerView>(R.id.formList).visibility = if (forms.isEmpty()) View.GONE else View.VISIBLE
```
seadowg: These lines need formatted (try the auto-formatter in Android Studio)

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java
```diff
@@ -85,7 +85,8 @@ public FormEntryPage clickOnForm(String formName) {
 
     private void clickOnFormButton(String formName) {
         assertFormExists(formName);
-        onData(withRowString(DatabaseFormColumns.DISPLAY_NAME, formName)).perform(click());
+        onView(withId(R.id.formList))
+                .perform(RecyclerViewActions.actionOnItem(hasDescendant(allOf(withId(R.id.form_title), withText(formName))), click()));
```
seadowg: I think here (and in the use of `actionOnItem`) below, we could just use `hasDescendant(withText(formName))` and avoid needing to specify the ID.

collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Inject
+    lateinit var settingsProvider: SettingsProvider
+
+    private val viewModel: BlankFormListViewModel by viewModels { viewModelFactory }
+
+    public override fun onCreate(bundle: Bundle?) {
+        super.onCreate(bundle)
+        DaggerUtils.getComponent(this).inject(this)
+
+        viewModel.formsToDisplay.observe(this) { forms -> showFormListDialog(forms) }
```
seadowg: I think this could crash if the forms load after this observation occurs as value of `formsToDisplay` could be `null`. The real problem there is just that `formsToDisplay` should be `LiveData<List<BlankFormListItem?>>` instead of `LiveData<List<BlankFormListItem>>` I think.

collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.os.Bundle
+import android.view.LayoutInflater
+import androidx.recyclerview.widget.DefaultItemAnimator
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+import org.odk.collect.android.R
+import org.odk.collect.android.adapters.SortDialogAdapter
+
+class FormListSortingDialog(
```
seadowg: Let's name this `FormListSortingBottomSheetDialog` - for a second there, I thought you'd changed the UI to a normal dialog.

collect_app/src/main/res/layout/activity_blank_form_list.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical">
+
+    <include layout="@layout/app_bar_with_progress_bar_layout" />
+
+    <org.odk.collect.android.formlists.FormListRecyclerView
+        android:id="@+id/formList"
```
seadowg: This should be snake cased really

testshared/build.gradle
```diff
@@ -51,4 +51,5 @@ dependencies {
     implementation(Dependencies.androidx_fragment_testing) {
         exclude group: 'androidx.test', module: 'monitor' //fixes issue https://github.com/android/android-test/issues/731
     }
+    implementation Dependencies.kotlinx_coroutines_test
```
seadowg: Do we need this?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java
```diff
@@ -53,28 +50,31 @@ public BlankFormSearchPage searchInBar(String query) {
...
+                .check(matches(hasDescendant(allOf(withId(R.id.map_button), withEffectiveVisibility(ViewMatchers.Visibility.GONE)))));
         return this;
     }
 
     public FormMapPage clickOnMapIconForForm(String formName) {
-        onData(allOf(is(instanceOf(Cursor.class)), CursorMatchers.withRowString(DatabaseFormColumns.DISPLAY_NAME, is(formName))))
-                .onChildView(withId(R.id.map_button))
-                .perform(click());
+        onView(withId(R.id.formList))
+                .perform(RecyclerViewActions.actionOnItem(hasDescendant(allOf(withId(R.id.form_title), withText(formName))), clickOnViewChild(R.id.map_button)));
```
seadowg: Could we use content description instead of ID here?

collect_app/src/main/res/layout/blank_form_list_item.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
```
seadowg, line 1: Please run auto-format on this to reorder attributes

seadowg (changes requested):
I've still got to have a deeper look at `BlankFormListViewModel` and `BlankFormListMenuDelegate` but thought I'd just give you my notes on everything else now seeing as there has been a big delay on getting feedback here.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
```diff
@@ -0,0 +1,363 @@
...
+import org.odk.collect.formstest.FormUtils
+import org.odk.collect.formstest.InMemFormsRepository
+import org.odk.collect.settings.keys.ProjectKeys
+import org.odk.collect.shared.settings.InMemSettings
+import org.odk.collect.testshared.BooleanChangeLock
+import org.odk.collect.testshared.TestScheduler
+
+@RunWith(AndroidJUnit4::class)
+class BlankFormListViewModelTest {
+    private val form1 = Form.Builder()
```
seadowg: There's a lot of noise here from setting up fixtures. Two things I'd think about here:

1. Not sharing fixtures between tests - I think just setting up state specifically for a test is more readable and makes it clearer why a fixture is set up in a certain way. For instance, only `deleted forms should be ignored` needs a deleted form to exist so in that test you can just set `formsRepository` up with a deleted form and a non-deleted form. A few of the tests probably don't even need there to be any forms. Additionally, sorting or filtering tests also only need two forms rather than three or four.
2. Using a default fixture/`copy` pattern like in [Fixtures.kt](https://github.com/getodk/collect/blob/master/geo/src/test/java/org/odk/collect/geo/support/Fixtures.kt) in `geo` - for this to work as nicely as it does there we'd need to convert `Form` to Kotlin and make it a data class, but I think that feels like a nice switch.

In my opinion, doing 2 makes 1 feel less like it adds noise to individual tests. I think generally in tests it's actually good to strive for as much isolation as possible while not obscuring too much setup in test helpers. It's a hard balance of course.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    @Test
+    fun `syncWithStorage should be triggered when viewModel is initialized`() {
+        verify(formsDirDiskFormsSynchronizer).synchronizeAndReturnError()
+    }
+
+    @Test
+    fun `syncWithStorage return correct result`() {
+        assertThat(viewModel.syncResult.value?.value, `is`("Result text"))
```
seadowg, line 61: Again, it might be clearer to only set up a sync result here. Right now it's not immediately obvious why we get `"Result test"`.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    @Test
+    fun `syncWithStorage should not be triggered when viewModel is initialized if another process operates on forms database`() {
```
seadowg: Maybe just `when viewModel is initialized if forms lock is locked`

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
```diff
@@ -0,0 +1,363 @@
...
+
+        assertThat(viewModel.isMatchExactlyEnabled(), `is`(true))
+    }
+
+    @Test
+    fun `isSyncingWithServer follows repository isSyncing`() {
+        val liveData = MutableLiveData(true)
+        whenever(syncRepository.isSyncing(projectId)).thenReturn(liveData)
+
+        val isSyncing = LiveDataTester().activate(viewModel.isSyncingWithServer())
```
seadowg: `LiveDataTester` should really be an instance level field so `teardown` can be called in an `@After`. It might be a good time to make a `LiveDataTesterRule` to make that all a little less boilerplate(y).

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+            sortAndFilter()
+        }
+
+    private val shouldHideOldFormVersions: Boolean
+        get() {
+            return generalSettings.getBoolean(ProjectKeys.KEY_HIDE_OLD_FORM_VERSIONS)
+        }
+
+    init {
+        loadFromDatabase()
```
seadowg, line 84: I don't think we need this here as `syncWithStorage` eventually calls `loadFromDatabase` itself. I can remove this call and no tests fail.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+    private fun logManualSyncWithServer() {
+        val uri = Uri.parse(generalSettings.getString(ProjectKeys.KEY_SERVER_URL))
+        val host = if (uri.host != null) uri.host else ""
+        val urlHash = getMd5Hash(ByteArrayInputStream(host!!.toByteArray())) ?: ""
+        analytics.logEvent(AnalyticsEvents.MATCH_EXACTLY_SYNC, "Manual", urlHash)
+    }
+
+    private fun sortAndFilter() {
+        _formsToDisplay.value = when (sortingOrder) {
+            0 -> _allForms.value.sortedBy { it.formName }
```
seadowg, line 223: Personally, I think we're probably fine here, but I wanted to highlight this to @lognaturel. Do you think there's any performance concerns around sorting/filtering in memory as opposed to in DB? We could definitely use [`asSequence`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-sequence.html) here to optimize this in memory at the very least.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegate.kt
```diff
@@ -0,0 +1,117 @@
...
+                    viewModel.sortingOrder = newSortingOrder
+                }.show()
+
+                true
+            }
+            else -> false
+        }
+    }
+
+    init {
```
seadowg: This should move to the top of the class.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt
```diff
@@ -0,0 +1,219 @@
...
+        menuDelegate.onPrepareOptionsMenu(menu)
+
+        menu.findItem(R.id.menu_filter).expandActionView()
+        menu.findItem(R.id.menu_filter).collapseActionView()
+
+        assertThat(menu.findItem(R.id.menu_refresh).isVisible, `is`(false))
+    }
+
+    @Test
+    fun `filterText should be cleared when menu is created`() {
```
seadowg, line 204: Why?

seadowg (changes requested):
Ok here are some more thoughts! Overall, I'm really liking the direction of this PR. Definitely keen to see us reusing a bunch of these components across the app!

collect_app/src/main/AndroidManifest.xml
grzesiek2010: I don't know why we have this. Maybe to avoid having duplicate activities when a user opens it directly fro an external app?  But then it should be added in all activities that are exported... Do you think we should get rid of it?

collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt
grzesiek2010: > I thought you'd changed the UI to a normal dialog.

Why? no it's still a bottom dialog.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
grzesiek2010: I don't think it's slower but we can ask the QA team to verify this case.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt
grzesiek2010: In order to clear it when the screen is rotated for example. I mentioned above that I simplified it a bit: 

> the new implementation is a little bit simpler - I removed reopening the sorting list dialog or the search bar after rotating a device. It was implemented in the past where we used to handle things like that but then we realized it doesn't make sense and had that conversation that we should handle cases that are crucial (for example answers in a form) but not such elements that a user can easily reopen and it's not a problem.

so we need to clear the text in view model too.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt
grzesiek2010: It's not required but I thought that maybe we should load data from database first in case when syncing would take a lot of time. I'm pretty sure the old approach worked like that. Do you think we should wait for syncing?

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java
grzesiek2010: What's the benefit? And which id are you talking about `R.id.formList` or `R.id.form_title` or `R.id.map_button`?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
grzesiek2010: But there are other injected objects so it's not like I can remove dagger from this class at all. Do you still think it doesn't make sense to use dagger for the future like that?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
grzesiek2010: That would require creating my own mediator? Is that what you meant?

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegate.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
grzesiek2010: Done.

collect_app/src/main/res/layout/blank_form_list_item.xml
grzesiek2010: Done.

collect_app/src/main/res/layout/activity_blank_form_list.xml
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListAdapter.kt
grzesiek2010: Right, done.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java
grzesiek2010: It looks like you don't like ids in espresso tests? I guess you think that in tests like that we don't need to verify on which position in a view (in this case `form_title`) that text is displayed just check it is somewhere there. I'm I right?

collect_app/src/main/java/org/odk/collect/android/formlist/FormListAdapter.kt
grzesiek2010: Ok.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
grzesiek2010: Done.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt
grzesiek2010: `formsToDisplay` has no initial value so observers won't be notified until data is loaded.

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt
grzesiek2010: I've improved it a bit. Maybe not entirely in a way you described because I wanted to avoid converting Form to kotlin but it should be a bit clearer now.

collect_app/src/main/AndroidManifest.xml
seadowg: Does seem a little risky to kill now on top of all the other changes... Let's leave it.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
seadowg: It's not a big deal, but it would be nice to avoid adding things to `AppDependencyModule` unless we really need them.

collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt
seadowg: Instead of just zipping the three `LiveData` objects together you could zip3 them then do a `Transformations.map` to get a `LiveData<Boolean>`:

```kotlin
val isLoading: LiveData<Boolean> = Transformations.map(LiveDataUtils.zip3(
    isFormLoadingRunning,
    isSyncingWithStorageRunning,
    isSyncingWithServerRunning,
)) { (one, two, three) -> one || two || three  }
```

Alternatively, I think you could use a count and a `LiveData<Boolean` (like a counting lock):

```kotlin

private var loadingCount = 0
val isLoading: LiveData<Boolean> = MutableLiveData<Boolean>(false)

private fun loadFromDatabase() {
    startLoading()
    scheduler.immediate(
        foreground = { .. },
        background = {
             finishLoading()
        }
    )
}

private fun startLoading() {
    loadingCount = loadingCount + 1
    isLoading.value = loadingCount > 0
}

private fun finishLoading() {
    loadingCount = loadingCount - 1
    isLoading.value = loadingCount > 0
}

File: collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java (hunk 11 above)
Comment (seadowg) [reply collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java]: > I guess you think that in tests like that we don't need to verify on which position in a view (in this case form_title) that text is displayed just check it is somewhere there

Correct. I'd tend to try and push testing around ordering or specific views (if it feels like there's a good reason to do so) down to lower levels like a unit test for a `View` or `Adapter`. Generally these tests are faster, but more importantly to me, the setup is also often simpler (you're not setting up state for the whole app, just the component in question).

File: collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt (hunk 12 above)
Comment (seadowg) [reply collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt]: Could we still change the type and correct the observer accordingly? I've realized that unless we're using `NonNullLiveData` we should always be marking the `LiveData` type as `?` as there's no (compiler level) protection against a `null` creeping in.

File: collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt (hunk 13 above)
Comment (seadowg) [reply collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt]: Because the name didn't mention "BottomSheet" and just said "Dialog". I think it's good to call out that this isn't a "modal" dialog (which would be my assumption from the name).

File: collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java (hunk 16 above)
Comment (seadowg) [reply collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java]: > And which id are you talking about R.id.formList or R.id.form_title or R.id.map_button?

Sorry that wasn't clear. I meant for the `map_button` (the content description should be something like "Open form map"). I think we should also be able to match the item through just the form name itself rather than needing to specify that the name is in the `form_title`.

> What's the benefit?

1. Using IDs increases coupling between the test and the actual implementation - we want to avoid as much of that as possible to make change as easy as possible later
2. Using content descriptions instead of IDs on views without text verifies that we've configured that view to work correctly with screen readers

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt (hunk 22 above)
Comment (seadowg) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt:84]: > that maybe we should load data from database first in case when syncing would take a lot of time

Hmmm yeah I think I agree with you. Can we add a test that checks that loading happens first, and then we sync (and reload)? That should be possible using `FakeScheduler`.

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt (hunk 25 above)
Comment (seadowg) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: Are we then doing extra work to prevent the filter text from being maintained on rotation/configuration change though?

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt (hunk 23 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt:223]: I've added it to the description above.

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt (hunk 7 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt:30]: Yeah but `BlankFormListViewModel.Factory` takes a lot of things in constructor which I would have to inject here anyway so it would be even worse than having this one field.

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt (hunk 25 above)
Comment (grzesiek2010) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: I clear the value stored in the viewmodel https://github.com/getodk/collect/pull/5077/files#diff-80daf2f5f67255e3fec799d7214bb0507e04c6b5c4f87075b318b6417014cd99R65 but the UI part is cleared automatically because it's recreated.

File: collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt (hunk 12 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt]: So you want to have type like `MutableLiveData<List<BlankFormListItem>?>` and always check if the list is not null?

File: collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java (hunk 16 above)
Comment (grzesiek2010) [reply collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java]: Ok, done.

File: collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt (hunk 13 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/formlists/FormListSortingDialog.kt]: Right, done.

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt (hunk 9 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt]: Done.

File: collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java (hunk 11 above)
Comment (grzesiek2010) [reply collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java]: Ok, done.

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt (hunk 7 above)
Comment (seadowg) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListActivity.kt:30]: Yeah, true. OK happy to go with it the way it is right now!

File: collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt (hunk 12 above)
Comment (seadowg) [reply collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt]: Yeah, I think that's best! It's annoying that `LiveData` doesn't support optional typing properly. It looks like some people are advocating a switch to [`StateFlow`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/) in Kotlin for that reason although I haven't done a lot of reading on it.

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt (hunk 25 above)
Comment (seadowg) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: Right but instead, could we just not clear the value in the `ViewModel` and filter text would be retained on rotation for free?

File: collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt (hunk 12 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt]: Ok so I added it in https://github.com/getodk/collect/pull/5077/commits/94243c02d87e4781e3c2aa219d9ed8bfaac3afd4

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt (hunk 25 above)
Comment (grzesiek2010) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: No because the search view is closed automatically and I don't want to reopen it like I don't want to reopen the sorting dialog (I think those things are not that important). So if I remove that clearing the result will be that we see closed search view but filtered forms.

Comment (seadowg) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: > So if I remove that clearing the result will be that we see closed search view but filtered forms.

Ah I see! I think adding a comment to the test explaining this would be helpful.

Comment (grzesiek2010) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListMenuDelegateTest.kt:204]: Done.

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt (hunk 4 above)
Comment (grzesiek2010) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt:38]: `FakeScheduler` is great if you need control over background tasks (postpone running them to check something before and then after). In this test I don't need it so I think using a simpler solution (according to https://developer.android.com/kotlin/coroutines/test) is better here.
Another thing is that `FakeScheduler` would need to be improved for this tests because when `BlankFormListViewModel` is created two background tasks are started (see https://github.com/getodk/collect/pull/5077/files#diff-753aa7ab27aa86da15dd4c38ca810b17bc884db22cec003d19c2ad5451a5c523R84)
but `FakeScheduler` keeps only one and once another one is started the first one is removed:
https://github.com/getodk/collect/blob/master/testshared/src/main/java/org/odk/collect/testshared/FakeScheduler.kt#L17

any thoughts?

Comment (grzesiek2010) [reply collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModelTest.kt]: Done.

File: testshared/build.gradle (hunk 15 above)
Comment (grzesiek2010) [reply testshared/build.gradle]: I needed this in my `TestScheduler` but I got rid of it so no no we don't.

File: collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt (hunk 22 above)
Comment (grzesiek2010) [reply collect_app/src/main/java/org/odk/collect/android/formlists/blankformlist/BlankFormListViewModel.kt:84]: Done.

File: collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormFixtures.kt [hunk 26]
```diff
@@ -0,0 +1,52 @@
+package org.odk.collect.android.formlists.blankformlist
+
+import org.odk.collect.forms.Form
+import org.odk.collect.formstest.FormUtils
+
+object BlankFormFixtures {
```
seadowg: This is definitely nicer, but for me, it's still hard to immediately see in each test what the purpose of each fixture is. For instance, `blankForm4` is important because it's deleted, but I can't tell that without clicking through and then reading the definition carefully. I think we could make it a little easier to understand how a fixture is being set up without needing to click through by using a single function with optional args here:

```kotlin
fun form(formId: String = "1", date: Long = 0, deleted: Boolean = false) {
    return Form.Builder()
        .dbId(1)
        .formId(formId)
        .version("1")
        .displayName("Form ${formId}")
        .date(date)
        .deleted(deleted)
        .formFilePath(FormUtils.createXFormFile(formId, "1").absolutePath)
        .build()
}
```

Then in tests it would be:

```kotlin
@Test
fun `deleted forms should be ignored`() {
    saveForms(
        BlankFormFixtures.form(formId = "1"), 
        BlankFormFixtures.form(formId = "2", deleted = true)
    )
    createViewModel()
        
    assertThat(viewModel.formsToDisplay.value!!.size, `is`(1))
    assertFormItem(viewModel.formsToDisplay.value!![0], BlankFormFixtures.blankForm1)
}
```

I feel like that makes it easier to understand the state that the test is setting up. What do you think?

collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt
```diff
@@ -45,7 +45,11 @@ class AndroidShortcutsActivity : AppCompatActivity() {
         super.onCreate(bundle)
         DaggerUtils.getComponent(this).inject(this)
 
-        viewModel.formsToDisplay.observe(this) { forms -> showFormListDialog(forms) }
+        viewModel.formsToDisplay.observe(this) { forms ->
+            forms?.let {
+                showFormListDialog(it)
```
seadowg: Just realized: this will end up triggering on `onResume` as well as triggering multiple times when the forms are updated via disk sync and db loading etc. We could show the dialog once and set an adapter that's updated with the `LiveData` or create a `DialogFragment` that does something similar. Alternatively, we could just show the forms as a list in `Activity` directly (with a basic `RecyclerView`/`Adapter` combo). I don't think there's any reason to use a dialog here.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt
grzesiek2010: > Just realized: this will end up triggering on onResume

This is ok because we need to create this dialog again to display it.

> as well as triggering multiple times when the forms are updated via disk sync and db loading

Yes this is not perfect but it would be a very rare case I guess.

> Alternatively, we could just show the forms as a list in Activity directly (with a basic RecyclerView/Adapter combo). I don't think there's any reason to use a dialog here.

I agree but this is something for a separate issue I think

collect_app/src/test/java/org/odk/collect/android/formlists/blankformlist/BlankFormFixtures.kt
grzesiek2010: Done.

collect_app/src/main/java/org/odk/collect/android/external/AndroidShortcutsActivity.kt
seadowg: > This is ok because we need to create this dialog again to display it.

Right but we're creating a new dialog and calling `show()` multiple times without dismissing the dialog that was previously shown. I tried it out with the debugger, and it looks like we create the dialog a minimum of two times (I'm guessing because of the double `loadFromDatabase` that happens in `BlankFormListViewModel`). However, it seems like this doesn't result in the user having to hit back multiple times to get out of the screen due to the listener passed to `setOnCancelListener` finishing the Activity. There's no visible bug here, but it does feel incorrect and could potentially lead to some confusion or problems in the future, so I think we should fix it before moving on.

grzesiek2010: Taking into account that:
- it looks like it's the last thing to fix
- I will be off next 2 weeks 
- the issue was about reworking `BlankFormListActivity` not `AndroidShortcutsActivity`

I think it's the best option to take a step back and keep the old behavior in `AndroidShortcutsActivity` what I did https://github.com/getodk/collect/pull/5077/commits/a04ff5b609abe0b9a94afc0dec3091d39df9ce84

When I'm back I can rework `AndroidShortcutsActivity` separately removing the dialog and displaying a list of forms in a recyclerview.

seadowg (approved):

kkrawczyk123:
@grzesiek2010 I have noticed that sorting forms by newest and the oldest works in the opposite way right now, see:
ezgif com-gif-maker (30)

The issue is not visible on master so it looks like a regression connected to PR.

seadowg:
@kkrawczyk123 the sort orders should be correct now.

srujner:
**Tested with Success!**
Verified on Androids 10,11 and 12

Verified Cases: 
- Issue with sorting is no longer reproducing;
- Performance for loading list, sorting and filtering is comparable with one on master branch;
- Regression check on: Sending finalized Forms, Editing saved Forms, Viewing sent Forms, Deleting saved Forms; 

kkrawczyk123:
Tested successfully also on Androids: 5.1 and 8.1

PR #5079, opened 2022-03-31
Add non `quick` appearance for select one from map widget

### LINKED ISSUE #4948: Add support for `select_one`/`select_one_from_file` with `map` appearance and without `quick` appearance

### LINKED ISSUE #5074: Crash after selecting feature using osmdroid
#### Problem description

```
java.lang.NullPointerException: Attempt to invoke interface method 'void org.osmdroid.api.IMapController.stopAnimation(boolean)' on a null object reference
	at org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.disableFollowLocation(MyLocationNewOverlay.java:408)
	at org.odk.collect.android.geo.OsmDroidMapFragment.enableLocationUpdates(OsmDroidMapFragment.java:438)
	at org.odk.collect.android.geo.OsmDroidMapFragment.onPause(OsmDroidMapFragment.java:155)
	at androidx.fragment.app.Fragment.performPause(Fragment.java:3200)
	at androidx.fragment.app.FragmentStateManager.pause(FragmentStateManager.java:630)
	at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:293)
	at androidx.fragment.app.FragmentStore.moveToExpectedState(FragmentStore.java:113)
```

The only way I've found to reproduce is to set the basemap as osm, use a form for selecting a feature (e.g. [this one](https://docs.google.com/spreadsheets/d/191Taxl2BU1UmBPpU4SLXEiP2cnhVx7mSe8TJeKnEDAI/edit#gid=0)) with https://github.com/getodk/collect/pull/5073 and select a feature. @seadowg mentioned [here](https://github.com/getodk/collect/pull/5073#issuecomment-1079742986) that it's existing behavior so it may be reproducible with the form instance map.

#### Expected behavior
No crash.

#### Other information 
The selection seems to work, there's just a crash after.

seadowg:
Closes #4948 
Closes #5074 
~~Blocked by #5073~~

In addition to the non `quick` appearance this also reorganizes the `geo` module (something @grzesiek2010 suggested earlier) and fixes the crash with OSM. Given there are a bunch of import changes this will definitely be easier to review commit by commit than using the "Files changed" tab.

Comments inline.

We'll skip QA for this and release as a beta.


androidshared/src/main/java/org/odk/collect/androidshared/system/ContextUtils.kt
```diff
@@ -12,4 +16,36 @@ object ContextUtils {
         context.theme.resolveAttribute(resId, outValue, true)
         return outValue.data
     }
+
+    /** Renders a Drawable (such as a vector drawable) into a Bitmap.  */
+    @JvmStatic
+    fun getBitmap(context: Context, drawableId: Int): Bitmap? {
+        val drawable = ContextCompat.getDrawable(context, drawableId)
+        if (drawable != null) {
```
grzesiek2010: Are you doing here anything special? If not you can probably use the `toBitmap` extension function:
`drawable.toBitmap()`

maps/src/test/java/org/odk/collect/maps/layers/MapFragmentReferenceLayerUtilsTest.kt
```diff
@@ -1,17 +1,16 @@
...
 import org.junit.runner.RunWith
-import org.odk.collect.android.geo.DirectoryReferenceLayerRepository
-import org.odk.collect.geo.maps.MapFragment
+import org.odk.collect.maps.MapFragment
 import org.odk.collect.shared.TempFiles.createTempDir
+import org.robolectric.RobolectricTestRunner
 import java.io.File
 
-@RunWith(AndroidJUnit4::class)
+@RunWith(RobolectricTestRunner::class)
```
grzesiek2010, line 13: Why did you change this?

geo/src/main/res/layout/property.xml
```diff
@@ -0,0 +1,31 @@
...
+        android:id="@+id/text"
+        style="?textAppearanceBody2"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/margin_small"
+        android:textColor="?colorOnSurface"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintStart_toEndOf="@id/icon"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_goneMarginStart="0dp"
```
grzesiek2010, line 28: Does this really change anything here?

geo/src/main/java/org/odk/collect/geo/selection/SelectionSummarySheet.kt
```diff
@@ -31,14 +33,32 @@ internal class SelectionSummarySheet(context: Context, attrs: AttributeSet?) :
...
+            property.text.text = it.text
+
+            if (it.icon != null) {
+                property.icon.setImageDrawable(ContextCompat.getDrawable(context, it.icon))
+                property.icon.background = null
+            } else {
+                property.icon.visibility = View.GONE
+            }
+
+            binding.properties.addView(property.root)
```
grzesiek2010, line 51: Shouldn't it be a list (recyclerview) instead of building it manually like that using `addView()`?

androidshared/src/main/java/org/odk/collect/androidshared/system/ContextUtils.kt
seadowg: I was just moving this around so hadn't given it a proper look, but I think you're right. It's only used for the OSM crosshairs anyway and switching it to using `toBitmap` seems to work!

maps/src/test/java/org/odk/collect/maps/layers/MapFragmentReferenceLayerUtilsTest.kt
seadowg: The `maps` module doesn't currently have any dependencies on AndroidX Test dependencies, so I used `RobolectricTestRunner` directly rather than having `AndroidJUnit4` delegate to it anyway. Annoyingly you can't just include `androidx.test.ext:junit` and use `AndroidJUnit4` in a JVM test without also depending on Robolectric, so because I didn't need anything else out of AndroidX Test I just opted to go with vanilla JUnit and Robolectric.

geo/src/main/res/layout/property.xml
seadowg: Yeah it's needed. It's definitely a strange part of `ConstraintLayout`. Basically, if the `icon` is `GONE` then `text` still has a margin (defined by `marginStart`) that we don't want. In a `RelativeLayout` or `LinearLayout` we could just flip where the margin lives (move it to `icon`), but in `ConstraintLayout` the margins need to live on the same view as the constraint, or they don't do anything. I was confused as to how I could get it to work and some Googling revealed that `goneMarginStart` is the answer. It's a way of stating "the margin when the target of the start constraint is gone".

geo/src/main/java/org/odk/collect/geo/selection/SelectionSummarySheet.kt
seadowg: Right now we're not optimizing for longer sets of properties (although we're interested to see if people end up with those) and we don't want the properties to be independently scrollable within the bottom sheet so a `LinearLayout` seems fine to me.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -106,14 +107,20 @@ internal class SelectChoicesMapData(private val resources: Resources, prompt: Fo
                 val latitude = geometry.split(" ")[0].toDouble()
                 val longitude = geometry.split(" ")[1].toDouble()
 
+                val properties = selectChoice.additionalChildren.filter {
```
lognaturel, line 110: I think we'll likely want to consider doing this on opening of the bottom sheet rather than preemptively. I expect that would have an appreciable effect on performance.

Less significant, but this filter then map does two passes over the properties, right? I do expect there will be cases where the property list will be significant. Of course if we just do this on bottomsheet open it's not as important. Probably no action to take here but wanted to share that I had to think through this in case it sparks other ideas.

lognaturel (approved):
I'm kind of blown away that we now have independent `map` and `geo` modules with the plan to further split maps. I keep feeling like we need to index these modules somewhere more prominent because they could legitimately be used in other projects.

I think we should talk about alternate ways of displaying the properties (e.g. maybe we eventually move towards more of a table look with alternating row shading) and some perf things I briefly highlighted. But otherwise this looks solid.

lognaturel:
I believe the order that properties are shown in is flipped from the order that they are defined in. E.g. with [this form](https://docs.google.com/spreadsheets/d/1j3COXsYsgA4tX4d922ho19Qm0jn9Xr7NnjxaTc6kJV4/edit#gid=759748424) I expect the order to be `name`, `description`, `annual_visits` but it's the opposite. Or maybe order is not preserved at all?

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
seadowg: Yeah, and as we've chatted about before, I think there are some good opportunities to have a lot of this happen off the UI thread (and with some loading UI state) as well.

lognaturel:
>  believe the order that properties are shown in is

Ooh. Wow. `pyxform` seems to sort them alphabetically on conversion. 😬😬😬 @seadowg https://github.com/XLSForm/pyxform/issues/602

kkrawczyk123:
**Verified on Androids 8.1 and 5.1 with cases:**
- Google Maps, Mapboxs, OSM
- offline layers
- select point, edit selection, finalize form
- screen rotation, minimization

I have noticed crashes on Android 5.1 that were not visible before so reported issue in #5084

srujner:
**Tested with success!**
Verified on Androids 10 and 11.

No crashes occurred

PR #5083, opened 2022-04-04
Improvements to `SelectOneFromMapWidget`

seadowg:
This adds a few things:

* A progress indicator is now shown when loading choices on the map (and also on the form map when loading instances)
* If a choice has been selected, clicking "Select Place" opens the map with that choice centered (and with its details showing)
* The widget now works properly with field list recomputations

Comments inline

It would be good to test all the above changes (including a quick look at the form map).

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
@@ -1,0 +1,23 @@
+package org.odk.collect.android.support.rules
+
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class RetryOnDeviceErrorRule : TestRule {
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                try {
+                    base.evaluate()
+                } catch (e: Throwable) {
+                    if (e::class.simpleName == "RootViewWithoutFocusException") {
+                        base.evaluate()
+                    } else {
+                        throw e
+                    }
+                }
+            }
+        }
+    }
+}
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
@@ -11,6 +11,7 @@ import org.odk.collect.android.external.InstanceProvider
 import org.odk.collect.androidshared.livedata.MutableNonNullLiveData
 import org.odk.collect.androidshared.livedata.NonNullLiveData
 import org.odk.collect.async.Scheduler
+import org.odk.collect.forms.Form
 import org.odk.collect.forms.FormsRepository
 import org.odk.collect.forms.instances.Instance
 import org.odk.collect.forms.instances.InstancesRepository
@@ -25,23 +26,19 @@ import java.util.Locale
 
 class FormMapViewModel(
     private val resources: Resources,
-    formId: Long,
-    formsRepository: FormsRepository,
+    private val formId: Long,
+    private val formsRepository: FormsRepository,
     private val instancesRepository: InstancesRepository,
     private val settingsProvider: SettingsProvider,
     private val scheduler: Scheduler
 ) : ViewModel(), SelectionMapData {
 
-    private var mapTitle = MutableLiveData<String>()
+    private var _form: Form? = null
+
+    private val mapTitle = MutableLiveData<String>()
     private var mappableItems = MutableNonNullLiveData<List<MappableSelectItem>>(emptyList())
     private var itemCount = MutableLiveData(0)
-    private val form = formsRepository.get(formId)!!
-
     private val isLoading = MutableNonNullLiveData(false)
-
-    init {
-        mapTitle.value = form.displayName
-    }
 
     override fun getMapTitle(): LiveData<String> {
         return mapTitle
@@ -68,6 +65,7 @@ class FormMapViewModel(
 
         scheduler.immediate(
             background = {
+                val form = _form ?: formsRepository.get(formId)!!.also { _form = it }
                 val instances = instancesRepository.getAllByFormId(form.formId)
                 val items = mutableListOf<MappableSelectItem>()
 
@@ -88,11 +86,12 @@ class FormMapViewModel(
                     }
                 }
 
-                Pair(items, instances.size)
+                Triple(form.displayName, items, instances.size)
             },
             foreground = {
-                mappableItems.value = it.first
-                itemCount.value = it.second
+                mapTitle.value = it.first
+                mappableItems.value = it.second
+                itemCount.value = it.third
                 isLoading.value = false
             }
         )
--- a/collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
@@ -37,7 +37,7 @@ class FormMapViewModel(
     private var itemCount = MutableLiveData(0)
     private val form = formsRepository.get(formId)!!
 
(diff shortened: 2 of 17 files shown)


lognaturel:
Some notes for my own sanity. @seadowg and I have discussed adding the following here (correct if I got it wrong):
- lazily get extra properties when the bottom sheet is shown (https://github.com/getodk/collect/pull/5079#discussion_r840980035)
- on open, if a value is selected, display as selected (big marker, centered at current zoom level) 

@seadowg, @getodk/testers haven't completed https://github.com/getodk/collect/pull/5079 yet. Should they maybe wait until this is in as well so they can go over the feature end-to-end?

seadowg:
> haven't completed https://github.com/getodk/collect/pull/5079 yet. Should they maybe wait until this is in as well so they can go over the feature end-to-end?

No I think it's fine to continue. Those changes will be pretty small and additional so might not even need QA.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -87,20 +93,47 @@ class SelectOneFromMapDialogFragment : MaterialFullScreenDialogFragment(), Fragm
...
+        const val ARG_SELECTED_INDEX = "selected_index"
     }
 }
 
-internal class SelectChoicesMapData(private val resources: Resources, prompt: FormEntryPrompt) :
+internal class SelectChoicesMapData(
+    private val resources: Resources,
+    scheduler: Scheduler,
+    prompt: FormEntryPrompt,
+    private val selectedIndex: Int?
```
seadowg, line 104: We need to pass the selected index through because the widget answer will be "dirty" (in the widget but not in the `FormEntryPrompt`) when return from the map. We ended up with something similar in the select minimal implementation - we pass through the current answer rather than relying on the `FormEntryPrompt`.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapWidget.kt
```diff
@@ -64,12 +70,14 @@ class SelectOneFromMapWidget(context: Context, questionDetails: QuestionDetails)
...
     override fun clearAnswer() {
         updateAnswer(null)
+        widgetValueChanged()
     }
 
     override fun setOnLongClickListener(l: OnLongClickListener?) {}
 
     override fun setData(answer: Any?) {
         updateAnswer(answer as SelectOneData)
+        widgetValueChanged()
```
seadowg, line 80: I realized we needed this added so that `SelectOneFromMapWidget` works with field lists properly (dependent computations for example)

material/src/main/java/org/odk/collect/material/MaterialProgressDialogFragment.java
```diff
@@ -134,4 +140,15 @@ private void setArgument(String key, Serializable value) {
     public View getDialogView() {
         return dialogView;
     }
+
+    public static void showOn(LifecycleOwner lifecycleOwner, NonNullLiveData<Boolean> liveData, FragmentManager fragmentManager, Supplier<MaterialProgressDialogFragment> supplier) {
```
seadowg, line 144: This feels like it's a pretty common pattern so nice to have a helper for it.

collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.runners.model.Statement
+
+class RetryOnDeviceErrorRule : TestRule {
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                try {
+                    base.evaluate()
+                } catch (e: Throwable) {
+                    if (e::class.simpleName == "RootViewWithoutFocusException") {
```
seadowg, line 14: This exception seems to be thrown when the tests can't focus on the app which seems to be a device error that we can't work around. I think always retrying in this scenario is safe as it's unlikely to happen in a genuine fail and if it is from an error in our test logic (trying to assert on the app when the menu drawer is still up), it should happen repeatedly (so we'll still see a fail).

There is a small risk that we have a test where we do something like close the notification drawer and that results in an animation we're not accounting for causing this failure sometimes and that that gets obscured. Again, this wouldn't be an error in the code but a small instability in our test code that's being hidden which I think I can live with.

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
```diff
@@ -55,29 +56,45 @@ class FormMapViewModel(
...
-                    // In GeoJSON, longitude comes before latitude.
-                    val lon = coordinates.getDouble(0)
-                    val lat = coordinates.getDouble(1)
-
-                    items.add(createItem(instance, lat, lon))
-                } catch (e: JSONException) {
-                    Timber.w("Invalid JSON in instances table: %s", instance.geometry)
+        isLoading.value = true
+
+        scheduler.immediate(
```
grzesiek2010, line 66: Shouldn't this use `viewmodelscope`? As i understand scheduler.immediate which launches `CoroutineScope` won't be canceled when `viewmodel` is destroyed or maybe I'm missing something?

grzesiek2010 (changes requested):

collect_app/src/main/java/org/odk/collect/android/formmanagement/formmap/FormMapViewModel.kt
seadowg: We haven't actually used the new `viewModelScope` yet - our `Scheduler` interface doesn't support it. I'm not sure if we get a lot of benefit here, as all it would do is cancel the task if you leave the screen while the instances are loading. It's potentially wasting a bit of CPU time to keep going, but it shouldn't have any negative consequences otherwise, as the view code will be detached from the `LiveData` object that the results get put into anyway. 

Maybe we should create an issue to create support for `viewModelScope` in some way in `Scheduler` and then come back for this and other `Scheduler` uses in `ViewModel` classes. How does that sound?

grzesiek2010: Ok we can have a separate issue. This case seems like a perfect example where `viewModelScope` should be used. Didn't it exist back then when you created that interface?

seadowg: Potentially! It was likely either undocumented or still alpha/beta. That, or I just missed it. It's important to remember that `viewModelScope` cancels coroutines on `ViewModel#onCleared` but only if the coroutine "cooperates" with being cancelled. By that, I mean we'd need to actually be checking `isActive` within the code to have cancellation avoid us doing extra work in the background. Without that, we're only really going to end up cancelling our `foreground` work, which will usually just amount to setting a value on `LiveData` (which should no longer be observed if the `ViewModel` has been destroyed). There's some good docs on this [here](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative).

I guess in this specific case (and in anywhere the background task is "quick"), I'd argue that we wouldn't really gain much from cancellation. Does that make sense? Thinking about it some more. I'm feeling like we should only really be introducing specific scopes/lifecycle aware tasks for longer running jobs (form download would be a good example) and it might be that we want to move all that stuff to `WorkManager` anyway.

grzesiek2010: Yes that's right `viewModelScope` is recommended everywhere for such cases (in tutorials/codelabs) but our code is not prepared to take advantage of it as you said.

lognaturel (approved):

lognaturel:
Really nice improvements to behavior! 🎉

I did a brief review of the code mostly to check for edge cases that might not be covered or things like that. @grzesiek2010 you'll probably want to do another pass.

@getodk/testers as the PR description says, some things to try out are select from map in field list, opening the map when there's already an item selected, loading a map with many items.

PR #5088, opened 2022-04-07
Fixed testing notifications

grzesiek2010:
I've ran tests locally and on firebase.

I've added some improvements to handling notification (commits 1-3). However I think they do not fix problems with tests, the real fix is the last commit where I added some more time to assert pages that might be open from notifications - it seems like it was the issue.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ErrorPage.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ErrorPage.kt
@@ -8,18 +8,12 @@ import androidx.test.espresso.matcher.ViewMatchers.withParent
 import org.hamcrest.Matchers.allOf
 import org.hamcrest.Matchers.instanceOf
 import org.odk.collect.android.R
-import org.odk.collect.android.support.WaitFor.waitFor
-import java.util.concurrent.Callable
 
 class ErrorPage : Page<ErrorPage>() {
 
     override fun assertOnPage(): ErrorPage {
-        return waitFor(
-            Callable {
-                assertText(R.string.errors)
-                this
-            }
-        )
+        assertText(R.string.errors)
+        return this
     }
 
     fun assertError(errorMessage: String): ErrorPage {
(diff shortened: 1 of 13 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/NotificationDrawer.kt
```diff
@@ -29,7 +30,7 @@ class NotificationDrawer {
         appName: String,
         title: String,
         body: String? = null,
-        subtext: String? = null
+        subtext: String = Project.DEMO_PROJECT_NAME
```
seadowg: I think I'd prefer this to be explicitly in the tests rather than a default. It seems strange for `NotificationDrawer` to know or care about `Project`.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ErrorPage.kt
```diff
@@ -8,12 +8,18 @@ import androidx.test.espresso.matcher.ViewMatchers.withParent
...
 import org.odk.collect.android.R
+import org.odk.collect.android.support.WaitFor.waitFor
+import java.util.concurrent.Callable
 
 class ErrorPage : Page<ErrorPage>() {
 
     override fun assertOnPage(): ErrorPage {
-        assertText(R.string.errors)
-        return this
+        return waitFor(
```
seadowg: Can we move these `waitFor` calls into the `NotificationDrawer`? I can totally understand us needing `waitFor` when returning from a notification, but we shouldn't need it when moving from one page within the app to the other.

seadowg (changes requested):

seadowg (approved):
@lognaturel I've updated this with additional fixes for the problems we were seeing with retries. We should merge this ASAP so that we get back to stable but good for you to double-check that nothing looks suspect.

lognaturel (approved):

PR #5095, opened 2022-04-11
Mapping performance improvements

### LINKED ISSUE #5087: Improve select from map performance with Mapbox and Google Maps
Collect v2022.2 beta

#### Problem description
Select one from map is prohibitively slow on Mapbox and Google Maps. A 1000-point form is almost unusable with Mapbox but fine with Google Maps. A 5000-point form is unusuable with both. It loads nearly instantly with osmdroid.

Use the thousand-point form [from the forum](https://forum.getodk.org/t/odk-collect-v2022-2-beta/36913/6). Add points using the script in the zip there or use [gilpin.zip](https://github.com/getodk/collect/files/8433510/gilpin.zip)) for 5000.

grzesiek2010:
Closes #5087 

I've tested the performance manually.

No.

No.


grzesiek2010:
On slack I said:

```
I think I know what makes mapbox so slow... the problem is that adding markers we create symbols for every marker:
    private Symbol createSymbol(SymbolManager symbolManager, MapPoint point, boolean draggable, @IconAnchor String iconAnchor, int iconDrawableId) {
        return symbolManager.create(new SymbolOptions()
            .withLatLng(toLatLng(point))
            .withIconImage(addIconImage(iconDrawableId))
            .withIconSize(1f)
            .withSymbolSortKey(10f)
            .withDraggable(draggable)
            .withTextOpacity(0f)
            .withIconAnchor(getIconAnchorValue(iconAnchor))
        );
    }

that symbolManager.create creates a new symbol and calls updateSource() what according to the comment
Trigger an update to the underlying source
what is not performant because we don't need to call that update for every marker. Fortunately there is a method create(List<S> optionsList) which we could use to add all markers at once and call that update just once.
Good news - the time of loading of 1k items goes down from 9s to 1s
Bad news it will require some more changes in our codebase
```

The last commit shows how displaying all markers in one go improves the loading time. In case of mapbox that time is now less than 1s where before it lasted 9s to load 1k items. When it comes to `osmdroid` and `google` there is no change (google even does not support adding multiple markers as a collection). That means we would only improve mapbox so the question is: does that make sense to implement such changes just for that one provider which we might remove in the future at all? The code at least in the current version is a bit more complex.

Another option would be to roll back changes implemented in google and osmdroid since there is no change in performance and keep it only in mapbox so that even if we remove it in the future we won't have those additional changes in other providers.

collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
```diff
@@ -356,12 +355,29 @@ private void resetLocationComponent() {
...
         int featureId = nextFeatureId++;
-        features.put(featureId, new MarkerFeature(featureId, symbolManager, point, draggable, iconAnchor));
+        features.put(featureId, new MarkerFeature(featureId, symbolManager, point, draggable, iconAnchor, iconDrawableId));
         return featureId;
     }
 
+    @Override public void displayMarkers() {
+        List<SymbolOptions> symbolOptions = new ArrayList<>();
+
+        for (Map.Entry<Integer, MapFeature> entry : features.entrySet()) {
```
grzesiek2010, line 367: This is pretty nasty as I said during the meeting... For example I had to change features to `LinkedHashMap` to make sure the order is preserved.
I don't love it but with current architecture this is the easiest solution. We might want to rework it in the future.

maps/src/main/java/org/odk/collect/maps/MapFragment.java
```diff
@@ -123,7 +123,7 @@ void addTo(
      * By default, the marker icon will be R.drawable.ic_map_point.
      * Returns a positive integer, the featureId for the newly added shape.
      */
-    int addMarker(MapPoint point, boolean draggable, @IconAnchor String iconAnchor);
+    int addMarker(MapPoint point, boolean draggable, @IconAnchor String iconAnchor, int iconDrawableId);
```
lognaturel, line 126: Consider overloading instead. `addMarker(point, iconDrawableId, iconAnchor)`, eg. That way you won't end up having to duplicate the resource.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -350,13 +350,7 @@ class SelectionMapFragment(
...
         for (item in items) {
             val point = MapPoint(item.latitude, item.longitude)
-            val featureId = map.addMarker(point, false, MapFragment.BOTTOM)
-
-            map.setMarkerIcon(
-                featureId,
-                if (featureId == selectedFeatureViewModel.getSelectedFeatureId()) item.largeIcon else item.smallIcon
-            )
-
+            val featureId = map.addMarker(point, false, MapFragment.BOTTOM, item.smallIcon)
```
lognaturel, line 349: This ends up working because currently `onFeatureClicked` always ends up getting called after this and adjusts the icon size.

And actually, I wonder whether the `setMarkerIcon` below ever actually changed icon size. As far as I can tell, the selected feature only gets set in `selectedFeatureViewModel` when `onFeatureClicked` is called the first time. In general I have a lot of confusion around this area: https://github.com/getodk/collect/pull/5099#discussion_r848921970

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -354,6 +350,9 @@ class SelectionMapFragment(
             itemsByFeatureId[featureId] = item
             points.add(point)
         }
+        selectedFeatureViewModel.getSelectedFeatureId()?.let {
```
lognaturel, line 354: I'm confused about why this is needed at all. I would expect https://github.com/getodk/collect/pull/5095/commits/14f6a6a218ad4ea8f9f8d8e584197b577e2c3339#diff-41e26ac0b5e19db6790e339c006bc7c1ce451d326733985f3309aa5f3f3d6d89R317 to do the actual feature click. And in fact, I would expect that `getSelectedFeatureId` is always null here because it gets set in `onFeatureClicked`.

collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
lognaturel: I agree this doesn't feel ideal. Did you consider adding a `addMarkers` method to the interface that takes in a `List<MapPoint>` and returns a `List<Integer>`? In the Google case, it would call `addMarker` in a loop and in the other two it would use the more performant add all functionality.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
grzesiek2010: It is here to keep already selected point highlighted (big marker and the bottom sheet) after recreating - rotating a device for example.

maps/src/main/java/org/odk/collect/maps/MapFragment.java
grzesiek2010: I don't know... ideally it was kotlin and I could use defaults.

collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
grzesiek2010: Yes. Maybe it would be a bit better it depends...
First of all passing a list of `MapPoint` would not be enough. I need at least icons for every point so that would need to be two list or a map which would be a ix of both.
Another problem is that I would end up with exactly the same nasty iterating through map of features to build the list of `symbolOptions` and then setup symbols in every `markerFeature` after displaying them.

grzesiek2010: This rather requires more changes in the way we create those features.

seadowg:
Closing this right now, and we'll come back to performance for hotfixes to the release.

Issue #5084, opened 2022-04-05
Collect is crashing when use select one from map widget on Google Maps on Android 5.1


kkrawczyk123:
#### Software and hardware versions 
Collect current master, Android v 5.1 device used...

#### Problem description
When testing #5079  I have noticed 2 crashes when used forms form #5073 when Google map is selected on Android 5.1 device only. 1st is happening when edit selected point and 2nd when try to finalize form after selecting point.

#### Steps to reproduce the problem
**Scenario 1:**
1. Open ex. map-internal-select form
2. Tap on Select place
3. Tap on one of points on the map
4. Confirm using Select button
5. Tap on Select place - Collect crashes

**Scenario 2:**
1. Open ex. map-internal-select form
2. Tap on Select place
3. Tap on one of points on the map
4. Confirm using Select button
5. Swipe to Save and Exit Screen

stacktrace:
`2022-04-05 10:05:06.064 11303-11303/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 11303
    java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter context
        at org.odk.collect.android.geo.MapsMarkerCache.getMarkerBitmap(MapsMarkerCache.kt)
        at org.odk.collect.android.geo.GoogleMapFragment.getBitmapDescriptor(GoogleMapFragment.java:648)
        at org.odk.collect.android.geo.GoogleMapFragment.access$100(GoogleMapFragment.java:70)
        at org.odk.collect.android.geo.GoogleMapFragment$MarkerFeature.setIcon(GoogleMapFragment.java:692)
        at org.odk.collect.android.geo.GoogleMapFragment.setMarkerIcon(GoogleMapFragment.java:270)
        at org.odk.collect.geo.selection.SelectionMapFragment$setUpSummarySheet$1.onStateChanged(SelectionMapFragment.kt:231)
        at com.google.android.material.bottomsheet.BottomSheetBehavior.setStateInternal(BottomSheetBehavior.java:1288)
        at com.google.android.material.bottomsheet.BottomSheetBehavior$SettleRunnable.run(BottomSheetBehavior.java:1878)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:135)
[...]

Issue #5075, opened 2022-03-28
Handle invalid geometries


lognaturel:
#### Software and hardware versions 
Collect v2022.1 and prior

#### Problem description
There are two overlapping problems:
- The `geometry` field is not validated prior to the corresponding choice being rendered. It could contain anything. We've seen crashes because of attempts to use geojson instead of ODK-formatted features, for example
- Mapbox and osmdroid crash when handed geometry that is out of bounds at least in the instance form map and possibly in other map uses. Google Maps appears to wrap around in the case of longitude and render off map in the case of latitude. This one may need to be spun off depending on where the proposed dialog described below is shown from.

#### Steps to reproduce the problem
The easiest is to use the select from map with a [form that has invalid points](https://docs.google.com/spreadsheets/d/1CDGcthyS7KJgnJikmJttLdwFS4xsSysDRuoxdVQS2zQ/edit?usp=sharing). You can also enter incorrect geometry (like `blah`).

#### Expected behavior

Shows a dialog (when points are invalid or geometry is invalid) that reads: "The choice with label <choice label> has invalid geometry (<geometry>)". Clicking OK returns to the form (from the dialog).

lognaturel:
There’s a somewhat related issue at the poles: https://github.com/osmdroid/osmdroid/issues/1339#issuecomment-491519929

seadowg:
Related to this we can also end up with `geometry` in an incorrect format (an example could be `0, 0, 0, 0`). We're seeing that appear in the wild in [this Crashlytics report](https://console.firebase.google.com/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/db72f993c1c08aa205771dfc1767bdbf?time=last-seven-days&versions=v2022.2.0%20(4418)&types=crash&sessionEventKey=625814AE002F00016447C11EC9338ADA_1664882394848775403).

PR #5098, opened 2022-04-12
Launch select from map with zoom on current location

### LINKED ISSUE #5093: Zoom to current location when launching select from map view
Currently we launch with a viewport that includes all of the available points. As briefly discussed in https://github.com/getodk/collect/issues/5087#issuecomment-1091928594, let's launch zoomed on the current location instead. It's likely users will want to toggle between the two views. In many use cases starting with current location is most helpful.

Note that this appears to be the current behavior with osmdroid.

lognaturel:
Closes #5093

Ran with osmdroid, Google Maps, Mapbox. Added test.

We briefly considered changing the behavior for the form instance map as well. That map is intended to be a summary of work done, though, whereas the select from map is more likely to be used for tasking upcoming work. It makes sense that they would start with a different zoom.

I also considered doing some additional work on making the zoom deterministic here. I've decided to do it separately in case it's controversial https://github.com/getodk/collect/pull/5099.

The intentional behavior change is that now all three mapping engines should start the select from map view zoomed in to the device location.

Any with select from map (https://forum.getodk.org/t/odk-collect-v2022-2-beta-select-from-map-geojson-datasets/36913/6)

Should be included in https://github.com/getodk/docs/issues/1458

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
@@ -53,7 +53,8 @@ class SelectOneFromMapDialogFragment : MaterialFullScreenDialogFragment(), Fragm
                 SelectionMapFragment(
                     SelectChoicesMapData(resources, scheduler, prompt, selectedIndex),
                     skipSummary = Appearances.hasAppearance(prompt, Appearances.QUICK),
-                    showNewItemButton = false
+                    showNewItemButton = false,
+                    zoomToFeatureBoundingBoxOnLoad = false
                 )
             }
             .build()
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
@@ -54,7 +54,7 @@ class SelectOneFromMapDialogFragment : MaterialFullScreenDialogFragment(), Fragm
                     SelectChoicesMapData(resources, scheduler, prompt, selectedIndex),
                     skipSummary = Appearances.hasAppearance(prompt, Appearances.QUICK),
                     showNewItemButton = false,
-                    zoomToFeatureBoundingBoxOnLoad = false
+                    zoomToFitItems = false
                 )
             }
             .build()
--- a/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
+++ b/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
@@ -36,7 +36,7 @@ class SelectionMapFragment(
     val selectionMapData: SelectionMapData,
     val skipSummary: Boolean = false,
     val showNewItemButton: Boolean = true,
-    val zoomToFeatureBoundingBoxOnLoad: Boolean = true
+    val zoomToFitItems: Boolean = true
 ) : Fragment() {
 
     @Inject
@@ -321,7 +321,7 @@ class SelectionMapFragment(
         if (selectedFeatureId != null) {
             onFeatureClicked(selectedFeatureId)
             viewportInitialized = true
-        } else if (zoomToFeatureBoundingBoxOnLoad && !viewportInitialized && points.isNotEmpty()) {
+        } else if (zoomToFitItems && !viewportInitialized && points.isNotEmpty()) {
             map.zoomToBoundingBox(points, 0.8, false)
             viewportInitialized = true
         }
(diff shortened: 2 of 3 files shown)


lognaturel:
One thing to consider about this implementation is that if no location is available on load it might feel jerky. That is, the viewport will be initialized to whatever the default center is and then all of a sudden when a location is acquired it will change. That would particularly be problematic in rural settings where there's only GPS to rely on.

An alternative would be to forget about the location change listener and at time of viewport initialization, go to the current location if there is one and fall back to the feature bounding box.

lognaturel:
This also possibly intersects with a conversation we've had on the forum about initial zoom levels: https://forum.getodk.org/t/geo-optimizing-zoom-levels-for-walking-surveyors/18941

I think we should start with this and explicitly ask field teams about how this default zoom level is working for them.

seadowg (approved):
I made one change to naming but looks good!

kkrawczyk123:
I have noticed crash on all my devices when used form with 1000 points and OSM Maps, Collect crashes just after tapping on Select place button. 

Stacktrace:
`2022-04-13 13:20:28.768 19972-19972/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 19972
    java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]
        at org.osmdroid.util.BoundingBox.set(BoundingBox.java:74)
        at org.osmdroid.util.BoundingBox.<init>(BoundingBox.java:49)
        at org.osmdroid.views.overlay.Marker.setPosition(Marker.java:192)
        at org.odk.collect.android.geo.OsmDroidMapFragment.createMarker(OsmDroidMapFragment.java:611)
        at org.odk.collect.android.geo.OsmDroidMapFragment.access$400(OsmDroidMapFragment.java:84)
        at org.odk.collect.android.geo.OsmDroidMapFragment$MarkerFeature.<init>(OsmDroidMapFragment.java:742)
        at org.odk.collect.android.geo.OsmDroidMapFragment.addMarker(OsmDroidMapFragment.java:322)
        at org.odk.collect.geo.selection.SelectionMapFragment.updateFeatures(SelectionMapFragment.kt:354)
        at org.odk.collect.geo.selection.SelectionMapFragment.updateItems(SelectionMapFragment.kt:317)
        at org.odk.collect.geo.selection.SelectionMapFragment.initMap$lambda-14(SelectionMapFragment.kt:195)
        at org.odk.collect.geo.selection.SelectionMapFragment.$r8$lambda$SlaLN6kg4v5ozQIoXqTju2L70Dw(SelectionMapFragment.kt)
        at org.odk.collect.geo.selection.SelectionMapFragment$$ExternalSyntheticLambda6.onChanged(Unknown Source)
        at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:151)
        at androidx.lifecycle.LiveData.setValue(LiveData.java:309)
        at org.odk.collect.androidshared.livedata.MutableNonNullLiveData.setValue(NonNullLiveData.kt:22)
        at org.odk.collect.android.widgets.items.SelectChoicesMapData._init_$lambda-1(SelectOneFromMapDialogFragment.kt:127)
        at org.odk.collect.android.widgets.items.SelectChoicesMapData.$r8$lambda$kznZHyHp7KqYIiw6zCS5HBSyEV8(SelectOneFromMapDialogFragment.kt)
        at org.odk.collect.android.widgets.items.SelectChoicesMapData$$ExternalSyntheticLambda0.accept(Unknown Source)
        at org.odk.collect.async.CoroutineScheduler$immediate$1.invokeSuspend(CoroutineScheduler.kt:17)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:706)`
        
It works well for forms with2 points. 
cc @seadowg @grzesiek2010

srujner:
Reproduced also on Androids 10 and 11

seadowg:
@srujner @kkrawczyk123  it looks like that crash is this issue: https://github.com/getodk/collect/issues/5075. I'm pretty sure @lognaturel's script/geojson file will always cause that - that form post should be updated to avoid people our beta testers running into it. I don't think we should be using it for QA either.

lognaturel:
Not quite the same, the points are valid in this case. It’s https://github.com/getodk/collect/pull/5073#issuecomment-1080952039 explanation at https://github.com/osmdroid/osmdroid/issues/1339#issuecomment-491519929 but fine to just not use that form.

kkrawczyk123:
**Tested with success!**

Verified cases:
- Zoom to current location
- zoom to all points
- screen rotation
- display points on map
- regression check on display points on map 
- regression check on geowidgets
- Google, Mapbox, OSM (Crashes from #5084 still visible, and crash mentioned above)
- offline layers

srujner:
**Tested with success!**

PR #5099, opened 2022-04-12
Make launch viewport for selection map deterministic

### LINKED ISSUE #5093: Zoom to current location when launching select from map view

lognaturel:
And addresses non-determinism described at https://github.com/getodk/collect/issues/5093#issuecomment-1095284696

This is going to conflict with https://github.com/getodk/collect/pull/5095! Those changes are more important. I do think it would be best to also fix the non-determinism if possible. Maybe take inspiration from here and do another PR once #5095 is merged?

Tried on device with all three mapping engines. Added a test.

Comments inline.

I think the regression risk is low. There aren't tests for all the possible combinations of start state (and I don't think there should be). There could be an inadvertent change to one of the combinations.

Any with a select one from map.

No.

### DIFF
--- a/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
+++ b/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
@@ -184,9 +184,6 @@ class SelectionMapFragment(
         }
 
         map.setGpsLocationEnabled(true)
-        map.setGpsLocationListener { point -> onLocationChanged(point) }
-
-        previousState?.let { restoreZoomFromPreviousState(it) }
 
         map.setFeatureClickListener(::onFeatureClicked)
         map.setClickListener { onClick() }
@@ -194,10 +191,8 @@ class SelectionMapFragment(
         selectionMapData.getMappableItems().observe(viewLifecycleOwner) {
             updateItems(it)
             updateCounts(binding)
-        }
 
-        selectedFeatureViewModel.getSelectedFeatureId()?.let {
-            onFeatureClicked(it)
+            initializeViewport()
         }
     }
 
@@ -210,14 +205,37 @@ class SelectionMapFragment(
         )
     }
 
-    private fun restoreZoomFromPreviousState(state: Bundle) {
+    private fun initializeViewport() {
+        previousState?.let {
+            if (restoreZoomFromPreviousState(it)) {
+                viewportInitialized = true
+            }
+        }
+
+        if (!viewportInitialized) {
+            itemsByFeatureId.filter { it.value.selected }.map { it.key }.firstOrNull()?.let {
+                onFeatureClicked(it)
+                viewportInitialized = true
+            }
+        }
+
+        if (!viewportInitialized && zoomToFeatureBoundingBoxOnLoad && points.isNotEmpty()) {
+            map.zoomToBoundingBox(points, 0.8, false)
+            viewportInitialized = true
+        }
+
+        map.setGpsLocationListener { point -> onLocationChanged(point) }
+    }
+
+    private fun restoreZoomFromPreviousState(state: Bundle): Boolean {
         val mapCenter: MapPoint? = state.getParcelable(MAP_CENTER_KEY)
         val mapZoom = state.getDouble(MAP_ZOOM_KEY)
 
         if (mapCenter != null) {
             map.zoomToPoint(mapCenter, mapZoom, false)
-            viewportInitialized = true // avoid recentering as soon as location is received
+            return true
         }
+        return false
     }
 
     private fun setUpSummarySheet(binding: SelectionMapLayoutBinding) {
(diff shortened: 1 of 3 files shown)


geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -183,20 +184,15 @@ class SelectionMapFragment(
...
 
         map.setFeatureClickListener(::onFeatureClicked)
         map.setClickListener { onClick() }
 
         selectionMapData.getMappableItems().observe(viewLifecycleOwner) {
             updateItems(it)
             updateCounts(binding)
-        }
 
-        selectedFeatureViewModel.getSelectedFeatureId()?.let {
```
lognaturel, line 198: I am pretty confident this is redundant. `updateItems` used to do this work. Now `initializeViewport` does.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -183,20 +184,15 @@ class SelectionMapFragment(
...
-            onFeatureClicked(it)
+            initializeViewport()
```
lognaturel, line 195: I was having trouble keeping track of all the different ways the viewport could be set at launch time. It makes sense to me to group all that together.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -209,14 +205,37 @@ class SelectionMapFragment(
...
+    }
+
+    private fun restoreZoomFromPreviousState(state: Bundle): Boolean {
         val mapCenter: MapPoint? = state.getParcelable(MAP_CENTER_KEY)
         val mapZoom = state.getDouble(MAP_ZOOM_KEY)
 
         if (mapCenter != null) {
             map.zoomToPoint(mapCenter, mapZoom, false)
-            viewportInitialized = true // avoid recentering as soon as location is received
+            return true
```
lognaturel, line 236: It made me uneasy to have this state change buried in here. Indicating that initialization is complete felt like it should go with all the different ways the viewport could be initialized.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -209,14 +205,37 @@ class SelectionMapFragment(
...
+                viewportInitialized = true
+            }
+        }
+
+        if (!viewportInitialized && zoomToFeatureBoundingBoxOnLoad && points.isNotEmpty()) {
+            map.zoomToBoundingBox(points, 0.8, false)
+            viewportInitialized = true
+        }
+
+        map.setGpsLocationListener { point -> onLocationChanged(point) }
```
lognaturel, line 227: `onLocationChanged` could set the gps location listener to null once the viewport is initialized, I think.

As I mentioned in https://github.com/getodk/collect/pull/5098#issuecomment-1097281775, maybe this whole strategy is problematic. An alternative would be to see if `map.getGpsLocation` is set and if so, center to that, otherwise fall back to the bounding box and forget about having a listener at all.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -209,14 +205,37 @@ class SelectionMapFragment(
...
+    private fun initializeViewport() {
+        previousState?.let {
+            if (restoreZoomFromPreviousState(it)) {
+                viewportInitialized = true
+            }
+        }
+
+        if (!viewportInitialized) {
+            itemsByFeatureId.filter { it.value.selected }.map { it.key }.firstOrNull()?.let {
+                onFeatureClicked(it)
```
lognaturel, line 217: I have some trouble understanding how the selected item is represented. It looks like if there's an item selected in the form, that item has its `selected` field set. Then this initial call on `onFeatureClicked` saves the id of that item in the viewmodel. And then the `selected` field on `MappableSelectItem` objects stays stale and is entirely ignored. This makes me uneasy. Can the viewmodel not be created with the selectedid and the `selected` property be removed on `MappableSelectItem`?

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -209,14 +205,37 @@ class SelectionMapFragment(
...
 
-    private fun restoreZoomFromPreviousState(state: Bundle) {
+    private fun initializeViewport() {
+        previousState?.let {
+            if (restoreZoomFromPreviousState(it)) {
+                viewportInitialized = true
+            }
+        }
+
+        if (!viewportInitialized) {
```
lognaturel, line 215: I added this thinking the intent was not to zoom to the selected feature if there was a viewport to restore. A warning, it would no work with https://github.com/getodk/collect/pull/5095/commits/0277ec3ef3e9a423bf8c32f4a46baa0848512947#diff-41e26ac0b5e19db6790e339c006bc7c1ce451d326733985f3309aa5f3f3d6d89L353 because then the selected feature would not get the bigger marker. Maybe this case should go first? The restore from saved state is on rotation, right?

lognaturel:
I think this is a distraction for now. I tried to do some more verification and the behavior seemed even less predictable. In particular, there seems to be a reason for the redundancy at https://github.com/getodk/collect/pull/5099#discussion_r848918072

PR #5100, opened 2022-04-13
Mapping performance improvements

grzesiek2010:
Work towards #5087 

I've tested the performance manually.

This pr contains three improvements:
1. Caching bitmaps (markers) in mapbox - this is what we already had for google so it seems like a nice improvement in mapbox too to avoid huge memory usage.
2. Fix for setting marker icons twice - before we would create every marker with a default icon and then set what made the process longer.
3. Fixed highlighting already selected items after rotating a device.

Displaying icons for items on a map might be affected so this is what should be tested (map of forms and choices). Nothing else should be affected.

No.

No.


maps/src/main/java/org/odk/collect/maps/MapFragment.java
```diff
@@ -123,7 +123,7 @@ void addTo(
      * By default, the marker icon will be R.drawable.ic_map_point.
```
seadowg: This comment needs updated (or have an overloaded version that uses the "default" described in the comment).

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -354,6 +350,9 @@ class SelectionMapFragment(
             itemsByFeatureId[featureId] = item
             points.add(point)
         }
+        selectedFeatureViewModel.getSelectedFeatureId()?.let {
```
seadowg: It looks like we're missing test coverage for this change (and the original placement of the code). I guess the code itself is for two things:

1. Keeping the currently selected item selected if the items change (not really a use case we currently have)
2. Keeping the currently selected item selected when the Fragment is recreated

I've added a test for **2** and verified that it forces us to move the code around by moving it back and seeing it fail.

In writing that test, I then realized there was a problem where if we started the fragment with a selection, selecting a new item and then selecting again would always result in the initial selection being displayed. I wrote a test for that case and then realized we don't need this code at all. That's all committed here.

seadowg (approved):

maps/src/main/java/org/odk/collect/maps/MapFragment.java
grzesiek2010: Fixed.

lognaturel:
@seadowg I am overjoyed to see you add PRs to milestones. 🎉😅

seadowg:
> I am overjoyed to see you add PRs to milestones. 🎉😅

Ah because it doesn't have an attached issue!

lognaturel:
> Ah because it doesn't have an attached issue!

I see, makes sense. At some point in time I remember @grzesiek2010 was removing an issue from the milestone and putting the PR in so there wasn't redundancy and the most "active thing" was visible in the milestone. That worked well for me though it is annoying to do by hand. Ah well, by the time we figure it out I'm sure GitHub will have a concept that makes both the issue and PR get visually liked to the milestone.

srujner:
**Tested with Success!**
Verified on Android 10 and 11

Verifies cases:
- View submissions on map;
- Google Maps, Mapbox, OSM, use offline layer;
- Regression checks on geowidgets, background location;
- Performance on Thousand points geojson widgets using Mapbox;
- Highlighting already selected items after rotating a device;
- Displaying icons for items on a map;
- Light and Dark mode;

kkrawczyk123:
Tested also on Androids: 5.1 and 8.1

I have also noticed another small screen rotation issue and reported in #5113

PR #5101, opened 2022-04-13
Make sure BottomSheetCallback is not called after onDestroy

### LINKED ISSUE #5084: Collect is crashing when use select one from map widget on Google Maps on Android 5.1

grzesiek2010:
Closes #5084

I've tested the fix manually.

The problem was that `BottomSheetCallback` was called after `onDestroy()`. The flow is:
1. When we confirm selecting an item by clicking the `Save` button then we also close the bottoms sheet: https://github.com/getodk/collect/blob/master/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt#L258
2. That triggers `BottomSheetCallback` which updates marker icon
3. If it happens after `onDestroy()` we have that crash

it should just fix the issue and have no side effects.

No.

No.

### DIFF
--- a/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
+++ b/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
@@ -54,6 +54,7 @@ class SelectionMapFragment(
 
     private lateinit var summarySheetBehavior: BottomSheetBehavior<*>
     private lateinit var summarySheet: SelectionSummarySheet
+    private lateinit var bottomSheetCallback: BottomSheetCallback
 
     private val itemsByFeatureId: MutableMap<Int, MappableSelectItem> = mutableMapOf()
 
@@ -153,6 +154,11 @@ class SelectionMapFragment(
         outState.putDouble(MAP_ZOOM_KEY, map.zoom)
     }
 
+    override fun onDestroy() {
+        summarySheetBehavior.removeBottomSheetCallback(bottomSheetCallback)
+        super.onDestroy()
+    }
+
     @SuppressLint("MissingPermission") // Permission handled in Constructor
     private fun initMap(newMapFragment: MapFragment, binding: SelectionMapLayoutBinding) {
         map = newMapFragment
@@ -235,7 +241,7 @@ class SelectionMapFragment(
             onBackPressedCallback
         )
 
-        summarySheetBehavior.addBottomSheetCallback(object : BottomSheetCallback() {
+        bottomSheetCallback = object : BottomSheetCallback() {
             override fun onStateChanged(bottomSheet: View, newState: Int) {
                 val selectedFeatureId = selectedFeatureViewModel.getSelectedFeatureId()
                 if (newState == BottomSheetBehavior.STATE_HIDDEN && selectedFeatureId != null) {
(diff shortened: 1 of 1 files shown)


seadowg:
Would an alternative here be to just remove https://github.com/getodk/collect/blob/master/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt#L258? We don't need to close the bottom sheet if we're leaving the page.

lognaturel (approved):

grzesiek2010:
> We don't need to close the bottom sheet if we're leaving the page.

I was thinking about it but I guess we need this in case of the map of forms then the fragment is not destroyed and when you click the button you open a new form and when you navigate back you don't want to see the same sheet still open. At least it was implemented like that. Is that a mistake?

seadowg:
>  I was thinking about it but I guess we need this in case of the map of forms then the fragment is not destroyed and when you click the button you open a new form and when you navigate back you don't want to see the same sheet still open.

Good point! Yeah, I think your solution is the best then.

seadowg (approved):

kkrawczyk123:
Tested with success!

Verified cases:
- scenarios from the issue
- select place, edit selection, edit submission
- Google Maps, Mapbox and OSM
- regression checks on Android 8.1

srujner:
Tested with success!

Verified with regression checks on Android 10 and 11

PR #5102, opened 2022-04-13
Fix `FormHierarchyPage` usage in tests

seadowg:
This adds missing `assertOnPage` calls to some actions that return `FormHierarchyPage` and updates tests that were incorrectly using those actions but getting away with it.

Ran Test Lab a few times.

Nothing much to discuss here. We need `assertOnPage` calls to check we're in the right place before doing anything else in a test.

Just test changes. This can be merged without QA.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
@@ -4,12 +4,15 @@ import androidx.annotation.NonNull;
 import androidx.test.espresso.contrib.RecyclerViewActions;
 
 import org.odk.collect.android.R;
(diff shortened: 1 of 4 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FormHierarchyPage.java
```diff
@@ -22,6 +25,12 @@ public FormHierarchyPage(String formName) {
     @NonNull
     @Override
     public FormHierarchyPage assertOnPage() {
+        // Make sure we wait for loading to finish
+        WaitFor.waitFor((Callable<Void>) () -> {
```
seadowg, line 29: This is duplicated from `FormEntryPage`. Good to have here as well as we know form loading can cause flakey interactions with Espresso.

lognaturel (approved):

PR #5105, opened 2022-04-14
Avoid calling permission listeners if Activity is finishing

seadowg:
Should prevent [this crash](https://console.firebase.google.com/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/32bfd02368a071b00cbc8c2e729c0dca?time=last-seven-days&versions=v2022.2.0%20(4418)&types=crash&sessionEventKey=6257F4300038000121BD5B5133923E95_1664850225155885106).

New tests.

I couldn't reproduce, but I can imagine there is a flow where the permission listener actions (`granted` or `denied`) can be called after the original Activity that kicked off the permission request has finished. This can result in crashes as the listener might try and interact with the Activity in some way that's now impossible (showing a dialog for instance). The simplest solution I could see is just to avoid calling the listener.

Good to check any of the major parts of the app where permissions are requested but since we don't have reproduction steps here, there isn't much to check.

### DIFF
--- a/permissions/src/main/java/org/odk/collect/permissions/PermissionsProvider.kt
+++ b/permissions/src/main/java/org/odk/collect/permissions/PermissionsProvider.kt
@@ -30,7 +30,11 @@ import timber.log.Timber
  * area so that classes don't have to deal with this responsibility; they just receive a callback
  * that tells them if they have been granted the permission they requested.
  */
-open class PermissionsProvider(private val permissionsChecker: PermissionsChecker) {
+open class PermissionsProvider(
+    private val permissionsChecker: PermissionsChecker,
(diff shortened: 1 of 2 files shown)


lognaturel (approved):

lognaturel:
@getodk/testers I think a small sanity check of the major paths that require permissions or any paths to permissions that you think might be interesting/unusual would have value but it looks quite low risk to me so no need for checks on multiple API levels.

kkrawczyk123:
**Tested with success!**

Verified cases:
- permission dialogs
- deny and allow 
- redirect to settings
- revoke permissions in device's settings
- contacts, location, camera, microphone, phone
- regression check on Android 5.1

srujner:
**Tested with success!**

PR #5110, opened 2022-04-19
Prevent crashes when geometry is invalid

### LINKED ISSUE #5075: Handle invalid geometries

seadowg:
Closes #5075 

New tests.

Comments inline.

Just checking that the issue is fixed should be enough here.

Any form with select one from map question (and geometry set up to be invalid).

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ErrorDialog.java
+++ b/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ErrorDialog.java
@@ -8,6 +8,8 @@ import android.os.Bundle;
 
 import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 
+import org.odk.collect.material.MaterialAlertDialogFragment;
+
 /*
 Copyright 2018 Theodoros Tyrovouzis
 
@@ -24,6 +26,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
  */
 
+/**
+ * @deprecated Should be replaced with {@link MaterialAlertDialogFragment}
+ */
+@Deprecated
 public class ErrorDialog extends DialogFragment {
     private static final String ARG_MESSAGE = "message";
 
--- a/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
@@ -45,13 +45,21 @@ class SelectOneFromMapDialogFragment : MaterialFullScreenDialogFragment(), Fragm
         super.onAttach(context)
         DaggerUtils.getComponent(context).inject(this)
 
+        val formIndex = requireArguments().getSerializable(ARG_FORM_INDEX) as FormIndex
+        val selectedIndex = requireArguments().getSerializable(ARG_SELECTED_INDEX) as Int?
+        val prompt = formEntryViewModel.getQuestionPrompt(formIndex)
+        val selectionMapData = SelectChoicesMapData(resources, scheduler, prompt, selectedIndex)
+
+        selectionMapData.hasInvalidGeometry().observe(this) {
+            if (it) {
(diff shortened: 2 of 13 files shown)


collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/ErrorDialog.java
```diff
@@ -24,6 +26,10 @@
 limitations under the License.
  */
 
+/**
+ * @deprecated Should be replaced with {@link MaterialAlertDialogFragment}
+ */
+@Deprecated
```
seadowg, line 32: I would have done the work to replace this instead of deprecating here, but wanted to avoid the extra risk given that this will be going into a hotfix.

collect_app/src/main/java/org/odk/collect/android/widgets/items/SelectOneFromMapDialogFragment.kt
```diff
@@ -103,43 +131,60 @@ internal class SelectChoicesMapData(
...
-                val longitude = geometry.split(" ")[1].toDouble()
+                val latitude: Double
+                val longitude: Double
+
+                try {
+                    latitude = geometry.split(" ")[0].toDouble()
+                    longitude = geometry.split(" ")[1].toDouble()
+
+                    if (latitude.absoluteValue > 90 || longitude.absoluteValue > 180) {
+                        return InvalidGeometry(prompt.getSelectChoiceText(selectChoice), geometry)
```
seadowg, line 181: I'm really not happy with this (and the below) early return but haven't really come up with anything better 🤔

collect_app/src/main/java/org/odk/collect/utilities/Result.java
```diff
@@ -5,7 +5,10 @@
 /**
  * Based on https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/ which is not
  * available to Java.
+ *
+ * @deprecated use {@link org.odk.collect.shared.result.Result} instead
  */
+@Deprecated
```
seadowg, line 11: Again, I would have done the work to replace this instead of deprecating here, but wanted to avoid the extra risk given that this will be going into a hotfix.

lognaturel (approved):

kkrawczyk123:
**Tested with success!**

Verified cases:
- Google, Mapbox, OSM (no differences in behavior)
- crashes not visible
- information about invalid geometry visible
- screen rotation, minimize app
- finalize, edit submission
- light/dark mode
- RTL language 

The only concern here is that issue mentions "OK" button on the displayed dialog and it has not been added. It's not a big deal as I am able to hide dialog after tapping outside of dialog, I just want to make sure that it was intentional as we don't really have any dialogs without buttons.

lognaturel:
> make sure that it was intentional as we don't really have any dialogs without buttons

Accidental but as you said, I don’t think it’s a big deal. I think we can go with it for now. 

Edit: @grzesiek2010 will add. I don’t think we’ll need additional verification!

srujner:
**Tested with success!**

PR #5114, opened 2022-04-20
Removed the custom splash screen

### LINKED ISSUE #254: Remove splash screen from ODK Collect

grzesiek2010:
Closes #254

I've played with the app to make sure it works without any crashes or bugs.

I removed the old custom splash screen and enabled the new one according to https://developer.android.com/guide/topics/ui/splash-screen. It should look like:

| Light mode  | Dark mode |
| Screenshot_1650922010 | Screenshot_1650922036 |

Please verify that everything works well after removing the custom splash screen. It's important to use different android versions and maybe different screen sizes.

No.

No.


collect_app/src/main/res/drawable/splash_screen_icon.xml
```diff
@@ -0,0 +1,314 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
grzesiek2010, line 1: @lognaturel I had to add a vector version of the icon to display it in the splash screen. I think it looks pretty good (similar to the original one) but let me know if you have a better one.

buildSrc/src/main/java/dependencies/Dependencies.kt
```diff
@@ -75,6 +75,7 @@ object Dependencies {
     const val slf4j_timber = "com.arcao:slf4j-timber:3.1@aar"
     const val emoji_java = "com.vdurmont:emoji-java:5.1.1"
     const val json_schema_validator = "com.networknt:json-schema-validator:1.0.66"
+    const val splashscreen = "androidx.core:core-splashscreen:1.0.0-beta02"
```
lognaturel, line 78: None of us love bringing in beta dependencies but I agree it's the way to go here.

lognaturel (approved):

collect_app/src/main/res/drawable/splash_screen_icon.xml
lognaturel: Works for me. I don't have a different one.

kkrawczyk123:
**Tested with success!**

Verified cases:
- app update when splash screen selected on the previous version (default and user's picture) 
- light/dark mode selected in device's settings
- screen rotation on new splash screen 
- minimize app when splash screen opened 
- import settings from old version when old splash screen is selected. 
- splash screen removed from user interface and access control/user settings

Collect icon is not visible on Android 5.1, only an empty page displays - I am not sure if we will want to fix it. Should I open a new issue? @grzesiek2010 @lognaturel

srujner:
**Tested with success!**

grzesiek2010:
> Collect icon is not visible on Android 5.1, only an empty page displays - I am not sure if we will want to fix it. Should I open a new issue?

I think we can ignore it:
> On API 21-22, The icon isn't displayed until the application starts, only the background is visible.

https://developer.android.com/reference/androidx/core/splashscreen/SplashScreen#themes

PR #5128, opened 2022-04-28
Fixed clearing features

### LINKED ISSUE #5127: Selected map layer is not displayed automatically in OpenStreetMaps
#### Problem description
Selected map layer is not displayed automatically in OpenStreetMap

1. Install ODK Collect and download the All Widgets form from the demo server
2. Add any map layer to layers directory on your device
3. In settings select OpenStreetMap and your layer 
4. Go to Fill Blank Form and click on the map button

#### Expected behavior
The selected layer should be displayed automatically.

grzesiek2010:
Closes #5127 

I've tested the fix manually.

My thoughts are in comments below so I'm not going to repeat them here.

Apart from verifying the fix itself please test geo widgets with OpenStreetMap (adding/removing points) to make sure everything works fine. Additionally please take a look at Leak Canary to check if there are maybe new leaks caused by maps.

The `All Widgets` form from the demo server.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -370,6 +370,7 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
         for (MapFeature feature : features.values()) {
             feature.dispose();
         }
+        map.invalidate();
         features.clear();
         nextFeatureId = 1;
     }
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -367,10 +367,8 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
 
     @Override
     public void clearFeatures() {
-        if (map != null) {  // during Robolectric tests, map will be null
-            for (MapFeature feature : features.values()) {
-                feature.dispose();
-            }
+        for (MapFeature feature : features.values()) {
+            feature.dispose();
         }
         features.clear();
         nextFeatureId = 1;
--- a/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
@@ -366,14 +366,6 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
     }
 
     @Override
-    public void removeFeature(int featureId) {
-        MapFeature feature = features.remove(featureId);
-        if (feature != null) {
-            feature.dispose();
-        }
-    }
-
-    @Override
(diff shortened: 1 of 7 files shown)


collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -381,10 +381,11 @@ public void removeFeature(int featureId) {
 
     @Override
     public void clearFeatures() {
-        map.getOverlays().clear();
```
grzesiek2010, line 384: I don't know why it was implemented in this way... clearing everything not just features like it is in google and mapbox and like I implemented it now.

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -175,12 +175,6 @@ public void onStop() {
         super.onStop();
     }
 
-    @Override
-    public void onDestroy() {
```
grzesiek2010, line 179: I had to remove that because it caused NPEs in OSM I don;t think it's needed, at least I'm not seeing any leaks.

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -381,10 +381,11 @@ public void removeFeature(int featureId) {
 
     @Override
     public void clearFeatures() {
-        map.getOverlays().clear();
-        addAttributionAndMapEventsOverlays();
-        map.getOverlays().add(myLocationOverlay);
-        map.invalidate();
+        if (map != null) {  // during Robolectric tests, map will be null
```
seadowg: Do we need this check here? We don't have any Robolectric tests for this class as far as I'm aware.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
grzesiek2010: Yes if we have tests that require checks like that it's probably only for Google maps so here it's not needed. Fixed.

seadowg (approved):

kkrawczyk123:
I have noticed a delay with showing/deleting single points from geopoint widget with map appearance when adding point by placement on map and when delete point using delete button. I was comparing with store version and I can see delays like 2-6s on both my devices. @grzesiek2010 Could you please take a look?

grzesiek2010:
> Could you please take a look?

ok I'm on it.

grzesiek2010:
Ok the problem has been fixed. Please continue testing.

srujner:
**Tested with Success!**
Verified on android 10 and 11

Verified Cases:
- #5127 is no longer reproducing;
- Regression checks on Geowidgets using OSM layer;
- Offline layers;
- Adding/Removing single points;
- Zero leaks at Canary Leaks app;

kkrawczyk123:
Verified also on Androids 5.1 and 8.1

PR #5130, opened 2022-04-29
Bump targetSdkVersion to API 31

### LINKED ISSUE #5129: Bump targetSdkVersion to API 31
#### Problem description
Every year we have to bump the `targetSdkVersion` to [Meet Google Play's target API level requirement](https://developer.android.com/google/play/requirements/target-sdk).
Theoretically we can keep releasing new versions of our app till November with the current `targetSdkVersion` which is 30 but I think this year we should do that earlier becasue:
- I remember that last year we procrastinated and the conclusion was that next time we should do that earlier
- It's better to do that at the begging of a release cycle to have time to catch bugs that might occur 
- It might be better to do that now when @kkrawczyk123 is still available and can help @srujner with this crazy annual case where anything might be broken

#### Expected behavior
The app should work well with `targetSdkVersion` 31 with no bugs/crashes/regressions

#### Things to do:
- [x] Bump `targetSdkVersion` to 31
- [x] Explicitly declare the [android:exported](https://developer.android.com/guide/topics/manifest/activity-element#exported) attribute for components that should be exported https://developer.android.com/about/versions/12/behavior-changes-12#exported
- [x] Specify pending intents mutability https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability
- [x] Double check toast text will fit https://developer.android.com/about/versions/12/behavior-changes-12#toast-redesign
- [x] Check behaviour of permissions and scheduled jobs (auto update, match exactly etc) when coming in and out of hibernation https://developer.android.com/topic/performance/app-hibernation#manually-invoke

@seadowg please when you are free also review the documentation https://developer.android.com/about/versions/12/behavior-changes-12 to make sure we do not miss anything obvious.

grzesiek2010:
Closes #5129 

I've reviewed the documentation https://developer.android.com/about/versions/12/behavior-changes-12#toast-redesign and implemented changes we identified in #5129 
I've also run tests on Android 12.

This is what we have to do every year as I discussed i the issue.

This might affect the app in any way. We have tried to identify as many features that might be broken as possible and fix it but this kind of change (bumping `targetSdkVersion` is tricky and it's not possible to list everything that should be tested).
Some ideas are:
- Toasts - they will contain the icon on Android 12 and they length will be restricted to two lines (see https://developer.android.com/about/versions/12/behavior-changes-12#toast-redesign)
- Testing the app on large screens and multi window mode (see https://developer.android.com/about/versions/12/behavior-changes-all#multi-window-mode)
- Camera preview on large screens (see https://developer.android.com/about/versions/12/behavior-changes-all#camera-preview)
- Microphone and camera indicators (see https://developer.android.com/about/versions/12/behavior-changes-all#multi-window-mode)

Generally it would be good for the QA team to read the whole doc:
- https://developer.android.com/about/versions/12/behavior-changes-all
- https://developer.android.com/about/versions/12/behavior-changes-12

to help you understand what kid of changes might occur while testing app on newer android versions (in this case android 12) and after bumping `targetSdkVersion`.

@seadowg if you have any other ideas please share.

No.

No.

### DIFF
--- a/location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
+++ b/location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
@@ -51,14 +51,12 @@ private class GnssStatusSatellitesLiveData(private val locationManager: Location
 
     private val gnssStatusCallback = object : GnssStatus.Callback() {
         override fun onSatelliteStatusChanged(status: GnssStatus) {
-            super.onSatelliteStatusChanged(status)
-            var usedInFix = 0
-            for (index in 0 until status.satelliteCount) {
-                if (status.usedInFix(index)) {
-                    usedInFix++
+            value = (0 until status.satelliteCount).fold(0) { count, index ->
+                when {
+                    status.usedInFix(index) -> count + 1
+                    else -> count
                 }
             }
-            value = usedInFix
         }
     }
 
--- a/location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
+++ b/location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
@@ -1,8 +1,11 @@
 package org.odk.collect.location
 
 import android.annotation.SuppressLint
+import android.location.GnssStatus
 import android.location.GpsStatus
 import android.location.LocationManager
+import android.os.Build
+import androidx.annotation.RequiresApi
 import org.odk.collect.androidshared.livedata.NonNullLiveData
 import org.odk.collect.location.satellites.SatelliteInfoClient
 
@@ -10,7 +13,13 @@ class GpsStatusSatelliteInfoClient(private val locationManager: LocationManager)
     SatelliteInfoClient {
 
     override val satellitesUsedInLastFix: NonNullLiveData<Int>
-        get() = GpsStatusSatellitesLiveData(locationManager)
+        get() {
+            return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+                GpsStatusSatellitesLiveData(locationManager)
+            } else {
+                GnssStatusSatellitesLiveData(locationManager)
+            }
+        }
 }
 
 @SuppressLint("MissingPermission")
@@ -33,5 +42,32 @@ private class GpsStatusSatellitesLiveData(private val locationManager: LocationM
             val usedInFix = status?.satellites?.filter { it.usedInFix() }?.count() ?: 0
             value = usedInFix
         }
+    }
+}
+
+@SuppressLint("MissingPermission")
+@RequiresApi(Build.VERSION_CODES.N)
+private class GnssStatusSatellitesLiveData(private val locationManager: LocationManager) :
+    NonNullLiveData<Int>(0) {
+
+    private val gnssStatusCallback = object : GnssStatus.Callback() {
+        override fun onSatelliteStatusChanged(status: GnssStatus) {
+            super.onSatelliteStatusChanged(status)
+            var usedInFix = 0
+            for (index in 0 until status.satelliteCount) {
+                if (status.usedInFix(index)) {
+                    usedInFix++
+                }
+            }
+            value = usedInFix
+        }
+    }
+
+    override fun onActive() {
+        locationManager.registerGnssStatusCallback(gnssStatusCallback, null)
+    }
+
+    override fun onInactive() {
+        locationManager.unregisterGnssStatusCallback(gnssStatusCallback)
     }
 }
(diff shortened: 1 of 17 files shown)


seadowg (approved):

srujner:
@grzesiek2010 ODK collect is crashing after clicking on the "Start GeoPoint" button for geopoint type with no appearance in All widgets. It is reproducible for all map sources only on Android 12. 
Here are some logs: 
crash1
crash2

grzesiek2010:
> ODK collect is crashing after clicking on the "Start GeoPoint" button for geopoint type

It should be fixed now. The problem was in observing the number of available satellites which we display in the dialog:
Screenshot_1652427379

I had to use a different provider to achieve that. Unfortunately I wasn't able to use it for all android versions because the new one is available on Android 7+ only. So for android 5-6 the old provider should be used and for android 7+ the new one which I've just implemented. You shouldn't see any differences during testing it's something that works under the hood but I'm just bringing it up to let you know that theoretically not only android 12 but all versions 7+ might be affected and should be tested to confirm that the number of satellites is displayed correctly.

PS this is the nature of bumping sdk versions. We knew the old provider was deprecated and we should replace it sooner or later but it wasn't mentioned anywhere that it will cause crashes on Android 12.

location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
```diff
@@ -35,3 +43,30 @@ private class GpsStatusSatellitesLiveData(private val locationManager: LocationM
...
 }
+
+@SuppressLint("MissingPermission")
+@RequiresApi(Build.VERSION_CODES.N)
+private class GnssStatusSatellitesLiveData(private val locationManager: LocationManager) :
+    NonNullLiveData<Int>(0) {
+
+    private val gnssStatusCallback = object : GnssStatus.Callback() {
+        override fun onSatelliteStatusChanged(status: GnssStatus) {
+            super.onSatelliteStatusChanged(status)
```
seadowg: I'm guessing you don't need this super call?

location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
```diff
@@ -35,3 +43,30 @@ private class GpsStatusSatellitesLiveData(private val locationManager: LocationM
...
+            var usedInFix = 0
```
seadowg: Not essential (and mostly subjective), but you can avoid needing any mutation here with a `fold`:

```kotlin
value = (0 until status.satelliteCount).fold(0) { count, index ->
    when {
        status.usedInFix(index) -> count + 1
       else -> count
    }
}
```

location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
grzesiek2010: Right

location/src/main/java/org/odk/collect/location/satellites/GpsStatusSatelliteInfoClient.kt
grzesiek2010: Done.

seadowg (approved):

srujner:
**Tested with Success!**
Verified on Android 10 and 12

Verified cases:

1. Maps:
- Select Map forms
- Offline Layers;
- Google Maps, Mapbox, OSM;
- View Submissions on map;
- Screen rotation, minimize app, using backward button,

2. Forms
- Sending finalized Forms, Editing saved Forms, Viewing sent Forms, Deleting saved Forms;
- All notifications: Tapping on notification, clicking on show details, exit from the error list;
- Match exactly and Failed match exactly;
- Automatic download and Failed automatic download,

3. Settings
- Adding project manually or through QR code;
- Switching through projects, adding duplicated projects;
- Invalid/Inverted QR code,

4. Camera
- Regression checks on Audio, Video and Image widgets;
- Using external app for audio recording, background recording;
- Using Front camera, selfie widget;
- Adding big pictures and using different image size;
- Checking all file formats
- Screen rotation, minimize app, using backward button,

5. Toast
- Displaying Toast messages;
- Displaying ODK Collect icon in toast messages;
- Displaying Toast messages longer than 2 lines;
- Toasts checked: save form, offline mode, form without instanceID, switch between projects, missing external app toast, invalid URL on fallback submission URL and server URL, invalid email address, experimental settings, set admin password, edit admin password, invalid admin password, disable admin password, scan invalid qr code, 

6. Other things
- Regression checks on All Widgets.xml;
- Exploratory testing;
- Permission dialogs;
- Project Management and Access control settings;
- Light and Dark theme.

kkrawczyk123:
Verified also on Androids: 5.1 and 8.1.

PR #5135, opened 2022-05-07
Add more info to `ExternalFilesUtils` crash

seadowg:
We've talked about doing this to get more insight into what might be causing this problem.

### DIFF
--- a/androidshared/src/main/java/org/odk/collect/androidshared/system/ExternalFilesUtils.kt
+++ b/androidshared/src/main/java/org/odk/collect/androidshared/system/ExternalFilesUtils.kt
@@ -14,7 +14,7 @@ object ExternalFilesUtils {
             throw IllegalStateException("External files dir is null!")
         } else {
             try {
-                val testFile = File(externalFilesDir.toString() + File.separator + ".test")
+                val testFile = File(externalFilesDir, ".test")
                 testFile.createNewFile()
                 testFile.delete()
             } catch (e: IOException) {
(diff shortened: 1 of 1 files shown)


androidshared/src/main/java/org/odk/collect/androidshared/system/ExternalFilesUtils.kt
```diff
@@ -8,13 +8,27 @@ object ExternalFilesUtils {
...
-            testFile.delete()
-        } catch (e: IOException) {
-            throw IllegalStateException("App can't write to storage!")
+        val externalFilesDir = context.getExternalFilesDir(null)
+
+        if (externalFilesDir == null) {
+            throw IllegalStateException("External files dir is null!")
+        } else {
+            try {
+                val testFile = File(externalFilesDir.toString() + File.separator + ".test")
```
grzesiek2010: You can simplify it and use:
`val testFile = File(externalFilesDir, ".test")`

androidshared/src/main/java/org/odk/collect/androidshared/system/ExternalFilesUtils.kt
```diff
@@ -8,13 +8,27 @@ object ExternalFilesUtils {
...
+                testFile.createNewFile()
+                testFile.delete()
+            } catch (e: IOException) {
+                if (!externalFilesDir.exists()) {
```
grzesiek2010, line 21: This you could check earlier without waiting for exception.

seadowg: This is probably paranoid, but I think I'd rather have the exception occur first, so we know we're in a crash scenario instead of deciding to crash ourselves. I'd of course expect the dir not existing to cause a crash, but I'm also worried that some magic might allow it to get created on the fly when written to on some devices.

grzesiek2010: ok

grzesiek2010 (approved):

Issue #5136, opened 2022-05-09
Redesign selection summary sheet


seadowg:
~~Blocked by #4943~~

The summary sheet should be redesigned to look like this when it's fully expanded:

<img width="352" alt="Screenshot 2022-06-03 at 10 27 16" src="https://user-images.githubusercontent.com/556280/171828028-43fc11b8-c0c2-47c6-bc3e-f298439d1bb7.png">

In addition:

* There should be a divider between each property 
* Clicking on an item on the map should open the bottom sheet ["collapsed"](https://developer.android.com/reference/com/google/android/material/bottomsheet/BottomSheetBehavior#STATE_COLLAPSED) and the peek height (the lowest part shown) should be:
    * the bottom of the only property for items with a single property
    * the middle of the second property for items with multiple properties

lognaturel:
Super small nit: I believe the horizontal rule is intended to be more subtle.

PR #5137, opened 2022-05-09
Move action/info above properties on selection map summary

seadowg:
Projects that have many properties run into problems with the "Save" button being hidden, so we're moving this around to fix that. I don't think it looks that great, but we've talked about doing a redesign for the next release, and I've created an issue for that (#5136).

Select one map:

Screenshot 2022-05-09 at 10 10 41

Form map:

Screenshot 2022-05-09 at 10 10 15

### DIFF
--- a/geo/src/main/res/layout/selection_summary_sheet_layout.xml
+++ b/geo/src/main/res/layout/selection_summary_sheet_layout.xml
@@ -37,6 +37,38 @@
         app:layout_constraintTop_toBottomOf="@id/guideline_top"
         tools:text="Name" />
 
+    <FrameLayout
+        android:id="@id/action_container"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


lognaturel:
Did you by any chance try grouping the horizontal rule with the title rather than putting it between the button and properties?

seadowg:
> Did you by any chance try grouping the horizontal rule with the title rather than putting it between the button and properties?

Yeah, I did. I preferred it the way it is. As I said, I'm not a fan of any of this really -  I think we need to rework the layout properly and with the long lists in mind. I think we can live with it for now though?

lognaturel (approved):

grzesiek2010 (approved):

kkrawczyk123:
**Tested with success!**

Verified cases:
- display points on map
- edit/view submission
- screen rotation when points selected 
- light mode and dark mode 
- form with select map

srujner:
**Tested with success!**
Verified also on Androids: 10 and 11

PR #5139, opened 2022-05-12
Remove `MapFragment` and `MapProvider` coupling

seadowg:
Work towards #4943 

While finishing off those clean-up tasks, I realized I needed to rework how `MapFragment` instances are managed on rotation (and other configuration changes) so I could move the code that handles retaining the map zoom/center. The best summary of  what and why is [here](https://github.com/getodk/collect/pull/5139#discussion_r871481103).

This also pulls `OsmDroidFragment` and `OsmDroidMapConfigurator` (as well as the OsmDroid SDK) out to a module to demonstrate that working.

Played around manually.

Comments inline.

Shouldn't change anything, but it would be good to verify anywhere in the app that uses maps with all the different engines.


collect_app/src/main/java/org/odk/collect/android/geo/MapProvider.java
```diff
@@ -47,7 +44,7 @@
  */
 @Deprecated
 @Singleton
-public class MapProvider implements MapFragmentFactory {
```
seadowg, line 50: Going down this path, we can probably get rid of the `MapProvider` entirely (or just rename a gutted version of it). After these changes, it's only being used to provide the list of different mapping engines (and their options) to the prefs UI.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -160,11 +166,21 @@ public class GoogleMapFragment extends SupportMapFragment implements
...
+                new GoogleMapTypeOption(GoogleMap.MAP_TYPE_HYBRID, R.string.hybrid),
+                new GoogleMapTypeOption(GoogleMap.MAP_TYPE_SATELLITE, R.string.satellite)
+        );
+
+        mapFragmentDelegate = new MapFragmentDelegate(configurator, settingsProvider.getUnprotectedSettings(), this::onConfigChanged);
     }
 
     @Override public void onStart() {
         super.onStart();
-        mapProvider.onMapFragmentStart(this);
```
seadowg, line 167: This is the code I really needed to be rid of. Delegating the config and listener registration to the singleton `MapProvider` requires it to have created all the `MapFragment` instances, so it can grab them out of the two maps (`listenersByMap` and `configuratorsByMap`) later. This means that recreated Fragments (instantiated during rotation for instance) would not work properly as they would never get registered with the `MapProvider`. As far as I can see, other than that being the way it works, there's no reason for any of that to be singleton state and the `MapFragment` instances should be able to handle their own configuration and listener registration. 

I've also added a `MapFragmentDelegate` to cut down any duplication that was added by pulling the code out of `MapProvider` and into the `MapFragment` implementations.

collect_app/src/main/java/org/odk/collect/android/geo/MapProvider.java
```diff
@@ -87,17 +84,17 @@ private static SourceOption[] initOptions() {
...
-                    KEY_MAPBOX_MAP_STYLE, R.string.basemap_source_mapbox,
-                    new MapboxUrlOption(Style.MAPBOX_STREETS, R.string.streets),
-                    new MapboxUrlOption(Style.LIGHT, R.string.light),
-                    new MapboxUrlOption(Style.DARK, R.string.dark),
-                    new MapboxUrlOption(Style.SATELLITE, R.string.satellite),
-                    new MapboxUrlOption(Style.SATELLITE_STREETS, R.string.hybrid),
-                    new MapboxUrlOption(Style.OUTDOORS, R.string.outdoors)
-                )
-            ),
+//            new SourceOption(BASEMAP_SOURCE_MAPBOX, R.string.basemap_source_mapbox,
```
seadowg: Ignoring Mapbox for now while we work on the SDK upgrade for that.

seadowg:
@lognaturel @grzesiek2010 I'll leave this a draft until the new mapbox changes come in so that I don't create any conflicts, but it would be great to get feedback on if this looks OK.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -160,11 +166,21 @@ public class GoogleMapFragment extends SupportMapFragment implements
     @Override public void onAttach(@NonNull Context context) {
         super.onAttach(context);
         DaggerUtils.getComponent(context).inject(this);
+
+        GoogleMapConfigurator configurator = new GoogleMapConfigurator(
```
grzesiek2010, line 171: Could this be a separate class (object)?

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
```diff
@@ -149,12 +165,67 @@ public void addTo(
...
-        mapProvider.onMapFragmentStart(this);
+        mapFragmentDelegate.onStart();
+    }
+
+    @NonNull
+    private MapConfigurator createConfigurator() {
+        org.odk.collect.shared.settings.Settings settings = settingsProvider.getUnprotectedSettings();
+        String basemap = settings.getString(KEY_BASEMAP_SOURCE);
+
+        MapConfigurator cftor = new OsmDroidMapConfigurator(
```
grzesiek2010: The same here it could be placed in a separate class. Current map fragments are big enough.

grzesiek2010 (changes requested):
Generally it looks good to me. You can even go ahead and merge it before my changes because even if you add changes in Mapbox conflicts should not be big.

collect_app/src/main/java/org/odk/collect/android/geo/OsmDroidMapFragment.java
seadowg: Agreed. This all lived in `MapProvider` before. I'm thinking we should switch Mapbox to this approach and then work out what the relationship between the `MapFragment` and `MapConfigurator` should look like. What do you think?

collect_app/src/main/java/org/odk/collect/android/geo/MapProvider.java
lognaturel: I spent a little bit of time thinking about whether there's some hidden advantage to having this singleton. I think it was just a convenient structure to give a single entry point into the map world before we started breaking things up into modules.

I'm in favor of going a little further here at least with renaming and no longer being a singleton. Can you please also update/remove the header comment?

seadowg:
@grzesiek2010 after our conversation yesterday around #5141, I'll include Mapbox in this PR (doing the same things I did for Google and OSM) to allow it to be moved to its own module. I think it also makes sense to move one of the other map implementations to its own module as a demo of how I see that working, and then you can move Mapbox in #5141.

grzesiek2010:
> @grzesiek2010 after our conversation yesterday around https://github.com/getodk/collect/pull/5141, I'll include Mapbox in this PR (doing the same things I did for Google and OSM) to allow it to be moved to its own module. I think it also makes sense to move one of the other map implementations to its own module as a demo of how I see that working, and then you can move Mapbox in https://github.com/getodk/collect/pull/5141.

Cool.

osmdroid/src/main/java/org/odk/collect/osmdroid/OsmDroidMapConfigurator.java
```diff
@@ -63,15 +58,14 @@ class OsmDroidMapConfigurator implements MapConfigurator {
...
+                context, prefKey, prefTitle, labelIds, values, settings
             ));
         }
         return Collections.emptyList();
     }
 
     @Override public Collection<String> getPrefKeys() {
-        return prefKey.isEmpty() ? ImmutableSet.of(KEY_REFERENCE_LAYER) :
-            ImmutableSet.of(prefKey, KEY_REFERENCE_LAYER);
+        return prefKey.isEmpty() ? setOf(KEY_REFERENCE_LAYER) : setOf(prefKey, KEY_REFERENCE_LAYER);
```
seadowg, line 68: Didn't feel worth bringing Guava over to `osmdroid`

docs/STATE.md
```diff
@@ -45,3 +45,4 @@ and update this document as the code evolves.
 * Replacing Rx (and other async work such as `AsyncTask`) with `LiveData` + `Scheduler` abstraction
 * Gradually removing use of `CursorLoader` (all remaining uses are in `CursorLoaderFactory`)
 * Using AndroidX Test in new local tests and migrating other local tests as we touch them (from classic Robolectric)
+* Replacing uses of Google Guava with similar helpers in Kotlin Standard Library
```
seadowg, line 48: Really don't think we need to be pulling in Guava any longer

osmdroid/src/main/java/org/odk/collect/osmdroid/OsmDroidMapFragment.java
```diff
@@ -86,17 +87,25 @@ public class OsmDroidMapFragment extends Fragment implements MapFragment,
...
+    public static final String KEY_WEB_MAP_SERVICE = "WEB_MAP_SERVICE";
 
     @Inject
     ReferenceLayerRepository referenceLayerRepository;
 
     @Inject
     LocationClient locationClient;
 
+    @Inject
+    MapConfigurator mapConfigurator;
```
seadowg, line 97: This could use the constructor with `FragmentFactory` but I'll leave that change right now. I can do that later as part of #4943 - I already have a branch running for that that depends on this PR.

osmdroid/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    implementation project(':shared')
+    implementation project(':androidshared')
+    implementation project(':maps')
+    implementation project(':location')
+    implementation project(':settings')
+
+    /**
+     * There seems to a weird dependency resolution problem if we don't use `api` here.
+     */
+    api Dependencies.osmdroid
```
seadowg: ~~I suspect this is Dagger's fault but need to do some digging. @grzesiek2010 if you are blocked I think it's worth merging with this PR as-is (as I'm away for a couple of days) and I can look at this when I get back.~~

Looks like this could be some weirdness in `implements` on `OsmDroidMapFragment` exposing types that don't exist in `collect_app` (as it doesn't have OsmDroid SDK types in its classpath any more). I've never run into this before, but I guess the fix is just to move those implementations to anonymous or inner classes.

strings/src/main/res/values/strings.xml
```diff
@@ -470,9 +470,9 @@
     <string name="terrain">Terrain</string>
     <string name="hybrid">Hybrid</string>
     <string name="satellite">Satellite</string>
-    <string name="light">Light</string>
-    <string name="dark">Dark</string>
-    <string name="outdoors">Outdoors</string>
+    <string name="light" tools:ignore="UnusedResources">Light</string>
```
grzesiek2010: You can remove `tools:ignore="UnusedResources"` here and below.

osmdroid/src/main/java/org/odk/collect/osmdroid/OsmDroidInitializer.kt
```diff
@@ -0,0 +1,10 @@
+package org.odk.collect.osmdroid
+
+import org.osmdroid.config.Configuration
+
+class OsmDroidInitializer(private val userAgent: String) {
```
grzesiek2010: Does this need to be a class with a constructor? Maybe just object and then `initialize` would take `userAgent` as parameter.

grzesiek2010: Does this need to be a class with a constructor? Maybe just object and then `initialize` would take `userAgent` as parameter.

osmdroid/src/main/java/org/odk/collect/osmdroid/WebMapService.java
```diff
@@ -1,18 +1,15 @@
...
 import java.io.Serializable;
 
-import static org.odk.collect.strings.localization.LocalizedApplicationKt.getLocalizedString;
-
 /**
  * A serializable definition of a Web Map Service in terms of its URL structure
  * and the parameters for fetching tiles from it.
  */
-class WebMapService implements Serializable {
+public class WebMapService implements Serializable {
```
grzesiek2010, line 12: It's up to you but you can convert it to kotlin while you are here.

settings.gradle
```diff
@@ -25,3 +25,4 @@ include ':androidtest'
 include ':settings'
 include ':servicetest'
 include ':maps'
+include ':osmdroid'
```
grzesiek2010, line 28: I would consider converting some of those classes you have moved to this new module to kotlin.

grzesiek2010, line 28: I would consider converting some of those classes you have moved to this new module to kotlin.

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+    @JvmStatic
+    fun createListPref(
+        context: Context,
+        key: String,
+        title: String,
+        labelIds: IntArray,
+        values: Array<String>,
+        settings: Settings,
+    ): ListPreference {
+        val labels = arrayOfNulls<String>(labelIds.size)
```
grzesiek2010: `arrayOfNulls`? why?

grzesiek2010: `arrayOfNulls`? why?

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+        return pref
+    }
+
+    private fun ensurePrefHasValidValue(
+        key: String,
+        validValues: Array<String>,
+        settings: Settings,
+    ) {
+        val value = settings.getString(key)
+        if (Arrays.asList(*validValues).indexOf(value) < 0) {
```
grzesiek2010: Why not just `validValues.indexOf`? Or you can even transform the whole check to `!validValues.contains(value)`?

grzesiek2010: Why not just `validValues.indexOf`?

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+            if (validValues.size > 0) {
```
grzesiek2010: Could be `validValues.isNotEmpty()`

grzesiek2010: Could be `validValues.isNotEmpty()`

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+        context: Context,
+        key: String,
+        title: String,
+        labels: Array<String?>,
+        values: Array<String>,
+        settings: Settings,
+    ): ListPreference {
+        ensurePrefHasValidValue(key, values, settings)
+        val pref = ListPreference(context)
+        pref.key = key
```
grzesiek2010: I would use `apply` or `also` to avoid repeating `pref.`

grzesiek2010: I would use `apply` or `also` to avoid repeating `pref.`

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Gets an integer value from the shared preferences.  If the preference has
+     * a string value, attempts to convert it to an integer.  If the preference
+     * is not found or is not a valid integer, returns the defaultValue.
+     */
+    @JvmStatic
+    fun getInt(key: String?, defaultValue: Int, settings: Settings): Int {
+        val value: Any? = settings.getAll().get(key)
```
grzesiek2010: You can use indexing operator `settings.getAll()[key]`

grzesiek2010: You can use indexing operator `settings.getAll()[key]`

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+    ): ListPreference {
+        val labels = arrayOfNulls<String>(labelIds.size)
+        for (i in labels.indices) {
+            labels[i] = context.getString(labelIds[i])
+        }
+        return createListPref(context, key, title, labels, values, settings)
+    }
+
+    /**
+     * Gets an integer value from the shared preferences.  If the preference has
```
grzesiek2010: Please remove double spaces in this comment.

grzesiek2010: Please remove double spaces in this comment.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
```diff
@@ -322,4 +333,40 @@ public PermissionsChecker providesPermissionChecker(@NonNull Context context) {
 
         return geoDependencyComponent;
     }
+
+    @NonNull
+    @Override
+    public OsmDroidDependencyComponent getOsmDroidDependencyComponent() {
```
grzesiek2010, line 275: We have more and more builders like this one I think it's time to start keeping them in separate classes to avoid blowing this class?

grzesiek2010, line 275: We have more and more builders like this one I think it's time to start keeping them in separate classes to avoid blowing this class?

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapConfigurator.java
```diff
@@ -62,11 +63,7 @@ private boolean isGooglePlayServicesAvailable(Context context) {
         return new PlayServicesChecker().isGooglePlayServicesAvailable(context);
     }
 
-    @Override public MapFragment createMapFragment(Context context) {
-        return new GoogleMapFragment();
-    }
-
-    @Override public List<Preference> createPrefs(Context context) {
+    @Override public List<Preference> createPrefs(Context context, Settings settings) {
```
grzesiek2010, line 67: You can convert this class to kotlin.

grzesiek2010, line 67: You can convert this class to kotlin.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapConfigurator.java
```diff
@@ -75,8 +72,8 @@ private boolean isGooglePlayServicesAvailable(Context context) {
         }
         String prefTitle = context.getString(
             R.string.map_style_label, context.getString(sourceLabelId));
-        return Collections.singletonList(PrefUtils.createListPref(
-            context, prefKey, prefTitle, labelIds, values
+        return Collections.singletonList(org.odk.collect.androidshared.ui.PrefUtils.createListPref(
```
grzesiek2010: Please fix this import.

grzesiek2010: Please fix this import.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -160,11 +167,21 @@ public class GoogleMapFragment extends SupportMapFragment implements
...
```
grzesiek2010, line 171: We construct `GoogleMapConfigurator` in two places in the same way. Maybe those parameters should be a part of that class and we do not need to pass them?

grzesiek2010, line 171: We construct `GoogleMapConfigurator` in two places in the same way. Maybe those parameters should be a part of that class and we do not need to pass them?

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -188,12 +205,7 @@ public class GoogleMapFragment extends SupportMapFragment implements
     }
 
     @Override public void applyConfig(Bundle config) {
-        mapType = config.getInt(KEY_MAP_TYPE, GoogleMap.MAP_TYPE_NORMAL);
-        referenceLayerFile = MapFragmentReferenceLayerUtils.getReferenceLayerFile(config, referenceLayerRepository);
-        if (map != null) {
-            map.setMapType(mapType);
-            loadReferenceOverlay();
-        }
+        throw new UnsupportedOperationException();
```
grzesiek2010: Now you should be able to get rid of this method at all.

grzesiek2010: Now you should be able to get rid of this method at all.

collect_app/src/main/java/org/odk/collect/android/geo/MapConfiguratorProvider.java
```diff
@@ -11,43 +12,24 @@
...
-@Deprecated
-@Singleton
-public class MapProvider implements MapFragmentFactory {
+import org.odk.collect.osmdroid.OsmDroidMapConfigurator;
+import org.odk.collect.osmdroid.OsmDroidMapConfigurator.WmsOption;
+import org.odk.collect.android.injection.DaggerUtils;
+import org.odk.collect.maps.MapConfigurator;
+import org.odk.collect.osmdroid.WebMapService;
+
+public class MapConfiguratorProvider {
```
grzesiek2010, line 31: Also here I would think about converting it to kotlin.

grzesiek2010, line 31: Also here I would think about converting it to kotlin.

collect_app/src/main/java/org/odk/collect/android/geo/MapFragmentFactoryImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.odk.collect.settings.keys.ProjectKeys.BASEMAP_SOURCE_OSM
+import org.odk.collect.settings.keys.ProjectKeys.BASEMAP_SOURCE_STAMEN
+import org.odk.collect.settings.keys.ProjectKeys.BASEMAP_SOURCE_USGS
+import org.odk.collect.settings.keys.ProjectKeys.KEY_BASEMAP_SOURCE
+import org.odk.collect.shared.settings.Settings
+
+class MapFragmentFactoryImpl(private val settings: Settings) : MapFragmentFactory {
+
+    override fun createMapFragment(context: Context): MapFragment {
+        return if (isBasemapOSM(settings.getString(KEY_BASEMAP_SOURCE))) {
```
grzesiek2010: It could be `when`:
```
        return when {
            isBasemapOSM(settings.getString(KEY_BASEMAP_SOURCE)) -> OsmDroidMapFragment()
            settings.getString(KEY_BASEMAP_SOURCE) == BASEMAP_SOURCE_MAPBOX -> MapboxMapFragment()
            else -> GoogleMapFragment()
        }
```

grzesiek2010: It could be `when`:
```
        return when {
            isBasemapOSM(settings.getString(KEY_BASEMAP_SOURCE)) -> OsmDroidMapFragment()
            settings.getString(KEY_BASEMAP_SOURCE) == BASEMAP_SOURCE_MAPBOX -> MapboxMapFragment()
            else -> GoogleMapFragment()
        }
```

collect_app/src/main/java/org/odk/collect/android/geo/MapboxMapFragment.java
```diff
@@ -212,6 +217,18 @@ public class MapboxMapFragment extends org.odk.collect.android.geo.mapboxsdk.Map
     @Override public void onAttach(@NonNull Context context) {
         super.onAttach(context);
         DaggerUtils.getComponent(context).inject(this);
+
+        MapboxMapConfigurator configurator = new MapboxMapConfigurator(
```
grzesiek2010, line 221: The same here it is constructed in the same way in two places so maybe it should be moved to `MapboxMapConfigurator`.

grzesiek2010, line 221: The same here it is constructed in the same way in two places so maybe it should be moved to `MapboxMapConfigurator`.

maps/src/main/java/org/odk/collect/maps/MapFragmentDelegate.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.odk.collect.shared.settings.Settings
+import org.odk.collect.shared.settings.Settings.OnSettingChangeListener
+import java.util.function.Consumer
+
+class MapFragmentDelegate(
+    private val configurator: MapConfigurator,
+    private val settings: Settings,
+    private val onConfigChanged: Consumer<Bundle>,
+) :
+    OnSettingChangeListener {
```
grzesiek2010: We usually do not add a new line in places like here I'm surprised it's even ok according to ktlint.

grzesiek2010: We usually do not add a new line in places like here I'm surprised it's even ok according to ktlint.

osmdroid/src/main/java/org/odk/collect/osmdroid/OsmDroidMapFragment.java
```diff
@@ -82,21 +82,28 @@
...
+
+    /**
+     * @deprecated {@link MapConfigurator} should encapsulate integration with
+     * {@link org.odk.collect.shared.settings.Settings}.
+     */
+    @Deprecated
```
grzesiek2010: But it's needed in `MapFragmentDelegate` so should this be deprecated?

grzesiek2010: But it's needed in `MapFragmentDelegate` so should this be deprecated?

grzesiek2010 (changes requested):

osmdroid/src/main/java/org/odk/collect/osmdroid/WebMapService.java
seadowg: I'm going to hold off on Kotlin conversions for this PR as I have a big branch that's dependent on this (which will finish off #4943). Can do a sweep through of Kotlin conversions at the end (definitely keen for these classes to switch, so we can use the `internal` keyword) so tell me off if I forget!

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
seadowg: It's what Kotlin conversion spat out for [this line](https://github.com/getodk/collect/blob/03137bfd820fc4ec88baf1f963c6855a0fad968c/collect_app/src/main/java/org/odk/collect/android/preferences/PrefUtils.java#L22) which does make sense I guess. I'll clean this class up though, so it's less obviously just a straight Java -> Kotlin conversion.

collect_app/src/main/java/org/odk/collect/android/application/Collect.java
seadowg: Yeah, agreed. Will create concrete classes for them in separate files.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
seadowg: It would be replaced with an `@Inject MapConfigurator mapConfigurator` like in `OsmDroidMapFragment` but as in https://github.com/getodk/collect/pull/5139#discussion_r876203520, I'm can avoid that entirely and use `FragmentFactory` in my follow-up PR.

osmdroid/src/main/java/org/odk/collect/osmdroid/OsmDroidMapFragment.java
seadowg: I think, eventually, `MapConfigurator` could handle the `Settings` listener registration as well. I agree it might be a bit much to keep this deprecated in the code though - it's not an essential change.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/NotificationDrawer.kt
```diff
@@ -53,7 +53,13 @@ class NotificationDrawer {
         destination: D
     ): D {
         val device = waitForNotification(appName)
-        val actionElement = device.findObject(By.text(actionText)) ?: device.findObject(By.text(actionText.uppercase()))
+
+        var actionElement = device.findObject(By.text(actionText)) ?: device.findObject(By.text(actionText.uppercase()))
+        if (actionElement == null) {
```
seadowg, line 58: Started running into this on Test Lab: looks from the video like the problem was the notification being "minimized" so that the action buttons were hidden. @grzesiek2010 as far as I understand, there isn't a way to guarantee that the notifications always show their action buttons (other than upping the "priority" which I'm not sure if we should do). Do you think it's something we should dig into more?

strings/src/main/res/values/untranslated.xml
```diff
@@ -25,7 +25,7 @@ the specific language governing permissions and limitations under the License. -
 
     <!-- For consistency, we don't translate these brand/organization names. -->
     <string name="basemap_source_google" translatable="false">Google</string>
-    <string name="basemap_source_mapbox" translatable="false">Mapbox</string>
+    <string name="basemap_source_mapbox" translatable="false" tools:ignore="UnusedResources">Mapbox</string>
```
grzesiek2010: `tools:ignore="UnusedResources"` is redundant here too.

androidshared/src/main/java/org/odk/collect/androidshared/ui/PrefUtils.kt
```diff
@@ -0,0 +1,79 @@
...
+        labelIds: IntArray,
+        values: Array<String>,
+        settings: Settings,
+    ): ListPreference {
+        val labels: Array<String?> = labelIds.map { context.getString(it) }.toTypedArray()
+        return createListPref(context, key, title, labels, values, settings)
+    }
+
+    /**
+     * Gets a integer value from the shared preferences. If the preference has
```
grzesiek2010: It was `Gets an integer` and you changed it to `Gets a integer`. My English grammar is not perfect but from what I know it should be `an`?

grzesiek2010 (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/NotificationDrawer.kt
grzesiek2010: What device did you use? I think notifications are minimized in API 31 and we need to adjust our tests to run them using that API but on earlier versions everything should be fine I guess.

seadowg: Test Lab was using API 29. Maybe it's based on how many notifications are already on screen or something. Either way, I guess this gets us ready for the later API levels.

grzesiek2010 (approved):

srujner:
**Tested with Success!** 
Verified on Androids: 10,11,12

Verified Cases:
    • Regression checks for GeoPoint, GeoTrace, GeoShape widgets; 
    • Select Map forms;
    • Offline Layers; 
    • Google Maps, Mapbox, OSM; 
    • Collecting Location in background; 
    • View Submissions on map; 
    • Screen rotation, minimize app, using backward button,

Issue #4940, opened 2021-12-20
Add Hilt to the project


grzesiek2010:
#### Problem description
We use Dagger for dependency injection. Hilt is built on top of the Dagger and it is a recommended tool now and it is everywhere (documentation, tutorials, articles). The goal of Hilt is:
- To simplify Dagger-related infrastructure for Android apps.
- To create a standard set of components and scopes to ease setup, readability, and code sharing between apps.
- To provide an easy way to provision different bindings to various build types, such as testing, debug, or release.

Some time ago I started a discussion on slack about it but we haven't done anything. Personally I'm a fan of this tool although I don't have hand-on experience with it, it feels like it makes life easier.

We can migrate from Dagger to Hilt (see https://dagger.dev/hilt/migration-guide.html) but It would require a lot of refactoring so for us maybe a better option would be to add Hilt to the project and support both gradually switching to Hilt.

What do you think @seadowg

seadowg:
I was just thinking about this! We should definitely have a proper discussion about this as part of our next release cycle and do some investigation, but I wanted to reply before you're away for the holidays. In short, I think we should look at and evaluate a few different options such as:

* Keeping Dagger2
* Moving to Hilt
* Moving to a Kotlin DI tool like Kodein or Koin

Whatever we decide, I think we could make the switch in our non `collect_app` modules first (as they have independent setup) and then maybe leave Dagger2 in `collect_app` until the job becomes small enough or we feel we have the time to do a full conversion there. 

Cards on the table: I'm not a fan of Dagger and whenever I've looked at Hilt I've not really felt like it improves on the problems I have (code generation, confusing compile time errors, need for dependency injection mechanism to be aware of components, trend towards tying yourself into Android specifics etc). It could be that that's changed over time though. I've found success in happiness previously by using less powerful but simpler (IMO) alternatives in Android such as (the now defunct) RoboGuice and Kodein DI. As I said at the beginning of this, I reckon we can probably compare different routes and come up with a good plan for the future.

grzesiek2010:
I have never used Koin or Kodein so I can't tell anything about my personal experiences with them vs Dagger. I can agree that dagger and its setup was complex and I think Hilt makes it a bit easier maybe not enough for you @seadowg but there is an improvement I think.

I would vote for sticking to dagger (I mean Hilt of course) because:
- it's the most popular DI tool that was used in ~95% of courses I watched/read in last 2 years + it's from Google
[...]

Issue #3883, opened 2020-05-29
Points entered are reset when select/deselect layer for Mapbox GeoTrace and GeoShape widgets


kkrawczyk123:
#### Software and hardware versions 
Collect v1.27.x, Android v5.1, 7.1, 10 probably all, device used...

#### Problem description
When testing #3880 I have noticed an issue that turned out to be visible on store version as well.
Points entered are reset when select/deselect layer for Mapbox GeoTrace and GeoShape widgets, gif attached:
![ezgif com-video-to-gif (2)](https://user-images.githubusercontent.com/33342488/83241581-523db500-a19b-11ea-990a-aee6883c0b3a.gif)

#### Steps to reproduce the problem
1. Select Mapbox in Map Settings 
2. Add an offline layer and select it
3. Open GeoTrace/GeoShape widget
4. Input any method and start to record location
5. When some points are already entered change layer to none
Points are gone/ Points entered displayed as 0 in the bottom of the page but polyline/polygon is visible

#### Expected behavior
- Points should be visible
- number of entered points should be visible
- user should be able to delete/save after stopping recording
- the results should be exactly the same as for Google map

#### Other information 
- the same issue is visible when an offline layer is not selected in settings but then is selected after recording points (opposite scenario to described)
- the issue occurs only for Mapbox

lognaturel:
Duplicate of #3515. The fundamental issue is with the way that the order of layers is established and maintained. See more of the discussion at https://github.com/getodk/collect/pull/3457#issuecomment-575847471

lognaturel:
More details about how this affects users: https://forum.getodk.org/t/collect-location-marker-disappears-on-map-after-sleep/28960/3

marenab:
Asking around to see if we can get some Mapboxers to help with this!

lognaturel:
Thank you, @marenab!

zestyping:
Yes, I'm looking at this issue.  Thanks ODK-bot!

grzesiek2010:
This issue does not occur anymore so I'm closing.

PR #3118, opened 2019-05-29
Show a warning instead of crashing if Mapbox cannot be initialized.

zestyping:
Issues: Closes #3115
Scope: MapboxMapFragment, UserInterfacePreferences
Requested reviewers: @lognaturel 

#### User-visible changes

When the user tries to launch a map with the Mapbox SDK selected, on a device that does not support the included Mapbox binary (`armeabi-v7a`), the app no longer crashes (as it did without this change).  Instead, a warning message appears.

When the user selects "Mapbox SDK" in the settings, a warning message appears.  The "Mapbox SDK" option is not disabled or omitted; it just generates a warning.

#### Notes for reviewers

I didn't bother to disable or remove the "Mapbox SDK" option because we're about to rewrite the map settings screens very soon.

In a first attempt, I had the `initMapbox()` method simply call `Mapbox.getInstance()` each time and catch the exception.  However, I discovered that after it fails the first time, it can be called a second time and it will return a non-null, but invalid, `Mapbox` instance without throwing an exception; any attempt to instantiate a map then results in a crash.  Therefore, `initMapbox()` remembers the result of the first attempt and does not try twice.

#### Verification performed

I built and ran the APK on an emulated x86-based device.  I went to the settings screen and selected "Mapbox SDK", and confirmed that I saw the warning message.  I went to a GeoPoint question and tried to open the map, and confirmed that it didn't crash and I saw the warning message.

PR #3108, opened 2019-05-28
Zero-byte stubs for unsupported architectures

yanokwa:
Closes #3097

#### What has been done to verify that this works as intended?
* File size is 8.3 MB on Mac. It's smaller in the Play Store.
* On armeabi-v7a with no libmapbox-gl.so binaries, there is a crash when I try to load the map. 
* On armeabi-v7a with just the armeabi-v7a libmapbox-gl binary (this PR), it works just fine.
* Play Store shows that this binary works on 5 native platforms (see screenshot below)

#### Why is this the best possible solution? Were any other approaches considered?
This approach is different from Ping's suggestions in #3097 in that I create my own aar with zero byte native binaries, rather than try to remove native binaries while packaging.

This allows us to ship this update to all architectures through the Play Store, but we only really include support for armeabi-v7a (by excluding everything else). The file size increase is minimal because we use zero-byte .so files.

If we could disable the Mapbox option on non-armeabi-v7 instead of crashing, that'd be better.
```
public static boolean isSupportedAbi(String Abi) {
        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
            return Arrays.asList(Build.SUPPORTED_ABIS).contains(Abi.toLowerCase());
        } else {
            return Build.CPU_ABI.equalsIgnoreCase(Abi);
        }
    }
```

#### Other
How I made nbi-stubs.
```
rm -rf nbi-stubs*
mkdir -p nbi-stubs/jni/arm64-v8a/;
mkdir -p nbi-stubs/jni/armeabi-v7a/;
mkdir -p nbi-stubs/jni/armeabi/;
mkdir -p nbi-stubs/jni/x86/;
mkdir -p nbi-stubs/jni/x86_64/;
touch nbi-stubs/jni/arm64-v8a/stub.so;
touch nbi-stubs/jni/armeabi-v7a/stub.so;
touch nbi-stubs/jni/armeabi/stub.so;
touch nbi-stubs/jni/x86/stub.so;
touch nbi-stubs/jni/x86_64/stub.so;
cd nbi-stubs;
zip -r ../nbi-stubs-1.0.0.aar *
cd .. 
rm -rf nbi-stubs/;
```

Play Store screenshots

**v1.22.0-beta.1**
Screen Shot 2019-05-28 at 2 23 44 PM

**this PR**
Screen Shot 2019-05-28 at 2 23 38 PM

PR #5141, opened 2022-05-16
Mapbox v10

### LINKED ISSUE #4893: Update Mapbox SDK
The Mapbox SDK has made two major changes in the last few updates:

1. Since 9.3.0, it uses a pre-built binary that's licensed under new Terms of Service ([changelog here](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.3.0))
1. Also since 9.3.0, the SDK is hosted in Mapbox's own maven repo which requires authentication via a Mapbox account to download ([docs here](https://docs.mapbox.com/android/maps/guides/install/))

I'll make two assumptions here around our reaction to this: we need to continue to support Mapbox, and we don't want to require contributors (or anyone building Collect) to have their own Mapbox accounts. Given that, before we update, we'll need to:

- [x] Confirm it's legally safe to use new versions of the SDK in the released versions of our app
- [ ] (At least) Allow Collect's `debug` build type to work without the Mapbox SDK dependency
- [ ] Add a way to do development with Mapbox (a build type/product flavour/build flag etc)

grzesiek2010:
Closes #4893 
~~Blocked by #5139~~

Most of the discussion around this PR is in the issue or here so no need to write this up really.


grzesiek2010:
Ok @lognaturel @seadowg it's time to decide how to handle Mapbox 10. As we talked recently options are:
- sharing our tokens so that everyone can build the app including CircleCi
- storing Mapbox 10 in our local repository
- having a separate flavor

I was more in favor of one of two first options because it would be easier. Then if it doesn't work for us or there is any problem with licenses we can add a separate flavor.
Please think about it one more time and we can discuss it.

collect_app/build.gradle
```diff
@@ -177,13 +173,6 @@ android {
         pickFirst 'draftv4/schema'
         pickFirst 'draftv3/schema'
 
-        // To ensure that ODK Collect is installable on all devices, we don't use
-        // abiFilters to exclude any ABIs; but to keep the APK slim, we include
-        // the Mapbox native library only for 32-bit and 64-bit ARM devices and
-        // omit it for all X86 devices.
-        exclude 'lib/x86/libmapbox-gl.so'
-        exclude 'lib/x86_64/libmapbox-gl.so'
```
grzesiek2010, line 185: I think you are not going to like it. Mapbox 10 has libs with different naming. I could of course update it here and delete too but that causes crashes when the app is started so it looks like we no longer can do tricks like that. That increases the size of our apk of course...

collect_app/build.gradle
```diff
@@ -115,7 +111,7 @@ android {
             minifyEnabled(true)
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
             resValue("string", "GOOGLE_MAPS_API_KEY", googleMapsApiKey)
-            buildConfigField 'String', "MAPBOX_ACCESS_TOKEN", '"' + mapboxToken + '"'
+            resValue("string", "mapbox_access_token", mapboxAccessToken)
```
grzesiek2010, line 114: This works out of the box. We generate a string named `mapbox_access_token` with a valid access token and mapbox takes it automatically to setup everything that it needs.

seadowg:
@grzesiek2010 is the compilation error on CI expected? I'm guessing it's because there isn't a Mapbox API key configured to grab the SDK.

collect_app/build.gradle
seadowg: What crash are you seeing? I think we'll want to do as much as can to avoid including unused `x86` libs.

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.kt
```diff
@@ -108,7 +107,6 @@ class ApplicationInitializer(
                 MapView(context).onCreate(null)
             }
             Configuration.getInstance().userAgentValue = userAgentProvider.userAgent
-            MapboxUtils.initMapbox()
```
seadowg, line 111: We just don't need this any longer? Definitely nice if so!

seadowg (dismissed):
Are you able to roll back to Kotlin conversions here? It makes it hard to understand the actual diff. For instance, Git is unable to track the changes you had to make to `MapboxInitializationFragment` https://github.com/getodk/collect/pull/5141/commits/d85866155bd80945245ff9841ceee8f1e50d9486.

We should potentially start doing our Kotlin conversions as follow-up PRs as I'm finding it's making reviews (and reviewing my own changes) much harder.

seadowg:
> sharing our tokens so that everyone can build the app including CircleCi
> storing Mapbox 10 in our local repository

I think these feel roughly equivalent: both would work but might result in Mapbox suspending our account/revoking keys, and we'd have to resort to doing something else eventually. An alternative to this that still keeps the build simple would be to require CI (which could be leaked easily unless we stop running CI for PRs) and developer to all have their own Mapbox account/API keys and create their own local `secrets.properties`.

> having a separate flavor

I still think something along these lines is my preferred option. We can add a product flavour for when Mapbox is included and when it isn't and have the former include the real implementation of `MapboxMapFragment` (and other Mapbox classes) and have the latter include a stub one. 

As I've thought about it more however, I think there might be a "simpler" way to do achieve the same goals: move Mapbox code to a module (we want to do this anyway for all the different `MapFragment` implementations) and only include it when we have an access token for the SDK. As an example, in our dependencies for `collect_app` we'd have something like :

```groovy
if (mapboxAccessToken != null) {
    implementation project(':mapbox')
}
```

This would allow us to have a module that would only compile when `mapboxAccessToken` is present. In other words, it's OK if it doesn't compile because of a missing dependency. Then when we're creating the Fragment in `collect_app` we can just load the class dynamically rather than statically:

```java
try {
    return this.getClassLoader().loadClass("org.odk.collect.mapbox.MapboxMapFragment").newInstance();
} catch (ClassNotFoundException e) {
     return null;
} catch (IllegalAccessException e) {
     return null;
} catch (InstantiationException e) {
     return null;
}
```

If we do go down this road (which I'm advocating for), we probably want to look at merging #5139 before this as those changes will be needed to allow `MapboxMapFragment` to be pulled out to a separate module.

collect_app/build.gradle
lognaturel: Agreed with @seadowg that more details are needed here. What's the binary size without the filtering?

grzesiek2010: The error is:
```
Caused by: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/org.odk.collect.android-fpP4T_tgz8x_GU7w9kKqvA==/base.apk"],nativeLibraryDirectories=[/data/app/org.odk.collect.android-fpP4T_tgz8x_GU7w9kKqvA==/lib/x86_64, /data/app/org.odk.collect.android-fpP4T_tgz8x_GU7w9kKqvA==/base.apk!/lib/x86_64, /system/lib64, /system/product/lib64]]] couldn't find "libmapbox-maps.so"
        at java.lang.Runtime.loadLibrary0(Runtime.java:1067)
        at java.lang.Runtime.loadLibrary0(Runtime.java:1007)
        at java.lang.System.loadLibrary(System.java:1667)
        at com.mapbox.common.loader.MapboxLibraryLoader.load(MapboxLibraryLoader.kt:19)
        at com.mapbox.maps.loader.MapboxMapsInitializer.create(MapboxMapsInitializer.java:19)
        at com.mapbox.maps.loader.MapboxMapsInitializer.create(MapboxMapsInitializer.java:16)
        at androidx.startup.AppInitializer.doInitialize(AppInitializer.java:180)
        at androidx.startup.AppInitializer.discoverAndInitialize(AppInitializer.java:238) 
        at androidx.startup.AppInitializer.discoverAndInitialize(AppInitializer.java:206) 
        at androidx.startup.InitializationProvider.onCreate(InitializationProvider.java:45) 
        at android.content.ContentProvider.attachInfo(ContentProvider.java:2092) 
        at android.content.ContentProvider.attachInfo(ContentProvider.java:2066) 
        at android.app.ActivityThread.installProvider(ActivityThread.java:6983) 
        at android.app.ActivityThread.installContentProviders(ActivityThread.java:6528) 
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6445) 
        at android.app.ActivityThread.access$1300(ActivityThread.java:219) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859) 
```

it takes place immediately when the app is started 

Here is the size of those files:
image

grzesiek2010:
> is the compilation error on CI expected? I'm guessing it's because there isn't a Mapbox API key configured to grab the SDK.

exactly there are no `MAPBOX_DOWNLOADS_TOKEN` so the app can't be built

collect_app/src/main/java/org/odk/collect/android/application/initialization/ApplicationInitializer.kt
grzesiek2010: We don not initialize mapbox like that now. That class was also used to handle cases where a user tried to use mapbox on unsupported devices or a user didn't have a valid access token (then null was returned and we were able to handle that cases). Now as all devices are supported because I removed excluding libraries for X86 devices + without a valid downloads token you won't build the app at all. It's still possible to have a supported device + valid downloads token but wrong access token but it doesn't make sense to check such strange cases I think.

grzesiek2010:
> Are you able to roll back to Kotlin conversions here? It makes it hard to understand the actual diff. For instance, Git is unable to track the changes you had to make to MapboxInitializationFragment https://github.com/getodk/collect/commit/d85866155bd80945245ff9841ceee8f1e50d9486.

I'm not sorry. Generally I always implement changes and once everything is done I add conversion in a separate commit but not this time sorry my mistake. To help you a bit with that class what I changed was:
- I've removed onStart/onResume/onPause/onStop/onSaveInstanceState/onLowMemory/onDestroyView methods because now it's handled automatically https://docs.mapbox.com/android/maps/guides/migrate-to-v10/#simplified-lifecycle-management
- I've updated loading styles https://docs.mapbox.com/android/maps/guides/migrate-to-v10/#map-styles

and that's it.

collect_app/build.gradle
lognaturel: If you haven't already, could you please review https://github.com/getodk/collect/pull/3108 and https://github.com/getodk/collect/pull/3118 to see if there are any ideas there?

The error you shared looks like it would be from a x86_64 device. Was it? I would not expect an arm device to give that error.

seadowg: If this error is indeed only on architectures with missing libs (which it looks like and hopefully is), then that pushes us more towards not including Mapbox in all builds (using product flavours or my preferred conditional dependency method). That would allow us to continue to exclude x86 libs as we'd just use the "non mapbox" build types on emulators.

seadowg:
Just a note about something to keep in mind: Chromebooks/Chrome OS devices are pretty likely to have x86. That's a real world example of x86 devices that we didn't really consider the first time we removed Mapbox's x86 libs (and therefore x86 support). I imagine we can get away with not supporting Mapbox on Chromebooks, but we should make sure the release build can start OK on them.

collect_app/build.gradle
grzesiek2010: > If this error is indeed only on architectures with missing libs

Yes the crash occurs only on x86 devices (if I remove those libs).

grzesiek2010:
@seadowg 
I moved mapbox code to its own module.
I wanted to try load classes dynamically like you said above: https://github.com/getodk/collect/pull/5141#issuecomment-1127891995

The problem is that we use that new module in more places than only to create `MapboxMapFragment` like in your example.
I'm going to list those places so that it's clear:
- creating `MapboxMapFragment` in https://github.com/getodk/collect/pull/5141/files#diff-f424798f5410bd1996038d93031423c03974cb72ff81112f1a58be8a03c228f0R21 we should be ale to handle this in the way you presented above
- creating `MapboxMapConfigurator` in https://github.com/getodk/collect/pull/5141/files#diff-3a00df4aebd5600fae250c3807e97417704f10649504cc5f3cad20652bfc2a3fR61 the same as above
- using MapboxMapConfigurator.DEFAULT_MAP_STYLE in [Defaults](https://github.com/getodk/collect/pull/5141/files#diff-210f5b7fc183e9c0fae301ac06622f84d636afb6d2740b21fa5c4599beeb64a0R58) I can replace it with a raw string `mapbox://styles/mapbox/streets-v11` so that `mapbox` module is not used there 
- `MapBoxInitializationFragment` in [main_menu](https://github.com/getodk/collect/pull/5141/files#diff-4beabbc156b71a2271e4b0d6f53adeb9f490488ab884ca0ac56c2933c5d06547R29) this would need to load that fragment dynamically so that if the `mapbox` module is not included there are no errors. BTW I need to double check if we still need this trick.
- places where we use classes from the new module to setup dagger:
https://github.com/getodk/collect/pull/5141/files#diff-19e8ed4f040ab190989099eb37cd00c559b474b8e802b01260614bf9b628773bR12 https://github.com/getodk/collect/pull/5141/files#diff-a2f0be8cb9ce6618f33c478f6ae0c69fc5e6eefa5807833ffc3f1f324369c5b3R281 this might be a problem

what do you think?

seadowg:
> creating MapboxMapConfigurator in https://github.com/getodk/collect/pull/5141/files#diff-3a00df4aebd5600fae250c3807e97417704f10649504cc5f3cad20652bfc2a3fR61 the same as above

Yeah, I think we can just dynamically add the Mapbox source option if the configurator is present.

> using MapboxMapConfigurator.DEFAULT_MAP_STYLE in [Defaults](https://github.com/getodk/collect/pull/5141/files#diff-210f5b7fc183e9c0fae301ac06622f84d636afb6d2740b21fa5c4599beeb64a0R58) I can replace it with a raw string mapbox://styles/mapbox/streets-v11 so that mapbox module is not used there

👍

> MapBoxInitializationFragment in [main_menu](https://github.com/getodk/collect/pull/5141/files#diff-4beabbc156b71a2271e4b0d6f53adeb9f490488ab884ca0ac56c2933c5d06547R29) this would need to load that fragment dynamically so that if the mapbox module is not included there are no errors. BTW I need to double check if we still need this trick.

Yes that sounds right. You'll have to add the Fragment with a transaction rather than loading it in XML using `name`.

> places where we use classes from the new module to setup dagger:
https://github.com/getodk/collect/pull/5141/files#diff-19e8ed4f040ab190989099eb37cd00c559b474b8e802b01260614bf9b628773bR12 https://github.com/getodk/collect/pull/5141/files#diff-a2f0be8cb9ce6618f33c478f6ae0c69fc5e6eefa5807833ffc3f1f324369c5b3R281 this might be a problem

Ooooft yeah that's a problem. I think what we'll have to do there is give the dependency module a constructor with args, and then grab the constructor dynamically. So the module would look like this:

```kotlin
@Module
class MapboxDependencyModule(
    private val settingsProvider: SettingsProvider,
    private val networkStateProvider: NetworkStateProvider,
    private val referenceLayerRepository: ReferenceLayerRepository
) {
    @Provides
    fun providesSettingsProvider(): SettingsProvider {
        return settingsProvider
    }

    @Provides
    fun providesNetworkStateProvider(): NetworkStateProvider {
        return networkStateProvider
    }

    @Provides
    fun providesReferenceLayerRepository(): ReferenceLayerRepository {
        return referenceLayerRepository
    }
}
```

I'd been thinking that giving these modules constructors would be nice anyway - it prevents us from having to create `Collect...` subclasses, so it's just less code overall. You can construct that without a static reference like this in Java:

```java
this.getClassLoader().loadClass("org.odk.collect.mapbox.MapboxDependencyModule").getConstructors()[0].newInstance(
    settingsProvider,
    networkStateProvider,
    referenceLayerRepository
);
```

Given we're using reflection to load and construct classes in multiple places, it might be a good idea to have a helper object that hides all this work. I could imagine a `MapboxClassInstanceCreator` that has methods like `createDependencyModule` and `createMapboxInitializationFragment`.

grzesiek2010:
I've fixed all steps mentioned above apart from DI. As you know i will be off next two weeks so if this blocks you somehow feel free to take over.

What's left:
- [x] fixing DI
- [x] removing x86 libs
- [x] making sure that everything works together

.circleci/test_modules.txt
```diff
@@ -12,5 +12,4 @@ upgrade:testDebug
 permissions:testDebug
 settings:testDebug
 maps:testDebug
-mapbox:testDebug
```
seadowg: We'll need to be careful to run tests ourselves when making changes to `mapbox`

README.md
```diff
@@ -153,15 +153,15 @@ Certain functions in ODK Collect depend on cloud services that require API keys
 
 **Google Maps API**: When the "Google Maps SDK" option is selected in the "User interface" settings, ODK Collect uses the Google Maps API for displaying maps in the geospatial widgets (GeoPoint, GeoTrace, and GeoShape).  To enable this API:
   1. [Get a Google Maps API key](https://developers.google.com/maps/documentation/android-api/signup).  Note that this requires a credit card number, though the card will not be charged immediately; some free API usage is permitted.  You should carefully read the terms before providing a credit card number.
-  1. Edit or create `collect_app/secrets.properties` and set the `GOOGLE_MAPS_API_KEY` property to your API key.  You should end up with a line that looks like this:
```
seadowg, line 156: There's no need `secrets.properties` to be in `collect_app`. Seems more intuitive to me for it to live in the root.

shared/src/main/java/org/odk/collect/shared/injection/MultiClassProvider.kt
```diff
@@ -0,0 +1,25 @@
+package org.odk.collect.shared
```
seadowg: This is basically a 25 line mini DI framework. I think there are definitely better solutions than this but lets finish the discussion in #4940 before thinking about that.

shared/src/main/java/org/odk/collect/shared/injection/ObjectProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.function.Supplier
+
+/**
+ * Allows a shared singleton to act as a provider for dependencies by implementing
+ * [ObjectProviderHost] and returning a [ObjectProvider] from
+ * [ObjectProviderHost.getMultiClassProvider].
+ */
+
+interface ObjectProvider {
```
seadowg, line 11: This is a 35 line mini DI framework. I think there are definitely better solutions than this but let's finish the discussion in #4940 before thinking about that.

collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
```diff
@@ -96,10 +97,10 @@ public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float ve
                 // Looks for user swipes. If the user has swiped, move to the appropriate screen.
 
                 // For all screens a swipe is left/right of at least .25" and up/down of less than .25" OR left/right of > .5"
-                int xpixellimit = (int) (ScreenUtils.xdpi(odkView.getContext()) * .25);
```
seadowg: A few tests were failing here due to `odkView` being null when we swipe back from the end screen. I'm not sure how we've got away with `SwipeHandler` not dealing with `EmptyView` and `FormEndView` for so long.

collect_app/src/androidTest/java/org/odk/collect/android/support/pages/ListPreferenceDialog.java
```diff
@@ -17,7 +17,6 @@ public ListPreferenceDialog<T> assertOnPage() {
     }
 
     public T clickOption(int option) {
-        clickOnString(option);
-        return page.assertOnPage();
+        return clickOnButtonInDialog(option, page);
```
seadowg, line 20: There's an issue referenced in this method.

mapbox/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,25 @@
...
+    <application>
+
+        <!-- Disable Mapbox initializers to prevent them running on devices that don't support it -->
+        <provider
+            android:name="androidx.startup.InitializationProvider"
+            android:authorities="${applicationId}.androidx-startup"
+            android:exported="false"
+            tools:node="merge">
+
+            <meta-data
```
seadowg, line 15: I couldn't find these initializer classes referenced in the Mapbox docs. This meant I had to debug inside Android library code to work out what initializers were being run. @lognaturel are we able to send some feedback to Mapbox about this?

shared/src/main/java/org/odk/collect/shared/injection/ObjectProvider.kt
lognaturel: In the mean time, consider adding to the comment to state it was introduced to solve a specific problem with Mapbox and is not intended to be used elsewhere.

README.md
lognaturel: It will be annoying to have to remember to move it and I guarantee some people who maintain forks (maybe even me!) will be super confused and have a hard time finding this note. Does it really seem worth it? I don’t feel strongly but do want to make sure you’ve considered that this has some costs for some types of users.

lognaturel (dismissed):
I’m delighted to say that now that y’all have done the work, it doesn’t look as scary or complicated as I had feared. I thought I’d want to review this interactively but I’m comfortable with the understanding I got async.

README.md
seadowg: Yeah that's a good point. The main reason for moving it was that `secrets.gradle` needs to access it, and it felt more natural for it to live in the top instead of having to reach down into a project. I'll think of a way to flag it during build (so it's obvious that `secrets.properties` needs to move) or move it back.

shared/src/main/java/org/odk/collect/shared/injection/ObjectProvider.kt
seadowg: I'm still thinking that if the result of #4940 is "keep Dagger" than I'll just replace this with a shared module for the Dagger setup. I'll leave this right now until we've made some decisions there.

README.md
```diff
@@ -153,18 +153,21 @@ Certain functions in ODK Collect depend on cloud services that require API keys
...
+  1. Edit or create `secrets.properties` and set the `GOOGLE_MAPS_API_KEY` property to your API key.  You should end up with a line that looks like this:
     ```
     GOOGLE_MAPS_API_KEY=AIbzvW8e0ub...
     ```
 
 **Mapbox Maps SDK for Android**: When the "Mapbox SDK" option is selected in the "User interface" settings, ODK Collect uses the Mapbox SDK for displaying maps in the geospatial widgets (GeoPoint, GeoTrace, and GeoShape).  To enable this API:
   1. [Create a Mapbox account](https://www.mapbox.com/signup/).  Note that signing up with the "Pay-As-You-Go" plan does not require a credit card.  Mapbox provides free API usage up to the monthly thresholds documented at [https://www.mapbox.com/pricing](https://www.mapbox.com/pricing).  If your usage exceeds these thresholds, you will receive e-mail with instructions on how to add a credit card for payment; services will remain live until the end of the 30-day billing term, after which the account will be deactivated and will require a credit card to reactivate.
```
seadowg, line 162: It looks to me like creating Mapbox accounts with `@gmail.com` addresses requires a credit card for an ID check (but my own domains seem to work fine) and this blocked @kkrawczyk123 from being able to set up an account. I'm hoping `@soldevelo` emails will work fine, but if not we might want to create an ODK dev account @lognaturel (probably using `@getodk.org`). For now, I've sent @kkrawczyk123 a `secrets.properties` file that with tokens generated in my dev account.

collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import androidx.fragment.app.Fragment
+import org.odk.collect.maps.MapConfigurator
+import org.odk.collect.maps.MapFragment
+
+object MapboxClassInstanceCreator {
+
+    @JvmStatic
+    fun isMapboxAvailable(): Boolean {
+        return try {
```
grzesiek2010: This could just return `createMapboxMapFragment() != null`

collect_app/src/main/java/org/odk/collect/android/geo/MapConfiguratorProvider.java
```diff
@@ -153,6 +155,16 @@ public static int[] getLabelIds() {
...
     }
 
+    private static boolean isMapboxSupported() {
+        List<String> mapboxAbis = Arrays.asList(
+                "arm64-v8a",
+                "armeabi-v7a"
+        );
+
+        return MapboxClassInstanceCreator.isMapboxAvailable()
+                && stream(Build.SUPPORTED_ABIS).anyMatch(abi -> mapboxAbis.contains(abi));
```
grzesiek2010: You can use method reference here `anyMatch(mapboxAbis::contains)`

grzesiek2010 (commented):
I've reviewed changes that were added when I was on leave. It looks great only two comments but nothing important.

dbemke:
I've found a crash after following those steps:

1. Select MapBox as a source of maps
2. Open form geojson with 1000 points 
[geojson.xml.txt](https://github.com/getodk/collect/files/9002380/geojson.xml.txt)
[thousand-points.geojson.txt](https://github.com/getodk/collect/files/9010719/thousand-points.geojson.txt)


3. Click on Select place in select from geojson
4. While map is loading hit back button

Stack trace from Logcat:
`2022-06-28 15:57:39.761 32216-6979/? E/collect.androi: No implementation found for void com.mapbox.maps.assets.AssetManagerProvider.finalize() (tried Java_com_mapbox_maps_assets_AssetManagerProvider_finalize and Java_com_mapbox_maps_assets_AssetManagerProvider_finalize__)
2022-06-28 15:57:39.761 32216-6979/? E/System: Uncaught exception thrown by finalizer
2022-06-28 15:57:39.761 32216-6979/? E/System: java.lang.UnsatisfiedLinkError: No implementation found for void com.mapbox.maps.assets.AssetManagerProvider.finalize() (tried Java_com_mapbox_maps_assets_AssetManagerProvider_finalize and Java_com_mapbox_maps_assets_AssetManagerProvider_finalize__)
        at com.mapbox.maps.assets.AssetManagerProvider.finalize(Native Method)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:291)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:278)
        at java.lang.Daemons$Daemon.run(Daemons.java:139)
        at java.lang.Thread.run(Thread.java:923)
2022-06-28 15:57:39.762 32216-6979/? E/collect.androi: No implementation found for void com.mapbox.maps.assets.AssetManagerProvider.finalize() (tried Java_com_mapbox_maps_assets_AssetManagerProvider_finalize and Java_com_mapbox_maps_assets_AssetManagerProvider_finalize__)
2022-06-28 15:57:39.762 32216-6979/? E/System: Uncaught exception thrown by finalizer
2022-06-28 15:57:39.763 32216-6979/? E/System: java.lang.UnsatisfiedLinkError: No implementation found for void com.mapbox.maps.assets.AssetManagerProvider.finalize() (tried Java_com_mapbox_maps_assets_AssetManagerProvider_finalize and Java_com_mapbox_maps_assets_AssetManagerProvider_finalize__)
        at com.mapbox.maps.assets.AssetManagerProvider.finalize(Native Method)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:291)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:278)
        at java.lang.Daemons$Daemon.run(Daemons.java:139)
        at `java.lang.Thread.run(Thread.java:923)``

On Google Maps there is no crash.

srujner:
Bug with Submissions: 
- Issue is only reproducing on Mapbox

**Steps to Reproduce:**

1. Go to Project Settings -> Maps and Select Mapbox as Source;
2. Download all Widgets form with Geopoint widget;
3. Fill few submissions and add answer to Geopoint;
4. Have at least two submissions with different statuses;
5. Click on the map icon next to the All Widgets form;
6. Click on the first created Submission and then try to click on another one (Only first submission will be clickable)
Screenshot_2022-06-28-15-59-00-815_org odk collect android


**Expected Behavior:**
User should be able to click on all submission points

**Additional information:**
- Issue is only reproducing on Mapbox

dbemke:
Layers in Geoshape and Geotrace widget on Mapbox don’t change the view of the map ( while the map is opened)

**Problem description:**

When layers are applied while the map is opened on Geoshape or Geotrace widget the view of the map doesn’t change. After that clicking back button and next clicking Start Geoshape/Geotrace- the layer is applied. Applying layer Countries closes the form and the user is moved to main menu. 

**Steps to reproduce:**

1. Go to Project Settings -> Maps and Select Mapbox as Source; 
2. Download all Widgets form with Geo widgets; 
3. Go to Geotrace/Geoshape widget
4. Click Start GoeTrace/Geoshape
5. Click on layer button → change layer to e.g. Control Room

**Expected behaviour:**
While maps are opened on Geoshape/ Geotrace widget , changing layers changes the view of the map.

https://user-images.githubusercontent.com/56479916/176446388-44d38a0d-f1ab-478a-a4a1-b29de97da2b9.mp4


**Other information:**
Goepoint widget works well.

grzesiek2010:
@srujner the bug you have reported should be fixed.

@dbemke I can't reproduce the crash you have described. Could you try again with the newest changes?

grzesiek2010:
@dbemke I've fixed the problem with changing layers that was a tricky one. Good catch!

dbemke:
Mapbox with applied vector map layer in Geopoint, Geoshape, Geotrace closes the form.

**Problem description:**

When Mapbox is used as a source of maps and there is a vector layer added and applied, the form crashes. User is moved to main menu.

**Steps to reproduce:**

1. Choose Mapbox as as source of maps in settings
2. Download all Widgets form with Geo widgets
3. Go to Geopoint/Geotrace/ Geoshape widget
4. Click Start
5. Click on layer button
6. Change layer to vector map e.g. 
[countries.mbtiles.txt](https://github.com/getodk/collect/files/9018438/countries.mbtiles.txt)


**Expected behavior:**

Layer is applied. The view of the map is changed. User is able to continue filling in a form.

@grzesiek2010 is already taking a look at this

grzesiek2010:
@dbemke 

The problem with vector tiles should be fixed.

grzesiek2010:
@srujner @dbemke 

I think this pull request also fixes this issue https://github.com/getodk/collect/issues/3883 Please verify. If not just ignore it wasn't the main goal.

dbemke:
@grzesiek2010 
verified on Android 11 and 8. Also verified https://github.com/getodk/collect/issues/3883 . Works as expected.

dbemke:
Mapbox with offline layers applied when ODK Collect was build without Internet access show a black background instead of a view of a map

**Problem description:**

When ODK Collect is built without access to the Internet and layers are added in directory, Mapbox is used as a source of maps on Geoshape and Geotrace widget or Geopoint with maps appearance,
after choosing a layer there is a black background instead of a view of a map.

**Steps to reproduce:**

    1. Build ODK Collect without access to the Internet
    2. Turn on Internet to download all Widgets form 
    3. Turn off the Internet
    4. Add offline layers in layers directory
    5. Go to ODK Collect app
    6. Choose Mapbox as as source of maps in settings 
    7. Go to All widgets form
    8. Go to Geotrace/ Geoshape widget or Geopoint with maps appearance
    9. Click Start 
    10. Click on layer button 
    11. Choose offline layer
    
**Expected behavior:**

Layer is applied. The view of the map is properly changed.

grzesiek2010:
@dbemke 
The issue with using mapbox offline is now fixed.

seadowg (dismissed):

srujner:
**Problem description:**
Issue can be reproduced only on **Android 5.1** and only for **Mapbox**.

When Mapbox is used as a source of maps and there is a submission created and the user clicks on submission and rotates the screen, the map is zooming out to the maximum value (map of the whole world is visible)

**Steps to reproduce:**

1. Choose Mapbox as a source of maps in settings;
2. Download all Widgets form with Geo widgets;
3. Create at least one submission point;
4. Click on the map icon next to the All Widgets form;
5. Click on the created Submission point and rotate the screen;
6. The map will zoom out to the maximum value

**Expected behavior:**
Map should stay at the same zoom value after rotating the screen

srujner:
**Problem description:**
Issue can be reproduced on **all devices** and only for **Mapbox**.

When Mapbox is used as a source of maps and there is a submission created and the user clicks on submission and rotates the screen, the dialog with information about submission disappears.

**Steps to reproduce:**

1. Choose Mapbox as a source of maps in settings;
2. Download all Widgets form with Geo widgets;
3. Create at least one submission point;
4. Click on the map icon next to the All Widgets form;
5. Click on the created Submission point and rotate the screen;
6. The dialog with additional information about submission will disappear.

https://user-images.githubusercontent.com/85225519/176883807-080365c2-3c8c-404f-aaac-7181efb4e2be.mp4

**Expected behavior:**
The dialog should not disappear after rotating the screen

seadowg:
It feels like there's only a couple of minor problems left (assuming everything other than the last two issues posted by @srujner are fixed). I'm thinking we can go ahead and merge this to unblock other PRs and create issues for the remaining problems. Sound good @grzesiek2010? If so, @srujner could you create issues for your last couple of comments?

grzesiek2010:
I'm not sure I would rather give the QA team some time till Monday at least if there is nothing big then we can file separate issues for everything.

lognaturel:
Maybe y’all can check in when @seadowg starts his Monday? QA might have made good progress by then and it’d be great to unblock other map-related things in flight.

seadowg:
@grzesiek2010 just dismissed reviews so that I remember to have another look at changes. Just request a review when you're done.

srujner:
**Tested with Success!**

Verified cases: 
- Issue [#3883](https://github.com/getodk/collect/issues/3883) is no longer reproducing;
- Five bugs will be created as separate issues, all minor things;
- Regression checks on Forms: geojson, gps_accuracy_threshold, map internal-select, placement_map, TestingGeoPointWidgetWithValue, OSM_buildings;    
- Select Map forms;    
- Offline Layers; 
- Placement by tapping, manual mode and automatic mode
- Google Maps, Mapbox, OSM;         
- View Submissions on map; 
- Sending Submission to Central;
- Switching between maps Google/OSM/Mapbox
- Map view with displayed points    
- Screen rotation, 
- Minimize app, using backward button, 
- Light and Dark mode,

dbemke:
**Tested with Success!**
Verified on Androids: 8.1, 11

seadowg (changes requested):
It looks like b2cea654a7e338398ddd6ecc0f22a12362b21139 and 34089a4eaa54e8b4dab9cccf3a383b85adb58ced make changes to `SelectionMapFragment`, so I'm guessing the issues they address (which I think are https://github.com/getodk/collect/pull/5141#issuecomment-1172217970 and https://github.com/getodk/collect/pull/5141#issuecomment-1172234222 respectively) are not actually Mapbox specific, and I'd suspect are also existing issues rather than new ones caused by this PR (given that the code being changed isn't Mapbox code). Is that right @grzesiek2010?

If so, I think we should leave those commits out and create issues. Those changes will cause some nasty conflicts with #5164 and the issues being fixed may also have already been fixed by those PRs. Additionally, #5164 introduces a way to test interactions with the `BottomSheetBehavior` which would be needed to write a test for b2cea654a7e338398ddd6ecc0f22a12362b21139 which is currently has no coverage.

mapbox/src/main/java/org/odk/collect/mapbox/MapBoxInitializationFragment.kt
```diff
@@ -66,13 +66,17 @@ class MapBoxInitializationFragment : Fragment() {
     }
 
     private fun initializeMapbox(context: Context) {
-        // Auto initialization is disabled to stop x86 devices from crashing so we do this manually
-        context.getState().let {
-            if (!it.get(KEY_MAPBOX_INITIALIZED, false)) {
-                AppInitializer.getInstance(context)
-                    .initializeComponent(MapboxMapsInitializer::class.java)
-                it.set(KEY_MAPBOX_INITIALIZED, true)
+        try {
```
seadowg: I can't work out why this didn't need a `try-catch` when it was called in `MapboxMapFragment`?

grzesiek2010:
https://github.com/getodk/collect/commit/34089a4eaa54e8b4dab9cccf3a383b85adb58ced yes it was an existing issue. the problem was that if we received current location before loading items on the map then other events like zooming to those points were ignored. However it wasn't visible on master at least on my device because somehow the process was a little bit faster (I think the new map fragment loads slower).

https://github.com/getodk/collect/commit/b2cea654a7e338398ddd6ecc0f22a12362b21139 here again the code was broken but it worked well somehow and only now the bug is visible

So both issues existed but both of them were let's say hidden.

seadowg:
> https://github.com/getodk/collect/commit/34089a4eaa54e8b4dab9cccf3a383b85adb58ced yes it was an existing issue. the problem was that if we received current location before loading items on the map then other events like zooming to those points were ignored. However it wasn't visible on master at least on my device because somehow the process was a little bit faster (I think the new map fragment loads slower).
>
> https://github.com/getodk/collect/commit/b2cea654a7e338398ddd6ecc0f22a12362b21139 here again the code was broken but it worked well somehow and only now the bug is visible
>
> So both issues existed but both of them were let's say hidden.

Thanks for the clarification. That makes sense. Can we pull those commits out for now (to a branch so that we don't lose the solutions you came up with), and then we can create issues or @getodk/testers can verify as part of a pass on #5164? That PR reworks `SelectionMapFragment` and adds a lot of tests to `SelectionMapFragment` so a rebase on top of these commits is going to be much harder than fixing the issues (and being able to write tests for them) there instead.

grzesiek2010:
> Can we pull those commits out for now

Those are pretty serious issues to me so I wouldn't release v2022.3 without it but if you want to just exclude it to avoid conflicts and then fix soon - I'm ok with it and I can do that.

mapbox/src/main/java/org/odk/collect/mapbox/MapBoxInitializationFragment.kt
grzesiek2010: It wasn't needed because on a device that does not support mapbox choosing mapbox is not possible so that class `MapboxMapFragment` was never used. However `MapBoxInitializationFragment` was always used if there were keys provided (what casued crashes on emulators for example). I improved it in https://github.com/getodk/collect/pull/5141/commits/24af4510bc9f0c43419c47deb8503001ddb2a7ef and now `try-catch` is not needed indeed because we check both keys and supported abis.

grzesiek2010:
@seadowg ok I reverted those two commits and I'm going to file separate issues for them.

seadowg:
> Those are pretty serious issues to me so I wouldn't release v2022.3 without it but if you want to just exclude it to avoid conflicts and then fix soon - I'm ok with it and I can do that.

Thanks! Agreed, we need to fix before releasing.

seadowg (approved):

PR #5149, opened 2022-05-26
Fixed duplicate forms

### LINKED ISSUE #5147: Forms inserted multiple times in db with match exactly
Collect v2022.2.2

#### Problem description
We've sometimes had reports of users getting duplicate formid/version errors. We thought the issue was mostly related to the server so for v2022.2.2, @grzesiek2010 made sure that on download a formid/version match results in an update, not a duplication. 

I just got the error with v2022.2.2 and it now looks like some kind of race condition to me.

I can't reproduce but here's what I did:
- Fresh 2022.2.2 install with no projects configured
- Scanned in the QR code for Foo at https://test.getodk.cloud/#/projects/149/app-users (access to one small form, one bigger one)
- Opened Filled Blank Form (I don't remember doing this particularly quickly. I've tried tapping on it as quickly as I can and can't reproduce)
- Filled out the `username` form, on finalization, got the error
- Switched form update mode to "manual", looked at Delete Saved Form, saw each form twice
- Looked at forms.db and saw
  ```
  1|2021 WHO Verbal Autopsy instrument||va_who_2021|2021102801|b8cb32fd35df50b9a0fbe86b5df5706d|1653063781657|2021 WHO Verbal Autopsy instrument-media|2021 WHO Verbal Autopsy instrument.xml||||b8cb32fd35df50b9a0fbe86b5df5706d.formdef||||
  2|2021 WHO Verbal Autopsy instrument||va_who_2021|2021102801|b8cb32fd35df50b9a0fbe86b5df5706d|1653063781657|2021 WHO Verbal Autopsy instrument-media|2021 WHO Verbal Autopsy instrument.xml||||b8cb32fd35df50b9a0fbe86b5df5706d.formdef||||
  3|username||username||b4b827d9630d452a12a0e78434457807|1653063781748|username-media|username.xml||||b4b827d9630d452a12a0e78434457807.formdef||||
  4|username||username||b4b827d9630d452a12a0e78434457807|1653063781748|username-media|username.xml||||b4b827d9630d452a12a0e78434457807.formdef||||
  ```

That App User code had definitely never been scanned by that install.

#### Expected behavior
This should never be a possible state.

#### Other information 
It feels like there's some sort of race condition we should identify and fix. It also feels like there should be a uniqueness constraint at the db level on md5 hash.

grzesiek2010:
Closes #5147 

I've verified the fix manually and added some automated tests.

I've added two levels of protection:
- I've removed scheduling tasks for every single setting if they are loaded in bulk: https://github.com/getodk/collect/pull/5149/commits/d68ff68e1f64b04b6f64007372c5ec816063d958. Thanks to that we don't have multiple tasks that might mess with each other
- I've added `UNIQUE ON CONFLICT IGNORE` to `md5Hash` column in forms db: https://github.com/getodk/collect/pull/5149/commits/c34fa2ff7a95fbd2af2415a88a331e94831d6796. Thanks to that we will avoid problems like that in the future if they are caused by something else. This also has one extra advantage - upgrading the db works in the way that we rename the current table then create a new one and copy rows to that new one: https://github.com/getodk/collect/pull/5149/commits/c34fa2ff7a95fbd2af2415a88a331e94831d6796#diff-7f1231dc28a397f6a9b04fbc5fbb27e1dfbf302e8f289ca457fee49500b1110eR242. That means if we have users with duplicate forms now and they are blocked, when they install a new version of the app with this fix the duplicate rows will be ignored.

It might be difficult for the QA team to reproduce the issue so I would focus on testing for regressions here only.

No.

No.


settings/src/main/java/org/odk/collect/settings/importing/SettingsImporter.kt
```diff
@@ -67,12 +67,8 @@ internal class SettingsImporter(
         loadDefaults(generalSettings, generalDefaults)
         loadDefaults(adminSettings, adminDefaults)
 
-        for ((key, value) in generalSettings.getAll()) {
```
grzesiek2010, line 70: I think it was just a mistake that we implemented it in this way. If we load settings in bulk we just need to update propertyManager and reschedule tasks once.
But if I'm wrong and missing something let me know.

lognaturel:
The two improvements look great to me. I can't think of a reason to process settings changes one by one when doing a bulk update. That's all I have, over to you, @seadowg.

settings/src/main/java/org/odk/collect/settings/importing/SettingsImporter.kt
seadowg: Yeah, it doesn't make sense to me any more either. I think `SettingsChangeHandler` might have been introduced to fire after specific keys changed and never then reworked for "bulk" changes.

formstest/src/main/java/org/odk/collect/formstest/FormsRepositoryTest.java
```diff
@@ -140,6 +140,17 @@ public void restore_marksDeletedAsFalse() {
         assertThat(formsRepository.get(1L).isDeleted(), is(false));
     }
 
+    @Test
+    public void save_multipleFormsWithTheSameHashIgnoresDuplicates() {
+        FormsRepository formsRepository = buildSubject();
+        Form form = FormUtils.buildForm("id", "version", getFormFilesPath()).build();
+
+        formsRepository.save(form);
+        formsRepository.save(form);
```
seadowg: Should we return the existing form when we insert a duplicate? I'm not sure about the consequences of just returning `null` (which can happen still result in a passing test).

collect_app/src/test/java/org/odk/collect/android/database/FormDatabaseMigratorTest.java
```diff
@@ -86,7 +118,8 @@ public void onUpgrade_fromVersion9() {
             assertThat(cursor.getString(cursor.getColumnIndex(AUTO_SEND)), is(contentValues.getAsString(AUTO_SEND)));
             assertThat(cursor.getString(cursor.getColumnIndex(AUTO_DELETE)), is(contentValues.getAsString(AUTO_DELETE)));
             assertThat(cursor.getString(cursor.getColumnIndex(GEOMETRY_XPATH)), is(contentValues.getAsString(GEOMETRY_XPATH)));
-            assertThat(cursor.isNull(cursor.getColumnIndex(DELETED_DATE)), is(true));
+            assertThat(cursor.getInt(cursor.getColumnIndex(DELETED_DATE)), is(0));
+            assertThat(cursor.getColumnIndex("deleted"), is(-1));
```
seadowg, line 122: Just to confirm: this is because we got rid of the `deleted` column, but it existed in version 9?

grzesiek2010: Yes.

formstest/src/main/java/org/odk/collect/formstest/FormsRepositoryTest.java
grzesiek2010: Hmmm it's not possible to return the existing one. To do that I would need to manually check if a form with a given md5 hash exist and return it without inserting.
I can replace `ON CONFLICT IGNORE` with `ON CONFLICT REPLACE` then nulls won't be returned but always the new one. What do you think?

seadowg: Hmmm not sure how I feel about the `REPLACE`. Could we do a check for an existing form based on md5 hash and return (if it does exist) that if insertion fails? In the conflict case, it looks like we just get back `-1` from `DatabaseFormsRepository#insertForm` to signal that no row was inserted.

grzesiek2010: Ok I added this https://github.com/getodk/collect/pull/5149/commits/a5b48f8647b2df7f618503f65db529745b87e7eb. Please take a look.

seadowg:
@grzesiek2010 looks like this PR is having the same problem that I ran into in #5139 on Test Lab (https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/5879985788461267514/details?stepId=bs.a128a1f931404148&testCaseId=2). I'll cherry-pick that commit into a new PR we can just merge so that the fix isn't blocked by QA.

grzesiek2010:
> @grzesiek2010 looks like this PR is having the same problem that I ran into in #5139 on Test Lab (https://console.firebase.google.com/u/0/project/api-project-322300403941/testlab/histories/bh.f6f8331e5ae6e371/matrices/5879985788461267514/details?stepId=bs.a128a1f931404148&testCaseId=2). I'll cherry-pick that commit into a new PR we can just merge so that the fix isn't blocked by QA.

The problem with notifications? If so ok we can merge it in a separate pr but I've ran tests for this pr too and then everything was fine so that must be random.

seadowg:
>  The problem with notifications? If so ok we can merge it in a separate pr but I've ran tests for this pr too and then everything was fine so that must be random.

Yeah thinking the same. PR here: https://github.com/getodk/collect/pull/5151.

seadowg (approved):

srujner:
**Tested with Success!**
Verified on Androids: 5.1, 10,11 and 12

Verified Cases:
- Issue #5147 is no longer reproducing;
- Installing the fix on project with duplicated forms upgrading db correctly - Duplicated forms are removed and single ones can be send without any error;
- Sending finalized Forms, Editing saved Forms, Viewing sent Forms, Deleting saved Forms;
- Match exactly and Failed match exactly;
- Adding project through QR code;
- Switching between projects, adding duplicated projects; 

PR #5153, opened 2022-05-31
Remove `logFormEvent` and `logServerEvent`

seadowg:
These events no longer work as Firebase limits the total number of param values to 200, so we end up only ever seeing a max of 200 forms or 200 servers for each event. Removing this, so we don't accidentally add any more of these in.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -520,8 +520,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
     // Precondition: the instance directory must be ready so that the audit file can be created
     private void formControllerAvailable(@NonNull FormController formController) {
-        AnalyticsUtils.setForm(formController);
-
         menuDelegate.formLoaded(formController);
 
         identityPromptViewModel.formLoaded(formController);
@@ -822,7 +820,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
 
         switch (requestCode) {
             case RequestCodes.OSM_CAPTURE:
-                AnalyticsUtils.logFormEvent(OPEN_MAP_KIT_RESPONSE);
+                Analytics.log(OPEN_MAP_KIT_RESPONSE);
                 setWidgetData(intent.getStringExtra("OSM_FILE_NAME"));
                 break;
             case RequestCodes.EX_ARBITRARY_FILE_CHOOSER:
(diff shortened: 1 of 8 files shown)


lognaturel:
Can you please add a link to docs about this? What do you think the actual behavior is when that limit is exceeded? Is there something new going on?

I thought that the 200 limit was just for the bar graph. I don't see a change from what I thought the behavior was previously. My belief is that the 200 most popular values are graphed and that others are logged but not displayed. It's also my belief that the additional values are accessible from e.g. Big Query.

I use the form and server hashes to get a general sense of whether we're dealing with one single project/form doing something vs. identifying a trend that lots of projects/forms follow. The cap at 200 doesn't seem like a problem for that and the technique still feels valuable. In practice I typically look at the top several forms/projects and then at whether or not there are 200+ represented.

seadowg:
> Can you please add a link to docs about this? What do you think the actual behavior is when that limit is exceeded? Is there something new going on?

I believe it's the limit of indices for custom dimensions ([here](https://support.google.com/analytics/answer/2709828?hl=en&ref_topic=2709827#zippy=%2Cin-this-article). We've been seeing this for a few months so it hasn't changed "recently".

> My belief is that the 200 most popular values are graphed and that others are logged but not displayed. It's also my belief that the additional values are accessible from e.g. Big Query.

Yes, as far as I've seen you can only see the most frequent 200 values (or indices if my assumption above is correct) for a our "form" or "server" custom dimensions. I'm not sure if the others are logged and available to BigQuery. My reading of the docs around the limits on indices would suggest they are not.

> I use the form and server hashes to get a general sense of whether we're dealing with one single project/form doing something vs. identifying a trend that lots of projects/forms follow. 

That's good to know. My thinking in us not using the custom dimensions any more is that we can no longer (as of a few months ago) answer the question "how many forms/servers are using this?" but you're right that we can answer the question "are just a few forms/servers using this?". I'll close this given it seems you're still getting some value from the custom dimensions.

lognaturel:
> we can answer the question "are just a few forms/servers using this?"

This is definitely the most important question in my mind. If you're happy enough with that, we can carry on without digging further.

Like I said, not sure it's worth it for either of us to dig further but I'm really confused. The docs you link to have a bar at the top that says "This article is about Universal Analytics properties, which will [stop processing data](https://support.google.com/analytics/answer/11583528) on July 1, 2023 (October 1, 2023 for Analytics 360 properties). If you haven't already, [start using a Google Analytics 4 property](https://support.google.com/analytics/answer/10089681)". I don't think that's what we use. I also don't see "custom parameter" anywhere in that doc. I see custom parameters described [here](https://firebase.google.com/docs/analytics/events?platform=web) but no mention of the 200 limit.

seadowg:
> I also don't see "custom parameter" anyw[here](https://firebase.google.com/docs/analytics/events?platform=web) in that doc. I see custom parameters described here but no mention of the 200 limit.

It's confusing as in Firebase it's referenced as a Custom "dimension" not a custom "parameter" and then in those docs it states:

> Custom parameters: Custom parameters can be used as [dimensions or metrics](https://support.google.com/analytics/answer/10075209) in [Analytics reports](https://support.google.com/analytics/answer/9212670).

I think the Firebase front end is confusing things for me here 🤔. But yes, let's avoid thinking about it as it seems like we're getting what we need right now!

PR #5154, opened 2022-05-31
Protect the app from importing keys and values that are not supported according to the json scheme

### LINKED ISSUE #4939: Protect the app from importing preferences that are not specified in our json scheme
#### Problem description
https://github.com/getodk/collect/pull/4924 introduces json scheme to validate preferences we are trying to import. The problem is that it will still be possible to import preferences other than those specified in the scheme what might cause problems like the one with `allow_other_ways_of_editing_form` described in https://github.com/getodk/collect/pull/4924#discussion_r771280301
Additionally preferences that should not be imported like the one mentioned above but exist in our app and rely on other preferences that we can import should have a proper updated state after importing those prefs.

### LINKED ISSUE #5096: Handle unsupported settings that might exist in QR codes
Collect all versions

#### Problem description
Currently we are not protected from saving settings that we do not support if such settings exist in qr codes. An example would be:
```
{
   "general":{
      "password":"koboPassword",
      "server_url":"https:\/\/kc.humanitarianresponse.info",
      "protocol":"anything",
      "username":"testUser"
   },
   "admin":{
      
   }
}
```

According to our scheme (see: https://github.com/getodk/collect/blob/master/settings/src/main/resources/client-settings.schema.json#L24) options that we support for `protocol` are `odk_default` and `google_sheets`. 

Scan the attached qr code:
qrcodeWithUnsupportedSettings

#### Expected behavior
If there is any other option provided like `anything` in the example above the default one should be used.

Additionally it would be good to display a toast if unsupported options exist in a qr code.

#### Other information 
https://github.com/getodk/collect/issues/5089

### LINKED ISSUE #5176: It is possible to add a project without setting current project id
#### Problem description
It is possible to add a project without setting current project id what causes crashes. It happens when a QR code does not contain project details so they need to be generated based on URL but the URL is invalid.
**IMPORTANT**: We already check URLs so not all types of error cause the issue but for example a URL like: `https:\/\/demo.getodk.org` does

1. Try to scan this qr code: 
settings
2. The first attempt will result with a message that the qr code is invalid
3. The second attempt will result with the `Duplicate Project` dialog - Select the first option (adding a duplicate project)

The same is possible when loading a QR code from a file (also two attempts are needed).

#### Expected behavior
The project should be added properly with no crashes. Even though the URL is invalid it should be imported so that a user can fix it in the app.

grzesiek2010:
Closes #5096 
Closes #4939 
Closes #5176

I've tested the app manually and improved automated tests.

I've extended `JsonSchemaSettingsValidator` and added methods to verify if existing keys/values are supported.

Please verify the case form the issue https://github.com/getodk/collect/issues/5096

No.

No.

### DIFF
--- a/settings/build.gradle
+++ b/settings/build.gradle
@@ -51,7 +51,6 @@ dependencies {
     implementation project(':projects')
 
     implementation Dependencies.json_schema_validator
-    implementation Dependencies.commons_io
 
     testImplementation Dependencies.junit
     testImplementation Dependencies.hamcrest
--- a/settings/build.gradle
+++ b/settings/build.gradle
@@ -51,6 +51,7 @@ dependencies {
     implementation project(':projects')
 
     implementation Dependencies.json_schema_validator
+    implementation Dependencies.commons_io
 
     testImplementation Dependencies.junit
     testImplementation Dependencies.hamcrest
--- a/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
+++ b/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
@@ -27,6 +27,24 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
         }
     }
 
+    override fun isKeySupported(parentJsonObjectName: String, key: String): Boolean {
+        return try {
+            schemaProvider().use { schemaStream ->
+                val schemaJsonObject = JSONObject(
+                    IOUtils.toString(schemaStream, Charset.defaultCharset())
+                )
+
+                return schemaJsonObject
+                    .getJSONObject("properties")
+                    .getJSONObject(parentJsonObjectName)
+                    .getJSONObject("properties")
+                    .has(key)
+            }
+        } catch (e: JsonParseException) {
+            false
+        }
+    }
+
     override fun isValueSupported(parentJsonObjectName: String, key: String, value: Any): Boolean {
         return try {
             schemaProvider().use { schemaStream ->
--- a/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
+++ b/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.networknt.schema.JsonSchemaFactory
 import com.networknt.schema.SpecVersion
 import com.networknt.schema.ValidatorTypeCode
-import org.apache.commons.io.IOUtils
 import org.json.JSONObject
 import org.odk.collect.settings.importing.SettingsValidator
 import java.io.InputStream
@@ -31,7 +30,7 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
         return try {
             schemaProvider().use { schemaStream ->
                 val schemaJsonObject = JSONObject(
-                    IOUtils.toString(schemaStream, Charset.defaultCharset())
+                    schemaStream.readBytes().toString(Charset.defaultCharset())
                 )
 
                 return schemaJsonObject
@@ -49,7 +48,7 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
         return try {
             schemaProvider().use { schemaStream ->
                 val schemaJsonObject = JSONObject(
-                    IOUtils.toString(schemaStream, Charset.defaultCharset())
+                    schemaStream.readBytes().toString(Charset.defaultCharset())
                 )
 
                 val settingJsonObject = schemaJsonObject
--- a/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
+++ b/settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
@@ -5,8 +5,11 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.networknt.schema.JsonSchemaFactory
 import com.networknt.schema.SpecVersion
 import com.networknt.schema.ValidatorTypeCode
+import org.apache.commons.io.IOUtils
+import org.json.JSONObject
 import org.odk.collect.settings.importing.SettingsValidator
 import java.io.InputStream
+import java.nio.charset.Charset
 
 internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> InputStream) :
     SettingsValidator {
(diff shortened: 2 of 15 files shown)


seadowg:
@grzesiek2010 is this meant to close https://github.com/getodk/collect/issues/4939?

grzesiek2010:
> @grzesiek2010 is this meant to close https://github.com/getodk/collect/issues/4939?

Yes, I've just fixed the description.

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
```diff
@@ -23,4 +26,53 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
             false
         }
     }
+
+    override fun isKeySupported(parentJsonObjectName: String, key: String): Boolean {
+        return try {
+            schemaProvider().use { schemaStream ->
+                val schemaJsonObject = JSONObject(
+                    IOUtils.toString(schemaStream, Charset.defaultCharset())
```
seadowg: Kotlin has [extensions for `InputStream`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-input-stream/) that let you avoid depending on Commons here. In general, we should be able to replace all the Guava and Commons calls with Kotlin standard library.

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
```diff
@@ -23,4 +26,53 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
             false
         }
     }
+
+    override fun isKeySupported(parentJsonObjectName: String, key: String): Boolean {
+        return try {
+            schemaProvider().use { schemaStream ->
```
seadowg: We're now reading the `schemaProvider` `InputStream` in a whole bunch of times, so I think it makes sense to cache it. To my mind, the simplest thing to do is to have a `lazy` property for the schema as a `String` and another as a `JSONObject` (that relies on the first).

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
```diff
@@ -23,4 +26,53 @@ internal class JsonSchemaSettingsValidator(private val schemaProvider: () -> Inp
...
+                val settingJsonObject = schemaJsonObject
+                    .getJSONObject("properties")
+                    .getJSONObject(parentJsonObjectName)
+                    .getJSONObject("properties")
+                    .getJSONObject(key)
+
+                return if (settingJsonObject.has("enum")) {
+                    val supportedValues = settingJsonObject.getJSONArray("enum")
+
+                    for (i in 0 until supportedValues.length()) {
```
seadowg: This could use `supportedValues.contains` instead I think?

settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
```diff
@@ -6,43 +6,47 @@ import org.junit.Test
...
+                    }
                 """
             ),
             equalTo(true)
         )
     }
 
     @Test
-    fun `returns false when json is invalid based on schema`() {
+    fun `isValidreturns false when json is invalid based on schema`() {
```
seadowg: Missing a space here.

seadowg (changes requested):

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
grzesiek2010: `JSONArray` does not provide such a method.

settings/src/test/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidatorTest.kt
grzesiek2010: Fixed.

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
grzesiek2010: Done.

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
grzesiek2010: Done.

settings/src/main/java/org/odk/collect/settings/validation/JsonSchemaSettingsValidator.kt
seadowg: Ach, sorry. Of course.

seadowg (approved):

settings/build.gradle
```diff
@@ -51,6 +51,7 @@ dependencies {
     implementation project(':projects')
 
     implementation Dependencies.json_schema_validator
+    implementation Dependencies.commons_io
```
seadowg: Shouldn't need this any longer

grzesiek2010: Right, fixed.

dbemke:
**Tested with Success!**

Verified on Android 8.1, 11

**Verified cases:**
- https://github.com/getodk/collect/issues/5176  
https://github.com/getodk/collect/issues/4939   
https://github.com/getodk/collect/issues/5096  are no longer reproducing
- importing QR code from a file, Google Drive and Central
- exporting QR code via Google Drive and email
- scanning QR code in horizontal and vertical way
- resetting all settings
- reconfiguring settings with QR code
- reconfiguring settings with big sized image
- reconfiguring settings with QR code containing sensitive information- admin and server password on Aggregate and ODK
- scanning valid, invalid QR code and inverted QR code
- updating settings from older versions
- adding duplicate project
- switching to existing project

srujner:
**Tested with Success!**

Verified on Android 10 and 12

Today I found one strange issue related to Android 5.1 only: #5178

grzesiek2010:
@seadowg there are new changes (for https://github.com/getodk/collect/issues/5176) here so you might want to take a look.

In the issue @lognaturel asked if https://github.com/getodk/collect/issues/5176 is a fix for [this Crashlytics report](https://console.firebase.google.com/u/2/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8d842445557fcf3942b4e99ba2f89aec?time=last-ninety-days&types=crash&sessionEventKey=62C64C2A033400015E5E410BADE4B60A_1695905622417877922). I said that the issue was probably responsible for just a part of the reports.
Now I'm thinking more about those crashes and I think that it might be caused by similar problems with QR codes that we are not able to detect and our scheme validator also does not deal with.

I wanted to propose one additional improvements which is wrapping [this](https://github.com/getodk/collect/blob/master/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt#L31) into `try-catch` and catch anything there. My thinking is that the mentioned method consists of many steps that could potentialy fail (for example creating `JSONObjects`). 
If that method somehow throws an exception we end up with that problem with no project id. So although we do not now for sure what might throw such exceptions it seeps to be possible.

What do you think @seadowg @lognaturel?

seadowg:
> I wanted to propose one additional improvements which is wrapping [this](https://github.com/getodk/collect/blob/master/settings/src/main/java/org/odk/collect/settings/ODKAppSettingsImporter.kt#L31) into try-catch and catch anything there. My thinking is that the mentioned method consists of many steps that could potentialy fail (for example creating JSONObjects).
If that method somehow throws an exception we end up with that problem with no project id. So although we do not know for sure what might throw such exceptions it seeps to be possible.

This probably makes sense. I think we'd also want to clear the new shared preferences bucket in the `catch` as well. A potentially nicer alternative would be to wrap `ProjectCreator#createNewProject`'s implementation in a "catch all" `try-catch` (and perform project repo/shared prefs clean up in the `catch`). Then we could remove the `try-catch` at `QrCodeProjectCreatorDialog#createProjectOrError` and just show an error when `createNewProject` returns `false` (which simplifies our UI code).

grzesiek2010:
> A potentially nicer alternative would be to wrap ProjectCreator#createNewProject's implementation in a "catch all" try-catch (and perform project repo/shared prefs clean up in the catch). Then we could remove the try-catch at QrCodeProjectCreatorDialog#createProjectOrError and just show an error when createNewProject returns false (which simplifies our UI code).

I already removed `try-catch` from `QrCodeProjectCreatorDialog#createProjectOrError` in this pr.
I think we should rather use `try-catch` in `ODKAppSettingsImporter#fromJSON` because this methods is the only place where `SettingsImporter.fromJSON` (the risky one) is called but `ODKAppSettingsImporter#fromJSON` itself is called from different places so i think it should catch exceptions and return false.

Then in `ProjectCreator#createNewProject` we can delete project / clear prefs if the returned value is false.

What do you think?

seadowg:
@grzesiek2010 ah sorry I just realized I was looking at the current `master`. Yeah, that sounds good!

grzesiek2010:
@seadowg new changes have been added. Please review.

seadowg (approved):

PR #5156, opened 2022-06-01
Fixed the dialog displayed when permission is denied

### LINKED ISSUE #5027: Camera permission dialog does not display when deny permission in project settings
Collect v2022.1.1, Android v all with permissions device used...

#### Problem description
When testing #4924 I have noticed that the camera permission dialog does not display after denying camera permissions like it does when performing the same steps when trying to add new project. The issue is visible on store version. see gif:
ezgif com-gif-maker (20)

1. Revoke camera permission or do fresh install
2. Go to Project settings/Project management/Reconfigure with QR code
3. Deny camera permission 

#### Expected behavior
Camera permissions dialog with options OK and Go to settings should display 

#### Other information 
DIalog displays on v2021.3.2

grzesiek2010:
Closes #5027

I decided to bring back the implementation of the dialog to what we had before merging https://github.com/getodk/collect/pull/4912/files#diff-071d32cd124445d9a343f84e2184cb8fb2422dd62ce8674596ad8142b70d14f6L227
Then we refactored it moving it to a dialog fragment add calling `action.denied` listener without waiting for closing the dialog. This caused the bug not only in the part decided by the issue but in all places where we relay on `denied` callback.

Theoretically we could use dialog fragment and add support for calling `action.denied` after closing it but it would require much more work what in this case doesn't make sense especially now when we know that sooner or later we need to rework the way we deal with permissions.

Please verify that the issue no longer exists. Additionally please make sure that the activity (with two tabs) is displayed correctly after granting the permission mentioned in the issue.

No.

No.


seadowg (approved):

srujner:
**Tested with Success!**
Verified on Androids: 10, 11 and 12

Verified Cases: 
- Issue #5027 is no longer reproducing;
- Camera Permission dialogs are displayed correctly; 

kkrawczyk123:
Verified also on Androids 5.1 and 8.1

PR #5158, opened 2022-06-03
Allow starting apps by package name on newer devices

### LINKED ISSUE #5157: Can't start external app on android 11+
#### Problem description
Can't start external app on android 11+
The problem has been originally reported on the forum: https://forum.getodk.org/t/external-app-doesn-t-opening/37644

1. Install https://play.google.com/store/apps/details?id=com.peace.ArMeasure&hl=de&gl=US
2. Test the attached form on android 11+ [exApp.xlsx](https://github.com/getodk/collect/files/8823956/exApp.xlsx)

grzesiek2010:
Closes #5157 

I've verified manually that this works well.

The problem is that:
> When an app targets Android 11 (API level 30) or higher and queries for information about the other apps that are installed on a device, the system filters this information by default. The limited package visibility reduces the number of apps that appear to be installed on a device, from your app's perspective.

see: https://developer.android.com/training/package-visibility

To solve the issue we have a few options:
- [Specific package names](https://developer.android.com/training/package-visibility/declaring#package-name) it is not going to work in our case since we don't know what apps our users want to start
- [QUERY_ALL_PACKAGES](https://developer.android.com/training/package-visibility/declaring) permissions (not recommended) so I don't want to use it. It would even probably require having a special permission from Google Play to publish such an app.
- [Packages that match an intent filter signature](https://developer.android.com/training/package-visibility/declaring#intent-filter-signature) this is what I used to solve the problem and it seems like the best option here

Please just verify that the issue is fixed and that will be enough.

No.

### DIFF
--- a/collect_app/src/main/AndroidManifest.xml
+++ b/collect_app/src/main/AndroidManifest.xml
@@ -315,4 +315,10 @@ the specific language governing permissions and limitations under the License.
         </activity>
 
     </application>
+
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.MAIN" />
+        </intent>
+    </queries>
 </manifest>


seadowg (changes requested):
I'm a little worried there might a deeper problem at play here. Any `activity` with a `intent-filter` should be exported by default (there's a good write-up on this and the changes for Android 12 around that [here](https://medium.com/androiddevelopers/lets-be-explicit-about-our-intent-filters-c5dbe2dbdce0)). Given that, I think whatever the underlying problem is here, it will probably still be present for similar scenarios where `MAIN` is not being used as the `action` on the `intent-filter`. Is it possible we're performing a query that we don't need to on the available packages, or that in this case the app on the other side is doing something incorrect in its manifest? 

It might be useful to create some demo apps to reproduce #5157, so we're able to be explicit about exactly what causes it.

grzesiek2010:
> I'm a little worried there might a deeper problem at play here. Any activity with a intent-filter should be exported by default (t[here](https://medium.com/androiddevelopers/lets-be-explicit-about-our-intent-filters-c5dbe2dbdce0)'s a good write-up on this and the changes for Android 12 around that here). Given that, I think whatever the underlying problem is here, it will probably still be present for similar scenarios where MAIN is not being used as the action on the intent-filter. Is it possible we're performing a query that we don't need to on the available packages, or that in this case the app on the other side is doing something incorrect in its manifest?

I decompiled the app we used for testing and its launcher activity is exported:

```
        <activity 
             android:label="@string/app_name" 
             android:name="com.peace.ArMeasure.ArActivity" 
             android:exported="true" 
             android:screenOrientation="portrait" 
             android:configChanges="orientation|screenSize">
            <intent-filter>
                <action android:name="android.intent.action.MAIN"/>
                <category android:name="android.intent.category.LAUNCHER"/>
            </intent-filter>
        </activity>
```

The problem is that in Collect we allow starting such exported activities 
- by name (any exported activity if a unique name is set) like in ODK Counter https://github.com/grzesiek2010/counter/blob/master/app/src/main/AndroidManifest.xml#L19
- or by package name (then only a launcher activity is started)

That activity does not have any unique name (like most of launcher activities I guess) so we can't use the first approach.
And since in android 11+ it is not possible to use package name like we used to do that in the past because of https://developer.android.com/training/package-visibility we can't use the second approach too.

but with my fix the second approach works well.

seadowg:
> I decompiled the app we used for testing and its launcher activity is exported:

Ah I see. I'd thought `ArMeasure` was the Activity name (due to the camel casing), but it's actually a package name.

> The problem is that in Collect we allow starting such exported activities
>
> * by name (any exported activity if a unique name is set) like in ODK Counter https://github.com/grzesiek2010/counter/blob/master/app/src/main/AndroidManifest.xml#L19
> * or by package name (then only a launcher activity is started)

Thanks this really helps me understand. I hadn't realized we supported the latter - that explains the special case of adding `MAIN` as an `action` we query into our manifest.

seadowg (approved):

srujner:
**Tested with Success!**
Verified on Android 10, 11 and 12

Verified cases:
- Fix works as intended on devices with Android 11 and 12
- Checked The form attached to the issue.

PR #5161, opened 2022-06-05
Clean up augmentations to AndroidX test code

seadowg:
For this, I went through `androidtest`, `fragmenttest` and `servicetest` and then checked that all the components were still needed (and removed them if not). I also made sure that everything was linked to a relevant discussion issue justifying its existence. 

The very good news here is that it looks like we no longer need the hacky test theme for `DialogFragment` which massively reduces the custom helpers we need for testing those.


collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.kt
```diff
@@ -56,4 +58,10 @@ class AndroidShortcutsTest {
         assertThat(shortcutTargetIntent.action, equalTo(Intent.ACTION_EDIT))
         assertThat(shortcutTargetIntent.data, equalTo(FormsContract.getUri("DEMO", formId)))
     }
+
+    @Test
+    fun blah() {
```
grzesiek2010: What is this? :smile:

grzesiek2010 (changes requested):

collect_app/src/androidTest/java/org/odk/collect/android/feature/external/AndroidShortcutsTest.kt
seadowg: haha doh! That's me checking if something works as I expect and then forgetting to get rid of it.

grzesiek2010 (approved):

PR #5163, opened 2022-06-20
Fix use of `CountingTaskExecutorRule`

seadowg:
Based on updates (inspired by a comment) to my [SO question](https://stackoverflow.com/questions/61872139/how-to-switch-out-androidx-architecture-component-task-executor-for-espresso-tes/62059803#62059803). I also realized while updating the Idling Resource that we weren't adding the `CountingTaskExecutorRule` to our chain so actually weren't switching Android's task executor out. This could explain some flakes we've been saying.


lognaturel (approved):
Very interesting!

PR #5164, opened 2022-06-20
Selection map cleanup

seadowg:
Closes #4943
Closes #5171 
Closes #5172 
~~Blocked by #5141~~

To understand changes here, it'd be best to review the checkmarks in the issue and then run through the PR commit by commit.

#### What has been done to verify that this works as intended?

New tests and verified manually.

#### Why is this the best possible solution? Were any other approaches considered?

The commits are mostly pretty small and iterative, so hopefully everything is clear. I'll leave comments inline for anything that stands out to me.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

It'd be good to review the selection map views (Form Map and Select One From Map) and also all the map engines in general (sorry 😭).

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#ui-components-style-guidelines)

Issue #5113, opened 2022-04-19
Deselected point is highlighted after screen rotation


kkrawczyk123:
#### Software and hardware versions 
Collect v2022.2x, Android all, device used...

#### Problem description
When testing #5100 I have noticed another small issue connected to select map and screen rotation. The issue is also visible on master. After deselecting previously saved point and screen rotation it is still highlighted and zoomed.

#### Steps to reproduce the problem
1. Open form with select-map
2. Select one place
3. Tap on select place again - see zoom to that point
4. Tap anywhere on the map, so point will not be zoomed
5. Rotate screen - see that point is zoomed and highlighted again 

#### Expected behavior
The unselected point should not be zoomed and highlighted after screen rotation

#### Other information 
just used form with select-map

PR #5167, opened 2022-06-22
Fix back button on select one from map

### LINKED ISSUE #5165: Pressing back with item selected closed Select One From Map widget map
Collect v2022.2

1. Open a form with a Select One From Map widget
2. Go to the question and hit Select Place
3. Select an item
4. Press back

The user is returned to the question.

#### Expected behavior

The summary sheet should close instead of returning to the question.

#### Other information 

Fixing this involves using the new Fragment 1.5.0 library (so that `DialogFragment` uses `ComponentDialog` under the hood) which is currently still an RC and uses some unstable dependencies.

seadowg:
Closes #5165 
~~Blocked by #5164 and on a stabler release of Fragment 1.5.0~~

The problem stemmed from the fact that when displaying in a `Dialog`, the back button will be handled by that and not the current `Activity`. The new `OnBackPressedDispatcher` API provides a neat way to deal with this - `ComponentDialog` owns an `OnBackPressedDispatcher` that can be passed through to any child `Fragment` objects and used instead of the host `Activity` one.

The alternative would be to pass custom listeners around which I'm assuming no one will be in favour of.

Should just fix the issue. Good to play around with both selection maps (select one from map and form map) paying special attention to the back button behaviour.


seadowg:
Fragments 1.5.3 is the current stable release, so this is hopefully unblocked.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -218,15 +220,15 @@ class SelectionMapFragment(
...
 
-        val onBackPressedCallback = object : OnBackPressedCallback(false) {
+        val closeSummarySheet = object : OnBackPressedCallback(false) {
             override fun handleOnBackPressed() {
                 summarySheetBehavior.state = STATE_HIDDEN
             }
         }
 
-        requireActivity().onBackPressedDispatcher.addCallback(
+        (onBackPressedDispatcher ?: requireActivity().onBackPressedDispatcher).addCallback(
```
seadowg: To my mind, using the host `Activity` `OnBackPressedDispatcher` means we only need to test this is configured correctly when we override the value (like in `SelectOneFromMapDialogFragment`).

buildSrc/src/main/java/dependencies/Versions.kt
```diff
@@ -5,12 +5,13 @@ object Versions {
     const val android_min_sdk = 21
     const val android_target_sdk = 31
 
-    const val androidx_fragment = "1.4.1"
+    const val androidx_fragment = "1.5.3"
```
seadowg, line 8: This version switched the return value of `DialogFragment#getDialog` (and the underlying dialog type used to implement `DialogFragment`) from `Dialog` to `ComponentDialog` which we need to be able to use the dialog's `OnBackPressedDispatcher`.

grzesiek2010 (approved):

lognaturel:
Thank you, @grzesiek2010! 😊

dbemke:
#### Android version
10, 13
#### Device used
Redmi 9T, Pixel 6a
#### Problem description
After rotating the screen in Select one from map form, the app crashes or sometimes the form is closed. In Google maps, Mapbox , Osm.
1. Go to Select one from map form.
2. Tap "Select place".
3. Rotate the screen.
#### Expected behavior
The app shouldn't crash.
#### Other information 
On master 4925521801d471055ad8d0aa1a61918bc0a45901 and the store version the issue doesn't occur.

dbemke:
#### Android version
10, 13
#### Device used
Redmi 9T, Pixel 6a
#### Problem description
If don’t keep activities is on, the app crashes or the form is closed after locking the screen, minimizing the app(Android 13 minimizing + going to different app and back to Collect), rotating the screen
1. In device settings don’t keep activities is on.
2. Go to Select one from map form.
3. Tap "Select place".
4. Lock and unlock the screen or minimize the app and go back to the app.
#### Expected behavior
The app shouldn’t crash or the form shouldn’t close.
#### Other information 
The issue occurs also on store version(v2022.3.6) but it doesn’t occur on master 4925521801d471055ad8d0aa1a61918bc0a45901.
I rebased the PR onto the master and the issue is present.

seadowg:
@dbemke those problems should be fixed now

dbemke:
#### Android version
8.1, 10, 11, 12, 13 (probably all)

#### Problem description
When Mapbox is chosen as a source of maps, after selecting a point in the form map, rotating and clicking the device back button the form closes. The form should not close – the bottom sheet should collapse.
1. In Collect settings choose Mapbox as a source of maps.
2. Go to All widgets.
3. Go to Geopoint widget (with no appearance).
4. Add a geo point and save the form.
5. Go to the form map of All widgets.
6. Tap the point (bottom sheet opens).
7. Rotate the screen.
8. Click the device back button.

#### Expected behavior
The bottom sheet should close. The form shouldn’t close.

seadowg:
@dbemke and definitely not happening in Google Maps or OSM right? That's what I'm seeing, but wanted to double check as it's not making much sense to me so far.

dbemke:
I checked again on the demo project and a different project and it occurs only in Mapbox. @srujner Can you check on Android 13?

srujner:
@seadowg @dbemke I can confirm that this only happens in Mapbox on Android 13.

seadowg:
@srujner @dbemke I haven't been able to quickly find any reason for Mapbox having this problem. Let's continue testing, and then if everything else is ok we can merge this and create a new issue for the Mapbox/rotate specific problem. My thinking is that the changes here still fix the bug for 99% of scenarios. Sound good?

dbemke:
@seadowg What do you think should be the default zoom in this situation? Kyiv is the point selected at the beginning and then I select and deselect a different point and rotate the screen.

https://user-images.githubusercontent.com/56479916/198031967-d63e1583-26df-4ef8-a45d-9752d5151024.mp4

seadowg:
@dbemke I think we already have an issue for that: https://github.com/getodk/collect/issues/5113

dbemke:
**Tested with Success!**

Verified on Android 10

**Verified cases:**
- issue #5165 and issues described in the PR are no longer reproducing
- forms: Select one from map, All widgets, placement_map, geojson, Encrypted geopoint, map-internal-select, S1FromMap, map-internal-select-localized
- map view and form map
- adding, saving, editing geo points
- Google maps, Mapbox, OSM
- clicking back button when a point is selected/ deselected
- clicking back button when the bottom sheet is expanded
- clicking back button after all buttons on a map
- regression checks: display submissions on map and Geo widgets
- don’t keep activities enabled/disabled
- offline layers
- dark/light mode
- rotating, minimizing, locking the screen

srujner:
**Tested with Success!**

Verified on Android 13

Issue with mapbox has been reported separately: https://github.com/getodk/collect/issues/5326

PR #5181, opened 2022-07-08
Exclude newer version of commons-io used in javarosa

### LINKED ISSUE #5178: Collect crashes when trying to Configure project with QR code or Entering Get Blank Form on Android 5.1.

Collect v current master Android v 5.1, 

**Problem description**

When testing https://github.com/getodk/collect/pull/5154 I have noticed crash on Android 5.1 device only when trying to Configure project with QR. This issue occurs both on master and #5154 on store version everything works fine.  
It's a little bit strange issue, because yesterday I scanned qr codes multiple times both on master and #5154 branch and everything works well. The second possible crash is after clicking on the Get Blank form. 
Issues does not occurs on all other devices.

**Steps to reproduce the problem with QR code**

1. Install ODK application;
2. Click on the "Configure with QR code" button;
3. Scan any valid QR code.

Screenshot_20220707-151127

**Steps to reproduce the problem with Get Blank Form**
1. Install ODK application;
2. Click on the "Try a demo" button;
3. Click on the "Get Blank Form" button;
4. Connecting to server dialog will show and after few seconds the application will crash or user will be redirected to the Main menu.

**Expected behavior**

Collect should not crash or redirect user to the main menu.

**Other information**
When trying to import QR code from drive nothing happens and QR code is not imported. 
Logs from Android studio:

--------- beginning of crash
2022-07-07 14:16:14.461 14893-14893/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.odk.collect.android, PID: 14893
    java.lang.NoClassDefFoundError: j$.util.function.Supplier$Wrapper
        at [org.apache.commons.io](http://org.apache.commons.io/).IOUtils$$ExternalSyntheticAPIConversion0.m(R8$$SyntheticClass)
        at [org.apache.commons.io](http://org.apache.commons.io/).IOUtils.<clinit>(IOUtils.java:183)
        at [org.apache.commons.io](http://org.apache.commons.io/).output.AbstractByteArrayOutputStream.needNewBuffer(AbstractByteArrayOutputStream.java:104)
        at [org.apache.commons.io](http://org.apache.commons.io/).output.ByteArrayOutputStream.<init>(ByteArrayOutputStream.java:50)
        at org.odk.collect.android.utilities.CompressionUtils.decompress(CompressionUtils.java:80)
        at org.odk.collect.android.projects.QrCodeProjectCreatorDialog.startScanning$lambda-4(QrCodeProjectCreatorDialog.kt:242)
        at org.odk.collect.android.projects.QrCodeProjectCreatorDialog.$r8$lambda$ByUYamdCK4t6CK6Mnx3PLcICPnE(QrCodeProjectCreatorDialog.kt)
        at org.odk.collect.android.projects.QrCodeProjectCreatorDialog$$ExternalSyntheticLambda4.onChanged(R8$$SyntheticClass)
        at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:151)
        at androidx.lifecycle.LiveData.setValue(LiveData.java:309)
        at androidx.lifecycle.MutableLiveData.setValue(MutableLiveData.java:50)
        at org.odk.collect.android.views.BarcodeViewDecoder$1.barcodeResult(BarcodeViewDecoder.java:21)
        at com.journeyapps.barcodescanner.DecoratedBarcodeView$WrappedCallback.barcodeResult(DecoratedBarcodeView.java:50)
        at com.journeyapps.barcodescanner.BarcodeView$1.handleMessage(BarcodeView.java:52)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:706)

grzesiek2010:
Closes #5178 

The problem was that in Javarosa we merged https://github.com/getodk/javarosa/pull/688 which added a new dependency `commons-io:commons-io:2.11.0`. This dependency became automatically available in Collect and caused conflicts since the version we use is 2.5 and we can't use a newer one, see: https://github.com/getodk/collect/pull/4158

I've verified that the crash does not occur anymore on Android 5.

I could also downgrade `commons-io` used in Javarosa to the version used in Collect and then maybe even remove it from Collect but I think those two project should have separate dependencies.

This should just fix the issue on Android 5 and have no side effects.

No.

No.

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -314,6 +314,7 @@ dependencies {
         exclude group: 'commons-logging'
     }
     implementation(Dependencies.javarosa) {
+        exclude group: 'commons-io'
         exclude group: 'joda-time'
         exclude group: 'org.slf4j'
         exclude group: 'org.hamcrest', module: 'hamcrest-all'


lognaturel:
That’s annoying!! We have an Android version check just for this in JR. I’m surprised it didn’t catch this. 

Maybe we should use the same version in JR instead?

grzesiek2010:
> Maybe we should use the same version in JR instead?

I thought about this but to me my current solution is better. If javarosa can use the newest version I don't think we should downgrade it.

lognaturel:
I guess because JavaRosa doesn't make use of any part of the API that is out of bounds for Collect, we're ok for now. And if we do try to use an API we can't use in Collect, it's at that point that the JR Android check will catch it.

Ok, works for me!

lognaturel (approved):

PR #5182, opened 2022-07-11
Use the same icon in all notifications

### LINKED ISSUE #4995: Foreground service notifications should use the same notification
Our foreground services currently have different notification icons - audio recording uses a specific recording one and location tracking uses the same icon as our other notifications (`R.drawable.ic_notification_small`). These should be made consistent, so it's clear to users when Collect is up to something in the background - they should all use `R.drawable.ic_notification_small`.

### Notes

We'll need a way to share the icon between multiple packages for this to work. It could be that using a "provider" interface that the `Application` can implement (similarly to `LocalizedApplication`) would solve this.

grzesiek2010:
Closes #4995 

The only thing of note here is the way we share the icon used in notifications across different modules. I decided to just place that icon in the `androidshared` module. I like this option more than for example initializing modules and injecting that icon because if we decided to chose the second option we would need to do that for other resources to be consistent (like for `ic_map_point` for example). In the future we would end up with bigger and bigger providers/initializers.

It's a safe fix that should have no side effects. Please just verify that all notifications use the same icon.

Any form with the [Audio Widget](https://docs.getodk.org/form-question-types/#audio-widgets).

No.


seadowg (changes requested):
I'd like to try and avoid these kinds of static dependencies as they prevent they pollute these modules with "Collect" specific links that would make it hard for another app to use the module (not essential, but I think a good standard to try and adhere to in terms of the levels of separation we have in our modules). However, our reliance on `strings` already breaks that. If we wanted that kind of separation, we'd also need a dynamic/inverted way of grabbing strings.

Just to keep the `androidshared` a little cleaner, could we introduce a `icons` module that we use to share icons between modules (like we do with `strings`)? That at least makes it very clear when a module needs access to external (to the module itself) icons.

settings.gradle
```diff
@@ -26,8 +26,9 @@ include ':settings'
 include ':servicetest'
 include ':maps'
 include ':osmdroid'
+include ':icons'
 
 apply from: 'secrets.gradle'
 if (getSecrets().getProperty('MAPBOX_DOWNLOADS_TOKEN', '') != '') {
     include ':mapbox'
-}
+}
```
seadowg: Looks like there's a missing new line at the end of the file here.

seadowg (approved):

grzesiek2010:
> Just to keep the androidshared a little cleaner, could we introduce a icons module that we use to share icons between modules (like we do with strings)? That at least makes it very clear when a module needs access to external (to the module itself) icons.

This should be ready. I will file a separate issue for moving other icons too.

settings.gradle
grzesiek2010: Why do we need that line?

seadowg: Git (and Github) warns against it (there's a good SO on why [here](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log)). You can set up Android Studio to make sure you always have a new line in its settings (Editor > General > On Save > Ensure every saved file ends with a line break).

grzesiek2010: Ok done.

dbemke:
**Tested with Success!**
Verified on Androids: 8.1, 11

**Verified icons in notifications:**

- for audio recording in Audio widget (All widgets form)
- for background audio recording
- for internal audio recording with other questions
- for location recording in Geopoint, Geoshape/Geotrace with automatic location recording
- form update failed/succeeded
- form upload succeeded
- when match exactly is enabled and disabled

srujner:
**Tested with Success!**
Verified also on Androids: 5.1, 10 and 12

PR #5188, opened 2022-07-14
Selection summary redesign

### LINKED ISSUE #5136: Redesign selection summary sheet

seadowg:
Closes #5136 

The selection summary now has an updated design based on what's specified in the issue. The one difference is that the title does not have the primary color (blue) background as it felt like that gave the sheet an older style (of Material Design) look.

Verified manually.

Not a lot to discuss here. Will leave comments on anything work highlighting.

This should be fairly low risk. It'd be good to check the select one from map and form map (shouldn't need to use different mapping engines).

### DIFF
--- a/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
+++ b/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
@@ -274,11 +274,12 @@ class SelectionMapFragment(
                 }
                 map.setMarkerIcon(featureId, item.largeIcon)
                 summarySheet.setItem(item)
+
+                summarySheetBehavior.state = STATE_COLLAPSED
                 summarySheet.viewTreeObserver.addOnGlobalLayoutListener(object : ViewTreeObserver.OnGlobalLayoutListener {
                     override fun onGlobalLayout() {
                         summarySheet.viewTreeObserver.removeOnGlobalLayoutListener(this)
                         summarySheetBehavior.peekHeight = summarySheet.peekHeight
-                        summarySheetBehavior.state = STATE_COLLAPSED
                     }
                 })
 
--- a/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
+++ b/geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
@@ -7,13 +7,14 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.ViewTreeObserver
 import androidx.activity.OnBackPressedCallback
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.LiveData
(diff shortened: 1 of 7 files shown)


geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -273,9 +274,16 @@ class SelectionMapFragment(
                 }
                 map.setMarkerIcon(featureId, item.largeIcon)
                 summarySheet.setItem(item)
-                selectedFeatureViewModel.setSelectedFeatureId(featureId)
 
-                summarySheetBehavior.state = STATE_EXPANDED
+                summarySheetBehavior.state = STATE_COLLAPSED
+                summarySheet.viewTreeObserver.addOnGlobalLayoutListener(object : ViewTreeObserver.OnGlobalLayoutListener {
```
seadowg, line 279: We need to use a `GlobalLayoutListener` here as otherwise the `peekHeight` could end up being calculated before the view has rendered and be inaccurate.

geo/src/main/java/org/odk/collect/geo/selection/SelectionSummarySheet.kt
```diff
@@ -20,6 +22,20 @@ internal class SelectionSummarySheet(context: Context, attrs: AttributeSet?) :
...
 
+    val peekHeight: Int
+        get() {
+            return when (binding.properties.size) {
+                0 -> binding.properties.top
+                1 -> binding.properties.top + binding.properties[0].bottom
+                else -> {
+                    val bottomOfFirstProp = binding.properties.top + binding.properties[0].bottom
+                    val secondPropHeight = binding.properties[1].bottom - binding.properties[1].top
+                    val enoughOfSecondPropToImplyMoreProps = (secondPropHeight / 12) * 7
```
seadowg, line 33: This is a bit of a magic number but seemed like the right ratio to me.

grzesiek2010 (approved):

lognaturel:
Merging so this is in beta. @getodk/testers please verify on `master`! It should be really quick -- not a lot of risk here.

dbemke:
**Tested with Success!**

Verified on Androids: 8.1, 11

**Verified cases:**

- color of the form icon of a selected item on map of forms
- divider between each property of the selected item on the map
- showing the bottom of the only property for items with a single property
- showing the middle of the second property for items with multiple properties
- the bottom sheet of an item on the map can be expanded
- light and dark mode

srujner:
**Tested with Success!**

PR #5189, opened 2022-07-15
Close databse connections for given project when it is removed

### LINKED ISSUE #4870: Collect crashes when Try a demo project after deleting all projects on Android 5.1

grzesiek2010:
Closes #4870 

I've verified the issue manually.

I think it's the only solution we just need to close database connections before removing them at all.

Please verify that the issue does not occur. You can also test deleting other projects (not only demo ones) to make sure ther is no regression.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
@@ -59,6 +59,22 @@ open class DatabaseConnection(
         private val toClose = mutableListOf<SQLiteOpenHelper>()
 
         @JvmStatic
+        fun cleanUp() {
+            val openHelpersToClear = mutableListOf<String>()
+
+            openHelpers.forEach { (databasePath, openHelper) ->
+                if (!File(databasePath).exists()) {
+                    openHelper.close()
+                    openHelpersToClear.add(databasePath)
+                }
+            }
+
+            openHelpersToClear.forEach {
+                openHelpers.remove(it)
+            }
+        }
+
+        @JvmStatic
         fun closeAll() {
             openHelpers.forEach { (_, openHelper) -> openHelper.close() }
             openHelpers.clear()
--- a/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/ProjectDeleter.kt
@@ -2,6 +2,7 @@ package org.odk.collect.android.projects
 
 import org.odk.collect.android.backgroundwork.FormUpdateScheduler
 import org.odk.collect.android.backgroundwork.InstanceSubmitScheduler
+import org.odk.collect.android.database.DatabaseConnection
 import org.odk.collect.android.utilities.ChangeLockProvider
 import org.odk.collect.forms.instances.Instance
 import org.odk.collect.forms.instances.InstancesRepository
@@ -59,6 +60,8 @@ class ProjectDeleter(
         projectsRepository.delete(currentProject.uuid)
 
         File(projectDirPath).deleteRecursively()
+
+        DatabaseConnection.cleanUp()
 
         return if (projectsRepository.getAll().isNotEmpty()) {
             val newProject = projectsRepository.getAll()[0]


seadowg (changes requested):
I'm seeing 226 test cases fail on Test Lab when running all the instrumentation tests for this. I remember that was a problem with a related change we tried to make around closing db connections. Are you seeing those fails as well?

grzesiek2010:
> Are you seeing those fails as well?

Yeah there was something wrong but I rebased onto master and it's green.

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
```diff
@@ -58,6 +58,22 @@ open class DatabaseConnection(
         private val openHelpers = mutableMapOf<String, SQLiteOpenHelper>()
         private val toClose = mutableListOf<SQLiteOpenHelper>()
 
+        @JvmStatic
+        fun close(projectId: String) {
```
seadowg: I'm not a huge fan of that this method cares about the project ID, as it doesn't really know (or doesn't need to know) anything about projects. Also, given the implementation, this parameter could really just be called `subDirectory` as if my db was in `/dir1/dir2/projectDir/` I coud pass `dir1` or `dir2` into this and it would close (and clear) the connection.

I'm wondering if an alternative approach would be to have a `cleanUp` method that closes and clears connections that are to no longer existing fils (sort of like we do just-in-time in when grabbing a db helper)? We could call this after deleting the project dir and it should then remove the helpers we want, without us needing to care about any identifiers in `DatabaseConnection`.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/database/DatabaseConnection.kt
grzesiek2010: So it's a similar solution it just check the file existence instead of finding it by project id. Updated, please take a look if it's what you expected.

seadowg (approved):

dbemke:
**Tested with Success!**

Verified on Androids: 8.1, 10, 11, 13 (on the emulator)

**Verified cases:**

- issue #4870 is no longer reproducing
- deleting the demo project and opening a new demo project
- deleting different projects and opening the demo project
- deleting a duplicate of a demo and a different project
- deleting manually entered project
- deleting Google Drive, Aggregate, Central projects
- trying to delete a project with unsent submissions
- deleting a project after resetting
- deleting an encrypted project
- trying to delete a project when background jobs are still running in the project
- deleting a Draft Testing Form
- adding manually/via QR code the same project which was deleted (demo and other)

srujner:
**Tested with Success!**

Verified on Android: 5.1

PR #5190, opened 2022-07-18
Use Material 3 Filled Button for selection summary action

seadowg:
We realized that our heavily customized Material Chip button looked exactly like [Material 3's Filled Button](https://m3.material.io/components/buttons/guideline). Before moving on, I wanted to investigate using that component and... success! It looks like we can use Material 3 components just fine with our `MaterialComponents` theme as long as we define the theme attributes that a component requires (an exhaustive list is [here](https://m3.material.io/libraries/mdc-android/getting-started)).

Verified manually.

Using Material 3 cuts down on a lot of customization here. We should think about adopting Material 3 components for other new UIs in the app from now on.

Should be very low risk but worth playing around with the selection summary sheet seeing as this is our first Material 3 component.

### DIFF
--- a/collect_app/src/main/res/values/theme.xml
+++ b/collect_app/src/main/res/values/theme.xml
@@ -46,6 +46,11 @@
         </item>
         <!--/Material theme attributes-->
 
+        <!-- Material3 attributes (needed for Material3 styles/components because theme is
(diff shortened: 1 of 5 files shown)


lognaturel (approved):
We called it!

I played around with it and don't see any risk. The font is a little different but seems fine to me. Merge now?

PR #5194, opened 2022-07-20
Fix checking supported abis

### LINKED ISSUE #5192: androidx.startup.StartupException when initializing mapbox
Collect v2022.3-beta.2

#### Problem description
https://console.firebase.google.com/u/1/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/eb9b8a0d7568854e2e09bc2b598fd781?time=last-seven-days&versions=v2022.3-beta.2%20(4459)&types=crash&sessionEventKey=62D23A36029200011E17896C87C1C642_1699263640304150428

```
Fatal Exception: androidx.startup.StartupException: androidx.startup.StartupException: androidx.startup.StartupException: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/system/framework/android.test.runner.jar", zip file "/system/framework/android.test.mock.jar", zip file "/system/framework/com.google.android.maps.jar", zip file "/system/framework/org.apache.http.legacy.boot.jar", zip file "/data/app/androidx.test.tools.crawler-MflW_00EOpJgclb9SgHiJg==/base.apk", zip file "/data/app/org.odk.collect.android-E9bdXKuvmIGldxw5iJpY7g==/base.apk"],nativeLibraryDirectories=[/data/app/androidx.test.tools.crawler-MflW_00EOpJgclb9SgHiJg==/lib/x86, /data/app/org.odk.collect.android-E9bdXKuvmIGldxw5iJpY7g==/lib/x86, /data/app/androidx.test.tools.crawler-MflW_00EOpJgclb9SgHiJg==/base.apk!/lib/x86, /data/app/org.odk.collect.android-E9bdXKuvmIGldxw5iJpY7g==/base.apk!/lib/x86, /system/lib, /system/vendor/lib]]] couldn't find "libmapbox-common.so"
       at androidx.startup.AppInitializer.doInitialize(AppInitializer.java:187)
       at androidx.startup.AppInitializer.doInitialize(AppInitializer.java:138)
       at androidx.startup.AppInitializer.initializeComponent(AppInitializer.java:117)
       at org.odk.collect.mapbox.MapBoxInitializationFragment.initializeMapbox(MapBoxInitializationFragment.kt:73)
       at org.odk.collect.mapbox.MapBoxInitializationFragment.onAttach(MapBoxInitializationFragment.kt:32)
       at androidx.fragment.app.Fragment.performAttach(Fragment.java:2954)
       at androidx.fragment.app.FragmentStateManager.attach(FragmentStateManager.java:463)
       at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:254)
       at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1840)
       at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1758)
       at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1701)
```

grzesiek2010:
Closes #5192 

I've run tests on firebase (where I was able to reproduce the issue).

As I said in the issue - `The most preferred ABI is the first element in the list` so we need to compare the first abi with the supported ones instead of comparing the whole list.

Please just verify that mapbox is still available on the devices you have.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
@@ -14,7 +14,7 @@ object MapboxClassInstanceCreator {
             "armeabi-v7a"
         )
 
-        return createMapboxMapFragment() != null && Build.SUPPORTED_ABIS.any(mapboxAbis::contains)
+        return createMapboxMapFragment() != null && mapboxAbis.contains(Build.SUPPORTED_ABIS.first())
     }
 
     fun createMapboxMapFragment(): MapFragment? {
--- a/collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
@@ -1,6 +1,5 @@
 package org.odk.collect.android.application
 
-import android.os.Build
 import androidx.fragment.app.Fragment
 import org.odk.collect.maps.MapConfigurator
 import org.odk.collect.maps.MapFragment
@@ -9,12 +8,12 @@ object MapboxClassInstanceCreator {
 
     @JvmStatic
     fun isMapboxAvailable(): Boolean {
-        val mapboxAbis = listOf(
-            "arm64-v8a",
-            "armeabi-v7a"
-        )
-
-        return createMapboxMapFragment() != null && mapboxAbis.contains(Build.SUPPORTED_ABIS.first())
+        return createMapboxMapFragment() != null && try {
+            System.loadLibrary("mapbox-common")
+            true
+        } catch (e: Throwable) {
+            false
+        }
     }
 
     fun createMapboxMapFragment(): MapFragment? {


collect_app/src/main/java/org/odk/collect/android/application/MapboxClassInstanceCreator.kt
```diff
@@ -14,7 +14,7 @@ object MapboxClassInstanceCreator {
             "armeabi-v7a"
         )
 
-        return createMapboxMapFragment() != null && Build.SUPPORTED_ABIS.any(mapboxAbis::contains)
+        return createMapboxMapFragment() != null && mapboxAbis.contains(Build.SUPPORTED_ABIS.first())
```
seadowg: This feels very unlikely, but can we protect from `SUPPORTED_ABIS` being empty here? I guess adding `!Build.SUPPORTED_ABIS.isEmpty()` as part of the `&&` expression would do.

seadowg (changes requested):
Just one wee tweak!

It's fascinating that the virtual device returns x86 and arm arches - hard to imagine a device that would do that out in the wild.

I'm also wondering if we should just attempt to load the library ourselves and catch the exception instead of trying to inspect the supported architectures here. How would you feel about that?

grzesiek2010:
> I'm also wondering if we should just attempt to load the library ourselves and catch the exception instead of trying to inspect the supported architectures here. How would you feel about that?

I don't have any strong feelings. Maybe... I've just added a fix with that approach so please take a a look if you like it I'm ok with it too.

seadowg:
> I don't have any strong feelings. Maybe... 

Same. Both approaches don't feel quite right, but I think I like using `System.loadLibrary` more. It's pretty consistent with how we detect if Mapbox is on the class path as well - seeing if something breaks.

seadowg (approved):

seadowg:
Marked this as `high priority` as other regression fixes rely on it

seadowg:
Marked this as `high priority` as other regression fixes rely on it

srujner:
**Tested with Success!**
Verified on Android 5.1, 10, 12

**Verified Cases:**
- Opening Mapbox on each device;
- Quick regression check on GeoWidgets;
- Offline Layers

dbemke:
Tested with Success!
Verified on Android 8.1, 11

Issue #5168, opened 2022-06-23
Placement by tapping doesn't work for OSM for GeoTrace and GeoShape widgets


dbemke:
**Software and hardware versions** 
Collect v2022.3, Android probably all

**Problem description**

In Geoshape widget and Geotrace widget placement by tapping doesn’t work for OSM. User is not able to set points on the map by tapping in Collect v2022.2.3. 
With OSM placement by tapping works in Collect v2022.2.3. 

**Steps to reproduce the problem**

1. Select OSM as a source of maps
2. Open form with Geo widgets
3. Go to Geotrace widget / Geoshape widget
4. Hit Start
5. Add points by method input- placement by tapping
6. Hit start
7. Tap on the map

**Expected behavior**

User is able to set points on map by tapping.

**Other information** 

Checked on All widgets form. Geoshape and Geotrace works on Google Maps.

https://user-images.githubusercontent.com/56479916/175320705-81dcb69e-f1fd-4d74-8fd8-556a7518d4b1.mp4

lognaturel:
This also feels like a good candidate for squeezing into this release or including in the point release. It’s been released but not for that long.

dbemke:
@seadowg  I think I’ve just found a mistake that I made. I’ve checked that again and (OSM) placement by tapping works well on Android 11 and 8.1 in Collect v2022.2.3. In the video there is Collect v2022.3 beta 0 so this might be regression?

seadowg:
@dbemke yeah if this worked in v2022.2.3 (the last released version) it should be a regression. Could you update the description with that new info and add the "regression" label to this issue please?

PR #5199, opened 2022-07-21
Redesign map buttons

### LINKED ISSUE #3202: Refresh map icons
~~Blocked by #4943~~

Collect v1.22.4 and older versions

#### Problem description
Icons in our map widgets don't look well, probably we should follow Google Maps design using circular icons.

| ODK Collect  | Google maps |
| Screenshot_2019-07-10-00-03-46  | Screenshot_2019-07-09-23-59-22  |

@seadowg if it's not in conflict with your work/plans I would do that.

seadowg:
Closes #3202 

The map buttons now use a [Material 3 FAB](https://m3.material.io/components/floating-action-button/overview) with different sizing inspired by Google Maps. I've also updated map layouts to use `ConstraintLayout`.

Verified manually.

Discussed in the issue and elsewhere.

It would be good to check all the map views now have redesigned buttons and that the arrangement makes sense when buttons are enabled and disabled. A general sweep of mapping features is a good idea, although we shouldn't need to check with different mapping engines (Google, OSM etc) as the map button code sits on top of those.


lognaturel:
The button changes look good to me. It's hard for me to judge how disruptive the changes to order will be so as you suggested, let's merge early in the next release cycle and get feedback. I think a good outcome is that save is always at the bottom near the thumb.

Since we're not pushing to get this into the current release, how about also updating the look of the "Record a Point" button in geotrace/shape manual selection? Also fine if you'd rather file a separate issue, maybe there's more thinking to do there.

~Not sure whether it's related to these changes but placement by tapping doesn't work with osmdroid at all. Not in `placement-map`, not in geotrace/shape. Haven't tried other engines.~ https://github.com/getodk/collect/issues/5168

Again, not sure it's related but the `placement-map` appearance has accuracy at the top with a while background whereas trace and shape use the primary color. I prefer the white background.

lognaturel:
One change this makes is not allowing vertical scroll on buttons to accommodate short screens.

I think that's ok. We know there are folks using small devices but my guess is that the would be less likely to do map-related work: it's just not practical to view maps without a little bit of context. This is something we can ask about when we do our first beta for the next release.

geo/src/main/res/layout/selection_map_layout.xml
```diff
@@ -35,16 +36,16 @@
             android:layout_above="@id/geometry_status"
             android:layout_below="@id/title" />
 
-        <org.odk.collect.androidshared.ui.multiclicksafe.MultiClickSafeImageButton
```
grzesiek2010, line 38: Here we used `MultiClickSafeImageButton` probably becasue it was possible to open form filling twice? I think we should protect users from doing that.

grzesiek2010 (changes requested):

grzesiek2010:
In landscape mode or probably in case of small screens it doesn't look good:
Screenshot_1662565554

seadowg:
> In landscape mode or probably in case of small screens it doesn't look good:

Good catch. I played around with smaller screens a lot, but not landscape. Interestingly, Google Maps seems to show/hide FABs selectively in different UIs (map vs location etc) in landscape. I'll have a play with doing something like that.

seadowg:
@lognaturel @grzesiek2010 I've added a layout specifically for landscape for geoshape/geotrace that puts the larger FABs in a horizontal row rather than a vertical column. Let me know what you think - I can do the same thing for the other map views if we feel like this works well.

grzesiek2010:
> I've added a layout specifically for landscape for geoshape/geotrace that puts the larger FABs in a horizontal row rather than a vertical column. Let me know what you think - I can do the same thing for the other map views if we feel like this works well.

It looks good to me.

geo/src/main/java/org/odk/collect/geo/selection/SelectionMapFragment.kt
```diff
@@ -176,7 +177,7 @@ class SelectionMapFragment(
         }
 
         if (showNewItemButton) {
-            binding.newItem.setOnClickListener {
+            binding.newItem.setMultiClickSafeOnClickListener {
```
grzesiek2010, line 180: I would also fix `binding.layerMenu` above. There it's probably also possible to open multiple dialogs?

grzesiek2010 (approved):

dbemke:
I’ve noticed that a disabled button is barely visible if the dark mode is applied when the map in the background is grey or light green or if the light mode is applied on e.g. Control room offline layer which is dark. 
darkmodebindisabled2
darkmodegreenbackground2
lightmodebindisabled2

dbemke:
Record a point button has the previous design (different style than other map buttons).
recordpoint

seadowg:
@dbemke ah this is fascinating. We don't actually switch theming on the maps, so dark theming really doesn't work on the buttons. Event the visible map buttons look really out of place. I'll have a quick look at seeing if we can switch theming for the maps, but I doubt we'll be able to for all 3 engines. I think we'll probably need to look at forcing a light theme on screens with maps.

seadowg:
@dbemke ok I've made it so that the map buttons use light theming even when dark theme is enabled (as the map itself is always light). The one exception is the manual record button in geo shape/trace. That currently uses an old button style, and I think we'd need to redesign it for it to be worth forcing the light theme.

dbemke:
On light maps the light theme looks nice and a disabled button is visible but it possible to change the style of the map e.g. in Google maps satellite or hybrid style there's dark green in the background.
stylesattelite2
stylehybrid2

seadowg:
@dbemke I've tweaked the transparency of disabled buttons to hopefully make them more (but not too) visible on in geo shape/trace. Take a look and let me know what you think. If it's good enough, I can make that change in the other map views as well.

dbemke:
@seadowg I've checked different backgrounds (light and dark) and now disabled buttons are visible in every case.

seadowg:
@dbemke ok that's fixed everywhere now.

srujner:
**Tested with Success!**

Verified on Emulator with Android 13

Verified Cases:

- Issue from the comments is no longer reproducing;
- All map views;
- Enabled and Disabled buttons;
- Different map sources;
- Light and Dark mode;
- Horizontal and Vertical View;
- Buttons arrangement; 
- Overall UI look;
- Quick regression check on GeoPoint widgets, Submissions Select map and External map widgets;
- Offline Layers 
- RTL language

dbemke:
**Tested with Success!**

Verified on Android: 10

PR #5206, opened 2022-07-23
 Add markers all at once in Mapbox

### LINKED ISSUE #5177: Crash of the app after minimizing the app or locking the screen while map is loading on geojson form with thousand points used with Mapbox.

Android All versions

**Problem description**

When Mapbox is used as a source of maps on geojson form and the map is loading, minimizing the app or locking the screen causes a crash of the app.

    1. Select Mapbox in Settings -> Maps 
    2. Download geojson form with thousand points 
[geojson.xml.txt](https://github.com/getodk/collect/files/9053826/geojson.xml.txt)
[thousand-points.geojson.txt](https://github.com/getodk/collect/files/9053827/thousand-points.geojson.txt)

    3. Click Select place
    4. Minimize the app/ Lock the screen
    5. Go back to the app
       
**Expected behavior**

Map should load. The app doesn’t crash.

**Other information**

Checked on geojson form with thousand points.Works well on Google Maps.

seadowg:
Closes #5177
~~Blocked by #5194~~

We'd talked about making this change before (in #5087) for example, but had decided to hold off until we heard feedback from users around performance. It seems that with the Mapbox v10 upgrade, minimizing and reopening the app can cause an ANR when adding large amounts of markers. I wasn't sure exactly why, but switching to adding markers in one go (against the Mapbox SDK) fixes this problem and **dramatically** speeds up loading for larger sets of items in selection maps (form map or select one from map widget).

Verified manually.

I did also try an approach of offloading some of the marker generation code to the background, but this didn't help - the ANR could still be reproduced. My theory is that the continual adding of markers (one by one) while the app was minimized caused problems for Mapbox (which seems very plausible).

I've had to make changes to the selection map for all the map engines, so we should do a QA pass on that again. 

### DIFF
--- a/collect_app/build.gradle
+++ b/collect_app/build.gradle
@@ -181,12 +181,14 @@ android {
         // abiFilters to exclude any ABIs; but to keep the APK slim, we include
         // the Mapbox native library only for 32-bit and 64-bit ARM devices and
         // omit it for all X86 devices.
-        exclude 'lib/x86/libmapbox-maps.so'
-        exclude 'lib/x86/libmapbox-common.so'
-        exclude 'lib/x86/libc++_shared.so'
-        exclude 'lib/x86_64/libmapbox-maps.so'
-        exclude 'lib/x86_64/libmapbox-common.so'
-        exclude 'lib/x86_64/libc++_shared.so'
+        if (!project.hasProperty("x86Libs")) {
+            exclude 'lib/x86/libmapbox-maps.so'
+            exclude 'lib/x86/libmapbox-common.so'
+            exclude 'lib/x86/libc++_shared.so'
+            exclude 'lib/x86_64/libmapbox-maps.so'
+            exclude 'lib/x86_64/libmapbox-common.so'
+            exclude 'lib/x86_64/libc++_shared.so'
+        }
     }
 
     compileOptions {
--- a/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
+++ b/collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
@@ -271,6 +271,17 @@ public class GoogleMapFragment extends SupportMapFragment implements
         return featureId;
     }
 
+    @Override
+    public List<Integer> addMarkers(List<MarkerDescription> markers) {
+        List<Integer> featureIds = new ArrayList<>();
+        for (MarkerDescription markerDescription : markers) {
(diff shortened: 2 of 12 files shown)


collect_app/build.gradle
```diff
@@ -181,12 +181,14 @@ android {
         // abiFilters to exclude any ABIs; but to keep the APK slim, we include
         // the Mapbox native library only for 32-bit and 64-bit ARM devices and
         // omit it for all X86 devices.
-        exclude 'lib/x86/libmapbox-maps.so'
-        exclude 'lib/x86/libmapbox-common.so'
-        exclude 'lib/x86/libc++_shared.so'
-        exclude 'lib/x86_64/libmapbox-maps.so'
-        exclude 'lib/x86_64/libmapbox-common.so'
-        exclude 'lib/x86_64/libc++_shared.so'
+        if (!project.hasProperty("x86Libs")) {
```
seadowg, line 184: I found it very helpful when trying to reproduce the original issue to be able to include x86 libs to try Mapbox on various different emulators.

collect_app/src/main/java/org/odk/collect/android/geo/GoogleMapFragment.java
```diff
@@ -271,6 +271,17 @@ public void onSaveInstanceState(@NonNull Bundle outState) {
         return featureId;
     }
 
+    @Override
+    public List<Integer> addMarkers(List<MarkerDescription> markers) {
```
seadowg, line 275: For the moment OSM and Google just call `addMarker` repeatedly from `addMarkers`. I'm not sure if either have the ability to have a similar implementation to Mapbox, but I didn't want to spend time on that now given I was focused on the crash.

grzesiek2010: I remember that at least in OSM it is possible to add a collection of markers. Could you file an issue for that?

grzesiek2010 (approved):

srujner:
**Tested with Success!**

Verified Cases:
- Issue https://github.com/getodk/collect/issues/5177 is no longer reproducing;
- Select Map forms;
- Offline Layers; 
- Google Maps, Mapbox, OSM;
- View Submissions on map; 
- Screen rotation, minimize app, using backward button,

dbemke:
**Tested with Success!**
Verified on Androids: 8.1, 11

PR #5212, opened 2022-07-26
Dependency updates for v2022.4 + cleaning

grzesiek2010:
I've ran automated tests and reviewed implemented chnages.

As usual I've updated dependencies for next release. This time I've additionally added some cleaning removing unused dependencies and rxjava which was used only in one place (`READ_PHONE_STATE` permission in forms) so I've replaced it with `livedata`.

This is a kind of change we usually do not test explicitly but merge and do that on a daily basis but this time I've added some changes that we should test first:
- verify that scanning QR codes works well especially on older devices (Android 5)
- verify that requesting `READ_PHONE_STATE` permission works as before in forms. This Permission is required if a from contains [phonenumber](https://docs.getodk.org/form-question-types/#metadata) hidden question. In such a case when a form is started a user should be asked for granting that permission.

No.

No.


collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
```diff
@@ -134,7 +133,7 @@ public List<String> getProperty(String propertyName) {
     @Override
     public String getSingularProperty(String propertyName) {
         if (!permissionsProvider.isReadPhoneStatePermissionGranted() && isPropertyDangerous(propertyName)) {
-            eventBus.post(new ReadPhoneStatePermissionRxEvent());
+            isPhoneNumberRequired.postValue(true);
```
seadowg, line 138: Could this be `setValue` or is `getSingularProperty` called from background (non-UI) threads?

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceUploaderAdapter.java
```diff
@@ -17,24 +17,13 @@
 
 import java.util.Date;
 
-import io.reactivex.disposables.CompositeDisposable;
-
 import static org.odk.collect.forms.instances.Instance.STATUS_SUBMISSION_FAILED;
 import static org.odk.collect.forms.instances.Instance.STATUS_SUBMITTED;
 
 public class InstanceUploaderAdapter extends CursorAdapter {
-    private final CompositeDisposable compositeDisposable;
```
seadowg, line 26: Was this just here and not being used for anything?

seadowg (approved):

seadowg:
🙏 Praise be to @grzesiek2010 for removing RxJava. I think you get to update `STATE.md` as well here.

collect_app/src/main/java/org/odk/collect/android/adapters/InstanceUploaderAdapter.java
grzesiek2010: Yes

collect_app/src/main/java/org/odk/collect/android/logic/PropertyManager.java
grzesiek2010: Yes it's used from background so it needs to be `postValue`

grzesiek2010:
> I think you get to update STATE.md as well here.

Right, done.

dbemke:
**Problem description**

After allowing ODK Collect to make and manage phone calls, opening another form results in continuous loading form state.

1. Go to the Metadata form.
[metadata.xml.txt](https://github.com/getodk/collect/files/9394548/metadata.xml.txt)

2. Allow ODK Collect to make and manage phone calls.
3. Exit the form (e.g click device back button and ignore changes).
4. Go to any other form (e.g. All widgets)

**Expected behavior**

Another form should open. The user should be able to complete the form.

**Other information** 

If the user minimizes the app (while loading form is on) and goes back to the form, the form works as expected.
On master allowing to make and manage phone calls doesn't result in continuous loading of forms, but it's possible to get this result by denying ODK Collect making and managing phone calls and in next dialog opening settings, allowing access to phone and going back to the form.

grzesiek2010:
The bug should be fixed. If there is anything else but it exists on master too please file a separate issue.

srujner:
**Tested with Success!**
Verified on Androids: 5.1, 10 and 12

Verified cases:

- Adding project through QR code; 
- Switching between projects, adding duplicated projects; 
- Invalid/Inverted QR code;
- Scanning barcodes and QR code with front and back camera;

- Issue found by @dbemke is no longer reproducing;
- Requesting Phone permission dialog;
- Allowing and Denying permissions in the ODK app and in the device settings;
- Screen rotation, minimize app, using backward button;
- Light and Dark theme.

dbemke:
**Tested with Success!**
Verified on Androids: 8.1, 11

PR #5215, opened 2022-07-27
Track usage of cellular_only

grzesiek2010:
I've just reviewed implemented changes.

Nothing to discuss here. We just want to know if that option is used.

It doesn't require testing.

No.

No.


seadowg (approved):
Would definitely be useful, but I'm interested to know if @lognaturel wants more info. Currently, this would answer the question "how often are forms sent with cellular only?". Would you want it to log a server event instead so that we can know if it's a small or large number of projects using this?

Another alternative would be to use a user property (like we do for non default theme etc) instead so that we could answer the question "how many devices have a project with cellular only set?".

lognaturel (approved):
Happy to start with this and then adjust based on what we find out. I'm expecting/hoping this will be very close to 0.

PR #5218, opened 2022-07-27
Update ISSUE_TEMPLATE.md

seadowg:
I've found that everyone is using slightly different formats for the version identifiers, and it's not always consistent what info we get. That makes sense to me as it's not really clear if we want the Android version or the ODK Collect version (although we do give an example). I've tried to make it clearer what data we want and split it out so that it should always be in a consistent order.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -10,11 +10,11 @@ Feature suggestions should be described [in the forum Features category](https:/
 
 -->
 
-### ODK Collect version
+#### ODK Collect version
 
-### Android version
+#### Android version
 
-### Device used
+#### Device used
 
 #### Problem description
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -10,8 +10,11 @@ Feature suggestions should be described [in the forum Features category](https:/
 
 -->
 
-#### Software and hardware versions 
-Collect v1.x.x, Android vx.x.x, device used...
+### ODK Collect version
```
grzesiek2010: Did you make this text bigger than `Problem description`/`Expected behavior`/`Other information` on purpose or is it a mistake?

seadowg: That's a mistake. Will fix.

grzesiek2010 (approved):

PR #5219, opened 2022-07-27
Retain accuracy for marker point in Mapbox

### LINKED ISSUE #5217: Saving point for the second time in Geopoint widget with maps/ placement-map changes altitude and accuracy to zero when used with Mapbox

Collect v2022.3 master f2ca, Android all

**Problem description**

Saving point for the second time in Geopoint widget with maps/ placement-map changes altitude and accuracy to zero when used with Mapbox.

https://user-images.githubusercontent.com/56479916/181252998-630c4626-9f3f-4feb-b964-b9a8e0fee49a.mp4

Steps to reproduce the problem

1. Choose Mapbox as a source of maps.
2. Go to All widgets form.
3. Go to Geopoint widget with maps/ placement-map
4. Save point by clicking save icon.
5. Click „View or Change Location”
6. Save point by clicking save icon.

**Expected behavior**

User should see latitude, longitude, altitude and accuracy values different from zero.

**Other information** 

I’ve checked this on Andoroid 8.1 and 11 Collect v2022.3 beta 2 and Collect v2022.2.3 - it works as expected.

seadowg:
Closes #5217 

Verified manually. Definitely feels like it would be nice to have tests for this, but we haven't ventured into writing tests in our mapping engine implementations.

Comments inline.

Just need to check the issue is fixed!

### DIFF
--- a/mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
+++ b/mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
@@ -295,21 +295,23 @@ class MapboxMapFragment :
             MapUtils.createPointAnnotations(requireContext(), pointAnnotationManager, markers)
 
         val featureIds = mutableListOf<Int>()
-        pointAnnotations.forEach {
-            val featureId = nextFeatureId++
-            val markerFeature = MarkerFeature(
-                requireContext(),
-                pointAnnotationManager,
-                it,
-                featureId,
-                featureClickListener,
-                featureDragEndListener,
-                MapPoint(it.point.latitude(), it.point.longitude())
-            )
+        markers.asSequence()
+            .zip(pointAnnotations.asSequence())
+            .forEach { (marker, pointAnnotation) ->
+                val featureId = nextFeatureId++
+                val markerFeature = MarkerFeature(
+                    requireContext(),
+                    pointAnnotationManager,
+                    pointAnnotation,
+                    featureId,
+                    featureClickListener,
+                    featureDragEndListener,
+                    marker.point
+                )
 
-            featureIds.add(featureId)
-            features[featureId] = markerFeature
-        }
+                featureIds.add(featureId)
+                features[featureId] = markerFeature
+            }
 
         return featureIds
     }


mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
```diff
@@ -295,21 +295,23 @@ class MapboxMapFragment :
...
-            val markerFeature = MarkerFeature(
-                requireContext(),
-                pointAnnotationManager,
-                it,
-                featureId,
-                featureClickListener,
-                featureDragEndListener,
-                MapPoint(it.point.latitude(), it.point.longitude())
-            )
+        markers.asSequence()
```
seadowg, line 298: Use [`Sequence`](https://kotlinlang.org/docs/sequences.html) here to avoid extra loops due to the `zip`.

mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
```diff
@@ -295,21 +295,23 @@ class MapboxMapFragment :
...
-        pointAnnotations.forEach {
-            val featureId = nextFeatureId++
-            val markerFeature = MarkerFeature(
-                requireContext(),
-                pointAnnotationManager,
-                it,
-                featureId,
-                featureClickListener,
-                featureDragEndListener,
-                MapPoint(it.point.latitude(), it.point.longitude())
```
seadowg, line 307: The problem was here: we were recreating the point that we save on the `MarkerFeature` and leaving altitude and accuracy behind.

lognaturel (approved):
This looks good to me. I could try this out, merge, release but I do have a slight preference for another round of QA. This also affected geotrace and geoshape, right?

mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
grzesiek2010: Hmm now this seems so complicated with that sequence and zip and it's even too late for me to wrap my head around it so I'm just going to ask: can't this whole function look like this?:
```
    override fun addMarkers(markers: List<MapFragment.MarkerDescription>): List<Int> {
        val featureIds = mutableListOf<Int>()
        markers.forEach {
            val featureId = nextFeatureId++
            val markerFeature = MarkerFeature(
                requireContext(),
                pointAnnotationManager,
                MapUtils.createPointAnnotation(pointAnnotationManager, it.point, it.isDraggable, it.iconAnchor, it.iconDrawableId, requireContext()),
                featureId,
                featureClickListener,
                featureDragEndListener,
                it.point
            )

            featureIds.add(featureId)
            features[featureId] = markerFeature
        }

        return featureIds
    }
```

lognaturel: You don't find the instantiation harder to read? I also would expect your version to be slower, @grzesiek2010, though I don't know if it'd be noticeable.

grzesiek2010: Why would my version be slower? Please notice that the original method has creating all `pointAnnotations` at the beggining:
```
        val pointAnnotations =
            MapUtils.createPointAnnotations(requireContext(), pointAnnotationManager, markers)
```
what I don't need so I removed it. Instead I create those `pointAnnotations` one by one what should take the same amount of time. Thanks to that I don't need that `asSequence` and `zip`. So I think my solution has even fewer operations and should be maybe even faster not slower.

seadowg: > Why would my version be slower?

Creating the point annotations one at a time is what caused #5177. That's why the code now creates all the point annotations at once (and up front as you say) with `PointAnnotationManager#create(options: List<PointAnnotationOptions>)`. I'm guessing that's what @lognaturel meant.

> Hmm now this seems so complicated with that sequence and zip

Agreed that it's a little awkward, especially mixing the more functional operations with a `forEach` here. I think I feel uncomfortable using `map` with side effects, but I guess I could have just split the side effect out to deal with that. I think we should leave this like this for now as the PR is in testing, but I think it'd be a cool candidate to play around with for a follow-up PR - it's got a bit of everything: needing to combine values to make other values, side effects and returning a subset values created. It's a nice opportunity to explore functional collection operations, sequences etc. Do you want to have a stab at a version you like more, and then we can discuss as a draft PR @grzesiek2010?

grzesiek2010: Ah right I missed the fact that creating all the points at once was an improvement. Ok I can take a look at it later to see if I can come up with something better (at least to me) and then we can discuss.

srujner:
**Tested with Success!!**
Verified on Androids 5.1, 10, 12

Verified cases:

- Issue https://github.com/getodk/collect/issues/5217 is no longer reproducing;
- View/Edit Submissions on map and send to Central;
- Google Maps, Mapbox, OSM; 
- Use offline layer; 
- Regression checks on geowidgets, background location;

dbemke:
**Tested with Success!!**
Verified on Androids 8.1, 11

mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
lognaturel: > Creating the point annotations one at a time is what caused https://github.com/getodk/collect/issues/5177.

I'll be honest, I forgot that was the root cause, it just looked like it would perform worse. 😄 If we do revisit this code, let's maybe write a comment about perf because it seems like the kind of thing that someone would restructure and accidentally cause a massive performance regression.

PR #5220, opened 2022-07-28
Track form edits

### LINKED ISSUE #5200: Capture analytics for form instance edits
Currently, form instances can always be edited whether they are finalized or not. This makes the meaning of finalization unclear and confusing. We are considering moving towards a model where finalized forms are immutable. It would be helpful to have a sense of how common edits are and what fraction of those edits happen on finalized forms.

grzesiek2010:
Closes #5200 

I've reviewed implemented changes and debugged the project.

Nothing to discuss here.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/InstanceChooserList.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/InstanceChooserList.java
@@ -33,13 +33,17 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 
 import org.odk.collect.android.R;
 import org.odk.collect.android.adapters.InstanceListCursorAdapter;
+import org.odk.collect.android.analytics.AnalyticsEvents;
+import org.odk.collect.android.analytics.AnalyticsUtils;
 import org.odk.collect.android.dao.CursorLoaderFactory;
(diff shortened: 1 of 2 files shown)


seadowg (changes requested):
If we fire these events after `FormEntryActivity#formControllerAvailable` (i.e. once the form has loaded), we could just call `logFormEvent` without needing to do any additional DB work as we'll have set the default param for `"form"`.

grzesiek2010:
> If we fire these events after FormEntryActivity#formControllerAvailable (i.e. once the form has loaded), we could just call logFormEvent without needing to do any additional DB work as we'll have set the default param for "form".

If I did that in `FormEntryActivity` then yeah I could use form controller to get form id and title instead of reading it from DB but then I would need to read status of a form from database. So it won't help my to avoid using DB + I wanted to avoid adding that code to already big enough `FormEntryActivity`.

seadowg (approved):
>  I wanted to avoid adding that code to already big enough FormEntryActivity.

Yeah fair enough. Don't feel strongly enough about that to block this. @lognaturel good to get your approval on this as well.

lognaturel (approved):
Works for me. I'm hopeful we won't keep this in for long. Thanks!

Issue #5229, opened 2022-08-01
Show loading state when calculating select choices


seadowg:
Like for form saves (#5162), we want to show a loading state in form entry while select choices are loading/calculated. We should show the same loading state as form saves (which uses the `TransparentProgressScreen` in the `form_entry.xml` layout) and should be shown while the choices are loaded.

This loading state should be visible for any kind of widget that needs to load choices. This means anything that calls `FormEntryPrompt#getSelectChoices` or loads choices using "fast external itemsets" or "pull data".

An example of the problems caused here would be [this ANR](https://console.firebase.google.com/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/17ef3e50444d4e36d4959333e3edec7b?time=last-seven-days&sessionEventKey=6541FF6101E50001252171458C8A1D19_1874840294361923375).

## Notes

One possible approach for this is to create a `getChoices` method on `FormEntryViewModel` that returns choices as a reactive value (like `LiveData`). This would allow easy integration of `FormEntryViewModel#isLoading`. This might mean loading choices outside the widget as using `LiveData` would require access to lifecycle at the call site of the `getChoices` method.

seadowg:
@lognaturel and I discussed this and decided that the cases we want to fix here are:
- The ANR example when logging the current screen
- Possible ANRs (and the stutter) when moving back or forwards in the form

We won't look at fixing potential ANRs/stutters that could happen when loading selects in a field list as it looks like the requires a larger rework of how Collect renders question pages that would look more like the "unidirectional data flow" models used by React/Redux apps. That's something we'll hopefully be able to evolve over the next few releases. I'll pull out a separate issue.

PR #5235, opened 2022-08-03
Stop early swipes causes crashes

### LINKED ISSUE #4896: Prevent swipes before form has loaded properly

seadowg:
Closes #4896

Verified manually (with modified code to create a window of time when you can swipe before the form has loaded).

The issue lays out a nice (if I do say so myself), but more time intensive route to solving this. We made some good progress on cleaning up the swipe code in the last release however, and I think we should see how further loading states (like we're adding in #5229) force us to keep making progress there.

Should have no effect although changing form entry swipe code is always a little scary. Good to put form entry and swiping/using next and back buttons through its paces.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
+++ b/collect_app/src/main/java/org/odk/collect/android/listeners/SwipeHandler.java
@@ -89,6 +89,10 @@ public class SwipeHandler {
 
         @Override
         public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY) {
+            if (view == null) {
+                return false;
+            }
+
             FlingRegister.flingDetected();
 
             if (e1 != null && e2 != null


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1492,7 +1492,7 @@ public void showView(SwipeHandler.View next, AnimationType from) {
         // adjust which view is in the layout container...
         SwipeHandler.View staleView = currentView;
         currentView = next;
-        swipeHandler.setView(currentView);
+//        swipeHandler.setView(currentView);
```
grzesiek2010: This should be reverted I guess.

seadowg: Ach I forgot to redo `git add` after taking this out again.

grzesiek2010 (approved):

grzesiek2010:
I think it doesn't need testing so I'm ok with merging but if you want the QA team to take a look at it please add `needs testing` label.

seadowg:
> I think it doesn't need testing so I'm ok with merging but if you want the QA team to take a look at it please add needs testing label.

I mostly agree with you, but I'm feeling nervous about it after my last set of changes to `FormEntryActivity` so going to let QA have a look.

dbemke:
**Tested with success!**

Verified on Androids: 8.1 and 11.

**Tested scenarios:**

- try to reproduce #4896 issue 
- swipe forward/backward 
- next/back button
- try with different forms 
- big forms
- Form Entry Settings: Check/Uncheck Moving backwards button
- Form Management: different Constraint processing behavior

srujner:
**Tested with success!**

PR #5237, opened 2022-08-03
Fall back to first project when no current project set

seadowg:
Should fix [this crash](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/8d842445557fcf3942b4e99ba2f89aec?time=last-seven-days&versions=v2022.3.0%20(4477)&types=crash&sessionEventKey=62EA01B7036F00010B09D6BBA9492EE2_1705956889396036664).

We most likely removed the ability for the user to get into the bad state that causes the crash, but didn't provide a fix for people that are already in that state. This PR attempts that

New tests.

One question is: why didn't we have `CurrentProjectProvider` behave like this before? Potentially, we didn't want to add overprotectiveness in that might allow us to make mistakes elsewhere. An alternative we could switch to if people would prefer we stay away from adding the fallback into `CurrentProjectProvider` would be to add in a new `Upgrade` (or app startup item) that fixes the problem by setting a current project if there isn't one, but there are projects.

This feels relatively safe, but I can't get past the feeling there may have been some reason we didn't want this behaviour in the first place. It'd be good to play around with setting up projects in the app, switching between them, app upgrades etc.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/projects/CurrentProjectProvider.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/projects/CurrentProjectProvider.kt
@@ -22,7 +22,13 @@ class CurrentProjectProvider(
                 throw IllegalStateException("Current project does not exist!")
             }
         } else {
-            throw IllegalStateException("No current project!")
+            val projects = projectsRepository.getAll()
+
+            if (projects.isNotEmpty()) {
+                return projects[0]
+            } else {
+                throw IllegalStateException("No current project!")
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


grzesiek2010 (approved):

lognaturel:
Seems like a natural thing to do. I’m not thinking of issues though I agree it’s strange the implementation wasn’t like that already.

srujner:
**Tested with Success!**

Verified cases:
- No issues or crashes found during testing;
- Update Application from older version;
- Switching between projects;
- Adding duplicated project;
- Adding project manually or through QR code; 
- Export/Import settings with QR code;
- Access Control - User Settings;

dbemke:
**Tested with success!**

Verified on Androids: 8.1 and 11

PR #5238, opened 2022-08-03
Fixed adding offline layers

### LINKED ISSUE #5236: Offline layers do not work with mapbox + satellite
#### ODK Collect version
v2022.3.0

#### Problem description
Offline layers do not work with mapbox + satellite

1.  Select mapbox + sattelites in Settings -> Maps
2.  Add and select the attached offline layers: [test.zip](https://github.com/getodk/collect/files/9251248/test.zip)
3. Open any map view and navigate to Corsica

grzesiek2010:
Closes #5236 

Adding offline layers is a bit problematic we have to add them above style layers (such a layer might consist of over 100 layers like streets for example) but below layers that represent points/lines (to keep points and lines visible) etc. The previous solution attempted to remember the index of the top style layer but it was faulty and selected offline layer was added below the last of layers that selected style layer consisted of. The current solution with using ids instead of indexes works well.

Please verify the case described in the issue and generally changing styles and offline layers. There is no need to test other providers than Mapbox.

No.

No.

### DIFF
--- a/mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
+++ b/mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
@@ -20,7 +20,7 @@ import com.mapbox.maps.MapboxMap
 import com.mapbox.maps.Style
 import com.mapbox.maps.dsl.cameraOptions
 import com.mapbox.maps.extension.style.layers.Layer
-import com.mapbox.maps.extension.style.layers.addLayerAt
+import com.mapbox.maps.extension.style.layers.addLayerAbove
 import com.mapbox.maps.extension.style.layers.generated.LineLayer
 import com.mapbox.maps.extension.style.layers.generated.RasterLayer
 import com.mapbox.maps.extension.style.sources.Source
@@ -95,7 +95,7 @@ class MapboxMapFragment :
     private var tileServer: TileHttpServer? = null
     private var referenceLayerFile: File? = null
     private var clientWantsLocationUpdates = false
-    private var offlineLayerPosition = -1
+    private var topStyleLayerId: String? = null
     private val locationCallback = MapboxLocationCallback(this)
     private var mapFragmentDelegate = MapFragmentDelegate(
         this,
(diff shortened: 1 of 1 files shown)


mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
```diff
@@ -95,7 +95,7 @@ class MapboxMapFragment :
     private var tileServer: TileHttpServer? = null
     private var referenceLayerFile: File? = null
     private var clientWantsLocationUpdates = false
-    private var offlineLayerPosition = -1
+    private var topStyleLayerId = ""
```
seadowg: Could this just be a nullable type? That would enforce that we don't use it before it's set.

seadowg (changes requested):

mapbox/src/main/java/org/odk/collect/mapbox/MapboxMapFragment.kt
grzesiek2010: It could be but I can't see any benefit in this case. Could you explain why it's better to you? It's always set before we use it. Even if it's not set we do not block adding offline layers.

seadowg: Ah will it not matter if we pass `""` to `addLayerAbove`?

grzesiek2010: It shouldn't I guess but ok let's not do that.

seadowg (approved):

srujner:
**Tested with Success!**

**Verified Cases:**

- Issue #5236  is no longer reproducing;
- Regression checks on GeoWidgets for Mapbox;
- Select Map forms;
- Changing map styles;
- Offline Layers in Collect and Project folder + Attached Layer;
- View Submission on map and sending submission to Central;
- Collecting Location in Background;
- External map widgets;
- Quick regression check on Google Maps and OSM
- Light and Dark mode;

dbemke:
**Tested with Success!**
Verified on Androids: 8.1, 11

 Issue #5240 still occurs on Android 11

Issue #5241, opened 2022-08-05
Null formController


grzesiek2010:
#### ODK Collect version
v2022.3.0

#### Android version
9, 10, 11 (maybe other too)

#### Problem description
https://console.firebase.google.com/u/1/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/4e09834954ffc3ee419ef8f9d316d2ed?versions=v2022.3.0%20(4477)&time=last-seven-days&types=crash&sessionEventKey=62EAAD7202EE000131F54D37A810577A_1706145731068221603

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.odk.collect.android.javarosawrapper.FormController.getEvent()' on a null object reference
       at org.odk.collect.android.activities.FormEntryActivity.onScreenRefresh(FormEntryActivity.java:1398)
       at org.odk.collect.android.formentry.FormIndexAnimationHandler.handle(FormIndexAnimationHandler.kt:23)
       at org.odk.collect.android.activities.FormEntryActivity.lambda$setupViewModels$7(FormEntryActivity.java:470)
       at org.odk.collect.android.activities.FormEntryActivity.$r8$lambda$T1V91JgXeCj2xNxpJOqRFsZ76xg(FormEntryActivity.java)
       at org.odk.collect.android.activities.FormEntryActivity$$InternalSyntheticLambda$1$c215df7b263d2a4690ddbf172f65e5f56e8023672153174df88d6c10c303a1fb$3.onChanged(FormEntryActivity.java:4)
       at androidx.lifecycle.LiveData.considerNotify(LiveData.java:133)
       at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:146)
       at androidx.lifecycle.LiveData$ObserverWrapper.activeStateChanged(LiveData.java:468)
       at androidx.lifecycle.LiveData$LifecycleBoundObserver.onStateChanged(LiveData.java:425)
       at androidx.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:354)
       at androidx.lifecycle.LifecycleRegistry.forwardPass(LifecycleRegistry.java:265)
       at androidx.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:307)
       at androidx.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:148)
[...]

PR #5243, opened 2022-08-05
Add logging around suspicious FormEntryActivity code

seadowg:
Adds logging that attempts to give us more info around https://github.com/getodk/collect/issues/5241.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -617,7 +617,6 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
             }
 
             // Not a restart from a screen orientation change (or other).
-            Timber.w("Setting formController to null");
             Collect.getInstance().setFormController(null);
 
             Intent intent = getIntent();
@@ -1396,6 +1395,11 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      */
     @Override
     public void onScreenRefresh() {
+        FormController formController = getFormController();
+        if (formController == null) {
+            Timber.w("FormController in the application class is null, FormController in formEntryViewModel is %s", (formEntryViewModel.isFormControllerSet() ? "not null" : "null"));
+        }
+
         int event = getFormController().getEvent();
 
         SwipeHandler.View current = createView(event, false);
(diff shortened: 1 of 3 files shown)


seadowg:
@grzesiek2010 @lognaturel feel free to add other logs in for things you're curious about here.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -253,6 +255,7 @@ protected void onCleared() {
     }
 
     private void updateIndex() {
+        Timber.w("Update formIndex");
```
grzesiek2010, line 258: I added this to check if the value in livedata is set by us or maybe it's a restored state as I described in my comment https://github.com/getodk/collect/issues/5241#issuecomment-1206333890

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1396,6 +1396,11 @@ public void onScreenChange(Direction direction) {
      */
     @Override
     public void onScreenRefresh() {
+        FormController formController = getFormController();
+        if (formController == null) {
+            Timber.w("FormController in the application class is null, FormController in formEntryViewModel is %s", (formEntryViewModel.isFormControllerSet() ? "not null" : "null"));
```
grzesiek2010, line 1400: Here I want to know if both formControllers are null or maybe there is only something wrong with the one stored in the application class.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -617,6 +617,7 @@ private void loadForm() {
             }
 
             // Not a restart from a screen orientation change (or other).
+            Timber.w("Setting formController to null");
             Collect.getInstance().setFormController(null);
```
grzesiek2010: I'm wondering if it wouldn't be better to add logs in that `setFormController` method to see the sequence of calls...
Maybe we first set it properly but then to null?

seadowg: Yeah that sounds good! Feel to make that change

lognaturel (approved):

PR #5248, opened 2022-08-11
Added more logs

grzesiek2010:
Just reviewed new logs.

Those are just some new logs that might help me to understand the root cause.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -370,6 +370,7 @@ public class FormEntryActivity extends CollectAbstractActivity implements Animat
      */
     @Override
     public void onCreate(Bundle savedInstanceState) {
+        Timber.w("onCreate");
         // Workaround for https://issuetracker.google.com/issues/37124582. Some widgets trigger
         // this issue by including WebViews
         if (Build.VERSION.SDK_INT >= 24) {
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -469,7 +470,7 @@ private void setupViewModels() {
         formEntryViewModel.getCurrentIndex().observe(this, index -> {
             if (index != null && Collect.getInstance().getFormController() == null) {
                 // https://github.com/getodk/collect/issues/5241
-                Timber.e("getCurrentIndex() firing with null getFormController()");
+                Timber.e("getCurrentIndex() firing with null getFormController(). The one stored in formEntryViewModel is %s", (formEntryViewModel.isFormControllerSet() ? "not null" : "null"));
```
seadowg, line 473: Maybe I'm wrong, but I don't think we want this to by dynamic or we'll end up with two events in Crashlytics?

grzesiek2010: I added it just to make sure but I'm pretty sure in all cases it's null controller in the application class and not null in view model. Even if I'm wrong and we end up with two separate events that's also nothing very bad I guess.

lognaturel: Seems ok to me. Either way I think we'll have to search/dig to find these non-fatals. Let's try it.

lognaturel (approved):

Issue #5240, opened 2022-08-04
Crash when restoring to some form entry screens


seadowg:
~~Blocked by #5338 (which adds testing for the flows we're changing)~~

Covers this [crash report](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/7ef6006d57c29eff231ee9f78d2f188e?time=last-seven-days&versions=v2022.3.0%20(4477)&types=crash&sessionEventKey=62EB9C8D00D100017025055AC3E08E99_1706408576049390202) (and [this one](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/7ef6006d57c29eff231ee9f78d2f188e?time=last-seven-days&versions=v2022.3.0%20(4477)&types=crash&sessionEventKey=62EB9C8D00D100017025055AC3E08E99_1706408576049390202)).

#### ODK Collect version

v2022.2.3+

#### Android version

Reproduced on Android 10, but likely a problem in all versions

#### Device used

Pixel 2 emulator

#### Problem description

The app crashes when "restoring" (being reopened after being kicked out of memory rather than closed by the user) if the user was last on some form entry screens. An example is the Select One From Map map view (used in the example below).

#### Steps to reproduce the problem

1. Set "Background process limit" to "No background processes" in Developer Options
2. Open a form with a Select One From Map question
3. Go to question and click "Select Place"
4. Switch to another app and then back again

The app crashes. This also happens (or the app behaves weirdly) when restoring to the:

* Quit form dialog - will show the hierarchy and then the dialog when you exit the hierarchy. Additionally, the dialog title will just be "no form loaded".
* The form hierarchy - will show normal form entry instead of the hierarchy
* Form saving in progress (harder to reproduce but does [happen in reality](https://console.firebase.google.com/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/b7ee403472493aaecd327e67d5560ce7?time=last-seven-days&versions=v2022.4.1%20(4543)&types=crash&sessionEventKey=63971D31037100013CB113600829941E_1754682268331899682)) - will crash

#### Expected behavior

For the moment, we're going to avoid restoring view state when we do a process restore. This means that, to the user, the flow will look exactly the same as loading a form from a save point. To be explicit, in the case where a user is filling a form:

1. User answers some questions
2. Process is killed (either via background process kill or changing permissions in Android settings)
[...]

PR #5250, opened 2022-08-12
Fix audit log event ordering

### LINKED ISSUE #5245: Audit log ordering is incorrect
#### ODK Collect version

v2022.3.1, v2022.3.0

#### Android version

Android 10 (but shouldn't matter)

#### Device used

Pixel 2 Emulator

#### Problem description

The first `question` event in the audit log comes before the `form start` event. This is especially problematic in cases where [`identify-user`](https://docs.getodk.org/form-audit-log/?highlight=audit#enumerator-identification) is enabled, as this first `question` event will be missing the `user` value.

1. Load a form that uses the audit log into Collect
2. Start form entry and save

The first event in the instance's `audit.csv` will be a `question` rather than a `form start`.
 
#### Other information 

Here's a log from v2022.2.3 generated using `IdentifyUserTest#openingForm_andThenEnteringIdentity_proceedsToForm`:

```csv
event,node,start,end,user
form start,,1660209573437,,Lucius
question,/data/what_up,1660209573448,1660209573895,Lucius
end screen,,1660209573903,1660209574158,Lucius
form save,,1660209574412,,Lucius
form exit,,1660209574412,,Lucius
form finalize,,1660209574412,,Lucius
```

And here's what we get from v2022.3.1:

```
event,node,start,end,user
question,/data/what_up,1660210097520,
form start,,1660210098279,,Lucius
question,/data/what_up,1660210098293,1660210098654,Lucius
end screen,,1660210098667,1660210098953,Lucius
form save,,1660210099260,,Lucius
form exit,,1660210099261,,Lucius
form finalize,,1660210099261,,Lucius
```

seadowg:
Closes #5245 

Added new tests and verified manually where I couldn't (comments inline on that).

This PR does the bare minimum to fix the issues. There's a nicer solution coincidently in the works (as part of the fix for #5240). Working on that is how I realized there was an issue with the log in the first place.

Things to focus on when checking this:
* form entry in general
* audit logging (especially with identify user and track change reason enabled)
* moving backwards enabled/disabled (with save points)
* editing forms and resuming from save points


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2176,59 +2175,50 @@ public void denied() {
...
+                                if (!allowMovingBackwards) {
+                                    // we aren't allowed to jump around the form so attempt to
+                                    // go directly to the question we were on last time the
+                                    // form was saved.
+                                    // TODO: revisit the fallback. If for some reason the index
+                                    // wasn't saved, we can now jump around which doesn't seem right.
+                                    FormIndex formIndex = SaveFormIndexTask.loadFormIndexFromFile();
+                                    if (formIndex != null) {
+                                        formController.jumpToIndex(formIndex);
+                                        formControllerAvailable(formController);
```
seadowg, line 2235: Replacing `onScreenRefresh` with `formControllerAvailable` is an untested change. We don't currently have a good way to test save point flows - I'll look at backfilling that for work on #5240. It might be that we want to remove this particular flow (returning to the exact index on a save point load when moving backwards is disabled). Otherwise, it looks like we might want to investigate if there is missing audit logging here.

seadowg:
I know we might need to wait a couple of days for review on this so going to mark this as "needs testing" now.

collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
grzesiek2010: But you were able to verify this change manually at least right?

seadowg: Yes. Sorry I should have said that. @getodk/testers to be very specific about the untested flow here: disable moving backwards and then reload a form from a save point (kill the app during form entry and then restart).

grzesiek2010 (dismissed):

dbemke:
In Collect v2022.2.3 in CSV file (e.g. form AuditTestLocationBackground) there is a different order of „question” and „location tracking enabled”. 
[AuditTestLocationBackground.xml.txt](https://github.com/getodk/collect/files/9351976/AuditTestLocationBackground.xml.txt)
In Collect v2022.2.3 first there was "location tracking enabled" and then "question".
auditcsv2022 2 3
In PR there is "question" and after that "location tracking enabled"
auditcsvpr

seadowg:
@srujner @dbemke looks like there's another problem with the log when editing a form. Here's an extract (I've cut out the first fill for clarity) of editing a form in v2022.2.3:

```
form resume,,1660760793169,,,,
jump,,1660760793183,1660760794203,,,
question,/data/what_up,1660760794238,1660760794924,,,
end screen,,1660760794930,1660760795211,,,
form save,,1660760795576,,,,
form exit,,1660760795576,,,,
form finalize,,1660760795577,,,,
```

And here's the same flow/flow on this branch:

```
form resume,,1660756947920,,,,
jump,,1660756947922,1660756948229,,,
question,/data/what_up,1660756947948,1660756948229,,,
question,/data/what_up,1660756948995,1660756949755,,,
end screen,,1660756949776,1660756950028,,,
form save,,1660756950378,,,,
form exit,,1660756950379,,,,
form finalize,,1660756950379,,,,
```

It seems like we end up with a `question` we don't want in there. I actually found this due to an existing test, but it was only failing intermittently. I'll look at a fix. I'm not sure how much hacking I can do before we need to make a larger change, however.

seadowg:
@srujner @dbemke that's ready to look at again. I've played around with a few different form filling scenarios (and added some more tests) to check that the event order looks good, but that's definitely the place to start exploring. I still feel like we're missing a lot of test coverage, and probably team knowledge of how the feature works, so best to verify things against older versions. Also, remember to play around with editing forms as well as filling in forms for the first time.

@grzesiek2010 I'll dismiss the review as it'd be good for you take another look. I'm not very happy with [de98455](https://github.com/getodk/collect/pull/5250/commits/de98455084a14e49ad29d251270689c7a0772a4e), but I'm trying to keep changes as minimal as possible, and all of this is getting moved around for #5240.

dbemke:
**Tested with Success!**

Verified on Androids: 8.1, 11

**Verified cases:**

- issue #5245 , the case of extra question event and location tracking after question event are no longer reproducing
- form entry in general
- audit log order of: start form, questions, location tracking, resume form, jump, save, exit, finalize
- saving and editing different forms: AuditTestLocationBackground, AuditTest, Track Changes Reason, trackingChangesAndLocationTestForm, Identify User, test audit repeats, Nested repeats for audit
- regression checks of forms
- resuming from a save point
- location and audio permission enabled, disabled
- moving backwards enabled, disabled

srujner:
**Tested with Success!**

Verified also on Androids: 5.1, 10 and 12

grzesiek2010 (approved):

PR #5247, opened 2022-08-11
Add more logging for mysterious reload crash

seadowg:
Add more logging for issue.

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkAll` and confirmed all checks still pass OR confirm CircleCI build passes and run `./gradlew connectedDebugAndroidTest` locally.
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/getodk/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html).
- [x] verified that any new UI elements use theme colors. [UI Components Style guidelines](https://github.com/getodk/collect/blob/master/docs/CODE-GUIDELINES.md#ui-components-style-guidelines)

PR #5252, opened 2022-08-16
Fixed logging non-fatal exceptions

grzesiek2010:
The main problem was that our naming was wrong we had `logFatal` which in fact logged non fatal exceptions and `logNonFatal` which logged simple messages

As a result we didn't log non fatal exceptions correctly. For example in https://github.com/getodk/collect/pull/5247/ we added:
`Timber.e("getCurrentIndex() firing with null getFormController()");`
what should result in displaying non fatal reports in crashlytics. Instead it's just a simple log we can see in other crashes but not  a separate report that we could track and analyze as we wanted.

According to https://github.com/getodk/collect/pull/5247/ non fatal exceptions should be logged calling `recordException()` method.

It's just a fix I've implemented according to https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=android

It doesn't require testing.

No.

No.


seadowg (approved):
Nice catch!

PR #5254, opened 2022-08-17
Fixed starting form filling twice

### LINKED ISSUE #5246: Ignoring or saving changes in a form entered by a created shortcut opens a new form if „Don’t keep activities” is on

Collect v2022.1.2, Android probably all

**Problem description**

When in a device settings „Don’t keep activities” is turned on, ignoring or saving the changes in a shortcut for a form from ODK Collect results in unexpected behavior. First, the form closes and immediately after that a new form is opened. 

https://user-images.githubusercontent.com/56479916/184151713-88fb47a1-f9d0-4689-8148-37417b3781cf.mp4

1. In device setting turn on „Don’t keep activities”.
2. Create a shortcut for any form from ODK Collect.
3. Go to the form by clicking the shortcut icon.
4. Click device back button.
5. Click „Ignore Changes” or „Save Changes”.

**Expected behavior**

The form should be closed.

**Other information** 

If the user chose „Save changes” the data is not lost.
Collect v2021.2.0 works as expected.

grzesiek2010:
Closes #5246 

I've verified the fix manually and added automated tests.

This is a change in the code that is responsible for staring forms not directly from the list of forms in the app but from shortcuts or external apps (see: https://github.com/grzesiek2010/collectTester)

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
+++ b/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
@@ -5,13 +5,10 @@ import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.result.contract.ActivityResultContracts
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
-import org.odk.collect.analytics.Analytics
 import org.odk.collect.android.R
 import org.odk.collect.android.activities.FormEntryActivity
-import org.odk.collect.android.analytics.AnalyticsEvents
 import org.odk.collect.android.injection.DaggerUtils
 import org.odk.collect.android.projects.CurrentProjectProvider
-import org.odk.collect.android.utilities.ApplicationConstants
 import org.odk.collect.projects.ProjectsRepository
 import javax.inject.Inject
 
@@ -54,8 +51,6 @@ class FormUriActivity : ComponentActivity() {
 
             if (projectId == currentProjectProvider.getCurrentProject().uuid) {
                 if (!formFillingAlreadyStarted) {
-                    logAnalytics(uriProjectId)
-
                     formFillingAlreadyStarted = true
                     openForm.launch(
                         Intent(this, FormEntryActivity::class.java).also {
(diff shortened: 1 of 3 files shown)


collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
```diff
@@ -49,13 +55,16 @@ class FormUriActivity : ComponentActivity() {
             logAnalytics(uriProjectId)
```
seadowg: Do you think we should move the analytics call so that it doesn't fire twice if the activity is recreated? Not a huge deal, but it does taint the numbers.

seadowg (approved):

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
grzesiek2010: Yeah why not. Done.

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
```diff
@@ -46,16 +52,19 @@ class FormUriActivity : ComponentActivity() {
...
             if (projectId == currentProjectProvider.getCurrentProject().uuid) {
-                openForm.launch(
-                    Intent(this, FormEntryActivity::class.java).also {
-                        it.action = intent.action
-                        it.data = uri
-                        intent.extras?.let { sourceExtras -> it.putExtras(sourceExtras) }
-                    },
-                )
+                if (!formFillingAlreadyStarted) {
+                    logAnalytics(uriProjectId)
```
seadowg: I don't think moving this here is correct, `logAnalytics` was being called in the case where `projectId` did not equal `currentProjectProvider.getCurrentProject().uuid` before. It might be that you just want to split out the different events in `logAnalytics` into individual calls.

seadowg (changes requested):

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
grzesiek2010: Yes but if I put it in that else block I would also need to make sure to log it only once not like now. Plus i'm wondering if we need it there... in that case we just display a dialog and that's it maybe we should have those logs only for cases where form filling is really started?

dbemke:
**Tested with Success!**

Verified on Androids: 8.1, 11 (unable to create a shortcut on Xiaomi Redmi 9T Android 10)

**Verified cases:**

- issue #5246 is no longer reproducing
- saving and ignoring changes in forms from shortcuts
- filling forms from shortcuts from different projects saving/finalizing/ignoring changes
- editing forms saved from shortcuts and saving/ignoring changes
- opening forms form shortcuts when ODK Collect is switched to different project
- resuming a form from a shortcut after closing it by killing the app
- changing project settings while filling a form from a shortcut and going back to the form
- starting a form from an external app: ODK Collect Intents Tester (checked also on Android 10)
- don’t keep activities enabled/disabled

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
seadowg: Yeah actually, now I think about it, I'm not sure if I particularly care about those analytics events any longer. All three events have fired 1000s of times, so there's no "dead code" and I don't think we're close to being able (or needing) to deprecate form URI actions without project IDs. Can we just kill analytics here rather than reworking the code for it @lognaturel?

srujner:
**Tested with Success!**

Verified on Android: 13 (On Emulator)

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
grzesiek2010: @lognaturel said that we can get rid of it so I will do that.

seadowg (approved):

PR #5255, opened 2022-08-17
More logs to determine the root cause of null formController

grzesiek2010:
I've just reviewed implemented changes.

Nothing to discuss here.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -183,6 +183,7 @@ import org.odk.collect.permissions.PermissionListener;
 import org.odk.collect.permissions.PermissionsChecker;
 import org.odk.collect.settings.keys.ProjectKeys;
 import org.odk.collect.settings.keys.ProtectedProjectKeys;
+import org.odk.collect.shared.strings.Md5;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


collect_app/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2105,6 +2105,8 @@ public void loadingComplete(FormLoaderTask task, FormDef formDef, String warning
         final FormController formController = task.getFormController();
 
         if (formController != null) {
+            Timber.w("Form %s started", formController.getFormTitle());
```
lognaturel: Can we hash this please? I know this kind of info sometimes shows up in crash logs but when we're explicitly logging I think we should hash.

grzesiek2010: Ok we can do that.

collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt
```diff
@@ -49,6 +50,7 @@ class FormUriActivity : ComponentActivity() {
             logAnalytics(uriProjectId)
 
             if (projectId == currentProjectProvider.getCurrentProject().uuid) {
+                Timber.w("Form started via VIEW/EDIT actions")
```
seadowg: Are we not already logging this [through an analytics event](https://github.com/getodk/collect/blob/55c00b8ce6f2a08d94a84713362f56be4ad0a3ce/collect_app/src/main/java/org/odk/collect/android/external/FormUriActivity.kt#L49)? @lognaturel pointed that out to me the other day.

grzesiek2010: Hmm let me check it...

grzesiek2010: Right that event should be visible so we do not need a new one. Fixed.

seadowg (approved):

seadowg:
@lognaturel merging this now, but take a look to check you're that ok with the logging.

lognaturel:
Works for me! Will be interesting.

PR #5259, opened 2022-08-23
Prevent `ConcurrentModificationException` when logging or flushing the `AuditEventLogger`

seadowg:
Should hopefully resolve [this crash](https://console.firebase.google.com/u/0/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/eebd7555f890d15034057bc131e622dd?time=last-seven-days&types=crash&sessionEventKey=630465E800C70001216501F62C12F425_1713447215830504361).

Added assertions in code to prevent us from logging or flushing with the audit log from background threads.

We could potentially have made `AuditEventLogger` "thread safe" by queuing/locking changes to the log itself, but that felt like a much more involved change. It looks like we introduced the only case of logging/flushing on a background thread in v2022.3.0, so making sure we went back to only calling `AuditEventLogger` from the UI thread made sense to me. When reviewing, it'd be good to verify that I'm right that we don't modify `AuditEventLogger` from a background thread anywhere else.

It'll be good to verify filling and editing forms with auditing both enabled and disabled.

### DIFF
--- a/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
+++ b/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
@@ -162,9 +162,11 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
         isLoading.setValue(true);
 
         scheduler.immediate((Supplier<Boolean>) () -> {
-            return updateAnswersForScreen(answers, evaluateConstraints);
+            return saveScreenAnswersToFormController(answers, evaluateConstraints);
         }, updateSuccess -> {
             isLoading.setValue(false);
+
+            formController.getAuditEventLogger().flush();
 
             if (updateSuccess) {
                 try {
@@ -183,9 +185,11 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
         isLoading.setValue(true);
 
         scheduler.immediate((Supplier<Boolean>) () -> {
-            return updateAnswersForScreen(answers);
+            return saveScreenAnswersToFormController(answers, false);
         }, updateSuccess -> {
             isLoading.setValue(false);
+
+            formController.getAuditEventLogger().flush();
 
             if (updateSuccess) {
                 try {
@@ -206,6 +210,13 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
     }
 
     public boolean updateAnswersForScreen(HashMap<FormIndex, IAnswerData> answers, Boolean evaluateConstraints) {
+        boolean success = saveScreenAnswersToFormController(answers, evaluateConstraints);
+        formController.getAuditEventLogger().flush();
+
+        return success;
+    }
+
+    private boolean saveScreenAnswersToFormController(HashMap<FormIndex, IAnswerData> answers, Boolean evaluateConstraints) {
         if (formController == null) {
             return false;
         }
@@ -221,7 +232,6 @@ public class FormEntryViewModel extends ViewModel implements RequiresFormControl
             return false;
         }
 
-        formController.getAuditEventLogger().flush();
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
```diff
@@ -162,10 +162,12 @@ public void moveForward(HashMap<FormIndex, IAnswerData> answers, Boolean evaluat
         isLoading.setValue(true);
 
         scheduler.immediate((Supplier<Boolean>) () -> {
-            return updateAnswersForScreen(answers, evaluateConstraints);
+            return saveScreenAnswersToFormController(answers, evaluateConstraints);
         }, updateSuccess -> {
             isLoading.setValue(false);
 
+            formController.getAuditEventLogger().flush();
```
seadowg, line 169: Move these `flush` calls to the `foreground` block so that they run on the UI thread (preventing concurrent `logEvent`/`flush` calls).

collect_app/src/main/java/org/odk/collect/android/formentry/audit/AuditEventLogger.java
```diff
@@ -57,6 +58,8 @@ public void logEvent(AuditEvent.AuditEventType eventType, boolean writeImmediate
      */
     public void logEvent(AuditEvent.AuditEventType eventType, FormIndex formIndex,
                          boolean writeImmediatelyToDisk, String questionAnswer, long currentTime, String changeReason) {
+        checkAndroidUIThread();
```
seadowg, line 61: This should prevent us from making the same mistake again. It might feel a little risking as it could cause a production crash if we end up calling `logEvent` or `flush` from a background thread in a flow that our automated or regression tests don't cover. However, I think in that case, it'll be faster and easier to see what's happening than have to wait for a crash/bug that is created by the concurrent access.

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
grzesiek2010: How did that happen? Were you able to reproduce? I;m thinking whether we should improve AuditEventLogger to protect it from cases like that instead.

seadowg: Wasn't able to reproduce, but I could see that we'd moved a `flush` call onto the background in v2022.3.0 (which could cause the exception). I do agree we should make it "thread safe" eventually (I comment about that [here](https://github.com/getodk/collect/pull/5259#discussion_r952459314)), but I think this is the smaller change we can make to prevent the crash.

grzesiek2010: For a hotfix ok but please file a separate issue because it's not a solution I love.

grzesiek2010 (approved):

collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryViewModel.java
seadowg: Sounds good, will do.

dbemke:
**Tested with Success!**

Verified on Androids: 8.1, 11

**Verified cases:**

    • filling forms with audit 
    • audit log order of: start form, questions, location tracking, resume form, jump, save, exit, finalize 
    • saving and editing different forms: AuditTestLocationBackground, AuditTest, Track Changes Reason, trackingChangesAndLocationTestForm, Identify User, test audit repeats, Nested repeats for audit 
    • regression checks of forms 
    • resuming from a save point 
    • location and audio permission enabled, disabled 
    • moving backwards enabled, disabled
    • don’t keep activities enabled/disabled

srujner:
**Tested with Success!**

Verified on Androids: 5.1, 10 and 12

PR #5260, opened 2022-08-23
Fixed logging errors

grzesiek2010:
I've reviewed implemented changes.

After merging https://github.com/getodk/collect/pull/5252 we ended up with many different errors logged in one category https://console.firebase.google.com/u/1/project/api-project-322300403941/crashlytics/app/android:org.odk.collect.android/issues/f940b5ab1c52a2583207fb130644f9ce?time=last-seven-days&versions=v2022.3.3%20(4490)&sessionEventKey=6304C495037400011B4C6EF52F2D7CBE_1713485720918396517

The problem was that if we log `Tiber.e` with just a message we need to create `throwable` because it's required in `crashlytics.recordException` (such a `throwable` is created in [CrashReportingTree](https://github.com/getodk/collect/pull/5252/files#diff-02662bcce6ec95546375a46512665a81ad590bebe3889f99c281098797ff2e16R13) for all types of messages) and we end up with one category of nonfatals.

To solve the issue I updated all `Timber.e` wrapping messages with `Error`.

It doesn't require testing.

No.

No.

### DIFF
--- a/collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
+++ b/collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
@@ -14,10 +14,10 @@ class RetryOnDeviceErrorRule : TestRule {
                     base.evaluate()
                 } catch (e: Exception) {
                     if (e is PerformException) {
-                        Timber.e("RetryOnDeviceErrorRule: Retrying due to PerformException!")
+                        Timber.e(Error("RetryOnDeviceErrorRule: Retrying due to PerformException!"))
                         base.evaluate()
                     } else if (e::class.simpleName == "RootViewWithoutFocusException") {
-                        Timber.e("RetryOnDeviceErrorRule: Retrying due to RootViewWithoutFocusException!")
+                        Timber.e(Error("RetryOnDeviceErrorRule: Retrying due to RootViewWithoutFocusException!"))
                         base.evaluate()
                     } else {
                         throw e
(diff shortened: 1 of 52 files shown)


collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
```diff
@@ -14,10 +14,10 @@ class RetryOnDeviceErrorRule : TestRule {
                     base.evaluate()
                 } catch (e: Exception) {
                     if (e is PerformException) {
-                        Timber.e("RetryOnDeviceErrorRule: Retrying due to PerformException!")
+                        Timber.e(Error("RetryOnDeviceErrorRule: Retrying due to PerformException!"))
```
seadowg: These should all be warnings instead of errors as they are test logging. Let's change them while we're here.

seadowg (changes requested):
There are a few places where you switch from Timber's formatting to using append (`+`) to build Strings. This could create a potential `NullPointerException` where one wouldn't have been possible before (I believe `String.format` can safely deal with `null` args). I think we should use `String.format` instead of `+` throughout.

Also, before merging, it would be good to verify that this gives us the output we want in Crashlytics (seeing as we've had to change this a few times to get it right). We could create a [test log](https://firebase.google.com/docs/crashlytics/get-started?platform=android#force-test-crash) to do that I think.

collect_app/src/androidTest/java/org/odk/collect/android/support/rules/RetryOnDeviceErrorRule.kt
grzesiek2010: Fixed.

grzesiek2010:
> There are a few places where you switch from Timber's formatting to using append (+) to build Strings. This could create a potential NullPointerException where one wouldn't have been possible before (I believe String.format can safely deal with null args). I think we should use String.format instead of + throughout.

The Timber's formatting used before is not null safe so there is no difference now.

> Also, before merging, it would be good to verify that this gives us the output we want in Crashlytics (seeing as we've had to change this a few times to get it right). We could create a [test log](https://firebase.google.com/docs/crashlytics/get-started?platform=android#force-test-crash) to do that I think.

I've tested it and it looks fine.

seadowg (approved):

